PR #1, opened 2017-03-16
Company add step 2

ztolley:
Created the step 2 for add uk company as best I can, but this will need changes to search and discussion around the quality of data and how that will have an adverse affect.


ztolley:
Added test coverage for company add steps 1 and 2

PR #2, opened 2017-03-20
Add ltd company

ztolley:
Refactored company repository to use generators
Added add ltd company
Removed set CH defaults and instead put the defaults in the form to make saving a company more generic.
Fix search box side effect from moving query to common

### DIFF
--- a/src/controllers/companycontroller.js
+++ b/src/controllers/companycontroller.js
@@ -2,13 +2,13 @@
 const express = require('express')
 const winston = require('winston')
 const companyRepository = require('../repositorys/companyrepository')
-const companyService = require('../services/companyservice')
 const metadataRepository = require('../repositorys/metadatarepository')
+const companyService = require('../services/companyservice')
 const companyFormattingService = require('../services/companyformattingservice')
 const { companyDetailLabels, chDetailLabels, companyTableHeadings, hqLabels } = require('../labels/companylabels')
 const formatDate = require('../lib/date').formatDate
-const router = express.Router()
 const { isBlank } = require('../lib/controllerutils')
+const router = express.Router()
 
 function getCommon (req, res, next) {
   const id = req.params.sourceId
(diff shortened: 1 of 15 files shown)


src/controllers/companycontroller.js
```diff
@@ -71,51 +54,67 @@ function getDetails (req, res, next) {
   }
 }
 
+// Figure out the business type using either the existing company business type
+// the companies house type or type sent in the request for a new company
+function calculateBusinessType (company, req) {
+  if (company && company.business_type && company.business_type !== null && typeof company.business_type === 'object') {
```
jim68000: Ouch. It pains me that JS can't safely handle undefs like this. Would a try/catch be better? I know it has performance implications.

src/controllers/companycontroller.js
```diff
@@ -71,51 +54,67 @@ function getDetails (req, res, next) {
...
+    return company.business_type
+  } else if (company && company.companies_house_data) {
+    for (const businessType of metadataRepository.businessTypeOptions) {
+      if (businessType.name.toLowerCase() === company.companies_house_data.company_category.toLowerCase()) {
+        return businessType
+      }
+    }
+  } else if (req.query && req.query.business_type) {
```
jim68000: Do we need a default action here?

src/controllers/companycontroller.js
```diff
@@ -71,51 +54,67 @@ function getDetails (req, res, next) {
...
+    res.locals.chDetailLabels = chDetailLabels
+    res.locals.chDetailDisplayOrder = ['name', 'company_number', 'registered_address', 'business_type']
+  }
+
+  const businessType = res.locals.business_type = calculateBusinessType(company, req)
+  const ukBased = res.locals.uk_based = (company.companies_house_data || company.uk_based || (req.query && req.query.country && req.query.country === 'uk'))
 
   let template
-  if (businessType === 'ltd') {
+  if (businessType.name.toLowerCase() === 'private limited company' || businessType.name.toLowerCase() === 'public limited company') {
```
jim68000: Shouldn't you use calculateBusinessType() here?

src/controllers/companycontroller.js
```diff
@@ -124,12 +123,23 @@ function postDetails (req, res, next) {
       winston.debug(error)
       if (error.errors) {
         winston.debug(error)
-        res.locals.errors = controllerUtils.transformErrors(error.errors)
-        res.locals.company = req.body
-        return editDetails(req, res)
+        // Leeloo has inconsistant structure to return errors.
+        // Get the errors and then re-render the edit page.
+        if (error.errors.errors) {
+          res.locals.errors = error.errors.errors
```
jim68000: !!!

ztolley: Yeah I know...

src/controllers/companycontroller.js
ztolley: We are using the calculated business type

src/controllers/companycontroller.js
ztolley: Defensive programming, plus sometimes we get a key with a null value. try catch wont work here.

PR #5, opened 2017-03-30
Updated company contacts to updated design

ztolley:
Refactored the company contact list to own controller to keep it simple
Introduced new style for card lists.

screen shot 2017-03-30 at 17 44 11


ztolley:
Design changed at last moment..

PR #9, opened 2017-04-04
Basic interaction list

ztolley:
Updated basic list of interactions using new card design.


test/controllers/companyinteractioncontroller.test.js
```diff
@@ -0,0 +1,217 @@
...
+        subject: 'Test subject',
+        date: '23 February 2017',
+        advisor: 'Fred Smith'
+      }]
+
+      addInteractionUrl = '/interaction/add?company=1234'
+    })
+
+    it('should render a list of interactions', function (done) {
+      renderContent({interactions, addInteractionUrl, company})
```
reupen: Mocha will handle promises automatically if you remove the `done` parameter and return the promise, e.g.:
```
    it('should render a list of interactions', function () {
      return renderContent({interactions, addInteractionUrl, company})
      .then((document) => {
        expect(document.getElementById('interaction-list')).to.not.be.null
      })
    })
```

src/controllers/companyinteractioncontroller.js
```diff
@@ -0,0 +1,39 @@
...
+
+  // build the data for the interaction list.
+  res.locals.interactions = company.interactions.map((interaction) => {
+    const type = (interaction.interaction_type.name === 'Service delivery') ? 'servicedelivery' : 'interaction'
+
+    return {
+      url: `/${type}/${interaction.id}/details`,
+      type: interaction.interaction_type.name,
+      subject: interaction.subject,
+      date: moment(interaction.date).format('DD MMMM YYYY'),
```
reupen: Would be nice to put the date format in config/global constants. Also the format should be `D MMMM YYYY` if writing out in full (is the style in the the GDS service manual).

src/controllers/companyinteractioncontroller.js
```diff
@@ -0,0 +1,39 @@
...
+      advisor: `${interaction.dit_advisor.first_name} ${interaction.dit_advisor.last_name}`
+    }
+  })
+
+  // Only allow a link to add an interaction if the company has contacts
+  if (company.contacts && company.contacts.length > 0) {
+    res.locals.addInteractionUrl = `/interaction/add?company=${company.id}`
```
reupen: Would be nice to have a message or something to say to add a contact in this case.

src/views/company/interactions.html
```diff
@@ -2,10 +2,32 @@
...
+  <li class="lined-list__item">
+    <div class="card interaction">
+      <a class="card__metadata" href="{{interaction.url}}">{{interaction.subject}}</a>
+      <div class="split-data">
+        <div>
+          <div class="card__metadata">{{interaction.type}}</div>
+          <div class="card__metadata">{{interaction.date}}</div>
+        </div>
+        <div>
+          <div class="card__metadata">{{interaction.advisor}}</div>
```
reupen: Not critical for now, but in the design there is a fourth item of text here, do we know what it is?

ztolley: The designs will change. The design I had showed a field that does not exist

src/controllers/companyinteractioncontroller.js
ztolley: This is temporary. There's a design to follow this that has that in it. I also need to handle other situations such as when there is no company details too

PR #12, opened 2017-04-07
Feature/dh 81 interaction detail

ztolley:


src/lib/textformatting.js
```diff
@@ -29,4 +28,29 @@ function newlineToBr (str) {
...
+    name += `${contact.first_name} `
+  }
+  if (contact.last_name) {
+    name += contact.last_name
+  }
+  return name.trim()
+}
+
+function getContactLink (object) {
+  if (object.contact && object.contact !== null) {
```
reupen: `null` is falsey, was there a different intention here?

src/repositorys/servicedeliveryrepository.js
```diff
@@ -11,29 +11,29 @@ function saveServiceDelivery (token, serviceDelivery) {
     options.url = `${config.apiRoot}/v2/service-delivery/${serviceDelivery.id}/`
     options.method = 'PUT'
   } else {
-    options.url = `${config.apiRoot}/v2/service-delivery/`
+    options.url = `${config.apiRoot}/v2/service_delivery/`
```
reupen: This has been renamed back now, right?

src/services/interactionformattingservice.js
```diff
@@ -0,0 +1,35 @@
+const {formatLongDate} = require('../lib/date')
+const {newlineToBr, getContactLink, getPropertyName} = require('../lib/textformatting')
+
+function getDisplayInteraction (interaction) {
+  const result = {
+    company: `<a href="/company/company_company/${interaction.company.id}/details">${interaction.company.name}</a>`,
```
reupen: Do we want HTML here?

src/views/interaction/interaction-details.html
```diff
@@ -2,7 +2,7 @@
 {% import "macros/trade.html" as trade %}
 
 {% block main %}
-  <h1 class="heading-xlarge record-title">Interaction details</h1>
-  {{ trade.keyvaluetable(displayValues, stripey=true, labels=labels) }}
-  <a class="button" href="/interaction/{{interaction.id}}/edit">Edit interaction details</a>
+  <h1 class="heading-xlarge record-title">Interaction</h1>
+  {{ trade.keyvaluetable(interactionDetails, stripey=true, labels=interactionLabels, keyorder=interactionDisplayOrder) }}
```
reupen: HTML isn't escaped in the table which causes problems when certain characters are in the data (e.g. `<` in subject or notes).

src/services/interactionformattingservice.js
ztolley: Yes. For keyvalue tables, that have a link in them, the only way to do that right now is like this. it will change.

src/lib/textformatting.js
ztolley: I had weird problems with things that had properties but set to null breaking stuff. Now I have tests to check those situations I can start to clean it up. starting with this

src/views/interaction/interaction-details.html
ztolley: I tried this real quick and it cause even more problems.  I'll merge this and raise a new ticket just for the notes field.

PR #13, opened 2017-04-10
Change address formatter to return null for empty address

ztolley:
Previously an empty address returned an empty string, which would be shown on the screen. This was inconsistent with the decision to not show empty fields.

### DIFF
--- a/src/services/companyformattingservice.js
+++ b/src/services/companyformattingservice.js
@@ -1,8 +1,9 @@
 const { companyDetailsLabels, chDetailsLabels, hqLabels } = require('../labels/companylabels')
-const getFormattedAddress = require('../lib/address').getFormattedAddress
-const {titleCase} = require('../lib/textformatting')
-const {formatLongDate} = require('../lib/date')
+const { getFormattedAddress } = require('../lib/address')
+const { titleCase } = require('../lib/textformatting')
+const { formatLongDate } = require('../lib/date')
 const { getPrimarySectorName } = require('../lib/transformsectors')
+
 
 const companyDetailsDisplayOrder = Object.keys(companyDetailsLabels)
 const chDetailsDisplayOrder = Object.keys(chDetailsLabels)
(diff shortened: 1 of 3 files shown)


test/lib/address.test.js
```diff
@@ -0,0 +1,48 @@
+/* globals expect: true, describe: true, it: true, beforeEach: true */
+const address = require('../../src/lib/address')
+
+describe('Address formatter', function () {
+  it('should format an address when', function () {
```
reupen: Something missed off after 'when'?

test/lib/address.test.js
```diff
@@ -0,0 +1,48 @@
...
+      trading_address_4: '',
+      trading_address_town: 'Big Town',
+      trading_address_county: 'Large County',
+      trading_address_postcode: 'LL1 1LL'
+    }
+
+    const actual = address.getFormattedAddress(source, 'trading')
+    expect(actual).equal('10 The Street, Warble, Big Town, Large County, LL1 1LL, United Kingdom')
+  })
+  it('should return null when the address is emptyu', function () {
```
reupen: Typo in 'emptyu'

src/services/companyformattingservice.js
```diff
@@ -95,18 +96,8 @@ function parseRelatedData (companies) {
...
-      address += titleCase(`${company[`${key}_address_county`]}, `)
-    }
-    if (company[`${key}_address_country`] && company[`${key}_address_country`].name && company[`${key}_address_country`].name.length > 0) {
-      address += titleCase(company[`${key}_address_country`].name)
-    } else if (address.length > 0) {
-      address += 'United Kingdom'
-    }
     return {
       name: `<a href="/company/company_company/${company.id}">${company.alias || company.name}</a>`,
       address
```
reupen: Is this right?

ztolley: Yes. Ugly but yes. If there is an address, but no country provided, add United Kingdom to what we display.

PR #27, opened 2017-04-21
Add/Edit interaction

ztolley:
Added the code and a multitude of tests needed for editing interactions and fixed a few links that didn't work.


src/controllers/interactioneditcontroller.js
```diff
@@ -0,0 +1,103 @@
...
+    try {
+      const token = req.session.token
+      const default_dit_advisor = req.session.user
+
+      // Work out what to use for the form data
+      // This can either be data recently posted, to be re-rendered with errors
+      // or an interaction that the user wishes to edit
+      // or a new interaction for a company or contact
+      if (typeof req.body === 'object' && Object.keys(req.body).length > 0) {
+        res.locals.formData = req.body
```
jim68000: This section of code is pretty dense and very repetitive: is there not an argument for breaking out those if/elses into function calls?

src/controllers/interactioneditcontroller.js
```diff
@@ -0,0 +1,103 @@
...
+        res.locals.interaction = yield interactionDataService.createBlankInteractionForContact(token, default_dit_advisor, req.query.interaction_type, req.query.contact)
+        res.locals.formData = interactionFormService.getInteractionAsFormData(res.locals.interaction)
+        res.locals.company = res.locals.interaction.company
+        res.locals.interaction_type = res.locals.interaction.interaction_type
+      } else {
+        return next('Unable to edit interaction')
+      }
+
+      if (req.query && req.query.company) {
+        res.locals.backUrl = `/company/company_company/${req.query.company}/interactions`
```
jim68000: This kind of routing probably shouldn't be done at this level - is this not the responsibility of the caller to figure out?

src/controllers/interactioneditcontroller.js
```diff
@@ -0,0 +1,103 @@
...
+        editDetails(req, res, next)
+      } else {
+        next(errors)
+      }
+    }
+  })
+}
+
+router.get(['/interaction/add/', '/interaction/:interactionId/edit', '/interaction/edit/'], editDetails)
+router.post(['/interaction/add/', '/interaction/:interactionId/edit', '/interaction/edit/'], postDetails)
```
jim68000: I think these are the cause of the issues I've raised above: I'd break these into different functions.

src/views/interaction/interaction-edit.html
```diff
@@ -13,52 +14,52 @@
...
-    {{ trade.date('date', label=labels.date, value=interaction.date ) }}
+    {{ trade.date('date', label=labels.date, value=formData.date ) }}
 
     {{ trade.autocomplete('dit_advisor',
       label=labels.dit_advisor,
-      value=interaction.dit_advisor.id,
-      displayvalue=interaction.dit_advisor.name,
+      value=formData.dit_advisor,
+      displayvalue=dit_advisor.name,
       hint="Please start typing to search for an advisor",
```
jim68000: According to Mills our standard spelling should be adviser. (Both are valid, advisor is possibly a bit more american). Which is awkward because we use advisor through the code in variable names etc

PR #28, opened 2017-05-02
Data Hub not datahub

jim68000:


reupen:
Could also update the description in package.json...

PR #29, opened 2017-05-02
Refactor company

ztolley:
Refactored the company add/edit/view
Benefits include simplified controllers that are easier to understand, less complex logic and increased test coverage for more scenarios. Also found bugs and fixed them along the way.

Also introduced yarn package management support to allow building with lock down package versions.


ztolley:
I know this is a massive pull request. I couldn't break it down into smaller requests without breaking the app

PR #39, opened 2017-05-03
Restored create SD page, SD details page and added new sector script

jim68000:
~Does not save yet - looks like Django is expecting an event even though they are optional~

Now saves and displays


reupen:
Getting this error when trying to create a service delivery for a contact:

```
TypeError: Cannot read property 'company' of undefined
    at data-hub-fe-beta2/src/controllers/servicedeliverycontroller.js:53:55
```

jim68000:
s/contactId/contact/g did the trick for that one @reupen

src/controllers/servicedeliverycontroller.js
```diff
@@ -18,7 +18,13 @@ function getCommon (req, res, next) {
   Q.spawn(function * () {
     try {
       const token = req.session.token
-      res.locals.serviceDelivery = yield serviceDeliveryService.getHydratedServiceDelivery(token, req.params.serviceDeliveryId)
+      // if we are creating a new service delivery then the id comes through as edit
+      // @TODO make the routes a bit more sensible
```
ztolley: Is there a card for this?

src/controllers/servicedeliverycontroller.js
```diff
@@ -77,14 +81,15 @@ function postServiceDeliveryEdit (req, res, next) {
       delete req.body.date_month
       delete req.body.date_day
 
-      req.body = nullEmptyFields(req.body)
+      // v2 endpoint rejects nulls
+      req.body = deleteNulls(nullEmptyFields(req.body))
```
ztolley: What happens if we want to clear the value in a field? Say if you want to to so there is no longer any notes (though I know technically every field is needed)

src/lib/propertyhelpers.js
```diff
@@ -18,6 +18,18 @@ function nullEmptyFields (data) {
   return cleanedObject
 }
 
+function deleteNulls (data) {
```
ztolley: See previous comment. Deleting null fields may in some situations result in an edit not being committed as the user intended.

test/lib/controllerutils.test.js
```diff
@@ -24,3 +24,100 @@ describe('isBlank', function () {
...
+    const source = [
+      {
+        'detail': 'This combination of service and service provider does not exist.',
+        'source': {
+          'pointer': '/data/relationships/service'
+        }
+      }
+    ]
+    const actual = transformV2Errors(source)
+    expect(actual.Alert).to.exist
```
ztolley: Why is alert Capital?

test/lib/controllerutils.test.js
```diff
@@ -24,3 +24,100 @@ describe('isBlank', function () {
...
+      {
+        'detail': "{'data': {'type': 'Country'}} has no key id",
+        'source': {
+          'pointer': '/data/relationships/country_of_interest'
+        }
+      }
+    ]
+    const actual = transformV2Errors(source)
+    expect((Object.keys(actual)).length).to.equal(9)
+    expect(actual.subject).to.be.defined
```
ztolley: I use 

expect(actual).to.have.property('subject') 

but I'm guessing thats just my style.

src/controllers/servicedeliverycontroller.js
jim68000: No, I'll add one to backlog as debt

src/controllers/servicedeliverycontroller.js
jim68000: That's not the way the API behaves. Subject, notes and date are mandatory fields, so the server rejects nulls for those, and if you try and send a null in the relationship fields it errors out.

jim68000:
Yeah, I realised what I actually need is integration tests so I can check that all the entrance points work with the pernicketty server

PR #47, opened 2017-05-09
gitignore addition

feedmypixel:
Add the intellij generated folder to `.gitignore`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,4 @@ reports
 .env
 npm*.log
 stats.json
+.idea


teneightfive:
This might be better place in the global gitignore of the developers machine as it's IDE specific? The project gitignore is probably better used for project specific files/folders that should be ignored from source like a temp build folder.

feedmypixel:
> This might be better place in the global gitignore of the developers machine as it's IDE specific? The project gitignore is probably better used for project specific files/folders that should be ignored from source like a temp build folder.

Good point. Done

PR #50, opened 2017-05-09
Feature/return of zendesk

jim68000:
Another commit to get the front end back to where it was before the last release.


src/views/support/bug.html
```diff
@@ -0,0 +1,62 @@
+{% extends "layouts/ukti.html" %}
+{% import "macros/trade.html" as trade %}
+
+{% block main %}
+
+  <h1 class="heading-xlarge">Report a problem or feedback</h1>
```
teneightfive: "Report a problem or leave feedback" instead?

src/views/support/bug.html
```diff
@@ -0,0 +1,62 @@
...
+          />
+          Feedback
+        </label>
+      </fieldset>
+    </div>
+
+    <input type="hidden" name="_csrf_token" value="{{data.csrfToken}}"/>
+    {{ trade.save(backUrl=referer, buttonText='Send') }}
+  </form>
+<script src="/javascripts/radio.bundle.js?{{ startTime }}"></script>
```
teneightfive: Might be worth adding a ticket to improve this method of cache busting. We might be forcing them to re-download an asset even if it hasn't changed but the server has been restarted.

src/views/support/thank.html
```diff
@@ -0,0 +1,31 @@
+{% extends "layouts/ukti.html" %}
+
+{% block main %}
+{% if not user or not user.name %}<a href="/">Home</a>{% endif %}
```
teneightfive: Do we need this? If we do, should it be higher up at a parent template level?

src/views/support/bug.html
```diff
@@ -0,0 +1,62 @@
...
+    </div>
+  {% endif %}
+
+  {{ trade.errorPanel( messages.formErrors ) }}
+
+  <form action="{{ action }}" method="post">
+    {{ trade.textbox('title', label="Title", hint="A brief summary of your feedback/problem", error=errors.title, autofocus=true, value=data.title) }}
+    {{ trade.textarea('description', label="Description", hint="If you were having a problem, explain what you did, what happened and what you expected to happen. If you want to provide feedback or a suggestion, describe it here.", error=errors.steps, autofocus=true, value=data.steps) }}
+    {{ trade.textbox('email', label="Email", hint="If you wish to be contacted, please provide your email address",error=errors.name, autofocus=true, value=data.name) }}
+    {{ trade.textbox('browser', label="Web browser", hint="The name of the web browser and version you were using", error=errors.browser, autofocus=true, value=data.browser) }}
```
teneightfive: Do we want to expose this field to the user or just capture it in a hidden field?

src/views/support/thank.html
```diff
@@ -0,0 +1,31 @@
...
+    {% endfor %}
+  </p>
+</div>
+{% endif %}
+
+{% if messages.success.length %}
+<div class="flash-message success-message">
+  <p class="flash-message-content">
+    {% for message in messages.success %}
+    {{ message }}<br>
```
teneightfive: Take a look at nunjucks [joiner](https://mozilla.github.io/nunjucks/templating.html#joiner-separator) for appending something within a loop.

src/views/support/bug.html
ztolley: The idea is to expose it with a default value and let them override it if this isn't the machine that had the problem

src/views/support/thank.html
ztolley: Good to know, not seen that one before.

src/views/support/bug.html
jim68000: That's an optimisation for later. I'll add something to the backlog.

src/views/support/thank.html
jim68000: That's a question for @ztolley really - I'm just copying and pasting from the old repo

src/views/support/thank.html
jim68000: Nice

src/views/support/bug.html
teneightfive: Wonder if that just confuses the interface a bit. Most people will have no idea what version of a browser or operating system they're on. If it helps us to understand a bug then it's useful to capture the useragent in a hidden field but it sounds like catering for a real edge case where somebody is recording a bug on a machine they didn't experience it on by displaying this field. 

Maybe focus on the majority of cases initially and look to add it if it's something people feel they want to provide/edit manually?

src/views/support/thank.html
teneightfive: Yeah, it's really handy when looping over content and you don't want to add the joiner to the last item.

jim68000:
Any objections if I merge this as is?

teneightfive:
I would say do the nunjucks joiner change and remove the Home link from the template in this PR then add tickets for the browser version and optimisation.

I'd also suggest rebasing on top of develop to remove the fixups and the merge commits.

src/views/support/bug.html
feedmypixel: FYI I have a nice little handrolled script that can be used in `npm scripts` that adds a hash to the filename and then replaces those references wherever required

src/views/support/thank.html
jim68000: Tidied up the header - I remember why this was here but it's better dealt with in the UKTI file

src/views/support/bug.html
jim68000: I'm inclined to leave this, especially as we've had a couple of instances of the site not rendering at all on one browser, so I'd like to leave the user the option to type something in here

jim68000:
I've done the nunjucks joiner, the removal of the logged out Home from  and raised the cache ticket. 

https://uktrade.atlassian.net/browse/DH-212

Not going to remove the user agent for reasons outlined above

src/views/support/bug.html
jim68000: One of our common failure cases is that the site won't render on a given device, so the ability to demarcate that is required. It could be handled in the description, but I think this a better forcing mechanism. For production Data Hub this is overkill, but at the point we are in the project I'd rather expose some technical stuff for the better trained beta participants.

PR #46, opened 2017-05-08
README updates

teneightfive:
A few tweaks to the installation instructions and how the development environment can be run.

PR #51, opened 2017-05-09
Environment Variables used in docker compose

feedmypixel:
The current `docker-compose` file doesn't run the app due to the necessary environment variables not being exposed. This work:

- exposes just the required environment variables to the containers provided via `docker-compose` in the `environment` property
- documents how to send through other environment variables to `docker-compose` containers via the `env_file` property and the `.env` file
- explains that you don't need to have a separate `redis` running if you work via `docker-compose`

This will work well with the work in #46 once this is merged in

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,11 +46,14 @@ To start the server just:
 
     docker-compose up
 
-The server starts in developer mode, which means that when you make local changes it will auto-compile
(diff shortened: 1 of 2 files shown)


feedmypixel:
> Wait for Dom's README merge to be done then you may need to re-edit the README file again with your change.

Done

PR #58, opened 2017-05-10
Remove fixed height on page footer

tyom:
### Before
image

### After
image

### DIFF
--- a/src/sass/_shame.scss
+++ b/src/sass/_shame.scss
@@ -102,10 +102,6 @@ hr {
   display: none;
 }
 
-.footer-wrapper {
-  height: 100px;
-}
-
 #details {
   .table--striped {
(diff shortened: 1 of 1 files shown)


src/sass/_shame.scss
```diff
@@ -102,10 +102,6 @@ hr {
   display: none;
 }
 
-.footer-wrapper {
```
ztolley: I know this is a bug and drives me mad too, but I'd look for the root cause and find out why it works at http://trade-elements.herokuapp.com/, but not here.

Also we want to remove the crown from the footer and change it to just have the feedback link

tyom: Trade Elements doesn't have height set on `.footer-wrapper` as far as I can see.

PR #60, opened 2017-05-10
Feature/dh 131 archive company

ztolley:
Allow a user to archive or unarchive a company.
Had to slightly refactor company to get the csrf token into the page. 
Can be used without javascript but best with, as this controls hiding and showing elements.

archive


src/views/company/archive.html
```diff
@@ -1,14 +1,35 @@
...
+    <a href="/company/unarchive/{{company.id}}">Unarchive</a>
+  </p>
+{% elif company.id %}
+  <div class="archive-panel js-hidden">
+    <p class="hidden archive-panel__show-button" aria-hidden="true">
+      <a href="#">Archive company</a>
+    </p>
+    <form action="/company/archive/{{company.id}}" class="section archive-panel__form " method="POST">
+      <input type="hidden" name="_csrf_token" value="{{csrfToken}}">
+      <div class='form-group {% if query.archiveerror %}form-group-error{% endif %}' id="archived_reason-wrapper">
```
teneightfive: Maybe tidy up the consistency of attribute quotes. Currently using a combo of single and double. Would say double is better for html/nunjucks files.

src/views/company/archive.html
```diff
@@ -1,14 +1,35 @@
...
+        <label class='form-label-bold' For='archived_reason'>
```
teneightfive: Lowercase the `for` attribute

test/javascripts/archiveform.test.js
```diff
@@ -0,0 +1,136 @@
...
+    archiveFormControl = new ArchiveForm(wrapper)
+    form = document.querySelector('.archive-panel__form')
+    showButton = document.querySelector('.archive-panel__show-button')
+    otherWrapper = document.querySelector('#archived_reason_other-wrapper')
+    otherField = document.querySelector('#archived_reason_other')
+    dropdownField = document.querySelector('#archived_reason')
+  })
+
+  it('should hide the form', function () {
+    expect(domTokenToArray(form.classList)).to.include('hidden')
```
teneightfive: Would it be better to check if the element is visible or not instead of it including a classname?

src/views/company/archive.html
ztolley: missed that, it had been a cut and past exercise from some react code

test/javascripts/archiveform.test.js
ztolley: That doesn't work as we are not loading in the entire css stack, just testing how we manipulate the dom

teneightfive: Yes, fair point. I guess the JS isn't changing the actual properties it's just adding a class.

PR #61, opened 2017-05-11
Change flow when saving contacts 

ztolley:
Changed the flow so when an existing contact is saved it still goes back to the contact detail screen but when adding new contact it goes go the company contact list.

Also added flash messages so the user knows why they are on the screen they are taken to.

addcontact
editcontact


test/controllers/contacteditcontroller.test.js
```diff
@@ -11,6 +11,10 @@ describe('Contact controller, edit', function () {
   let saveContactFormStub
   let getContactAsFormDataStub
   let company
+  const next = function (error) {
+    console.log(error)
```
teneightfive: Does it need this console log?

ztolley: it helps when debugging tests. Seems to kick out more info than just mocha alone

tyom:
It might be cleaner to rebase on develop rather than merge it into your branch.

PR #62, opened 2017-05-11
Feature/dh 229 search results not displaying

feedmypixel:
search results from the add company were not displaying in the frontend this work:

- fixes the bug arounds assigning properties to `doc_type` in the requestBody which meant the results were not showing on the front end
- tidies up code that wasn't being used
- add nock for mocking api calls
- add tests for search functionality

## Of Note
There is work `//TODO` around the endpoint being used here for search details. This work just involves a bug fix. See Jira `DH-203`

addnewcompanysearch


feedmypixel:
> can you explain the use of nock to me and how it compares to proxyquire/sinon.

I haven't used proxyquire before, I have used sinon. nock is great for stubbing requests with mock data.

https://github.com/node-nock/nock
> Nock is an HTTP mocking and expectations library for Node.js
> Nock can be used to test modules that perform HTTP requests in isolation.

A quick look over proxyquire, it is a stubbing approach to functions and modules? and sinon is used for spies and stubs. Feels like only one of those is needed? Especially if your code is written in a way that you can easily stub work

So for example in the code in [companycontrollerch.test.js](https://github.com/uktrade/data-hub-fe-beta2/blob/develop/test/controllers/companycontrollerch.test.js#L46-L64)

It looks like you don't need proxyquire if the methods are exposed, you can just stub with sinon like:

```
const companyservice = require('../services/companyservice')
const getInflatedDitCompanyStub = sinon.stub(companyservice, 'getInflatedDitCompany', () => { company_number: '1234' })
```

One other thing here, It looks like you are not testing the internals of the method `getInflatedDitCompany` you are just stubbing a response when it is called. So if something changed internally in `getInflatedDitCompany` the tests wouldn't fail.

> If they are 2 approaches that do the same can we stick with proxyquire and sinon for consistancy.
Agreed we shouldn't have multiple things doing the same thing but unless I'm not understanding it they are doing different things.

test/services/searchservice.test.js
```diff
@@ -0,0 +1,78 @@
+/* globals expect: true, describe: true, it: true */
+const nock = require('nock')
+const config = require('../../src/config')
+const searchService = require('../../src/services/searchservice')
+const companyMockAPIResponse = require('../data/search-response-company')
+const companiesHousePrivateLtdMockAPIResponse = require('../data/search-response-companies-house-private-ltd')
+const companiesHousePublicLtdMockAPIResponse = require('../data/search-response-companies-house-public-ltd')
+const facets = {
```
tyom: Could we add a new line to separate requires from local definitions?

test/services/searchservice.test.js
```diff
@@ -0,0 +1,78 @@
...
+      searchService.search({
+        token: 'token',
+        term: 'testTerm',
+        filters: ['company_company', 'company_companieshousecompany']
+      })
+        .then((response) => {
+          expect(response).to.deep.equal(expectedResponse)
+        })
+    })
+    it('Should return expected format from search method for a companies house public ltd', function () {
```
tyom: I think it would help readability if we added new lines between `it()`s.

PR #67, opened 2017-05-12
Investment journey (start and create stages)

teneightfive:
This adds the start of the journey for adding a new investment
project.

The journey has a few different scenarios:

- Starting with a UK based company and searching for foreign source
  of equity
- Starting with a foreign company and choosing it as the equity source
- Starting with a foreign company and searching for a different equity
  source


src/views/investment/create.html
```diff
@@ -5,145 +5,36 @@
...
-    <h1 class="page-heading">Create an investment project</h1>
-
-    {% if messages.info.length %}
-      {% for message in messages.info %}
-        <p class="infostrip">{{message}}</p>
-      {% endfor %}
-    {% endif %}
-
-    {{ trade.errorPanel( errors ) }}
+  <h1 class="page-heading">Add new investment project</h1>
```
ztolley: what happened to the flash messages? There is a general fragment you can include to display them. messages.html i think

teneightfive: I'm not sure there's anything that warrants it at this stage. Check out the branch and try it out and see what you think.

[UK company journey](http://localhost:3000/investment/start?client-company=dcdabbc9-1781-e411-8955-e4115bead28a)

[Foreign company journey](http://localhost:3000/investment/start?client-company=af91b667-a098-e211-a939-e4115bead28a)

src/views/investment/start.html
```diff
@@ -0,0 +1,136 @@
...
+                <li class="search-result__attribute search-result__attribute--inline">{TODO} investment projects</li>
+                <li class="search-result__attribute search-result__attribute--inline">{TODO} (D - post managed)</li>
+                <li class="search-result__attribute search-result__attribute--inline">Relationship manager: {TODO}</li>
+              </ul>
+            </a>
+          </li>
+        {% endfor %}
+      </ol>
+
+      <div class="pagination">
```
tyom: Could hide pagination if there are no pages.

teneightfive: 👍

src/controllers/investment/create.controller.js
```diff
@@ -0,0 +1,21 @@
...
+
+  if (!equityCompanyId) {
+    return res.redirect('/investment/start')
+  }
+
+  getInflatedDitCompany(req.session.token, equityCompanyId).then((equityCompany) => {
+    res.render('investment/create', {
+      equityCompany
+    })
+  }).catch(error => res.render('error', { error }))
```
feedmypixel: do we want to be exposing this error to the client? also should we be logging it?

src/views/investment/start.html
```diff
@@ -0,0 +1,138 @@
...
+                <li class="search-result__attribute search-result__attribute--inline">{TODO} investment projects</li>
+                <li class="search-result__attribute search-result__attribute--inline">{TODO} (D - post managed)</li>
+                <li class="search-result__attribute search-result__attribute--inline">Relationship manager: {TODO}</li>
+              </ul>
+            </a>
+          </li>
+        {% endfor %}
+      </ol>
+
+      {% if pagination %}
```
tyom: `pagination` does exist though. You should check for its length.

src/controllers/investment/start.controller.js
```diff
@@ -0,0 +1,75 @@
...
+  if (searchTerm) {
+    promises.push(search({
+      token: req.session.token,
+      page: req.query.page,
+      term: searchTerm
+    }))
+  }
+
+  Promise.all(promises).then((result) => {
+    let clientCompany, searchResult, pagination
```
feedmypixel: what are the thoughts around comma separated `let`? IMO I prefer them on separate lines

src/controllers/investment/start.controller.js
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    res.render('investment/start', {
+      clientCompany,
+      searchTerm,
+      searchResult,
+      pagination,
+      findInvestment
+    })
+  }).catch(error => res.render('error', { error }))
```
feedmypixel: same as comment above, do we want to be exposing this error to the client? also should we be logging it?

src/views/investment/start.html
teneightfive: Good point. Fixed.

src/controllers/investment/start.controller.js
```diff
@@ -0,0 +1,75 @@
...
+      page: req.query.page,
+      term: searchTerm
+    }))
+  }
+
+  Promise.all(promises).then((result) => {
+    let clientCompany, searchResult, pagination
+    let findInvestment = req.query['find-investment']
+
+    if (clientCompanyId) {
```
feedmypixel: these if/else blocks don't feel very clear. Can this work be simplified somehow?

src/controllers/investment/create.controller.js
teneightfive: It might be worth taking a look at error handling in a separate change.

Maybe doing something like having a generic error view that if not in prod it will display the stack trace of the error but in prod give a more generic error page.

@ztolley do we have any generic error handling in the server? As in, are all errors logged out at least to the console?

src/controllers/investment/start.controller.js
teneightfive: I'm not precious to be honest, this just felt a little neater at the time. Can put each definition on a new line if we prefer that standard?

cc @ztolley @tyom

src/controllers/investment/start.controller.js
teneightfive: Yeah the logic here felt a bit tricker and tried it a few ways. Combing the promises felt better than nested promises and unfortunately they're not both required.

`Promise.all` only returns an array so it needed some extra logic to decide what the items in the array actually were. Any thoughts on how this could be done instead?

src/controllers/investment/create.controller.js
feedmypixel: If we can send in `next/done` with the `req` and `res` we can just call it with the `error` and let it drop through to a generic error handler, which as @teneightfive said could work with winston and logging levels

src/controllers/investment/start.controller.js
tyom: I don't have a strong preference on this.

src/controllers/investment/create.controller.js
teneightfive: Something [like this](https://github.com/nhsuk/betahealth/blob/develop/config/express.js#L139-L174) for handling errors?

src/controllers/investment/start.controller.js
```diff
@@ -0,0 +1,75 @@
...
+
+  if (searchTerm) {
+    promises.push(search({
+      token: req.session.token,
+      page: req.query.page,
+      term: searchTerm
+    }))
+  }
+
+  Promise.all(promises).then((result) => {
```
tyom: What do we think about convention to put `then`, `catch` on new lines? IMO it makes it easier to read as steps. e.g.
```js
Promose.all(promises)
  .then(result => ...)
  .catch(error => ...)
```

src/controllers/investment/create.controller.js
teneightfive: And then just throw an error in `.catch()`?

src/controllers/investment/start.controller.js
teneightfive: Am happy to change to that as a convention.

src/controllers/investment/start.controller.js
```diff
@@ -0,0 +1,75 @@
...
+function postHandler (req, res) {
+  const isEquitySource = req.body['is-equity-source']
+  const clientCompany = req.body['client-company']
+
+  if (isEquitySource === 'true') {
+    res.redirect(`/investment/create?equity-company=${clientCompany}`)
+  } else if (isEquitySource === 'false') {
+    res.redirect(`/investment/start?client-company=${clientCompany}&find-investment=true`)
+  } else {
+    res.redirect(`/investment/start?client-company=${clientCompany}`)
```
teneightfive: @feedmypixel @tyom @ztolley am not sure about this right now. Basically the form doesn't have validation. I could render the same view but would mean a call to the company service again, or I could pass a query param back to the get handler but that feels nasty.

@ztolley what's been the approach with validation so far?

src/controllers/investment/create.controller.js
feedmypixel: Yup something like that. I always refer back to these for error handling
- https://expressjs.com/en/guide/error-handling.html
-https://github.com/expressjs/express/blob/8eb95ae57973b2cbe7778bc2e10450a380ca2efe/examples/error-pages/index.js

feedmypixel: > And then just throw an error in .catch()?


for the function signature 
```
```

then the catch
```
.catch((error) => {
   done(error)
})
```

teneightfive: Cool. I'll just make this throw an error for now then and can do the error handling in another change.

src/controllers/investment/start.controller.js
feedmypixel: I'd be tempted to go for
```
Promise
   .all(promises)
  .then(result => ...)
  .catch(error => ...)
```

feedmypixel: Not a good example to demonstrate this

works better

```
someArray
   .map()
   .filter()
   .reduce()
```

teneightfive: I think I prefer the function call on the same line. You might also have something like this:

```
asyncFunction()
  .then(result => ...)
  .catch(error => ...)
```

So keeping the function call on the same line feels more consistent.

src/controllers/investment/start.controller.js
feedmypixel: My vote is for `post > get > redirect`. On an error from the post then just drop through to the next middleware route handler via `next()`

```
router.post('/investment/start', postHandler, postErrorHandler)
```

in the `postErrorHandler` you can render the view with an errors you would like, which if needed can be sent through via `next(error)` call.

Any success redirects can be handled in the `postHandler`

teneightfive: Yeah, I think that feels better. Just means it'll need a call to the company service as well.

src/controllers/investment/start.controller.js
tyom: `.map` is a prototype method on `someArray`, as is `filter` and `reduce`.

`.all` is a method on `Promise` while `then/catch` are on its prototype and expect a promise. So they are not quite the "steps" I was referring to.

I'd say keep `.all()` on the same line as `Promise` 😄

src/controllers/investment/start.controller.js
ztolley: So far validation has been the job of the API
So we just passed it to the API and if it failed, attach that to the responce object and re-render the form

teneightfive: I think with this set of views though we don't need to be sending anything to the API yet. Have a look at the change and see if that makes sense now.

teneightfive:
@jim68000 regarding the API changes. Can we add an endpoint to return only foreign companies? Or maybe the ability to pass an option to the companies endpoint.

Also, the current visuals show data that I'm not is available yet. Do we have any of these:
- Number of investments in the UK
- Account tier
- Client relationship manager

jim68000:
The foreign only/uk only is a search change - I'm going to catch up with Pawel on Monday to find out where he's got to

Number of investments in UK is this call: GET /v3/investment/{companyId}/projects 

I *think* the Account Tier and Client Relationship Manager are the same as One List Status and One List Account Manager, which we do have, but I need to double check that with the Investment guys on Tuesdays

src/views/investment/create.html
ztolley: Just wondered why they were removed when it's easy to add and tbh we need to start putting more of them in

src/controllers/investment/start.controller.js
ztolley: Please look at contact edit controller to see how we've done this before. I want to be consistent. If we want to re-architect how we handle errors and forms lets agree the approach before to refactor everything

src/controllers/investment/start.controller.js
tyom: What about something like this:

```js
if (clientCompanyId) {
  promises.push(getInflatedDitCompany(req.session.token, clientCompanyId))
} else {
  promises.push(Promise.resolve())
}
```

and then:

```js
Promise.all(promises).then(([clientCompany, searchResult]) => {
  let findInvestment = req.query['find-investment']

  const pagination = getPagination(req, searchResult)

  if (clientCompanyId && clientCompany.uk_based) {
    findInvestment = true
  }

  res.render('investment/start', {
    clientCompany,
    searchTerm,
    searchResult,
    pagination,
    findInvestment
  })
})
```

Effectively always give the same number of promises to `.all()`.

feedmypixel: ok so my thoughts would be to use a promise based generator. Q.spawn is already in use on the project, although my preference is [co](https://github.com/tj/co)

**caveat this is untested code

```

  genCSRF(req, res)

  return getInvestmentPageDetail(req, clientCompanyId)
    .then(({clientCompany, searchTerm, searchResult, pagination, findInvestment}) => {
      res.render('investment/start', {
        clientCompany,
        searchTerm,
        searchResult,
        pagination,
        findInvestment
      })
    })
    .catch((error) => next(error))
}
```

```
const getInvestmentPageDetail = function (req, clientCompanyId) {
    Q.spawn(function * () {
      try {
        let searchResult
        let pagination
        let findInvestment = req.query['find-investment']
        let clientCompany

        if (clientCompanyId) {
          clientCompany = yield getInflatedDitCompany(req.session.token, clientCompanyId)
        }

        if (searchTerm) {
          searchResult = yield search({
            token: req.session.token,
            page: req.query.page,
            term: searchTerm
          })
          pagination = getPagination(req, searchResult)
        }

        resolve({
          clientCompany,
          searchTerm,
          searchResult,
          pagination,
          findInvestment
        })
      } catch (error) {
        winston.error(error)
        reject(error)
      }
    })
  })
}```

File: src/views/investment/create.html (hunk 1 above)
Comment (teneightfive) [inline src/views/investment/create.html]: I removed what wasn't necessary on this screen at this stage. They may well need to be present on a different step. The views and code that's there already for investment doesn't seem to be working and seems based on a previous version of the journey. So am just building up the latest version of the journey as I go.

File: src/controllers/investment-start.controller.js [hunk 10]
```diff
@@ -0,0 +1,95 @@
...
+function getHandler (req, res, next) {
+  const clientCompanyId = req.query['client-company']
+  const searchTerm = req.query['q']
+  const promises = []
+
+  genCSRF(req, res)
+
+  if (clientCompanyId) {
+    promises.push(getInflatedDitCompany(req.session.token, clientCompanyId))
+    // promises.push(getInvestmentSummary(req.session.token, clientCompanyId))
```
tyom: Remove?

src/views/investment/start.njk
```diff
@@ -0,0 +1,150 @@
...
+    <table class="table--key-value">
+      <tbody>
+        <tr>
+          <td>{{ clientCompany.name }}</td>
+        </tr>
+        <tr>
+          <td>{{ clientCompany.registered_address_country.name }}</td>
+        </tr>
+        <tr>
+          <td>{{ 'No' if clientCompanyInvestments.count == 0 else clientCompanyInvestments.count }} investment project{{ 's' if clientCompanyInvestments.count != 1 }} in the UK</td>
```
tyom: Same as above, could use `default`.

src/views/investment/start.njk
```diff
@@ -0,0 +1,150 @@
...
+              {% endif %}
+            {% endfor %}
+          </div>
+        </div>
+      {% endif %}
+    {% endif %}
+
+  {% else %}
+
+    <form action="" class="section archive-panel__form" method="POST">
```
tyom: Don't need empty `action` attribute. Pretty sure it's invalid in HTML5.

src/views/investment/create.njk
```diff
@@ -3,147 +3,26 @@
...
+    <table class="table--key-value">
+      <tbody>
+        <tr>
+          <td>{{ equityCompany.name }}</td>
+        </tr>
+        <tr>
+          <td>{{ equityCompany.registered_address_country.name }}</td>
+        </tr>
+        <tr>
+          <td>{{ 'No' if equityCompanyInvestments.count == 0 else equityCompanyInvestments.count }} investment project{{ 's' if clientCompanyInvestments.count != 1 }} in the UK</td>
```
teneightfive: First point is easier to read but not sure adding a pluralise filter will necessarily be cleaner right now so am going to leave as currently is.

src/controllers/investment-create.controller.js
```diff
@@ -0,0 +1,30 @@
+const router = require('express').Router()
+
+const { getInflatedDitCompany } = require('../services/company.service')
+const { getCompanyInvestmentProjects } = require('../repos/investment.repo')
+
+function getHandler (req, res, next) {
+  const equityCompanyId = req.query['equity-company']
+  const promises = [
+    getInflatedDitCompany(req.session.token, equityCompanyId),
+    // TODO: remove this when the company endpoint returns a list of invesments
```
feedmypixel: is it worth raising this as an issue? Or are we going to keep on top of //TODO's

src/controllers/investment-create.controller.js
```diff
@@ -0,0 +1,30 @@
+const router = require('express').Router()
+
+const { getInflatedDitCompany } = require('../services/company.service')
+const { getCompanyInvestmentProjects } = require('../repos/investment.repo')
+
+function getHandler (req, res, next) {
+  const equityCompanyId = req.query['equity-company']
+  const promises = [
```
feedmypixel: is there a better name? `promises` is a little generic

src/controllers/investment-create.controller.js
```diff
@@ -0,0 +1,30 @@
...
+    getInflatedDitCompany(req.session.token, equityCompanyId),
+    // TODO: remove this when the company endpoint returns a list of invesments
+    getCompanyInvestmentProjects(req.session.token, equityCompanyId)
+  ]
+
+  if (!equityCompanyId) {
+    return res.redirect('/investment/start')
```
feedmypixel: Shame we don't have reverse routing in node. (in scala this could be `controllers.routes.investment.start.url`)
Not for this PR but seeing this makes me feel that maybe these strings should be in some form of config? Would need trying out to see if thats a little overkill

src/controllers/investment-start.controller.js
```diff
@@ -0,0 +1,90 @@
...
+function getHandler (req, res, next) {
+  const clientCompanyId = req.query['client-company']
+  const searchTerm = req.query['q']
+  const promises = []
+
+  genCSRF(req, res)
+
+  if (clientCompanyId) {
+    promises.push(getInflatedDitCompany(req.session.token, clientCompanyId))
+    // TODO: remove this when the company endpoint returns a list of invesments
```
feedmypixel: same comment as above on `//TODO's`

src/controllers/investment-start.controller.js
```diff
@@ -0,0 +1,90 @@
...
+    promises.push(getCompanyInvestmentProjects(req.session.token, clientCompanyId))
+  } else {
+    promises.push(Promise.resolve())
```
feedmypixel: can be `promises.push(Promise.resolve(), Promise.resolve())`

src/controllers/investment-create.controller.js
teneightfive: Yep a good thought. If we want to change the url of a particular view we don't want all links to that page to 404.

src/controllers/investment-start.controller.js
```diff
@@ -0,0 +1,90 @@
...
+      page: req.query.page,
+      term: searchTerm
+    }))
+  } else {
+    promises.push(Promise.resolve())
+  }
+
+  Promise.all(promises)
+    .then(([clientCompany, clientCompanyInvestments, searchResult]) => {
+      let showSearch = req.query['show-search'] || false
```
feedmypixel: `req.query['show-search']` should give you `undefined` as it won't be on the object if nothing is there. Although you may be wanting to explicitly set this to a `Boolean`

src/controllers/investment-create.controller.js
teneightfive: Not sure. I went with this pattern for now as we're still building out the rest of the journey and not turning it all on just yet. 

If it was a new set of features we were releasing but wanted to come back and improve I'd say somewhere like github issues or JIRA would be more appropriate.

src/controllers/investment-create.controller.js
teneightfive: Any suggestions? `apiCalls`?

src/controllers/investment-start.controller.js
teneightfive: Yep, wanted to explicitly set this to either `true` or `false`. That's what the current test checks for too.

src/sass/_search-result.scss
```diff
@@ -0,0 +1,46 @@
+.search-result {
+  margin: 0 0 10px;
+  padding: 0;
+  list-style: none;
+}
+
+.search-result__link {
+  border: 1px solid $grey-2;
```
feedmypixel: what are peoples thoughts @teneightfive @tyom @ztolley about explicit imports in `sass` files when variables and mixins are used in a file?

so `@import("trade-elements/dist/sass/colours/_palette.scss")` at the top of the file here

src/controllers/investment-create.controller.js
jim68000: +1 for centralised route config. And all embedded strings, tbh

src/sass/_search-result.scss
```diff
@@ -0,0 +1,46 @@
+.search-result {
+  margin: 0 0 10px;
```
feedmypixel: There are a lot of magic numbers around. It would be good to get some rhythm setup with `$gutter` etc across the sass

src/sass/_search-result.scss
teneightfive: Yes. We should be declaring specific dependencies for a component. Will help with making components more modular.

Will add the dependencies for this file.

src/sass/_search-result.scss
teneightfive: Agreed. @ztolley are there any generic spacing units currently in use that could be used instead?

src/controllers/investment-create.controller.js
feedmypixel: `companyDetails` ?

src/controllers/investment-create.controller.js
feedmypixel: created
- `Jira > DH-244`  Centralise Router config
- `Jira > DH-245` Centralise Embedded strings

src/controllers/investment-create.controller.js
teneightfive: But it's not just related to the company details. The search result is also appended to the `promises` array, which is independent of the company.

src/controllers/investment-create.controller.js
teneightfive: Ah. I also just created #78 as an issue on the repo...

src/sass/_search-result.scss
feedmypixel: created the following around this
- `Jira > DH-246` White space and Vertical Rhythm parity needed

src/controllers/investment-create.controller.js
feedmypixel: I have referenced the Jira in the Issue. I don't have a preference. May be worth asking Tonia what works for her?

src/controllers/investment-create.controller.js
feedmypixel: `companyDetailsAndSearchResult`? If it says what is in on the tin I then don;t have to open the tin to look...

src/controllers/investment-create.controller.js
jim68000: As we don't have integration between Jira and GH issues (maybe that's a @TODO?), I'd prefer Jira for visibility

feedmypixel:
Good stuff @teneightfive 

A few thoughts from this Pr:
- Big PR's are difficult to go through especially if you haven't had your head in the code. Small and often wins every time, means less things get through. I appreciate this isn't always possible with chunks of work
- Once we have more markup in we should start looking at abstracting it into views

src/views/investment/create.njk
```diff
@@ -3,147 +3,31 @@
...
+        <table class="table--key-value">
+          <tbody>
+            <tr>
+              <td>{{ equityCompany.name }}</td>
+            </tr>
+            <tr>
+              <td>{{ equityCompany.registered_address_country.name }}</td>
+            </tr>
+            <tr>
+              <td>{{ equityCompanyInvestments.count|default('No', true) }} investment project{{ 's' if equityCompanyInvestments.count != 1 }} in the UK</td>
```
jim68000: Inclined to agree that a pluralise filter might be nicer here. OK for now. Maybe raise an issue?

teneightfive: Have raised a [JIRA ticket](https://uktrade.atlassian.net/browse/DH-247) for it.

teneightfive:
>Big PR's are difficult to go through especially if you haven't had your head in the code. Small and often wins every time, means less things are missed. I appreciate this isn't always possible with chunks of work

@feedmypixel absolutely agree. I'm not a fan of big PRs either. This started to spiral a bit but was all related to one chunk of work. 

Tried to reduce the complexity by not including the logic for the next step yet. It will be a separate change off the back of this work.

>Once we have more markup in we should start looking at abstracting it into smaller chunks for reuse to be consumed by views/pages

Definitely. Once we've printed everything out and have an understanding of what components we currently need we'll start to see what has lots of cases of re-use so can structure the macros/markup appropriately.

teneightfive:
Have created a [ticket in JIRA](https://uktrade.atlassian.net/browse/DH-248) for adding foreign company search support.

src/sass/_search-result.scss
```diff
@@ -0,0 +1,49 @@
...
+  display: block;
+  text-transform: capitalize;
+
+  strong {
+    @include bold-14;
+    text-transform: inherit;
+  }
+}
+
+.search-result__attribute--inline {
```
tyom: What does this do?

teneightfive: Changes the display of the attributes to inline.

tyom: Thanks Dom. Why does it need to be inline? It seems that the first list item is always block so it never actually inlines it.

PR #68, opened 2017-05-12
Fixed a validation issue with Zendesk JSON

jim68000:
If an email is present it must validate - setting it to Anonymous was a no-no

### DIFF
--- a/src/controllers/support.controller.js
+++ b/src/controllers/support.controller.js
@@ -56,7 +56,7 @@ function postBug (req, res) {
   }
   const ticket = {
     requester: {
-      email: (req.body.email && req.body.email.length > 0) ? req.body.email : 'Anonymous'
+      name: 'Data Hub user'
     },
     subject: req.body.title,
     comment: {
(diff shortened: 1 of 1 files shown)


src/controllers/support.controller.js
```diff
@@ -68,11 +68,14 @@ function postBug (req, res) {
...
   }
+  if (req.body.email && req.body.email.length > 0) {
+    ticket.requester.email = req.body.email
+  }
   return postToZen(ticket)
     .then(({data}) => {
       req.flash('success-message', `Created new bug, reference number ${data.ticket.id}`)
       res.redirect('/support/thank')
-    })
+    }).catch(function (error) { res.render('support/bug', {}, {errors: {message: error.message}}) })
```
tyom: Perhaps, drop `.catch` on the new line to improve readability? And maybe format the function body. 

Could also use arrow function as it's typically preferred in callbacks.

jim68000: Will do

PR #70, opened 2017-05-15
Updated the contact interactions list to match the style used for company

ztolley:


test/controllers/contact-interaction.controller.test.js
```diff
@@ -0,0 +1,200 @@
...
+  let interaction
+  let formatted
+  let contactInteractionController
+  let contact
+  let company
+
+  beforeEach(function () {
+    contactInteractionController = require('../../src/controllers/contact-interaction.controller')
+    company = { id: '555', name: 'Fred ltd' }
+    contact = {
```
feedmypixel: Are these common? can they be abstracted into the `/test/data` folder so they can be used by others?

test/services/interaction-formatting.service.test.js
```diff
@@ -229,4 +229,74 @@ describe('Interaction formatting service', function () {
       expect(actual).to.deep.equal(expected)
     })
   })
+  describe('Contact interactions', function () {
+    it('should return the required fields for the display interaction', function () {
+      const expected = {
```
feedmypixel: suggestion const name of `expectedContactInteraction`

test/services/interaction-formatting.service.test.js
```diff
@@ -229,4 +229,74 @@ describe('Interaction formatting service', function () {
...
+        url: '/interaction/22651151-2149-465e-871b-ac45bc568a62/details',
+        subject: 'Subject 1234',
+        interaction_type: 'Email',
+        advisor: 'John Brown',
+        date: '14 Feb 2017',
+        service: 'service name',
+        notes: 'Here are some notes<br/>line 2.',
+        dit_team: 'team name'
+      }
+      const actual = interactionFormattingService.getDisplayContactInteraction(interaction)
```
feedmypixel: suggestion const name of `contactInteraction`

test/controllers/contact-interaction.controller.test.js
ztolley: What do you suggest we name the files? contact.js? Interactions.js? formattedInteractions? I'm not using Nock to fake the network, I'm using stubs.. unless you think I should be mocking the api calls and not the immediate dependency? The tricky bit there is tests start to overlap as we have a test suite just for the formatter (or should)

src/controllers/contact-interaction.controller.js
```diff
@@ -2,7 +2,7 @@ const express = require('express')
 const winston = require('winston')
 const Q = require('q')
 
-const { getDisplayContactInteraction } = require('../services/interaction-formatting.service')
+const interactionFormattingService = require('../services/interaction-formatting.service')
```
tyom: The destructured version removes the need to repeat the service name further down. Should it be kept as is?

src/services/contact-data.service.js
```diff
@@ -53,28 +56,22 @@ function getContactInteractionsAndServiceDeliveries (token, contactId) {
 
         // Parse the service delivery results into something that can be displayed
         const parsedServiceDeliverys = serviceDeliverys.map((serviceDelivery) => {
-          return {
-            id: serviceDelivery.id,
-            date: serviceDelivery.attributes.date,
-            created_on: serviceDelivery.attributes.date,
-            notes: serviceDelivery.attributes.notes,
-            subject: serviceDelivery.attributes.subject,
+          return Object.assign({}, {id: serviceDelivery.id}, serviceDelivery.attributes, {
```
tyom: You could move `serviceDelivery.id` to the object literal you pass next.
```js
return Object.assign({},
  serviceDelivery.attributes,
  {
    id: serviceDelivery.id,
    interaction_type: { id: null, name: 'Service delivery' },
    // …
  }
})
```

src/views/contact/interactions.njk
```diff
@@ -3,27 +3,46 @@
 
 {% block tabmain %}
 <h2 class="heading-medium">Contact interactions</h2>
-{% if displayInteractions.length > 0 %}
+{% if interactions.length > 0 %}
```
tyom: Could drop `> 0` as zero length is falsey.

src/views/contact/interactions.njk
```diff
@@ -3,27 +3,46 @@
...
 <ul id="interaction-list" class="lined-list">
-  {% for interaction in displayInteractions %}
+  {% for interaction in interactions %}
   <li class="lined-list__item">
-    <div class="card interaction">
-      <a class="card__metadata" href="{{interaction.url}}">{{interaction.subject}}</a>
+    <div class="card expanding-card interaction" id="{{interaction.id}}">
```
tyom: Feels like it should be either `card` or `expanding-card` component. Or perhaps `--expanding` should be the modifier of `card` component?

src/views/contact/interactions.njk
```diff
@@ -3,27 +3,46 @@
...
+      <a class="card__metadata expanding-card__button" href="#{{interaction.id}}">
+        <div class="expanding-card__indicator"></div>
+        {{interaction.subject}}
+      </a>
       <div class="split-data">
         <div>
-          <div class="card__metadata">{{interaction.interaction_type}}</div>
-          <div class="card__metadata">{{interaction.date}}</div>
+          <div class="card__metadata"><strong>Type:</strong> {{interaction.interaction_type}}</div>
```
tyom: I think it would be more readable if `strong` was on the new line. I generally avoid putting nested HTML tags on single line, unless they are really simple.

src/controllers/contact-interaction.controller.js
ztolley: If I didn't change it it wasn't picking up the stubbed function

test/controllers/contact-interaction.controller.test.js
feedmypixel: stubbing sounds fine. My thought was can these be used elsewhere if so they should be in a files. I'd vote for `.json` files and yes `contact.js`, `interactions.js` (add `-<type>` if they are of a specific type).

feedmypixel: Whats formatting the `formattedInteractions`?

src/services/interaction-formatting.service.js
```diff
@@ -63,11 +63,15 @@ function getDisplayContactInteraction (interaction) {
...
   const result = {
+    id: interaction.id,
     url: `/${type}/${interaction.id}/details`,
     interaction_type: interaction.interaction_type.name,
     subject: interaction.subject,
-    date: formatLongDate(interaction.date),
-    advisor: getPropertyName(interaction, 'dit_advisor')
+    date: formatMediumDate(interaction.date),
+    advisor: getPropertyName(interaction, 'dit_advisor'),
+    notes: newlineToBr(interaction.notes),
```
teneightfive: Do we need to be converting the response here? Is this something that can be done with template filters rather than at this point?

src/views/contact/interactions.njk
teneightfive: `card card--expandable` and then toggling a state class like `is-expanded` feels like a good BEM based approach.

src/services/interaction-formatting.service.js
ztolley: I think for now it's fine here. I don't like putting any sort of logic in the view if I can help it and this has to test a few things before returning the value.

teneightfive: Okay. I think formatting changes are clearer when handled at view level using filters though. But seeing as we're already using this method lets keep it.

tyom:
Could we use kebab-case for test data file names?

src/javascripts/expanding-cards.js
```diff
@@ -24,8 +24,8 @@ class ExpandingCard {
    */
   cacheElements (element) {
     this.wrapperElement = element
-    this.toggleButton = element.querySelector('.expanding-card__button')
-    this.revealElement = element.querySelector('.expanding-card__reveal')
+    this.toggleButton = element.querySelector('.card--expanding__button')
```
tyom: I don't think this is the BEM way. Modifiers can't be elements. An element can be a modifier however. Element should be part of a component and modifiers should only modify existing component or element.

src/sass/_shame.scss
```diff
@@ -249,28 +249,28 @@ hr {
 }
 
 
-a.expanding-card__button {
+a.card--expanding__button {
```
tyom: Is qualifier required here?

src/sass/_shame.scss
```diff
@@ -249,28 +249,28 @@ hr {
...
-  .expanding-card__button-indicator::before {
+  .card--expanding__button-indicator::before {
     font-family: "FontAwesome";
     width: 10px;
     display: inline-block;
     content: "\f0da";
   }
 
-  .expanding-card--open .expanding-card__button-indicator::before {
+  .card--expanding--open .card--expanding__button-indicator::before {
```
tyom: This is not BEM. State should be expressed as `is-open` which is chained to block/element. And modifier should never be used on its own.

src/sass/_shame.scss
```diff
@@ -249,28 +249,28 @@ hr {
...
     content: "\f0d7";
   }
 
-  a.expanding-card__button {
+  a.card--expanding__button {
```
tyom: Is qualifier required?

src/javascripts/expandable-card.js
```diff
@@ -0,0 +1,95 @@
+/* eslint no-new: 0 */
+const { hide, toggleVisible, toggleClass, addClass, createElementFromMarkup } = require('../lib/element-stuff')
```
tyom: Just a recommendation to expand these long destructuring to improve readability. Pretty common pattern.

```js
const {
  hide,
  toggleVisible,
  toggleClass,
  addClass,
  createElementFromMarkup
} = require('../lib/element-stuff')
```

PR #76, opened 2017-05-16
generic error logging and tests

feedmypixel:
This work is the first cut at generic error logging for the application. 

### 404
If no route is found it will drop through to the `notFound` middleware, then the `catchAll` middleware and will serve a `404` page.

### !404
In your controller you can now call `next()` with an instantiated `Error` object as the first argument and it will drop through to the `catchAll` middleware. By providing a value to the `error.status` property on the Error object you can set the response status, otherwise without a `.status` value this will default to `500`

```
.catch(next)
```

the stack trace will now show only in the browser when in **development** mode
screencapture-localhost-3000-dfdfd-1495032660523

## Of Note
- winston will log `error` for `500's` and `info` for anything else.
- The copy/humanised error messages are currently just placeholder copy
- Its worth have a read of the [expressjs error-handling docs](https://expressjs.com/en/guide/error-handling.html)

#### 500
screen shot 2017-05-16 at 17 02 40

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -1,8 +1,10 @@
 const port = process.env.PORT || 3000
-const defaultLogLevel = (process.env.NODE_ENV === 'development') ? 'debug' : 'error'
+const isDev = (process.env.NODE_ENV === 'development')
+const defaultLogLevel = isDev ? 'debug' : 'error'
 
 module.exports = {
   env: process.env.NODE_ENV,
+  isDev,
   port,
   apiRoot: process.env.API_ROOT || 'http://localhost:8000',
   api: {
--- a/src/middleware/errors.js
+++ b/src/middleware/errors.js
@@ -1,0 +1,34 @@
+const config = require('../config')
+const winston = require('winston')
+
+function notFound (req, res, next) {
+  const error = new Error('Not Found')
+  error.status = 404
+
+  next(error)
+}
+
+function catchAll (error, req, res, next) {
+  const statusCode = error.status = (error.status || 500)
+  const statusMessage = statusCode === 404
+    ? 'Sorry we couldn\'t find that page!'
+    : 'Sorry something has gone wrong!'
+
+  if (res.headersSent) {
+    return next(error)
+  }
+
+  winston[statusCode === 500 ? 'error' : 'info'](error)
+
+  res.status(statusCode)
+    .render('errors/index', {
+      statusCode,
+      statusMessage,
+      devErrorDetail: config.isDev && error
+    })
+}
+
+module.exports = {
+  notFound,
+  catchAll
+}
--- a/src/sass/application.scss
+++ b/src/sass/application.scss
@@ -16,6 +16,7 @@ $images-dir: '/images/';
 @import 'shame';
 @import 'splitheader';
 @import 'tables';
+@import 'pre';
 
 .table--readonly {
   border-left-width: 5px;
--- a/src/server.js
+++ b/src/server.js
@@ -37,9 +37,9 @@ const interactionEditController = require('./controllers/interaction-edit.contro
 const serviceDeliveryController = require('./controllers/service-delivery.controller')
 const supportController = require('./controllers/support.controller')
 
+const isDev = config.isDev
 const app = express()
 app.disable('x-powered-by')
-const isDev = app.get('env') === 'development'
 winston.level = config.logLevel
 
 const RedisStore = redisCrypto(session)
--- a/src/server.js
+++ b/src/server.js
@@ -20,6 +20,7 @@ const metadata = require('./repos/metadata.repo')
 const user = require('./middleware/user')
 const auth = require('./middleware/auth')
 const csrf = require('./middleware/csrf')
+const errors = require('./middleware/errors')
 
 const apiController = require('./controllers/api.controller')
 const contactController = require('./controllers/contact.controller')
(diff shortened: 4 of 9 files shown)


src/middleware/errors.js
```diff
@@ -0,0 +1,32 @@
...
+    : 'Sorry something has gone wrong!'
+
+  if (res.headersSent) {
+    return next(error)
+  }
+
+  winston[statusCode === 500 ? 'error' : 'info'](error)
+
+  res.status(statusCode)
+    .render('errors/index.njk', {
```
teneightfive: I think you can drop the file extension here.

src/middleware/errors.js
```diff
@@ -0,0 +1,32 @@
...
+      statusCode,
+      statusMessage
```
teneightfive: Would be nice to pass the the error object back if in development mode. Something like this:

```js
res.render(template, {
  ...
  error: config.env === 'development' ? error : {}
});
```

src/views/errors/index.njk
```diff
@@ -0,0 +1,6 @@
+{% extends "layouts/ukti.njk" %}
+
+{% block main %}
+  <h1 class="heading-medium error">{{ statusCode }}</h1>
```
teneightfive: I would be tempted to drop the error code and use the message as the heading. Codes like 404/500 will be meaningless to most users. 

Then display the error code and full error message if passed to the view when in development mode.

src/middleware/errors.js
feedmypixel: Can do. FYI it is being logged out in the console, do we also want to see this in the browser window?

src/views/errors/index.njk
feedmypixel: I spoke with mills yesterday. I think there needs to be a second card to design and content generic errors

feedmypixel: I have created `Jira > DH-243` around this

src/middleware/errors.js
teneightfive: Don't think it will do any harm in development mode. Saves you having to switch back to the console if it's something obvious.

test/middleware/errors.test.js
```diff
@@ -0,0 +1,113 @@
+/* globals expect: true, describe: true, it: true, before: true, after: true, sinon: true */
```
tyom: Can rebase and remove these now.

test/middleware/errors.test.js
```diff
@@ -0,0 +1,113 @@
...
+      errors.notFound(null, null, nextSpy)
+
+      expect(nextSpy.calledOnce).to.be.true
+      expect(nextSpy.args[0][0] instanceof Error).to.be.true
+      expect(nextSpy.args[0][0].message).to.equal('Not Found')
+      expect(nextSpy.args[0][0].status).to.equal(404)
+      done()
+    })
+  })
+  describe('catchAll method', function () {
```
tyom: New line?

src/server.js
```diff
@@ -36,9 +37,9 @@ const interactionEditController = require('./controllers/interaction-edit.contro
 const serviceDeliveryController = require('./controllers/service-delivery.controller')
 const supportController = require('./controllers/support.controller')
 
+const isDev = config.isDev
```
teneightfive: We can probably use `config.isDev` where this variable was previously being used and remove this declaration.

feedmypixel: Could do, I cached in a `const` as its being used 3 times in this file

src/config.js
```diff
@@ -1,8 +1,10 @@
 const port = process.env.PORT || 3000
-const defaultLogLevel = (process.env.NODE_ENV === 'development') ? 'debug' : 'error'
+const isDev = (process.env.NODE_ENV === 'development')
```
tyom: Is it worth checking for `process.env.NODE_ENV !== 'production'` as `development` is not always set.

teneightfive: Yes, just spotted this breaks some other things in `server.js` (nunjucks watching) unless you specifically set env to `development`. @feedmypixel maybe do a quick patch.

src/views/errors/index.njk
```diff
@@ -0,0 +1,13 @@
...
+  <p>{{ statusMessage }}</p>
+
+  {% if devErrorDetail %}
+    <ul>
```
tyom: `ul`? Is that the right element? Would a `dl` be better suited?

src/config.js
feedmypixel: So I did add `"develop": "NODE_ENV='development' parallel sass:watch js:client:watch js:server:watch"` in `package.json` but removed it as it felt like `NODE_ENV="development" npm run develope` was the better way. Thoughts?

src/sass/application.scss
```diff
@@ -16,6 +16,7 @@ $images-dir: '/images/';
 @import 'shame';
 @import 'splitheader';
 @import 'tables';
+@import 'pre';
```
tyom: You could define a component like `error-details` and name file accordingly. I would try to avoid file names that contain other class names.

src/config.js
tyom: Most projects I worked on would set production environment rather than development.

src/sass/application.scss
feedmypixel: yeah thats me trying to be too generic - will do

PR #77, opened 2017-05-16
Test setup

teneightfive:
Removes some of the repetition in test files and moves it to mocha configuration.

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,7 +59,7 @@
     "test": "mocha --require test/common.js --ui bdd test/**/*.test.js",
     "coverage": "nyc --reporter=html --reporter=text --reporter=lcov mocha --require test/common.js --ui bdd test/**/*.test.js",
     "lint": "npm run lint:js && npm run lint:sass",
-    "lint:js": "standard",
+    "lint:js": "eslint ./test ./src *.js",
     "lint:js:modified": "CHANGED_FILES=$(git diff --staged --name-only | grep .*.js$); if [ $CHANGED_FILES ]; then eslint $CHANGED_FILES; fi",
     "lint:sass": "sass-lint -c .sass-lint.yml 'src/**/*.scss' -v -q",
     "lint:sass:modified": "sass-lint -c .sass-lint.yml $(git diff --staged --name-only | grep .*.scss) -v -q",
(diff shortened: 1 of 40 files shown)


package.json
```diff
@@ -83,7 +83,10 @@
   "standard": {
     "parser": "babel-eslint",
     "globals": [
+      "$",
```
feedmypixel: can we also add in `before` and `after` here? I am using them in PR [#76](https://github.com/uktrade/data-hub-fe-beta2/pull/76/files#diff-1046f6ecb8f561f85d0f7a7be8f407ddR12)

test/controllers/company-archive.controller.test.js
```diff
@@ -34,7 +30,7 @@ describe('Company controller, archive', function () {
     getViewCompanyLinkStub = sinon.stub().returns('/testurl')
     companyRepositoryArchiveCompanyStub = sinon.stub().resolves(null)
     companyRepositoryUnArchiveCompanyStub = sinon.stub().resolves(null)
-    companyArchiveController = proxyquire('../../src/controllers/company-archive.controller', {
+    companyArchiveController = proxyquire(`${appFolder}/controllers/company-archive.controller`, {
```
feedmypixel: It feels like we are loosing something with this global, something is lost in the communication of the files location. 
Maybe just a `root` variable would communicate better? `${root}/src/controllers/company-archive.controller` 

But then, is it worth it for saving a few `../../` especially when it is used in some places and not in others.

e.g:
```
...
interactionDataService = proxyquire(`${appFolder}/services/interaction-data.service`, {
        '../repos/company.repo': {
          getDitCompany: getDitCompanyStub
        },
...
```

teneightfive: Yeah, I wasn't totally convinced either. I could try switching to using root. Would then also mean we can remove `testFolder`. 

Although that does mean that if we change those locations we'll need to update that path across all the test files again.

It seems we may restructure to separate the client from the server at some point so would mean just one change to `appFolder`. Maybe changing the name to something clearer, `serverSrc`?

feedmypixel: > Although that does mean that if we change those locations we'll need to update that path across all the test files again.

I'd call YAGNI on it. When we do lets approach it then. 

IMO there isn't a huge gain for changing `../../` to a variable because moving this signpost into a variable is a disadvantage and causes extra cognitive load. 
Any changes we can use `search & replace`

teneightfive: The problem is, when defining proxyquire globally it uses it's current path as the route where it calculate relative paths from. 

So a file in `./test/controllers/ctrl.test.js` which wants to reference `./src/controllers/ctrl.js` will need `../src/controllers/ctrl.js` as the relative path which feels more confusing. 

Think I'm going to remove `appFolder` and `testFolder` and use `root`. I don't think having a `root` variable would cause much cognitive load when looking at file references in test files.

feedmypixel: > Think I'm going to remove appFolder and testFolder and use root. I don't think having a root variable would cause much cognitive load when looking at file references in test files.

Agreed

PR #79, opened 2017-05-17
generic error logging and tests PR suggestions

feedmypixel:
- Flip `isDev` config value to check for `!production` so "development" doesn't have to be explicitly set for `NODE_ENV`
- Style and markup improvements around the stack errors component

screencapture-localhost-3000-dfdf-1495039224123

### DIFF
--- a/src/sass/_errors.scss
+++ b/src/sass/_errors.scss
@@ -1,0 +1,16 @@
+.stack-trace {
+  margin-bottom: $gutter;
+}
+
+.stack-trace__heading {
+  @include heading-24;
+  margin-bottom: $gutter/2;
+}
+
+.stack-trace__details {
+  @include core-16;
+  font-family: monospace;
+  overflow: auto;
+  word-wrap: normal;
+  white-space: pre;
+}
--- a/src/views/errors/index.njk
+++ b/src/views/errors/index.njk
@@ -5,9 +5,9 @@
   <p>{{ statusMessage }}</p>
 
   {% if devErrorDetail %}
-    <ul>
-      <li>{{ devErrorDetail.message }}</li>
-      <li><pre class="horizontal-scroll">{{ devErrorDetail.stack }}</pre></li>
-    </ul>
+    <dl class="stack-trace">
+      <dt class="stack-trace__heading">{{ devErrorDetail.message }}</dt>
+      <dd class="stack-trace__details">{{ devErrorDetail.stack }}</dd>
+    </dl>
   {% endif %}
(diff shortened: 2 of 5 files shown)


tyom:
You can't do fixups on commits that have already been merged.

feedmypixel:
> You can't do fixups on commits that have already been merged.

ah school boy `--fixup` mistake. Fixed.

src/views/errors/index.njk
```diff
@@ -5,9 +5,9 @@
   <p>{{ statusMessage }}</p>
 
   {% if devErrorDetail %}
-    <ul>
-      <li>{{ devErrorDetail.message }}</li>
-      <li><pre class="horizontal-scroll">{{ devErrorDetail.stack }}</pre></li>
-    </ul>
+    <dl>
+      <dt>{{ devErrorDetail.message }}</dt>
+      <dd><pre class="error-details">{{ devErrorDetail.stack }}</pre></dd>
```
tyom: Why not give `dl` the class name `error-details`. That way you can also style the rest of the component if needed. You could also drop the `pre` tag and apply the styles directly on `dd`. Maybe adding BEM elements for that component.

feedmypixel: so as we don't have other styling currently (I can jazz it up) add styles like this?
```
.stack-trace {
}
.stack-trace__heading {
}
.stack-trace__details {
  overflow: auto;
  word-wrap: normal;
  white-space: pre;
}
```

src/sass/_error-details.scss
```diff
@@ -1,4 +1,4 @@
-.horizontal-scroll {
+.error-details {
   overflow: auto;
   word-wrap: normal;
   white-space: pre;
```
tyom: Should add `font-family: monospace` since it's reset to `inherit` and dial down font size a bit (.9em?). Could also add a bottom margin to separate stack from error message.

src/views/errors/index.njk
tyom: 👍

src/sass/_errors.scss
```diff
@@ -0,0 +1,15 @@
+.stack-trace {
+  @include heading-24;
```
tyom: Maybe move this to `stack-trace__heading`?

feedmypixel: ah yeah that should be there! ta

teneightfive: We should also include the dependencies for this file. 

```scss
@import 'typography'
```

PR #82, opened 2017-05-18
Add support for including components using a nunjucks extension

teneightfive:
This adds the ability to separate components into their own `.njk` files and call them from any view, include or macro using a [custom nunjucks tag](https://mozilla.github.io/nunjucks/api.html#custom-tags).

Custom tags also allow easier reading of view files. It clearly shows a component is being loaded and avoids confusion with nunjucks' double brace syntax which could be calling a global function/variable, a macro or a local variable.

It also includes moving the nunjucks specific config to its own file to increase maintainability and reduce the configuration happening in `server.js`.

## Including components

To include a component with its default state or that expects no data:

```njk
{% component 'person' %}
```

To include a component with data:

```njk
{% component 'person', {
  name: 'Barry',
  age: '55'
} %}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,6 +127,29 @@ yarn run lint
 
 See the [contributing guide](./CONTRIBUTING.md).
 
+## Components
+
+This app includes support for including components using a custom [nunjucks
+tag](https://mozilla.github.io/nunjucks/api.html#custom-tags). This method
+allows components to be available in all layouts, views, includes and macros
+and allows each component to be a separating entity which makes maintaining
+and testing them easier.
+
+To include a component with its default state or one that expects no data:
+
+```njk
+{% component 'person' %}
+```
+
+To include a component and pass data to it:
+
+```njk
+{% component 'person', {
+  name: 'Barry',
+  age: '55'
+} %}
+```
+
 ## Deployment
 
 Commits to `develop` are automatically deployed to a heroku instance. Pull
--- a/config/nunjucks.js
+++ b/config/nunjucks.js
@@ -1,0 +1,63 @@
+const nunjucks = require('nunjucks')
+const winston = require('winston')
+
+const COMPONENTS_PATH = '_components/' // relative to views path
+const COMPONENT_EXT = 'njk'
+
+function ComponentExtension (env) {
+  this.tags = ['component']
+
+  this.parse = function parse (parser, nodes) {
+    const tok = parser.nextToken()
+    const args = parser.parseSignature(null, true)
+
+    parser.advanceAfterBlockEnd(tok.value)
+
+    return new nodes.CallExtension(this, 'run', args)
+  }
+
+  this.run = function run (context, name, locals = {}) {
+    let result = ''
+
+    try {
+      result = env.render(`${COMPONENTS_PATH}${name}.${COMPONENT_EXT}`, locals)
+    } catch (e) {
+      if (e.message.includes('template not found')) {
+        result = `Component '${name}' does not exist`
+      } else {
+        result = e.message
+      }
+
+      winston.error(result)
+    }
+
+    return new nunjucks.runtime.SafeString(result)
+  }
+}
+
+module.exports = (app, config) => {
+  const env = nunjucks.configure([
+    `${config.root}/src/views`,
+    `${config.root}/node_modules/@uktrade/trade_elements/dist/nunjucks`
+  ], {
+    autoescape: true,
+    express: app,
+    watch: config.isDev
+  })
+
+  // Custom filters
+  const filters = require(`@uktrade/trade_elements/dist/nunjucks/filters`)
+
+  filters.stringify = JSON.stringify
+
+  Object.keys(filters).forEach((filterName) => {
+    env.addFilter(filterName, filters[filterName])
+  })
+
+  // Custom extensions
+  env.addExtension('ComponentExtension', new ComponentExtension(env))
+
+  // Global variables
+
+  return env
+}
--- a/src/config.js
+++ b/src/config.js
@@ -1,8 +1,11 @@
+const path = require('path')
+
 const port = process.env.PORT || 3000
 const isDev = (process.env.NODE_ENV !== 'production')
 const defaultLogLevel = isDev ? 'debug' : 'error'
 
 module.exports = {
+  root: path.normalize(`${__dirname}/..`),
   env: process.env.NODE_ENV,
   isDev,
   port,
--- a/src/server.js
+++ b/src/server.js
@@ -4,15 +4,14 @@ const compression = require('compression')
 const config = require('./config')
 const express = require('express')
 const flash = require('connect-flash')
-const nunjucks = require('nunjucks')
 const redis = require('redis')
 const redisCrypto = require('connect-redis-crypto')
 const session = require('express-session')
 const url = require('url')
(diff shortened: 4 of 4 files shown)


config/nunjucks.js
```diff
@@ -0,0 +1,67 @@
+const nunjucks = require('nunjucks')
+const changeCase = require('change-case')
+const winston = require('winston')
+
+const COMPONENTS_PATH = '_components/' // relative to views path
+const COMPONENT_EXT = 'njk'
```
feedmypixel: I'd suggest `NUNJUCKS_FILE_EXT`

config/nunjucks.js
```diff
@@ -0,0 +1,67 @@
...
+
+      winston.error(result)
+    }
+
+    return new nunjucks.runtime.SafeString(result)
+  }
+}
+
+module.exports = (app, config) => {
+  const env = nunjucks.configure([
```
feedmypixel: you could also include `app.set('view engine', 'njk')` in this file? from [server.js](https://github.com/uktrade/data-hub-fe-beta2/blob/develop/src/server.js#L102)

config/nunjucks.js
```diff
@@ -0,0 +1,67 @@
+const nunjucks = require('nunjucks')
+const changeCase = require('change-case')
+const winston = require('winston')
+
+const COMPONENTS_PATH = '_components/' // relative to views path
+const COMPONENT_EXT = 'njk'
+
+function ComponentExtension (env) {
```
feedmypixel: Semantics but it feels like using `class` here is more inline with what I have seen around the app

config/nunjucks.js
```diff
@@ -0,0 +1,67 @@
...
+  this.run = function run (context, data, args) {
+    const name = data
+    const ctx = args || {}
+
+    let result = ''
+
+    try {
+      result = env.render(`${COMPONENTS_PATH}${changeCase.paramCase(name)}.${COMPONENT_EXT}`, ctx)
+    } catch (e) {
+      if (e.message.indexOf('template not found') > -1) {
```
tyom: `e.message.includes('template not found')`?

config/nunjucks.js
```diff
@@ -0,0 +1,67 @@
...
+  const env = nunjucks.configure([
+    `${config.root}/src/views`,
+    `${config.root}/node_modules/@uktrade/trade_elements/dist/nunjucks`
+  ], {
+    autoescape: true,
+    express: app,
+    watch: config.isDev
+  })
+
+  // Custom filters
```
feedmypixel: you could abstract this into a function and then the name of the function `addCustomFilters` would remove the need for the comment

```
function addCustomFilters () {
  const tradeElementsFilters = require(`${config.root}/node_modules/@uktrade/trade_elements/dist/nunjucks/filters`)
  const dataHubFilters = require(`${config.root}/src/nunjucks/filters`)

  const filters = Object.assign(dataHubFilters, tradeElementsFilters)

  Object.keys(filters).forEach((filterName) => {
    env.addFilter(filterName, filters[filterName])
  })
}
```

then

```
// src/nunjucks/filters

let filters = {}

filters.stringify = JSON.stringify

module.exports = filters

```

config/nunjucks.js
```diff
@@ -0,0 +1,67 @@
...
+    const args = parser.parseSignature(null, true)
+
+    parser.advanceAfterBlockEnd(tok.value)
+
+    return new nodes.CallExtension(this, 'run', args)
+  }
+
+  this.run = function run (context, data, args) {
+    const name = data
+    const ctx = args || {}
```
tyom: Could initialise `args` param with default `{}` in function signature.

config/nunjucks.js
```diff
@@ -0,0 +1,67 @@
...
+
+    let result = ''
+
+    try {
+      result = env.render(`${COMPONENTS_PATH}${changeCase.paramCase(name)}.${COMPONENT_EXT}`, ctx)
```
feedmypixel: do we need the `changeCase` dependency here? Tighten up what we accept then no need to convert to paramCase. The component name needs to be the filename to work.

teneightfive: Could drop it. This was actually from a previous project where the components could also come from an API response so wanted to always convert to the expected case of the frontend. Could have a use for that here in the future too but can drop it for now.

config/nunjucks.js
```diff
@@ -0,0 +1,67 @@
+const nunjucks = require('nunjucks')
+const changeCase = require('change-case')
```
tyom:
```js
const kebabCase = require('lodash/kebabCase')
```
No additional dependencies as lodash is already required.

config/nunjucks.js
teneightfive: I decided to leave that there as it wasn't specifically related to nunjucks configuration. It's more for express' view rendering.

config/nunjucks.js
teneightfive: As discussed probably okay to leave as is as relates specifically to the components section.

config/nunjucks.js
teneightfive: I think if we start to add more filters locally within this app we should extract it out into its own module but I think it's okay to leave here for now as the majority are coming externally from trade elements anyway. 

Maybe we can review the stringify filter use and see if we can then remove the need for any local filters.

config/nunjucks.js
teneightfive: Have removed this dep as don't really need it in this scenario.

feedmypixel:
It would also be good to have this documented in the `README`. Just what is in the PR description would be great, under a `nunjucks` > `components` section.

README.md
```diff
@@ -46,14 +46,14 @@ To start the server just:
 
     docker-compose up
 
-The server starts in developer mode, which means that when you make local changes it will auto-compile 
-sass or javavscript, and will restart nodejs when server side changes are made. A container with redis will also start, this is linked to the data hub container.
```
feedmypixel: @teneightfive are you using the `.editorconfig`?

teneightfive: Yep. Not sure why it doesn't respect the settings...

config/nunjucks.js
```diff
@@ -0,0 +1,65 @@
...
+    `${config.root}/src/views`,
+    `${config.root}/node_modules/@uktrade/trade_elements/dist/nunjucks`
+  ], {
+    autoescape: true,
+    express: app,
+    watch: config.isDev
+  })
+
+  // Custom filters
+  const filters = require(`${config.root}/node_modules/@uktrade/trade_elements/dist/nunjucks/filters`)
```
tyom: Do you need `${config.root}/node_modules` bit? Could you just import by name `@uktrade/trade_elements/dist/nunjucks/filters`?

config/nunjucks.js
```diff
@@ -0,0 +1,65 @@
...
+  this.parse = function parse (parser, nodes) {
+    const tok = parser.nextToken()
+    const args = parser.parseSignature(null, true)
+
+    parser.advanceAfterBlockEnd(tok.value)
+
+    return new nodes.CallExtension(this, 'run', args)
+  }
+
+  this.run = function run (context, data, ctx = {}) {
```
tyom: Should `ctx` be called `locals` to differentiate it from the first parameter `context`?

PR #83, opened 2017-05-18
Google Tag Manager

feedmypixel:
- add google tag manager snippet 
- add google tag manager no script snippet
- refactor `GOOGLE_TAG_MANAGER` env name

### DIFF
--- a/src/views/layouts/ukti.njk
+++ b/src/views/layouts/ukti.njk
@@ -4,18 +4,22 @@
 {% block service_name %}Data Hub{% endblock %}
 
 {% block head_styles %}
-<!-- This is a test -->
 <link href="{{ asset_path | default('/') }}css/application.css" media="screen" rel="stylesheet" />
 {% endblock %}
 
 {% block head_script %}
- {% if user %}
+  {% include "_includes/google-tag-manager-snippet.njk" %}
+  {% if user %}
     <script>
       if (!window.DIT) window.DIT = {};
       window.DIT.user = {{ user | stringify | safe }};
     </script>
   {% endif %}
     <script src="{{ asset_path }}javascripts/common.js?{{ startTime }}"></script>
+{% endblock %}
+
+{% block body_start %}
+  {% include "_includes/google-tag-manager-no-script.njk" %}
 {% endblock %}
 
 {% block content %}
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ This file expects the following environment variables:
 | POSTCODE_KEY | Part of the frontend looks up addresses for postcodes using [getaddress.io](https://getaddress.io/). Obtain a key for the service and set it here |
 | API_CLIENT_ID | Half the credentials needed to talk to the back end |
 | API_CLIET_SECRET | The second half of the credentials needed to talk to the backend |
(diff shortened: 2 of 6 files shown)


feedmypixel:
@jim68000 FYI, as spoke about this PR involves a change of the env `GOOGLE_TAG_MANAGER` -> `GOOGLE_TAG_MANAGER_KEY`

src/views/layouts/ukti.njk
```diff
@@ -4,12 +4,12 @@
 {% block service_name %}Data Hub{% endblock %}
 
 {% block head_styles %}
-<!-- This is a test -->
 <link href="{{ asset_path | default('/') }}css/application.css" media="screen" rel="stylesheet" />
 {% endblock %}
 
 {% block head_script %}
- {% if user %}
+  {% include "includes/googleTagManagerSnippet.njk" %}
```
teneightfive: I'm wondering if we even need to have these as includes?

src/views/includes/googleTagManagerSnippet.njk
```diff
@@ -0,0 +1,11 @@
+{% if googleTagManagerKey %}
```
teneightfive: I find it useful to prefix folders for includes, macros, components, etc with something like an underscore so it keeps shared files grouped away from theme based folders within the view folder. I've started to use that approach for the [components PR](https://github.com/uktrade/data-hub-fe-beta2/pull/82) so might be worth following?

src/views/layouts/ukti.njk
tyom: Good point. It's only ever included in one place, isn't it?

teneightfive: Yep, that was my thought. It can still be useful having includes even if they're only included once if they remove large, self-contained chunks of code from a layout but I'm not sure if this would be fine within the base layout on its own.

Maybe move to an include if we start including lots of tracking blocks further down the line.

feedmypixel: After discussion with @teneightfive the abstraction into includes is for communication of what this is rather than because it is used in multiple places. I tried it straight in the template but it looks quite busy. Or at least what it is is communicated better via the include file name.

src/views/includes/googleTagManagerSnippet.njk
feedmypixel: nice idea changed to `_includes`

feedmypixel:
Also changed filenames to kebab case

src/views/includes/googleTagManagerSnippet.njk
feedmypixel: I wonder if this a symptom of having `macros`, `includes`, `templates`, `components` in the views folder? Maybe a future folder structure could be
```
views
└── _layouts
└── _partials
│       └── components
│       └── includes
│       └── templates
└── company
└── contact
...




```

feedmypixel:
> @jim68000 FYI, as spoke about this PR involves a change of the env GOOGLE_TAG_MANAGER -> GOOGLE_TAG_MANAGER_KEY

this has been changed in the relevant environments and will sit silent until this work has been released

PR #91, opened 2017-05-19
contact add form missing company name

feedmypixel:
- expose `companyUrl` to the view layer
- refactor of `getViewCompanyLink` method
- add tests around `buildCompanyUrl`
- add `npm test:watch` task

screen shot 2017-05-19 at 12 35 13

### DIFF
--- a/src/controllers/company-archive.controller.js
+++ b/src/controllers/company-archive.controller.js
@@ -10,7 +10,7 @@ function archiveCompany (req, res, next) {
   Q.spawn(function * () {
     try {
       const company = yield companyRepository.getDitCompany(req.session.token, req.params.id)
-      const url = companyService.getViewCompanyLink(company)
+      const url = companyService.buildCompanyUrl(company)
       const reason = (req.body.archived_reason !== 'Other') ? req.body.archived_reason : req.body.archived_reason_other
 
       if (reason.length > 0) {
@@ -32,7 +32,7 @@ function unarchiveCompany (req, res, next) {
   Q.spawn(function * () {
     try {
       const company = yield companyRepository.getDitCompany(req.session.token, req.params.id)
-      const url = companyService.getViewCompanyLink(company)
+      const url = companyService.buildCompanyUrl(company)
       yield companyRepository.unarchiveCompany(req.session.token, company.id)
       req.flash('success-message', 'Updated company record')
       res.redirect(url)
--- a/src/services/company.service.js
+++ b/src/services/company.service.js
@@ -99,21 +99,26 @@ function getCompanyForSource (token, id, source) {
 }
 
 /**
- * Pass an API formatted company record in and return a url to view that company
- * depending on it's type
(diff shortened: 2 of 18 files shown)


src/services/company.service.js
```diff
@@ -99,21 +99,26 @@ function getCompanyForSource (token, id, source) {
...
+  const companyPath = '/company/view'
+  const businessType = company.business_type && company.business_type.name.toLowerCase()
+  let urlPath
+
   if (!company.uk_based) {
-    return `/company/view/foreign/${company.id}`
-  } else if (company.business_type.name.toLowerCase() === 'private limited company' || company.business_type.name.toLowerCase() === 'public limited company') {
-    return `/company/view/ltd/${company.id}`
+    urlPath = `${companyPath}/foreign/${company.id}`
+  } else if (businessType === 'private limited company' || businessType === 'public limited company') {
```
teneightfive: Could this just check for `businessType` containing `limited company` or are there other options that would pass that check?

feedmypixel: @ztolley has the knowledge if that would work

tyom: Could just return the URL for each condition instead of creating `urlPath`.

```js
if (!company.uk_based) {
  return `...`
}
if (businessType === 'private limited company' /* ... */) {
  return `...`
}
```

feedmypixel: Yeah that is similar to what was there. I quite like the return being at the bottom of a method, it being the last thing, making it a little easier to read.

More of a preference thing.

ztolley: I think that they are the only forms of limited.

src/controllers/company-archive.controller.js
```diff
@@ -10,7 +10,7 @@ function archiveCompany (req, res, next) {
   Q.spawn(function * () {
     try {
       const company = yield companyRepository.getDitCompany(req.session.token, req.params.id)
-      const url = companyService.getViewCompanyLink(company)
+      const url = companyService.buildCompanyUrl(company)
```
ztolley: build company url is a little none specific, there are lots of urls relating to company, like edit company, view company contacts, though you could argue it's the default url for a company. Not a blocker though

src/services/company.service.js
ztolley: that would work

ztolley: Doesn't matter either way, this way could be argued as easier to step through a debug.

tyom: I agree to disagree 😄 

I prefer return early pattern. IMO it's easier to read, fewer lines and no jumping `if/else if/else` when scanning the code.

src/controllers/company-archive.controller.js
feedmypixel: 👍 Yeah I would go with it being the default. To then have the following if needed
- `buildEditCompanyUrl`
- `buildCompanyContactsUrl`

PR #94, opened 2017-05-19
Investment Project Details page

tyom:
This adds the details page for investment project.

Local URL: http://localhost:3000/investment/f22ae6ac-b269-4fe5-aeba-d6a605b9a7a7/details

Note, some of the investment project values are currently missing.


src/controllers/investment-details.controller.js
```diff
@@ -0,0 +1,67 @@
...
+const localNavItems = [
+  { label: 'Project details', slug: 'details' },
+  { label: 'Client', url: 'client' },
+  { label: 'Project team', url: 'team' },
+  { label: 'Interactions', url: 'interactions' },
+  { label: 'Documents', url: 'documents' },
+  { label: 'Evaluation', url: 'evaluation' },
+  { label: 'Audit history', url: 'audit' }
+].map(item => {
+  item.id = kebabCase(item.label)
```
feedmypixel: snazzy, but you could just hardcode the `id` in? (sorry boring old man here)

src/controllers/investment-details.controller.js
```diff
@@ -0,0 +1,67 @@
...
+        tab: 'project-details'
+      })
+    } catch (error) {
+      winston.error(error)
+      next(error)
+    }
+  })
+}
+
+router.get('/investment/:id', (req, res) => res.redirect(`/investment/${req.params.id}/details`))
```
feedmypixel: can this be abstracted into a method, purely for using the name on the method to show what it does?

src/views/investment/_base.njk
```diff
@@ -0,0 +1,34 @@
...
+
+  {% block projectStatus %}
+    {{ projectStatus(project) }}
+  {% endblock %}
+
+  <div class="grid-row">
+    {% block localNav %}
+      <nav class="local-nav column-one-quarter">
+        {% for item in localNavItems %}
+          <a href="/investment/{{project.id}}/{{ item.slug }}" class="local-nav__link{% if tab == item.id %} is-active{% endif %}">
```
feedmypixel: suggestion `tab` changed to `currentNavItem`

src/controllers/investment-details.controller.js
tyom: This is a remnant from previous implementation. I'll just use the `slug` instead.

src/views/investment/details.njk
```diff
@@ -1,44 +1,158 @@
...
+        </tr>
+      </tbody>
+    </table>
+    <p>
+      <a href="/investment/{{ project.id }}/details/edit" class="button-secondary">Edit summary</a>
+    </p>
 
-<h1>{{ projectTitle }}</h1>
+    <h2 class="heading-medium">Value</h2>
+    <table class="table--key-value  table--striped">
```
jim68000: There is a striped table macro in trade elements. Are you holding off using that until we do modularisation properly?

src/controllers/investment-details.controller.js
```diff
@@ -0,0 +1,67 @@
...
+
+      res.render('investment/details', {
+        project,
+        value,
+        requirements,
+        localNavItems,
+        tab: 'project-details'
+      })
+    } catch (error) {
+      winston.error(error)
```
teneightfive: You don't need to log this as it should already be being handled in the express error handler.

src/controllers/investment-details.controller.js
```diff
@@ -0,0 +1,67 @@
...
+    try {
+      const project = yield getInvestmentProjectSummary(req.session.token, req.params.id)
+
+      res.render('investment/details-edit', {
+        project,
+        localNavItems,
+        tab: 'project-details'
+      })
+    } catch (error) {
+      winston.error(error)
```
teneightfive: Same as above re error logging.

src/sass/_local-nav.scss
```diff
@@ -0,0 +1,25 @@
+.local-nav {
```
teneightfive: Are there not styles for the local nav already? Or is this part of moving to BEM based components?

teneightfive: It's also missing deps.

src/sass/_status-bar.scss
```diff
@@ -0,0 +1,42 @@
+@import 'colours/palette';
+@import 'typography';
```
teneightfive: Are these all the deps needed? Feels like its missing the dependencies for `%contain-floats`, and `visually-hidden` and the gutter vars.

src/views/investment/details.njk
tyom: I'm not convinced there's a point to abstract a simple HTML table like that.

src/views/investment/_macros.njk
```diff
@@ -0,0 +1,21 @@
+{% macro projectStatus(project) %}
```
teneightfive: Could this be done in the new component style instead of a macro?

src/controllers/investment-details.controller.js
feedmypixel: good catch... 💥 

you can just `} catch(next)`

src/sass/_local-nav.scss
tyom: > Are there not styles for the local nav already? Or is this part of moving to BEM based components?

There's [leftnav.scss](https://github.com/uktrade/data-hub-fe-beta2/blob/develop/src/sass/_leftnav.scss) but it's also styling `.right-content` which is not used on this page.

I'm using a standard grid with a simplified `local-nav` component here.

src/controllers/investment-details.controller.js
tyom: > you can just `} catch(next)`

I don't think you can do that. Syntax error.

src/controllers/investment-details.controller.js
```diff
@@ -0,0 +1,64 @@
...
+      next(error)
+    }
+  })
+}
+
+function redirectToDetails (req, res) {
+  return res.redirect(`/investment/${req.params.id}/details`)
+}
+
+router.get('/investment/:id', (req, res) => redirectToDetails)
```
tyom: That's a mistake on my part. Good spot!

src/views/_components/project-status.njk
```diff
@@ -0,0 +1,19 @@
+{% if project_code %}
+  <div class="status-bar grid-row">
+    <dl class="status-bar__section column-one-quarter">
+      <dt>Project code</dt>
+      <dd class="status-bar__section-title">
+        {{ project_code }}
+      </dd>
+      <dd>
+        <a href="#">Close project</a>
```
tyom: It's a placeholder. I'll remove.

README.md
```diff
@@ -150,6 +150,30 @@ To include a component and pass data to it:
...
+{% component 'person', personData, gender='male' %} 
+```
+
+Is the same as:
+
+```njk
+{% component 'person', {
+  name: 'Barry',
+  age: '55',
+  gender: 'male
```
teneightfive: Missing ending `'`

src/views/_components/project-status.njk
```diff
@@ -0,0 +1,34 @@
...
+ # @return {string} HTML for project status
+ #
+ # @example
+ #   {% component 'project-status', {
+ #     projectCode: 'DHP-0000001
+ #     phaseName: 'Prospect'
+ #   } %}
+ #}
+
+{% if projectCode %}
```
teneightfive: Do we need to check for the presence of this if it's a required param?

If it's always expected we shouldn't need this.

src/views/investment/_base.njk
```diff
@@ -0,0 +1,34 @@
...
+          </a>
+        {% endfor %}
+      </nav>
+    {% endblock %}
+
+    {% block mainContent %}
+    {% endblock %}
+  </div>
+
+  <script src="/javascripts/investment.bundle.js"></script>
```
teneightfive: Should we add this to the bottom of the layout. Can use the current `body_end_scripts` block.

src/views/_components/project-status.njk
tyom: One thing is to say it's required, another is to handle user error :)

I think it's good to handle cases when no sensible input is provided.

src/views/investment/_base.njk
tyom: This is probably something we could do we a more general layout refactor we have planned.

teneightfive: True. But will be easier to refactor if we're adding them all to the same block for now.

src/controllers/investment-details.controller.js
```diff
@@ -0,0 +1,103 @@
...
+    'R&D budget': data.r_and_d_budget,
+    'Non-FDI R&D project': data.non_fdi_r_and_d_budget,
+    'New-to-world tech': data.new_tech_to_uk,
+    'Export revenue': data.export_revenue
+  }
+}
+
+function formatRequirementsData (data) {
+  return {
+    'Main strategic driver': data.strategic_drivers,
```
jim68000: Strategic drivers is an array, so this should be pluralised

src/controllers/investment-details.controller.js
```diff
@@ -0,0 +1,103 @@
...
+    'Client requirements': data.client_requirements,
+    'Competitor countries': data.competitor_countries,
+    'Possible UK locations': data.uk_region_locations,
+    'Investment location': data.average_salary,
```
jim68000: average_salary for location? is this padding for a missing value?

tyom: Seems to be missing. I'll change to `null` to make it obvious.

README.md
```diff
@@ -150,6 +150,30 @@ To include a component and pass data to it:
 } %}
 ```
 
+Component can take multiple arguments. It will combine them in single object:
+
+```js
+res.render('some-page', {
+  personData: {
```
feedmypixel: has this been done for clarity of data when sending to the view/component? It feels a little magic to do this and will cause nested items to override top level properties of the same name?

If I sent through the following to a view 
```
{
  personData: {
    name: 'Barry',
    age: 55    
  }
}
```
I would expect to be able to access the following in the view:
```
personData.name
```
and not
```
name
```

src/controllers/investment-details.controller.js
```diff
@@ -0,0 +1,103 @@
...
+
+const {buildCompanyUrl} = require('../services/company.service')
+const {
+  getInvestmentProjectSummary,
+  getInvestmentValue,
+  getInvestmentRequirements
+} = require('../repos/investment.repo')
+
+const localNavItems = [
+  { label: 'Project details', slug: 'details' },
```
feedmypixel: small thing but would aid visibility to line up properties here

src/controllers/investment-details.controller.js
```diff
@@ -0,0 +1,103 @@
...
+    'R&D budget': data.r_and_d_budget,
+    'Non-FDI R&D project': data.non_fdi_r_and_d_budget,
+    'New-to-world tech': data.new_tech_to_uk,
+    'Export revenue': data.export_revenue
+  }
+}
+
+function formatRequirementsData (data) {
+  return {
+    'Main strategic drivers': data.strategic_drivers,
```
feedmypixel: feels like the strings for the properties in `formatRequirementsData`, `formatValueData` and `formatProjectData` belong in the view layer

src/views/investment/_base.njk
feedmypixel: > will be easier to refactor if we're adding them all to the same block for now.

+1, I'd go with what have and then at least it is all consistent in being the way we don't want to do it

test/controllers/investment-details.controller.test.js
```diff
@@ -0,0 +1,109 @@
...
+          { label: 'Interactions', slug: 'interactions' },
+          { label: 'Documents', slug: 'documents' },
+          { label: 'Evaluation', slug: 'evaluation' },
+          { label: 'Audit history', slug: 'audit' }
+        ],
+        project: {
+          'Anonymous description': null,
+          'Business activity': undefined,
+          'Client': {
+            name: 'Samsung SDS',
```
feedmypixel: is this all mock data?

src/controllers/investment-details.controller.js
feedmypixel: > feels like the strings for the properties in formatRequirementsData, formatValueData and formatProjectData belong in the view layer

But I see why they are here maybe something that `Jira > DH-245` could address?

README.md
tyom: > has this been done for clarity of data when sending to the view/component?

It combines top level objects into one so a component could receive an object from locals with additional custom properties. e.g.
```js
const dish = {
  ingredients: ['mushrooms', 'carrot', 'onion'],
  preparation: 'wok'
}
```
```njk
{% component 'plate', dish, { salt: true, pepper: false} %}
```
is the same as:

```njk
{% component 'plate', {
  ingredients: ['mushroom', 'carrot', 'onion'],
  preparation: 'wok',
  salt: true,
  pepper: false
} %}
```

src/controllers/investment-details.controller.js
tyom: Feels like it's quite unconventional to line up properties in JS. That smells of Ruby.

src/views/investment/_base.njk
tyom: Bu it is consistent with how it's currently done. End of `main` block or some such.

src/controllers/investment-details.controller.js
tyom: Yes, it's a wider problem, but here I'm trying to get rid of dependencies like `labels/investment-label`.

feedmypixel: > Yes, it's a wider problem, but here I'm trying to get rid of dependencies like labels/investment-label.

+1

README.md
feedmypixel: > It combines top level objects into one so a component could receive an object from locals with additional custom properties. e.g.

+ 1

PR #95, opened 2017-05-19
fix circle ci integration

feedmypixel:
Fix circle integration

- show test results in **Test Summary** section in `CircleCi`
- make coverage report available in **Artifacts** section in `CircleCi`

You can see [the code coverage](https://531-83823675-gh.circle-artifacts.com/0/tmp/circle-artifacts.BqnjfA4/coverage/index.html) from the **Artifacts** section.

## Of Note
This is first cut, it feels like we should also be able to see code coverage in the test summary section, even if its just percentages which then click through to the above code report

screen shot 2017-05-19 at 17 30 10

screen shot 2017-05-19 at 17 31 03

screen shot 2017-05-19 at 17 27 15

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,7 @@ reports
 .env
 npm*.log
 stats.json
+
+.nyc_output
+coverage
+test-results.xml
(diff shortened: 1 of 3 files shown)


feedmypixel:
After reading a few things at the weekend I have a few things I'd like to add this, will re-raise later today

PR #97, opened 2017-05-19
Add support for conditional subfields

teneightfive:
This allows conditional content to be shown/hidden based on the
value of a particular field

It also starts to create a base javasript app file that is shared
by the whole application

conditional-subfield

### DIFF
--- a/src/javascripts/app.js
+++ b/src/javascripts/app.js
@@ -1,0 +1,3 @@
+const ConditionalSubfields = require('./modules/conditional-subfields')
+
+ConditionalSubfields.init()
--- a/src/javascripts/modules/conditional-subfields.js
+++ b/src/javascripts/modules/conditional-subfields.js
@@ -1,0 +1,134 @@
+/**
+ * Conditional sub fields
+ *
+ * Supports the ability to conditionally show and hide content based on the
+ * value of a field.
+ *
+ * The element to be shown requires 2 attributes:
+ * [data-controlled-by] - this is the name of the field which controls it
+ * [data-control-value] - the value by which this element should be displayed
+ *
+ * It also requires a basic bit of style to work:
+ *
+ * [data-controlled-by] {
+ *   display: none;
+ * }
+ * [data-controlled-by].is-active {
+ *   display: block;
+ * }
+ *
+ * Note: it is recommend to only apply this if javascript is enabled
+ *
+ * @example
+ *
+ * <label>
+ *   <input name="control-field" value="yes" type="radio">Yes
+ * </label>
+ * <label>
+ *   <input name="control-field" value="no" type="radio">No
+ * </label>
+ *
+ * <div class="js-ConditionalSubfield" data-controlled-by="control-field" data-control-value="yes">
+ *   Visible when `control-field` equals `yes`
+ * </div>
+ *
+ */
+
+const uniq = require('lodash/uniq')
+
+const { addClass, removeClass } = require('../../lib/element-stuff.js')
+
+const ConditionalSubfields = {
+  selector: 'js-ConditionalSubfield',
+  activeClass: 'is-active',
+  onChangeHandler: null,
+  controllers: [],
+  wrapper: null,
+
+  init (wrapper = document) {
+    this.wrapper = wrapper
+
+    this.cacheEls()
+
+    if (this.controllers.length) {
+      this.bindEvents()
+      this.render()
+    }
+  },
+
+  cacheEls () {
+    const conditionalSubFields = this.wrapper.getElementsByClassName(this.selector)
+
+    this.onChangeHandler = this.onChange.bind(this)
+    this.controllers = uniq([...conditionalSubFields].map((controller, i) => {
+      return controller.getAttribute('data-controlled-by')
+    }))
+  },
+
+  bindEvents () {
+    this.wrapper.addEventListener('change', this.onChangeHandler)
+  },
+
+  render () {
+    this.controllers.forEach((controller) => {
+      const fields = this.wrapper.querySelectorAll(`[name="${controller}"]`)
+
+      fields.forEach(this._handleField)
+    })
+  },
+
+  destroy () {
+    this.wrapper.removeEventListener('change', this.onChangeHandler)
+  },
+
+  onChange (evt) {
+    const field = evt.target
+
+    if (this.controllers.includes(field.name)) {
+      this._handleField(field)
+    }
+  },
+
+  _handleField (controlInput) {
+    const subFields = this.wrapper.querySelectorAll(`[data-controlled-by="${controlInput.name}"]`)
+
+    subFields.forEach((subField, i) => {
+      const controlInputValue = subField.getAttribute('data-control-value') + ''
+      let isVisible = false
+
+      if (controlInput.tagName === 'SELECT') {
+        isVisible = controlInput.value === controlInputValue
+      } else if (controlInputValue && controlInputValue !== '*' && controlInput.checked) {
+        isVisible = controlInput.value === controlInputValue
+      }
+
+      this._toggleSubField(subField, isVisible)
+    })
+  },
+
+  _toggleSubField (subField, isVisible) {
+    if (isVisible) {
+      addClass(subField, this.activeClass)
+    } else {
+      removeClass(subField, this.activeClass)
+    }
+
+    subField.setAttribute('aria-expanded', isVisible)
+    subField.setAttribute('aria-hidden', !isVisible)
+
+    if (!isVisible && !subField.getAttribute('data-persist-values')) {
+      const children = subField.querySelectorAll('input, select, checkbox, textarea')
+
+      children.forEach((field, i) => {
+        field.value = ''
+        field.checked = false
+
+        const event = this.wrapper.createEvent('HTMLEvents')
+        event.initEvent('change', true, false)
+        field.dispatchEvent(event)
+      })
+    }
+  }
+}
+
+module.exports = ConditionalSubfields
--- a/src/sass/application.scss
+++ b/src/sass/application.scss
@@ -22,6 +22,8 @@ $images-dir: '/images/';
 @import 'status-bar';
 @import 'tables';
 
+@import 'objects/conditional-subfields';
+
 @import 'trumps/layout';
 
 .table--readonly {
--- a/src/sass/objects/conditional-subfields.scss
+++ b/src/sass/objects/conditional-subfields.scss
@@ -1,0 +1,7 @@
+.js-enabled .js-ConditionalSubfield {
+  display: none;
+
+  &.is-active {
+    display: block;
+  }
+}
--- a/src/views/layouts/ukti.njk
+++ b/src/views/layouts/ukti.njk
@@ -9,13 +9,15 @@
 
 {% block head_script %}
   {% include "_includes/google-tag-manager-snippet.njk" %}
+
   {% if user %}
     <script>
       if (!window.DIT) window.DIT = {};
       window.DIT.user = {{ user | stringify | safe }};
     </script>
   {% endif %}
-    <script src="{{ asset_path }}javascripts/common.js?{{ startTime }}"></script>
+
+  <script src="{{ asset_path }}javascripts/common.js?{{ startTime }}"></script>
 {% endblock %}
 
 {% block body_start %}
@@ -44,4 +46,8 @@
   <div class="container">
     {% block main %}{% endblock %}
   </div>
+{% endblock %}
+
+{% block body_end_scripts %}
+  <script src="{{ asset_path }}javascripts/app.bundle.js?{{ startTime }}"></script>
 {% endblock %}
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -4,6 +4,7 @@ const prod = process.env.NODE_ENV === 'production'
 module.exports = {
   devtool: prod ? 'hidden-source-map' : 'source-map',
   entry: {
+    app: './src/javascripts/app.js',
     'company-investment-tab-form': './src/javascripts/company-investment-tab-form',
     'company-add': './src/javascripts/company-add',
     'company-edit': './src/javascripts/company-edit',


src/javascripts/app.js
```diff
@@ -0,0 +1,7 @@
+const $ = require('jquery')
```
ztolley: Do we really need jquery?

src/sass/application.scss
```diff
@@ -20,6 +20,9 @@ $images-dir: '/images/';
 @import 'splitheader';
 @import 'tables';
 
+// Objects
```
ztolley: Objects? So these are different to components ?

Also you can see it's an object in the path

src/javascripts/modules/conditional-subfields.js
```diff
@@ -0,0 +1,118 @@
...
+    } else if (controlInputValue && controlInputValue !== '*') {
+      $controlInput = $controlInput.filter(`[value="${controlInputValue}"]`)
+      $controlInput.attr('aria-controls', $field.attr('id'))
+      isVisible = $controlInput.is(':checked')
+    }
+
+    this._toggleField($field, isVisible)
+  },
+
+  _toggleField ($field, isVisible) {
```
ztolley: This was already in element-stuff   - hide - show

src/javascripts/modules/conditional-subfields.js
```diff
@@ -0,0 +1,118 @@
...
+    this.conditionalFields = $(this.el)
+  },
+
+  bindEvents () {
+    const controllers = $.unique(this.conditionalFields.map((i, controller) => {
+      return $(controller).data('controlled-by')
+    }))
+
+    $.each(controllers, (i, name) => {
+      $(`[name="${name}"]`).on('change.ConditionalSubfields', $.proxy(this.handleVisibility, this))
```
ztolley: You can use class properties instead of proxy or bind.  handleVisibility = (event) => {

src/javascripts/app.js
jim68000: Deeply unconvinced we need jQuery. As far as I can see it's used below as a replacement for query selector and each and for binding - as we can do that natively, I'd say we shouldn't use it

src/sass/application.scss
teneightfive: Yes. This is based on the [ITCSS](http://www.creativebloq.com/web-design/manage-large-css-projects-itcss-101517528) methodology. 

Fair point about the comment, probably doesn't need. Was just there more as a placeholder so it's clear where to add other objects as ITCSS requires components/units/objects/trumps/etc to be defined in a specific order.

src/sass/objects/conditional-subfields.scss
```diff
@@ -0,0 +1,7 @@
+.js-enabled [data-controlled-by] {
```
feedmypixel: could we just have a `js-hidden` class?

src/sass/objects/conditional-subfields.scss
```diff
@@ -0,0 +1,7 @@
...
+  display: none;
+
+  &.is-active {
```
feedmypixel: this `is-active` could be attached to a component

src/javascripts/modules/conditional-subfields.js
```diff
@@ -0,0 +1,118 @@
...
+ */
+
+const $ = require('jquery')
+
+const ConditionalSubfields = {
+  el: '[data-controlled-by]',
+  activeClass: 'is-active',
+
+  init () {
+    this.cacheEls()
```
feedmypixel: We need to check for the existence of the markup hook? and then return if its not there? rather than going into `bindEvents` or `render` methods and running the `each` to then not apply the code?

feedmypixel:
@teneightfive can you add a gif/screenshot of this component please for context

src/sass/objects/conditional-subfields.scss
teneightfive: We could but it then requires adding `js-hidden` to each element which felt messier. 

This way there is a default state for the component with a state class when open. Maybe `is-open` would be better than `is-active`?

src/sass/objects/conditional-subfields.scss
teneightfive: It didn't feel worthy of turning into a component in its own right. But if we refactor to remove this attribute anyway it may need some style to apply its default and open state.

src/javascripts/modules/conditional-subfields.js
teneightfive: Yes, that would save it running methods it doesn't need to.

teneightfive:
@feedmypixel 

>It feels like the inputs should be specifically controlling something, rather than an element being controlled by inputs.

I did have this thought originally, but it would mean adding support to all form components to pass some sort of object that would then display the correct element. It would also mean making sure that the target is applied to a specific form element within a fieldset, ie each radio/checkbox in a multi select to allow different values to show/hide multiple elements.

Moving this logic to the element being controlled made more sense when considering those options as it meant only applying that logic in one place rather than 2. Happy to have another rethink of how that might work though if we think it's a better route?

>The 2nd observation is I find the object literal way of doing this quite different to the way we are doing things on the back end with function declarations.

Am happy to change this to function declarations if we want to be consistent. Object literal made sense as didn't see a need to instantiate multiple instances of it. It also uses a set of methods based on [Heisenberg.js](https://github.com/Heisenbergjs/heisenberg) which I think reads really nicely if you structure the logic of your modules in a consistent way.

src/javascripts/app.js
feedmypixel: + 1 for no jQuery

src/javascripts/modules/conditional-subfields.js
feedmypixel: \+ 1 Feels like it would be a nice pattern for Frontend work, in an `init` method check for existence of elements, return if none

src/sass/objects/conditional-subfields.scss
feedmypixel: feels like there could become a lot of `display:hidden` specific component rules, when we could just have one `js-hidden`. It also communicates in the markup a little better

src/sass/objects/conditional-subfields.scss
feedmypixel: it feels like the default, no JavaScript is open, so potentially an `is-closed` class, but then that could just be an `is-hidden` class?

src/javascripts/modules/conditional-subfields.js
```diff
@@ -0,0 +1,129 @@
...
+  init () {
+    if (document.contains(document.querySelector(this.el))) {
+      this.cacheEls()
+      this.bindEvents()
+      this.render()
+    }
+  },
+
+  cacheEls () {
+    console.log('doing')
```
tyom: Forgot to remove?

src/javascripts/modules/conditional-subfields.js
```diff
@@ -0,0 +1,129 @@
...
+const uniq = require('lodash/uniq')
+
+const { addClass, removeClass } = require('../../lib/element-stuff.js')
+
+const ConditionalSubfields = {
+  el: '[data-controlled-by]',
+  activeClass: 'is-active',
+
+  init () {
+    if (document.contains(document.querySelector(this.el))) {
```
tyom: Could we avoid calling `querySelector` twice? Once here and once again in `cacheEls`?

Perhaps call `cacheEls` and then check the property of cached elements has length in `init` before binding events and rendering?

src/javascripts/modules/conditional-subfields.js
```diff
@@ -0,0 +1,129 @@
...
+      this.render()
+    }
+  },
+
+  cacheEls () {
+    console.log('doing')
+    const conditionalSubFields = document.querySelectorAll(this.el)
+
+    this.onChangeHandler = this.onChange.bind(this)
+    this.controllers = uniq(Array.prototype.slice.call(conditionalSubFields).map((controller, i) => {
```
tyom: Could we initialise all properties with default value. It would make it easier to see what properties are used and would give default values.

tyom: We're passing this through Babel. Could just spread it.
```js
uniq(...conditionalSubFields).map()
```

src/javascripts/modules/conditional-subfields.js
```diff
@@ -0,0 +1,129 @@
...
+      removeClass($subField, this.activeClass)
+    }
+
+    $subField.setAttribute('aria-expanded', isVisible)
+    $subField.setAttribute('aria-hidden', !isVisible)
+
+    if (!isVisible && !$subField.getAttribute('data-persist-values')) {
+      const $children = $subField.querySelectorAll('input, select, checkbox, textarea')
+
+      Array.prototype.slice.call($children).forEach((field, i) => {
```
tyom: Spread here too? In fact `NodeList` has `forEach` method so you don't even need to do that.

src/javascripts/modules/conditional-subfields.js
teneightfive: Yes, I think the thought was to prevent the other logic happening in `cacheEls` by calling that here. It checks for any presence of that selector first. Then runs the 3 functions.

Unless we also check in `cacheEls` before setting the other 2 properties. But then feels like we're doing the same check in 2 places?

src/javascripts/modules/conditional-subfields.js
```diff
@@ -0,0 +1,129 @@
...
+        isVisible = $controlInput.value === controlInputValue
+      }
+
+      this._toggleSubField($subField, isVisible)
+    })
+  },
+
+  _toggleSubField ($subField, isVisible) {
+    if (isVisible) {
+      addClass($subField, this.activeClass)
```
tyom: There's a `toggleClass` in `lib/element-stuff`. Would that work?

src/javascripts/modules/conditional-subfields.js
```diff
@@ -0,0 +1,129 @@
...
+    }))
+  },
+
+  bindEvents () {
+    document.addEventListener('change', this.onChangeHandler)
+  },
+
+  render () {
+    this.controllers.forEach((controller, i) => {
+      const $field = document.querySelectorAll(`[name="${controller}"]`)
```
tyom: Looks more like `$fields`.

src/javascripts/modules/conditional-subfields.js
tyom: If you initialise the property `cacheEls` caches to you can just check for its length before running other methods. Even if you don't, `querySelectorAll` always returns an array.

src/javascripts/modules/conditional-subfields.js
teneightfive: It doesn't support passing `state` so used an explicit add/remove style here.

src/javascripts/app.js
```diff
@@ -0,0 +1,14 @@
+/* eslint no-unused-vars: 0 */
+const ConditionalSubfields = require('./modules/conditional-subfields')
+
+function ready (fn) {
```
tyom: Where's this used?

src/javascripts/app.js
```diff
@@ -0,0 +1,14 @@
...
+  if (document.readyState !== 'loading') {
+    fn()
+  } else {
+    document.addEventListener('DOMContentLoaded', fn)
+  }
+}
+
+document.addEventListener('DOMContentLoaded', function () {
```
tyom: As this file is included at the bottom of the page before `</body>` DOM would already be loaded.

src/javascripts/modules/conditional-subfields.js
```diff
@@ -0,0 +1,131 @@
...
+      this.bindEvents()
+      this.render()
+    }
+  },
+
+  cacheEls () {
+    const conditionalSubFields = document.getElementsByClassName(this.selector)
+
+    this.onChangeHandler = this.onChange.bind(this)
+    this.controllers = uniq([...conditionalSubFields].map((controller, i) => {
```
tyom: `i` redundant?

src/javascripts/modules/conditional-subfields.js
```diff
@@ -0,0 +1,134 @@
...
+ *
+ * <div data-controlled-by="control-field" data-control-value="yes">
+ *   Visible when `control-field` equals `yes`
+ * </div>
+ *
+ */
+
+const uniq = require('lodash/uniq')
+
+const { addClass, removeClass } = require('../../lib/element-stuff.js')
```
feedmypixel: can we give a more descriptive name rather than stuff? `element-helpers.js`?

src/javascripts/modules/conditional-subfields.js
```diff
@@ -0,0 +1,134 @@
...
+
+const ConditionalSubfields = {
```
feedmypixel: Maybe part of another PR but worth documenting the use of [heisenberg](https://github.com/Heisenbergjs/heisenberg) for file structure?

src/sass/application.scss
feedmypixel: > Also you can see it's an object in the path

I'd agree with this. But we should also document the approach `ITCSS` or a flavour of

src/javascripts/modules/conditional-subfields.js
```diff
@@ -0,0 +1,134 @@
...
+  selector: 'js-ConditionalSubfield',
```
feedmypixel: does this need to be a property on the object? its just being used in once place

src/javascripts/modules/conditional-subfields.js
tyom: Heisenberg is dead. Last updated 4 years go.

src/javascripts/modules/conditional-subfields.js
```diff
@@ -0,0 +1,134 @@
...
+    }))
+  },
+
+  bindEvents () {
+    this.wrapper.addEventListener('change', this.onChangeHandler)
+  },
+
+  render () {
+    this.controllers.forEach((controller) => {
+      const $fields = this.wrapper.querySelectorAll(`[name="${controller}"]`)
```
feedmypixel: is the `$` convention a hangover from jQuery?

src/javascripts/modules/conditional-subfields.js
feedmypixel: I don't see `js-ConditionalSubfield` referenced in markup examples anywhere? should I?

src/javascripts/modules/conditional-subfields.js
```diff
@@ -0,0 +1,134 @@
...
+      this.bindEvents()
+      this.render()
+    }
+  },
+
+  cacheEls () {
+    const conditionalSubFields = this.wrapper.getElementsByClassName(this.selector)
+
+    this.onChangeHandler = this.onChange.bind(this)
+    this.controllers = uniq([...conditionalSubFields].map((controller, i) => {
```
feedmypixel: do we need to `uniq` this? what are we guarding for here? I feel if a developer puts something wrong in the view I.e two `data-controlled-by` elements with the same value then it should be corrected in the view, rather than just the first `data-controlled-by` element working

src/javascripts/modules/conditional-subfields.js
feedmypixel: you could `Array.from` this nodeList then you don't have to call Array.prototype methods to process it

src/javascripts/modules/conditional-subfields.js
teneightfive: By all means. Maybe a separate change. This library already existed so I was utilising it. 

It's basically our current equivalent to jQuery for some DOM manipulation at the moment. I'm still not convinced on the roll our own framework. Having something like a lightweight version of jQuery if we're doing a lot of DOM works gives us a better chance of avoiding bugs we haven't encountered in browsers we haven't tested in yet.

We're also not running any kind of browser testing which means we can't guarantee each change still works. I think the benefit of using community supported packages to keep our code smaller and more maintainable could soon outweigh a bandwidth argument. 

Even with jQuery right now our whole unminified document is still less than 100k.

src/javascripts/modules/conditional-subfields.js
teneightfive: Ha. Yes, maybe we need to decide on a JS module structure (or framework) going forward rather than doing everything adhoc on a view specific basis.

src/javascripts/modules/conditional-subfields.js
teneightfive: The idea of it is so that it's clear to see what the hook for this module is. In some cases it may be used in other parts of the module so it's more a pattern than a necessity.

teneightfive: It hasn't been used yet. There is a separate branch which I've been rebasing on top of this which includes use of this module. Wanted to keep this is a separate change so it was easier to understand rather than bundling in a new feature branch.

teneightfive: I should probably add it to the example in the comments

src/javascripts/modules/conditional-subfields.js
teneightfive: Yes. There are instances where we have multiple elements controlled by one field so we don't want to include the same field name multiple times in this object.

src/javascripts/modules/conditional-subfields.js
teneightfive: A little. Maybe we should drop, but I think it sometimes clearly identifies that the variable is an element. Should we drop?

src/javascripts/modules/conditional-subfields.js
tyom: Is that a SUIT CSS style class name?

src/javascripts/modules/conditional-subfields.js
teneightfive: Are you suggesting this:

```js
fields.forEach((field) => {
  this._handleField(field)
})
```

tyom:
```js
fields.forEach(this._handleField)
```
?

src/javascripts/modules/conditional-subfields.js
feedmypixel: > Ha. Yes, maybe we need to decide on a JS module structure (or framework) going forward rather than doing everything adhoc on a view specific basis.

yeah it would be worth getting the principles agreed and committed

feedmypixel: > Heisenberg is dead. Last updated 4 years go.

Lol! Admittedly I was blindly following @teneightfive advice. Plus I did enjoy Breaking Bad.

src/javascripts/modules/conditional-subfields.js
feedmypixel: > I should probably add it to the example in the comments

That works for me

src/javascripts/modules/conditional-subfields.js
tyom: > Plus I did enjoy Breaking Bad.
<details>
  <summary>Spoiler alert</summary>
  They both died round about the same time
</details>

src/javascripts/modules/conditional-subfields.js
feedmypixel: seeing where `js-ConditionalSubfield` is in the markup that makes sense. I was thinking `js-ConditionalSubfield` was a wrapper around everything

PR #99, opened 2017-05-22
Improve Investment Details page

tyom:


src/views/_components/key-value-table.njk
```diff
@@ -0,0 +1,31 @@
+{##
+ # @param {object} [items] An object of items with label as key
+ # @param {string} [variant] A table
+ #
+ # @return {string} HTML for key/value table
+ #
+ # @example
+ #   {% component 'key-value-table', variant='striped' items={
```
teneightfive: Missing comma after second arg.

src/views/_components/key-value-table.njk
```diff
@@ -0,0 +1,31 @@
...
+ #     'Key #1: '80%',
+ #     'Key #2: '20%'
+ #   } %}
+ #}
+
+{% if items|length %}
+  <table class="table--key-value{{ ' table--' + variant if variant }}">
```
teneightfive: Does this really need to be a table? Doesn't feel like tabular data. 

Maybe a `dl` would be a more suitable element.

src/views/_components/project-status.njk
```diff
@@ -1,15 +1,30 @@
-{% if project_code %}
+{##
+ # @param {string} [id] Project id
+ # @param {string} [name] Project name
+ # @param {string} [projectCode] *required* Project code
```
teneightfive: What's out pattern for documenting component variables?

I've previously used `(optional)` for data this isn't required and used the assumption everything else is required.

src/views/_components/project-status.njk
```diff
@@ -1,15 +1,30 @@
...
+ # @param {string} [phaseName] Name of project phase
+ #
+ # @return {string} HTML for project status
+ #
+ # @example
+ #   {% component 'project-status', {
```
teneightfive: `project-status` feels like a content specific name for this component.

Maybe `status-bar` to match the class names?

config/nunjucks.js
```diff
@@ -16,11 +16,12 @@ function ComponentExtension (env) {
     return new nodes.CallExtension(this, 'run', args)
   }
 
-  this.run = function run (context, name, locals = {}) {
+  this.run = function run (context, name, ...locals) {
```
teneightfive: If we're changing how we can call components we should document in the readme.

src/views/_components/key-value-table.njk
tyom: Just moving to existing component structure. We can decide whether this should be a table in a different PR.

config/nunjucks.js
tyom: Done.

tyom:
Moving these changes in #94.

PR #101, opened 2017-05-22
check for name value before lowercasing

feedmypixel:
Fix for error on home page. The position of this check had changed hence it was throwing this error

### DIFF
--- a/src/services/company.service.js
+++ b/src/services/company.service.js
@@ -107,7 +107,7 @@ function getCompanyForSource (token, id, source) {
  */
 function buildCompanyUrl (company) {
   const companyPath = '/company/view'
-  const businessType = company.business_type && company.business_type.name.toLowerCase()
+  const businessType = company.business_type && (company.business_type.name && company.business_type.name.toLowerCase())
   let urlPath
 
   if (!company.uk_based) {


src/services/company.service.js
```diff
@@ -107,7 +107,7 @@ function getCompanyForSource (token, id, source) {
  */
 function buildCompanyUrl (company) {
   const companyPath = '/company/view'
-  const businessType = company.business_type && company.business_type.name.toLowerCase()
+  const businessType = company.business_type && (company.business_type.name && company.business_type.name.toLowerCase())
```
tyom: What about:
```js
const businessType = lodash.get(company, 'business_type.name', null)
```

feedmypixel: Didn't realise we had lodash, was toying with the idea of 
```
const {business_type: { name }} = company.business_type.name
name.toLowerCase()
```

tyom: That doesn't give you default value.

ztolley: The problem is though that the name HAS to be there otherwise it can't build a link. So when getting the data from the server we now need to go through each of the contacts, get it's company and business type , and go look at the metadata to pad it out to get the name.

I still can't understand how it worked in the first place if the data isn't the response.

It's not working until you can login, goto the homepage and click on a company link for a contact

feedmypixel: > I still can't understand how it worked in the first place if the data isn't the response.

the check for this moved position out of the if and to the top of the method.

feedmypixel:
> That doesn't give you default value.

true - it will just assign `undefined`

feedmypixel:
> Doesn't fix the problem

Do you have any more info?

PR #102, opened 2017-05-22
Company exports

ztolley:
Based on a design by Steph before she left. The Original screen included an export account manager, which does not seem to exist, and the countries they are exporting to and want to export to.

No knows that the client relationship manager is, and it might be something completely different, so left his out for now.

companyexports

Kept it simple for now and we can fin


src/controllers/company-export.controller.js
```diff
@@ -0,0 +1,100 @@
...
+const exportDetailsLabels = {
+  exportToCountries: 'Currently exporting to',
+  futureInterestCountries: 'Future countries of interest'
+}
+
+function toArray (formValue) {
+  if (!formValue) {
+    return []
+  } else if (Array.isArray(formValue)) {
+    return formValue.filter(item => item.length > 0)
```
jim68000: Is this actually a toArray() function? It looks more like it's detecting something that is an array and filtering empty values/ Maybe rename?

src/controllers/company-export.controller.js
```diff
@@ -0,0 +1,100 @@
...
+ *  HTTP Get call to get export data relating to a company
+ */
+function view (req, res, next) {
+  Q.spawn(function * () {
+    try {
+      const company = res.locals.company
+
+      res.locals.exportDetails = {
+        exportToCountries: company.export_to_countries.map(country => country.name).join(', '),
+        futureInterestCountries: company.future_interest_countries.map(country => country.name).join()
```
jim68000: Missing ', ' in the join?

src/lib/controller-utils.js
```diff
@@ -109,8 +109,7 @@ function flattenIdFields (data) {
       if (Array.isArray(fieldValue)) {
         // Scan through the array of values, strip out any that are null, empty or have a null or empty id
         data[fieldName] = fieldValue
-          .filter(item =>
-            ((item && item.id && item.id !== null && item.id.length > 0) || (item !== null && item.length > 0)))
+          .filter(item => (item && item.id))
```
jim68000: Happy as I am to see the removal of the test on the left, what's the reason for trimming off all the other nullity tests? Do we filter stuff further upstream?

src/services/company.service.js
```diff
@@ -106,19 +106,23 @@ function getCompanyForSource (token, id, source) {
...
-  if (!company.uk_based) {
-    urlPath = `${companyPath}/foreign/${company.id}`
-  } else if (businessType.includes('limited company')) {
-    urlPath = `${companyPath}/ltd/${company.id}`
-  } else {
-    urlPath = `${companyPath}/ukother/${company.id}`
-  }
+  try {
+    const companyPath = '/company/view'
+    const businessType = company.business_type && company.business_type.name.toLowerCase()
```
jim68000: If company.business_type is undef here businessType is also undef, which leads to a ``` Cannot read property 'includes' of undefined ``` error in the tests below. We should probably have a sensible default here instead.

ztolley: this may be fixed with Ben's fix

src/lib/controller-utils.js
ztolley: @tyom informed me on another piece of code that the were not needed

src/controllers/company-export.controller.js
```diff
@@ -0,0 +1,100 @@
...
+}
+
+function common (req, res, next) {
+  Q.spawn(function * () {
+    try {
+      res.locals.tab = 'exports'
+      res.locals.company = yield companyService.getInflatedDitCompany(req.session.token, req.params.id)
+      companyService.getCommonTitlesAndlinks(req, res, res.locals.company)
+      next()
+    } catch (error) {
```
teneightfive: This can just be `catch (next)`

src/controllers/company-export.controller.js
```diff
@@ -0,0 +1,100 @@
...
+
+      res.locals.exportDetails = {
+        exportToCountries: company.export_to_countries.map(country => country.name).join(', '),
+        futureInterestCountries: company.future_interest_countries.map(country => country.name).join()
+      }
+
+      res.locals.exportDetailsLabels = exportDetailsLabels
+      res.locals.exportDetailsDisplayOrder = ['exportToCountries', 'futureInterestCountries']
+
+      res.render('company/exports-view')
```
teneightfive: What's the reason for using `res.locals` as opposed to [passing locals directly to the view](https://expressjs.com/en/api.html#res.render) using the express `res.render` method?

teneightfive: Doesn't setting `res.locals` mean that they would also be available to any other middleware called after setting them?

src/controllers/company-export.controller.js
```diff
@@ -0,0 +1,100 @@
...
+    } catch (error) {
```
teneightfive: Same as above, can just be `catch (next)`

src/controllers/company-export.controller.js
```diff
@@ -0,0 +1,100 @@
...
+      const company = yield companyRepo.getDitCompany(req.session.token, req.params.id)
+      controllerUtils.flattenIdFields(company)
+      const postData = Object.assign({}, company, {
+        export_to_countries: toArray(req.body.export_to_countries),
+        future_interest_countries: toArray(req.body.future_interest_countries)
+      })
+
+      yield companyRepo.saveCompany(req.session.token, postData)
+      res.redirect(`/company-exports/view/${req.params.id}`)
+    } catch (errors) {
```
teneightfive: Same as above

src/javascripts/country-list.js
```diff
@@ -0,0 +1,45 @@
+/* eslint no-new: 0 */
+const { insertAfter } = require('../lib/element-stuff')
+const guid = require('@uktrade/trade_elements').guid
```
teneightfive: Can use destructuring to be consistent with line above:

```js
const { guid } = require('@uktrade/trade_elements')
```

src/services/company.service.js
```diff
@@ -110,15 +110,18 @@ function buildCompanyUrl (company) {
...
   let urlPath
 
-  if (!company.uk_based) {
-    urlPath = `${companyPath}/foreign/${company.id}`
-  } else if (businessType.includes('limited company')) {
-    urlPath = `${companyPath}/ltd/${company.id}`
-  } else {
-    urlPath = `${companyPath}/ukother/${company.id}`
-  }
+    if (!company.uk_based) {
```
teneightfive: Is this missing the opening `try` line?

src/services/company.service.js
```diff
@@ -110,15 +110,18 @@ function buildCompanyUrl (company) {
...
+    } else if (businessType.includes('limited company')) {
+      urlPath = `${companyPath}/ltd/${company.id}`
+    } else {
+      urlPath = `${companyPath}/ukother/${company.id}`
+    }
 
-  return urlPath
+    return urlPath
+  } catch (error) {
+    return '/'
```
teneightfive: Should we use the new method for handling errors and send the error to the `next` function?

src/views/company/exports-edit.njk
```diff
@@ -0,0 +1,61 @@
+{% extends "./index.njk" %}
+{% import "macros/trade.html" as trade %}
+
+{% block tabmain %}
+{% include "messages.njk" %}
```
teneightfive: Should be indented in line with rest of block contents.

src/views/company/exports-view.njk
```diff
@@ -0,0 +1,19 @@
+{% extends "./index.njk" %}
+{% import "macros/trade.html" as trade %}
+
+{% block tabmain %}
+{% include "messages.njk" %}
```
teneightfive: Also needs indenting.

test/controllers/company-export.controller.test.js
```diff
@@ -0,0 +1,297 @@
...
+
+    controllerUtils = require(`${root}/src/lib/controller-utils`)
+    sinon.spy(controllerUtils, 'flattenIdFields')
+
+    company = require('../data/api-response-intermediary-company.json')
+
+    company.export_to_countries = [{id: '1234', name: 'France'}, {id: '2234', name: 'Italy'}]
+    company.future_interest_countries = [{id: '4321', name: 'Germany'}]
+
+    companyService.getInflatedDitCompany = sinon.stub().resolves(company)
```
teneightfive: Might be worth using `sinon.sandbox.create()` in a `beforeEach` block.

webpack.config.js
```diff
@@ -14,7 +14,8 @@ module.exports = {
     index: './src/javascripts/index',
     search: './src/javascripts/search',
     'service-delivery': './src/javascripts/service-delivery',
-    'archive-form': './src/javascripts/archive-form'
+    'archive-form': './src/javascripts/archive-form',
+    'country-list': './src/javascripts/country-list'
```
teneightfive: Feels like we should start moving to a set of bundles rather than lots of page specific files dotted around different views.

src/controllers/company-export.controller.js
ztolley: You only use catch(next) in a promise chain, not regular try catch

src/controllers/company-export.controller.js
ztolley: You only use catch(next) in a promise chain, not regular try catch

src/controllers/company-export.controller.js
ztolley: You only use catch(next) in a promise chain, not regular try catch

test/controllers/company-export.controller.test.js
ztolley: Only learnt about that after the tests were done

webpack.config.js
ztolley: indeed, but not in this ticket

tyom:
Somehow my JS lint fix commit is in your branch even though it's been merged to develop. Have you rebased recently?

src/controllers/company-export.controller.js
```diff
@@ -0,0 +1,121 @@
...
+  if (containsFormData(req)) {
+    data.export_to_countries = req.body.export_to_countries
+    data.future_interest_countries = req.body.future_interest_countries
+  } else {
+    data.export_to_countries = res.locals.company.export_to_countries.map(country => country.id)
+    data.future_interest_countries = res.locals.company.future_interest_countries.map(country => country.id)
+  }
+
+  data.exportDetailsLabels = exportDetailsLabels
+  data.countryOptions = countryOptions
```
tyom: Could initialise `data` above with these.

src/javascripts/country-list.js
```diff
@@ -0,0 +1,45 @@
+/* eslint no-new: 0 */
```
tyom: Needed?

tyom:
I don't seem to get any countries in select boxes.

test/javascripts/country-list.test.js
```diff
@@ -0,0 +1,69 @@
...
+    <div class="form-group">
+      <label class="form-label-bold">Test dropdown</label>
+      <select name="test" class="form-control">
+        <option value="">Select an option</option>
+        <option value="one">One</option>
+        <option value="two">Two</option>
+      </select>
+      <p><a class="add-button" href="addcountry">Add another country</a></p>
+    </div>
+  </div`
```
tyom: Missing `>`.

test/javascripts/country-list.test.js
```diff
@@ -0,0 +1,69 @@
+/* eslint no-new: 0 */
```
tyom: Needed?

src/views/company/exports-edit.njk
```diff
@@ -0,0 +1,61 @@
...
+      {% else %}
+        {{ trade.dropdown("export_to_countries",
+          label="Specify if this company is currently exporting to a market (optional)",
+          emptyLabel="Select country",
+          id="export_to_countries_new",
+          value="",
+          options=countryOptions)
+        }}
+      {% endif %}
+      <p><a class="add-button" href="addcountry">Add another country</a></p>
```
tyom: Does it work without JS?

src/controllers/company-export.controller.js
teneightfive: Yes, my bad.

test/javascripts/country-list.test.js
ztolley: Appears standard changed their rules on that. Removing

src/controllers/company-export.controller.js
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  return [formValue]
+}
+
+/**
+ * Common method for view and edit to gather information required for company layout
+ *
+ */
+function common (req, res, next) {
```
teneightfive: I've been trying a different method for handling post and get with the same context which feels slightly neater. Basically all of the things required for the view are set in the get handler and then the post handler processes the form and either redirects if successful or sets errors on `res.locals` and calls the `next()` function.

Means we don't need this shared middleware which gets called on both. What do you think?

src/sass/_shame.scss
```diff
@@ -248,3 +248,19 @@ hr {
     }
   }
 }
+
+#company-exports {
+  th {
+    white-space: nowrap;
```
teneightfive: Feels like this could be placed somewhere else rather than shame. Maybe set on a table unit as it's probably a property we want on all tables.

src/sass/_shame.scss
```diff
@@ -248,3 +248,19 @@ hr {
...
+  }
+}
+
+
+// In country lists, hide all but the first label
+.country-list > .form-group > .form-label-bold {
+  display: block;
+}
+
+.country-list > .form-group ~ .form-group .form-label-bold {
```
teneightfive: Should these not be wrapped in a `.js-enabled` class?

src/views/company/exports-edit.njk
```diff
@@ -0,0 +1,61 @@
+{% extends "./index.njk" %}
+{% import "macros/trade.html" as trade %}
+
+{% block tabmain %}
+  {% include "messages.njk" %}
+  <h2 class="heading-medium">An an export summary</h2>
+  {{ trade.errorPanel( errors, labels=exportDetailsLabels ) }}
```
teneightfive: Maybe give things at the same indentation level some space to breath so it's easier to scan the file.

src/views/company/exports-edit.njk
```diff
@@ -0,0 +1,61 @@
...
+  <form method="POST">
+    <input type="hidden" name="id" value="{{company.id}}"/>
+    <input type="hidden" name="_csrf_token" value="{{csrfToken}}"/>
```
teneightfive: The name of this has changed to just `_csrf`.

teneightfive: Also mentioned to tyom that it's good to give variables a space either side of the braces. I find it's easier to read that way.

src/sass/_shame.scss
ztolley: 99% of the time it's ok to wrap the title but for a couple of tables, including this one, it really looks odd to split it over 2 lines. When we create better tables we could revisit

src/javascripts/country-list.js
```diff
@@ -0,0 +1,46 @@
...
+
+  /**
+   * Handle selection of a country on a new country dropdown
+   *
+   * When the add country button is pressed, clone the last country dropdown
+   * and reset it's value.
+   *
+   * @memberof AddCountry
+   */
+  handleAddCountry (event) {
```
teneightfive: Is this something we can make more generic? Feels like we will have other examples where we'll need to "add another".

src/javascripts/country-list.js
```diff
@@ -0,0 +1,46 @@
...
+const { guid } = require('@uktrade/trade_elements')
+
+class CountryList {
+  constructor (wrapperElement) {
+    this.wrapperElement = wrapperElement
+    this.addEventHandlers()
+  }
+
+  addEventHandlers () {
+    this.wrapperElement.querySelector('.country-list__add')
```
teneightfive: We should use `js-` prefixed classes for javascript hooks and not hook onto classes used for style.

src/javascripts/country-list.js
```diff
@@ -0,0 +1,46 @@
...
+    const newFormControl = newFormGroup.querySelector('.form-control')
+    newFormControl.setAttribute('id', newId)
+    newFormControl.value = ''
+
+    insertAfter(newFormGroup, lastFormGroup)
+  }
+}
+
+if (typeof document !== 'undefined') {
+  document.querySelectorAll('.country-list-js')
```
teneightfive: Would be better to move the js bit to a prefix: `js-`

src/sass/_country-list.scss
```diff
@@ -0,0 +1,17 @@
+// In country lists, hide all but the first label
+.country-list > .form-group > .form-label-bold {
+  display: block;
+}
+
+.country-list > .form-group ~ .form-group .form-label-bold {
```
teneightfive: Should this be nested within `.js-enabled`?

src/sass/application.scss
```diff
@@ -6,6 +6,7 @@ $images-dir: '/images/';
 // application specific
 @import 'address';
 @import 'card';
+@import 'country-list';
```
teneightfive: This name feels too tied to the content. Is it really specific to country? Feels like it's more of an add pattern than a country pattern.

ztolley:
Country list is a dropdown to be consistent with other country drop downs in screens like company edit and contact edit.

src/views/company/exports-edit.njk
```diff
@@ -0,0 +1,64 @@
...
+      {% else %}
+        {{ trade.dropdown("future_interest_countries",
+          label="Future countries of interest if known (optional)",
+          emptyLabel="Select country",
+          id="future_interest_countries_new",
+          value="",
+          options=countryOptions)
+        }}
+      {% endif %}
+      <input class="country-list__add" type="submit" name="addFutureInterestCountry" value="Add country">
```
teneightfive: Mentioned in the scss too but this classname feels too tied to its contents. The pattern is more about adding another. This could even just be a modifier on `button` to style more like a link.

src/views/company/exports-view.njk
```diff
@@ -0,0 +1,18 @@
...
+  <h2 class="heading-medium">Exports</h2>
+  {{ trade.keyvaluetable(exportDetails, id="company-exports", labels=exportDetailsLabels, keyorder=exportDetailsDisplayOrder) }}
+  <p class="actions">
+    <a href="/company-exports/edit/{{company.id}}" class="button button-secondary">Edit summary</a>
+  </p>
+
+  <div class="section">
+    <h2 class="heading-medium">Export wins</h2>
+    <p>Data Hub does not yet cater for logging Export Wins. You can
+      log an export win for this company <a href="http://www.exportwins.service.trade.gov.uk/">here</a>.
```
teneightfive: "Click here" could possibly be improved. Maybe:

```
Data Hub does not currently support recording Export Wins. <a href="">Record an export win</a> for this company.
```

src/javascripts/country-list.js
ztolley: for every element within a component you are adding an event handler?

src/sass/_country-list.scss
ztolley: Nope

src/sass/application.scss
ztolley: I'll change it to add another dropdown or add another control

src/views/company/exports-view.njk
ztolley: Original design had the whole line a link. I'll change the wording to use 'click here' if that is what is takes to get this merged

teneightfive: I didn't mean to use _click here_. I put a suggestion of something that reads clearer above and that would also make the link more descriptive if somebody using a screenreader was listening to all the links on a page.

src/javascripts/country-list.js
teneightfive: If the javascript module is trying to hook onto a specific element it should be clear from the markup that changing or moving it would affect that JS module. 

If you hook onto a class that is used for styling, if somebody changes that style based class names or moves them to a different element it's not clear that it would affect anything in javascript.

So no, not every element, only elements that require a hook in the javascript.

src/views/company/exports-view.njk
ztolley: The wording there then doesn't make sense, you tell the user they can't add an export win and then in the next sentence you tell them they can, but don't say they are going to another place.. I'd rather leave it as it is for now and bring it up with content and see if we can improve the wording.

src/sass/application.scss
ztolley: Not sure what to name it. It's not just a select list of a value list, it would need a name to indicate it's a list of  shared values for a name that only needs one visible label.

src/lib/controller-utils.js
jim68000: OK

src/views/company/exports-view.njk
teneightfive: I think it makes more sense than a _click here_ link. I was thinking more in terms of a screenreader reading through the links on a page only. That text describes what is going to happen. You could add (you are going to leave this site) to be more specific but I really don't think the current text is clear. And worth addressing now because it may not be picked up for a while.

package.json
```diff
@@ -59,6 +59,7 @@
     "build": "npm run clean && npm run sass:compile && npm run js",
     "test": "mocha --require test/common.js --ui bdd test/**/*.test.js",
     "test:watch": "npm test -- -w",
+    "coverage": "nyc --reporter=html --reporter=text --reporter=lcov mocha --require test/common.js --ui bdd test/**/*.test.js",
```
teneightfive: I wouldn't add this to this change for now. Feels unrelated and better as a separate PR.

src/javascripts/add-another-field.js
```diff
@@ -0,0 +1,44 @@
...
+const { guid } = require('@uktrade/trade_elements')
+
+class AddAnotherField {
+  constructor (wrapperElement) {
+    this.wrapperElement = wrapperElement
+    this.addEventHandlers()
+  }
+
+  addEventHandlers () {
+    this.wrapperElement.querySelector('.js-add_button')
```
teneightfive: Would be better to be consistent with case style. Below using kebab case and this is using a combination of snake and kebab.

src/views/company/exports-edit.njk
```diff
@@ -0,0 +1,64 @@
...
+      {% else %}
+        {{ trade.dropdown("export_to_countries",
+          label="Specify if this company is currently exporting to a market (optional)",
+          emptyLabel="Select country",
+          id="export_to_countries_new",
+          value="",
+          options=countryOptions)
+        }}
+      {% endif %}
+      <input class="button--link js-add_button" type="submit" name="addExportToCountry" value="Add country">
```
teneightfive: Thinking about it, shouldn't we be adding this button with javascript? Am assuming without JS it is redundant? Would also mean it doesn't need another js class and would be easier to hook on to?

src/javascripts/add-another-field.js
```diff
@@ -0,0 +1,44 @@
...
+   *
+   *
+   * @memberof AddAnotherField
+   */
+  handleAddFormGroup (event) {
+    event.preventDefault()
+    event.target.blur()
+
+    const newId = guid()
+    const formGroups = this.wrapperElement.querySelectorAll('.form-group')
```
teneightfive: Still using a lot of style based classes within this module. Would be better to not bind things to style classes. Will cause issues if we use a different naming convention for components.

src/views/company/exports-edit.njk
ztolley: Without js this submits the form and tells the controller to add another country so it is used.

src/javascripts/add-another-field.js
ztolley: I worry there will be alot of duplication if we have one set of classes for js and another for styling

src/javascripts/add-another-field.js
ztolley: What would you call it?

src/views/company/exports-edit.njk
teneightfive: Ah okay, yep then keep it here.

src/javascripts/add-another-field.js
teneightfive: But the point isn't about duplication. You can argue that BEM duplicates a lot of characters but it brings with it scope. 

The idea for not using style based classes in JS is that it keeps the two things separate. So that if the elements or class names used in the markup to apply the presentation layer change, the javascript functionality doesn't have to be changed as well or worse, end up broken.

PR #103, opened 2017-05-22
Fix 500 on search

feedmypixel:
Send errors to generic error middleware, update tests where necessary.

### DIFF
--- a/src/middleware/user.js
+++ b/src/middleware/user.js
@@ -6,18 +6,18 @@ module.exports = (req, res, next) => {
   const user = req.session.user
 
   if (token && !user) {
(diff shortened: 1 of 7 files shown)


src/middleware/user.js
```diff
@@ -15,9 +15,8 @@ module.exports = (req, res, next) => {
 
       res.locals.user = req.session.user
       next()
-    }).catch((error) => {
-      res.render('error', { error })
     })
+      .catch(next)
```
teneightfive: If we're using this style of indentation maybe make the `.then()` match and drop it onto a new line and indent the block.

PR #104, opened 2017-05-22
Fix for clean on windows

ztolley:
To make the tool build on windows all that was needed was to update the clean script.

Git hooks still dont work and short of moving the logic into a js script I'm not sure it could be OS agnostic.

### DIFF
--- a/src/services/company.service.js
+++ b/src/services/company.service.js
@@ -106,19 +106,21 @@ function getCompanyForSource (token, id, source) {
  * @returns {string} urlPath
  */
 function buildCompanyUrl (company) {
-  const companyPath = '/company/view'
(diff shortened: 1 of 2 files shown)


src/services/company.service.js
```diff
@@ -106,19 +106,21 @@ function getCompanyForSource (token, id, source) {
...
+      urlPath = `${companyPath}/foreign/${company.id}`
+    } else if (businessType.includes('limited company')) {
+      urlPath = `${companyPath}/ltd/${company.id}`
+    } else {
+      urlPath = `${companyPath}/ukother/${company.id}`
+    }
 
-  return urlPath
+    return urlPath
+  } catch (error) {}
```
teneightfive: Should we not return the error?

PR #109, opened 2017-05-23
Move CSRF token out of controllers

### LINKED ISSUE #87: Tidy up CSRF work
At the moment there is a [genCSRF](https://github.com/uktrade/data-hub-fe-beta2/blob/develop/src/lib/controller-utils.js#L122-L127) utility method that is called within a controller that [requires CSRF](https://github.com/uktrade/data-hub-fe-beta2/blob/8a37b17463faaea98023768bd350339a15c543e6/src/controllers/investment-start.controller.js#L14)

There is also the [csrf application level middleware](https://github.com/uktrade/data-hub-fe-beta2/blob/develop/src/server.js#L130) that is a middleware check placed on all routes within the application.

## Suggestion
Move these two things into one middleware that is then placed on routes that require csrf tokens. 
- We then don't run middleware on routes that don't require it
- Explicitly set the csrf requirement on the route rather than in the route handler method
- Have one place for csrf work

```
router.get('/investment/start', csrf, getHandler)
router.post('/investment/start', csrf, postHandler)
```

teneightfive:
CSRF tokens are set by calling a function within controllers that
require them. This changes moves that logic to an app level
middleware that sets a CSRF token on the `res.locals objects` which
is then available in a view.

It currently sets the token globally making it accessible to all views. We
could look to switch this to only calling the csrf middleware on controllers
that need it though.

Closes #87


src/controllers/company-add.controller.js
```diff
@@ -2,7 +2,7 @@ const express = require('express')
 const Q = require('q')
 const winston = require('winston')
 const { ukOtherCompanyOptions, foreignOtherCompanyOptions } = require('../options')
-const { isBlank, toQueryString, genCSRF } = require('../lib/controller-utils')
+const { isBlank, toQueryString } = require('../lib/controller-utils')
```
tyom: Do we need to remove `genCSRF` function from `controller-utils`?

jim68000: Do we need to newline this as 

```
const { isBlank,
toQueryString } = require('../lib/controller-utils')
```

As we've done elsewhere, or is two too few?

teneightfive: We don't have a specific pattern as far as I'm aware. It's normally something dictated by the linter but we don't have a rule around this. 

I sometimes do it based on line length. Happy to go with the majority. I'd go more with this if we split onto multiple lines though:

```js
const {
  isBlank,
  toQueryString
} = require('../lib/controller-utils')
```

teneightfive: It has been: https://github.com/uktrade/data-hub-fe-beta2/pull/109/files#diff-8e1b32bba2191e4b068dc3124f4860df

tyom: I think it's short enough in this case. No need to split.

test/middleware/csrf-token.test.js
```diff
@@ -0,0 +1,25 @@
...
+    this.csrfMiddleware = require(`${root}/src/middleware/csrf-token`)()
+    this.sandbox = sinon.sandbox.create()
+  })
+
+  afterEach(() => {
+    this.sandbox.restore()
+  })
+
+  describe('set a CSRF token', () => {
+    it('should set the csrf token on response locals object', () => {
```
feedmypixel: can we also test for an error being thrown with incorrect `csrfToken`

teneightfive:
### Note

Decided to keep the setting of the csrf token as a global middleware for now given the amount of controllers that currently use forms. This avoids the extra bloat in the majority of controllers in this app. This could be changed in the future however.

PR #111, opened 2017-05-24
refactor and fix for the way doc_type is passed to the search endpoint

feedmypixel:
The way `doc_type` was being passed to the api `/search` endpoint was causing a 500 error. 
This work fixes that issue and also refactors the server side and client side code around the selection of facets on the `/search` page.

There was also a bug where you could add multiple `doc_type` to the url path this has also been fixed.

## Of Note
@elcct has a pr up on [data-hub-leeloo](https://github.com/uktrade/data-hub-leeloo/pull/212) that:
>It also adds 3 search endpoints for basic and detailed search - specs are available in data-hub-api-spec repository.

Spec information in [search.yml](https://github.com/uktrade/data-hub-api-spec/blob/master/generator/specs/paths/search.yml)

## TODO
Once this has been released we can send `company` and `contact` to the api which means more complexity can be removed from the frontend code:
- The [buildSearchFilters](https://github.com/uktrade/data-hub-fe-beta2/compare/feature/DH-258-refactor-doc_type-passed-to-search?expand=1#diff-b20e67cec3704e32aa6b08c2a7a007cfR10) can be removed. 
- The [buildFacetViewDataObj](https://github.com/uktrade/data-hub-fe-beta2/compare/feature/DH-258-refactor-doc_type-passed-to-search?expand=1#diff-d59a93984165d0bd91213bc63b7cdfa2R60) can be updated.

search

### DIFF
--- a/src/controllers/search.controller.js
+++ b/src/controllers/search.controller.js
@@ -7,15 +7,26 @@ const Q = require('q')
 
 const router = express.Router()
 
-function get (req, res, next) {
-  const filters = Object.assign({}, req.query)
-  delete filters.term
-  delete filters.page
+function buildSearchFilters (query) {
+  const docType = query.doc_type
+  const filters = []
 
+  if (docType === 'company') {
+    filters.push('company_company', 'company_companieshousecompany')
+  } else if (docType) {
+    filters.push(docType)
+  }
+
+  return filters
+}
+
+function get (req, res, next) {
   if (!req.query.term || req.query.term.length === 0) {
     res.render('search/facet-search', { result: null, pagination: null, params: req.query })
     return
   }
+
+  const filters = buildSearchFilters(req.query)
 
   searchService.search({
     token: req.session.token,
@@ -25,7 +36,11 @@ function get (req, res, next) {
   })
     .then((result) => {
       const pagination = getPagination(req, result)
-      res.render('search/facet-search', { result, pagination, params: req.query })
+      res.render('search/facet-search', {
+        result,
+        pagination,
+        params: req.query
+      })
     })
     .catch(next)
 }
--- a/src/javascripts/facets.js
+++ b/src/javascripts/facets.js
@@ -1,5 +1,5 @@
 /* eslint new-cap: 0 */
-const getQueryParam = require('../lib/url-stuff').getQueryParam
+const {getQueryParam, buildQueryString} = require('../lib/url-helpers')
 const {toggleClass, show} = require('../lib/element-stuff')
 
 const term = getQueryParam('term')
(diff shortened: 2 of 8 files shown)


tyom:
Seems a bit strange that we change the search value when switching filters (substituting space with +).

src/javascripts/facets.js
```diff
@@ -13,53 +13,56 @@ class Facets {
...
+    this.collapseButtons.forEach((collapseButton) => {
+      collapseButton.addEventListener('click', this.toggleFacet, false)
+    })
   }
 
-  selectOptionHandler = () => {
-    let url = `?term=${term}`
+  selectOptionHandler = (event) => {
+    const input = event.target || event.srcElement
+    let queryParams = {
```
tyom: Can be a `const`. You're not reassigning it.

src/javascripts/facets.js
```diff
@@ -13,53 +13,56 @@ class Facets {
...
+    const queryParams = {
+      term: decodeURIComponent(term)
+    }
 
-  clearFacetSelection = (event) => {
-    const facetWrapper = event.target.parentElement.parentElement.parentElement
-    const checkedInputs = facetWrapper.querySelectorAll('input[type=checkbox]:checked')
-    for (let pos = 0; pos < checkedInputs.length; pos += 1) {
-      const input = checkedInputs.item(pos)
+    this.checkboxInputs.forEach((input) => {
```
tyom: You probably don't need to reset each input as you're refreshing the page with the new query string anyway.

src/javascripts/facets.js
```diff
@@ -13,53 +13,56 @@ class Facets {
...
-    toggleClass(facetWrapper, 'collapse')
+    window.location.href = buildQueryString(queryParams)
+  }
 
-    const control = facetWrapper.querySelector('.collapse-filter-js')
-    toggleClass(control, 'fa-chevron-up')
-    toggleClass(control, 'fa-chevron-down')
+  toggleFacet = () => {
+    toggleClass(this.collapsibleElem, 'collapse')
+    toggleClass(this.controlElem, 'fa-chevron-up')
```
tyom: Maybe off-topic, but do we even need this expand/collapse functionality on category filter box?

src/javascripts/facets.js
```diff
@@ -13,53 +13,56 @@ class Facets {
 
   initElements (targetElement) {
     this.element = targetElement
-    this.clearButtons = this.element.querySelectorAll('.clear-filter-js')
-    this.collapseButtons = this.element.querySelectorAll('.collapse-filter-js')
+    this.clearButtons = Array.from(this.element.querySelectorAll('.clear-filter-js'))
+    this.collapseButtons = Array.from(this.element.querySelectorAll('.collapse-filter-js'))
+    this.checkboxInputs = Array.from(this.element.querySelectorAll('input[type=checkbox]'))
+    this.controlElem = this.element.querySelector('.collapse-filter-js')
```
tyom: Should the class name start with `js-` as well to be consistent?

src/controllers/search.controller.js
```diff
@@ -7,16 +7,27 @@ const Q = require('q')
 
 const router = express.Router()
 
-function get (req, res, next) {
-  const filters = Object.assign({}, req.query)
-  delete filters.term
-  delete filters.page
+function buildSearchFilters (query) {
+  const docType = query.doc_type && query.doc_type
+  let filters = []
```
tyom: Can be a `const`.

src/controllers/search.controller.js
```diff
@@ -7,16 +7,27 @@ const Q = require('q')
 
 const router = express.Router()
 
-function get (req, res, next) {
-  const filters = Object.assign({}, req.query)
-  delete filters.term
-  delete filters.page
+function buildSearchFilters (query) {
+  const docType = query.doc_type && query.doc_type
```
tyom: Double checking?

src/javascripts/facets.js
feedmypixel: That means the clearing will then rely on the page being refreshed, which feels a little odd

src/javascripts/facets.js
feedmypixel: agreed - FYI this is all being redesigned

src/controllers/search.controller.js
feedmypixel: if its not falsey then assign it

feedmypixel:
> Seems a bit strange that we change the search value when switching filters (substituting space with +).

yeah that is because the initial query is not encoded, but running the filters uses `buildQueryString` that then encodes the query string

src/javascripts/facets.js
tyom: You're probably right, but since we don't store the state in the URL at this point it's safe to assume refresh means reset.

src/controllers/search.controller.js
tyom: Do you need to check it before assignment?

tyom:
> yeah that is because the initial query is not encoded, but running the filters uses buildQueryString that then encodes the query string

Could we do something about that. Changing user's input without explanation is wrong.

feedmypixel:
> Could we do something about that. Changing user's input without explanation is wrong.

So interestingly we are using a `get` for what should really be a `post` on `/search` I'll refactor this in the morning and add it to this PR

src/controllers/search.controller.js
feedmypixel: actually (just opened IDE to look at this) this is left over from old code - no need for excessive checking!

feedmypixel:
> Could we do something about that. Changing user's input without explanation is wrong.

done the url is now being ran through `encodeURI`

src/controllers/search.controller.js
```diff
@@ -7,16 +7,27 @@ const Q = require('q')
...
 
-function get (req, res, next) {
-  const filters = Object.assign({}, req.query)
-  delete filters.term
-  delete filters.page
+function buildSearchFilters (query) {
+  const docType = query.doc_type
+  const filters = []
+
+  if (docType && docType === 'company') {
```
tyom: You could just check `docType === 'company'` without checking if its truthy first.  `(undefined === 'company') === false`.

PR #112, opened 2017-05-24
Add CSRF token to login form

tyom:
Log in page is missing CSRF token and therefore failing.


src/views/login.njk
```diff
@@ -28,6 +28,7 @@
   {{ trade.errorPanel( messages.formErrors ) }}
 
   <form action="{{ action }}" method="post">
+    <input type="hidden" name="_csrf" value="{{csrfToken}}">
```
teneightfive: Maybe add space between the brackets and the name though.

PR #115, opened 2017-05-25
add the yarn error log to gitignore

feedmypixel:
Does what it says on the tin...


ztolley:
Personally I have it on my global ignore list

PR #118, opened 2017-05-25
Investment project creation

teneightfive:
This adds the create screen for an investment project.

It includes additions to the investment repo and investment
service as well as the create controller.

department of international trade

### DIFF
--- a/src/controllers/investment-create.controller.js
+++ b/src/controllers/investment-create.controller.js
@@ -1,7 +1,9 @@
 const router = require('express').Router()
 
 const { getInflatedDitCompany } = require('../services/company.service')
-const { getCompanyInvestmentProjects } = require('../repos/investment.repo')
+const { getCompanyInvestmentProjects, saveInvestmentProject } = require('../repos/investment.repo')
+const { transformForApi } = require('../services/investment-formatting.service')
+const metadataRepo = require('../repos/metadata.repo')
 
 function getHandler (req, res, next) {
   const equityCompanyId = req.query['equity-company']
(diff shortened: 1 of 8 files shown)


src/views/investment/create.njk
```diff
@@ -1,31 +1,222 @@
...
+        }}
+
+        {{ trade.textarea('description',
+          value=form.state['description'],
+          label='Description')
+        }}
+
+        {% component 'form-multiple-choice', {
+          variant: 'radio',
+          label: 'Have we signed a non-disclosure agreement (NBA) with the client?',
```
reupen: Typo in NBA (should be NDA).

test/controllers/investment-start.controller.test.js
```diff
@@ -48,6 +48,24 @@ describe('Investment start controller', () => {
...
+    describe('when no company ID is present', () => {
+      it('should render company details and search', (done) => {
+        this.controller.getHandler({
+          session: {
+            token
+          },
+          query: {}
+        }, {
+          render: (template, data) => {
+            expect(data.clientCompany).to.be.undefined
```
tyom: You may wanna wrap it in a `try/catch` block otherwise you'll get obscure "Timeout of 2000ms exceeded" when tests fail.

```js
render: (template, data) => {
  try {
    expect(data.clientCompany).to.be.undefined
    expect(data.clientCompanyInvestments).to.be.undefined
    expect(data.showSearch).to.equal(true)
    done()
  } catch (e) {
    done(e)
  }
}
```

src/services/investment-formatting.service.js
```diff
@@ -55,4 +55,58 @@ function getClosedInvestmentProjects (investmentProjects) {
...
+  if (body['is-relationship-manager']) {
+    project['client_relationship_manager'] = body['is-relationship-manager']
+  }
+
+  if (body['is-referral-source']) {
+    project['referral_source_advisor'] = body['is-referral-source']
+  }
+
+  if (body['nda_signed']) {
+    project['project_shareable'] = false
```
jim68000: Isn't project shareable semi-independent from the NDA field? IIRC if you say NDA not signed, _then_ you get a radio pair that says whether the project is shareable.  This may have changed.

teneightfive: Yes, but the API returns an error if you don't set a value for `project_shareable`. So this sets it to false if the NDA has been signed. 

Maybe that's something we need to change on the API side actually? `project_shareable` should either be set to `false` if `nda_signed` is `true` or it shouldn't be required.

src/controllers/investment-create.controller.js
```diff
@@ -17,14 +19,59 @@ function getHandler (req, res, next) {
...
+        }
+      })
+
+      const investmentTypes = metadataRepo.investmentTypeOptions.map((type) => {
+        return {
+          value: type.id,
+          label: type.name
+        }
+      })
+      const form = res.locals.form || {}
```
tyom: New line?

src/services/investment-formatting.service.js
reupen: I thought project shareable was being brought up to the top level? That's why it's required at the moment...

teneightfive: @reupen I assumed that if we've signed an NDA then it's not shareable with our partners?

That field is to set whether we can share it with partners and therefore only applies if we didn't need to sign an NDA.

reupen: Logically speaking, I don't see why the NDA couldn't say it can be shared with our partners. Anyway, the markings on the design up on the wall seem to have it brought out to the top level.

teneightfive: Will bring that out to the same level.

PR #122, opened 2017-05-25
Use our own layout template and styles to improve modularity

tyom:
- Create new Nunjucks base layout template and derived sub templates
- Include new styles for the new layout
- Update the views to use the new layout

### How new inheritance works
New layout inheritance


README.md
```diff
@@ -174,6 +174,43 @@ Is the same as:
...
+  - `body_footer` - wraps site footer container (inside body > footer)
+  - `body_footer_content` - contains content inside site footer
+
+### Base template variables
+
+Base layout checks for certain variables.
+
+- `siteTitle` {string} - name of the site. Defaults to 'Department for International Trade'.
+- `serviceTitle` {string} - name of the service.
+- `phaseBanner` {string} - phase of the project. Possible values: `alpha` and `beta`.
```
teneightfive: If it's expecting a _phase_ maybe change this variable to `phase` instead of `phaseBanner`.

Do we even need to include this as a variable if it's not being used anywhere else? Could it just be a block? Or an include loaded into a block?

src/sass/environment/_global.scss
```diff
@@ -0,0 +1,52 @@
...
+    background-color: $focus-colour;
+    outline: 3px solid $focus-colour;
+    text-decoration: none;
+  }
+}
+
+input,
+textarea {
+  &:focus {
+    outline: 3px solid $yellow;
```
teneightfive: Should this use `$focus-colour` like the above properties?

src/sass/environment/_global.scss
```diff
@@ -0,0 +1,52 @@
...
+  min-height: 100vh;
+}
+
+.u-visuallyhidden,
+.u-visually-hidden {
+  @include visually-hidden;
+}
+
+a {
+  color: $govuk-blue;
```
teneightfive: Do we have semantic variable names for link states?

teneightfive: Maybe use `$link-colour` and `$link-hover-colour`

src/sass/layout/_global-footer.scss
```diff
@@ -0,0 +1,98 @@
...
+  @include site-width-container;
+  padding-top: $default-spacing-unit * 4;
+  padding-bottom: $default-spacing-unit * 4;
+}
+
+.global-footer__links,
+.global-footer__copyright {
+  @include media(tablet) {
+    display: inline-block;
+    margin-right: -4px;
```
teneightfive: Use `$default-spacing-unit`?

src/sass/layout/_global-header.scss
```diff
@@ -0,0 +1,40 @@
...
+@import '../environment/tools/mixins/typography';
+
+.global-header {
+  background-color: $black;
+  color: $white;
+}
+
+.global-header__inner {
+  @include site-width-container;
+  padding: 10px 0;
```
teneightfive: Magic numbers. Maybe use a combination of `$baseline-grid-unit` to help vertical rhythm.

src/sass/layout/_global-nav.scss
```diff
@@ -0,0 +1,31 @@
+@import '../environment/settings/colours';
+@import '../environment/tools/mixins/layout';
+
+.global-nav {
+  @include clearfix;
+  padding: 10px 0;
```
teneightfive: Also consider `$baseline-grid-unit` in this file

src/sass/layout/_main-content.scss
```diff
@@ -0,0 +1,12 @@
+@import '../environment/settings/colours';
+@import '../environment/tools/mixins/layout';
+
+.main-content {
+  background-color: $white;
+  min-height: 60vh;
+
+  main {
```
teneightfive: Styling an element doesn't feel like BEM. We should also look at a way to not restrict the page width.

src/sass/layout/_notification-banner.scss
```diff
@@ -0,0 +1,17 @@
+@import '../environment/settings/colours';
+@import '../environment/tools/mixins/layout';
+
+.notification-banner {
+  background-color: $light-blue-25;
+}
+
+.notification-banner--js-hidden {
```
teneightfive: Should this be a modifier or should it be a state class like `is-hidden` instead?

README.md
tyom: The idea is to use one string to add/change/remove project phase. Block is used to allow customisation/replacement. I think in this case it can work just as a variable.

So you think we should not rely on variables for layout?

src/sass/layout/_global-footer.scss
tyom: This is meant to [reset browser's default space size](https://css-tricks.com/fighting-the-space-between-inline-block-elements/#article-header-id-1). It's not really linked to `$default-spacing-unit` in any way. If that variable changes it could potentially break the layout.

src/sass/layout/_main-content.scss
tyom: True. But there can only be one `main` on page I felt it could be an exception.

Your thoughts @feedmypixel?

src/sass/layout/_notification-banner.scss
tyom: `is-hidden` is meant to be set by application at runtime. A utility class could be a better solution. I'll update.

ztolley:
Expanding cards broken, was the hidden class removed or renamed?

tyom:
> _messages.njk seems an add place for this file. Shouldn't this be a component and then use that in the layouts?

Potentially. But I tried to keep the focus on layout and leave the current template structure mostly intact (except for renaming the partials).

> Project-status ? Is this name too tied to the use case and not generic enough?

Potentially. Again it's out of scope for this change. It was introduced in Investment Details PR.

> Should component njk files start with _ or is that just layouts?

Components are all stored in `_component` directory. The main purpose of prefixing those files (or directories) is to differentiate them from page views.

> Remove the search header and ukti layouts then update pages that use the search header to switch to the regular header. The search bar has been removed from the header.

Not sure I follow. Could you explain it in more detail?

> Contact/_layout and company/_layout-edit.njk is inconsistent and will this make it hard for some IDE's to find the file via search (I know atom and visual studio code is ok but what about intellj?) Better to follow the convention used for things like company.repo.js or _contact-edit-layout.njk?

It's quite consistent in that they both start with `_layout` to convey the fact that they are layouts for those sections. In the case of companies they two diverge for the type of page. I'm not a big fan of adding the current directory into the file name. Another way to name those files I considered is to use type extension, e.g. `_base.layout.njk` or `company/_edit.layout.njk`.

In IntelliJ one can type the directory followed by slash to start search of files within that directory. That works well for me. @teneightfive, @feedmypixel  what are your thoughts?

tyom:
> Expanding cards broken, was the hidden class removed or renamed?

I'll take a look. Thanks for spotting.

src/sass/layout/_global-footer.scss
```diff
@@ -0,0 +1,98 @@
...
+  }
+}
+
+.global-footer__links {
+  @include media(tablet) {
+    width: 75%;
+  }
+}
+
+.global-footer__list {
```
teneightfive: Worth considering whether this style is actually an element within the global footer block or whether it could be a block in its own right. Maybe just `.list` with a modifier to remove the `list-style`.

src/sass/layout/_global-footer.scss
```diff
@@ -0,0 +1,98 @@
...
+  margin: 0;
+  padding: 0;
+
+  li {
+    margin-right: 1em;
+    display: inline-block;
+  }
+}
+
+.global-footer__open-government-licence {
```
teneightfive: Same with this, does the style applied tied to it being a child element of the footer block or could it be a block in its own right.

src/sass/trumps/_utility.scss
```diff
@@ -0,0 +1,18 @@
+.u-visuallyhidden,
```
teneightfive: Can drop this one.

src/sass/layout/_main-content.scss
feedmypixel: Id put the `display: block` and any base element work in a `elements/_base.scss` or `elements/_main.scss` with the element selector. But the width or any non generic element styles should be on a class selector in the relevant folder. Of which Layout feels correct. ITCSS, says `/objects` for layout, but that name has no real meaning here.

src/sass/layout/_phase-banner.scss
```diff
@@ -0,0 +1,32 @@
+@import '../environment/settings/colours';
```
feedmypixel: do these need to be in an `environment` folder? how about just `/settings` and `/tools`

src/sass/layout/_phase-banner.scss
```diff
@@ -0,0 +1,32 @@
...
+  padding-top: $baseline-grid-unit * 2;
+  padding-bottom: $baseline-grid-unit * 2;
+  border-bottom: 1px solid $grey-2;
+}
+
+// TODO: remove qualifier
+span.phase-banner__tag {
+  background-color: $govuk-blue;
+  display: inline-block;
+  padding: 3px 6px 2px;
```
feedmypixel: best a separate PR but we need to fix all these CSS magic numbers that are around

src/sass/trumps/_utility.scss
```diff
@@ -0,0 +1,18 @@
...
+.u-visually-hidden {
+  @include visually-hidden;
+}
+
+.u-hidden {
+  display: none;
```
feedmypixel: With these being trumps, should they all have `!important`?

README.md
```diff
@@ -174,6 +174,44 @@ Is the same as:
 } %}
 ```
 
+## Templates
+
+Templates extensively use Nunjuck's [template inheritance](https://mozilla.github.io/nunjucks/templating.html#template-inheritance).
```
feedmypixel: do we need "extensively" here?

README.md
```diff
@@ -174,6 +174,44 @@ Is the same as:
...
+There are several top level block which are used for injecting content during rendering. Each subsequent template
```
feedmypixel: `block` should be `blocks`

README.md
```diff
@@ -174,6 +174,44 @@ Is the same as:
...
+that extends the base layout can include additional blocks. 
+
```
feedmypixel: I'd say its worth getting your diagram from the PR in the README as it demonstrates the template hierarchy really well

src/sass/_shame.scss
```diff
@@ -12,10 +12,6 @@
 
 }
 
-#content {
-  min-height: 600px;
-}
-
 .data {
   .data__description {
```
feedmypixel: does this need to be nested?

src/sass/_shame.scss
```diff
@@ -248,3 +244,14 @@ hr {
     }
   }
 }
+
+.js-enabled .block-label.selection-button-radio {
```
feedmypixel: can you add a refactor `/TODO` on this? feels like it needs it

src/sass/application.scss
```diff
@@ -3,7 +3,15 @@ $images-dir: '/images/';
 @import 'trade-elements';
 @import 'font-awesome';
 
-// application specific
+@import 'environment/global';
+@import 'layout';
+@import 'trumps/utility';
+
+@import 'objects/conditional-subfields';
```
feedmypixel: are we happy with what the folder `object` means? any better wording?

src/sass/application.scss
```diff
@@ -3,7 +3,15 @@ $images-dir: '/images/';
 @import 'trade-elements';
 @import 'font-awesome';
 
-// application specific
+@import 'environment/global';
+@import 'layout';
+@import 'trumps/utility';
```
feedmypixel: should live with other trumps at the bottom
itcss-layers

src/sass/application.scss
```diff
@@ -3,7 +3,15 @@ $images-dir: '/images/';
 @import 'trade-elements';
 @import 'font-awesome';
 
-// application specific
+@import 'environment/global';
```
feedmypixel: I'd add a newline between sections

src/sass/environment/_global.scss
```diff
@@ -0,0 +1,55 @@
...
+  margin: 0;
+  padding: 0;
+  display: flex;
+  flex-flow: column;
+  min-height: 100vh;
+}
+
+a {
+  color: $link-colour;
+  text-decoration-skip: ink;
```
feedmypixel: nice

src/sass/environment/_global.scss
```diff
@@ -0,0 +1,55 @@
+@import "settings/colours";
+@import "settings/typography";
+@import "tools/mixins/addons";
+
```
feedmypixel: If we are going with a loose ITCSS approach this feels like it should be in the `/elements` folder either as individual files around concerns (bit OTT) or a `/elements/_base.scss`

src/sass/environment/settings/_colours.scss
```diff
@@ -0,0 +1,76 @@
...
+$green-50: #7fb299;
+$green-25: #cce0d6;
+$turquoise: #28a197;
+$turquoise-50: #95d0cb;
+$turquoise-25: #d5ecea;
+$light-blue: #2b8cc4;
+$light-blue-50: #96c6e2;
+$light-blue-25: #d5e8f3;
+
+// Semantic colour names
```
feedmypixel: I'm not convinced these add meaning. I still need to find the variable to see what colour it actually is...

src/sass/_shame.scss
ztolley: This is why it's in shame :/

src/sass/environment/settings/_layout.scss
```diff
@@ -0,0 +1,4 @@
+@import "./typography";
+
+$site-width: 960px;
+$gutter: $baseline-grid-unit * 8;
```
feedmypixel: thats a big gutter, is `32px` correct?

ztolley:
re my comment about search header. Short version, we should not be using the search header anymore, so update the pages that use it and remove it.

src/sass/environment/settings/_typography.scss
```diff
@@ -0,0 +1,105 @@
...
+$default-spacing-unit: $baseline-grid-unit * 4;
+$measure: 40em; // comfy reading width of ~ 55-65 characters
+
+$type-settings-map: (
+  16: -1,
+  20: 0,
+  24: 1,
+  36: 2,
+  64: 3,
+);
```
feedmypixel: These are funny type sizes I'd expect to see something more traditional, along the lines of:

- 12
- 14
- 18
- 24
- 30
- 36
- 48
- 60
- 72

Maybe for another PR?

ztolley:
Filenames, ok I dont mind not having the directory in the name, it was others I was worried about. Visual studio code is fine with it. not sure about using .layout.njk but see what people thing. Main concern is wanting to have a naming convention that is predictable. Makes it easy to add new files and also makes it easy to find things

src/sass/layout/_global-footer.scss
```diff
@@ -0,0 +1,98 @@
...
+    width: 75%;
+  }
+}
+
+.global-footer__list {
+  list-style: none;
+  margin: 0;
+  padding: 0;
+
+  li {
```
feedmypixel: needs a BEM className

src/sass/layout/_global-footer.scss
```diff
@@ -0,0 +1,98 @@
...
+    margin-right: 1em;
+    display: inline-block;
+  }
+}
+
+.global-footer__open-government-licence {
+  margin-top: 1em;
```
feedmypixel: are we `em` or `px`, need to be consistent. I vote for px then use [postcss-pxtorem](https://github.com/cuth/postcss-pxtorem) with fallback on for Ie8 (keeps px values)

src/sass/layout/_global-footer.scss
```diff
@@ -0,0 +1,98 @@
...
+
+  @include media(tablet) {
+    padding-left: 50px;
+  }
+
+  p {
```
feedmypixel: needs BEM className

src/sass/layout/_global-nav.scss
```diff
@@ -0,0 +1,31 @@
...
+
+.global-nav__navigation {
+  float: left;
+}
+
+.global-nav__account {
+  float: right;
+  color: $grey-1;
+
+  > li {
```
feedmypixel: needs BEM className

src/sass/layout/_global-footer.scss
tyom: Unless we have a use for a generic class like that I'd avoid premature generalisation.

src/sass/layout/_phase-banner.scss
tyom: I'd agree with you. Previously he had those styles under environment but I'm not convinced they should be. @teneightfive do you have a reason for them to remain under `environment/`?

src/sass/layout/_phase-banner.scss
tyom: These particular magic numbers are the tweaks to visual appearance of that particular element rendered using smaller font sizes.

README.md
tyom: Probably not. Just wanted to stress that we use `extend` and `super` and chained inheritance rather than a flatter variant often used with more blocks on main layout instead.

src/sass/_shame.scss
tyom: I'd rather not change `_shame.scss` unnecessarily. It's out of scope for this work. I removed that particular rule because it was interfering with the new changes.

src/sass/_shame.scss
tyom: The whole `_shame.scss` should go sooner rather than later.

src/sass/environment/settings/_colours.scss
tyom: Not convinced either. Just copied these over from existing styles.

src/sass/environment/settings/_typography.scss
```diff
@@ -0,0 +1,105 @@
+$dit-font-stack: Helvetica, Arial, sans-serif;
```
jim68000: I thought we were Arial first? (Now I have to take the Arial or Helvetica test to prove I know the difference... http://www.ironicsans.com/helvarialquiz/

src/sass/environment/settings/_layout.scss
tyom: Hmm.. Trade Element's gutter is `30px` which is what we're using. These settings are not actually being used yet, even though they are declared as dependencies across other includes. I'm going to go and try to remove the stuff that's not being used directly by the application. We can tackle CSS in the following PRs and rely on Trade Elements for now.

src/sass/layout/_global-footer.scss
jim68000: 👍

src/sass/application.scss
teneightfive: It's a tricky one but it is the term used in ITCSS so we can clearly point at that as the source of this file structure.

src/sass/layout/_global-footer.scss
tyom: I'm trying to avoid bloating this PR with stuff not directly concerning the layout. I'll add a TODO to address this when we deal with CSS specifically in the upcoming PRs.

src/sass/environment/settings/_colours.scss
teneightfive: I think the purpose of them is to show what the decided colours for certain patterns are. 

So if within a component someone is styling a link, for example a button that should look like a link, they can clearly see there is a definition for the link colour, rather than having to either look through the rest of the sass to find what colour is being used for links or just pick one themselves.

The idea is basically to breed consistency. There are potentially some that could be culled though, for example `$alpha-colour` if it's only being used in one place?

src/sass/environment/settings/_typography.scss
tyom: My bad. Misread. Although I doubt there's a system out there that has Helvetica Neue but not Arial.

src/sass/environment/settings/_typography.scss
teneightfive: I think these have been picked from a vertical rhythm sizing scale that we used at NHS. These are probably up for change based on this app but should probably come as a separate change and more clearly state why the decided sizes are chosen.

Probably along with switching use of `core-19` to `core-font(19)`.

src/sass/layout/_phase-banner.scss
teneightfive: It was only because we were following the ITCSS model. If we want to simplify to start with I could live with moving them to `/settings`

src/sass/trumps/_utility.scss
```diff
@@ -0,0 +1,19 @@
...
+
+.u-visually-hidden {
+  @include visually-hidden;
+}
+
+.u-hidden {
+  display: none !important;
+}
+
+.js-enabled .u-js-hidden {
```
teneightfive: What about writing this as instead:

```scss
.u-js-hidden {
  .js-enabled & {
    display: none !important;
  }
}
```

Reasons behind it is that you can see all the utility classes by just scanning the left hand side of this file.

src/sass/_shame.scss
feedmypixel: >The whole _shame.scss should go sooner rather than later.
💥

src/sass/application.scss
feedmypixel: yeah I think `object` and `generic` in ITCSS are a little meaningless - but adding this in FED principles works for me

feedmypixel:
> re my comment about search header. Short version, we should not be using the search header anymore, so update the pages that use it and remove it.

FYI the work I'm doing around search removes all this

src/sass/environment/settings/_typography.scss
tyom: I'd suggest we look into this as part of future CSS-related changes. They are here temporary to plug dependency holes.

feedmypixel:
We need to write an integration test setup card.

PR #123, opened 2017-05-26
Move to Contact API V3

marcofucci:
This refactors the contact services, controllers and related resources to use the new Contact API V3.

**Before**
The API returned complete nested objects eg.
```js
{
  "id": "8d4c80b1-2399-4b45-9d41-192648d3233a",
  "first_name": "Oratio",
  "last_name": "Nelson",
  "company": {
    "alias": "...",
    "business_type": "...",
    "sector": "...",
    "registered_address_country": "...",
    "registered_address_1": "75 Stramford Road",
    "registered_address_town": "London",
    "trading_address_country": "...",
    "trading_address_1": "1 Hello st.",
    "trading_address_town": "Dublin",
    "uk_region": "..."
  }
  ...
}
```
or just flat nested ones eg.

```js
{
  "id": "8d4c80b1-2399-4b45-9d41-192648d3233a",
  "first_name": "Oratio",
  "last_name": "Nelson",
  "company": "s1c8d226-12ed-4d75-a3cd-6ec315a08b74"
  ...
}
```

**After**
The API v3 now return a very light version of nested objects composed of just `id` and `name`. 
```js
{
  "id": "479766dc-855a-4131-be8f-e404063a8ddc",
  "first_name": "Oratio",
  "last_name": "Nelson",
  "company": {
      "id": "d15e46f1-89d7-4edb-9f31-e758ae2f9d54",
      "name": "name6"
  },
  ...
}
```
This is so that responses stay slim but it does mean that sometimes the frontend has to make additional calls.

**So this PR:**
- starts using v3
- changes the display contact functions as before they could access contact.company to get the company address, now they need an extra `company` param
- converts a flat object to a nested one before sending it to the backend to save the object (see `convertNestedObjects`)
- updates now use `PATCH` instead of `PUT`

**Note**
To get the company contacts, the code is still requesting the company object via API and looping over company.contacts. Moving to the new `/v3/contact?company_id=...` endpoint is not part of this PR.

### DIFF
--- a/src/controllers/contact-edit.controller.js
+++ b/src/controllers/contact-edit.controller.js
@@ -28,13 +28,16 @@ function editDetails (req, res, next) {
       // or a contact that the user wishes to edit
       // or a new contact for a company
       if (containsFormData(req)) {
+        res.locals.company = res.locals.company
+          ? res.locals.company : yield companyRepository.getDitCompany(token, req.body.company)
         res.locals.formData = req.body
-        res.locals.company = yield companyRepository.getDitCompany(token, req.body.company)
       } else if (res.locals.contact) {
+        res.locals.company = res.locals.company
+          ? res.locals.company : yield companyRepository.getDitCompany(token, res.locals.contact.company.id)
         res.locals.formData = contactFormService.getContactAsFormData(res.locals.contact)
-        res.locals.company = res.locals.contact.company
       } else if (req.query.company) {
-        res.locals.company = yield companyRepository.getDitCompany(token, req.query.company)
+        res.locals.company = res.locals.company
+          ? res.locals.company : yield companyRepository.getDitCompany(token, req.query.company)
         res.locals.formData = { company: res.locals.company.id }
       } else {
         return next('Unable to edit contact')
--- a/src/controllers/contact-edit.controller.js
+++ b/src/controllers/contact-edit.controller.js
@@ -27,20 +27,23 @@ function editDetails (req, res, next) {
       // This can either be data recently posted, to be re-rendered with errors
       // or a contact that the user wishes to edit
       // or a new contact for a company
+      let companyId
+
       if (containsFormData(req)) {
-        res.locals.company = res.locals.company
-          ? res.locals.company : yield companyRepository.getDitCompany(token, req.body.company)
+        companyId = req.body.company
         res.locals.formData = req.body
       } else if (res.locals.contact) {
-        res.locals.company = res.locals.company
-          ? res.locals.company : yield companyRepository.getDitCompany(token, res.locals.contact.company.id)
+        companyId = res.locals.contact.company.id
         res.locals.formData = contactFormService.getContactAsFormData(res.locals.contact)
       } else if (req.query.company) {
-        res.locals.company = res.locals.company
-          ? res.locals.company : yield companyRepository.getDitCompany(token, req.query.company)
-        res.locals.formData = { company: res.locals.company.id }
+        companyId = req.query.company
+        res.locals.formData = { company: req.query.company }
       } else {
         return next('Unable to edit contact')
+      }
+
+      if (!res.locals.company) {
+        res.locals.company = yield companyRepository.getDitCompany(token, companyId)
       }
 
       if (req.params.contactId) {
--- a/src/controllers/contact.controller.js
+++ b/src/controllers/contact.controller.js
@@ -2,6 +2,7 @@ const express = require('express')
 const Q = require('q')
 
 const contactRepository = require('../repos/contact.repo')
+const companyRepository = require('../repos/company.repo')
 const contactFormattingService = require('../services/contact-formatting.service')
 const companyService = require('../services/company.service')
 const { contactDetailsLabels } = require('../labels/contact-labels')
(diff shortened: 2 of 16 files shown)


src/controllers/contact-edit.controller.js
```diff
@@ -28,13 +28,16 @@ function editDetails (req, res, next) {
       // or a contact that the user wishes to edit
       // or a new contact for a company
       if (containsFormData(req)) {
+        res.locals.company = res.locals.company
```
teneightfive: Why do we need to check whether `res.locals.company` already exists?

marcofucci: because it might have been set in another place and we don't want to make an extra call if it's not necessary.

src/services/contact-form.service.js
```diff
@@ -1,5 +1,8 @@
 const Q = require('q')
-const { getPropertyId, nullEmptyFields, convertYesNoToBoolean } = require('../lib/property-helpers')
+const {
+  getPropertyId, nullEmptyFields,
```
teneightfive: If we're using multiline syntax I'd drop each var onto a new line.

src/controllers/contact-edit.controller.js
teneightfive: Not sure why it would have been set on `res.locals` by that point though. I don't think we're really utilising middleware in that way yet. `res.locals` is currently just being used instead of [passing data directly](https://expressjs.com/en/api.html#res.render) to the view.

We're working on moving this to a format of sending data to the view locally rather than adding to `res.locals`.

So for now I wouldn't worry about this check.

marcofucci: Because of this line: https://github.com/uktrade/data-hub-fe-beta2/blob/contact-v3/src/controllers/contact-edit.controller.js#L103

Sometimes it's been set by the `getCommon` but not in all the cases.

src/services/contact-form.service.js
marcofucci: 👍

src/controllers/contact-edit.controller.js
teneightfive: Yep, I see now.

src/lib/property-helpers.js
```diff
@@ -101,4 +123,13 @@ function hasProperty (object, property) {
   object[property] !== null)
 }
 
-module.exports = { getPropertyId, getPropertyName, nullEmptyFields, convertYesNoToBoolean, deleteNulls, hasProperty, hasObjectProperty }
+module.exports = {
```
jim68000: As future @todo it would be nice to alphabetise these when we expand things like this

marcofucci: 👍

src/controllers/contact-edit.controller.js
```diff
@@ -28,13 +28,16 @@ function editDetails (req, res, next) {
...
+          ? res.locals.company : yield companyRepository.getDitCompany(token, req.body.company)
         res.locals.formData = req.body
-        res.locals.company = yield companyRepository.getDitCompany(token, req.body.company)
       } else if (res.locals.contact) {
+        res.locals.company = res.locals.company
+          ? res.locals.company : yield companyRepository.getDitCompany(token, res.locals.contact.company.id)
```
tyom: A bit ternary heavy. Any way to make it more readable? You could, maybe, set the argument you pass to `companyRepository.getDitCompany` as variable and call it in the end?

src/controllers/contact.controller.js
```diff
@@ -31,7 +33,9 @@ function getCommon (req, res, next) {
 function getDetails (req, res, next) {
   try {
     res.locals.tab = 'details'
-    res.locals.contactDetails = contactFormattingService.getDisplayContact(res.locals.contact)
+    res.locals.contactDetails = contactFormattingService.getDisplayContact(
+      res.locals.contact, res.locals.company
```
tyom: Could better to read each argument on its own line since you're breaking lines.

src/lib/property-helpers.js
```diff
@@ -82,6 +82,28 @@ function convertYesNoToBoolean (object) {
 }
 
 /**
+* Convert fk relationships from flat to 1-level deep
+*
+* @param {Object} object
+* @param {Array} props
```
tyom: As these params guaranteed to default to their types if empty? Is it worth initialising them in the function signature below?

src/services/contact-data.service.js
```diff
@@ -67,17 +66,7 @@ function getContactInteractionsAndServiceDeliveries (token, contactId) {
...
-          return Object.assign({}, interaction, {
-            interaction_type: interactionDataService.getInteractionType(interaction.interaction_type),
-            dit_advisor: advisorHash[interaction.dit_advisor],
-            service: serviceOffers.find((option) => option.id === interaction.service),
-            dit_team: metadataRepository.teams.find((option) => option.id === interaction.dit_team)
-          })
-        })
-
-        const combinedIteractions = [...parsedInteractions, ...parsedServiceDeliverys]
+        const combinedIteractions = [...interactions, ...parsedServiceDeliverys]
```
tyom: deliveries

marcofucci:
@tyom all done

marcofucci:
Are you happy @tyom, @teneightfive ?

PR #127, opened 2017-05-30
Added codecov integration

ztolley:
Integrated code coverage reports into PR's. I think we also use this for other datahub projects


teneightfive:
Not sure I'm a fan of automated comments on PR...can make it quite noisy.

teneightfive:
I also think it might be worth looking at [codeclimate](https://codeclimate.com/) as it includes some extra features like quality and duplicated/similar code.

PR #129, opened 2017-05-30
Allow an investment project summary to be edited

teneightfive:
Adds the ability to edit the project summary screen and update it.

### DIFF
--- a/src/controllers/investment-create.controller.js
+++ b/src/controllers/investment-create.controller.js
@@ -1,24 +1,29 @@
 const router = require('express').Router()
 
 const { getInflatedDitCompany } = require('../services/company.service')
-const { getCompanyInvestmentProjects, saveInvestmentProject } = require('../repos/investment.repo')
-const { transformForApi } = require('../services/investment-formatting.service')
+const {
+  getCompanyInvestmentProjects,
+  getInvestmentProjectSummary,
+  createInvestmentProject,
+  updateInvestmentProject
+} = require('../repos/investment.repo')
+const { transformToApi, transformFromApi } = require('../services/investment-formatting.service')
 const metadataRepo = require('../repos/metadata.repo')
 
 function getHandler (req, res, next) {
-  const equityCompanyId = req.query['equity-company']
+  const equityCompanyId = res.locals.equityCompanyId || req.query['equity-company']
   const promises = [
     getInflatedDitCompany(req.session.token, equityCompanyId),
-    // TODO: remove this when the company endpoint returns a list of invesments
     getCompanyInvestmentProjects(req.session.token, equityCompanyId)
   ]
 
   if (!equityCompanyId) {
-    return res.redirect('/investment/start')
+    res.redirect('/investment/start')
   }
 
   Promise.all(promises)
     .then(([equityCompany, equityCompanyInvestments]) => {
+      const form = res.locals.form || {}
       const contacts = equityCompany.contacts.map((contact) => {
         return {
           id: contact.id,
@@ -31,7 +36,6 @@ function getHandler (req, res, next) {
           label: type.name
         }
       })
-      const form = res.locals.form || {}
 
       form.options = {
         contacts,
@@ -51,9 +55,17 @@ function getHandler (req, res, next) {
 }
 
 function postHandler (req, res, next) {
-  const formattedProject = transformForApi(req.body)
+  const formattedBody = transformToApi(req.body)
+  const projectId = res.locals.projectId
+  let saveMethod
 
-  saveInvestmentProject(req.session.token, formattedProject)
+  if (projectId) {
+    saveMethod = updateInvestmentProject(req.session.token, projectId, formattedBody)
+  } else {
+    saveMethod = createInvestmentProject(req.session.token, formattedBody)
+  }
+
+  saveMethod
     .then((result) => {
       res.redirect(`/investment/${result.id}`)
     })
@@ -63,6 +75,7 @@ function postHandler (req, res, next) {
 
         res.locals.form.errors = err.error
         res.locals.form.state = req.body
+
         next()
       } else {
         next(err)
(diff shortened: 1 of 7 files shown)


src/controllers/investment-create.controller.js
```diff
@@ -1,24 +1,29 @@
...
+  getInvestmentProjectSummary,
+  saveInvestmentProject,
+  editInvestmentProject
+} = require('../repos/investment.repo')
+const { transformForApi, transformFromApi } = require('../services/investment-formatting.service')
 const metadataRepo = require('../repos/metadata.repo')
 
 function getHandler (req, res, next) {
-  const equityCompanyId = req.query['equity-company']
+  const equityCompanyId = res.locals.equityCompany || req.query['equity-company']
```
feedmypixel: do we need to support both?

src/controllers/investment-create.controller.js
```diff
@@ -63,14 +75,32 @@ function postHandler (req, res, next) {
...
+}
+
 router.get('/investment/create', getHandler)
 router.post('/investment/create', postHandler, getHandler)
 
-module.exports = { router, getHandler, postHandler }
+router.get('/investment/:id/details/edit', editHandler, getHandler)
+router.post('/investment/:id/details/edit', editHandler, postHandler, getHandler)
+
+module.exports = { router, getHandler, postHandler, editHandler }
```
feedmypixel: newlines for these?

src/services/investment-formatting.service.js
```diff
@@ -78,6 +84,8 @@ function transformForApi (body) {
     project['referral_source_advisor'] = body['is-referral-source']
   }
 
+  project['estimated_land_date'] = `${body['land-date_year']}-${body['land-date_month']}-01`
+
   Object.keys(project).forEach((key) => {
```
feedmypixel: Not part of this PR but it feels like the data needs to be normalised to either an array or an object, not both.

src/services/investment-formatting.service.js
```diff
@@ -87,22 +95,56 @@ function transformForApi (body) {
       project[key] = [{
         id: project[key]
       }]
+    } else if (remove.includes(key)) {
```
feedmypixel: can you remove unwanted keys before they get iterated over?

src/services/investment-formatting.service.js
```diff
@@ -87,22 +95,56 @@ function transformForApi (body) {
...
 
-  delete project['land-date_year']
-  delete project['land-date_month']
-  delete project['is-relationship-manager']
-  delete project['is-referral-source']
+  const date = new Date(formatted['estimated_land_date'])
+  formatted['land-date_year'] = date.getFullYear()
+  formatted['land-date_month'] = date.getMonth() + 1 // month is zero based index
+
+  Object.keys(formatted).forEach((key) => {
```
feedmypixel: This is also being used above - can this be abstracted into a function?

src/services/investment-formatting.service.js
```diff
@@ -87,22 +95,56 @@ function transformForApi (body) {
...
+  return formatted
 }
 
 module.exports = {
   getInvestmentDetailsDisplay,
   getOpenInvestmentProjects,
   getClosedInvestmentProjects,
-  transformForApi
+  transformForApi,
+  transformFromApi
```
feedmypixel: the wording of these methods are very similar and I'm not sure what they do or how they are different. Could a rename make that clearer?

feedmypixel: If I have understood right maybe these are clearer and less likely to be mixed up?
- ` transformToApi`
- `transformFromApi`

src/views/investment/create.njk
```diff
@@ -2,27 +2,32 @@
...
+    <table class="table--key-value">
+      <tbody>
+        <tr>
+          <td>{{ equityCompany.name }}</td>
+        </tr>
+        <tr>
+          <td>{{ equityCompany.registered_address_country.name }}</td>
+        </tr>
+        <tr>
+          <td>{{ equityCompanyInvestments.count|default('No', true) }} investment project{{ 's' if equityCompanyInvestments.count != 1 }} in the UK</td>
```
feedmypixel: FYI there is [pluralise filter](https://github.com/uktrade/data-hub-fe-beta2/pull/139/files#diff-a34f9a72bb41fd9597d5ec8aa14d4e1fR18) coming

teneightfive: _applaudes_

src/controllers/investment-create.controller.js
teneightfive: Reason being is one is coming from the `editHandler` middleware (which thinking about it, I may rename to `editMiddleware` to be more clear) when editing a project summary, and the other is being set as a querystring param. 

When editing I don't really want to redirect to the get with the querystring as you'd lose the values in `res.locals`.

I could refactor so that there is a middleware on all routes that looks for the querystring params and also sets it on to `res.locals` so that this function only expects that value to come from one place?

src/services/investment-formatting.service.js
teneightfive: I don't really like what these functions are doing to be honest. It's compensating for the way the API sends that data back. We have to expand a string into an object in some cases where I think it might be nicer for the API to accept a string or an object, and if a string, treat it as the ID value.

src/services/investment-formatting.service.js
teneightfive: Yes that's probably more efficient.

src/repos/investment.repo.js
```diff
@@ -17,11 +17,19 @@ function getInvestmentRequirements (token, investmentId) {
   return authorisedRequest(token, `${config.apiRoot}/v3/investment/${investmentId}/requirements`)
 }
 
-function saveInvestmentProject (token, project) {
+function saveInvestmentProject (token, body) {
```
tyom: I wonder if create/update would be more appropriate verbs.

src/controllers/investment-create.controller.js
feedmypixel: >When editing I don't really want to redirect to the get with the querystring as you'd lose the values in res.locals.

sounds sensible

>I could refactor so that there is a middleware on all routes that looks for the querystring params and also sets it on to res.locals so that this function only expects that value to come from one place?

Feels like you would expect a query to be on req.query rather than res.locals. Which would make it a little magic IMO. 

It sounds like for the moment we have two ways in.

Maybe this is a bigger thing in that we have a `getHandler` which is doing the get work for two things. Render a company page and edit a company page. Potentially we have a handler for the display, and then if it is an edit it falls through to an edit handler?

*caveat I have not had my head in the code so take whichever route you feel is sensible*

src/views/investment/create.njk
feedmypixel: 🙇

src/controllers/investment-create.controller.js
teneightfive: The reason for using the same handler is there is next to no difference apart from where the equity company ID is coming from, hence the thought to use the same handler but have the edit route add an additional piece of middleware.

Let me see if I can refactor it a bit.

feedmypixel: I quite like the middleware pattern on routes. If it doesn't do something in the first handler just drop it through to the next, same principle as the error handling.

This approach also works really nicely when doing errors with `post -> redirect get`

teneightfive: Yep, am doing that with the post/get actions. The post route includes the postaction and then drops through to the get action. Does just mean you have to set things on `res.locals` though.

teneightfive: Had a chat with @tyom and for now I think we're going to leave it rather than adding more middleware. It's something we'll probably revisit when we look to combine a lot of the add/edit flows that we currently have.

PR #130, opened 2017-05-30
Add configuration to carry out more checks in code climate

ztolley:


jim68000:
I like that your codeclimate PR fails in codeclimate.

PR #131, opened 2017-05-31
Fix issues for IE8 and below

tyom:
- Stop serving JS to older IEs
- Fix HTML5 tags for older IEs (shiv)

This change removes Webpack's CommonsChunkPlugin as it was interfering with IE shiv that we pre-bundle with Webpack. At this point the commons JS was very small and wasn't really very useful (we don't serve many vendor files). The way the bundles are build (one per file) also makes it harder to configure commons chunks. It's designed to work with bigger bundles which import other files from within rather than JS file per bundle as it's currently configured.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -16,7 +16,8 @@ module.exports = {
     search: './src/javascripts/search',
     'service-delivery': './src/javascripts/service-delivery',
     'archive-form': './src/javascripts/archive-form',
-    'add-another-field': './src/javascripts/add-another-field'
+    'add-another-field': './src/javascripts/add-another-field',
+    ie: ['html5shiv']
   },
   output: {
     path: 'build/javascripts',
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -56,10 +56,8 @@ module.exports = {
       sourceMap: false,
       dead_code: true
     }),
-    new webpack.optimize.DedupePlugin(),
-    new webpack.optimize.CommonsChunkPlugin('common.js')
+    new webpack.optimize.DedupePlugin()
   ] : [
-    new webpack.optimize.DedupePlugin(),
-    new webpack.optimize.CommonsChunkPlugin('common.js')
+    new webpack.optimize.DedupePlugin()
   ]
 }
--- a/package.json
+++ b/package.json
@@ -28,6 +28,7 @@
     "express": "^4.14.0",
     "express-session": "^1.14.1",
     "font-awesome": "^4.6.3",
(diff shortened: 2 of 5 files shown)


webpack.config.js
```diff
@@ -55,10 +56,8 @@ module.exports = {
       sourceMap: false,
       dead_code: true
     }),
-    new webpack.optimize.DedupePlugin(),
-    new webpack.optimize.CommonsChunkPlugin('common.js')
+    new webpack.optimize.DedupePlugin()
```
ztolley: Not a bad thing, just curios why dropping the common plugin, is it just because the benefits are so little?

webpack.config.js
```diff
@@ -55,10 +56,8 @@ module.exports = {
...
   ] : [
-    new webpack.optimize.DedupePlugin(),
-    new webpack.optimize.CommonsChunkPlugin('common.js')
+    new webpack.optimize.DedupePlugin()
```
ztolley: Not in the scope of this card but should we be switching to a more recent version of webpack and use tree shaking?

webpack.config.js
```diff
@@ -16,7 +16,8 @@ module.exports = {
     search: './src/javascripts/search',
     'service-delivery': './src/javascripts/service-delivery',
     'archive-form': './src/javascripts/archive-form',
-    'add-another-field': './src/javascripts/add-another-field'
+    'add-another-field': './src/javascripts/add-another-field',
```
teneightfive: I do think we should actually enforce trailing comma rather than enforcing no trailing comma. It would avoid longer diffs like this.

webpack.config.js
tyom: I tried to explain it in PR description. Basically, yeah, we're not getting any benefits at this stage. And this change actually makes it harder to implement as we want to exclude IE bundle from commons.

webpack.config.js
tyom: Yeah, we should switch to Webpack 2 at some point.

webpack.config.js
ztolley: originally it was there because we put a whole ton of shivs in for ie but then I think I eventually pulled that out so yay to no common

webpack.config.js
ztolley: As much as I love Standard js... I'm inclined to agree it makes merges easier to have trailing

teneightfive: Not to mention when coding and using IDE shortcuts to duplicate a line. With this rule set you have to manually add/remove that comma when doing this.

PR #132, opened 2017-05-31
use double quotes in Sass

tyom:


src/sass/_deprecated/_flash.scss
```diff
@@ -2,7 +2,7 @@
   padding: 15px 20px;
 
   &.flash--info {
-    background-color: #E5EEF6;
-    border-left: 8px solid #005EA5;
+    background-color: #e5eef6;
```
ztolley: Would it be better to make this a variable?

tyom: It would be. But it's out of scope of this PR. I tried to pick low hanging fruit here. There is future work to fix style linting warnings (and make them into errors). But it's a much larger piece of work.

Issue #98, opened 2017-05-19
Move development dependencies to devDependencies in package.json


teneightfive:
## Problem

We currently list the majority of dependencies in `dependencies` in the package manifest. However, some of these packages, like mocha, are only required during development.

## Solution

Move development only dependencies to `devDependencies` key in package manifest.

PR #134, opened 2017-05-31
Move postcss-cli from devDeps to deps

tyom:
The `postcss` command needs to exist when deploying.


teneightfive:
We should also clean up what's currently in deps. Feels like quite a few should only exist in devDeps.

tyom:
> We should also clean up what's currently in deps. Feels like quite a few should only exist in devDeps.

Yes, we should. Should we do it in this PR?

teneightfive:
I reckon so as we're already playing with moving deps between the two. Then just change the description to be less specific.

tyom:
> I reckon so as we're already playing with moving deps between the two. Then just change the description to be less specific.

Due to the way we deploy certain dev packages are required in production. Perhaps we should do it separately as it might require more work than it seems on the surface. There's #98 for this already. 

This is supposed to be a fix for broken build.

teneightfive:
>Due to the way we deploy certain dev packages are required in production. Perhaps we should do it separately as it might require more work than it seems on the surface. There's #98 for this already.

Agreed.

PR #135, opened 2017-05-31
Remove dead code (Company Investment)

tyom:
I can't find evidence that this code is being used so I'm removing the dead weight.


teneightfive:
This may well be needed as a reference when adding the _investment pipeline_.

tyom:
> This may well be needed as a reference when adding the investment pipeline.

Could we find it in history instead?

PR #137, opened 2017-05-31
Convert interaction and service delivery pages to use new layout

tyom:


src/views/interaction/interaction-edit.njk
```diff
@@ -13,6 +12,10 @@
 
   <h1 class="heading-xlarge">Add new interaction</h1>
   {{ trade.errorPanel( errors, labels=labels ) }}
+{% endblock %}
+
+{% block body_main_content %}
+  {{ formData | log | safe }}
```
ztolley: Stray logging statement?

tyom: Maybe. But it was already there (line 5).

tyom:
> As a follow up I was thinking to convert the key value tables in the details pages to use the component version that looks ready

We should do that as part of our Trade Elements move/conversion work.

PR #139, opened 2017-05-31
Search results Header and v3 api work

feedmypixel:
Fundamentally this works is for `Jira > DH-242` **Create faceted search results header**. But due to [yak shaving](https://en.wiktionary.org/wiki/yak_shaving) it has encompassed a lot more.

- Create faceted search results header
- Update Search to work with `v3/search` endpoint and response
- Refactor Search, pagination components and bring them over from Trade Elements
- Create Search Nav, Results Summary components
- Refactor associated nunjucks filter and bring over from Trade Elements
- Refactor, modularise and BEM associated styles
- Add [Joi validation](https://github.com/hapijs/joi/tree/v10.5.2) with [Celebrate Joi Validation middleware](https://github.com/continuationlabs/celebrate) for query, params, headers, body validation (potential for validation schemas becoming separate files if we move folder structure to being about features not concerns)

### Of Note
- This has been ran against a local checkout of [data-hub-leeloo/pull/212](https://github.com/uktrade/data-hub-leeloo/pull/212) and works as expected
- I have removed the display of **Incorportation Date** as this is now not available directly from `/v3/search`

searchresults


src/sass/components/_lists.scss
```diff
@@ -0,0 +1,9 @@
+.list {
```
teneightfive: Should this be a modifier? Feels like the default should be with `list-style: disc` and to remove the list style we add a modifier?

Or the component could be more specific like `list-compact`?

src/sass/components/_results.scss
```diff
@@ -0,0 +1,26 @@
+.results-list {
```
teneightfive: I've realised that this feels similar to the [search result](https://github.com/uktrade/data-hub-fe-beta2/blob/develop/src/sass/_deprecated/_search-result.scss) component that has been moved to `_deprecated` even though it's a more recent edition using BEM naming convention. 

Maybe we should look at combing the two and using a modifier. The existing one adds a border around the element and fits a different use case where we want the whole result to be the clickable action.

src/sass/elements/_form.scss
```diff
@@ -0,0 +1,6 @@
+input[type="search"]::-webkit-search-decoration,
```
teneightfive: Is this something that could live in the reset file or base?

src/views/_components/pagination.njk
```diff
@@ -0,0 +1,25 @@
+{##
+ # @param {object} pagination - An array of objects with properties {label, link, currentPage}
```
teneightfive: So a pattern @tyom and I started to use was to use the property `children` when a component listed over a set of objects. 

It meant components were slightly more flexible with each other and could often mean you could nest components within components more easily. It also meant that if you see a `children` property the component will be looping over some other type of object. Might be worth thinking about whether that's still a good pattern to use and if we want to stick to it.

src/sass/components/_pagination.scss
```diff
@@ -0,0 +1,14 @@
+@import "../settings/colours";
+
+.pagination-nav {
```
teneightfive: Is this not a modifier on a list? It seems to add a border to the top and give them more space so feels like it could be applied to the original component.

The name also feels too heavily tied to the content it contains. Could this style not be used for something other than pagination?

src/views/_components/results-summary.njk
```diff
@@ -0,0 +1,20 @@
+{##
+ # @param {int}    total      - Total number of results
+ # @param {string} searchTerm - Search term used to obtain results
+ #
+ # @return {string} HTML for results summary
+ #
+ # @example
+ #   {% component 'results-summary', {
+ #     total: results.count,
```
teneightfive: Might be more useful to pass a string or int into the example so it's more obvious.

src/sass/components/_lists.scss
tyom: Doesn't feel like this is a component. There's a [list object](https://github.com/uktrade/data-hub-fe-beta2/blob/develop/src/sass/objects/_list.scss) that you may find fits the bill.

src/sass/elements/_headings.scss
```diff
@@ -0,0 +1,13 @@
...
+h3,
+h4,
+h5,
+h6 {
+  display: inline-block;
+  margin-bottom: $gutter/2;
+}
+
+h3 {
+  @include core-19();
```
tyom: Shouldn't use empty `()` for mixins without args. 

Also, can we start moving away from `code-n` mixins. They are not very good.

config/nunjucks/filters.js
```diff
@@ -0,0 +1,26 @@
+const nunjucks = require('nunjucks')
+const filters = {}
+
+filters.stringify = JSON.stringify
```
tyom: Could just initialise them with the object. Wouldn't it be nicer to read?

```js
const filters = {
  stringify: JSON.stringify,

  removeNull(array) {
    return array.filter(item => item)
  }

  // …
}
```

src/sass/components/_results.scss
feedmypixel: yeah I also noticed `.search-result` when doing this. Feels like we need to point out the differences to design and see if they can be aligned. If not then what you have suggested sounds like a plan.

src/sass/components/_pagination.scss
feedmypixel: Yeah its a little tricky, I think this would fall under a design refactor and alignment for `list` and `list--inline`. The `text-align: center;` and random padding made me feel like it was specific to `.pagination-nav`. Move it to `_deprecated` to indicate it needs a look at?

src/sass/components/_lists.scss
feedmypixel: I have presumed our default list style is without `list-style: disc` same again maybe needs aligning with design?

config/nunjucks/filters.js
feedmypixel: Once Trade elements filters have been moved over I'd quite like to make this a `Map`

src/views/_components/pagination.njk
feedmypixel: Lets have a catch up when we are face to face

config/nunjucks/filters.js
```diff
@@ -0,0 +1,27 @@
...
+    )
+  },
+
+  removeNull: (array) => {
+    return array.filter(item => item)
+  },
+
+  pluralise: (string, count) => {
+    if (count !== 1) {
+      string += '\'s'
```
tyom: Is apostrophe needed in pluralised word?

config/nunjucks/index.js
```diff
@@ -36,24 +36,29 @@ function ComponentExtension (env) {
   }
 }
 
+function customFilters (env) {
+  const tradeElementsFilters = require(`@uktrade/trade_elements/dist/nunjucks/filters`)
+  const dataHubFilters = require('./filters')
+
+  const filters = Object.assign(tradeElementsFilters, dataHubFilters)
+
+  Object.keys(filters).forEach((filterName) => {
```
tyom: Lodash could make it easier to read:
```js
const {forOwn} = require('lodash')

forOwn(filters, (filterFunc, filterName) => {
  env.addFilter(filterName, filterFunc)
})
```

config/nunjucks/index.js
```diff
@@ -36,24 +36,29 @@ function ComponentExtension (env) {
   }
 }
 
+function customFilters (env) {
```
tyom: Should function be called something like `addCustomFilters`?

src/views/_components/search-bar.njk
```diff
@@ -0,0 +1,42 @@
+{##
+ # @param {string} action - Search forms action attribute value
+ # @param {string} method - Search forms method attribute value
+ # @param {string} label  - Search forms label text
+ # @param {string} searchTerm  - Search term used for search forms input value
+ # @param {string} placeHolder  - Text used for Search inputs placeholder, defaults to label
```
tyom: [JSDoc syntax for optional params](http://usejsdoc.org/tags-param.html#optional-parameters-and-default-values) is `[placeHolder=label]`.

Isn't "placeholder" one word?

src/views/_components/search-bar.njk
```diff
@@ -0,0 +1,42 @@
...
+ #    method: 'get',
+ #    label: 'Search...',
+ #    searchTerm: searchTerm
+ #   } %}
+ #}
+
+{% set placeHolder = placeHolder | default(label) %}
+{% set submitText = submitText | default('Search') %}
+
+<form class="search-bar"
```
tyom: Probably OK on one line, it's not that long. But if multiline I think it's better to drop each attribute on its own line followed by closing `>`.

```njk
<form
  class="search-bar"
  action="{{ action }}"
  method="{{ method }}"
  role="search"
>
```

Makes it easier to scan and more consistent with your `input` below.

src/sass/components/_results.scss
teneightfive: I showed Tom N this variation and he was happy with it. But they also have different uses. But feel like they may well be similar markup so could combine to 1 component?

src/sass/elements/_headings.scss
teneightfive: @tyom feels like this might a separate PR to strip all those out and replace with `core-font(x)` and error any further uses of it.

config/nunjucks/filters.js
```diff
@@ -0,0 +1,27 @@
+const nunjucks = require('nunjucks')
+const filters = {
+
+  stringify: JSON.stringify,
+
+  highlight: (string, searchTerm) => {
+    const regEx = new RegExp(`(${searchTerm})`, 'gi')
+
+    return new nunjucks.runtime.SafeString(
+      string.replace(regEx, '<strong>$1</strong>')
```
teneightfive: Think we discussed making this class based in case we decide on a different style.

config/nunjucks/index.js
teneightfive: I wonder whether it even needs to be a separate function - it's only 6 lines of code. And we've done a lot of abstraction by moving nunjucks config out of `server.js` and filters into their own module. This feels like 1 level too far.

src/controllers/search.controller.js
```diff
@@ -1,68 +1,91 @@
 const express = require('express')
 const companyRepository = require('../repos/company.repo')
-const { buildCompanyUrl } = require('../services/company.service')
+const {buildCompanyUrl} = require('../services/company.service')
```
teneightfive: What's our pattern for this. I've been using a space between curly braces up until now. It's not something that we've defined in the linter but maybe should for consistency?

src/controllers/search.controller.js
```diff
@@ -1,68 +1,91 @@
...
+  const searchTerm = req.query.term
 
   searchService.search({
     token: req.session.token,
-    term: req.query.term,
-    page: req.query.page,
-    filters
+    searchTerm,
+    searchType,
+    page: req.query.page
```
teneightfive: Is it worth being consistent and defining this after line 48 too?

teneightfive: That way all the request params being used are declared in one place.

src/controllers/search.controller.js
```diff
@@ -1,68 +1,91 @@
...
+      next(error)
+    }
+  })
 }
 
-router.get('/search', get)
-router.get('/viewcompanyresult/:source/:id', viewCompanyResult)
+router.get('/search/:searchType?', searchActionValidationSchema, indexAction, searchAction)
+
+router.get('/viewcompanyresult/:id', viewCompanyResult) // TODO is this in the right controller
```
teneightfive: I'm still not convinced by what we're doing here either. 

I don't think we need to be conveying what type a company is in the url. We should be using more generic urls like `/company/:companyId`.

That's probably a separate change though.

src/sass/components/_search-bar.scss
```diff
@@ -0,0 +1,36 @@
...
+  position: relative;
+}
+
+.search-bar__input {
+  background-color: $white;
+  border: 1px solid $grey-5;
+  color: $black;
+  box-sizing: border-box;
+  width: 100%;
+  @include core-font(19);
```
teneightfive: I find it easier to read when mixins are written before declarations. 

We should consider enabling the [sass-lint rule](https://github.com/sasstools/sass-lint/blob/master/docs/rules/mixins-before-declarations.md) for it.

src/sass/components/_results.scss
teneightfive: Also missing dependency that includes `$gutter`. Or is that coming from trade elements?

src/sass/components/_search-nav.scss
```diff
@@ -0,0 +1,23 @@
+@import "../settings/colours";
+
+.search-nav {
```
teneightfive: Wondering whether this style is specific to _search_ or whether it's just a tab component?

src/services/search.service.js
```diff
@@ -1,26 +1,24 @@
...
-    offset: (page * limit) - limit,
-    doc_type: filters.length && filters
+    offset: (page * limit) - limit
   }
+
   const options = {
-    url: `${config.apiRoot}/search/`,
-    body: requestBody,
-    method: 'POST'
+    url: `${config.apiRoot}/v3/search${buildQueryString(params)}`,
```
teneightfive: Feels like we're lacking some consistency with the use of _services_ and _repos_ at the moment. 

Should we add a ticket to discuss how we want to structure calls to APIs and formatting of data?

src/views/_components/search-bar.njk
```diff
@@ -0,0 +1,42 @@
...
+ # @param {string} submitText  - Text used for Search submit, defaults to "Search"
+ #
+ # @return {string} HTML for search bar
+ #
+ # @example
+ #   {% component 'search-bar', {
+ #    action: '/search/company',
+ #    method: 'get',
+ #    label: 'Search...',
+ #    searchTerm: searchTerm
```
teneightfive: Would also replace this with a string to avoid use of variables in components.

src/views/_components/search-bar.njk
```diff
@@ -0,0 +1,42 @@
...
+ #   } %}
+ #}
+
+{% set placeHolder = placeHolder | default(label) %}
+{% set submitText = submitText | default('Search') %}
+
+<form class="search-bar"
+      action="{{ action }}"
+      method="{{ method }}"
```
teneightfive: Could this have a default value and method be optional. Feels like in most cases it would be `GET`.

src/views/_components/search-bar.njk
```diff
@@ -0,0 +1,42 @@
...
+    <label class="visually-hidden" for="search">
+      {{ label }}
+    </label>
+    <div class="search-bar__input-wrapper">
+      <input class="search-bar__input"
+             id="search"
+             type="search"
+             name="term"
+             value="{{ searchTerm }}"
+             placeholder="{{ placeHolder }}"
```
teneightfive: Does it always have to include a placeholder? Might be better to make this completely optional and when calling it can be set to the same value of label if that's the desire.

Might also be worth considering if the label should always be hidden of if that could be a property on the component.

src/views/index.njk
```diff
@@ -3,7 +3,11 @@
 
 {% block body_main_content %}
 
-  {{ trade.searchbar('term', label="Search for company name or contact", value=query.term) }}
+  {% component 'search-bar', {
+    action: '/search/company',
+    method: 'get',
+    label: 'Search...'
```
teneightfive: Should use the previous label rather than an ellipsis. Helps to make it clearer what can be searched for whilst it's still limited.

src/views/search/_layout.njk
```diff
@@ -0,0 +1,52 @@
+{% extends "layouts/_base-two-column.njk" %}
+
+  {% block main_heading %}
+  <div class="grid-row">
```
teneightfive: Needs an extra indentation.

src/views/search/_layout.njk
```diff
@@ -0,0 +1,52 @@
...
+          } %}
+
+          {% block results %}
+          {% endblock %}
+        </div>
+      </div>
+    {% endif %}
+
+    <div class="grid-row">
+    <div class="column-full">
```
teneightfive: Needs an extra indentation

src/views/search/_layout.njk
```diff
@@ -0,0 +1,52 @@
...
+
+  {% block main_grid_left_column %}
+    &nbsp; {# // TODO filters #}
+  {% endblock %}
+
+  {% block main_grid_right_column %}
+    {% if results %}
+      <div class="grid-row">
+        <div class="column-full">
+          {% block resultsBar %}
```
teneightfive: We should decide on case convention for nunjucks blocks. We're using snake case in the base layouts but camel case here.

Nunjucks tends to use camel case for variables and block names.

src/views/search/results-company.njk
```diff
@@ -0,0 +1,51 @@
...
+  <div class="results-bar">
+    <a class="button button-secondary" href="/company/add-step-1/">
+      Add new company
+    </a>
+  </div>
+{% endblock %}
+
+
+{% block results %}
+<ol class="results-list">
```
teneightfive: Missing indentation

src/views/search/results-company.njk
```diff
@@ -0,0 +1,51 @@
...
+  {% for company in results.companies %}
+    <li class="results-list__item">
+      {% if company.companies_house_data %}
```
teneightfive: This doesn't look easy to read by nesting the opening html tag within an if/else block. 

Might be easier to set that value of the href to a variables using [set](https://mozilla.github.io/nunjucks/templating.html#set) within the if/else block and indenting the markup at the same level.

Should also probably nest the anchor within the h3 if we're not including anything else inside the anchor tag.

src/views/search/results-contact.njk
```diff
@@ -0,0 +1,36 @@
+{% extends "./_layout.njk" %}
+
+{% block results %}
+<ol class="results-list">
+  {% for contact in results.contacts %}
+  <li class="results-list__item">
+    <a href="/contact/{{ contact.id }}/details?term={{ searchTerm }}">
+      <h3>{{ (contact.first_name + ' ' + contact.last_name) | highlight(searchTerm) }}</h3>
```
teneightfive: Should we be applying any style to this element? What if we want the result to be displayed the same but using a `h2` or `h4`?

src/views/search/results-company.njk
```diff
@@ -0,0 +1,51 @@
...
+        <a href="/company/view/ch/{{ company.companies_house_data.company_number }}">
+      {% else %}
+        <a href="/viewcompanyresult/{{ company.id }}">
+      {% endif %}
+        <h3>{{ company.name | highlight(searchTerm) }}</h3>
+      </a>
+
+      {% if company.companies_house_data %}
+        <div>
```
teneightfive: Should these be paragraphs instead? Or maybe even list items within a `dl` with the description of the metadata as the `dt` but hidden? Could you reuse [metadata-list](https://github.com/uktrade/data-hub-fe-beta2/blob/develop/src/sass/_deprecated/_metadata-list.scss)? (Another component that probably shouldn't be in deprecated)

src/views/search/results-contact.njk
```diff
@@ -0,0 +1,36 @@
...
+    <div>
+      {% for addressDetail in [
+        contact.registered_address_1,
+        contact.registered_address_2,
+        contact.registered_address_3,
+        contact.registered_address_4,
+        contact.registered_address_county,
+        contact.registered_address_town,
+        contact.registered_address_postcode
+      ] | removeNull %}
```
teneightfive: Could consider using [whitespace control](https://mozilla.github.io/nunjucks/templating.html#whitespace-control) here

config/nunjucks/index.js
feedmypixel: Abstraction into a function to provide high level instructions is one thing I quite like. It doesn't just have to be for DRY purposes, communication and abstraction is the aim here. 

So in this instance you can scan the `module.exports` function and see it has a call to a function named `addCustomFilters` and that tells you all you need. If you wish to delve into the function you can. Rather than having to scan the `module.exports` function and reverse engineer a block of code.

config/nunjucks/filters.js
feedmypixel: Yeah I had that thought whilst refactoring it but decided to stay with `<strong>`. We can do that if the designs need it in the future or do you have a potential generic class name in mind? feels like maybe a `.text-` helper?

src/controllers/search.controller.js
feedmypixel: let me know and ill set it on my IDE 

> It's not something that we've defined in the linter but maybe should for consistency?

[object-curly-spacing](http://eslint.org/docs/rules/object-curly-spacing) would do it

feedmypixel: IDE switched to have spaces within Object literal braces

src/controllers/search.controller.js
feedmypixel: `searchTerm` and `searchType` have been cached as local consts because they are used multiple times. `token` and `page` are used on once by being sent in to the `search` method.

src/controllers/search.controller.js
feedmypixel: see reply on above comment

src/sass/components/_results.scss
feedmypixel: yeah `$gutter` is from govuk_elements via trade elements.

src/services/search.service.js
feedmypixel: > Should we add a ticket to discuss how we want to structure calls to APIs and formatting of data?

+ 1

src/views/_components/search-bar.njk
feedmypixel: > Does it always have to include a placeholder? Might be better to make this completely optional and when calling it can be set to the same value of label if that's the desire.

placeholder defaults to label

>Might also be worth considering if the label should always be hidden of if that could be a property on the component.

Sounds good. For this PR I'd call YAGNI

src/views/index.njk
feedmypixel: From the designs Tom has put in `Search...` I'll clarify with him. I think it is due to other things coming in, `projects`, `events`, `advisors`

feedmypixel: As discussed with @teneightfive will update to `Search for company name or contact` for now and revist when new search things come in, to increase or reduce message.

src/sass/components/_results.scss
feedmypixel: > But feel like they may well be similar markup so could combine to 1 component?

Worth raising a card to do this once this work is in. Feels like once we know how all look we can combine where required.

src/sass/components/_pagination.scss
teneightfive: Yeah moving to deprecated could work for now to get this work in. A good idea to put things in there that we're not happy with just yet.

src/views/_components/search-bar.njk
teneightfive: Fair point. One to discuss if we find a need to a search component that needs a visible label.

In fact, that's what I've used for the investment journey search. But that's not using components yet so will possibly revisit that and can address it then.

src/controllers/search.controller.js
```diff
@@ -4,65 +4,90 @@ const { buildCompanyUrl } = require('../services/company.service')
...
+    )
   }
+})
+
+function searchAction (req, res, next) {
+  const searchType = req.params.searchType
+  const searchTerm = req.query.term
 
-  const filters = buildSearchFilters(req.query)
+  console.log(searchType, '==================================')
```
teneightfive: 🚨

src/controllers/search.controller.js
teneightfive: I suppose I was thinking more for readability but leave it as is.

src/controllers/search.controller.js
teneightfive: Which reply?

src/sass/components/_results.scss
```diff
@@ -0,0 +1,26 @@
...
+.results-list__item {
+  margin-bottom: $gutter;
+}
+
+.results-bar {
+  text-align: right;
+}
+
+.results-summary {
+  @include core-19;
```
teneightfive: Do we want to use the new mixins for fonts in this file too?

src/views/_components/pagination.njk
```diff
@@ -0,0 +1,25 @@
...
+ #   {% component 'pagination', {
+ #     pagination: pagination
+ #   } %}
+ #}
+{% if pagination | length %}
+  <nav class="pagination-nav">
+    <ul class="list-inline">
+      {% for page in pagination %}
+        <li class="pagination-nav__item">
+        {% if page.currentPage %}
```
teneightfive: Would indent this a further level so it's easier to read/scan

src/views/_components/search-bar.njk
```diff
@@ -0,0 +1,45 @@
+{##
+ # @param {string} action - Search forms action attribute value
+ # @param {string} method - Search forms method attribute value
```
teneightfive: Might be worth using the default value syntax to highlight one is being set (`method=get`). Would that also now make it optional?

src/views/_components/search-bar.njk
```diff
@@ -0,0 +1,45 @@
...
+ # @example
+ #   {% component 'search-bar', {
+ #    action: '/search/company',
+ #    method: 'get',
+ #    label: 'Search...',
+ #    searchTerm: 'example search term'
+ #   } %}
+ #}
+
+{% set method = method | default('GET') %}
```
teneightfive: Lowercase the default value

src/views/_components/search-bar.njk
```diff
@@ -0,0 +1,45 @@
+{##
+ # @param {string} action - Search forms action attribute value
+ # @param {string} method - Search forms method attribute value
+ # @param {string} label  - Search forms label text
+ # @param {string} searchTerm  - Search term used for search forms input value
+ # @param {string} [placeholder=label]  - Text used for Search inputs placeholder, defaults to label
+ # @param {string} submitText  - Text used for Search submit, defaults to "Search"
```
teneightfive: Same as above re the default value

src/views/_components/search-bar.njk
```diff
@@ -0,0 +1,45 @@
...
+      {{ label }}
+    </label>
+    <div class="search-bar__input-wrapper">
+      <input class="search-bar__input"
+             id="search"
+             type="search"
+             name="term"
+             value="{{ searchTerm }}"
+             placeholder="{{ placeholder }}"
+      />
```
teneightfive: We don't need the self closing slash

src/views/index.njk
```diff
@@ -3,7 +3,11 @@
 
 {% block body_main_content %}
 
-  {{ trade.searchbar('term', label="Search for company name or contact", value=query.term) }}
+  {% component 'search-bar', {
+    action: '/search/company',
+    method: 'get',
```
teneightfive: Can possibly remove this now we have the default value? Although it is nice to be clear when calling the component.

src/views/search/_layout.njk
```diff
@@ -0,0 +1,55 @@
+{% extends "layouts/_base-two-column.njk" %}
+
+
+{{ searchType }}
```
teneightfive: Is this a debugging slip?

src/views/search/_layout.njk
```diff
@@ -0,0 +1,55 @@
...
+
+  {% block main_heading %}
+    <div class="grid-row">
+      <div class="column-full">
+        {% component 'search-bar', {
+          action: '/search/' + searchType | default('company'),
+          method: 'get',
```
teneightfive: Could also remove this if we want to fall back to the default.

config/nunjucks/index.js
```diff
@@ -36,24 +36,30 @@ function ComponentExtension (env) {
...
 }
 
+function addCustomFilters (env) {
+  const tradeElementsFilters = require(`@uktrade/trade_elements/dist/nunjucks/filters`)
+  const dataHubFilters = require('./filters')
+
+  const filters = Object.assign(tradeElementsFilters, dataHubFilters)
+
+  Object.keys(filters)
+    .forEach((filter) => {
```
teneightfive: Does this need to be on a new line?

config/nunjucks/index.js
```diff
@@ -36,24 +36,30 @@ function ComponentExtension (env) {
   }
 }
 
+function addCustomFilters (env) {
```
teneightfive: Still not convinced by abstracting to a function in this instance. Feels like if we start by doing this we should move all the other lines within the module out to their own function which would leave us with `configureNunjucks()`, `addExtensions()` and `addGlobals()` each only doing a few lines of code. I feel its clear enough what they're doing when left within the module and almost easier to scan the module that way rather than having to move up and down the file to see what each function is doing. We've also already abstracted the heavy lifting of the filters and extensions to their own functions/modules.

src/controllers/search.controller.js
feedmypixel: :man_facepalming:

src/sass/components/_search-nav.scss
feedmypixel: Another to refactor I feel once we have examples of all "tabby" potentials. This has just been brought over from Trade Elements and refactored/BEM'd

src/views/search/results-contact.njk
feedmypixel: Potentially, but one for another PR I feel

src/controllers/search.controller.js
feedmypixel: Oh strange, I'm sure this was asked twice! So the param `searchType` is `company`, `contact` it is the `entity` used in `/v3/search`. It felt like this was nice to have it as a param for search. Your in `/search`, or your in `/search/company` or your in `/search/contact`. The current section is also highlighted on the tab in the search nav and the term a query in the url

Maybe it is a bigger url structure question

src/views/index.njk
feedmypixel: feels nicer to have it in

src/views/search/_layout.njk
feedmypixel: feels nicer to have it in

src/controllers/search.controller.js
teneightfive: Think we have have misunderstood each other a bit. I was referring to the `/viewcompanyresult ` route and how it takes an ID, does an API call to get a company decides on a url and then redirects to it.

Also that a route concerned with getting a company is in the search controller. This feels like it's the concern of a company controller. At the moment we're doing multiple calls to get a company from the API but in different feature locations.

A more generic url also feels like a better approach. So we always have `/company/:id` regardless of what type. Rather than `/viewcompanyresult/:id` which redirects to `/company/view/ukother/:id` or `company/view/foreign/:id`.

It's definitely something for a different PR though and possibly easier to explain in person.

config/nunjucks/index.js
tyom: I'm not a fun of the side effects in this function. Could we make it pure if we need to make it a function?

src/sass/components/_search-bar.scss
teneightfive: Do you want to move the mixin to the top of the function though so we start using that pattern now?

src/sass/settings/_colours.scss
```diff
@@ -7,6 +7,7 @@ $grey-1: #6f777b;
 $grey-2: #bfc1c3;
 $grey-3: #dee0e2;
 $grey-4: #f8f8f8;
+$grey-5: #9b9b9b;
```
teneightfive: Should we be introducing another grey to the colour palette? I would speak to Tom about whether the same outcome could be achieved using one of the existing greys. 

They are also currently in order of dark to light but this 5th item feels like it fits somewhere in the middle.

src/views/_components/search-bar.njk
```diff
@@ -0,0 +1,45 @@
...
+    </label>
+    <div class="search-bar__input-wrapper">
+      <input class="search-bar__input"
+             id="search"
+             type="search"
+             name="term"
+             value="{{ searchTerm }}"
+             placeholder="{{ placeholder }}"
+      >
+      <input class="search-bar__submit" type="submit" value="{{ submitText }}"/>
```
teneightfive: Sorry, missed this one earlier but also doesn't need the `/`

src/views/_components/search-nav.njk
```diff
@@ -0,0 +1,33 @@
...
+ #    searchEntityResults: [{count:1, text:"Companies", entity:"company"}],
+ #    searchType: "company",
+ #    searchTerm: "example search term"
+ #  } %}
+ #}
+
+{% if searchEntityResults | length %}
+  <ul class="search-nav">
+    {% for searchEntityResult in searchEntityResults %}
+      {% set currentPage = searchEntityResult.entity === searchType %}
```
teneightfive: Would `isCurrentPage` be a better name for this? `currentPage` could suggest that it would be an int value rather than a bool.

src/views/search/results-contact.njk
teneightfive: Feels like its concerned with this PR if we're adding the style for this component though. 

If we're thinking in terms of BEM we should probably be applying a specific style to how we want the heading in the result to be displayed.

Also feels like there might even be two components here: `result-list` (which could be a more generically named component) and `result-item`?

test/components/search-bar.test.js
```diff
@@ -0,0 +1,55 @@
...
+    expect(searchInputLabel.textContent).to.equal(mockData.label)
+    expect(searchInput.value).to.equal(mockData.searchTerm)
+    expect(searchInput.placeholder).to.equal(mockData.label)
+    expect(searchSubmit.value).to.equal('Search')
+  })
+
+  it('should render as expected with overridden defaults', () => {
+    const mockData = {
+      action: '/example/action',
+      method: 'get',
```
teneightfive: Should this not be set to something other than the default if we're testing that it will carry this value?

test/components/search-bar.test.js
```diff
@@ -0,0 +1,55 @@
+const { renderComponentToDom } = require('../component-helper')
+
+describe('Search Bar component', () => {
+  it('should render as expected with defaults', () => {
+    const mockData = {
+      action: '/example/action',
+      method: 'get',
```
teneightfive: Should we remove `method` and `label` if they have default values?

test/controllers/search.controller.test.js
```diff
@@ -144,3 +129,173 @@ describe('Search controller', function () {
...
+              token: '1234'
+            },
+            query: {
+              term: 'mock term'
+            }
+          }, {}, this.next
+        )
+
+        expect(this.next.calledOnce).to.be.true
+        done()
```
teneightfive: Does it need the done callback?

test/lib/pagination.test.js
```diff
@@ -10,7 +10,7 @@ describe('Pagination', () => {
       }
 
       const result = {
-        total: 1000
+        count: 1000
```
teneightfive: This line is repeated a lot of times. Can it be moved to a higher scope and overridden if necessary within a test block?

src/controllers/search.controller.js
feedmypixel: Yes it appears we are talking about totally separate things :)
>Also that a route concerned with getting a company is in the search controller.

yup I added the TODO around this its in the wrong place


Basically yes, the `/viewcomapnyresult` work needs a look at. There was some redirect stuff that I removed from `viewCompanyResult` and gave it a little spruce but didn't want to carry on down a refactor hole.

src/lib/controller-utils.js
```diff
@@ -61,6 +61,7 @@ function transformV2Errors (sourceErrors) {
   return errors
 }
 
+// TODO this is very similar to /src/lib/url-helpers.buildQueryString Maybe time for a common folder?
```
feedmypixel: maybe `shared` is better, with things in folders and semantic filenames

`shared/url-helpers/build-query-string.js`

I hear you though, I'm the same with `/lib` it can quite quickly become a dumping ground for everything!

src/sass/settings/_colours.scss
feedmypixel: took it from the sketch file. I was going to have a chat with him when in to see where he is getting the colours from.

feedmypixel: A side thought it would be good to have a dh/designers group on here and people on it who wanted to be @jim68000 can we make this so?

test/controllers/search.controller.test.js
feedmypixel: yup

config/nunjucks/filters.js
feedmypixel: have gone with `results-highlight` for the moment can be changed if it is used in multiple places and if search results and results can be combined

config/nunjucks/index.js
```diff
@@ -36,24 +36,29 @@ function ComponentExtension (env) {
...
+  const tradeElementsFilters = require(`@uktrade/trade_elements/dist/nunjucks/filters`)
+  const dataHubFilters = require('./filters')
+
+  const filters = Object.assign({}, tradeElementsFilters, dataHubFilters)
+
+  Object.keys(filters).forEach((filter) => {
+    return env.addFilter(filter, filters[filter])
```
tyom: Don't need to return this either.

config/nunjucks/filters.js
```diff
@@ -0,0 +1,27 @@
...
+    )
+  },
+
+  removeNull: (array) => {
+    return array.filter(item => item)
+  },
+
+  pluralise: (string, count) => {
+    if (count !== 1) {
+      string += 's'
```
teneightfive: Not sure this covers all bases. What about `delivery` and `deliveries`? Here is the [django implementation](https://github.com/django/django/blob/master/django/template/defaultfilters.py#L844)

tyom: Alternatively you can pass another argument with custom word for pluralisation. I think that would be easier.

feedmypixel: I'm going to call YAGNI on this - lets add this when required by the frontend

> Alternatively you can pass another argument with custom word for pluralisation. I think that would be easier.

sounds good

tyom: Or just use [pluralize](https://www.npmjs.com/package/pluralize) :)

teneightfive: >Or just use pluralize :)

That. Also, I can see us needing it -> `company` and `companies`.

And thinking of, we could do better with our results text using pluralize and say _10 companies_ or _1 contact_ rather than a generic _5 results_?

src/sass/components/_pagination.scss
```diff
@@ -0,0 +1,14 @@
+@import "../settings/colours";
+@import "../tools/mixins/typography";
+
+.pagination-nav {
```
tyom: We should call components by their file name. Why not just `pagination`?

src/sass/components/_pagination.scss
```diff
@@ -0,0 +1,14 @@
...
+  margin: 40px auto;
+  padding-top: 20px;
```
tyom: What's this padding for?

src/sass/components/_pagination.scss
```diff
@@ -0,0 +1,14 @@
+@import "../settings/colours";
+@import "../tools/mixins/typography";
+
+.pagination-nav {
+  margin: 40px auto;
```
tyom: `margin: auto` is redundant without box dimensions.

src/sass/components/_results.scss
tyom: Selector and component file name should be the same.

src/sass/settings/_colours.scss
teneightfive: Yes would be useful to loop them into the conversations.

I would say for now though we should pick one of the existing greys that fits best rather than adding to the colour palette. Sketch files can often get littered with variations of colours so I think we should be sensible when picking colours from them and seeing if we can find a similar match in the existing palette. 

Looking where it's used `$grey-1` would work for the input border (or even `$black` which I think is used for input borders elsewhere) and possibly `$grey-1` or `$grey-2` for the borders as we've used that colour previously for borders.

src/sass/components/_results.scss
```diff
@@ -0,0 +1,39 @@
+@import "../tools/functions/typography";
+@import "../settings/colours";
+
+.results-list {
+  border-top: 1px solid $grey-5;
+  list-style: none;
+  margin-top: 10px;
+}
+
+.results-item {
```
tyom: Is this a different block or is this an element of this one? If it's a block it should have its own file.

config/nunjucks/index.js
teneightfive: Shall we move this back into the module for now then?

src/sass/components/_results.scss
```diff
@@ -0,0 +1,39 @@
...
+.results-item__header {
+  color: $link-colour;
+  margin-bottom: 10px;
+}
+
+.results-highlight {
+  @include bold-font(20);
+}
+
+.results-bar {
```
tyom: BEM?

src/sass/components/_results.scss
```diff
@@ -0,0 +1,39 @@
...
+  text-align: right;
+}
+
+.results-summary {
```
tyom: BEM?

src/sass/components/_results.scss
```diff
@@ -0,0 +1,39 @@
...
+  padding: 20px 0;
+  border-bottom: 1px solid $grey-5;
+}
+
+.results-item__header {
+  color: $link-colour;
+  margin-bottom: 10px;
+}
+
+.results-highlight {
```
tyom: BEM?

src/sass/components/_search-bar.scss
```diff
@@ -0,0 +1,36 @@
...
+}
+
+.search-bar__submit {
+  position: absolute;
+  width: 50px;
+  height: 100%;
+  top: 0;
+  right: 0;
+  border: 0;
+  text-indent: -500px;
```
tyom: I think you can use `@include hide-text(50px, 100%)` from `tools/mixins/addons`.

src/views/_components/search-bar.njk
```diff
@@ -0,0 +1,45 @@
...
+  action="{{ action }}"
+  method="{{ method }}"
+  role="search"
+>
+  <div class="search-bar__wrapper">
+    <label class="visually-hidden" for="search">
+      {{ label }}
+    </label>
+    <div class="search-bar__input-wrapper">
+      <input class="search-bar__input"
```
tyom: Should drop the `class` attribute on new line to be consistent.

src/views/_components/pagination.njk
tyom: Does it need to have `pagination` object? Could we just pass individual things to it (`label`, `link`, `currentPage`)? What other parameters could it have?

config/nunjucks/filters.js
feedmypixel: >  10 companies or 1 contact rather than a generic 5 results?

yup nice

src/sass/components/_results.scss
feedmypixel: this is a different component not a child

src/sass/components/_results.scss
feedmypixel: this is a different component not a child

src/sass/components/_results.scss
feedmypixel: this is a different component not a child

src/sass/components/_results.scss
feedmypixel: Different block - will do

src/sass/components/_pagination.scss
feedmypixel: `pagination-nav` due to`pagination` being in use in Trade Elements. Filename changed to `pagination-nav` and `//TODO added`

src/views/search/results-company.njk
```diff
@@ -0,0 +1,56 @@
...
+          <dd>
+            {% for addressDetail in [
+              company.registered_address_1,
+              company.registered_address_2,
+              company.registered_address_3,
+              company.registered_address_4,
+              company.registered_address_county,
+              company.registered_address_town,
+              company.registered_address_postcode
+            ] | removeNull %}
```
tyom: Do you need `removeNull` here if you're checking for falsy value below?

src/views/search/results-contact.njk
```diff
@@ -0,0 +1,42 @@
...
+      <dl>
+        {% if (contact.company) %}
+          <dd>{{ contact.company.name }}</dd>
+        {% endif %}
+        {% if results.job_title %}
+          <dd>{{ results.job_title }}</dd>
+        {% endif %}
+
+        <dd>
+          {%- for addressDetail in [
```
tyom: This looks exactly the same as one in `results-company.njk`. Could we make a macro to keep it DRY?

src/sass/components/_pagination.scss
tyom: Could we import individual trade elements instead of importing them all in one go?

src/sass/components/_results.scss
tyom: Will it ever be used outside of results component?

src/sass/components/_results.scss
tyom: Is it really standalone?

src/sass/components/_results.scss
tyom: Premature componentanisation? 😄

src/views/search/results-company.njk
feedmypixel: yeah this is to remove any nulls before the code enters the loop, otherwise there may be empty values after non empty values which cause problems with the `loop.last` comma insertion.

so
`[null, null, 'ben', 'null', null']` would give you `ben,`
where as 
`['ben']` would give you `ben`

src/sass/components/_search-bar.scss
```diff
@@ -0,0 +1,38 @@
...
+  position: relative;
+}
+
+.search-bar__input {
+  background-color: $white;
+  border: 1px solid $grey-5;
+  color: $black;
+  box-sizing: border-box;
+  width: 100%;
+  @include core-font(20);
```
tyom: Mixins should go first. Then individual properties can be overridden.

src/sass/components/_search-bar.scss
```diff
@@ -0,0 +1,38 @@
...
+
+.search-bar__submit {
+  position: absolute;
+  width: 50px;
+  height: 100%;
+  top: 0;
+  right: 0;
+  border: 0;
+  padding: 0;
+  @include hide-text(50px, 100%);
```
tyom: Mixins should go first.

src/sass/components/_search-bar.scss
```diff
@@ -0,0 +1,38 @@
...
+  color: $black;
+  box-sizing: border-box;
+  width: 100%;
+  @include core-font(20);
+  padding: 6px 50px 6px 6px;
+}
+
+.search-bar__submit {
+  position: absolute;
+  width: 50px;
```
tyom: You don't need width and height any more.

src/sass/_components.scss
```diff
@@ -1,2 +1,9 @@
 @import "components/local-nav";
+@import "components/pagination-nav";
+@import "components/results-bar";
+@import "components/results-list";
```
tyom: I'm not sure why we need the whole load of `results-*` components. Aren't they all part of one `results` component?

src/sass/_components.scss
```diff
@@ -1,2 +1,9 @@
...
+@import "components/results-item";
+@import "components/results-summary";
+@import "components/search-bar";
```
tyom: Same with `search-bar`. Could this be single component? Do we intend to use them individually?

src/sass/_components.scss
feedmypixel: yeah - they were in one file but you recommended if different blocks then split them out to separate files. I don't feel like it is a massive BEM block with children it feels like separate blocks in the same family. Personally I would stick with what I had - them all in a results files

src/sass/_components.scss
feedmypixel: Maybe these (this and the results) are better to sit with each other and work out?

feedmypixel: So search bar is only on the results page. Search can sit by itself

config/nunjucks/index.js
teneightfive: Think this still needs looking at. It should either return the combined filters be called something like `getFilters()` and the loop and `.addFilter()` call remains in the module or we leave the whole block within the module.

src/views/_components/pagination.njk
feedmypixel: >Does it need to have pagination object? Could we just pass individual things to it (label, link, currentPage)? What other parameters could it have?

Maybe I think we will know when we get an overview of all the components we have and can group them and pass information in like this

src/sass/trumps/_text.scss
```diff
@@ -0,0 +1,5 @@
+@import "../tools/functions/typography";
+
+.text-highlight {
+  font-weight: bold !important;
```
teneightfive: Does this need `!important`? The idea of trumps being at the bottom of the cascade should negate the need for it shouldn't it?

src/services/search.service.js
```diff
@@ -1,26 +1,24 @@
-const authorisedRequest = require('../lib/authorised-request')
-const config = require('../config')
-const includes = require('lodash/includes')
 const Q = require('q')
 const winston = require('winston')
+const authorisedRequest = require('../lib/authorised-request')
+const config = require('../config')
+const {buildQueryString} = require('../lib/url-helpers')
```
teneightfive: Space between braces

src/views/_components/results-summary.njk
```diff
@@ -0,0 +1,24 @@
...
+ # @param {string} resultType - Type of results that have been searched for
+ # @param [{string}] pluralisedResultType - Pluralised result type for when the appending of an "s" won't do
+ # @param {int} total - Total number of results
+ # @param {string} searchTerm - Search term used to obtain results
+ #
+ # @return {string} HTML for results summary
+ #
+ # @example
+ #   {% component 'results-summary', {
+ #     resultType: 'company',
```
teneightfive: 👍

src/views/_macros/address.njk
```diff
@@ -0,0 +1,23 @@
+{##
+ # @param {string[]}   addressDetails - address parts
```
teneightfive: Drop the `[]`

src/sass/_components.scss
tyom: Apologies for the confusion. I merely asked 
> Is this a different block or is this an element of this one? If it's a block it should have its own file.

I didn't mean to split them into new components as they all appear to be part of one component.

src/views/_macros/address.njk
```diff
@@ -0,0 +1,23 @@
...
+ # @return {string} HTML address
+ #
+ # @example
+ #   {{ formatter([
+ #     'first line of address',
+ #     'second line of address',
+ #     'AL1 1DF'
+ #   ]) }}
+ #}
+{% macro formatter(addressDetails, searchTerm) %}
```
teneightfive: Should this actually be a macro or could this be a function or filter in nunjucks?

src/views/search/results-contact.njk
```diff
@@ -0,0 +1,44 @@
...
+    resultType: searchType,
+    total: results.count,
+    searchTerm: searchTerm
+  } %}
+{% endblock %}
+
+{% block results %}
+  <ol class="results-list">
+    {% for contact in results.contacts %}
+    <li class="results-item">
```
teneightfive: Needs indenting

src/views/search/results-contact.njk
```diff
@@ -0,0 +1,44 @@
...
+      <h3 class="results-item__header">
+        <a href="/contact/{{ contact.id }}/details?term={{ searchTerm }}">
+          {{ (contact.first_name + ' ' + contact.last_name) | highlight(searchTerm) }}
+        </a>
+      </h3>
+
+      <dl>
+        {% if (contact.company) %}
+          <dd>{{ contact.company.name }}</dd>
```
teneightfive: If this is using a `dl` shouldn't we add a `dt` and visually hide?

config/nunjucks/index.js
feedmypixel: If we need to work with multiples of `addGlobals`, `addExtensions` then we should abstract the work into a method which gets these multiples, loops over them and applies them to a method.

src/sass/settings/_colours.scss
feedmypixel: I choose `$grey-1` :)

src/views/_macros/address.njk
tyom: Why? Isn't it array of strings?

feedmypixel: JSDOC says `{string[]}` is an array of strings (looked it up special like)

src/views/search/_layout.njk
```diff
@@ -0,0 +1,51 @@
+{% extends "layouts/_base-two-column.njk" %}
+{% from "_macros/address.njk" import formatter as addressFormatter %}
```
tyom: Could just name the macro `addressFormatter` since it's specific to addresses, then
```njk
{% from "_macros/address.njk" import addressFormatter %}
```

Maybe stick that macro in some generic macros file. Do we need this many macro files?

src/sass/trumps/_text.scss
tyom: Except when specificity wins.

feedmypixel: always felt trumps should have `!important` TBH

src/sass/_components.scss
feedmypixel: ok no worries ill stick them back - It id feel quite overkill 👍

src/services/search.service.js
```diff
@@ -57,33 +55,14 @@ function suggestCompany (token, term, types) {
...
 function searchLimited (token, term) {
   return new Promise((resolve, reject) => {
     Q.spawn(function * () {
       try {
-        const allResults = yield search({token, term, filters: ['company_company', 'company_companieshousecompany']})
+        const allResults = yield search({ token, term, filters: ['company_company', 'company_companieshousecompany'] })
         const filtered = allResults.hits.filter(result => (result._type === 'company_companieshousecompany' ||
-          (result._source.business_type && result._source.business_type.toLowerCase() === 'private limited company') ||
-          (result._source.business_type && result._source.business_type.toLowerCase() === 'public limited company')))
+        (result._source.business_type && result._source.business_type.toLowerCase() === 'private limited company') ||
```
tyom: Could use Lodash's get to make it more succinct:
```js
const filtered = allResults.hits.filter(result => {
  const businessType = get(result, '_source.business_type', '').toLowerCase()
  return result._type === 'company_companieshousecompany' && ['private limited company', 'public limited company'].includes(businessType)
})
```

feedmypixel: Looks good. FYI this is just a white space change. I've had nothing to do with this method. Will revert in the morning. The only real change I wanted here was the spaces either side of the curlys

feedmypixel: Also `company_companieshousecompany` has gone away. This needs fixing all over the app.

src/sass/components/_lists.scss
feedmypixel: aligned to use `.list-inline` as suggested by @tyom

src/views/search/_layout.njk
feedmypixel: aligned with `snake_case`

src/views/search/results-company.njk
feedmypixel: Made them `dl` with `u-hidden` `dt`

config/nunjucks/index.js
teneightfive: Had a thought last night that if we're using local filters should we import the ones from trade elements if we're talking about moving that dependency over anyway. 

Still not sure I agree that this function should be concerned with the nunjucks env. 

If we import the filters from trade elements then we can define the require import at the top of the file (`dataHubFilters` or just `filters`) and then all that needs to be in the module is:

```
Object.keys(filters).forEach((filter) => {
  env.addFilter(filter, filters[filter])
})
```

This feels much cleaner and that we possibly don't need that abstraction any more?

src/views/_components/pagination.njk
teneightfive: I'm be tempted to go with @tyom's suggestion now for clarity. If the component needs to support other params we add them. 

But we should avoid just passing an object to the component that then expects properties if possible.

src/views/_components/pagination.njk
```diff
@@ -0,0 +1,25 @@
+{##
+ # @param {object} pagination - An array of objects with properties {label, link, currentPage}
+ #
+ # @return {string} HTML for pagination navigation
+ #
+ # @example
+ #   {% component 'pagination', {
+ #     pagination: pagination
```
teneightfive: This example isn't very clear what it should expect. If we go with params we should give examples of what they should contain rather than variables.

src/views/_components/search-bar.njk
```diff
@@ -0,0 +1,46 @@
+{##
+ # @param {string} action - Search forms action attribute value
```
teneightfive: We should be consistent with the spacing method we use for component docs. Sometimes we've used spaces to align the description but here it uses a combo of a single space of a double space before the `-`

src/views/_components/search-bar.njk
```diff
@@ -0,0 +1,46 @@
...
+ # @param {string} [method='get'] - Search forms method attribute value
+ # @param {string} label  - Search forms label text
+ # @param {string} searchTerm  - Search term used for search forms input value
+ # @param {string} [placeholder=label]  - Text used for Search inputs placeholder, defaults to label
+ # @param {string} [submitText='Search']  - Text used for Search submit, defaults to "Search"
```
teneightfive: Don't think the default needs the quotes

teneightfive: And can also remove the `, defaults to "Search"` at the end of the description

src/views/_components/search-bar.njk
```diff
@@ -0,0 +1,46 @@
+{##
+ # @param {string} action - Search forms action attribute value
+ # @param {string} [method='get'] - Search forms method attribute value
+ # @param {string} label  - Search forms label text
+ # @param {string} searchTerm  - Search term used for search forms input value
+ # @param {string} [placeholder=label]  - Text used for Search inputs placeholder, defaults to label
```
teneightfive: Can also remove the `, defaults to...` from the end of the description

src/views/investment/start.njk
```diff
@@ -106,19 +106,9 @@
...
-                <div class="pagination__page">{{ page.label }}</div>
-              {% else %}
-                <a href="?{{ page.link }}" class="pagination__page">{{ page.label }}</a>
-              {% endif %}
-            {% endfor %}
-          </div>
-        </div>
-      {% endif %}
+      {% component 'pagination', {
+        pagination: pagination
```
teneightfive: As tyom suggested I think this would also read better if it's something like this:

```js
{% component 'pagination', {
  children: [{ link: '/link-path', label: 'Link text' }],
  currentPage: '/page'
} %}
```

src/views/search/results-company.njk
```diff
@@ -0,0 +1,58 @@
...
+        {% endif %}
+
+        <h3 class="results-item__header">
+          <a href="{{ companyUrl }}">
+            {{ company.name | highlight(searchTerm) }}
+          </a>
+        </h3>
+
+        <dl>
+          <dt class="u-hidden">Company Details</dt>
```
teneightfive: Should probably have two `dt`s in this instance. One for `Company number` and one for `(Company) address`?

src/views/search/results-contact.njk
```diff
@@ -0,0 +1,45 @@
...
+    {% for contact in results.contacts %}
+      <li class="results-item">
+        <h3 class="results-item__header">
+          <a href="/contact/{{ contact.id }}/details?term={{ searchTerm }}">
+            {{ (contact.first_name + ' ' + contact.last_name) | highlight(searchTerm) }}
+          </a>
+        </h3>
+
+        <dl>
+          <dt class="u-hidden">Contact Details</dt>
```
teneightfive: Same here about number of `dt`s. Would make more sense to have `Company name`, `Job title`, and `Address`

src/views/_components/pagination.njk
feedmypixel: Agreed but I feel this is another ticket. This component is consuming what is provided by [pagination.js](https://github.com/uktrade/data-hub-fe-beta2/blob/develop/src/lib/pagination.js). The ticket would be to refactor `pagination.js` and the component.

config/nunjucks/index.js
feedmypixel: Yeah that feels good and agreed the extra work at the moment is because we are combining from Trade Element filters. Lets do this when:

> If we import the filters from trade elements

test/controllers/search.controller.test.js
teneightfive: I'm not sure you do in this instance as nothing is being done asynchronously. If you remove the `done()` callback from this line and don't pass it in as a param on line 165 the test should still work as expected.

src/views/_components/pagination.njk
teneightfive: Okay, if we keep it as an array of objects lets change the name to something clearer like `children`?

And should we update the type to be `{object[]}`?

src/views/_components/search-bar.njk
tyom: Could probably remove it "default to …" from description too if you're adding it in param name.

config/nunjucks/index.js
teneightfive: We should minimise the change to this file in this instance then if we're planning to do that. Otherwise when we import we have to undo this change and just creates extra noise.

I'd either say import it with this change as we're creating the filters file or leave this code within the module and import them in a small change after.

teneightfive: This change could just be 2 new lines and 1 removed line. Simple and easier to see the diff.

src/views/investment/start.njk
feedmypixel: Yeah this is nice. I think this falls under the above comment as it changes the interface.

>Agreed but I feel this is another ticket. This component is consuming what is provided by [pagination.js](https://github.com/uktrade/data-hub-fe-beta2/blob/develop/src/lib/pagination.js). The ticket would be to refactor `pagination.js` and the component.

test/controllers/search.controller.test.js
feedmypixel: my bad yes that works when the argument is removed

src/views/_components/pagination.njk
```diff
@@ -0,0 +1,31 @@
+{##
+ # @param {object[]} pagination An array of pagination page objects
+ #
+ # @return {string} HTML for pagination navigation
+ #
+ # @example
+ #   {% component 'pagination', {
+ #     pages: [
```
teneightfive: This doesn't match the param name above or what's used in the component. I prefer `pages` to `pagination` though so think we should change to that.

src/views/_components/search-bar.njk
```diff
@@ -0,0 +1,46 @@
+{##
+ # @param {string} action Search forms action attribute value
+ # @param {string} [method='get'] Search forms method attribute value
```
teneightfive: Doesn't need quotes around default value

src/views/_components/search-bar.njk
```diff
@@ -0,0 +1,46 @@
...
+ # @param {string} label Search forms label text
+ # @param {string} searchTerm Search term used for search forms input value
+ # @param {string} [placeholder=label] Text used for Search inputs placeholder
```
teneightfive: Just realised this default value could be confused with being a string. Not sure there's a good way to convey this though?

src/views/_macros/address.njk
teneightfive: Lack of JSDoc on my part there

src/views/_components/search-bar.njk
feedmypixel: Yeah that is what it is officially [optional-parameters-and-default-values](http://usejsdoc.org/tags-param.html#optional-parameters-and-default-values)

src/views/_components/pagination.njk
```diff
@@ -0,0 +1,31 @@
...
+ #       {
+ #         label: '2',
+ #         link: '?term=example&page=2',
+ #         currentPage: 'true'
+ #       }
+ #     ]
+ #   } %}
+ #}
+{% if pagination | length %}
```
teneightfive: These references still need updating to `pages`

src/views/search/results-contact.njk
```diff
@@ -0,0 +1,47 @@
...
+      <li class="results-list__item">
+        <h3 class="results-list__item-header">
+          <a href="/contact/{{ contact.id }}/details?term={{ searchTerm }}">
+            {{ (contact.first_name + ' ' + contact.last_name) | highlight(searchTerm) }}
+          </a>
+        </h3>
+
+        <dl>
+          {% if (contact.company) %}
+            <dt class="visually-hidden">Company</dt>
```
teneightfive: Should probably use the `u-` prefix version of these classes here so we don't have to change them when we remove the govuk-elements/trade-elements dependencies.

src/views/_components/results-summary.njk
```diff
@@ -0,0 +1,24 @@
+{##
+ # @param {string} resultType Type of results that have been searched for
+ # @param [{string}] pluralisedResultType Pluralised result type for when the appending of an "s" won't do
```
teneightfive: What's the `[{string}]` syntax? Is it meant to be conveying optional? If so, the `[]` should wrap the name not the type.

Might be worth setting a default of `results` for this too.

src/views/_components/results-summary.njk
```diff
@@ -0,0 +1,24 @@
+{##
+ # @param {string} resultType Type of results that have been searched for
```
teneightfive: Should this have a default? Possible `result`.

src/views/_components/search-nav.njk
```diff
@@ -0,0 +1,33 @@
...
+ #    searchTerm: "example search term"
+ #  } %}
+ #}
+
+{% if searchEntityResults | length %}
+  <ul class="search-nav">
+    {% for searchEntityResult in searchEntityResults %}
+      {% set isCurrentPage = searchEntityResult.entity === searchType %}
+
+      <li class="search-nav__item{{ " search-nav__item--selected" if isCurrentPage }}">
```
teneightfive: Could use single quotes for the nunjucks var to stop it messing with the syntax highlighting?

src/views/investment/start.njk
teneightfive: Need to update to:

```njk
{% component 'pagination', {
  pages: pagination
} %}
```

config/nunjucks/filters.js
```diff
@@ -0,0 +1,31 @@
+const nunjucks = require('nunjucks')
+const filters = {
+
```
tyom: Unnecessary new line?

config/nunjucks/index.js
```diff
@@ -43,16 +43,16 @@ module.exports = (app, config) => {
...
+  const filters = Object.assign({}, tradeElementsFilters, dataHubFilters)
 
   // Custom filters
-  const filters = require(`@uktrade/trade_elements/dist/nunjucks/filters`)
-
-  filters.stringify = JSON.stringify
-
-  Object.keys(filters).forEach((filterName) => {
-    env.addFilter(filterName, filters[filterName])
+  Object.keys(filters).forEach((filter) => {
```
tyom: Could use Lodash's `forOwn`
```js
forOwn(filters, (filterFunc, filterName) => {
  env.addFilter(filterName, filterFunc)
}
```

feedmypixel: I prefer vanilla for this TBH

src/controllers/search.controller.js
```diff
@@ -4,65 +4,88 @@ const { buildCompanyUrl } = require('../services/company.service')
...
-  const docType = query.doc_type
-  const filters = []
+function buildSearchEntityResultsData (searchEntityResults) {
+  const navItemText = {
+    company: 'Companies',
+    contact: 'Contacts'
+  }
+
+  return searchEntityResults.map((searchEntityResult) => {
+    return Object.assign({}, searchEntityResult,
```
tyom: Would it read better if each arg was on new line rather than some of them?
```js
Object.assign(
  {},
  searchEntityResult,
  {
    text: '…'
  }
)
```

src/views/_components/results-summary.njk
feedmypixel: :) yes, I did that but removed it as it looked noisy - will add in

src/sass/settings/_buttons.scss
```diff
@@ -0,0 +1 @@
+$search-button-width: 50px;
```
tyom: Does this need to be a global setting? If it does this file should be loaded from the `_settings.scss` loader.

src/views/search/_layout.njk
```diff
@@ -0,0 +1,51 @@
...
+  {% endblock %}
+
+
+  {% block main_grid_left_column %}
+    &nbsp; {# // TODO filters #}
+  {% endblock %}
+
+  {% block main_grid_right_column %}
+    {% if results %}
+      <div class="grid-row">
```
tyom: Why do you need grid row/column here? Wouldn't it have the same effect without?

src/views/search/_layout.njk
```diff
@@ -0,0 +1,51 @@
...
+          {% endblock %}
+
+          {% block results %}
+          {% endblock %}
+        </div>
+      </div>
+    {% endif %}
+
+    <div class="grid-row">
+      <div class="column-full">
```
tyom: Same here.

src/views/search/results-company.njk
```diff
@@ -0,0 +1,59 @@
+{% extends "./_layout.njk" %}
+
+{% block action_bar %}
```
tyom: Could get around without all these blocks by using `{{ super() }}`.

src/views/_components/results-summary.njk
```diff
@@ -0,0 +1,24 @@
+{##
+ # @param {string} [resultType=results] Type of results that have been searched for
```
teneightfive: Should be `result` not `results` 😳 . Might be worth putting something in the doc that it should be in singular form.

feedmypixel:
Thanks for all the 👀  on this

PR #143, opened 2017-05-31
Remember requested url when redirecting unauthorised requests

teneightfive:
When a user requests an authorised url and is not authorised this
will store that url and redirect back there after logging in.

Will default to the homepage if there was no requested url.

### DIFF
--- a/src/controllers/login.controller.js
+++ b/src/controllers/login.controller.js
@@ -1,16 +1,15 @@
 /* eslint new-cap: 0 */
 
-const express = require('express')
+const router = require('express').Router()
 const rp = require('request-promise')
-const config = require('../config')
+const { get } = require('lodash')
 
-const router = express.Router()
+const config = require('../config')
 
 function authenticate (username, password) {
   const options = {
     method: 'POST',
-    url: (config.apiRoot + config.api.authUrl),
-
+    url: config.apiRoot + config.api.authUrl,
     headers: {
       'cache-control': 'no-cache',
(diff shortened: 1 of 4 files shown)


src/controllers/login.controller.js
```diff
@@ -41,7 +41,7 @@ function loginToApi (req, res, next) {
   authenticate(req.body.username, req.body.password)
     .then((data) => {
       req.session.token = data.access_token
-      res.redirect('/')
+      res.redirect(req.session.returnTo || '/')
     })
     .catch((error) => {
       if (error.response && error.response.statusCode && error.response.statusCode === 401) {
```
tyom: Unrelated but still your could improve it by using Lodash's `get`.
```js
const statusCode = get(error, 'response.statusCode')
if (statusCode === 401) { /* … */ }
```

src/middleware/auth.js
```diff
@@ -1,14 +1,16 @@
...
 module.exports = function auth (req, res, next) {
-  winston.debug('auth:start')
+  const url = req.url
 
-  if (!req.url.startsWith('/ping.xml') && !req.url.startsWith('/support') && req.url !== '/login' && req.url !== '/error' && !req.session.token) {
-    winston.debug('auth:redirect login')
-    res.redirect('/login')
-    return
+  if (
+    url.startsWith('/ping') ||
```
tyom: Could name these in a variable to improve readability.
```js
const passThrough = /^\/(support|ping|login)\b/.test(url) || req.session.token
```

PR #147, opened 2017-06-01
A new sortable table component

ztolley:
Loosely based on the sortable table in Trade elements but with many improvements and made simpler. Adopts the new component framework.

The example shown isn't in the project (we all know why) but it demonstrates the control.

If this works it could be used to replace existing tables.
sortable

### DIFF
--- a/config/nunjucks/filters.js
+++ b/config/nunjucks/filters.js
@@ -1,4 +1,6 @@
 const nunjucks = require('nunjucks')
+const _ = require('lodash')
+
 const {
   isArray,
   isPlainObject,
@@ -13,6 +15,10 @@ function isNotEmpty (value) {
 
 const filters = {
   stringify: JSON.stringify,
+
+  values: _.values,
+
+  keys: _.keys,
 
   highlight: (string, searchTerm) => {
     const regEx = new RegExp(`(${searchTerm})`, 'gi')
--- a/src/javascripts/app.js
+++ b/src/javascripts/app.js
@@ -1,3 +1,5 @@
 const ConditionalSubfields = require('./modules/conditional-subfields')
+const SortableTable = require('./modules/sortable-table')
 
 ConditionalSubfields.init()
+SortableTable.init()
--- a/src/javascripts/modules/sortable-table.js
+++ b/src/javascripts/modules/sortable-table.js
@@ -1,0 +1,148 @@
+const { guid } = require('@uktrade/trade_elements')
+
+const { parseDateString } = require('../../lib/date')
+const { addClass, removeClass } = require('../../lib/element-stuff')
+
+function isDate (text) {
+  return parseDateString(text) !== null
+}
+
+function convertToDate (dateString) {
+  return parseDateString(dateString)
+}
+
+const ASC_CLASS = 'is-ascending'
+const DESC_CLASS = 'is-descending'
+
+class SortableTable {
+  constructor (element, document) {
+    this.element = element
+    this.cacheElements()
+    this.attachEvents()
+    this.parseTable()
+    this.sortAsc = true
+    this.currentKey = null
+    this.document = document
+  }
+
+  cacheElements () {
+    this.headingElements = this.element.querySelectorAll('thead th')
+    this.tableBody = this.element.querySelector('tbody')
+  }
+
+  attachEvents () {
+    this.headingElements.forEach((element) => {
+      element.addEventListener('click', this.handleHeadingClick.bind(this))
+    })
+  }
+
+  parseTable () {
+    this.keys = SortableTable.assignTableKeys(this.headingElements)
+    this.data = SortableTable.parseTableBody(this.element, this.keys)
+  }
+
+  handleHeadingClick (event) {
+    const key = event.target.getAttribute('data-key')
+
+    if (this.currentKey && key === this.currentKey) {
+      this.sortAsc = !this.sortAsc
+    } else {
+      this.sortAsc = true
+      this.currentKey = key
+    }
+
+    this.data = SortableTable.sort(this.data, key, this.sortAsc)
+    this.tableBody.innerHTML = SortableTable.render(this.data)
+
+    this.clearSortClasses()
+
+    if (this.sortAsc) {
+      addClass(event.target, ASC_CLASS)
+    } else {
+      addClass(event.target, DESC_CLASS)
+    }
+  }
+
+  clearSortClasses () {
+    removeClass(this.headingElements, ASC_CLASS)
+    removeClass(this.headingElements, DESC_CLASS)
+  }
+
+  static assignTableKeys (headingElements) {
+    let keys = []
+    // parse the headings into an array of keys.
+    headingElements.forEach((headingElement) => {
+      const key = guid()
+      keys.push(key)
+      headingElement.setAttribute('data-key', key)
+    })
+
+    return keys
+  }
+
+  static parseTableBody (tableElement, keys) {
+    let rowsData = []
+    const rows = tableElement.querySelectorAll('tbody tr')
+
+    for (let rowPos = 0; rowPos < rows.length; rowPos += 1) {
+      const rowElement = rows.item(rowPos)
+      const rowCells = rowElement.querySelectorAll('td')
+      rowsData[rowPos] = {}
+      for (let col = 0; col < keys.length; col += 1) {
+        rowsData[rowPos][keys[col]] = rowCells[col].textContent
+      }
+    }
+
+    return rowsData
+  }
+
+  static sort (data, key, sortAsc = true) {
+    return data.sort((a, b) => {
+      let aValue = a[key]
+      let bValue = b[key]
+
+      if (isDate(aValue) && isDate(bValue)) {
+        aValue = convertToDate(aValue)
+        bValue = convertToDate(bValue)
+      }
+
+      let result
+      if (aValue < bValue) {
+        result = -1
+      } else if (aValue > bValue) {
+        result = 1
+      } else {
+        result = 0
+      }
+
+      if (!sortAsc) {
+        result = result - (result * 2)
+      }
+
+      return result
+    })
+  }
+
+  static render (data) {
+    let html = ''
+    for (const record of data) {
+      let string = '<tr>'
+      for (const key in record) {
+        string += `<td>${record[key]}</td>`
+      }
+      string += '</tr>'
+      html += string
+    }
+    return html
+  }
+
+  static init () {
+    const elements = document.querySelectorAll('.js-SortableTable')
+    for (let pos = 0; pos < elements.length; pos += 1) {
+      const element = elements.item(pos)
+      new SortableTable(element)   // eslint-disable-line no-new
+    }
+  }
+}
+
+module.exports = SortableTable
--- a/src/lib/date.js
+++ b/src/lib/date.js
@@ -1,20 +1,37 @@
 const { longDateFormat, mediumDateFormat } = require('../config')
-const moment = require('moment')
+const format = require('date-fns/format')
 
-function formatLongDate (date) {
-  if (!date) {
-    return null
+function formatLongDate (dateString = []) {
+  if (dateString) {
+    return format(parseDateString(dateString), longDateFormat)
   }
-  return moment(date).format(longDateFormat)
+
+  return null
 }
 
-function formatMediumDate (date) {
-  if (!date) {
-    return null
+function formatMediumDate (dateString = []) {
+  if (dateString) {
+    return format(parseDateString(dateString), mediumDateFormat)
   }
-  return moment(date).format(mediumDateFormat)
+
+  return null
+}
+
+function parseDateString (dateString) {
+  if (dateString.indexOf('/') !== -1) {
+    const parts = dateString.split('/')
+    return new Date(parts[2], parts[1] - 1, parts[0])
+  } else {
+    const date = new Date(dateString)
+    if (date.toString() !== 'Invalid Date') {
+      return date
+    }
+  }
+  return null
 }
 
 module.exports = {
-  formatLongDate, formatMediumDate,
+  formatLongDate,
+  formatMediumDate,
+  parseDateString,
 }
--- a/src/views/_components/data-table.njk
+++ b/src/views/_components/data-table.njk
@@ -1,0 +1,55 @@
+{##
+ # @param {object[]} data              Objects to show in the table
+ # @param {object}   [columns]         Key/value pair of columns to show and their label
+ # @param {string}   [variant=striped] Variant of table to display
+ # @param {boolean}  [sortable=false]  Adds ability to sort each column
+ #
+ # @return {string} HTML for table
+ #
+ # @example
+ #  {% component 'data-table', {
+ #     variant: 'striped',
+ #     columns: {
+ #       name: 'Name',
+ #       addr: 'Address'
+ #     },
+ #     data: [{
+ #       id: '134',
+ #       name: 'Fred Bloggs',
+ #       addr: '10 Fred Road, Humbleton, Berks, SL5 5RR'
+ #     }, {
+ #       id: '4444',
+ #       name: 'Jane Smith',
+ #       addr: '32 Palace Place, London, SE1 1TR'
+ #     }]
+ #  } %}
+ #}
+{% if data.length %}
+  {% set sortable = sortable | default(false) %}
+  {% set keys = data[0] | keys %}
+  {% set headings = data[0] | keys %}
+
+  {% if columns %}
+    {% set keys = columns | keys %}
+    {% set headings = columns | values %}
+  {% endif %}
+
+  <table class="data-table{% if sortable %} is-sortable js-SortableTable{% endif %}{{ ' data-table--' + variant if variant }}">
+    <thead>
+      <tr>
+        {% for heading in headings %}
+          <th>{{ heading }}</th>
+        {% endfor %}
+      </tr>
+    </thead>
+    <tbody>
+      {% for record in data %}
+        <tr>
+          {% for key in keys %}
+            <td>{{ record[key] }}</td>
+          {% endfor %}
+        </tr>
+      {% endfor %}
+    </tbody>
+  </table>
+{% endif %}
--- a/test/components/data-table.test.js
+++ b/test/components/data-table.test.js
@@ -1,0 +1,170 @@
+const { renderComponentToDom } = require('../component-helper')
+
+const data = [{
+  id: '1234',
+  name: 'Fred Smith',
+  address: '10 The Street',
+  uk: true,
+}, {
+  id: '5678',
+  name: 'John Brown',
+  address: '10 The Street',
+  uk: true,
+}]
+
+const columns = {
+  'name': 'Name',
+  'address': 'Address',
+}
+
+describe('Data table component', () => {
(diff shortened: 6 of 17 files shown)


src/sass/components/_sortable-table.scss
```diff
@@ -0,0 +1,17 @@
+.data-table--sortable {
```
tyom: The file name doesn't correspond to component name. Should it be a modifier if there is no component called `data-table`?

ztolley: Happy to hear views on naming. Think of it as a regular table (which inherits the table styles from GDS) that is made sortable

data-table{% if sortable %} js-data-table--sortable data-table--sortable {% endif %}

_data-table.scss


Would this be better?

teneightfive:
The js data hook could just be `js-SortableTable`. Doesn't need to include BEM style class names.

config/nunjucks.js
```diff
@@ -50,6 +50,7 @@ module.exports = (app, config) => {
   const filters = require(`@uktrade/trade_elements/dist/nunjucks/filters`)
 
   filters.stringify = JSON.stringify
+  filters.values = Object.values
```
teneightfive: @feedmypixel has extracted the filters out into their own module so beware you don't loose this when you rebase

src/javascripts/sortable-table.js
```diff
@@ -0,0 +1,148 @@
...
+function convertToDate (dateString) {
+  return parseDateString(dateString)
+}
+
+const ASC_CLASS = 'sort-asc'
+const DESC_CLASS = 'sort-desc'
+
+class TableSort {
+  constructor (element, document) {
+    console.log(element)
```
teneightfive: 🚨

src/javascripts/sortable-table.js
```diff
@@ -0,0 +1,148 @@
...
+
+function isDate (text) {
+  return parseDateString(text) !== null
+}
+
+function convertToDate (dateString) {
+  return parseDateString(dateString)
+}
+
+const ASC_CLASS = 'sort-asc'
```
teneightfive: Might be nicer to use state style classes for these. So `is-ascending` and `is-descending`.

src/javascripts/sortable-table.js
```diff
@@ -0,0 +1,148 @@
...
+const DESC_CLASS = 'sort-desc'
+
+class TableSort {
+  constructor (element, document) {
+    console.log(element)
+    this.element = element
+    this.cacheElements()
```
teneightfive: Don't know whether we want to be consistent with how we name our methods on modules. That's why I quite like [heisenbergjs](http://heisenbergjs.github.io/heisenberg/) (yes @tyom, I know it's old skool). But at least when you look in a module you always expect to see common method names and then immediately know what they're doing. So:

- `cacheEls()` - to cache any DOM elements used in the module
- `bindEvents()` - binds events to DOM elements
- `render()` - calls on first run of the module and can often run through and set the elements initial state

src/javascripts/sortable-table.js
```diff
@@ -0,0 +1,148 @@
...
+      }
+      string += '</tr>'
+      html += string
+    }
+    return html
+  }
+}
+
+if (typeof document !== 'undefined') {
+  const elements = document.querySelectorAll('.js-sortable-table')
```
teneightfive: I think it's good to use a consistent name for the js hook, class name and file name. At the moment you've got `js-sortable-table`, `TableSort` and `sortable-table`. It's also good to pascal case the js hook using the same name as the class name so it's easy to find. 

I'd suggest using `SortableTable` for the class name, `js-SortableTable` for the hook and `sortable-table.js` for the file name.

src/lib/date.js
```diff
@@ -15,6 +15,19 @@ function formatMediumDate (date) {
   return moment(date).format(mediumDateFormat)
 }
 
+function parseDateString (dateString) {
```
teneightfive: Could we not use moment for this? Seems like we're already including it as a dependency in this file anyway?

src/lib/date.js
```diff
@@ -15,6 +15,19 @@ function formatMediumDate (date) {
...
+    if (date.toString() !== 'Invalid Date') {
+      return date
+    }
+  }
+  return null
+}
+
 module.exports = {
-  formatLongDate, formatMediumDate
+  formatLongDate, formatMediumDate, parseDateString
```
teneightfive: New line for each?

src/sass/components/_sortable-table.scss
teneightfive: I would say use `_data-table.scss` for the filename and then possibly a state class for it being sortable: `is-sortable`. That could apply the cursor style and then as suggested above, use state classes for the cells of `is-asc` and `is-desc` that apply the icon style?

src/views/_components/data-table.njk
```diff
@@ -0,0 +1,68 @@
+{##
+ # @param {array[object]}     Data An array of objects to show in the table
```
teneightfive: I think the [JSDoc style](http://usejsdoc.org/tags-param.html#parameters-with-properties) for this is `{object[]}`

src/views/_components/data-table.njk
```diff
@@ -0,0 +1,68 @@
...
+ #       name: 'Fred Bloggs',
+ #       addr: '10 Fred Road, Humbleton, Berks, SL5 5RR'
+ #     },{
+ #       id: '4444',
+ #       name: 'Jane Smith',
+ #       addr: '32 Palace Place, London, SE1 1TR'
+ #     }]
+ #  %}
+ #}
+{% if (data and data.length > 0) or columns %}
```
teneightfive: Think you can just do `data.length`

test/dom-helpers.js
```diff
@@ -0,0 +1,10 @@
+function domTokenToArray (obj) {
```
teneightfive: Could probably just add this to [component-helper.js](https://github.com/uktrade/data-hub-fe-beta2/blob/develop/test/component-helper.js) rather than a new file

webpack.config.js
```diff
@@ -16,6 +16,7 @@ module.exports = {
     'service-delivery': './src/javascripts/service-delivery',
     'archive-form': './src/javascripts/archive-form',
     'add-another-field': './src/javascripts/add-another-field',
+    'sortable-table': './src/javascripts/sortable-table',
```
teneightfive: Feels like we should start moving these to a combined output file rather than multiple modules. [app.js](https://github.com/uktrade/data-hub-fe-beta2/blob/develop/src/javascripts/app.js) is probably the best place.

src/javascripts/sortable-table.js
tyom: I don't know why we need to use classes sugar anyway, especially on the client.

Also, have you noticed the size of the compiled JS?
image

That's minified. Moment is not tiny 😞

tyom: We should take a look at https://date-fns.org/

It's modular so you can pick the bits you need instead of importing the whole library to format date in one place.

test/javascripts/sortable-table.test.js
```diff
@@ -0,0 +1,197 @@
+const jsdom = require('jsdom')
+const { JSDOM } = jsdom
+const TableSort = require(`${root}/src/javascripts/sortable-table`)
+const { domTokenToArray } = require('../dom-helpers')
+
+const FIRST_NAME_SELECTOR = '.js-sortable-table tbody tr:first-child td:first-child'
+const LAST_NAME_SELECTOR = '.js-sortable-table tbody tr:last-child td:first-child'
+const HOOK = '.js-sortable-table'
+
+const HTML = `<html><body><table class="js-sortable-table">
```
tyom: Could you format it properly?

src/views/_components/data-table.njk
ztolley: I thought that too but no

src/javascripts/sortable-table.js
ztolley: Even if you strip out moment and use date-fns, the file ends up being 59K, 53K bigger than it needs to be. This is due to the fact there is no tree shaking taking place in webpack.

I'm switching to date-fns for now but to bring the file size down further we Must move to webpack 2.

webpack.config.js
ztolley: outside the scope of this card

teneightfive: I don't think it is. There is now a shared js file that we should include modules in so I would say we use that method moving forward rather than creating more separate files.

src/javascripts/modules/sortable-table.js
```diff
@@ -0,0 +1,147 @@
...
+  handleHeadingClick (event) {
+    const key = event.target.getAttribute('data-key')
+
+    if (!this.currentKey) {
+      this.sortAsc = true
+      this.currentKey = key
+    } else if (key === this.currentKey) {
+      this.sortAsc = !this.sortAsc
+    } else {
+      this.sortAsc = true
```
teneightfive: These are set to the same as the first block. Can we remove the first block in that case?

src/javascripts/modules/sortable-table.js
```diff
@@ -0,0 +1,147 @@
...
+      }
+
+      return result
+    })
+  }
+
+  static render (data) {
+    let html = ''
+    for (const record of data) {
+      let string = '<tr>'
```
teneightfive: Should we build this with `document.createElement` instead of using strings?

src/views/_components/data-table.njk
```diff
@@ -0,0 +1,68 @@
+{##
+ # @param {object[]]}     Data An array of objects to show in the table
+ # @param {string}            Variant A string or defining a style variation to apply to the table
+ #                            currently this only supports 'stripey'
+ # @param {object}  columns  	A key value pair object that lists the columns to show and their associated labels
+ #                            e.g { firstname: 'First name' }. If this isn't passed then all columns will display
+ #                            and use the keys as the headers
+ # @param {string} class
+ # @param {string} id
+ # @param {boolean} sortable default true
```
teneightfive: Should be more like this for optional default values: `@param {boolean} [sortable=true] Whether the table is sortable`

src/views/_components/data-table.njk
```diff
@@ -0,0 +1,68 @@
...
+  {% if not columns and data %}
+    {% set keys = data[0] | keys %}
+    {% set headings = data[0] | keys %}
+  {% else %}
+    {% set keys = columns | keys %}
+    {% set headings = columns | values %}
+  {% endif %}
+
+  <table
+    class="data-table{% if sortable %} is-sortable js-SortableTable{% endif %}{{ ' table--' + variant if variant }}{{ class if class }}"
```
teneightfive: This class doesn't match the hook in the JS module.

teneightfive: Does the class call need `if class`? If it has no value nunjucks won't render anything.

src/views/_components/data-table.njk
```diff
@@ -0,0 +1,68 @@
...
+    id="{{ id }}">
```
teneightfive: If dropping attributes onto new line then move the closing tag to a new line too.

src/javascripts/modules/sortable-table.js
ztolley: Building a string and then setting that as an innerHTML value is, in this case, simpler to read code and faster than lots of dom manupulation (even with fragments off the page)

src/views/_components/data-table.njk
```diff
@@ -0,0 +1,69 @@
+{##
+ # @param {object[]]}     Data An array of objects to show in the table
```
teneightfive: Got an extra `]` here

src/views/_components/data-table.njk
```diff
@@ -0,0 +1,69 @@
...
+ # @param {string}            Variant A string or defining a style variation to apply to the table
+ #                            currently this only supports 'stripey'
+ # @param {object}  columns  	A key value pair object that lists the columns to show and their associated labels
+ #                            e.g { firstname: 'First name' }. If this isn't passed then all columns will display
+ #                            and use the keys as the headers
+ # @param {string} class
```
teneightfive: Do we want to be consistent and add a description for these params?

test/components/data-table.test.js
```diff
@@ -0,0 +1,130 @@
...
+    const component = renderComponentToDom('data-table', { columns })
+
+    expect(component.className).to.contain('data-table')
+    const tableHeadings = component.querySelectorAll('thead tr th')
+
+    expect(tableHeadings.length).to.equal(2)
+    expect(tableHeadings[0].textContent).to.equal('Name')
+    expect(tableHeadings[1].textContent).to.equal('Address')
+  })
+  it('should render table rows', () => {
```
teneightfive: Do we want to give these blocks some white space when they're indented to the same level?

test/javascripts/sortable-table.test.js
```diff
@@ -0,0 +1,202 @@
+const jsdom = require('jsdom')
+const { JSDOM } = jsdom
+const SortableTable = require(`~/src/javascripts/modules/sortable-table`)
+const { domTokenToArray } = require('../component-helper')
```
teneightfive: Use wavy syntax for module path here too?

config/nunjucks/filters.js
```diff
@@ -1,6 +1,7 @@
 const nunjucks = require('nunjucks')
 const filters = {
   stringify: JSON.stringify,
+  values: Object.values,
```
feedmypixel: newline?

src/javascripts/app.js
```diff
@@ -1,3 +1,5 @@
 const ConditionalSubfields = require('./modules/conditional-subfields')
-
 ConditionalSubfields.init()
+
+const SortableTable = require('./modules/sortable-table')
```
feedmypixel: feels like consts at the top and inits below? or is that me being oldSkool?

src/javascripts/modules/sortable-table.js
```diff
@@ -0,0 +1,144 @@
...
+    this.document = document
+  }
+
+  cacheElements () {
+    this.headingElements = this.element.querySelectorAll('thead th')
+    this.tableBody = this.element.querySelector('tbody')
+  }
+
+  attachEvents () {
+    for (let pos = 0; pos < this.headingElements.length; pos += 1) {
```
feedmypixel: could be a `forEach`

src/javascripts/modules/sortable-table.js
feedmypixel: I'd vote for `document.createElement` html being built with strings feels very jQuery, maybe thats just me

src/javascripts/modules/sortable-table.js
```diff
@@ -0,0 +1,144 @@
...
+      let aValue = a[key]
+      let bValue = b[key]
+
+      if (isDate(aValue) && isDate(bValue)) {
+        aValue = convertToDate(aValue)
+        bValue = convertToDate(bValue)
+      }
+
+      let result = 0
+      if (aValue < bValue) result = -1
```
feedmypixel: not a fan of one line if conditionals

src/lib/date.js
```diff
@@ -1,20 +1,44 @@
...
+  if (!dateString || dateString.length === 0) {
     return null
   }
-  return moment(date).format(longDateFormat)
+  const date = parseDateString(dateString)
+  if (date) {
+    return format(date, longDateFormat)
+  }
+  return null
 }
```
feedmypixel: can this just be 
```
function formatLongDate (dateString=[]) {
 const date = parseDateString(dateString)

  if (date) {
    return format(date, longDateFormat)
  }

  return null
}
```

test/javascripts/archive-form.test.js
```diff
@@ -2,6 +2,7 @@
 const jsdom = require('jsdom')
 const { JSDOM } = jsdom
 const ArchiveForm = require('~/src/javascripts/archive-form')
+const { domTokenToArray } = require('../component-helper')
```
feedmypixel: you could use `~` here

src/javascripts/app.js
teneightfive: Yep, I agree, lets do that for consistency.

src/javascripts/modules/sortable-table.js
teneightfive: I'm kind of leaning towards that too.

src/lib/date.js
```diff
@@ -1,20 +1,44 @@
...
+    }
+  }
+  return null
 }
 
 module.exports = {
-  formatLongDate, formatMediumDate
+  formatLongDate,
+  formatMediumDate,
+  parseDateString
```
teneightfive: Missing trailing comma. Shouldn't the linter now pick this up?

ztolley: it was done before the linter changes merged

src/javascripts/modules/sortable-table.js
ztolley: fine

config/nunjucks/filters.js
```diff
@@ -1,6 +1,8 @@
 const nunjucks = require('nunjucks')
+
 const filters = {
   stringify: JSON.stringify,
+  values: Object.values,
```
teneightfive: I think @feedmypixel meant a space between each filter so above this line. But I also like the space after the require so keep that :)

src/views/_components/data-table.njk
```diff
@@ -0,0 +1,70 @@
+{##
+ # @param {object[]}     Data An array of objects to show in the table
+ # @param {string}            Variant A string or defining a style variation to apply to the table
+ #                            currently this only supports 'stripey'
+ # @param {object}  columns  	A key value pair object that lists the columns to show and their associated labels
+ #                            e.g { firstname: 'First name' }. If this isn't passed then all columns will display
+ #                            and use the keys as the headers
+ # @param {string} class      A css classname to add to the table element of the control
+ # @param {string} id         A value to assign to the table id attribute
+ # @param {boolean} [sortable=true]  Should the table apply the 'sortable-table' decorator, allowing users to
```
teneightfive: You can drop the `Default is set to true` as this is described in JSDoc style: `[sortable=true]`

src/views/_components/data-table.njk
```diff
@@ -0,0 +1,70 @@
...
+ #                                   sort row order in real time. Default is set to true
+
+ # @return {string} HTML for sortable table
+ #
+ # @example
+ #  {% component 'sortable-table',
```
teneightfive: This doesn't look like the right syntax for calling a component. This would be a better method of calling this component with this data:

```njk
# @example
#  {% component 'sortable-table', {
#     variant: 'stripey',
#     id: 'example-table',
#     columns: {
#       name: 'Name',
#       addr: 'Address'
#     },
#     data: [{
#       id: '134',
#       name: 'Fred Bloggs',
#       addr: '10 Fred Road, Humbleton, Berks, SL5 5RR'
#     }, {
#       id: '4444',
#       name: 'Jane Smith',
#       addr: '32 Palace Place, London, SE1 1TR'
#     }]
#  %}
```

webpack.config.js
teneightfive: We don't need this anymore now that it's being including the global `app.js` file.

src/views/_components/data-table.njk
ztolley: that looks right, will update and correct

PR #148, opened 2017-06-01
Add adviser search to investment creation and edit

teneightfive:
The ability to search and add an adviser was missing. This adds
that ability and the ability to set the value again when
editing.

This code may change when the journey for creating and editing
an investment project gets split into multiple sections.

advisor-search

### DIFF
--- a/src/controllers/investment-create.controller.js
+++ b/src/controllers/investment-create.controller.js
@@ -8,13 +8,15 @@ const {
   updateInvestmentProject
 } = require('../repos/investment.repo')
 const { transformToApi, transformFromApi } = require('../services/investment-formatting.service')
+const { getAdvisors } = require('../repos/advisor.repo')
 const metadataRepo = require('../repos/metadata.repo')
 
 function getHandler (req, res, next) {
   const equityCompanyId = res.locals.equityCompanyId || req.query['equity-company']
   const promises = [
     getInflatedDitCompany(req.session.token, equityCompanyId),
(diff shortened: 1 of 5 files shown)


src/controllers/investment-create.controller.js
```diff
@@ -8,21 +8,23 @@ const {
...
+    getAdvisors(req.session.token)
   ]
 
   if (!equityCompanyId) {
     res.redirect('/investment/start')
   }
 
   Promise.all(promises)
-    .then(([equityCompany, equityCompanyInvestments]) => {
+    .then(([equityCompany, equityCompanyInvestments, advisorResponse]) => {
```
jim68000: Just as an FYI, we're changing the spelling throughout the site from Advisor to Adviser (both are valid, but Advisor is more usually American)

teneightfive: Yes. Good point, will update throughout.

teneightfive: Actually...have just had a quick look at the amount of occurrences of `advisor` and it's around 285 in 40 files. I think we should do that as a separate change and PR which should be easier to see what's changed.

It could confuse what's going on in this change. Thoughts?

tyom: Definitely a separate PR.

feedmypixel: separate PR defo

teneightfive: I'll add it to this one then :trollface:

PR #149, opened 2017-06-01
Simplify investment project transforms

teneightfive:
Make the code that transform data for investment projects to and from
the API easier to understand and check not fall over when arrays
are empty.


src/controllers/investment-create.controller.js
```diff
@@ -92,6 +92,7 @@ function editMiddleware (req, res, next) {
       res.locals.equityCompanyId = projectData.investor_company.id
       res.locals.projectId = projectData.id
 
+      // return res.send(transformFromApi(projectData))
```
tyom: 🚨

PR #150, opened 2017-06-02
Use wavy for local require

tyom:
We set project root in test globals to be able to require local modules without
having to specify relative paths using 'require(`${root}/module`)'.
Unfortunately it breaks IDEs 'navigate to declaration' functionality.

wavy is a small package that symlinks module named ~ to current project directory which
allows to do the same using `require('~/module')` anywhere in the project while
preserving IDEs ability to resolve modules and navigate to its declaration.


feedmypixel:
yeah this is really nice

Issue #78, opened 2017-05-17
Centralised router config


teneightfive:
## Problem

We currently refer to routes directly as strings when calling `res.redirect`. If routes change this will mean we'll have to make sure all occurrences of the original path will need to be changed or we'll end up with broken redirects.

## Suggestion

Either create a centralised router object or expose a `url` property on each controller which can be referenced when calling `res.redirect` or in tests.

## Further suggestion

It might also be cleaner to have all routes defined in the same place rather than within each controller so it's easier to see all available routes in one place. Controllers could then expose just their methods instead of the router.

feedmypixel:
`Jira > DH-244` also contains info around this need

rafenden:
`urls` module was introduced to address this issue.

PR #151, opened 2017-06-02
Move investment controllers under sub-app

tyom:
Create investment sub-app to hold common data and include sub routes.


src/controllers/investment/edit.controller.js
```diff
@@ -0,0 +1,59 @@
+const router = require('express').Router()
+
+const { getInvestmentProjectSummary } = require('../../repos/investment.repo')
+const { transformFromApi, formatProjectStatusData } = require('../../services/investment-formatting.service')
+const { populateFormMiddleware, postFormMiddleware } = require('./form')
```
jim68000: Is there a rule on when to split destructures like this into separate lines? More than 2?

src/controllers/investment/index.js
```diff
@@ -0,0 +1,33 @@
+const router = require('express').Router()
+
+function getLocalNav (req, res, next) {
+  res.locals.localNavItems = [
+    { label: 'Project details', slug: 'details' },
```
jim68000: Wondering if we shouldn't keep the tabs here commented out until we have the associated pages sorted.

src/controllers/investment/index.js
```diff
@@ -0,0 +1,33 @@
...
+    { label: 'Client', slug: 'client' },
+    { label: 'Project team', slug: 'team' },
+    { label: 'Interactions', slug: 'interactions' },
+    { label: 'Documents', slug: 'documents' },
```
jim68000: This one has no clear set of requirements, and may not be required, so should be removed

src/controllers/investment/index.js
```diff
@@ -0,0 +1,33 @@
+const router = require('express').Router()
+
+function getLocalNav (req, res, next) {
+  res.locals.localNavItems = [
+    { label: 'Project details', slug: 'details' },
+    { label: 'Client', slug: 'client' },
```
jim68000: This currently just replicates part of the details screen. Check with Vince but I (and Dom) propose removing it

src/lib/controller-utils.js
```diff
@@ -135,13 +135,18 @@ function containsFormData (req) {
   return (typeof req.body === 'object' && Object.keys(req.body).length > 0)
 }
 
+function isValidGuid (string) {
```
jim68000: I don't think this is going to be a problem for us, but this doesn't actually check for a correct GUID. Real GUIDs have the form

xxxxxxxx-xxxx-Mxxx-Nxxx-xxxxxxxxxxxx

Where M & N are versions of the GUID being used, so can't be 0-f. (the actually rules are bitmaps - see wiki page)

Last place I worked we managed to introduce a bug because the node.js side of the business was generating UUIDs randomly and the C# side expected a Microsoft GUID, so a proportion of attempts to submit a GUID failed.

src/controllers/investment/edit.controller.js
tyom: No hard rule. I'm inclined to say around 80 char length. I also tend to always split them at the end of the file for module exports.

src/lib/controller-utils.js
tyom: It's meant to loosely check GUID signature rather than validate its particular type. I use it to identify IDs from other slugs.

src/controllers/investment/index.js
tyom: The Project details page works here so I'm leaving it.

src/controllers/investment/details.controller.js
```diff
@@ -0,0 +1,28 @@
...
+const { getProjectDetails } = require('./projectDetails')
+const { isValidGuid } = require('../../lib/controller-utils')
+
+function getDetailsHandler (req, res) {
+  return res.render('investment/details', {
+    currentNavItem: 'details',
+  })
+}
+
+function redirectToDetails (req, res, next) {
```
ztolley: What is this purpose of this?

src/lib/controller-utils.js
ztolley: That gets around a BIG problem had with false positive matches on urls I ran into previous.

But I would say that can this be a short term fix and longer term we need to revisit how we build url's as @tyom and @teneightfive have already said.

package.json
```diff
@@ -63,7 +63,7 @@
     "clean": "rm -rf build",
     "start": "node --use-strict src/server.js",
     "build": "npm run clean && npm run sass:compile && npm run js",
-    "test": "mocha --recursive --require test/common.js --ui bdd test",
+    "test": "mocha",
```
feedmypixel: 👍

src/controllers/investment/create.controller.js
```diff
@@ -0,0 +1,32 @@
...
+  }
+
+  return res.render('investment/create')
+}
+
+function postHandler (req, res) {
+  res.locals.localNavItems = null
+
+  if (res.locals.form.errors) {
+    res.render('investment/create')
```
feedmypixel: best practice for returns on these?

src/services/investment-formatting.service.js
```diff
@@ -1,60 +1,6 @@
-const { formatLongDate } = require('../lib/date')
+const moment = require('moment')
```
ztolley: In the data table branch I've switched the date library to use the date fns library to be lighter than moment.

Do we want to keep the formatLongDate function or switch this code to date-=I fns and remove formatLongDate all together?

Not all now, but lets decide going forward.

src/controllers/investment/details.controller.js
tyom: It redirects to `/details` if ID checks out.

src/services/investment-formatting.service.js
tyom: Since your branch is not merged yet I'd say we do another pass at this and remove moment completely. What do you say?

src/controllers/investment/form.js
```diff
@@ -0,0 +1,98 @@
...
+    res.redirect('/investment/start')
+  }
+
+  const promises = [
+    getEquityCompanyDetails(req.session.token, equityCompanyId),
+    getAdvisors(req.session.token),
+  ]
+
+  Promise.all(promises)
+    .then(([ { equityCompany, equityCompanyInvestment }, advisorsResponse ]) => {
```
feedmypixel: would this be easier to read on separate lines?

src/controllers/investment/index.js
```diff
@@ -0,0 +1,31 @@
...
+    return res.redirect(`/investment/start`)
+  }
+  next()
+}
+
+router.use(
+  '/investment',
+  handleEmpty,
+  getLocalNav,
+  require('./start.controller').router,
```
feedmypixel: doesn't feel like I would be expecting requires down here

src/controllers/investment/projectDetails.js
```diff
@@ -0,0 +1,45 @@
...
+      const valueData = yield getInvestmentValue(req.session.token, req.params.id)
+      const requirementsData = yield getInvestmentRequirements(req.session.token, req.params.id)
+
+      res.locals.investmentProject = {
+        meta: formatProjectStatusData(projectData),
+        details: formatProjectData(projectData),
+        value: formatValueData(valueData),
+        requirements: formatRequirementsData(requirementsData),
+      }
+      res.locals.equityCompanyId = projectData.investor_company
```
feedmypixel: newline?

src/controllers/investment/index.js
tyom: We selectively use this pattern in the main app. Should we import those controllers first?

src/sass/components/_local-nav.scss
```diff
@@ -34,4 +34,11 @@
     text-decoration: none;
     font-weight: 600;
   }
+
+  &--background {
```
feedmypixel: I realise this was here before but I feel using the `parent` sass selector like this:
- introduces cognitive load when reading the sass
- makes searching for classNames in sass harder

Personally I keep them for psuedo classes only
Thoughts?

tyom: I'm OK with defining modifiers like this as they always attached to a block or element. So once you find the main block/element it's a matter of scanning the rule set to see any modifiers.

I would only reserve this approach for small rule sets. Bigger chunk would go under their own declaration.

@teneightfive Thoughts?

src/views/investment/start.njk
```diff
@@ -141,5 +141,12 @@
       {{ trade.save(buttonText="Continue") }}
     </form>
 
+  {% else %}
+
+    <p class="flash flash--info">
+      Client company is required
```
feedmypixel: should this be hradcoded?

tyom: Well, this is a temporary measure to avoid confusing user if id is missing. In future I would like to show company search if client company id is missing.

src/controllers/investment/index.js
teneightfive: Yes, I was going to say the same. We should be consistent with the use of requires and import libraries at the top of the file first.

src/controllers/investment/index.js
```diff
@@ -0,0 +1,35 @@
...
+  require('./create.controller').router,
+  require('./details.controller').router,
+  require('./edit.controller').router,
```
teneightfive: This trailing comma breaks the app. 🚨

tyom: Doesn't break in Node 8 😛

src/controllers/investment/create.controller.js
```diff
@@ -0,0 +1,31 @@
+const router = require('express').Router()
+
+const { populateFormMiddleware, postFormMiddleware } = require('./form')
+
+function getHandler (req, res) {
+  res.locals.localNavItems = null
```
teneightfive: Not sure I like that we have to set this value to null in all the views that don't use the left nav.

Would it not be better to override the block in the view or use a different layout for views that require the left nav?

src/controllers/investment/details.controller.js
```diff
@@ -0,0 +1,28 @@
+const router = require('express').Router()
+
+const { getProjectDetails } = require('./projectDetails')
+const { isValidGuid } = require('../../lib/controller-utils')
+
+function getDetailsHandler (req, res) {
+  return res.render('investment/details', {
+    currentNavItem: 'details',
```
teneightfive: Can we do something more globally for navigation? Possibly pass the current path in to a nunjucks global and check that value against the url used for a navigation item?

teneightfive: Could be a separate change though.

src/controllers/investment/details.controller.js
teneightfive: Shouldn't we check if it's valid in the param middleware? Then this can be a straight redirect?

src/controllers/investment/edit.controller.js
teneightfive: Wonder if there's a better name/pattern for the `./form` file. `shared.js`? `investment.shared.js`?

src/controllers/investment/details.controller.js
```diff
@@ -0,0 +1,28 @@
+const router = require('express').Router()
+
+const { getProjectDetails } = require('./projectDetails')
```
teneightfive: The file name for `projectDetails.js` isn't consistent with the rest of the app. We use kebab case elsewhere so should stick to that.

src/controllers/investment/form.js
```diff
@@ -0,0 +1,104 @@
...
+    getAdvisors(req.session.token),
+  ]
+
+  Promise.all(promises)
+    .then(([
+      {
+        equityCompany,
+        equityCompanyInvestment,
+      },
+        advisorsResponse,
```
teneightfive: The indentation seems odd here. Why are `equityCompany` and `equityCompanyInvestment` also in an object?

src/controllers/investment/form.js
```diff
@@ -0,0 +1,104 @@
...
+      next()
+    })
+}
+
+function postFormMiddleware (req, res, next) {
+  const formattedBody = transformToApi(req.body)
+  const projectId = res.locals.projectId || req.params.id
+  let saveMethod
+
+  res.locals.form = res.locals.form || {}
```
teneightfive: Does this need to be defined here? If it's only being used within the `catch()` can't it be defined within that?

src/sass/components/_local-nav.scss
teneightfive: I'd lean towards @feedmypixel's side. I personally don't like using parent selector for modifiers for the exact reasons he mentioned. 

I don't see the benefit you get from using them. I find it quicker to read when you have the block alongside the modifier in the declaration and I don't think we want to hinder people searching on a modifier class in a codebase if we can.

src/services/investment-formatting.service.js
teneightfive: @ztolley your switch was for the frontend though where we are more concerned with file size from dependencies. As this is running on the server we can be less concerned about the size of that dependency, no?

src/views/_components/form-multiple-choice.njk
```diff
@@ -31,7 +31,7 @@
 
 {% set hintId  = 'hint-' + name %}
 {% set variant = variant | default('radio') %}
-{% set value   = value | default('') | string %}
+{% set value   = value | default('', true) | string %}
```
teneightfive: Would this not set a value that had `true` to checked when we didn't want it?

src/controllers/investment/details.controller.js
tyom: I had an idea of a more generic current nav item detection but then I realised I was going down the rabbit hole and decided to drop out in favour of doing smaller incremental refactorings.

src/views/_components/form-multiple-choice.njk
ztolley: Yes size is not so important, but as we found with my PR, a piece of code that was being used for backend crept into front end, partially due to my assumption that tree shaking was setup

src/controllers/investment/details.controller.js
teneightfive: Yep. Agreed. Let's do that later on.

I do think we need to look at the other change re nav though - not having to override it in views that don't review it. We should do that at the template level.

src/controllers/investment/create.controller.js
tyom: We only need to set it in controllers, since that's where menus are defined.

We could override them in views but then it might be weird to have that data in view floating around as it's not relevant.

src/controllers/investment/form.js
tyom: This is defined in `investment.repo` and because it's a generic function that is expected to be used on its own I felt it would be weird to return an array from promise response.

src/controllers/investment/details.controller.js
tyom: `:id` would still need to be defined on the route somewhere so not much point.

src/controllers/investment/edit.controller.js
tyom: Maybe. This is meant to be shared for routes concerned with forms though.

src/services/investment-formatting.service.js
tyom: I still think we should move to to fns-date over Moment. I think it can do everything moment can in terms of functionality but it's modular, more performant and is immutable. AFAIK the only thing it doesn't have is an alternative to Moment timezone.

Why would we want to depend on two competing libraries?

src/views/_components/form-multiple-choice.njk
tyom: @ztolley Is this meant to be a comment about Moment?

ztolley: I couldn't comment on the comment about moment (github) so this is as close as I could get

src/controllers/investment/create.controller.js
```diff
@@ -0,0 +1,27 @@
...
+
+function getHandler (req, res) {
+  if (!res.locals.equityCompany) {
+    return res.redirect('/investment/start')
+  }
+
+  return res.render('investment/create')
+}
+
+function postHandler (req, res) {
```
teneightfive: This is doing exactly the same as the edit controller post handler. Do we not want to combine this or find a way to reuse the code?

src/controllers/investment/details.controller.js
```diff
@@ -0,0 +1,28 @@
...
+    currentNavItem: 'details',
+  })
+}
+
+function redirectToDetails (req, res, next) {
+  const id = req.params.id
+  if (isValidGuid(id)) {
+    return res.redirect(`/investment/${id}/details`)
+  }
+  return next()
```
teneightfive: Do we want to pass an error to next here saying that it's not a valid ID?

src/controllers/investment/index.js
```diff
@@ -0,0 +1,40 @@
+const router = require('express').Router()
+
+const startController = require('./start.controller')
+const createController = require('./create.controller')
+const detailsController = require('./details.controller')
+const editController = require('./edit.controller')
+
+function getLocalNav (req, res, next) {
```
teneightfive: Have started to think about how we name variables and might be useful to use `getLocalNavMiddleware` in variable names to be clear about what things are.

@feedmypixel thoughts? (I noticed you being specific in tests and found it quite useful when reading the code)

src/controllers/investment/index.js
```diff
@@ -0,0 +1,40 @@
...
+  getLocalNav,
+  startController.router,
+  createController.router,
+  detailsController.router,
+  editController.router
+)
+
+module.exports = {
+  router,
+  getLocalNav,
```
teneightfive: Wonder if it's better to abstract these into a middleware module so they're easier to test and don't have to be exported on this module?

src/controllers/investment/shared.js
```diff
@@ -0,0 +1,144 @@
...
+const metadataRepo = require('../../repos/metadata.repo')
+const {
+  createInvestmentProject,
+  updateInvestmentProject,
+  getEquityCompanyDetails,
+  getInvestmentProjectSummary,
+  getInvestmentValue,
+  getInvestmentRequirements,
+} = require('../../repos/investment.repo')
+
```
teneightfive: Do we need this new line?

src/controllers/investment/shared.js
```diff
@@ -0,0 +1,144 @@
...
+  ]
+
+  Promise.all(promises)
+    .then(([
+      {
+        equityCompany,
+        equityCompanyInvestment,
+      },
+      advisersResponse,
+     ]) => {
```
teneightfive: Indentation is still incorrect here. Has an extra space.

src/controllers/investment/start.controller.js
```diff
@@ -1,15 +1,17 @@
...
+const { getCompanyInvestmentProjects } = require('../../repos/investment.repo')
+const { searchForeignCompany } = require('../../services/search.service')
+const { getPagination } = require('../../lib/pagination')
 
 function getHandler (req, res, next) {
   const clientCompanyId = req.query['client-company']
   const searchTerm = req.query['q']
   const promises = []
 
+  res.locals.localNavItems = null
```
teneightfive: Did we not find a way to not require settings this back to null to hide the navigation block?

src/sass/components/_local-nav.scss
```diff
@@ -35,3 +35,11 @@
     font-weight: 600;
   }
 }
+
+.local-nav__link--background {
```
teneightfive: Why do we need this extra state? I thought at the moment we just remove the left nav when editing?

src/controllers/investment/create.controller.js
teneightfive: I wouldn't say it's weird to use a different layout for create/edit that doesn't include the left nav at this stage.

It feels stranger to set it in middleware that all routes inherit but then set it to null in controllers that don't require that piece of layout.

Layouts choices should be defined in views and based on what layout they inherit from in my opinion.

src/controllers/investment/shared.js
```diff
@@ -0,0 +1,144 @@
...
+    saveMethod = createInvestmentProject(req.session.token, formattedBody)
+  }
+
+  saveMethod
+    .then((result) => {
+      res.locals.resultId = result.id
+      next()
+    })
+    .catch((err) => {
+      res.locals.form = res.locals.form || {}
```
teneightfive: This could even go one block lower, no?

It's only being reference inside the if statement.

src/sass/components/_local-nav.scss
tyom: The designs suggest that the navigation persists and changes state.

src/controllers/investment/create.controller.js
tyom: I think these handlers are so small that it makes sense to keep them individual to each controller rather than cross import from other controllers.

Also the edit handler is slightly different as it's currently setting navigation state in locals.

src/controllers/investment/details.controller.js
tyom: We throw an error but it will cause a `500` rather than more appropriate `404` if it follows the middleware chain.

src/controllers/investment/shared.js
tyom: I think it helps to read it. In fact I'd add one to the destructuring above.

src/controllers/investment/index.js
tyom: We have `shared.js` I suppose.

src/controllers/investment/details.controller.js
teneightfive: Have you tried that route? 

[http://localhost:3000/investment/736ad86f-da29-4715-a213cec6ff6/details](http://localhost:3000/investment/736ad86f-da29-4715-a213cec6ff6/details) still loads that view but with empty details.

Might need to check this is performing correctly.

teneightfive: I think calling `next()` just passes it on to the next middleware which is the get handler, and not right through to the `404` handler.

src/controllers/investment/index.js
teneightfive: Maybe park this for a later change. We've also got an issue (#78) to look at having a centralised router which this could cross over with.

src/controllers/investment/details.controller.js
tyom: Which is fine because there may be other middleware which it may need to pass, such as a logger.

PR #153, opened 2017-06-02
Lint object curly spacing

tyom:
So we never have to mention it in PRs again.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -43,7 +43,8 @@ module.exports = {
     new webpack.DefinePlugin({
       'process.env': {
         'NODE_ENV': JSON.stringify('production')
-      }}),
+      }
+    }),
     new webpack.optimize.UglifyJsPlugin({
       compress: {
         warnings: false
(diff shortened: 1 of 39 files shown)


teneightfive:
👍

webpack.config.js
```diff
@@ -43,7 +43,7 @@ module.exports = {
     new webpack.DefinePlugin({
       'process.env': {
         'NODE_ENV': JSON.stringify('production')
-      }}),
+      } }),
```
teneightfive: This one actually feels like it should be on the next line down.

tyom: Just that one? It's auto`--fix`ed. But I'll fixed it for you. ✨

teneightfive: The linter probably wouldn't spot that though. Just the lack of space.

It should be this though:
```
new webpack.DefinePlugin({
  'process.env': {
    'NODE_ENV': JSON.stringify('production')
  }
}),
```

tyom: Thanks Dom ✋

PR #154, opened 2017-06-02
Investment audit.

ztolley:
The work for this is done but can't be merged yet, but feel free to start reviewing.

Produces a lined list of items on the screen, listing the person who made a change, how many items they changed and when.

### DIFF
--- a/src/controllers/investment/audit.controller.js
+++ b/src/controllers/investment/audit.controller.js
@@ -1,0 +1,44 @@
+const router = require('express').Router()
+const Q = require('q')
+const { get } = require('lodash')
+
+const { getProjectDetails } = require('./shared')
+const { formatLongDate } = require('../../lib/date')
+const { getInvestmentProjectAuditLog } = require('../../repos/investment.repo')
+
+function formatAuditLog (logEntry) {
+  return {
+    name: logEntry.user.name,
+    timestamp: formatLongDate(logEntry.timestamp),
+    changes: (logEntry.changes && Object.keys(logEntry.changes).length) || 0,
+  }
+}
+
+function getInvestmentAudit (req, res, next) {
+  Q.spawn(function * () {
+    try {
+      if (get(res, 'locals.investmentProject')) {
+        const rawAuditLog = yield getInvestmentProjectAuditLog(req.session.token, req.params.id)
+        const auditLog = rawAuditLog.map(formatAuditLog)
+
+        return res.render('investment/audit', {
+          currentNavItem: 'audit',
+          auditLog,
+        })
+      }
+
+      return next()
+    } catch (error) {
+      next(error)
+    }
+  })
+}
+
+router.param('id', getProjectDetails)
+router.get('/:id/audit', getInvestmentAudit)
+
+module.exports = {
+  router,
+  getInvestmentAudit,
+  formatAuditLog,
+}
--- a/src/controllers/investment/index.js
+++ b/src/controllers/investment/index.js
@@ -5,6 +5,7 @@ const startController = require('./start.controller')
 const createController = require('./create.controller')
 const detailsController = require('./details.controller')
 const editController = require('./edit.controller')
+const auditController = require('./audit.controller')
 
 function handleEmptyMiddleware (req, res, next) {
   if (req.path === '/') {
@@ -20,7 +21,8 @@ router.use(
   startController.router,
   createController.router,
   detailsController.router,
-  editController.router
+  editController.router,
+  auditController.router
 )
 
 module.exports = {
--- a/src/repos/investment.repo.js
+++ b/src/repos/investment.repo.js
@@ -49,6 +49,13 @@ function updateInvestmentProject (token, investmentId, body) {
   })
 }
 
+function getInvestmentProjectAuditLog (token, investmentId) {
+  return authorisedRequest(token, `${config.apiRoot}/v3/investment/${investmentId}/audit`)
+    .then((data) => {
+      return data.results
+    })
+}
+
 module.exports = {
   getCompanyInvestmentProjects,
   getInvestmentProjectSummary,
@@ -57,4 +64,5 @@ module.exports = {
   getEquityCompanyDetails,
   createInvestmentProject,
   updateInvestmentProject,
+  getInvestmentProjectAuditLog,
 }
--- a/src/sass/objects/_list.scss
+++ b/src/sass/objects/_list.scss
@@ -1,3 +1,6 @@
+@import "../settings/colours";
+@import "../settings/layout";
+
 .list-unstyled,
 .list-inline {
   list-style: none;
@@ -8,4 +11,13 @@
 .list-inline li {
   margin-right: 1em;
   display: inline-block;
+}
+
+.list-lined li {
+  border-bottom: 1px solid $border-colour;
+  padding: $gutter-half $gutter-half $gutter-half 0;
+
+  &:first-child {
+    border-top: 1px solid $border-colour;
+  }
 }
--- a/src/views/investment/audit.njk
+++ b/src/views/investment/audit.njk
@@ -1,0 +1,13 @@
+{% extends "investment/_layout.njk" %}
+
+{% block main_grid_right_column %}
+  <h2 class="heading-medium">Audit history</h2>
+
+  <ul class="list-lined">
+    {% for logEntry in auditLog %}
+      <li>
+        User {{ logEntry.name }} changed {{ logEntry.changes }} {{ 'item' | pluralise(logEntry.changes) }} on {{ logEntry.timestamp }}
+      </li>
+    {% endfor %}
+  </ul>
+{% endblock %}
--- a/test/controllers/investment/audit.controller.test.js
+++ b/test/controllers/investment/audit.controller.test.js
@@ -1,0 +1,212 @@
+const investmentProjectAuditData = require('~/test/data/investment/audit-log.json')
+
+const token = 'abcd'
+
+describe('Investment audit controller', () => {
+  beforeEach(() => {
+    this.sandbox = sinon.sandbox.create()
+    this.next = this.sandbox.stub()
+    this.getInvestmentProjectAuditLog = this.sandbox.stub().resolves(investmentProjectAuditData.results)
+
+    this.controller = proxyquire('~/src/controllers/investment/audit.controller', {
+      '../../repos/investment.repo': {
+        getInvestmentProjectAuditLog: this.getInvestmentProjectAuditLog,
+      },
+    })
+  })
+
+  it('should get the audit log from the API', (done) => {
+    this.controller.getInvestmentAudit({
+      session: {
+        token,
+      },
+      params: {
+        id: '9999',
+      },
+    }, {
+      locals: {
+        investmentProject: {},
+      },
+      render: (template, data) => {
+        try {
+          expect(this.getInvestmentProjectAuditLog).to.be.calledWith(token, '9999')
+          done()
+        } catch (error) {
+          done(error)
+        }
+      },
+    }, this.next)
+  })
+  it('should send a parsed copy of audit data to the view', (done) => {
+    const expected = [{
+      name: 'Duke Ellington',
+      timestamp: '2 June 2017',
+      changes: 3,
+    }, {
+      name: 'Fred Smith',
+      timestamp: '2 June 2017',
+      changes: 5,
+    }]
+
+    this.controller.getInvestmentAudit({
+      session: {
+        token,
+      },
+      params: {
+        id: '9999',
+      },
+    }, {
+      locals: {
+        investmentProject: {},
+      },
+      render: (template, data) => {
+        try {
+          expect(data.auditLog).to.deep.equal(expected)
+          done()
+        } catch (error) {
+          done(error)
+        }
+      },
+    }, this.next)
+  })
+  it('should handle audit entry containing an empty timestamp', (done) => {
+    const badDate = [{
+      user: {
+        id: '41212312312321',
+        name: 'Fred Smith',
+      },
+      timestamp: '',
+      comment: 'Optional changeset comment - we can stick anything here',
+      changes: {
+        fieldName1: [false, true],
+      },
+    }]
+
+    this.getInvestmentProjectAuditLog = this.sandbox.stub().resolves(badDate)
+
+    this.controller = proxyquire('~/src/controllers/investment/audit.controller', {
+      '../../repos/investment.repo': {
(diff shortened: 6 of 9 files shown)


src/server.js
```diff
@@ -136,6 +136,7 @@ app.use(require('./controllers/company-export.controller').router)
 app.use(investmentStartController.router)
 app.use(investmentCreateController.router)
 app.use(investmentDetailsController.router)
+app.use(require('./controllers/investment-audit.controller').router)
```
teneightfive: We should pick a consistent method for adding these routers. We should probably use the method of defining the module at the top and keeping `require()`s together and referencing it using the const in `app.use()`

src/views/investment/audit.njk
```diff
@@ -0,0 +1,12 @@
+{% extends "investment/_layout.njk" %}
+
+{% block main_grid_right_column %}
+  <h2 class="heading-medium">Audit history</h2>
+  <ul class="dh-lined-list">
```
teneightfive: Breathing space

test/controllers/investment-audit.controller.test.js
```diff
@@ -0,0 +1,197 @@
+const investmentProjectAuditData = require('../data/investment/audit-log.json')
+
+const token = 'abcd'
+
+describe('Investment audit controller', () => {
+  beforeEach(() => {
+    this.sandbox = sinon.sandbox.create()
+    this.next = function (error) { throw new Error(error) }
```
teneightfive: Might be better to just use:
```js
this.next = this.sandbox.stub()
```

And then check what that stub gets called with.

test/controllers/investment-audit.controller.test.js
```diff
@@ -0,0 +1,197 @@
...
+        try {
+          expect(data.auditLog).to.deep.equal(expected)
+          done()
+        } catch (error) {
+          done(error)
+        }
+      },
+    }, this.next)
+  })
+  it('should handle when provided with an invalid date', (done) => {
```
teneightfive: Handle what?

test/repos/investment.repo.test.js
```diff
@@ -87,4 +89,17 @@ describe('Investment repository', () => {
       return expect(actual).to.eventually.deep.equal(investmentProjectSummaryData)
     })
   })
+
+  describe('#getInvestmentProjectAudit', () => {
+    // /v3/investment/${investmentProjectSummaryData.id}/audit
```
teneightfive: Do we need this comment?

test/controllers/investment-audit.controller.test.js
ztolley: we done see errors though when we do this

src/views/investment/audit.njk
ztolley: I like to call it efficient

test/controllers/investment-audit.controller.test.js
teneightfive: Why do we need to see errors though? The unit test should just test that an error was thrown?

src/sass/objects/_list.scss
```diff
@@ -9,3 +9,18 @@
   margin-right: 1em;
   display: inline-block;
 }
+
+.dh-lined-list {
```
teneightfive: Have we started to introduce this prefix now then? Seems like the rest of the files doesn't use it.

We should also use the object name first like above. So would be `list-lined` in this instance.

I also wonder if it's a modifier rather than a separate block too?

@feedmypixel @tyom thoughts?

ztolley: No one else has but I figured we had to start somewhere

ztolley: Also there is a lined list already have a lined list but the styles aren't fully bem so they would clash and break so the namespacing fixes it

src/controllers/investment-audit.controller.js
```diff
@@ -0,0 +1,40 @@
...
+const { formatLongDate } = require('../lib/date')
+const { getInvestmentProjectAuditLog } = require('../repos/investment.repo')
+
+const auditLogColumns = {
+  date: 'Date',
+  name: 'Name',
+  count: 'Item change count',
+}
+
+function formatAuditLog (logEntry) {
```
jim68000: Irritating thing Dariusz mentioned this morning: the audit endpoint looks at all changes to the table. Of course, the last_modified timestamp is part of the InvestmentProject table, so the change count will always be 1 more than required...

src/controllers/investment/audit.controller.js
```diff
@@ -0,0 +1,40 @@
...
+        auditLog,
+        auditLogColumns,
+      })
+    } catch (error) {
+      next(error)
+    }
+  })
+}
+
+router.get('/investment/:id/audit', getInvestmentAudit)
```
tyom: Perhaps you could rename it to `getInvestmentAudit` handler, or just `getHandler` to be consistent with the rest of investment controllers.

src/controllers/investment/audit.controller.js
```diff
@@ -0,0 +1,40 @@
...
+    name: logEntry.user.name,
+    timestamp: formatLongDate(logEntry.timestamp),
+    changes: (logEntry.changes && typeof logEntry.changes === 'object') ? Object.keys(logEntry.changes).length : 0,
+  }
+}
+
+function getInvestmentAudit (req, res, next) {
+  Q.spawn(function * () {
+    try {
+      const rawAuditLog = yield getInvestmentProjectAuditLog(req.session.token, req.params.id)
```
tyom: I would suggest extracting this logic in its own function so that this handler is only concerned with view rendering.

src/repos/investment.repo.js
```diff
@@ -49,6 +49,13 @@ function updateInvestmentProject (token, investmentId, body) {
   })
 }
 
+function getInvestmentProjectAuditLog (token, investmentId) {
+  return authorisedRequest(token, `${config.apiRoot}/v3/investment/${investmentId}/audit`)
+  .then((data) => {
```
tyom: Should be indented.

src/sass/objects/_list.scss
```diff
@@ -9,3 +9,19 @@
   margin-right: 1em;
   display: inline-block;
 }
+
+
+.dh-list {
+  list-style: none;
```
tyom: There's `list-unstyled` a the top of this file that does exactly this.

src/sass/objects/_list.scss
```diff
@@ -9,3 +9,19 @@
...
+  padding: 0;
+  margin: 0;
+}
+
+.dh-list--lined .dh-list__item {
```
tyom: Have we decided to prefix our objects/components? We should be consistent.

tyom: I think we do need to start prefixing our components to avoid clashes with existing Trade Elements. We need to agree on prefixes. @teneightfive @feedmypixel

src/controllers/investment/audit.controller.js
```diff
@@ -0,0 +1,47 @@
...
+    } catch (error) {
+      next(error)
+    }
+  })
+}
+
+router.param('id', getProjectDetails)
+router.get('/:id/audit', getInvestmentAudit)
+
+module.exports = { router, getInvestmentAudit, formatAuditLog }
```
teneightfive: We should use a new line pattern now to reduce diff noise in the future.

tyom:
Could you clean up your history please.

src/controllers/investment/index.js
```diff
@@ -20,7 +21,8 @@ router.use(
   startController.router,
   createController.router,
   detailsController.router,
-  editController.router
+  editController.router,
+  auditController.router
```
teneightfive: Missing trailing comma. Has this not been rebased on develop and being picked up when committing by the commit hooks?

src/sass/objects/_list.scss
```diff
@@ -9,3 +9,19 @@
...
+  border-bottom: 1px solid $border-colour;
```
teneightfive: Does the list need a top border to surround the element?

teneightfive: Should also declare dependencies if we're now using colours and gutters.

src/sass/objects/_list.scss
ztolley: I think we need to phase that old style out

src/controllers/investment/index.js
ztolley: Wasn't picked up. maybe interactive rebase bypasses this?

src/controllers/investment/index.js
```diff
@@ -20,7 +21,8 @@ router.use(
   startController.router,
   createController.router,
   detailsController.router,
-  editController.router
+  editController.router,
```
teneightfive: Interesting that CI didn't pick this up with @tyom's branch actually...

src/sass/objects/_list.scss
```diff
@@ -9,3 +9,12 @@
   margin-right: 1em;
   display: inline-block;
 }
+
+.list--lined .list__item {
+  border-bottom: 1px solid $border-colour;
```
teneightfive: Need to add dependencies to top of file.

```scss
@import "../settings/colours";
@import "../settings/layout";
```

src/controllers/investment/index.js
ztolley: Can't comma dangle a function call

src/controllers/investment/index.js
ztolley: can't comma dangle function call

src/sass/objects/_list.scss
ztolley: done

src/controllers/investment/audit.controller.js
```diff
@@ -0,0 +1,44 @@
+const router = require('express').Router()
+const Q = require('q')
+const { get } = require('lodash')
+
+const { getProjectDetails } = require('./shared')
+const { formatLongDate } = require('../../lib/date')
+const { getInvestmentProjectAuditLog } = require('../../repos/investment.repo')
+
+function formatAuditLog (logEntry) {
```
teneightfive: We might want to abstract this out into the investment service later on as we're doing some formatting for details in there. Not for now though.

src/controllers/investment/index.js
teneightfive: Ah of course!

src/controllers/investment/index.js
teneightfive: Of course! My bad

teneightfive:
@ztolley make sure you rebase these changes before merging.

src/sass/objects/_list.scss
```diff
@@ -9,3 +12,12 @@
   margin-right: 1em;
   display: inline-block;
 }
+
+.list--lined .list__item {
```
ztolley: The top is the old way, the bottom the new way

src/controllers/investment/audit.controller.js
```diff
@@ -0,0 +1,44 @@
...
+  return {
+    name: logEntry.user.name,
+    timestamp: formatLongDate(logEntry.timestamp),
+    changes: (logEntry.changes && typeof logEntry.changes === 'object') ? Object.keys(logEntry.changes).length : 0,
```
tyom: I think you could make it more readable with Lodash's `get`:
```js
Object.keys(get(logEntry, 'changes', {})).length
```

src/controllers/investment/audit.controller.js
```diff
@@ -0,0 +1,44 @@
...
+  }
+}
+
+function getInvestmentAudit (req, res, next) {
+  Q.spawn(function * () {
+    try {
+      if (get(res, 'locals.investmentProject')) {
+        const rawAuditLog = yield getInvestmentProjectAuditLog(req.session.token, req.params.id)
```
tyom: Is it worth extracting this logic to somewhere more generic like repos?

src/sass/objects/_list.scss
tyom: Seems like objects should be simple enough to not require modifiers. Modifiers are typically associated with components (blocks).

ztolley: so back to .list-lined ?

src/controllers/investment/audit.controller.js
ztolley: doesn't work if the value of changes is null

src/controllers/investment/audit.controller.js
ztolley: ?

src/controllers/investment/audit.controller.js
feedmypixel: Do you need to cater for it being different types? Just make it only accept one thing. There is a lot of this around the codebase. IMO KISS and write code as if it is typed rather than catering for all situations.
`(logEntry.changes && Object.keys(logEntry.changes).length) || 0`

src/sass/objects/_list.scss
teneightfive: What about simply:

```scss
.list-lined li {
  border-bottom: 1px solid $border-colour;
  padding: $gutter-half $gutter-half $gutter-half 0;

  &:first-child {
    border-top: 1px solid $border-colour;
  }
}
```

It's consistent with what we have for now and can address objects later when we have more components.

feedmypixel: I usually write modifiers like:

```
.list--lined 
  .list__item {
  ...
  }
}

```

test/controllers/investment/audit.controller.test.js
```diff
@@ -0,0 +1,212 @@
...
+          done()
+        } catch (error) {
+          done(error)
+        }
+      },
+    }, this.next)
+  })
+  it('should send a parsed copy of audit data to the view', (done) => {
+    const expected = [{
+      name: 'Duke Ellington',
```
feedmypixel: 👍

src/controllers/investment/audit.controller.js
tyom: You could remove `null` values before doing that operation
```js
Object.keys(_.get(_.pick(logEntry), 'changes', {})).length
```
Which would make it a little bit more complex but I still think it's easier to read as it's declarative.

src/controllers/investment/audit.controller.js
tyom: Nevermind.

src/sass/objects/_list.scss
teneightfive: @feedmypixel I think the discussion is more around whether this fits the traditional component style of BEM as it's sat within _objects_ which could be treated as more high level.

It also feels like it could live better within _elements_ too.

PR #155, opened 2017-06-02
Patch/add search info to investment

feedmypixel:
- update add required arguments for `/v3/search`
- fix component name
- send in new arguments to component

### DIFF
--- a/src/controllers/investment-start.controller.js
+++ b/src/controllers/investment-start.controller.js
@@ -2,14 +2,13 @@ const router = require('express').Router()
 
 const { getInflatedDitCompany } = require('../services/company.service')
 const { getCompanyInvestmentProjects } = require('../repos/investment.repo')
-const { search } = require('../services/search.service')
+const { searchForeignCompany } = require('../services/search.service')
 const { getPagination } = require('../lib/pagination')
 
 function getHandler (req, res, next) {
   const clientCompanyId = req.query['client-company']
   const searchTerm = req.query['q']
   const promises = []
-  const searchType = 'company'
 
   if (clientCompanyId) {
     promises.push(getInflatedDitCompany(req.session.token, clientCompanyId))
@@ -20,12 +19,10 @@ function getHandler (req, res, next) {
   }
 
   if (searchTerm) {
-    promises.push(search({
+    promises.push(searchForeignCompany({
       token: req.session.token,
       page: req.query.page,
       searchTerm,
-      searchType,
-      uk_based: false
     }))
   } else {
     promises.push(Promise.resolve())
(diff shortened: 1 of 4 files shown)


feedmypixel:
To obtain foreign companies you need to `POST` the body param `uk_based:false`. The `searchService.search` needs some work to accept a method.

src/controllers/investment-start.controller.js
```diff
@@ -9,6 +9,7 @@ function getHandler (req, res, next) {
   const clientCompanyId = req.query['client-company']
   const searchTerm = req.query['q']
   const promises = []
+  const searchType = 'company'
```
teneightfive: Do we need to cache this variable actually? And doesn't search default to company anyway?

feedmypixel: nope old code - thats now been removed

src/services/search.service.js
```diff
@@ -25,6 +25,33 @@ function search ({ token, searchTerm, searchType, limit = 10, page = 1 }) {
...
 }
 
+// TODO discuss why this needs to be a POST rather than a GET like the default `/v3/search`
+// TODO this work has been abstracted for easy removal of code once this api call has has been changed to GET
+// TODO we have query params on a POST's url
+// TODO `uk_based` param is a string on the backend when it should be a boolean
+function searchForeignCompany ({ token, searchTerm, page = 1, limit = 10 }) {
+  const queryParams = {
+    offset: (page * limit) - limit,
+    limit
```
teneightfive: What's the deal with trailing commas now. I've started to add them but maybe we need a PR and rule change to enforce them if we would prefer that standard?

src/services/search.service.js
```diff
@@ -73,4 +100,4 @@ function searchLimited (token, term) {
   })
 }
 
-module.exports = { search, suggestCompany, searchLimited }
+module.exports = { search, searchForeignCompany, suggestCompany, searchLimited }
```
teneightfive: Multiple lines now?

feedmypixel:
@tyom @teneightfive I have updated this work to work with the `/v3/search/company` `post` endpoint. There are various things I have added `TODOs` for:

- discuss why this needs to be a POST rather than a GET like the default `/v3/search`
- this work has been abstracted for easy removal of code once this api call has has been changed to GET
- we have query params on a POST's url
- we have a param named `original_query` that is set to `term` on the backend, this needs aligning to other endpoints and changed to `term` in the request

Of not this work is also waiting for this PR to be merged https://github.com/uktrade/data-hub-leeloo/pull/237

feedmypixel:
I have also created `Jira > DH-293` around this work

PR #156, opened 2017-06-05
Rename advisor to adviser

teneightfive:
Although both are [accepted spellings](http://grammarist.com/spelling/adviser-advisor/) adviser is the primary spelling
therefore we should be using that spelling where necessary.

Advisor is more commonly used in the U.S. whereas adviser is the
preferred spelling in the U.K.


teneightfive:
@reupen are we planning to update the spellings in API fields?

reupen:
@teneightfive Yes, I started that a couple of weeks ago but renaming the actual Django model was proving difficult. Still, renaming it in all the endpoints should be fine. I can get started on that now so you can finish the renaming here?

Also the commit message has the renaming the the wrong way round which is a bit unfortunate :(

teneightfive:
Cool. `advisor` is also in a few property names, were you going to rename those too?]

Will edit the commit message - good spot!

reupen:
> Cool. advisor is also in a few property names, were you going to rename those too?]
Yep, URLs + request and response bodies.

PR #158, opened 2017-06-05
Remove codeclimate duplicates

ztolley:
Code climate was producing too many false positives


tyom:
Are you meant to be committing `package-lock.json` in this PR?

PR #159, opened 2017-06-05
Enforce comma dangle

teneightfive:
This turns the comma dangle rule on for eslint and fixes previous
occurrences that break the new rule.

The [reason](https://medium.com/@nikgraf/why-you-should-enforce-dangling-commas-for-multiline-statements-d034c98e36f8) we won't to enforce this rule is it means cleaner diffs
when adding to an array or object and easier code manipulation.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -2,7 +2,16 @@
   "extends": "standard",
   "parser": "babel-eslint",
   "rules": {
-    "comma-dangle": "off",
+    "comma-dangle": [
+      "error",
+      {
+        "arrays": "always-multiline",
+        "objects": "always-multiline",
+        "imports": "never",
+        "exports": "never",
+        "functions": "ignore"
+      }
+    ],
     "object-curly-spacing": ["error", "always"]
   }
 }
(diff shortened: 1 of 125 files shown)


.eslintrc
```diff
@@ -2,7 +2,16 @@
...
+      "error",
+      {
+        "arrays": "always-multiline",
+        "objects": "always-multiline",
+        "imports": "never",
+        "exports": "never",
+        "functions": "ignore"
+      }
+    ],
     "object-curly-spacing": ["error", "always"]
```
feedmypixel: comma needed? :trollface:

feedmypixel:
Personally I am fine either way. I have it inbuilt to not dangle my commas due to IE bugs over the years. But I do understand the world has moved on

teneightfive:
@feedmypixel Yep, we all remember IE8 joy but as long as we're passing our files through a minifier it should remove those trailing commas which means IE8 will be fine.

.eslintrc
tyom: Not in JSON.

PR #160, opened 2017-06-05
Feature/add default search nav aggregations

feedmypixel:
This work:
- adds default `result.aggregations` if they have not been provided by the api
- adds number formatting for the count results
- adds filter tests

searchtweaks

### DIFF
--- a/test/config/nunjucks/filters.test.js
+++ b/test/config/nunjucks/filters.test.js
@@ -1,21 +1,12 @@
 const filters = require('~/config/nunjucks/filters')
 
 describe('nunjucks filters', () => {
-  beforeEach(() => {
-    this.sandbox = sinon.sandbox.create()
-    this.filters = filters
-  })
-
-  afterEach(() => {
-    this.sandbox.restore()
-  })
-
   describe('#highlight', () => {
     it('should render string with highlight', () => {
       const searchTerm = 'example term'
       const mockString = `we should see ${searchTerm} highlighted here`
 
-      const highlightedString = this.filters.highlight(mockString, searchTerm)
+      const highlightedString = filters.highlight(mockString, searchTerm)
 
       expect(highlightedString.val).to.equal(`we should see <span class="results-highlight">${searchTerm}</span> highlighted here`)
     })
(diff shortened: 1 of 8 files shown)


test/config/nunjucks/filters.test.js
```diff
@@ -0,0 +1,102 @@
+
+describe('nunjucks filters', () => {
+  beforeEach(() => {
+    this.sandbox = sinon.sandbox.create()
+    this.filters = require('~/config/nunjucks/filters')
```
teneightfive: You probably don't need to require the filters file before every test if its not being modified. I tend to add it in the `beforeEach` hook if it gets changed during some tests.

You can just require the file once at the top and call the methods from the global const.

test/config/nunjucks/filters.test.js
```diff
@@ -0,0 +1,102 @@
...
+      const mockString = `we should see ${searchTerm} highlighted here`
+
+      const highlightedString = this.filters.highlight(mockString, searchTerm)
+
+      expect(highlightedString.val).to.equal(`we should see <span class="results-highlight">${searchTerm}</span> highlighted here`)
+    })
+
+    it('should render string without highlight', () => {
+      const searchTerm = 'example term'
+      const mockString = `we should not see another term highlighted here`
```
teneightfive: Doesn't need template strings here.

test/config/nunjucks/filters.test.js
```diff
@@ -0,0 +1,103 @@
+const filters = require('~/config/nunjucks/filters')
+
+describe('nunjucks filters', () => {
+  beforeEach(() => {
+    this.sandbox = sinon.sandbox.create()
+    this.filters = filters
```
teneightfive: You can also just do away with adding it to this context all together and call the methods using`filters.highlight()` instead of `this.filters.highlight()`

I tend to add it to the test context before every test only if I need to override it for certain tests. Adding it in the `beforeEach` will then make sure it's reset to that default state before each test. Normally handy if you're stubbing methods using `proxyquire`. 

In this instance you don't need any of that so can simply require at the top and call the methods normally.

PR #162, opened 2017-06-06
update the pluralise filters to use the new signature

feedmypixel:
update the pluralise filters to use the new signature


src/views/index.njk
```diff
@@ -30,7 +30,7 @@
           </ol>
         {% else %}
           <p class="dashboard-section-intro">
-            No interactions have been added in the last {{ totalDays | pluralise( 'day') }}
+            No interactions have been added in the last {{ totalDays }}{{ 'day' | pluralise(totalDays) }}
```
teneightfive: No space between number and word

PR #164, opened 2017-06-06
List of investment projects for a given company

teneightfive:
This adds a list of investment projects for a company and displays them within the company view.

This change also restructures some of the sass components to use a more consistent layout for displaying a list of results.

## Screenshot
company-investments


src/controllers/company-investments.controller.js
```diff
@@ -0,0 +1,32 @@
...
+const { getInflatedDitCompany } = require('../services/company.service')
+const { getCompanyInvestmentProjects } = require('../repos/investment.repo')
+
+function getAction (req, res, next) {
+  const token = req.session.token
+  const companyId = req.params.id
+
+  Q.spawn(function * () {
+    try {
+      const company = yield getInflatedDitCompany(token, companyId)
```
tyom: Looks like this is doing exactly the same thing as [new function](https://github.com/uktrade/data-hub-fe-beta2/pull/151/files#diff-8613cd0b59bd8c0edaa4e1dd1139b5baR21) in `investment.repo`.

We should extract it and use it in all appropriate places instead of redefining the logic.

src/sass/components/_results-item.scss
```diff
@@ -0,0 +1,38 @@
+@import "../settings/colours";
+@import "../settings/layout";
+@import "../tools/functions/typography";
+
+.results-item {
```
tyom: Feels weird bringing `results-item` back as a component. Isn't it always part of `results-list` component?

src/sass/components/_results-item.scss
```diff
@@ -0,0 +1,38 @@
...
+  text-decoration: none;
+
+  &:hover,
+  &:focus {
+    background: $grey-4;
+    border-color: $grey-1;
+  }
+}
+
+  .results-item__header {
```
tyom: Unintended whitespace?

src/controllers/company-investments.controller.js
teneightfive: Yep, I can use that method instead.

teneightfive: There was talk with @jim68000 do return investments on a company object too. In a similar fashion to how we return contacts on a company object. Which would remove the need for these 2 calls.

src/views/investment/start.njk
```diff
@@ -63,19 +63,19 @@
...
 
       {% if searchResult.count > 0 %}
-        <ol class="results-list results-list--block">
+        <ol>
           {% for result in searchResult.results %}
-            <li class="search-result">
-              <a href="/investment/create?equity-company={{ result.id }}" class="search-result__link">
-                <h3 class="search-result__title">{{ result.name | highlightTerm(searchTerm) | safe }}</h3>
+            <li>
+              <a href="/investment/create?equity-company={{ result.id }}" class="results-item results-item--link">
```
tyom: I think it should be: `ol.results-list`, `li.results-list__item` and `a.results-list__link`.

src/sass/components/_results-item.scss
teneightfive: So this may need some discussion. I've actually removed `results-list` as a component as the style it was providing wasn't needed. 

`results-item` is then the only thing that is a component that provides layout. It also had a new modifier which is used for the block style listing on investment results and the classNames are also moved onto the `<a>` element.

May need some further thought but felt like `results-list` wasn't necessary as a sass component at the moment.

src/sass/components/_results-item.scss
teneightfive: It's actually intentional. To show the parent/child relationship of Block->Element. 

What are the thoughts @feedmypixel @ztolley? Maybe it's not something we adopt. Or if we do, it should be added to the linter.

tyom: Oh no. Please let's not go down this road again.

src/views/investment/start.njk
teneightfive: My thought behind this was that the style being applied it more of a modifier than a child element. It's happening to the same root element rather than it being an element in its own right. 

We could do it so the modifier is on the `results-list` component but still means having `results-list__item` on either an `li` or `a` depending on the type.

src/sass/components/_results-item.scss
tyom: Let's have a a chat about list components.

teneightfive: Okay. Thinking about it, this might not always have to be a list.

For example, with investment projects, each project could be an `<article>` but use the same class to apply the style?

src/sass/components/_results-item.scss
teneightfive: I'll remove to prevent your tears

src/views/investment/start.njk
tyom: Why is it a modifier though? It's a different HTML element, has different attributes and functionality (active link vs inert list item). I reckon it should be an element not a modifier.

src/sass/components/_results-item.scss
tyom: To me `results-item` suggests a part of sequence (a list). Whether it's an actual HTML list element is not important. Semantically it's still part of sequence, which should always have container `results-list`.

```pug
section.results-list
  article.results-list__item
  article.results-list__item
```
Can work just as well.

src/views/investment/start.njk
teneightfive: You can't have this inside the other without also adding a modifier to either the list item or the parent to remove the style applied to the list item though.

So rather than applying new style to the link element and applying a modifier to the list element I made it a modifier. Maybe `--link` is misleading. Maybe something like `--block` would be easier to understand.

It changes the visual layout from this:
screen shot 2017-06-06 at 18 11 06

To this:
screen shot 2017-06-06 at 18 11 20

And this is done using the same markup so felt more like a modifier than a completely new element.

src/sass/components/_results-item.scss
teneightfive: Even if the block isn't applying any style? It also limits the number of elements we can use within each item. That short markup makes sense but feels like it needs to be extracted if we have this:

```njk
section.results-list
  article.results-list__item
    h2.results-list__item__header
    p.results-list__item__lede
  article.results-list__item
    h2.results-list__item__header
    p.results-list__item__lede
```

It then feels more like the item is a block in its own right.

src/views/investment/start.njk
tyom: Seems to me this could be better expressed by styling `li`s as list items and introducing `a` as a separate element that has styles specific to it rather than mixing list item styles together with anchor. That way you can add/remove/hide elements without having to worry about updating to the correct modifiers.

src/sass/components/_results-item.scss
teneightfive: Maybe we're getting hung up on it being contained in a list when really the block style doesn't have to be that.

It could just be called `.result` or `.card` with modifiers.

tyom: The block can apply styles such as margins and list styling though. Then items either inherit styles from the container or define styles specific to that list item.

Also, we don't want to have sub-elements in BEM. Can scope those elements to their block `results-list`

```pug
section.results-list
  article.results-list__item
    h2.results-list__header
    p.results-list__lede
  article.results-list__item
    h2.results-list__header
    p.results-list__lede
```
It doesn't have to have direct mapping to HTML structure.

src/views/investment/start.njk
teneightfive: But you can't add a separate element for the `a` without also creating a modifier for the `li`.

src/controllers/company-investments.controller.js
jim68000: It's in V3 of the API, but I just parked that so that Reupen can build the endpoints for archiving and evaluation.

jim68000:
NB, this is probably not the best named PR. 'Pipeline' is for the view of investment projects across company; this is the investments tab

teneightfive:
>NB, this is probably not the best named PR

Changed :)

config/nunjucks/filters.js
```diff
@@ -6,7 +6,7 @@ const filters = {
     const regEx = new RegExp(`(${searchTerm})`, 'gi')
 
     return new nunjucks.runtime.SafeString(
-      string.replace(regEx, '<span class="results-highlight">$1</span>')
+      string.replace(regEx, '<span class="u-highlight">$1</span>')
```
feedmypixel: 👍

src/sass/trumps/_utility.scss
```diff
@@ -19,3 +19,7 @@
     display: none !important;
   }
 }
+
+.u-highlight {
+  font-weight: bold;
```
feedmypixel: `!important` needed here?

src/views/investment/start.njk
```diff
@@ -65,19 +65,19 @@
...
+                <h3 class="results-item__header">{{ result.name | highlight(searchTerm) }}</h3>
 
                 <dl class="metadata-list">
-                  <dt class="metadata-list__term">Registered address</dt>
+                  <dt class="metadata-list__term u-visually-hidden">Registered address</dt>
                   <dd class="metadata-list__value metadata-list__value--block">
                     {% set comma = joiner() %}
                     {% for part in [result.registered_address_1, result.registered_address_2, result.registered_address_town, result.registered_address_postcode] -%}
```
feedmypixel: Maybe not part of this PR but you could use the [addressFormatter](https://github.com/uktrade/data-hub-fe-beta2/blob/develop/src/views/_macros/address.njk) macro here

src/sass/trumps/_utility.scss
teneightfive: I'll add for consistency but I'm not sure I'm convinced that utilities are trumps that require an important flag.

They are far enough down that cascade that it should win and we shouldn't be style element selectors if we're using BEM properly anyway.

teneightfive: I also have a feeling this particular highlight may come back and bite us at some point so might need to revisit.

If within a particular component we might want it displayed differently...but then I suppose that could also be a modifier.

src/sass/components/_results-item.scss
teneightfive: >h2.results-list__item-header

This suggests the element is a child of `results-list__item` rather than `results-list` and feels like a hack to not use `results-list__item__header` or `results-item__header` which is what it essentially is.

src/views/investment/start.njk
teneightfive: Have created #169 to make sure this gets looked at.

PR #165, opened 2017-06-07
Full width layouts

### LINKED ISSUE #121: Use full width of the browser
## Problem

We currently restrict the page to the same width as is being used for GOV.UK and public facing services. This isn't as big a concern for this app as it's an internal facing admin interface.

## Solution

We can there make use of the full width of the browser. We should define our own grid and remove [the grid being imported](https://github.com/uktrade/trade-elements/blob/master/src/sass/ukti-template.scss#L11) from govuk elements.

teneightfive:
This adds the ability to use a full width layout instead of the one restricted to `960px`. The majority of the views now inherit this full width layout. For the time being the login view has been kept at a restricted with.

This PR has been separated from setup work that has now been done in #173.

## BLOCKED REASON

This change is temporarily on hold whilst we decide if it will have larger layout implications. Some views may look unusable on much larger resolutions so we should consider those properly before merging.

## Issue addressed

Closes #121

### DIFF
--- a/assets/stylesheets/_layout.scss
+++ b/assets/stylesheets/_layout.scss
@@ -1,1 +1,2 @@
 @import "layout/assemblies";
+@import "layout/containers";
--- a/assets/stylesheets/layout/_containers.scss
+++ b/assets/stylesheets/layout/_containers.scss
@@ -1,0 +1,14 @@
+@import "../tools/mixins/layout";
+
+.l-container {
+  @include site-width-container;
+}
+
+.l-container--restricted {
+  max-width: $site-width;
+
+  @include media($min-width: ($site-width + $gutter * 2)) {
+    margin-left: auto;
+    margin-right: auto;
+  }
+}
--- a/assets/stylesheets/objects/base/_main-content.scss
+++ b/assets/stylesheets/objects/base/_main-content.scss
@@ -1,4 +1,5 @@
 @import "../../settings";
+@import "../../tools/mixins/layout";
 
 .main-content {
   position: relative;
@@ -11,17 +12,15 @@
     position: absolute;
     z-index: 10;
     top: 0;
-    right: 0;
-    left: 0;
+    right: ($gutter / 2);
+    left: ($gutter / 2);
     border-top: 10px solid $govuk-blue;
     max-width: 100%;
-    width: $site-width;
     margin: 0 auto;
 
-    @include media($max-width: $site-width) {
-      width: auto;
-      left: $default-spacing-unit;
-      right: $default-spacing-unit;
+    @include media(desktop) {
+      left: $gutter;
+      right: $gutter;
     }
   }
 }
--- a/assets/stylesheets/tools/mixins/_layout.scss
+++ b/assets/stylesheets/tools/mixins/_layout.scss
@@ -56,15 +56,12 @@
 }
 
 @mixin site-width-container {
-  margin: 0 ($gutter / 2);
-  max-width: $site-width;
+  margin-left: ($gutter / 2);
+  margin-right: ($gutter / 2);
 
   @include media(desktop) {
-    margin: 0 $gutter;
-  }
-
-  @include media($min-width: ($site-width + $gutter * 2)) {
-    margin: 0 auto;
+    margin-left: $gutter;
+    margin-right: $gutter;
   }
 }
 
--- a/assets/stylesheets/trumps/_layout.scss
+++ b/assets/stylesheets/trumps/_layout.scss
@@ -1,18 +1,3 @@
-@import "../tools/mixins/layout";
-
-.l-container {
-  margin-left: auto;
-  margin-right: auto;
-  padding-left: $default-spacing-unit;
-  padding-right: $default-spacing-unit;
-  max-width: $site-width;
-
-  @include media($min-width: $site-width) {
-    padding-left: 0;
-    padding-right: 0;
-  }
-}
-
 .column-two-quarters,
 .column-three-quarters {
   > h1,


config/nunjucks/index.js
```diff
@@ -59,6 +59,10 @@ module.exports = (app, config) => {
   env.addExtension('ComponentExtension', new ComponentExtension(env))
 
   // Global variables
+  env.addGlobal('siteTitle', 'Data Hub')
```
teneightfive: Was considering setting these in the global config file and passing in here but this is the only place they'd be used and they wouldn't be overridden per environment.

src/sass/tools/mixins/_layout.scss
```diff
@@ -56,15 +56,24 @@
...
+    margin-left: $gutter;
+    margin-right: $gutter;
   }
 
   @include media($min-width: ($site-width + $gutter * 2)) {
-    margin: 0 auto;
+    margin-left: auto;
+    margin-right: auto;
+
+    .is-fluid & {
```
teneightfive: Not sure if this is the best class name...open to suggestions...

config/nunjucks/index.js
tyom: Feels a bit strange to set global site variables in Nunjucks config file. Should these be a separate globals config file perhaps?

src/sass/tools/mixins/_layout.scss
jim68000: .is-fullwidth?

src/views/_layouts/dit-base.njk
```diff
@@ -1,6 +1,6 @@
 {%- set assetPath = asset_path | default('/') -%}
 <!DOCTYPE html>
-<html lang="en" class="no-js">
+<html lang="en" class="no-js{{ bodyClass }}">
```
tyom: Is it actually the `bodyClass` or a `rootClass`?

tyom: Should add a space `no-js` and `bodyClass`.

config/nunjucks/index.js
```diff
@@ -59,6 +59,10 @@ module.exports = (app, config) => {
...
+  env.addGlobal('projectPhase', 'beta')
+  env.addGlobal('feedbackLink', '/support/bug')
+  env.addGlobal('bodyClass', ' is-fluid')
```
tyom: I would suggest avoiding spaces in globals like this. Better add space in class declaration in HTML. Browser happily ignores them.

src/sass/tools/mixins/_layout.scss
tyom: `layout-fluid`?

src/views/_layouts/dit-base.njk
```diff
@@ -61,7 +61,7 @@
 
       {% block body_main %}
         <div class="main-content">
-          <main id="content" role="main" class="main-content__inner">
+          <main id="main-content" role="main" class="main-content__inner">
```
tyom: The old ID actually had a bunch of styles associated with it. For example, bottom padding to allow space between content and footer.

src/sass/tools/mixins/_layout.scss
feedmypixel: do we have a need for both? is this just because of
> For the time being the login view has been kept at a restricted with.

config/nunjucks/index.js
teneightfive: You thinking similar method to the [filters file](https://github.com/uktrade/data-hub-fe-beta2/blob/32a805bf77b5437740e4fa8998ddf1598b08107f/config/nunjucks/filters.js)? Or outside of nunjucks?

The are tied to the nunjucks layouts so didn't feel too strange to me to have them set as nunjucks globals. The problem with setting them in the template is [they can't be overridden](https://github.com/mozilla/nunjucks/issues/406) in subsequent layouts that inherit from it.

src/sass/tools/mixins/_layout.scss
teneightfive: @feedmypixel Yes it's for that reason for now as it didn't feel like the login view warranted taking the full width.

We also need to think of this stuff as a base for future services too as we may extract that out to the next incarnation of trade elements. 

>.is-fullwidth?

@feedmypixel raised a good point that we may still want a max-width for really, really big resolutions so it might not always be full width.

Almost feels like the width restricted version should be a modifier?

src/views/_layouts/dit-base.njk
teneightfive: >Should add a space no-js and bodyClass

It's currently added in the variable but you're right, it's probably better to set here.

src/views/_layouts/dit-base.njk
teneightfive: Okay, I'll replace those too as we want to moving away from those, right?

teneightfive:
>Agree with @tyom - the globals need to go in a global config. I'm sure we'll find other things to put in it

@jim68000 I made a comment above about this. They are specifically concerned with the layout rather than anywhere else in the app which is why I placed them inside the nunjucks config initially. We could do something similar to filters but they are nicely abstracted so they can be tested but not sure it's worth us testing global layout defaults?

src/views/_layouts/dit-base.njk
tyom: Just add the bits we need from there. We do want to move away from those eventually.

teneightfive: Done.

config/nunjucks/index.js
tyom: I was just thinking to have a file like `config/template-globals.js` (or JSON) that would have a list of global variable pairs like
```json
{
  "projectPhase": "beta",
  "feedbackLink": "/support/bug"
}
```
And in this file we would just loop through the keys and add those variables to `env`.
```js
forOwn(require('../template-globals.json'), (value, key) => {
  env.addGlobal(key, value)
})
```

src/sass/tools/mixins/_layout.scss
tyom: I don't think it's appropriate to use state class name for layout modifier.

teneightfive: What about a layout prefix (`l-`) - `l-fluid` or `l-contained` if we use fluid width as the default?

ztolley:
definitely needs everyone to have a look and figure out how to make search look less baron

teneightfive:
This is being closed for now as there is still some work to be done around elements to make them use the full width of the browser effectively.

Issue #74, opened 2017-05-16
Folder structure


feedmypixel:
## Problem
At the moment we are organising our code around roles

- `src/controllers`
- `src/repos`
- `src/services`

This involves an amount of digging around to find work and relies on naming conventions being followed to find associated files. That or `ctrl + click` throughs in the code.

## Solution/Suggestion
> Organize your Files Around Features, Not Roles

[Project Structure](https://blog.risingstack.com/node-hero-node-js-project-structure-tutorial/)
Then it will be a lot simplier to see what does what and what is connected.

- `src/contacts`
- `src/interactions`
- `src/ltd-company`
- `src/foreign`

Then if wanted we can follow a naming convention of `.controller`, `.service`, `.repo` and group any other files that may be needed with the structure
```
src
└── contacts
     contact.controller.js
     contact.service.js
     contact.repo.js
     contact.test.js
     README.md
```
> Groupings/Naming all up for grabs

teneightfive:
Definitely agree that we should move to feature grouping as the app grows. Involves a lot of jumping around when building one feature at the moment. 

This is a suggestion for overall folder structure:

```
/config
  express.js
  index.js
  nunjucks.js
/[assets|client]
  /javascripts
  /stylesheets
/[server|app|src]
  /[feature|component|product]
    feature-name.controller.js
    feature-name.controller.test.js
    feature-name.service.js
    feature-name.service.test.js
    feature-name.repo.js
    feature-name.repo.test.js
    feature-name.view.njk
  /lib
  /public
  /views
    /_components
    /_includes
    /_layouts
```

Not convinced on where the generic layouts/components/includes should live though.

feedmypixel:
this also covers discussion from #73

tyom:
Can we rename 'views' to 'templates' please?

Layouts and macros are not views.

teneightfive:
Yes, templates probably does make more sense.

teneightfive:
For reference #151 starts to look at using express sub apps for routing and contained logic on investment.

teneightfive:
After some discussion we came to a consensus of a proposed structure for now:

```
/config
  express.js
  index.js
  nunjucks.js
/assets
  /javascripts
  /stylesheets
/src
  /apps
    /[feature]
      feature-name.controller.js
      feature-name.service.js
      feature-name.repo.js
      index.js
      labels.js
      router.js

  /middleware
  /lib
  /templates
    /_components
    /_includes
    /_layouts
[...]

PR #168, opened 2017-06-07
fix for add company search

feedmypixel:
This work involves:
- updating the search used on `/company/add-step-2/` to work with the new search api.
- adding markup to display `Private/Limited` companies
- updating tests and adding test data fixtures

### Of Note
- I have moved a portion of the simple logic into the template away from the controller
- There are parts of this code that I am not happy with and need to be addressed at a future point.
In particular the tests are just testing that methods have been called and could do with a refactor
- I would like to approach this in a `middleware` fashion similar to work from @tyom 

add-company

### DIFF
--- a/src/controllers/company-add.controller.js
+++ b/src/controllers/company-add.controller.js
@@ -1,11 +1,11 @@
 const express = require('express')
 const Q = require('q')
-const winston = require('winston')
 const { ukOtherCompanyOptions, foreignOtherCompanyOptions } = require('../options')
+const { getCHCompany } = require('../repos/company.repo')
 const { isBlank, toQueryString } = require('../lib/controller-utils')
-const searchService = require('../services/search.service')
-const companyService = require('../services/company.service')
-const companyFormattingService = require('../services/company-formatting.service')
+const { searchLimitedCompanies } = require('../services/search.service')
+const { getDisplayCH, getDisplayCompany } = require('../services/company-formatting.service')
+const { getPagination } = require('../lib/pagination')
 const { companyDetailsLabels, chDetailsLabels, companyTypeOptions } = require('../labels/company-labels')
 
 const router = express.Router()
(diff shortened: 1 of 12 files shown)


data.json
```diff
@@ -0,0 +1,26 @@
+{
```
jim68000: Does this need to be in the root folder?

teneightfive:
>I would like to approach this in a middleware fashion similar to work from @tyom

I think this is looking like a nice approach that we can start to use for features/sections of the site. It will also address issue #74.

data.json
```diff
@@ -0,0 +1,26 @@
...
+  "sic_code_2": "",
+  "sic_code_3": "",
+  "sic_code_4": "",
+  "uri": "http://business.data.gov.uk/id/company/06960817",
+  "incorporation_date": "2009-07-13",
+  "registered_address_country": {
+    "id": "80756b9a-5d95-e211-a939-e4115bead28a",
+    "name": "United Kingdom"
+  }
+}
```
teneightfive: Missing new line

data.json
teneightfive: Where is this file being used?

teneightfive: Feels like it might have been an accidental save...

src/controllers/company-add.controller.js
```diff
@@ -1,10 +1,9 @@
 const express = require('express')
 const Q = require('q')
-const winston = require('winston')
 const { ukOtherCompanyOptions, foreignOtherCompanyOptions } = require('../options')
+const companyRepository = require('../repos/company.repo')
```
teneightfive: Use destructuring for consistency?
```js
const { getCHCompany } = require('../repos/company.repo')
```

src/controllers/investment-start.controller.js
```diff
@@ -19,10 +19,11 @@ function getHandler (req, res, next) {
   }
 
   if (searchTerm) {
-    promises.push(searchForeignCompany({
+    promises.push(searchCompany({
       token: req.session.token,
       page: req.query.page,
       searchTerm,
+      isUkBased: false,
```
teneightfive: `isUKBased`?

src/services/search.service.js
```diff
@@ -52,6 +51,20 @@ function searchForeignCompany ({ token, searchTerm, page = 1, limit = 10 }) {
...
 }
 
+function searchLimitedCompany ({ token, searchTerm }) {
+  return searchCompany({
+    token,
+    searchTerm,
+    isUkBased: true,
+  })
+    .then((result) => {
+      return result.results.filter((company) => {
```
teneightfive: Will this not return issues if using pagination?

src/services/search.service.js
```diff
@@ -29,14 +28,14 @@ function search ({ token, searchTerm, searchType, limit = 10, page = 1 }) {
 // TODO this work has been abstracted for easy removal of code once this api call has has been changed to GET
 // TODO we have query params on a POST's url
 // TODO we have a param named `original_query` that is set to `term` on the backend, this needs aligning to other endpoints and changed to `term` in the request
-function searchForeignCompany ({ token, searchTerm, page = 1, limit = 10 }) {
+function searchCompany ({ token, searchTerm, isUkBased, page = 1, limit = 10 }) {
```
teneightfive: I quite liked the abstraction before by having a different method to call to return foreign companies.

Maybe we could keep the separate method but have it call a shared search method passing `uk_based: false` in.

We may have a need for a `searchUKCompany()` later on as well as `searchLimitedCompany()`.

On a side note, would the method names be better if they were plurals? We are looking through a list of companies rather than looking for a single record.

src/views/company/add-step-2.njk
```diff
@@ -10,48 +10,71 @@
...
         <li class="results-list__result">
-          {% if hit.selected %}
-            <a class="open" href="{{closeLink}}">
-              <h3 class="result-title">{{hit.name}}</h3>
+          {% set isSelected = currentlySelected and (currentlySelected or company.company_number == currentlySelected) %}
+
+          {% if isSelected %}
+            {% set openHref = '/company/add-step-2/?term=' + searchTerm + '&business_type=' + businessType + '&country=' + country %}
+
+            <a class="open" href="{{ openHref }}">
```
teneightfive: Could this be this instead: 
```njk
<a href="/company/add-step-2/?term={{ searchTerm }}&business_type={{ businessType }}&country={{ country}}">
```

src/views/company/add-step-2.njk
```diff
@@ -10,48 +10,71 @@
...
+              <h3 class="result-title">{{company.name}}</h3>
             </a>
           {% else %}
-            <a class="closed" href="{{hit.url}}">
-              <h3 class="result-title">{{hit.name}}</h3>
+            {% set closedHref = '/company/add-step-2/?term=' + searchTerm + '&business_type=' + businessType + '&country=' + country + '&selected=' + company.company_number if company.company_number else company.id %}
+
+            <a class="closed" href="{{ closedHref }}">
```
teneightfive: Same as open href

src/controllers/investment/start.controller.js
```diff
@@ -19,10 +19,11 @@ function getHandler (req, res, next) {
   }
 
   if (searchTerm) {
-    promises.push(searchForeignCompany({
+    promises.push(searchCompany({
       token: req.session.token,
       page: req.query.page,
       searchTerm,
+      isUkBased: false,
```
teneightfive: `isUKBased`?

tyom:
Should we remove the left-hand column in search results?

src/services/search.service.js
feedmypixel: yeah was going to leave pagination until another day hence this. I have fixed and added pagination although I haven't seen it working due not being able to find results in the data base that need to paginate

src/views/company/add-step-2.njk
feedmypixel: yeah it started like that 😄 but felt a little hard to read hence abstracting it out to a var - will move back if you feel it reads better?

feedmypixel:
> Should we remove the left-hand column in search results?

@tyom yes I was thinking this with the main results also. There are filters coming in in the left column but it would be nice to have search full width all over. Maybe not part of this PR though?

data.json
feedmypixel: Yup my bad 👍

src/services/search.service.js
```diff
@@ -52,57 +47,33 @@ function searchForeignCompany ({ token, searchTerm, page = 1, limit = 10 }) {
...
-          (result._source.business_type && result._source.business_type.toLowerCase() === 'private limited company') ||
-          (result._source.business_type && result._source.business_type.toLowerCase() === 'public limited company')))
+function searchLimitedCompanies (options) {
+  const optionsUkBasedTrue = Object.assign({}, options, { isUkBased: true })
+
+  return searchCompanies(optionsUkBasedTrue)
+    .then((result) => {
+      const limitedCompanies = result.results.filter((company) => {
+        return company.company_number ||
+          (company.business_type && company.business_type.name.toLowerCase().includes('limited company'))
```
reupen: There aren't Companies House companies in search at the moment, so we have a bit of a problem... This is just getting Data Hub companies that are UK companies.

feedmypixel: Ok - but this isn't an issue with this work right its the data?

reupen: The data isn't wrong, it's just that Companies House companies weren't really considered with the new search. But, yes, I'd fix it separately once they're in search and we know how it works.

PR #171, opened 2017-06-08
Allow objects to be filtered by removeFalsey filter

tyom:
Nunjucks allows to use `length` filter to check if object is empty. This allows us to check objects have values before passing them to components. If objects has `undefined` or `null` values it might be considered not worthy of showing. Allowing to filter out those values gives ability to only show values that are explicitly set.

```njk
{% set obj = { a: null, b: undefined, c: 'value', d: false, e: '' } %}
{% set arr = [null, undefined, 'value', false, ''] %}

{% if obj | removeFalsey | length %}
  {{ obj | dump }} {# { c: 'value', d: false } #}
{% endif %}
{{ arr | dump }} {# [ 'value', false ] #}
```

### DIFF
--- a/config/nunjucks/filters.js
+++ b/config/nunjucks/filters.js
@@ -2,9 +2,14 @@ const nunjucks = require('nunjucks')
 const {
   isArray,
   isPlainObject,
+  isEmpty,
   pickBy,
   isNil,
 } = require('lodash')
+
+function isNotEmpty (value) {
+  return !isNil(value) && !/^\s*$/.test(value) && !(isPlainObject(value) && isEmpty(value))
+}
 
 const filters = {
   stringify: JSON.stringify,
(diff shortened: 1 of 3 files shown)


tyom:
Hmm.. Maybe `removeFalsey` is not such a great name for this filter.

config/nunjucks/filters.js
```diff
@@ -1,4 +1,15 @@
 const nunjucks = require('nunjucks')
+const {
+  isArray,
+  isPlainObject,
+  pickBy,
+  isNil,
+} = require('lodash')
+
+function isNotEmpty (value) {
+  return !isNil(value) && !/^\s*$/.test(value)
```
feedmypixel: do we also need to be checking for empty arrays here?

config/nunjucks/filters.js
```diff
@@ -10,8 +21,14 @@ const filters = {
     )
   },
 
-  removeFalsey: (array) => {
-    return array.filter(item => item)
+  removeEmpty: (collection) => {
```
feedmypixel: Potentially could be named `removeNilAndEmpty`?

tyom: Nil implies empty. I'd go for a shorter filter name.

config/nunjucks/filters.js
tyom: Good point. Empty objects too.

tyom: Actually this Regex test takes cares of empty arrays because its called on `[].toString()`.

I've added the test for empty objects.

config/nunjucks/filters.js
feedmypixel: `removeNil` ? feels like `removeEmpty` only covers half of the cases `('', [])`. Agreed `removeNilAndEmpty` is a little verbose but it "does what it says on the tin"™

PR #173, opened 2017-06-08
Global layout tweaks

teneightfive:
This adds:

- a change to the main content ID to avoid style clash with trade elements
- a small refactor of the layouts folder to make it more consistent with how we structure includes, marcos and components in the views folder
- moving template defaults into [nunjucks globals](https://mozilla.github.io/nunjucks/api.html#addglobal) to allow them to be overridden. See [nunjucks issue #406](https://github.com/mozilla/nunjucks/issues/406) for why this needs to be done.
---
**Note:** This PR has been spilt out from #165 to merge initial changes and leave a full width layout fix to a later date.


config/nunjucks/globals.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
+  'siteTitle': 'Data Hub',
```
tyom: Maybe drop the quotes for keys.

PR #176, opened 2017-06-08
Set correct view locals for a company view

teneightfive:
A function wasn't being called in the investment list view which
prevented some information being displayed in the view.

### DIFF
--- a/src/controllers/company-investments.controller.js
+++ b/src/controllers/company-investments.controller.js
@@ -1,7 +1,7 @@
 const router = require('express').Router()
 const Q = require('q')
 
-const { getInflatedDitCompany } = require('../services/company.service')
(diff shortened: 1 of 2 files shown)


src/controllers/company-investments.controller.js
```diff
@@ -1,7 +1,7 @@
 const router = require('express').Router()
 const Q = require('q')
 
-const { getInflatedDitCompany } = require('../services/company.service')
+const { getInflatedDitCompany, getCommonTitlesAndlinks } = require('../services/company.service')
```
tyom: `AndLinks`

teneightfive: That's the name of the method...don't think that's in scope for this change.

The whole company section will need a refactor to a similar structure as we now have investment.

tyom: 😞

PR #184, opened 2017-06-09
Use POST to unarchive contacts and companies when calling the back end

reupen:
GET for unarchiving is being removed from the back end in favour of POST.


teneightfive:
@reupen this will require some changes on the frontend so am assuming we should hold off merging this until we've implemented them and do a simultaneous release?

reupen:
@teneightfive The back end supports both GET and POST at the moment. Will remove GET there once this is in.

teneightfive:
>The back end supports both GET and POST at the moment.
👍

PR #185, opened 2017-06-09
Initial work for Investment value

tyom:
- [x] Create/edit investment value
- [x] Allow to transform data independent of labels
- [x] Display correct form labels

image


src/controllers/investment/audit.controller.js
```diff
@@ -2,7 +2,7 @@ const router = require('express').Router()
 const Q = require('q')
 const { get } = require('lodash')
 
-const { getProjectDetails } = require('./shared')
+const { getProjectDetails } = require('./middleware')
```
teneightfive: Would a better filename be `shared.middleware` to use the same pattern as `form.middleware`?

tyom: I was thinking that without specific name it would be collecting all default ones.

`form.middleware.js`, `something.middleware.js`, `middleware.js`

src/services/investment-formatting.service.js
```diff
@@ -3,6 +3,8 @@ const { mapValues, get } = require('lodash')
 const { buildCompanyUrl } = require('./company.service')
 
 function transformToApi (body) {
+  if (!body) { return }
```
teneightfive: Do we want to keep blocks expanded?

src/views/investment/value-edit.njk
```diff
@@ -0,0 +1,130 @@
...
+      ) }}
+    </div>
+    <div class="form-group">
+      {{ trade.textbox(
+          'foreign_equity_investment',
+          value=form.state['foreign_equity_investment'],
+          label='Foreign equity investment'
+      ) }}
+    </div>
+    <div class="form-group">
```
teneightfive: Aren't the `form-group` classes already applied within the macro/component?

src/controllers/investment/audit.controller.js
teneightfive: Feels a bit too vague. I think we should actually apply the naming convention to the rest of our middleware like we do with controllers: `[name].middleware.js`.

PR #187, opened 2017-06-09
Remaining work around investment value

tyom:
- Format data for humans in value view
- Add average salary dropdown field
- Add infostrips for empty value and requiremetns

### DIFF
--- a/src/views/investment/details.njk
+++ b/src/views/investment/details.njk
@@ -22,6 +22,9 @@
     {% component 'key-value-table', items=values, variant='striped' %}
     {{ renderButton('value', 'Edit value', true) }}
   {% else %}
+    <p class="infostrip">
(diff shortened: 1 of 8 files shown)


src/views/investment/details.njk
```diff
@@ -22,6 +22,9 @@
     {% component 'key-value-table', items=values, variant='striped' %}
     {{ renderButton('value', 'Edit value', true) }}
   {% else %}
+    <p class="infostrip">
+      Please complete ’Total investment’ and ‘Number of new jobs’ to move to Assign PM stage
```
teneightfive: `’` is the wrong way around on `’Total investment’`.

tyom:
> Am slightly worried about test coverage slipping for these features. Should make sure we follow up with tests within the investment app.

Definitely. I'm aware of missing tests and am planning to add them as soon as I can.

PR #188, opened 2017-06-09
Remove dead link, fix cookies link

tyom:
😳


teneightfive:
Don't think we need this anymore as #189 removes the whole block for now.

tyom:
It removed it from `datahub-base.njk`. It still remains in `dit-base.njk`.

teneightfive:
>It still remains in dit-base.njk.

I see what you're referring to now...

PR #189, opened 2017-06-10
Layout updates

teneightfive:
This change includes some updates to the header/footer layouts:

- Option to remove phase banner and display phase tag next to site title
- Ability to add a proposition menu to global header
- Remove crown copyright from footer until we know we can include it
- Copy tweaks to sign in/sign out

### Before
before-login

### After
login-after

### Before
before

### After
after

### DIFF
--- a/src/sass/components/_proposition-menu.scss
+++ b/src/sass/components/_proposition-menu.scss
@@ -1,0 +1,61 @@
+@import "../settings/colours";
+@import "../settings/layout";
+@import "../tools/mixins/addons";
+@import "../tools/mixins/conditionals";
+@import "../tools/mixins/layout";
+
+.proposition-menu {
+  @include bold-font(14);
+  color: $white;
+  margin: 0;
+  padding: $baseline-grid-unit 0 0;
+
+  @include media(tablet) {
+    @include bold-font(16);
+    float: right;
+  }
+}
+
+.proposition-menu__item {
+  border-bottom: 1px solid $grey-1;
+  display: block;
+
+  @include media(tablet) {
+    border-bottom: 0;
+    float: left;
+    padding-left: $default-spacing-unit;
+
+    &:first-child {
+      padding-left: 0;
+    }
+  }
+}
+
+.proposition-menu__link {
+  color: inherit;
+  display: block;
+  padding: $baseline-grid-unit;
+  text-decoration: none;
+
+  &:active,
+  &:visited,
+  &:hover {
+    color: inherit;
+  }
+
+  &:focus {
+    color: $black;
+  }
+
+  &:hover {
+    text-decoration: underline;
+  }
+
+  @include media(tablet) {
+    padding: 0;
+  }
+
+  &.is-active {
+    color: $light-blue;
+  }
+}
--- a/src/sass/objects/layout/_global-header.scss
+++ b/src/sass/objects/layout/_global-header.scss
@@ -8,38 +8,61 @@
 }
 
 .global-header__inner {
+  @include clearfix;
   @include site-width-container;
   padding: ($baseline-grid-unit * 3) 0;
 }
 
-.global-header__link {
+.global-header__site-title {
+  word-spacing: -4px;
+
+  @include media(tablet) {
+    float: left;
+  }
+
+  .phase-badge {
+    margin-top: 5px;
+  }
+}
+
+.global-header__logo {
+  border-bottom: 1px solid transparent;
   display: inline-block;
   color: inherit;
-  font-size: 1.6em;
-  font-weight: 600;
   text-decoration: none;
-  padding: 0 $baseline-grid-unit;
-  margin: 0 (-$baseline-grid-unit);
+  padding: 0;
+  margin: 0 $default-spacing-unit 0 0;
 
   &:hover,
   &:visited,
   &:focus {
-    text-decoration: underline;
     color: inherit;
     background-color: transparent;
+  }
+
+  &:hover {
+    border-color: inherit;
+  }
+}
+
+.global-header__site-name {
+  @include bold-font(36);
+  letter-spacing: 1px;
+  line-height: .9;
+
+  @include media(tablet) {
+    line-height: .9;
+    font-size: 30px;
   }
 }
 
 .global-header__service-name {
-  @include core-font(16);
-  color: $grey-1;
-  display: block;
-  margin-top: $baseline-grid-unit * 2;
+  @include core-font(24);
+  color: inherit;
+  line-height: .9;
+  margin-left: $baseline-grid-unit * 2;
 
   @include media(tablet) {
-    @include core-font(20);
-    display: inline;
-    margin-left: $baseline-grid-unit * 4;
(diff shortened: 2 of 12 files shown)


src/sass/components/_proposition-menu.scss
```diff
@@ -0,0 +1,58 @@
...
+.proposition-menu {
+  @include bold-font(14);
+  color: $white;
+  margin: 0;
+  padding: $baseline-grid-unit 0 0;
+
+  @include media(tablet) {
+    @include bold-font(16);
+    float: right;
+
```
ztolley: dead space?

src/sass/objects/layout/_global-header.scss
```diff
@@ -8,38 +8,55 @@
...
     color: inherit;
     background-color: transparent;
   }
+
+  &:hover {
+    border-color: inherit;
+  }
+
+  @include media(tablet) {
+    line-height: 0.9;
```
ztolley: isn't your line height already 0.9?

src/sass/objects/layout/_global-header.scss
```diff
@@ -8,38 +8,55 @@
...
+  display: inline;
+  line-height: 1;
+  margin-left: $baseline-grid-unit;
 
   @include media(tablet) {
-    @include core-font(20);
-    display: inline;
-    margin-left: $baseline-grid-unit * 4;
-    margin-top: 0;
+    line-height: 1;
```
ztolley: isn't the line height already 1?

src/sass/objects/layout/_global-header.scss
teneightfive: Nope. At this point it is `48px` which is equivalent to `1.6`.

src/sass/objects/layout/_global-header.scss
teneightfive: Nope. It is `32px` with a font-size of `20px` which is equivalent to `1.6`.

src/views/_layouts/datahub-base.njk
```diff
@@ -27,3 +56,5 @@
     {% endblock %}
   <!--<![endif]-->
 {% endblock %}
+
+{% block body_footer_content %}{% endblock %}
```
tyom: Maybe remove the whole `body_footer`?

src/sass/components/_phase-tag.scss
```diff
@@ -0,0 +1,22 @@
...
+  background-color: $govuk-blue;
+  color: $white;
+  cursor: default;
+  display: inline-block;
+  font-weight: 600;
+  letter-spacing: 1px;
+  margin: 0 ($default-spacing-unit / 2) 0 0;
+  padding: 2px 6px 0px;
+
+  .global-header & {
```
tyom: I'm inclined to think that component shouldn't not about all the various contexts it may appear. Instead it should have generic styles and each context would then adapt it to a specific use case.

teneightfive: Yeah wasn't happy with this bit. Where would these changes live. Possibly a modifier?

src/sass/objects/layout/_global-header.scss
```diff
@@ -8,38 +8,55 @@
...
+    font-size: 30px;
+  }
 }
 
 .global-header__service-name {
-  @include core-font(16);
-  color: $grey-1;
-  display: block;
-  margin-top: $baseline-grid-unit * 2;
+  @include core-font(20);
```
tyom: Hmm. I'm not sure the font size/alignment works. Why are there three different font sizes in title?

src/sass/components/_phase-tag.scss
tyom: Maybe `global-header` component can define these rules?

src/views/_layouts/datahub-base.njk
teneightfive: Not sure. Leaving the grey footer possibly helps signify the end of the page at this stage. 

And we may add some content to this section like more support links, links to a guide...

Would be less of a jump back if we leave the footer container for now.

src/views/_layouts/datahub-base.njk
```diff
@@ -17,6 +17,35 @@
...
+        {% if user %}
+          <li class="proposition-menu__item">
+            <a href="/myaccount" class="proposition-menu__link">{{ user.name }}</a>
+          </li>
+        {% endif %}
+        <li class="proposition-menu__item">
+          {% if user %}
+            <a href="/login/signout" class="proposition-menu__link">Sign out</a>
+          {% else %}
+            <a href="/login" class="proposition-menu__link">Sign in</a>
```
tyom: Is this needed? AFAIK we always redirect to `/login` when signed out anyway.

src/sass/components/_phase-tag.scss
teneightfive: Isn't it clearer to show how a component would change within its context rather than declaring those rules within a separate component.

Maybe a modifier would be best as it may not apply just to when it sits within that component.

src/sass/objects/layout/_global-header.scss
teneightfive: Not sure it works visually?

src/views/_layouts/datahub-base.njk
teneightfive: Yes. I noticed if you click the support link (which doesn't require you to be logged in) you'll end up trapped without a way to clearly get back to a page where you can sign in.

src/sass/components/_phase-tag.scss
tyom: But then component would have to know about the context outside of its own definition. IMO that's not a modular approach.

tyom: Does it even need these styles? It seems fine if I disable them in dev tools.

src/sass/objects/layout/_global-header.scss
tyom: I think the font size is jumping too much which distracts me personally and probably breaks some typographic rules.

src/views/_layouts/datahub-base.njk
tyom: Fair enough.

src/views/_layouts/datahub-base.njk
tyom: Dunno. I think it's just another thing that may confuse user (is something meant to be there?)

src/sass/components/_phase-tag.scss
teneightfive: Seems the float and display can go but the margin does align it correctly. Let me have a look at the phase banner version see if we can come up with a happy medium.

src/sass/objects/layout/_global-header.scss
teneightfive: What about this:

screen shot 2017-06-12 at 14 30 44

teneightfive: Or even lighter weight for service name:

screen shot 2017-06-12 at 15 02 49

PR #195, opened 2017-06-13
Add missing checkboxes in investment value

tyom:
image

### DIFF
--- a/src/controllers/investment/labels.js
+++ b/src/controllers/investment/labels.js
@@ -31,6 +31,8 @@ const labels = {
       non_fdi_r_and_d_budget: 'Is this project associated with a non-FDI R&D project?',
       new_tech_to_uk: 'Does the project bring ‘New To World’ Technology, IP or Business Model to the UK site?',
       export_revenue: 'Will the UK company export a significant proportion of their products and services produced in the UK as a result of the FDI project?',
+      client_cannot_provide_total_investment: 'Client cannot provide this information',
+      client_cannot_provide_foreign_investment: 'Client cannot provide this information',
     },
   },
   requirementsLabels: {
--- a/src/javascripts/modules/label-select.js
+++ b/src/javascripts/modules/label-select.js
@@ -1,0 +1,50 @@
+const { addClass, removeClass } = require('../../lib/element-stuff.js')
+
+function getControlLabel (control) {
+  if (control.type !== 'checkbox') { return }
+
+  const labelEl = control.parentElement
+
+  if (labelEl.className.indexOf('selection-button-checkbox') < 0) { return }
+
+  return labelEl
+}
+
+const LabelSelect = {
+  init () {
+    this.bindEvents()
+  },
+
+  bindEvents () {
+    document.addEventListener('change', this.onChangeHandler)
+    document.addEventListener('focus', this.onFocusHandler, true)
+  },
+
+  onChangeHandler (evt) {
+    const target = evt.target || evt.srcElement
+    const labelEl = getControlLabel(target)
+    const selectedClassName = 'selected'
+
+    if (!labelEl) { return }
+
+    if (target.checked) {
+      addClass(labelEl, selectedClassName)
+    } else {
+      removeClass(labelEl, selectedClassName)
+    }
+  },
+
+  onFocusHandler (evt) {
+    const target = evt.target || evt.srcElement
+    const focusedEl = getControlLabel(target)
+    const focusedClassName = 'focused'
+
+    if (!focusedEl) { return }
+
+    addClass(focusedEl, focusedClassName)
+
+    target.addEventListener('blur', () => removeClass(focusedEl, focusedClassName))
+  },
+}
+
+module.exports = LabelSelect
--- a/src/controllers/investment/form.middleware.js
+++ b/src/controllers/investment/form.middleware.js
@@ -149,6 +149,8 @@ function investmentValueFormPostMiddleware (req, res, next) {
     average_salary: {
       id: req.body.average_salary,
     },
+    client_cannot_provide_total_investment: req.body.client_cannot_provide_total_investment === 'on',
+    client_cannot_provide_foreign_investment: req.body.client_cannot_provide_foreign_investment === 'on',
   })
 
   updateInvestmentValue(req.session.token, res.locals.projectId, formattedBody)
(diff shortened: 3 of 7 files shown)


src/controllers/investment/labels.js
```diff
@@ -31,6 +31,8 @@ const labels = {
       non_fdi_r_and_d_budget: 'Is this project associated with a non-FDI R&D project?',
       new_tech_to_uk: 'Does the project bring ‘New To World’ Technology, IP or Business Model to the UK site?',
       export_revenue: 'Will the UK company export a significant proportion of their products and services produced in the UK as a result of the FDI project?',
+      client_cannot_provide_total_investment: 'Client cannot provide this information',
```
tyom: I think the labels should describe the value they are sending. Here they are relying on parent context which would be an accessibility issue.

@jim68000 @uktrade/datahub-design

src/javascripts/modules/label-select.js
```diff
@@ -0,0 +1,48 @@
...
+    }
+  },
+
+  onFocusHandler (evt) {
+    const target = evt.target || evt.srcElement
+    const focusedEl = getControlLabel(target)
+
+    if (!focusedEl) { return }
+
+    addClass(focusedEl, 'focused')
```
tyom: Yeah, we should. But this is based on existing Trade Elements. I think we should address that when we rebuild them in BEM.

src/javascripts/modules/label-select.js
```diff
@@ -0,0 +1,50 @@
+const { addClass, removeClass } = require('../../lib/element-stuff.js')
+
+function getControlLabel (control) {
+  if (control.type !== 'checkbox') { return }
+
+  const labelEl = control.parentElement
```
feedmypixel: Will we always be nesting checkboxes in labels? will they always be the direct parent? I'd be tempted to use a `js-` selector here so we don't have to rely on the makup structure

src/javascripts/modules/label-select.js
```diff
@@ -0,0 +1,50 @@
...
+
+  if (labelEl.className.indexOf('selection-button-checkbox') < 0) { return }
```
feedmypixel: As discussed we need a lib for a `hasClass`

feedmypixel: would ` control.querySelector('.selection-button-checkbox')` do for this function? Unless I'm missing something

src/javascripts/modules/label-select.js
```diff
@@ -0,0 +1,50 @@
...
+
+  return labelEl
+}
+
```
feedmypixel: I'm more of a fan of literal object or just functions in a file in comparison to the class method. I don't really feel we need a `class` or make use of it and if its objet literal or a list of functions is just preference. We should agree a way though and align anything in `/src/javascripts/modules` to that way

src/javascripts/modules/label-select.js
```diff
@@ -0,0 +1,50 @@
...
+const LabelSelect = {
+  init () {
+    this.bindEvents()
```
feedmypixel: Feels like a nice pattern is to check at the start for the relevant `js-` elements and return if they are not present

src/javascripts/modules/label-select.js
tyom: > would control.querySelector('.selection-button-checkbox') do for this function? 

That would only search descendants of that element.

src/javascripts/modules/label-select.js
tyom: I think we should [stop using ES6 classes](https://medium.com/javascript-scene/the-two-pillars-of-javascript-ee6f3281e7f3).

src/javascripts/modules/label-select.js
tyom: In this case we're relying on events. The benefit of not checking for existing elements is that they can be introduced after the function is executed and it would still work.

The handler are doing the necessary checks.

src/javascripts/modules/label-select.js
tyom: I'm not very happy about the whole implementation of custom checkboxes/radios. This change is attempting to retrofit missing functionality. Adding `js-` prefix would create and additional dependency which could be forgotten during authoring and break the component.

Perhaps a better approach would be to refactor this component to apply custom styles to JS only controls prefixed with `js-`. But that's outside of the scope of this change.

I would suggest we thoroughly review all our form controls in forthcoming refactoring sessions as we move Trade Elements into the code base.

feedmypixel: Agreed, sounds sensible to me

src/javascripts/modules/label-select.js
feedmypixel: > I think we should stop using ES6 classes.
+1

src/javascripts/modules/label-select.js
feedmypixel: Sorry I have misunderstood this. It feels like in the `onChangeHandler` you could just assign the `const labelEl` via a `js-hook` but then this would come under the refactor task. It feels like `getControlLabel` could potentially go with a refactor?

feedmypixel: Actually I see whats happening, your change and focus are `document` events.

src/javascripts/modules/label-select.js
feedmypixel: It feels a little like you are setting events and relying on the change handlers, rather than doing nothing if the elements are not present on the page. But then as you have said you essentially have event delegation running on the document. My question would be do you need this on all pages in the application?

feedmypixel: Also another potential benefit for a return if there are no elements, is that it communicates to the dev upfront - you need these elements to proceed.

tyom: I don't see why not. These handlers are only executed on `change` and `focus` globally. I'd say that since this custom checkbox is meant to replace default checkboxes it's fine to always expect them to appear anywhere in the document without having to worry to include the script first.

Essentially it's a global component.

feedmypixel:
As discussed lets work out a blueprint for the client-side JavaScript. I like this object literal approach. Lets keep an eye on if we need to build things as globals with a delegate event on the document (due to JavaScript placing elements on the page) or if we need to build things that just initialise on page load dependent on them being on the page.

LGTM 👍

PR #201, opened 2017-06-14
Investment requirements

tyom:
- [x] Investment requirements key/value table
- [x] Investment requirements form
  - [x] Ability to add basic data
  - [x] Conditional field and progressive disclosure for UK company to be invested in

requirements ng

### DIFF
--- a/config/nunjucks/filters.js
+++ b/config/nunjucks/filters.js
@@ -9,6 +9,7 @@ const {
   keys,
   values,
   flatten,
+  map,
 } = require('lodash')
 
 function isNotEmpty (value) {
@@ -19,10 +20,9 @@ const filters = {
   stringify: JSON.stringify,
 
   values,
-
   keys,
-
   flatten,
+  map,
 
   highlight: (string, searchTerm) => {
     const regEx = new RegExp(`(${searchTerm})`, 'gi')
--- a/src/controllers/investment/form.middleware.js
+++ b/src/controllers/investment/form.middleware.js
@@ -1,5 +1,5 @@
 const Q = require('q')
-const { get } = require('lodash')
+const { get, flatten } = require('lodash')
 const metadataRepo = require('../../repos/metadata.repo')
 const { getAdvisers } = require('../../repos/adviser.repo')
 const { isValidGuid } = require('../../lib/controller-utils')
@@ -8,6 +8,7 @@ const interactionFormattingService = require('../../services/interaction-formatt
 const {
   valueLabels,
   interactionsLabels,
+  requirementsLabels,
 } = require('./labels')
 const {
   createInvestmentInteraction,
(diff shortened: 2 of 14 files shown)


config/nunjucks/filters.js
```diff
@@ -20,6 +20,8 @@ const filters = {
 
   keys: _.keys,
 
+  map: _.map,
```
teneightfive: I think I've removed underscore as a default dependency and just used destructuring like above in another PR. Maybe rebase on develop to make sure you have the latest changes.

feedmypixel: Maybe not part of this PR but can we call this `lodash`? rather than `_` to align to the rest of the app

src/controllers/investment/form.middleware.js
```diff
@@ -227,11 +243,37 @@ function interactionDetailsFormPostMiddleware (req, res, next) {
...
+      res.locals.form.errors = err.error
+      res.locals.form.state = req.body
+
+      next()
+    } else {
+      next(err)
+    }
+  })
+}
+
```
feedmypixel: as you mentioned it is starting to feel like these could be split out into separate files or feature groupings. Something like: `/middleware/form/<name>.js` folder/filenames. Maybe easier as a different PR

tyom: I'd say a different PR would be more appropriate.

config/nunjucks/filters.js
```diff
@@ -19,10 +20,9 @@ const filters = {
   stringify: JSON.stringify,
 
   values,
-
   keys,
-
```
teneightfive: Any reason for removing the breathing space? Shouldn't we be consistent and remove the line between all methods in this file in that case?

tyom: I thought it looked odd with shorthanded properties.

```js
  stringify: JSON.stringify,

  values,
  
  keys,
  
  flatten,
  
  map,

  highlight: (string, searchTerm) => { }
```

Too much breathing space.

src/controllers/investment/shared.middleware.js
```diff
@@ -29,10 +29,11 @@ function getProjectDetails (req, res, next, id = req.params.id) {
   Q.spawn(function * () {
     try {
       const projectData = yield getInvestmentProjectSummary(req.session.token, req.params.id)
-      res.locals.valueData = yield getInvestmentValue(req.session.token, req.params.id)
+      const valueData = yield getInvestmentValue(req.session.token, req.params.id)
       res.locals.requirementsData = yield getInvestmentRequirements(req.session.token, req.params.id)
       res.locals.teamData = yield getInvestmentTeam(req.session.token, req.params.id)
       res.locals.projectData = projectData
+      res.locals.valueData = valueData
```
tyom: `valueData` is cached because it's used further down in `projectStatus` locals.

Issue #167, opened 2017-06-07
URL structure


teneightfive:
## Problem

The app currently uses a few variations for URLs that could follow a similar pattern. For example some URLs require a trailing slash, whereas other require no trailing slash. If the pattern doesn't match it will currently 404.

We also use inconsistent patterns for creating/viewing/editing. For example, some view URLs include the text `view` in path (`/company/view/foreign/:id`), others don't (`/investment/:id`).

## Solution

We should use a consistent pattern when referencing URLs so that when URLs are shared it is more clear what they are going to link to and it allows an expectation to be set when navigating the site and doing similar actions that the URL structure will be the same.

A [google sheet](https://docs.google.com/spreadsheets/d/1TXW2IKjKe6Ue0csBh4z7X0Gk6kvnwfd4n_afsV4sehY/edit#gid=0) (you will need access to the DIT Google Drive area) has been created to start to map the existing URLs to proposed future URLs and to have discussions around the proposals.

PR #205, opened 2017-06-15
Config structure

### LINKED ISSUE #81: Extract logger
## The problem

Logging is currently handled by [winston logger](https://github.com/winstonjs/winston) which is loaded in to files each time logging is required. The problem with this is any global configuration is lost. 

## Possible solution

Create a separate logger module which has generic configuration for winston and exposes its methods. This can then be included where logging is required and will persist the default configuration for the app.

teneightfive:
This restructures some of the configuration so that its easier to maintain:

- Move config to new config folder
- Move logger to own config
- Move routing to a single place
- Remove trailing slashes from URLs
- Log all requests to console transport

## Issues

This PR closes and starts to address some issues:

- closes #81 
- starts to address #78 and #74


src/controllers/my-account.controller.js
```diff
@@ -6,6 +6,6 @@ function myAccount (req, res) {
   res.render('myaccount/index')
 }
 
-router.get('/', myAccount)
+router.get('/myaccount', myAccount)
```
tyom: Could hyphenate.

src/controllers/support.controller.js
```diff
@@ -84,8 +84,10 @@ function thank (req, res) {
...
 }
 
-router.get('/bug', getBug)
-router.post('/bug', postBug)
-router.get('/thank', thank)
+router
+  .route('/support/bug')
+  .get(getBug)
+  .post(postBug)
+router.get('/support/thank', thank)
```
tyom: New line?

src/controllers/my-account.controller.js
teneightfive: Was going to do as a separate change. Relates to #167

src/controllers/support.controller.js
teneightfive: Even if there's only method?

tyom: It's a different route though. Just feels squashed like that.

Also I wonder if we should standardise the way we name terminal middleware (i.e. the ones that render page). In investment we use `{verb}Handler` and as well as `something{Verb}Handler` and `something{Verb}`.

Out of scope for this PR but something to think about.

teneightfive: Definitely thing we should use a better naming convention for handlers. Add a github issue?

tyom: [Added](https://github.com/uktrade/data-hub-fe-beta2/issues/208).

config/logger.js
```diff
@@ -0,0 +1,36 @@
...
+
+if (config.env === 'production') {
+  // TODO: add a transport to log errors in service like Sentry
+  // before we go live with real users
+  exceptionTransports.push(
+    new (winston.transports.Console)({
+      json: true,
+      timestamp: true,
+      colorize: false,
+      stringify: function stringify (obj) {
```
feedmypixel: could/should this be a boolean? [console-transport](https://github.com/winstonjs/winston/blob/master/docs/transports.md#console-transport)

config/logger.js
```diff
@@ -0,0 +1,36 @@
...
+  level: config.logLevel,
+  json: false,
+  colorize: true,
+  handleExceptions: true,
+}))
+
+if (config.env === 'production') {
+  // TODO: add a transport to log errors in service like Sentry
+  // before we go live with real users
+  exceptionTransports.push(
```
feedmypixel: does this need sending into the logger transports property on [#L31](https://github.com/uktrade/data-hub-fe-beta2/pull/205/files#diff-a9cb773d3e55b76418348b85705637c9R31) ?

config/routes/index.js
```diff
@@ -0,0 +1,34 @@
...
+  '../../src/controllers/company-interaction.controller',
+  '../../src/controllers/company-contact.controller',
+  '../../src/controllers/company-investments.controller',
+  '../../src/controllers/company-export.controller',
+  '../../src/controllers/company-add.controller',
+  '../../src/controllers/company-archive.controller',
+  '../../src/controllers/investment',
+]
+
+controllers.forEach((ctrl) => {
```
feedmypixel: 👍  a recursive directory read of files would mean we would have to do nothing :)

config/routes/index.js
```diff
@@ -0,0 +1,34 @@
...
+  '../../src/controllers/company-foreign.controller',
+  '../../src/controllers/company-ltd.controller',
+  '../../src/controllers/company-ukother.controller',
+  '../../src/controllers/company-interaction.controller',
+  '../../src/controllers/company-contact.controller',
+  '../../src/controllers/company-investments.controller',
+  '../../src/controllers/company-export.controller',
+  '../../src/controllers/company-add.controller',
+  '../../src/controllers/company-archive.controller',
+  '../../src/controllers/investment',
```
feedmypixel: maybe be explicit here with the filename?

config/logger.js
teneightfive: Good spot

config/routes/index.js
teneightfive: You mean `'../../src/controllers/investment/index'`? We don't use that pattern anywhere else that uses an index file at the moment though. I'm not sure it needs it.

- Nunjucks [example in server file](https://github.com/uktrade/data-hub-fe-beta2/blob/develop/src/server.js#L14)
-

config/routes/index.js
teneightfive: I think once we get to a point where most things are a _sub-app_ like investment we could just run through a directory and expect the same router format.

Right now we're loading routers in different ways. If we ran through everything file in the directory and sub directories now we'd be including all the investment controllers too whilst we only need 1 of those files at the moment.

If we have an expected structure in the future we could load that. Was thinking that each sub-app directory could contain a `router.js` file that we could load and would mean we don't have to define them each time we add one.

feedmypixel: sorry I wasn't serious with this I was just thinking how automated we could make it :) (should of put a :trollface: on it)

teneightfive: Haha, I know but I'd also been thinking the exact same thing :) 

But realised it wasn't possibly just yet...

I want to get to a place where there's very little of this type of setting up.

src/controllers/support.controller.js
feedmypixel: I quite like the pattern that is happening with `consts` of [singles at the top and multiples below](https://github.com/uktrade/data-hub-fe-beta2/blob/develop/src/controllers/investment/edit.controller.js#L3-L10)

Maybe this pattern could be followed through with routes?

feedmypixel:
LGTM ✅

PR #207, opened 2017-06-16
Remove government crest

teneightfive:
This change updates a few things around the icons used for the site. It also adds some better metadata to the base template for when the link is shared.

## Favicon

It removes the crown and replaces with the DIT crest.

favicon-192x192

## Opengraph/twitter/apple icon

It replaces the white crown on a black background with the DIT crest plus department name.

apple-touch-icon

## Logo

It also adds the crest in to the logo in the header and removes the DIT abbreviation favouring a visually hidden element containing the full department name.

screen shot 2017-06-22 at 15 22 04


src/middleware/check-secure.js
```diff
@@ -0,0 +1,60 @@
+/* eslint-disable no-param-reassign, no-restricted-syntax */
```
feedmypixel: whats this for? Should it be part of this PR? are we using Azure?

test/middleware/check-secure.test.js
```diff
@@ -0,0 +1,101 @@
+describe('Check secure middleware', () => {
```
feedmypixel: this feels like it should be in a separate PR. Also have you heard of [helmet](https://github.com/helmetjs/helmet)?

src/middleware/check-secure.js
feedmypixel: this feels like it should be in a separate PR. Also have you heard of [helmet](https://github.com/helmetjs/helmet)?

teneightfive: Yep, was thinking we might need to add helmet at some point too.

The purpose of this middleware though is to check if the site is being served securely or not 
and then set a property on the request object for future middleware.

That property is used as part of the locals that are set so that's why its part of this PR. Could split it out but its tied to the functionality being added.

tyom:
2^KB favicon 🤔 
Are you sure we can't optimise it further?

src/middleware/check-secure.js
ztolley: Didn't we already have force https?

And yeah, I don't see how this relates to the crown, should be it's own PR to argue over

teneightfive: This middleware doesn't force https. It checks whether we are using https and uses that value to determine how to set some of the values in the locals middleware. The check will be similar so we can possibly amend the force https middleware to use this value rather than do the check itself. 

I can do this as a separate change but this PR will depend on it.

teneightfive:
>Are you sure we can't optimise it further?

@tyom I can optimise now but I started to look at how we're optimising/minimising/revisioning assets in a separate PR. Would be good to run all assets through some sort of optimisation during the build step.

teneightfive:
>The icon it almost impossible to make out

@ztolley I agree that its not as clear as the crown but seeing as we shouldn't be using the crown at the moment this seemed like the best option. I haven't used the version that includes the border as this reduced the clarity even more.

Am open to a unique favicon for Data Hub but this could be a good interim step?

config/nunjucks/globals.js
```diff
@@ -2,5 +2,6 @@ module.exports = {
   siteTitle: 'DIT',
   serviceTitle: 'Data Hub',
   projectPhase: 'alpha',
+  description: 'Data Hub | Department for International Trade',
```
tyom: Maybe use a word like "by" instead of pipe in description. Better yet, provide an actual description of what Data Hub is (if we can summarise it in one sentence).

src/sass/objects/layout/_global-header.scss
```diff
@@ -21,28 +21,47 @@
...
+  background: url("../../assets/images/dit-crest__white.png") 6px 0 no-repeat $black;
+  background-size: contain;
+  border-left: 2px solid $white;
+  display: inline-block;
+  height: 30px;
+  margin-right: 8px;
+  width: 36px;
+  vertical-align: text-bottom;
+
+  @media(min-resolution: 192dpi) {
```
feedmypixel: can this not be done in the media mixin?

src/views/_layouts/datahub-base.njk
```diff
@@ -1,5 +1,7 @@
 {% extends "_layouts/dit-base.njk" %}
 
+{% set title = [title, 'DIT Data Hub'] | flatten | removeNilAndEmpty | join(' - ') %}
```
feedmypixel: feels like this title work could be abstracted? maybe something to discuss in the future?

teneightfive: Yeah. Its set here as it feels related only to Data hub. I thought about moving it to the global locals middleware but that gets called before the views so can't be combined with the join. 

It could be a filter but that feels overkill for now.

PR #209, opened 2017-06-16
Feature/investment interaction tests

feedmypixel:
This work covers:

- investment/interaction tests
- improvements to investment/interaction work

## Of Note
- I have also set the timezone for circleci to `Europe/London` due to this work involving some tests that work with dates


src/controllers/investment/form.middleware.js
```diff
@@ -169,32 +169,34 @@ function investmentValueFormPostMiddleware (req, res, next) {
...
-        contacts,
-        interactionTypes,
-      }
-
-      next()
-    } catch (error) {
-      next(error)
-    }
+      .then(next)
+      .then(resolve)
```
ztolley: Dont need to resolve change to test as discussed

const req = {
  params: {
    id: 1
  }
}

const res = {
  locals: {}
}

In the test
controller.method(req, res, () => {
  expect (res.locals).to.have.property(x)
})

src/lib/date/toISOString.js
```diff
@@ -1,10 +1,20 @@
-module.exports = function toISOString (day, month, year) {
+module.exports = function toISOString (year, month, day, hours = 12, mins = 0) {
```
ztolley: Bit confusing to have a date.js file and a date folder with this in. Should be one or the other

src/controllers/investment/interaction/create.controller.js
```diff
@@ -6,13 +7,14 @@ const {
...
   res.locals.title.unshift('Add interaction')
-  return res.render('investment/interactions/create')
+
+  return res.render('investment/interaction/create')
 }
 
-function postAddInteractionHandler (req, res, next) {
-  if (res.locals.form.errors) {
+function createPostInteractionHandler (req, res, next) {
+  if (size(res.locals.form.errors)) {
```
tyom: Would `!isEmpty` read better?

src/lib/date/toISOString.js
feedmypixel: this is for future splitting out of functionality

src/controllers/investment/interaction/edit.controller.js
```diff
@@ -9,13 +10,14 @@ const {
   interactionDetailsFormPostMiddleware,
 } = require('../form.middleware')
 
-function getEditInteractionHandler (req, res, next) {
+function editGetInteractionHandler (req, res, next) {
   res.locals.title.unshift('Edit interaction')
-  res.render('investment/interactions/edit')
+
```
teneightfive: We should decide on new line pattern. Feels like some of us prefer a bit of space between settings locals and calling the render method and others don't. We'll end up just polluting diffs if we change it to our preference each time we touch a method.

package.json
```diff
@@ -66,7 +66,7 @@
     "clean": "rm -rf build",
     "start": "node --use-strict src/server.js",
     "build": "npm run clean && npm run sass:compile && npm run js",
-    "test": "mocha test/unit/**/*.test.js",
+    "test": "mocha --recursive test/unit/",
```
feedmypixel: Had to change this back due to `/test/unit/controllers/investment/interactions` folders

PR #210, opened 2017-06-16
middleware to trim strings found in req.body

feedmypixel:
This work trims whitespace from strings found in `req.body` posts. 

Currently you cannot login when your username has trailing whitespace.

### DIFF
--- a/src/middleware/trimBodyParams.js
+++ b/src/middleware/trimBodyParams.js
@@ -1,0 +1,30 @@
+const { size } = require('lodash')
+
+/**
+ * Trim whitespace from req.body strings
+ * @param obj
+ * @returns {*}
+ */
+function trim (obj) {
+  for (let prop in obj) {
+    let value = obj[prop]
+
+    if (typeof value === 'string') {
+      obj[prop] = value.trim()
+    } else if (value instanceof Array) {
+      obj[prop] = trim(value)
+    }
+  }
+
+  return obj
+}
+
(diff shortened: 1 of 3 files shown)


src/middleware/trimBodyParams.js
```diff
@@ -0,0 +1,30 @@
+const { size } = require('lodash')
+
+/**
+ * Trim whitespace from req.body strings
+ * @param obj
+ * @returns {*}
+ */
+function trim (obj) {
+  for (let prop in obj) {
```
ztolley: use const

teneightfive:
Not sure we need our own middleware for this. If we start using [express validator](https://github.com/ctavan/express-validator) across the board for validation and sanitization it includes a [trim](https://github.com/ctavan/express-validator#sanitizer) method to do this.

Equally Joi also has a [trim](https://github.com/hapijs/joi/blob/master/API.md#stringtrim) method and we always include this as a dependency.

feedmypixel:
> Equally `Joi` also has a trim method and we always include this as a dependency.

I initially started with `Joi` but getting `Joi` to run globally and locally on a controller didn't work. Which then made me think if `Joi` is the right thing for this. Maybe this is something that can go in now and we can refactor/remove when 

> If we start using express validator across the board for validation and sanitization it includes a trim method to do this.

feedmypixel:
closing this in favour of adding trim on the login controller with `Joi`

PR #211, opened 2017-06-16
e2e testing using Nightwatch and Cucumber

tyom:
This is an experiment to add e2e testing alongside with unit tests using Nightwatch.js writen in Cucumber.

The following new ENV variables are required:
```
QA_USER_EMAIL=
QA_USER_PASSWORD=
QA_SELENIUM_PORT=4444
QA_SELENIUM_HOST=127.0.0.1
QA_HOST=http://localhost:3000
```

`QA_USER_EMAIL` can be set to `cmuuser1test@mobile.ukti.gov.uk`
`QA_HOST` can be pointed to `https://rhod.datahub.dev.uktrade.io`

The password is usual.


test/acceptance/page_objects/Company.js
```diff
@@ -0,0 +1,143 @@
...
+    {
+      findCompany (name) {
+        return this
+          .setValue('#search', name)
+          .submitForm('@searchForm')
+      },
+
+      createForeignOrg (name) {
+        this.newlyCreatedCompanyName = name
+        return this
```
feedmypixel: can this be split out into 3 separate journeys? rather than using comments to show the separation?

##### Pseudo code:

```
return this
   . createForeignOrgStep1()
   . createForeignOrgStep2()
   . createForeignOrgStep3()
```

package.json
```diff
@@ -95,15 +96,21 @@
   "author": "Zac Tolley <zac@thetolleys.com>",
   "license": "MIT",
   "engines": {
-    "node": "6.10.2"
+    "node": "8.1.0"
```
tyom: Doing now.

.gitignore
```diff
@@ -27,11 +27,15 @@ node_modules/
 test.sh
 apitest.js
 reports
+.idea/
```
feedmypixel: I tried this a while ago. The general consensus was things like this should be in your own global git ignore. This is for project work only

test/acceptance/page_objects/Company.js
tyom: But they are never used separately. Why separate in three separate steps now if we don't need it? Besides, that'd just pollute the `commands` in page objects file.

test/acceptance/features/step_definitions/create-a-company.js
```diff
@@ -0,0 +1,50 @@
+const shortId = require('shortid')
+const { client } = require('nightwatch-cucumber')
+const { defineSupportCode } = require('cucumber')
+
+defineSupportCode(({ Given, Then, When }) => {
+  const Company = client.page.Company()
+
+  Given(/^I am an authenticated user on Data Hub website$/, async () => {
+    await client.page.Login().navigate().logIn()
```
tyom: Erm.. Not sure. I read it as single statement saying **log me in**.

test/acceptance/features/step_definitions/create-a-company.js
```diff
@@ -0,0 +1,50 @@
+const shortId = require('shortid')
+const { client } = require('nightwatch-cucumber')
+const { defineSupportCode } = require('cucumber')
+
+defineSupportCode(({ Given, Then, When }) => {
```
tyom: Why?

.gitignore
tyom: They are in mine. But I added it as I was on-boarding Venu in our repo. I didn't want to figure out how to add it to global gitignore on Windows so I felt it was easier and safer to just add it here.

But to be honest, following that consensus, half of that git ignore file should be in one's globals.

feedmypixel: > But to be honest, following that consensus, half of that git ignore file should be in one's globals.
Maybe that needs doing as a separate PR?

test/acceptance/page_objects/Company.js
feedmypixel: > Besides, that'd just pollute the commands in page objects file.

Agreed, that sounds like a bad plan

> Why separate in three separate steps now if we don't need it? 

It was more to communicate via code instead of comments.

> But they are never used separately.

ok sounds overkill

.gitignore
tyom: Removing them? I don't see any harm on over specifying project's git ignore. I'd rather stuff doesn't get added to repo's history instead of relying on individual developers having their global git ignores in check.

feedmypixel:
LGTM 👍

PR #212, opened 2017-06-18
Feature/investment archive

ztolley:
Created a common archive entity form component based on updated designs that move the archive function into the body of the details page for investment project.

Migrated the company and contact screens to use the same component

Moved the archive component js into application bundle.

Not that the component does not handle being passed in a value as the current flow does not support this.  The next phase of this would be to add this feature in and migrate company pages to a shared controller and follow the investment pattern, which would make it easier to work with this component.

*Archive investment project*
archive-invest

*Archive company*
archive-company

*Archive contact*
archive-contact


src/javascripts/modules/archive-form.js
```diff
@@ -0,0 +1,187 @@
...
+  }
+
+  /**
+   * Called when the reason dropdown value is changed so that
+   * if can be decided if the other reason field should be displayed.
+   *
+   *
+   * @memberof ArchiveForm
+   */
+  reasonUpdate () {
```
teneightfive: This action can be handled by the [conditional subfield module](https://github.com/uktrade/data-hub-fe-beta2/blob/80b1ce61661f1dc98162dac938f1531f01efe19f/src/javascripts/modules/conditional-subfields.js).

src/views/_components/archive-form.njk
```diff
@@ -0,0 +1,42 @@
...
+  <input type="hidden" name="_csrf" value="{{ csrfToken }}">
+  <fieldset class="{% if error %}error{% endif %} js-hidden">
+    <legend class="visually-hidden" aria-visible="false">Archive reason</legend>
+    <div class="form-group {% if error %}form-group-error{% endif %}" id="archived_reason-wrapper">
+      <label class="form-label-bold" for="archived_reason">
+        {{ label }}
+        {% if error %}<span class="error-message js-errorMessage">{{ error }}</span>{% endif %}
+      </label>
+
+      <select id="archived_reason" class="form-control" name="archived_reason">
```
teneightfive: Consider changing this to a radio input if there are only 2 options. The [multiple choice component](https://github.com/uktrade/data-hub-fe-beta2/blob/80b1ce61661f1dc98162dac938f1531f01efe19f/src/views/_components/form-multiple-choice.njk) would give you the markup you need.

src/views/_components/archive-form.njk
```diff
@@ -0,0 +1,42 @@
...
+        <option value="">Select a value</option>
+        {% for option in options %}
+          <option value="{{ option }}">{{ option }}</option>
+        {% endfor %}
+        <option value="Other">Other</option>
+      </select>
+    </div>
+
+    <div class="form-group">
```
teneightfive: Do we not have a textbox component we can use here?

src/views/_components/archive-form.njk
```diff
@@ -0,0 +1,42 @@
...
+      <label class="form-label-bold" for="archived_reason_other">Other</label>
+      <input id="archived_reason_other" class="form-control" name="archived_reason_other" type="text">
+    </div>
+  </fieldset>
+  <div class="save-bar">
```
teneightfive: Is there not a component in trade elements for this?

src/views/company/_layout-view.njk
```diff
@@ -7,7 +7,14 @@
     <span class="post-header">{{ headingAddress }}</span>
   </h2>
 
-  {% include "./_archive.njk" %}
+    {% if company.archived %}
```
teneightfive: Should be dropped back by 2 spaces.

src/views/company/_layout-view.njk
```diff
@@ -7,7 +7,14 @@
...
+    <p class="infostrip">
+      This company has been archived on {{company.archived_on | formatDate}} by {{company.archived_by.first_name}} {{company.archived_by.last_name}}. <br>
```
teneightfive: `This company was archived on...`?

src/views/_components/archive-form.njk
```diff
@@ -0,0 +1,42 @@
...
+        {{ label }}
+        {% if error %}<span class="error-message js-errorMessage">{{ error }}</span>{% endif %}
+      </label>
+
+      <select id="archived_reason" class="form-control" name="archived_reason">
+        <option value="">Select a value</option>
+        {% for option in options %}
+          <option value="{{ option }}">{{ option }}</option>
+        {% endfor %}
+        <option value="Other">Other</option>
```
teneightfive: Should _other_ always be an option? Could this be determined by the properties passed in to the component?

src/views/company/details-foreign.njk
```diff
@@ -14,3 +14,17 @@
     <p class="footnote">You do not have perission to edit the Account management section</strong></p>
   </div>
 {% endblock %}
+
+{% if not company.archived %}
+  <h2 class="heading-medium">Archive company</h2>
+  <p>Archive this company if it is no longer required or active.</p>
+
+  {% component 'archive-form',
+    {
```
teneightfive: Does this need to be on a new line? I find it neater to do:

```
{% component 'component-name', {
  ...
} %}
```

src/javascripts/modules/archive-form.js
```diff
@@ -0,0 +1,187 @@
+/* eslint no-new: 0 */
```
tyom: Why do we use `new` if linter doesn't want us to?

ztolley: then I suggest we upate the linter to allow it if we want to use classes

src/views/company/_layout-view.njk
ztolley: Yup... This company was archived on 12th April 2017 13:25 by Zac Tolley. Makes sense no?

ztolley: Yeah that too

src/javascripts/modules/archive-form.js
tyom: I would suggest we [don't use ES6 classes](https://github.com/joshburgess/not-awesome-es6-classes).

src/views/_components/archive-form.njk
ztolley: was trying to remove dependency on trade elements. Though that would mean I should add a save bar to our new components.

teneightfive: I would stick with consistency for now so when we replace a component from trade elements its easier to see where it is used and replace in those locations.

Otherwise we'll have to hunt down similarities too.

src/views/_components/archive-form.njk
```diff
@@ -0,0 +1,34 @@
...
+ #
+ # @example
+ # {% component 'archive-form', {
+ #    label: 'Archive reason',
+ #    options: ['Reason 1', 'Reason 2'],
+ #    error: 'Provide a value'
+ #  } %}
+ #
+#}
+{% set radioOptions = options | arrayToLabelValues %}
```
feedmypixel: can this be done before the data is sent to the component?

config/nunjucks/filters.js
```diff
@@ -57,6 +57,28 @@ const filters = {
...
   },
+
+  pushToArray: (value, newItem) => {
+    if (Array.isArray(value)) {
+      value.push(newItem)
+    }
+    return value
+  },
+
+  arrayToLabelValues: (items) => {
```
feedmypixel: Can this be done in the controller layer?
Same as above it feels like you don't need to check if its an array. IMO only accept one thing then it will blow up when the right thing is not sent in, rather than guarding things in the methods.

src/controllers/investment/archive.controller.js
```diff
@@ -0,0 +1,54 @@
...
+const investmentRepository = require('../../repos/investment.repo')
+const { detailsGetHandler } = require('./details.controller')
+
+const router = express.Router()
+
+function archiveInvestmentProjectHandler (req, res, next) {
+  Q.spawn(function * () {
+    try {
+      // Archive the project.
+      const reason = (req.body.archived_reason === 'Other') ? req.body.archived_reason_other : req.body.archived_reason
```
feedmypixel: can this just be one property thats set on the body?

src/controllers/investment/archive.controller.js
```diff
@@ -0,0 +1,54 @@
...
+      yield investmentRepository.archiveInvestmentProject(req.session.token, req.params.id, reason)
```
feedmypixel: could this be a promise? and then render the page on success and render an error on fail?

src/controllers/investment/archive.controller.js
```diff
@@ -0,0 +1,54 @@
...
+
+      res.locals.projectData = Object.assign({}, res.locals.projectData, {
+        archived: true,
+        archived_reason: reason,
+        archived_by: req.session.user,
+        archived_on: new Date(),
+      })
+      next()
+    } catch (err) {
+      if (err.statusCode === 400) {
```
feedmypixel: this catch seems to be handling an error from `archiveInvestmentProject` and any `q.spawn`errors? If thats correct that feels like its doing too much.

feedmypixel: It would be worth looking over the `POST -> redirect GET` pattern that is being used around the new structure in Investments [interaction/edit.controller.js](https://github.com/uktrade/data-hub-fe-beta2/blob/develop/src/controllers/investment/interaction/edit.controller.js#L13-L39)

src/javascripts/modules/archive-form.js
```diff
@@ -0,0 +1,153 @@
...
+  /**
+   *
+   * Finds important elements for this component and caches them
+   * to make it easier to use them
+   */
+  cacheElements () {
+    this.archiveForm = this.wrapper.querySelector(this.selector)
+    this.document = (typeof document !== 'undefined') ? document : this.archiveForm.ownerDocument
+
+    this.reasonLabel = this.archiveForm.querySelector('[for="archived_reason"]')
```
feedmypixel: change to `archivedReasonLabel`> `reasonLabel` makes me think what reason?

src/javascripts/modules/archive-form.js
```diff
@@ -0,0 +1,153 @@
...
+    this.reasonInputElement = this.archiveForm.querySelector('[name=archived_reason]')
+    this.otherTextInput = this.archiveForm.querySelector('[name=archived_reason_other]')
```
feedmypixel: what does other mean? IS there a better name?

src/views/_components/archive-form.njk
ztolley: Ideally the archive component just wants a list of reasons to give. Translating the simple array to a list of label and value options is required to put this into the multiple choice component and not sure the thing calling the archive form should care about that. Same for adding the other option, it's not something the caller cares about but part of the implementation of the component.

ztolley:
screen shot 2017-06-22 at 15 43 41

ztolley:
@feedmypixel I was thinking after this card maybe I'd go back and talk to @teneightfive about changing the multiple choice card to see if it makes sense to allow it to show an 'other' option, but then the issue is the field names

config/nunjucks/filters.js
```diff
@@ -57,6 +57,22 @@ const filters = {
   formatNumber: (number, locales = 'en-GB') => {
     return number.toLocaleString(locales)
   },
+
+  pushToArray: (value, newItem) => {
+    value.push(newItem)
+    return value
```
teneightfive: Do you need to return the value? Is this for subsequent filters?

src/javascripts/modules/archive-form.js
```diff
@@ -0,0 +1,151 @@
...
+
+    // If the form contains an error, show it expanded so the user
+    // can see the error, otherwise hide it until asked to show it.
+    if (this.errorElement) {
+      this.showForm()
+    } else {
+      this.hideForm()
+    }
+
+    return this
```
teneightfive: Does it need to return anything?

src/javascripts/modules/archive-form.js
```diff
@@ -0,0 +1,151 @@
...
+   *
+   * Finds important elements for this component and caches them
+   * to make it easier to use them
+   */
+  cacheElements () {
+    this.archiveForm = this.wrapper.querySelector(this.selector)
+    this.document = (typeof document !== 'undefined') ? document : this.archiveForm.ownerDocument
+
+    this.archivedReasonLabel = this.archiveForm.querySelector('[for="archived_reason"]')
+    this.reasonInputElement = this.archiveForm.querySelector('[name=archived_reason]')
```
teneightfive: Not sure its a good idea to tie all these to the field name. What if the field name changes because the name changes in the api? 

It would mean updating this file as well.

src/sass/_deprecated/_shame.scss
```diff
@@ -276,3 +276,7 @@ select.form-control {
     box-shadow: 0 0 0 3px $focus-colour;
   }
 }
+
+fieldset.error {
+  margin-bottom: 20px;
```
teneightfive: We need [lobotomized owl](https://alistapart.com/article/axiomatic-css-and-lobotomized-owls). :)

PR #213, opened 2017-06-19
Upgrade to Node 8

tyom:


tyom:
> Do we need anything in the README saying which version we support?

Updated README.

PR #215, opened 2017-06-19
Refactor SSL middleware checks

teneightfive:
This change creates a new piece of middleware which adds the ability
to set a property on the request object that subsequent middlewares
can use and also a method which can enforce HTTPS.

This middleware is now only being called when the app is not
in a development mode.


src/middleware/sslify.js
```diff
@@ -0,0 +1,107 @@
...
+ * @param {Hash} [options]
+ * @return {Hash}
+ * @api private
+ */
+function applyOptions (options) {
+  const settings = {}
+  options = options || {}
+
+  for (const option in defaults) {
+    if ({}.hasOwnProperty.call(defaults, option)) {
```
tyom: Isn't this the same as doing the following?
```js
const settings = Object.assign(defaults, options)
```

teneightfive: Have removed the need for the whole `applyOptions` method.

teneightfive:
Being closed in favour of #217. We probably don't need to set a property on the request object after having a think about it.

PR #218, opened 2017-06-21
Use Webpack 3

tyom:
- Upgrade to Webpack 3
- Import JS files via main app loader file (slightly restructure modules to have consistent init method).


webpack.config.js
```diff
@@ -33,13 +21,22 @@ module.exports = {
...
   },
-  plugins: prod ? [
+  output: {
+    path: path.resolve(__dirname, 'build/javascripts'),
+    filename: '[name].bundle.js',
+  },
+  plugins: [],
+}
+
+if (isProd) {
```
feedmypixel: I wonder if this should be in the `/config/webpack` folder as:
- `webpack.config.dev.js`
- `webpack.config.prod.js`

and referred to from `webpack.config.js`

tyom: The config file is so small at the moment I think it might be overkill to split it into three files.

src/javascripts/app.js
```diff
@@ -4,7 +4,27 @@ const SortableTable = require('./modules/sortable-table')
 const DetailsList = require('./modules/details-list')
 const LabelSelect = require('./modules/label-select')
 
+const AddAnotherField = require('./_deprecated/add-another-field')
+const ArchiveForm = require('./_deprecated/archive-form')
+const CompanyAdd = require('./_deprecated/company-add')
+const CompanyEdit = require('./_deprecated/company-edit')
+const ContactEdit = require('./_deprecated/contact-edit')
+const ExpandableCard = require('./_deprecated/expandable-card')
+/* eslint no-new: 0 */
```
teneightfive: Don't think we need this at the moment.

webpack.config.js
teneightfive: I think its overkill for this change but would be worth considering splitting them out into individual files for different environments. Not convinced they should live in `/config` though...

PR #232, opened 2017-06-23
Ensure the form element exists before executing

teneightfive:


src/javascripts/modules/archive-form.js
```diff
@@ -14,6 +14,9 @@ const ArchiveForm = {
   init (wrapper = document) {
     this.wrapper = wrapper
     this.cacheElements()
+
+    if (!this.archiveForm) { return }
```
feedmypixel: I'd really like to do a basic className check in all `init` methods for the element that is used in frontend code. If here continue, otherwise don't.

PR #249, opened 2017-06-23
Move search feature into its own sub-app

teneightfive:
This continues on the folder structure work to move features
into their own express sub-apps to keep concerns neatly separated.

## Issues

Continues to address issue #74.


src/apps/search/router.js
```diff
@@ -0,0 +1,11 @@
+const router = require('express').Router()
+
+const searchController = require('./search.controller')
```
tyom: Could import those individually to save repeating `searchController`.

teneightfive:
>Other than the fact the support stuff is wrapped up in it, fine.

Yep, was waiting for the other PR to be merged. I've been creating a new branch for each app on top of the previous one so there is less chance of conflict. Then rebasing which normally automatically removes that commit and cleans up the history.

PR #250, opened 2017-06-23
Suggestion for textbox component

ztolley:
Starting look at how to refactor the components, this is a basic example as a talking point

### DIFF
--- a/assets/stylesheets/components/_textbox.scss
+++ b/assets/stylesheets/components/_textbox.scss
@@ -1,8 +1,14 @@
 @import "../generic/form-control";
 @import "../settings/colours";
+@import "../settings/layout";
 
 .textbox {
-  @extend %form-control;
+  margin-bottom: $form-vertical-spacing;
+  box-sizing: border-box;
+
+  &.has-error {
+    border-left: 2px solid $error-colour;
+  }
 }
 
 .textbox__label {
@@ -16,22 +22,22 @@
 .textbox__hint {
   color: $secondary-text-colour;
   display: block;
-  font-weight: normal;
+  font-weight: 400;
 }
 
 .textbox__input {
   border: 2px solid $black;
   display: block;
-  font-weight: 400px;
+  font-weight: 400;
   padding: 5px 4px 4px;
   width: 100%;
-  max-width: 540px;
+  max-width: $max-field-width;
 
   &:focus {
     outline: 3px solid $yellow;
   }
 
-  .has-error & {
-    border: 2px solid $error-colour;
+  .has-error {
+    border-color: $error-colour;
   }
 }
--- a/assets/stylesheets/components/_textbox.scss
+++ b/assets/stylesheets/components/_textbox.scss
@@ -1,0 +1,37 @@
+@import "../generic/form-control";
+@import "../settings/colours";
+
+.textbox {
+  @extend %form-control;
+}
+
+.textbox__label {
+  color: $black;
+  display: block;
+  font-weight: 700;
+  line-height: 1.5;
+  text-transform: none;
+}
+
+.textbox__hint {
+  color: $secondary-text-colour;
+  display: block;
+  font-weight: normal;
+}
+
+.textbox__input {
+  border: 2px solid $black;
+  display: block;
+  font-weight: 400px;
+  padding: 5px 4px 4px;
+  width: 100%;
+  max-width: 540px;
+
+  &:focus {
+    outline: 3px solid $yellow;
+  }
+
+  .has-error & {
+    border: 2px solid $error-colour;
+  }
+}
--- a/assets/stylesheets/generic/_form-control.scss
+++ b/assets/stylesheets/generic/_form-control.scss
@@ -1,8 +1,0 @@
-%form-control {
-  margin-bottom: 2rem;
-  box-sizing: border-box;
-
-  &.has-error {
-    border-left: 2px solid $error-colour;
-  }
-}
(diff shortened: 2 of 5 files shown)


ztolley:
I want to use post css and set the component css file as a dependency, but thats another discussion.

assets/stylesheets/components/_textbox.scss
```diff
@@ -0,0 +1,37 @@
+@import "../generic/form-control";
+@import "../settings/colours";
+
+.textbox {
+  @extend %form-control;
```
tyom: Components should never rely on placeholder classes. They should just include necessary styles instead of grouping them with other external components (messing the order they are placed). gzip can handle repetition.

assets/stylesheets/generic/_form-control.scss
```diff
@@ -0,0 +1,8 @@
+%form-control {
```
tyom: Feels like this should be a normal object. We don't need placeholder selectors.

assets/stylesheets/components/_textbox.scss
ztolley: The reason wasn't to reduce repetition but rather reduce having to make the same change in multiple places. But an alternative would be to use variables

ztolley:
Looked at Googles Material Design web component library and was interesting to see how they went full BEM and have very similar requirements.

https://getmdl.io/components/index.html#textfields-section

assets/stylesheets/generic/_form-control.scss
ztolley: I started down that path but then you'd have a component compose of form-control and textfield. So, as said elsewhere, alternative is live with repetition but use variables.

assets/stylesheets/components/_textbox.scss
```diff
@@ -0,0 +1,37 @@
...
+.textbox__hint {
+  color: $secondary-text-colour;
+  display: block;
+  font-weight: normal;
+}
+
+.textbox__input {
+  border: 2px solid $black;
+  display: block;
+  font-weight: 400px;
```
teneightfive: Does it need the `px`?

assets/stylesheets/components/_textbox.scss
teneightfive: I agree with @tyom. We should avoid using `@extend` too.

assets/stylesheets/components/_textbox.scss
```diff
@@ -0,0 +1,37 @@
...
+  display: block;
+  font-weight: 700;
+  line-height: 1.5;
+  text-transform: none;
+}
+
+.textbox__hint {
+  color: $secondary-text-colour;
+  display: block;
+  font-weight: normal;
```
teneightfive: Above you're using numerical values for font-weight and here its text value. Maybe stick to one method.

assets/stylesheets/components/_textbox.scss
```diff
@@ -0,0 +1,37 @@
...
+}
+
+.textbox__input {
+  border: 2px solid $black;
+  display: block;
+  font-weight: 400px;
+  padding: 5px 4px 4px;
+  width: 100%;
+  max-width: 540px;
```
teneightfive: What is this width based on? If its a max width we want to set all form fields too it should live in settings.

assets/stylesheets/components/_textbox.scss
```diff
@@ -0,0 +1,37 @@
...
+
+  &:focus {
+    outline: 3px solid $yellow;
+  }
+
+  .has-error & {
+    border: 2px solid $error-colour;
```
teneightfive: Can just do:

```
border-color: $error-colour;
```

assets/stylesheets/generic/_form-control.scss
teneightfive: Placeholders also wouldn't live in `generic`. Probably better placed on `tools`.

src/views/_components/textbox.njk
```diff
@@ -0,0 +1,58 @@
+{##
+ # Textbox
+ #
+ # @param {string} [variant]  Type of text box (default|multiline)
```
teneightfive: I know these are copied from the example I gave but I now realise the JSDoc syntax was incorrect. Required params shouldn't have the square brackets.

src/views/_components/textbox.njk
```diff
@@ -0,0 +1,58 @@
...
+ #   {% component 'textbox', {
+ #     variant:  'multiline',
+ #     label: 'How old are you?',
+ #     name:  'age'
+ #   } %}
+ #}
+{% set hintId = 'hint-' + name %}
+
+<div class="textbox{{ ' has-error' if error }}">
+  <label for="input-{{ name }}" class="textbox__label">
```
teneightfive: Label could be a separate _block_.

src/views/_components/textbox.njk
```diff
@@ -0,0 +1,58 @@
...
+ # @example
+ #   {% component 'textbox', {
+ #     variant:  'multiline',
+ #     label: 'How old are you?',
+ #     name:  'age'
+ #   } %}
+ #}
+{% set hintId = 'hint-' + name %}
+
+<div class="textbox{{ ' has-error' if error }}">
```
teneightfive: I think it can be useful to prefix form components with a prefix `form-`. Thoughts?

assets/stylesheets/settings/_layout.scss
```diff
@@ -2,3 +2,5 @@
 
 $site-width: 960px;
 $gutter: $baseline-grid-unit * 8;
+$max-field-width: 400px;
```
teneightfive: Is it worth having a `_forms.scss` settings file for these things?

Issue #206, opened 2017-06-15
Asset revisioning


teneightfive:
## The problem

Our asset revisioning is using a simple time based state from when the server starts. This causes cached assets to be re-downloaded even if nothing has changed. 

## Solution

To prevent this we should revision assets based on their contents. There are [gulp plugins](https://www.npmjs.com/package/gulp-rev-all) or we could investigate using webpack to append a hash to the filename.

We'll also need to create a manifest mapping filenames to their new filename which a middleware will need to check when in production.

tyom:
Fixed in #251.

PR #251, opened 2017-06-23
Use Webpack to manage assets

tyom:
- Webpack loads Sass files
- Webpack loads fonts and images referenced in packages or `/assets/images` directory
- New `getAssetPath` function in global `locals` to get path to compiled asset
- Optimise images on build
- Refactor npm run scripts
- Integrate BrowserSync with Webpack
- Add asset revision hashes

Addresses #206


assets/stylesheets/application.scss
```diff
@@ -1,7 +1,9 @@
 $images-dir: "/images/";
 
-@import "trade-elements";
-@import "font-awesome";
+@import "../../node_modules/@uktrade/trade_elements/dist/sass/trade-elements";
```
teneightfive: Can we not also do `~@uktrade/trade_elements/...`?

assets/stylesheets/objects/layout/_global-header.scss
```diff
@@ -24,7 +24,7 @@
 }
 
 .global-header__crest {
-  background: url("../../assets/images/dit-crest__white.png") 6px 0 no-repeat $black;
+  background: url("../../../images/dit-crest__white.png") 6px 0 no-repeat $black;
```
teneightfive: This nesting seems quite deep. Shouldn't we have stylesheets and images at the same level? 

Something like this:
```
/assets
  /stylesheets
    application.css
  /images
    ...
```

package.json
```diff
@@ -1,11 +1,11 @@
 {
-  "name": "data-hub-frontend",
+  "name": "data-hub-fe-beta2",
```
teneightfive: Might want to rebase on top of develop as this reverts and earlier change from @feedmypixel

assets/stylesheets/objects/layout/_global-header.scss
tyom: They are like that, aren't they?

package.json
tyom: Hmm.. It is up to date with `develop`.

src/server.js
```diff
@@ -106,6 +106,7 @@ app.use(favicon(path.join(config.root, 'public/assets/images', 'favicon.ico')))
 app.use(express.static(path.join(config.root, 'public')))
 app.use('/javascripts', express.static(path.join(config.root, 'build/javascripts')))
 app.use('/css', express.static(path.join(config.root, 'build/css')))
+app.use('/images', express.static(path.join(config.root, 'build/images')))
```
teneightfive: Can we condense these 3 lines to be?

```
app.use('/assets', express.static(path.join(config.root, 'build')))
```

package.json
```diff
@@ -96,7 +95,8 @@
     "css:autoprefixer": "postcss build/css/*.css --use autoprefixer -d build/css",
     "sync": "browser-sync start --proxy http://localhost:3000 --files build/css/*.css build/javascripts/*.js",
     "circle": "npm run lint && nyc --reporter=lcov npm run test:unit -- --reporter mocha-circleci-reporter && npm run sass && npm run js && codeclimate-test-reporter < coverage/lcov.info",
-    "heroku-postbuild": "npm rebuild node-sass && npm run sass && npm run js"
+    "heroku-postbuild": "npm rebuild node-sass && npm run sass && npm run js",
+    "watch:test": "npm run test:unit -- -w"
```
teneightfive: Do we want to use this watch pattern for `client` and `js`, etc?

package.json
tyom: Somehow overwritten. I'll fix.

src/views/_layouts/dit-base.njk
```diff
@@ -1,4 +1,4 @@
-{%- set assetPath = asset_path | default('/') -%}
+{%- set publicPath = asset_path | default('/') -%}
```
teneightfive: Should we try and remove `asset_path` and `publicPath` all together and have 1 way to load assets?

Maybe a function on locals like `assetPath('filename.ext')` which checks the manifest rather than a filter?

assets/stylesheets/objects/layout/_global-header.scss
teneightfive: Not sure why we need to go back 3 directories then?

src/views/_layouts/dit-base.njk
tyom: `assetPath` is not the same as `publicPath`. One is Webpack the other is Express.

package.json
teneightfive: Maybe bin these changes. You probably amended it wrong during a conflict.

src/views/_layouts/dit-base.njk
tyom: Started with a function, then switched to filter. No I look at it and think that maybe function call is better.

assets/stylesheets/objects/layout/_global-header.scss
tyom: Because `_global-header` is in `stylesheets/objects/layout/`.

src/views/_layouts/dit-base.njk
teneightfive: I think we should load assets the same way though. In `datahub-base.njk` we're doing:

```njk
<script src="{{ 'ie.js' | asset }}"></script>
```

but in `dit-base.njk` we're doing:

```njk
<link href="{{ publicPath }}css/application.css" media="screen" rel="stylesheet">
```

teneightfive: I think there are also some references to js files in this file that may have been missed

assets/stylesheets/objects/layout/_global-header.scss
teneightfive: So this isn't looking for the compiled path?

tyom: No, this is taking source asset and Webpack then substitutes it with the compiled one.

tyom:
> Following on from what Ben said. I think that this is best done after we merge trade elements in 

Sure.

> and I'm not convinced there is a benefit to handling styles in webpack right now and the whole setting the whole app styles as a dependency in app.js feels off.
> Why don't we discuss styles in general after we;ve reorganised the code and are starting on components as it would be way nicer to use weback and have each component define it's style dependency.

The benefit of doing it in Webpack is to have single place to configure and manage asset dependencies. Hash them, configure them.

Going forward we may decide to reference styles from within individual components.

src/server.js
tyom: I removed `/assets` from the URLs. Felt it was unnecessary.

config/nunjucks/globals.js
```diff
@@ -1,6 +1,27 @@
...
+}
+
 module.exports = {
   serviceTitle: 'Data Hub',
   projectPhase: 'alpha',
   description: 'Data Hub is a customer relationship, project management and analytical tool for Department for International Trade.',
   feedbackLink: '/support/bug',
+
+  assetPath (asset) {
+    const assetsUrl = process.env.assetsUrl || this.ctx.BASE_URL || '/'
```
teneightfive: We should try and keep properties set on `process.env` in one place in the config file so its easier to see what env vars can be set. 

We should also use const case for environment variables.

config/nunjucks/globals.js
```diff
@@ -1,6 +1,27 @@
...
+  logger.error('Manifest file is not found. Ensure assets are built.')
+}
+
 module.exports = {
   serviceTitle: 'Data Hub',
   projectPhase: 'alpha',
   description: 'Data Hub is a customer relationship, project management and analytical tool for Department for International Trade.',
   feedbackLink: '/support/bug',
+
+  assetPath (asset) {
```
teneightfive: I'm not sure `globals.js` is the best place for this method. Wondering whether it should be its own middleware or set in the locals middleware?

config/nunjucks/globals.js
teneightfive: If we're adding new env vars we should explain them in the readme too.

src/server.js
```diff
@@ -28,6 +25,8 @@ const logger = require('../config/logger')
 
 const router = require('../config/routes')
 
+const { redisStore } = require('./store')
```
teneightfive: Move to config, like the logger and where we might end up having express config?

teneightfive: Making call it something more explicit like `session-store.js` too.

teneightfive: Feels like this would also be neater as a separate PR. This PR is snaking quite a bit now...

tyom: Moved out in #262.

config/nunjucks/globals.js
tyom: We could make a middleware to automatically match asset paths to those defined in manifest and return them if found but that means that all assets would need to be referenced by their chunk name, e.g. `app.js` instead of full path. And I feel that it might be a bit too magical. It might be better to grasp what's going on by having an explicit global Nunjucks function. At least for now.

teneightfive: I didn't mean make a magic middleware I just meant I don't think it should live as a nunjucks _global_. 

Feels better suited in `locals` middleware of a separate middleware that sets a method on `res.locals`?

config/nunjucks/globals.js
```diff
@@ -1,6 +1,28 @@
...
+    const assetsUrl = config.assetsHost || this.ctx.BASE_URL || '/'
```
teneightfive: Also because this is falling back on the `BASE_URL` value which is set in locals anyway. Maybe it would be better in the same location.

webpack.config.js
```diff
@@ -1,24 +1,78 @@
 const webpack = require('webpack')
+const merge = require('webpack-merge')
 const path = require('path')
+const ExtractTextPlugin = require('extract-text-webpack-plugin')
+const WebpackAssetsManifest = require('webpack-assets-manifest')
+const BrowserSyncPlugin = require('browser-sync-webpack-plugin')
 
 const isProd = process.env.NODE_ENV === 'production'
```
teneightfive: Use `!config.isDev`? Trying to reduce using `process.env` all over the app.

teneightfive: Use `!config.isDev`? Trying to reduce using `process.env` all over the app.

src/server.js
```diff
@@ -53,14 +53,12 @@ app.set('view engine', 'njk')
...
-app.use(favicon(path.join(config.root, 'public/assets/images', 'favicon.ico')))
+app.use(favicon(path.join(config.root, 'public/images', 'favicon.ico')))
 app.use(express.static(path.join(config.root, 'public')))
 app.use('/javascripts', express.static(path.join(config.root, 'build/javascripts')))
 app.use('/css', express.static(path.join(config.root, 'build/css')))
-
-app.use('/images', express.static(`${__dirname}/../assets/images/_deprecated/`))
-
-app.use('/fonts', express.static(`${__dirname}/../node_modules/font-awesome/fonts`))
+app.use('/images', express.static(path.join(config.root, 'build/images')))
```
teneightfive: Wondering whether `build` should be a dot folder (`.build`)? Might be easier to see it contains generated assets and keeps the output folder at the top of the list in most IDEs.

webpack.config.js
```diff
@@ -27,17 +81,38 @@ const webpackConfig = {
...
+      ],
+    }, {
+      reload: false,
+    }),
+  ],
+})
 
-  webpackConfig.plugins.push(
+const prod = merge.smart(common, {
+  devtool: 'hidden-source-map',
```
teneightfive: Seems to still be outputting sourcemap for CSS in production build mode

package.json
tyom: Not sure what you mean.

config/index.js
```diff
@@ -40,3 +41,7 @@ module.exports = {
   longDateFormat: 'D MMMM YYYY',
   mediumDateFormat: 'D MMM YYYY',
 }
+
+config.buildDir = path.join(config.root, '.build')
```
teneightfive: What about moving the root out to a var so this can still defined in the object?

```js
const root = path.normalize(`${__dirname}/..`)

const config = {
  root,
  ...
  buildDir: path.join(root, '.build'),
}
```

nodemon.json
```diff
@@ -0,0 +1,13 @@
+{
+  "restartable": "rs",
+  "watch": [
+    "src/**/*.js"
+  ],
+  "ignore": [
+    "build/"
```
teneightfive: `".build/"`?

src/server.js
```diff
@@ -53,14 +53,12 @@ app.set('view engine', 'njk')
...
-app.use('/javascripts', express.static(path.join(config.root, 'build/javascripts')))
-app.use('/css', express.static(path.join(config.root, 'build/css')))
-
-app.use('/images', express.static(`${__dirname}/../assets/images/_deprecated/`))
-
-app.use('/fonts', express.static(`${__dirname}/../node_modules/font-awesome/fonts`))
+app.use('/js', express.static(path.join(config.buildDir, 'js')))
+app.use('/css', express.static(path.join(config.buildDir, 'css')))
+app.use('/images', express.static(path.join(config.buildDir, 'images')))
+app.use('/fonts', express.static(path.join(config.buildDir, 'node_modules/font-awesome/fonts')))
```
teneightfive: What are we even still using fontawesome for?!

Might be another PR to see if we can remove the need for it...

webpack.config.js
```diff
@@ -53,7 +130,12 @@ if (isProd) {
...
       dead_code: true,
     }),
-  )
-}
+    new ExtractTextPlugin('css/[name].[chunkhash:8].css'),
+  ],
+})
 
-module.exports = webpackConfig
+if (config.isDev) {
```
teneightfive: Wondering if your original way around was better? 

Should the default be prod or develop? To get the develop config this setup would rely on `NODE_ENV=development` being specifically set.

config/index.js
tyom: Good idea. Much better!

src/server.js
tyom: I agree. It's used for dropdown icons and maybe something else. We should get rid of it. Perhaps move to SVG icons.

webpack.config.js
tyom: I've another condig prop with explicit checking of prod.

config/index.js
```diff
@@ -1,6 +1,7 @@
 const path = require('path')
 
 const isDev = (process.env.NODE_ENV !== 'production')
+const isProd = (process.env.NODE_ENV === 'production')
```
teneightfive: This can probably just be defined inline in the property definition.

The `isDev` variable is used elsewhere which is why it's defined outside the object.

src/middleware/locals.js
```diff
@@ -4,20 +4,40 @@ const config = require('../../config')
 
 const startTime = new Date().getTime()
 
+let webpackManifest = {}
```
feedmypixel: does this need a value assigned to it?

src/middleware/locals.js
```diff
@@ -4,20 +4,40 @@ const config = require('../../config')
...
+try {
+  webpackManifest = require(`${config.buildDir}/manifest.json`)
+} catch (err) {
+  logger.error('Manifest file is not found. Ensure assets are built.')
```
feedmypixel: Is having no manifest a valid option? does this need to return?

src/middleware/locals.js
```diff
@@ -4,20 +4,40 @@ const config = require('../../config')
...
+    startTime: startTime,
+    referer: req.headers.referer,
+    env: config.env,
+    googleTagManagerKey: config.googleTagManagerKey,
+    query: req.query,
+    currentPath: req.path,
+    BASE_URL: baseUrl,
+    CANONICAL_URL: baseUrl + req.originalUrl,
+
+    getAssetPath (asset) {
```
feedmypixel: going forward maybe this could be moved into its own file in a `src/middleware/locals` folder?

webpack.config.js
```diff
@@ -1,24 +1,78 @@
...
+      },
+      {
+        test: /\.scss$/,
+        use: ExtractTextPlugin.extract({
+          fallback: 'style-loader',
+          use: [
+            {
+              loader: 'css-loader',
+              options: {
+                sourceMap: config.isDev,
```
feedmypixel: do we need both of these in config? Can we just negate `config.isProd`? potentially for a different PR

webpack.config.js
```diff
@@ -27,17 +81,38 @@ const webpackConfig = {
...
-if (isProd) {
-  webpackConfig.devtool = 'hidden-source-map'
+const develop = merge.smart(common, {
+  output: {
+    filename: 'js/[name].js',
+  },
+  plugins: [
+    new ExtractTextPlugin('css/[name].css'),
+    new BrowserSyncPlugin({
+      port: 3001,
```
feedmypixel: should this be referenced from config?

webpack.config.js
```diff
@@ -27,17 +81,38 @@ const webpackConfig = {
...
+      proxy: 'http://localhost:3000',
```
feedmypixel: should this be referenced from config?

webpack.config.js
```diff
@@ -50,10 +125,15 @@ if (isProd) {
...
     }),
-  )
-}
+    new ExtractTextPlugin('css/[name].[chunkhash:8].css'),
+  ],
+})
 
-module.exports = webpackConfig
+if (config.isProd) {
+  module.exports = prod
```
feedmypixel: sat here by themselves `prod` and `develop` feel like they need a little more explanation. Maybe `prodConfig` and `devConfig`?

config/index.js
```diff
@@ -30,6 +33,7 @@ module.exports = {
     // 2 hour timeout
     ttl: process.env.SESSION_TTL || (2 * 60 * 60 * 1000),
   },
+  assetsHost: process.env.ASSETS_HOST,
```
tyom: I'm getting the base URL in locals middleware where there's special logic to obtain it.

src/middleware/locals.js
tyom: Makes it easier to deal with missing property when we do `webpackManifest[asset]` in `getAssetPath`. Just returns `undefined` instead of throwing an error.

src/middleware/locals.js
tyom: It is a valid option if you wish to serve static files from other Express static locations (i.e. `public/`)

webpack.config.js
tyom: I find it easier to read and the logic for them is quite different.

webpack.config.js
tyom: Nah. This is BrowserSync spefic and is only used in this one place.

webpack.config.js
tyom: Hmm.. You are in a file called `webpack.config.js`. Feel like that's already setting the scope for what this is.

PR #252, opened 2017-06-23
Move service deliveries into its own sub-app

teneightfive:
Continuing with the sub-apps for features work, this moves service
deliveries and its concerns into one folder.

## Issues

Continues to address #74.


src/apps/service-deliveries/router.js
```diff
@@ -0,0 +1,12 @@
+const router = require('express').Router()
+
+const serviceDeliveriesController = require('./service-deliveries.controller')
+
+router.get('/servicedelivery/:serviceDeliveryId/*', serviceDeliveriesController.getCommon)
+router.get(['/servicedelivery/:serviceDeliveryId/edit', '/servicedelivery/edit/'], serviceDeliveriesController.getServiceDeliveryEdit)
```
feedmypixel: putting some newlines in may make this easier on the eye?

PR #256, opened 2017-06-24
Move api.controller to feature apps

tyom:
Addresses #74 and #78.


src/apps/api/router.js
```diff
@@ -0,0 +1,11 @@
+const router = require('express').Router()
+const { postcodeLookupHandler } = require('./api.controller')
+
+router
+  .route('/postcodelookup/:postcode')
+  .get(postcodeLookupHandler)
```
ztolley: I like this but then isn't this different from the syntax used in PR 255,

router.get('/myaccount', getHandler)

Can we pick one style and all stick with it?

ztolley: nvm  I see we use a mix of the two, especially in contacts

teneightfive: The pattern I stick to is if there's only one method per route, use shorthand:

```js
router.get('/route', handler)
```

but if there's multiple methods for a route, use this syntax:

```js
router
  .route('/route')
  .get(getHandler)
  .post(postHandler)
```

Based on that style, we should probably use shorthand here.

ztolley: Yup I'm following that

tyom: Updated.

PR #258, opened 2017-06-24
Fix broken archive form as a result of merge with trade elements.

ztolley:
New style components don't seem to be able to call macros for the components merged from trade elements.

As no other new components use those macros it seems simpler to just not call them


teneightfive:
Could also just import the macro in this component. That's the difference with the extension way too - you can include components within components easily.

PR #260, opened 2017-06-24
Move interactions into a sub-app

teneightfive:
Continuing with the folder structure work this moves all of the
interactions work into a sub folder.

## Issues

Addresses #74 and #78.


src/apps/interactions/controllers/index.js
```diff
@@ -0,0 +1,7 @@
+const interactions = require('./controllers/interactions.controller')
```
tyom: Shouldn't this be `require('./interactions.controller')`? I noticed it in other feature apps.

src/apps/interactions/controllers/index.js
```diff
@@ -0,0 +1,7 @@
...
+const edit = require('./controllers/edit.controller')
+
+module.exports = {
+  interactions,
```
tyom: Wonder if this should be called `add` and the details and common should go in its own controller file.

src/apps/interactions/router.js
```diff
@@ -0,0 +1,17 @@
+const router = require('express').Router()
+
+const detailsController = require('./controllers/details.controller')
+const editController = require('./controllers/edit.controller')
+
+router.get('/interaction/add-step-1/', detailsController.getAddStep1)
+router.post('/interaction/add-step-1/', detailsController.postAddStep1)
```
tyom: These (and the ones below) could be grouped under `router.route` method.

src/apps/interactions/router.js
```diff
@@ -0,0 +1,17 @@
...
+router.get('/interaction/:interactionId/*', detailsController.getCommon)
```
tyom: Maybe use `router.param` to fetch interaction details by id.

src/apps/interactions/router.js
```diff
@@ -0,0 +1,17 @@
...
+router.get('/interaction/:interactionId/details', detailsController.getInteractionDetails)
+
+router.get(['/interaction/add/', '/interaction/:interactionId/edit', '/interaction/edit/'], editController.editDetails)
+router.post(['/interaction/add/', '/interaction/:interactionId/edit', '/interaction/edit/'], editController.postDetails)
+
+module.exports = {
+  // path: '/contact',
```
tyom: Should it be mounted under `/interactions`?

src/apps/interactions/router.js
```diff
@@ -0,0 +1,27 @@
+const router = require('express').Router()
+
+const detailsController = require('./controllers/details.controller')
+const editController = require('./controllers/edit.controller')
+
+router.param('interactionId', detailsController.getCommon)
+
+router
```
ztolley: personal preference, but I like 

router.route('/company/edit/ltd/:id')
  .get(foreignController.editCommon, ltdController.editDetails)

I like having the methods indented.

src/apps/interactions/router.js
```diff
@@ -0,0 +1,27 @@
...
+  .route('/add-step-1/')
+  .get(detailsController.getAddStep1)
+  .post(detailsController.postAddStep1)
+
+router
+  .route([
```
ztolley: bit ugly

teneightfive: Why? That's also not a very useful statement 🙄

src/apps/interactions/router.js
teneightfive: We haven't used this pattern elsewhere. I'm sticking with the current pattern.

If we decide to switch we can do it across the board.

teneightfive:
There are a few things within the interactions app that could be refactored/tidied up but I don't want to bundle it with this change and let the PR snowball.

We may also feel that interactions is best suited within company in the same style we have for investment and contacts.

PR #265, opened 2017-06-25
Move investment projects work into sub-app

teneightfive:
This moves all of the investment work into one feature app. This
builds on top of all the other work to move code into feature
groupings.

## Issues

This addresses #74 and #78.


src/apps/investment-projects/controllers/archive.controller.js
```diff
@@ -1,11 +1,6 @@
-/* eslint new-cap: 0 */
-const express = require('express')
 const { get } = require('lodash')
-const { getProjectDetails } = require('./shared.middleware')
-const investmentRepository = require('../../repos/investment.repo')
-const { detailsGetHandler } = require('./details.controller')
 
-const router = express.Router()
+const investmentRepository = require('../investment-projects.repo')
```
ztolley: I'm wondering if, when it's below an /app, it should just be repository, there's only ever one.

/apps/investment/repository.js

teneightfive: >I'm wondering if, when it's below an /app, it should just be repository, there's only ever one.

I was thinking the exact same thing. But sticking to the pattern we have for now.

I think once they've all in feature apps we can review naming conventions. I think it would be useful to have naming conventions for middleware/controller methods that are exported too.

Currently some of the older apps have middleware bundled in to controllers so we'll probably want to split those out too.

PR #267, opened 2017-06-26
Move remaining repos/services

teneightfive:
This moves the leftover services and repos into either a feature
app location of a general library.

It also removes services that no longer seem to have a use.

## Issues

This addresses #74 and #78.


src/apps/api/api.controller.js
```diff
@@ -1,5 +1,5 @@
 const { lookupAddress } = require('./api.service')
-const metadata = require('../../repos/metadata.repo')
+const metadata = require('../../lib/metadata')
```
tyom: `lib` 🥇

teneightfive: It's temporary 😳

PR #271, opened 2017-06-26
Cleanup dead code

teneightfive:
There are a few libraries and modules that are no longer being used
across the app so this change removes those to clean up the
codebase.


src/apps/companies/controllers/add.js
```diff
@@ -62,10 +63,10 @@ function postAddStepOne (req, res, next) {
   }
 
   if (req.body.business_type === 'ukother' || req.body.business_type === 'foreign') {
-    return res.redirect(`/company/add/${req.body.business_type}?${toQueryString(params)}`)
+    return res.redirect(`/company/add/${req.body.business_type}${buildQueryString(params)}`)
```
tyom: `/company/add/${req.body.business_type + buildQueryString(params)}`?

src/apps/companies/controllers/add.js
```diff
@@ -62,10 +63,10 @@ function postAddStepOne (req, res, next) {
...
   }
 
-  return res.redirect(`/company/add-step-2/?${toQueryString(params)}`)
+  return res.redirect(`/company/add-step-2/${buildQueryString(params)}`)
```
tyom: Maybe cache it to avoid repeat calls?

src/lib/text-formatting.js
```diff
@@ -46,4 +30,4 @@ function getContactLink (object) {
   return null
 }
 
-module.exports = { titleCase, newlineToBr, getContactName, getContactLink }
+module.exports = { newlineToBr, getContactName, getContactLink }
```
tyom: Line per member.

PR #272, opened 2017-06-27
Refactor routing

tyom:
- create main app routers automatically from module directories in `apps`/
- Rename controllers, services and repos in `apps` to be more consistent and easier to read

The whole sub app can be defined in single `index.js`file. If some of the features need to be split 
out into additional files they are added as plurals of type (e.g. `services.js`, controllers.js` etc) 
which in turn an be split further by creating a directory with that plural name and new file names 
with the file type concerns (e.g. `controllers/details.js`).

The type extensions have been dropped as they are becoming less useful in new structure. The files are becoming more predictable to find. Each directory can have a router as on of the exports in its index file or a separate file. So when importing these files one can:

```js
const { router } = require('./apps/account') // from index file
const { router } = require('./apps/api') // from router.js via index
const router = require('./apps/api/router') // from router.js
const companiesControllers = require('./apps/companies/controllers')
const { addController } = require('./apps/companies/controllers')
```

## Issues

Addresses #74.


tyom:
> Looks fine and I like the expansion theory, only comment is that an entity only ever has 1 repository so I'd call it repository and not repos

And when you want to add another property you'd have to rename that file and update it in multiple places. Why not agree to always treat them as containers for multiple properties with at least one?

Issue #226, opened 2017-06-22
Remove Trade Elements dependency


feedmypixel:
Once we have brought anything that is used in Data Hub in from Trade elements e need to remove the Trade Elements dependency

ztolley:
Discussed with Ben.

Ben will work on this card with the view to move the trade elements code into datahub and make it work as is.

A separate list of cards will be raised for each component to convert the macro syntax to the new component syntax.

PR #283, opened 2017-06-28
Remove unused SASS and GOV mixins

feedmypixel:
- Remove as many mixins provided by gov as possible,
- Remove any css that is not being used by JavaScript or view files

## Of Note
The only GOV mixin I haven't managed to remove is `button`
This removes a lot of SASS work! But I am pretty confident from using a number of scripts that there shouldn't be too many styling regressions from this work. If there they can be fixed as and when.


feedmypixel:
work associated with https://github.com/uktrade/data-hub-frontend/issues/226

assets/stylesheets/_deprecated/components/_pagination.scss
```diff
@@ -1,32 +0,0 @@
-.pagination {
```
teneightfive: There's a comment in [paginvation-nav](https://github.com/uktrade/data-hub-frontend/blob/develop/assets/stylesheets/components/_pagination-nav.scss#L5) to remove some declarations when this is gone. Might be worth checking.

assets/stylesheets/_deprecated/components/_searchbar.scss
```diff
@@ -20,7 +19,7 @@
 
   .searchbar__input {
     box-sizing: border-box;
-    font-size: em(19);
+    font-size: 19px;
```
teneightfive: Do we want to use the font mixin?

feedmypixel: I was thinking of using [postcss-pxtorem](https://github.com/cuth/postcss-pxtorem) thoughts?

Issue #242, opened 2017-06-23
Date entry macro refactor


ztolley:
Refactor the macro from the old Trade Elements style to the newer data hub component style

jim68000:
Is this just a matter of porting stuff to the new component tag?

ztolley:
@tyom can give you an overview of the differences between the old macro style I used and the component tools he created

ztolley:
We need to talk about the impact of moving from the css naming scheme used by GDS to BEM, the two are completely different and would require re-implementation of GDS styles

teneightfive:
@jim68000 there are a few differences with using macros and a custom extension.

- Using the custom extension offers a clearer syntax from looking at views.

  It's easier to see this is a component:
  ```njk
    {% component 'table' %}
  ```

  than:

  ```njk
  {{ table }}
  ```

  Macros syntax is the same as calling variables and methods.

- Using macros [requires importing](https://github.com/uktrade/data-hub-frontend/blob/38cb234c84c17e4d9ca6448570901c4ff49ff8a6/src/views/_layouts/datahub-base.njk#L4-L12) the macros file either globally in the template or in a macro that also references another macro. 

  With a custom extension they tag can be called in any layout, view, include or another component without having to load anything else.

PR #285, opened 2017-06-29
Form components

tyom:
This adds new composable Nunjucks macros for form elements as well as new component styles. New styles do not rely on external dependencies further improving the cascade and number of rules defined for each element.

form-components

[Demo](https://datahub-beta2-pr-285.herokuapp.com/components/form)

Addresses #226, #242.

### DIFF
--- a/assets/stylesheets/_layout.scss
+++ b/assets/stylesheets/_layout.scss
@@ -1,0 +1,7 @@
+@import "layout/global-header";
+@import "layout/global-footer";
+@import "layout/global-nav";
+@import "layout/main-content";
+@import "layout/notification-banner";
+@import "layout/phase-banner";
+@import "layout/skip-links";
--- a/assets/stylesheets/_layout.scss
+++ b/assets/stylesheets/_layout.scss
@@ -1,7 +1,1 @@
-@import "layout/global-header";
-@import "layout/global-footer";
-@import "layout/global-nav";
-@import "layout/main-content";
-@import "layout/notification-banner";
-@import "layout/phase-banner";
-@import "layout/skip-links";
+@import "layout/assemblies";
--- a/assets/stylesheets/components/form/_error-summary.scss
+++ b/assets/stylesheets/components/form/_error-summary.scss
@@ -1,0 +1,26 @@
+@import "../../settings";
+
+.c-error-summary {
+  border: 5px solid $error-colour;
+  padding: $default-spacing-unit;
+  max-width: 40em;
+}
+
+.c-error-summary__heading {
+  margin-top: 0;
+}
+
+.c-error-summary__list {
+  font-weight: 600;
+  list-style: square;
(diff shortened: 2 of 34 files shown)


assets/stylesheets/components/form/_error-summary.scss
```diff
@@ -0,0 +1,26 @@
+@import "../../settings";
+
+.c-error-summary {
+  border: 5px solid $error-colour;
+  padding: 20px;
+  max-width: 40em;
+
+  > .heading-medium {
```
teneightfive: Aren't these legacy class names? Maybe see if we can remove their need.

assets/stylesheets/_layout.scss
```diff
@@ -0,0 +1,9 @@
+@import "layout/global-header";
```
teneightfive: Not sure moving these to just `layout` feels right. They feel almost like self-contained blocks (or components) and not tied to global layout.

assets/stylesheets/components/form/_error-summary.scss
```diff
@@ -0,0 +1,26 @@
...
+  font-weight: 600;
+  list-style: square;
+  padding-left: 20px;
+  color: $error-colour;
+
+  a {
+    color: $error-colour;
+  }
+
+  li {
```
teneightfive: If using BEM should this not be `.c-error-summary__list-item`?

src/apps/auth/views/sign-in.njk
```diff
@@ -1,36 +1,32 @@
...
-        {% for message in messages.success %}
-          {{ message }}<br>
-        {% endfor %}
-      </p>
-    </div>
-  {% endif %}
-
-  {{ errorPanel( messages.formErrors ) }}
+  {% call Form(formData) %}
+    {{ flashMessage('success', messages) if messages.success|length }}
```
teneightfive: Changes to these are being picked up in #294 so can possibly just drop this change for now.

assets/stylesheets/components/form/_error-summary.scss
tyom: I think we would need to remove them once we have the typography sorted. Otherwise we'd have to redeclare all the heading styles individually for each component. Although it might be a good idea to use a BEM class to apply `error-summary` specific styles.

assets/stylesheets/_layout.scss
tyom: I wonder if they should go under components.

assets/stylesheets/components/form/_error-summary.scss
```diff
@@ -0,0 +1,26 @@
+@import "../../settings";
+
+.c-error-summary {
+  border: 5px solid $error-colour;
+  padding: 20px;
```
teneightfive: Magic numbers.

PR #287, opened 2017-06-29
Fix bug where investment/start was redirected by earlier routes

tyom:
Move the `/start` further up the chain to ensure the page is not redirected by `redirectToDetails`.


feedmypixel:
LGTM 👍

PR #288, opened 2017-06-30
let assets passthrough so they are not recorded on locals.returnTo

feedmypixel:
### For consideration

This was originally for a bug fix that now does not exist. It feels like it may be worth keeping the part that places client side assets within an `assets` folder?

@tyom thoughts?


feedmypixel:
This isn't quite right - sorry closing for the moment

PR #289, opened 2017-06-30
Don't record assets in returnTo property

feedmypixel:
### Problem
Assets were being stored in `locals.returnTo` property

### Solution
This work:
- enables the application to reference client side assets from `/assets` folder
- add `assets` to auth url check so assets are not stored in returnTo locals property


tyom:
I was trying to avoid us having extra `assets/` directory in public when I removed it earlier.

src/middleware/auth.js
```diff
@@ -1,6 +1,6 @@
 module.exports = function auth (req, res, next) {
   const url = req.url
-  const passThrough = /^\/(support|ping|login)\b/.test(url) || req.session.token
+  const passThrough = /^\/(support|ping|login|assets)\b/.test(url) || req.session.token
```
teneightfive: We also need to add `public` and `favicon.ico` to this exempt list.

src/server.js
```diff
@@ -53,12 +53,9 @@ app.set('view engine', 'njk')
 nunjucks(app, config)
 
 // Static files
-app.use(favicon(path.join(config.root, 'public/images', 'favicon.ico')))
+app.use(favicon(path.join(config.root, 'public/assets/images', 'favicon.ico')))
 app.use(express.static(path.join(config.root, 'public')))
```
teneightfive: Might need to set this as `/public` so that it can easily be added to the exempt list?

feedmypixel:
Closing for the moment as the bug is not 100% reproducible

PR #278, opened 2017-06-27
Use sensible name for URLs

teneightfive:
This starts to use a sensible and predictable URL structure for the feature apps.

### Auth

`/login` => `/sign-in`
`/login/signout` => `/sign-out`

### Account

`/myaccount` => `/profile`

### Support

`/support/bug` => `/support/feedback`

### Contacts

`/contact/add` => `/contacts/create`
`/contact/:id/details` => `/contacts/:id`
`/contact/:id/edit` => `/contacts/:id/edit`
`/contact/interactions/:id` => `/contacts/:id/interactions`
`/contact/:id/archive` => `/contacts/:id/archive`
`/contact/:id/unarchive` => `/contacts/:id/unarchive`

### Investment projects

`/investment/start` => `/investment-projects/create/1`
`/investment/create` => `/investment-projects/create/2`
`/investment/:id/interaction` => `/investment-projects/:id/interactions`
`/investment/:id/interaction/create` => `/investment-projects/:id/interactions/create`
`/investment/:id/interaction/:interactionId/edit` => `/investment-projects/:id/interactions/:interactionId/edit`
`/investment/:id/*` => `/investment-projects/:id/*`

### Company interactions

`/interaction/add-step-1` => `/interactions/create/1`
`/interaction/add` => `/interactions/create/2`
`/interaction/:id/details` => `/interactions/:id`
`/interaction/:id/edit` => `/interactions/:id/edit`

## Issues

Addresses #167.

PR #291, opened 2017-06-30
Fix: remove unused JavaScript file

feedmypixel:
Remove investment bundle that is no longer being used


teneightfive:
This is already being addressed in #278 but we could merge this so its a clearer history.

feedmypixel:
> This is already being addressed in #278 but we could merge this so its a clearer history.

No need for it to be clearer closing this

PR #294, opened 2017-06-30
Refactor app wide activity notifications

teneightfive:
This strips back flash messages to always being displayed in the same
place and using a consistent style for them.

## Screenshot

screen shot 2017-07-03 at 16 16 36


assets/stylesheets/components/_messages.scss
```diff
@@ -0,0 +1,28 @@
+.messages {
+  margin-top: 1em;
```
tyom: Ideally this should go in assemblies (which is not merged yet).

assets/stylesheets/components/_messages.scss
```diff
@@ -0,0 +1,52 @@
...
+
+  + * {
+    margin-top: $default-spacing-unit;
+  }
+}
+
+.c-messages__close {
+  @include bold-font(16);
+  position: absolute;
+  top: $default-spacing-unit;
```
tyom: Alignment is a bit off. I wonder if there's a better way to align the button. e.g. set top, bottom to `0` and absolute element height with margin set to `auto`. That way it would always centre it vertically. Or perhaps flexbox?

assets/stylesheets/components/_messages.scss
```diff
@@ -0,0 +1,52 @@
...
+  // between elements
+  margin-top: $default-spacing-unit;
+
+  &:focus {
+    outline: none;
+  }
+}
+
+.c-messages__item {
+  @include bold-font(20);
```
tyom: Do we want to adjust font size for smaller screens? Do we have to use these terrible typography mixins?

tyom:
Could add an example to components page alongside the form components to document this.

assets/javascripts/modules/messages.js
```diff
@@ -0,0 +1,54 @@
...
+
+  render () {
+    Array.from(this.messages).forEach(this.appendClose.bind(this))
+  },
+}
+
+module.exports = {
+  init () {
+    const elements = document.querySelectorAll('.js-Messages')
+    for (let pos = 0; pos < elements.length; pos += 1) {
```
tyom: Might be easier to write as
```js
const elements = document.querySelectorAll('.js-Messages')
Array.from(elements).forEach(element => Messages.init(element))
```

assets/stylesheets/components/_messages.scss
teneightfive: I was thinking of just sticking it to the top rather than having it vertically centred. But agreed that it looks slightly off when only a one line message. It's also because I reduced the font size as it has less importance.

assets/stylesheets/components/_messages.scss
teneightfive: >Do we have to use these terrible typography mixins?

🤣 

It's more for consistency at the moment. If you want to do a change to strip out their removal then at least we know where they're used. 

>Do we want to adjust font size for smaller screens?

I would say yes. But mainly because we're using those font mixins elsewhere and reducing the font size on smaller screens too.

assets/stylesheets/components/_messages.scss
tyom: Maybe tweak the line height to match the error message.

PR #296, opened 2017-07-03
Move CH details into own sub app and details controller

ztolley:
This is step 1 in a number of steps to refactor companies.
Step 2, which will follow soon after, is to refactor company details into a single details controller.


src/apps/ch/controllers.js
```diff
@@ -0,0 +1,30 @@
+const { title } = require('case')
+
+const { getCHCompany } = require('./repos')
+const { getDisplayCH } = require('./services/formatting')
+const { chDetailsLabels } = require('./labels')
+
+const chDetailsDisplayOrder = ['name', 'company_number', 'registered_address', 'business_type', 'company_status', 'incorporation_date', 'sic_code']
```
tyom: Multiline might be easier to read.

src/apps/ch/services/formatting.js
```diff
@@ -0,0 +1,29 @@
...
+    company_number: companyHouseData.company_number,
+    business_type: companyHouseData.company_category,
+    company_status: companyHouseData.company_status,
+    registered_address: getFormattedAddress(companyHouseData, 'registered'),
+    incorporation_date: formatLongDate(companyHouseData.incorporation_date),
+  }
+
+  displayCH.sic_code = []
+  if (companyHouseData.sic_code_1 && companyHouseData.sic_code_1.length > 0) displayCH.sic_code.push(companyHouseData.sic_code_1)
+  if (companyHouseData.sic_code_2 && companyHouseData.sic_code_1.length > 0) displayCH.sic_code.push(companyHouseData.sic_code_2)
```
tyom: Shouldn't it be `sic_code_2.length`?

Also, might be easier to use Lodash `get`:
```js
if (get(companyHouseData, `sic_code_1`, []).length) {...}
```

src/apps/ch/views/details.njk
```diff
@@ -0,0 +1,28 @@
...
+{% endblock %}
+
+{% block main_grid_left_column %}
+  <nav class="local-nav">
+    <a class="local-nav__link is-active" href="/ch/{{company.company_number}}/details">Company details</a>
+  </nav>
+{% endblock %}
+
+{% block main_grid_right_column %}
+  {% include "_includes/messages.njk" %}
```
tyom: Dom refactored flash messages in #294. So might want to keep an eye on that.

src/apps/ch/index.js
```diff
@@ -0,0 +1,6 @@
+const router = require('./router')
+
+module.exports = {
+  mountpath: '/ch',
```
tyom: Can we call it `company-house` so it's obvious what it is?

src/apps/companies/labels.js
```diff
@@ -53,4 +45,4 @@ const accountManagementDisplayLabels = {
   oneListAccountManager: 'One List account manager',
 }
 
-module.exports = { companyDetailsLabels, chDetailsLabels, companyTypeOptions, hqLabels, accountManagementDisplayLabels }
+module.exports = { companyDetailsLabels, companyTypeOptions, hqLabels, accountManagementDisplayLabels }
```
tyom: I think we've been using multiline exports as convention.

src/apps/ch/controllers.js
```diff
@@ -0,0 +1,30 @@
+const { title } = require('case')
+
+const { getCHCompany } = require('./repos')
```
teneightfive: Can we use a clearer name than `CH` when referring to companies house. Maybe `CompaniesHouse`.

When I first started I had no idea what it referred to.

src/apps/ch/index.js
teneightfive: I don't think we should have a separate app and a separate URL for companies house companies. Can we not do something on the ID structure to determine where to get the data from and keep a consistent URL structure and a consistent set of controllers and views? 

The view can then handle what data it needs to display.

If we use separate URLs we're back to the same scenario as before (`/view/ch/[id]`, `/view/ukother/[id]`, `/view/foreign/[id]`).

src/apps/ch/services/formatting.js
```diff
@@ -0,0 +1,29 @@
...
+    name: title(companyHouseData.name),
+    company_number: companyHouseData.company_number,
+    business_type: companyHouseData.company_category,
+    company_status: companyHouseData.company_status,
+    registered_address: getFormattedAddress(companyHouseData, 'registered'),
+    incorporation_date: formatLongDate(companyHouseData.incorporation_date),
+  }
+
+  displayCH.sic_code = []
+  if (companyHouseData.sic_code_1 && companyHouseData.sic_code_1.length > 0) displayCH.sic_code.push(companyHouseData.sic_code_1)
```
teneightfive: Could check `companyHouseData.sic_code_1.length > 0` once and do a switch on `companyHouseData.sic_code_1` instead of repeating this each time.

teneightfive: Or just add each value to an array and use a [filter](https://lodash.com/docs/4.17.4#filter) to remove empty values.

src/apps/ch/index.js
ztolley: Techincally a CH is a seperate entity. It makes things simpler if we treat it as such, rather than have to figure out which thing to do base on the stucture of an id, plus we will probably already have to use isGiud for some routes so you are making it more difficult for yourself.

src/apps/ch/controllers.js
ztolley: I used tyom's suggestion company-house

src/apps/ch/services/formatting.js
ztolley: seems like a long way around doing a simple thing, Tyom's suggestion for syntax change actually is nice and readable.

src/apps/ch/controllers.js
teneightfive: The [organisation](https://www.gov.uk/government/organisations/companies-house) is called `companies-house` though so might be better suited. `company-house` isn't a thing.

src/apps/ch/services/formatting.js
teneightfive: Even with `get`, this feels quite repetitive:

```js
displayCH.sic_code = []
if (get(companyHouseData, `sic_code_1`, []).length) {
  displayCH.sic_code.push(companyHouseData.sic_code_1)
}
if (get(companyHouseData, `sic_code_2`, []).length) {
  displayCH.sic_code.push(companyHouseData.sic_code_2)
}
if (get(companyHouseData, `sic_code_3`, []).length) {
  displayCH.sic_code.push(companyHouseData.sic_code_3)
}
if (get(companyHouseData, `sic_code_4`, []).length) {
  displayCH.sic_code.push(companyHouseData.sic_code_4)
}
```

could just be:

```js
displayCH.sic_code = filter([
  companyHouseData.sic_code_1,
  companyHouseData.sic_code_2,
  companyHouseData.sic_code_3,
  companyHouseData.sic_code_4,
])
```

src/apps/company-house/services/formatting.js
```diff
@@ -0,0 +1,30 @@
...
+    name: title(companyHouseData.name),
+    company_number: companyHouseData.company_number,
+    business_type: companyHouseData.company_category,
+    company_status: companyHouseData.company_status,
+    registered_address: getFormattedAddress(companyHouseData, 'registered'),
+    incorporation_date: formatLongDate(companyHouseData.incorporation_date),
+  }
+
+  displayCH.sic_code = []
+  if (get(companyHouseData, `sic_code_1`, []).length) displayCH.sic_code.push(companyHouseData.sic_code_1)
```
teneightfive: Even with `get`, this feels quite repetitive:

```js
displayCH.sic_code = []
if (get(companyHouseData, `sic_code_1`, []).length) {
  displayCH.sic_code.push(companyHouseData.sic_code_1)
}
if (get(companyHouseData, `sic_code_2`, []).length) {
  displayCH.sic_code.push(companyHouseData.sic_code_2)
}
if (get(companyHouseData, `sic_code_3`, []).length) {
  displayCH.sic_code.push(companyHouseData.sic_code_3)
}
if (get(companyHouseData, `sic_code_4`, []).length) {
  displayCH.sic_code.push(companyHouseData.sic_code_4)
}
```

could just be:

```js
displayCH.sic_code = filter([
  companyHouseData.sic_code_1,
  companyHouseData.sic_code_2,
  companyHouseData.sic_code_3,
  companyHouseData.sic_code_4,
])
```

ztolley: That is quite neat

src/apps/company-house/index.js
```diff
@@ -0,0 +1,6 @@
+const router = require('./router')
+
+module.exports = {
+  mountpath: '/company-house',
```
teneightfive: I wrote this in another stream and think it got lost:

>The [organisation](https://www.gov.uk/government/organisations/companies-house) is called `companies-house` though so might be better suited. `company-house` isn't a thing.

src/apps/company-house/controllers.js
```diff
@@ -0,0 +1,38 @@
+const { title } = require('case')
+
+const { getCHCompany } = require('./repos')
```
teneightfive: It would be clearer to use companiesHouse rather than CH.

ztolley: getCompaniesHouseCompany is a little.. verbose

ztolley: Switching to that naming for other things get Reall long, i.e. CompaniesHouseDetailsDisplayOrder

teneightfive: Seeing as this method is actually already within the `companies-house` app does it need to contain the verbose name in this instance?

It could just be `getCompany()`.

src/apps/companies-house/controllers.js
```diff
@@ -0,0 +1,38 @@
...
+    const companiesHouse = await getCompaniesHouse(req.session.token, req.params.id)
+    const headingName = title(companiesHouse.name)
+
+    res.render('ch/views/details', {
+      headingName,
+      tab: 'details',
+      companiesHouseDetails: getDisplayCompaniesHouse(companiesHouse),
+      companiesHouseDetailsDisplayOrder,
+      companiesHouseDetailsLabels,
+      addUrl: `/company/add/ltd/${companiesHouse.company_number}`,
```
teneightfive: Should this be being created in the controller? 

If we fix the URL structure before this PR it could be a relative URL within the view.

src/apps/companies-house/controllers.js
```diff
@@ -0,0 +1,38 @@
...
+  'incorporation_date',
+  'sic_code',
+]
+
+async function getDetails (req, res, next) {
+  try {
+    const companiesHouse = await getCompaniesHouse(req.session.token, req.params.id)
+    const headingName = title(companiesHouse.name)
+
+    res.render('ch/views/details', {
```
teneightfive: This path looks incorrect now the app has been renamed.

src/apps/companies-house/index.js
```diff
@@ -0,0 +1,6 @@
+const router = require('./router')
+
+module.exports = {
+  mountpath: '/companies-house',
```
teneightfive: I'm still not convinced about having a separate app and set of routes of companies house companies.

Are these companies returned in the search results? And if we have a separate sub-app are we expecting to have a separate _collection_ at this mount path which returns all companies house organisations?

Or do we want these companies to also be returned as part of the _companies_ collection?

src/apps/companies-house/views/details.njk
```diff
@@ -0,0 +1,26 @@
...
+    <a class="local-nav__link is-active" href="/companies-house/{{companiesHouse.company_number}}/details">Company details</a>
+  </nav>
+{% endblock %}
+
+{% block main_grid_right_column %}
+  <h2 class="heading-medium">Company summary</h2>
+
+  <div class="dark-panel">
+    {{ keyvaluetable(companiesHouseDetails, labels=companiesHouseDetailsLabels, keyorder=companiesHouseDetailsDisplayOrder, readonly=true) }}
+    <p class="footnote">Powered by data from <strong>Companies House</strong></p>
```
teneightfive: Is _powered_ the right term? Maybe just saying `Companies house data` is enough.

src/apps/companies-house/views/details.njk
```diff
@@ -0,0 +1,26 @@
...
+  </div>
+
+  <p><a href="{{addUrl}}" class="button">Add</a></p>
```
teneightfive: What does this add? Does it add another companies house company? Or does it create it as a data hub company? If the latter what is the difference and how does the user know they're doing something different?

src/apps/companies/controllers/ltd.js
```diff
@@ -2,6 +2,8 @@ const companyService = require('../services/data')
 const companyFormService = require('../services/form')
 const companyFormattingService = require('../services/formatting')
 const companyRepository = require('../repos')
+const { getCHCompany } = require('../../companies-house/repos')
```
teneightfive: We could drop the `CH` from this method if we keep it as a separate sub-app as the companies house part is inferred from the sub-app.

src/apps/companies/controllers/ltd.js
```diff
@@ -18,7 +20,7 @@ async function getDetails (req, res, next) {
     res.locals.companyDetailsLabels = companyDetailsLabels
 
     if (company.companies_house_data) {
-      res.locals.chDetails = companyFormattingService.getDisplayCH(company.companies_house_data)
+      res.locals.chDetails = getDisplayCH(company.companies_house_data)
```
teneightfive: Do we also display companies house data in data hub companies?

src/apps/companies/router.js
```diff
@@ -42,7 +39,7 @@ router
   .post(foreignController.postDetails)
 
 router
-  .route('/company/add/ltd')
+  .route('/company/add/ltd/:company_number')
```
teneightfive: Should we stick to using camel case for variables?

src/apps/companies-house/index.js
ztolley: Separate

src/apps/companies/router.js
ztolley: Yes , it's the var name in the model

src/apps/companies-house/views/details.njk
ztolley: That's how it is now and is the agreed wording

src/apps/companies-house/views/details.njk
ztolley: A datahub company

ztolley: This is where they add data to a companies house company and it turns into a cdms company. It's nothing now

src/apps/companies-house/controllers.js
ztolley: It's not relative as it's a link to a different sub app

src/apps/companies/controllers/ltd.js
ztolley: The cdms company app calls this repo

src/apps/companies/controllers/ltd.js
ztolley: Yes

src/apps/companies/router.js
ztolley: I mean no.... the variable has this case as it's the case in the model and it keeps the property name consistent with how we refer all other model property names

src/apps/companies-house/controllers.js
ztolley: It could be in the. Jew if you want. Ur this was just a refactor and that is the existing code

ztolley: Autocorrect gone mad

src/apps/companies-house/index.js
ztolley: Separate entities in search, separate entities in the db, the have different menus and behaviour.. it's almost comparing java to JavaScript

src/apps/companies-house/views/details.njk
teneightfive: _Add_ on its own isn't very descriptive about the action happening though. Something like `Add company to Data Hub` would give a clearer meaning of what this action does.

It's clear to you and I because we've written the code but we should make actions and links as descriptive as possible.

src/apps/companies/controllers/ltd.js
teneightfive: I was suggesting that because the repo is already within the `companies-house` app that infers that you are getting a companies house company. We can then change the method name to `getCompany` removing the use of `CH` as we've agreed `CH` isn't very clear in variable names.

src/apps/companies/router.js
teneightfive: The variable is used in [the ltd controller](https://github.com/uktrade/data-hub-frontend/pull/296/files#diff-67462a4fc7d1a7b4bc29463b9f4864d5R59) so doesn't need to match the naming convention used on the model in the API.

We should use a consistent naming convention in the frontend and [all other](https://github.com/uktrade/data-hub-frontend/blob/develop/src/apps/search/router.js#L10) param variables names are using camel case.

src/apps/companies-house/index.js
teneightfive: >It's almost comparing java to JavaScript
Not sure I see the comparison 🤔 

The thought behind my suggestion was that we could remove the need to do have to check what type a company is when trying to build the URL in things like search and handle that in the company app in param middleware. This could still load a different view which would cater for different menus.

I guess the question is whether we want to create a separate collection at `/companies-house` listing only companies house companies. If we do, then it warrants the separate app.

If we want to combine companies house companies in the `/companies` collection and flag to the user they are different then we should try and keep them in the companies app.

src/apps/companies-house/services/formatting.js
```diff
@@ -0,0 +1,31 @@
+const { title } = require('case')
```
feedmypixel: Same as above with title, should we not be displaying the companies name as it has been entered?

src/apps/companies-house/services/formatting.js
```diff
@@ -0,0 +1,31 @@
...
+const { filter } = require('lodash')
+
+const { getFormattedAddress } = require('../../../lib/address')
+const { formatLongDate } = require('../../../../common/date')
+
+function getDisplayCompaniesHouse (companiesHouseData) {
+  if (!companiesHouseData) return null
```
feedmypixel: missing curlies

src/apps/companies-house/services/formatting.js
```diff
@@ -0,0 +1,31 @@
...
+  const displayCompaniesHouse = {
+    name: title(companiesHouseData.name),
+    company_number: companiesHouseData.company_number,
+    business_type: companiesHouseData.company_category,
+    company_status: companiesHouseData.company_status,
+    registered_address: getFormattedAddress(companiesHouseData, 'registered'),
+    incorporation_date: formatLongDate(companiesHouseData.incorporation_date),
+  }
+
+  displayCompaniesHouse.sic_code = filter([
```
feedmypixel: does this need to be separate from the above object literal?

src/apps/companies-house/views/details.njk
feedmypixel: govuk className - we are trying to move away from these right? Or is that out of scope of this PR?

src/apps/companies-house/views/details.njk
feedmypixel: can we have a variable name that explains a bit more what this is? maybe `addCompaniesHouseCompanyUrl`?

src/apps/companies/controllers/ltd.js
```diff
@@ -54,8 +56,8 @@ function editCommon (req, res, next) {
 
 async function addDetails (req, res, next) {
   try {
-    res.locals.chCompany = await companyRepository.getCHCompany(req.session.token, req.params.company_number)
-    res.locals.chDetails = companyFormattingService.getDisplayCH(res.locals.chCompany)
+    res.locals.chCompany = await getCHCompany(req.session.token, req.params.company_number)
```
feedmypixel: IMO I would stop using the `ch` abbreviation and change it to `CompaniesHouse` everywhere

src/apps/companies/router.js
feedmypixel: I'd question that. Variables on our side are being written/refactored as camelCase. Exposing the model to the front end is not ideal. If needed you could created a adapter object?

src/apps/companies/services/formatting.js
```diff
@@ -1,36 +1,11 @@
-const { title } = require('case')
-
-const { companyDetailsLabels, chDetailsLabels, hqLabels } = require('../labels')
+const { companyDetailsLabels, hqLabels } = require('../labels')
```
feedmypixel: whats `hq`? potentially an opportunity to express meaning through a more verbose variable name? 😁  Stops me having to dig into the required module to understand what it does

test/unit/apps/companies-house/controllers.test.js
```diff
@@ -0,0 +1,102 @@
...
+  afterEach(() => {
+    this.sandbox.restore()
+  })
+
+  it('should get companies house details', (done) => {
+    this.controller.getDetails(
+      this.req,
+      {
+        render: () => {
+          expect(this.getCompaniesHouseStub).to.be.calledWith(this.token, this.id)
```
feedmypixel: can you also check the return value?

test/unit/apps/companies-house/controllers.test.js
```diff
@@ -0,0 +1,102 @@
...
+          done()
+        },
+      },
+      done
```
feedmypixel: Why are you passing `done` in as the `next` parameter? Is this meant to be named `next`? we have been using a `this.next` pattern and stubbing/spying it in other tests

test/unit/apps/companies-house/repos.test.js
```diff
@@ -0,0 +1,27 @@
...
+  afterEach(() => {
+    this.sandbox.restore()
+  })
+
+  it('should call the API with the correct url', () => {
+    return this.repos.getCompaniesHouse('1234', '4321')
+      .then(() => {
+        expect(this.authSub).to.be.calledWith('1234', 'http://test.com/v3/ch-company/4321')
+      })
+  })
```
feedmypixel: so this test is sending in and mocking values and testing that when you call a method via another with values that the inner method is called with those values. Can you explain the value here?

test/unit/data/companies-house.json
```diff
@@ -0,0 +1,22 @@
...
+  "registered_address_county": "",
+  "registered_address_postcode": "S20 1ED",
+  "company_number": "02658484",
+  "company_category": "Private Limited Company",
+  "company_status": "Active",
+  "sic_code_1": "82990 - Other business support service activities n.e.c.",
+  "sic_code_2": "82991 - Other business support service activities n.e.c.",
+  "sic_code_3": "",
+  "sic_code_4": "",
+  "uri": "http://business.data.gov.uk/id/company/07937720",
```
feedmypixel: should this url be in a fixture?

src/apps/companies-house/controllers.js
```diff
@@ -0,0 +1,38 @@
+const { title } = require('case')
```
ztolley: Companies house names would be displayed all in caps I think. I was trying to be consistent with the use of title in other places.

src/apps/companies-house/index.js
ztolley: The next step was going to be to simplify the company url thing, so it does away with ltd, foreign, ukother, but a companies house thing is a seperate entity. The only way they are related is you display companies house data at the top of a company details screen, but it is simply related to a company just like a contact is.

src/apps/companies-house/views/details.njk
ztolley: I'm changing it, as suggested, to build the url in the view instead as it's clearer that way

src/apps/companies-house/controllers.js
```diff
@@ -0,0 +1,38 @@
...
+
+const { getCompaniesHouse } = require('./repos')
+const { getDisplayCompaniesHouse } = require('./services/formatting')
+const { companiesHouseDetailsLabels } = require('./labels')
+
+const companiesHouseDetailsDisplayOrder = [
```
ztolley: You mean move the display order variable out?

src/apps/companies-house/views/details.njk
ztolley: out of scope for the PR. If it should change to a new one you have done, tell me.

src/apps/companies-house/views/details.njk
ztolley: changing it to be in the view

test/unit/apps/companies-house/controllers.test.js
ztolley: Ah, I found this is a really simple way to get the error output IF there is one. If I stub next I see a test fails but dont know why, if I do this then by convention the code sends the error to next

test/unit/apps/companies-house/repos.test.js
ztolley: Its just to test to confirm we call the api with the right url, and seemed to be the simplest way to do this. The reason I mock data is .. well it just seemed simple enough. But yes you could just say resolves and not mock data but then that might break it. I wanted to make it as real as possible.

test/unit/data/companies-house.json
ztolley: I'll make it more Fake

src/apps/companies-house/controllers.js
teneightfive: >My company is called feedMyPixel Ltd. I wouldn't want that title cased.

That does raise a good point actually. Mine is `Ten EightFive Ltd` so changing to title case would change my company name. It seems that they're always uppercase on companies house. Maybe we just keep that case rather than changing the case?

PR #302, opened 2017-07-05
Use constant case for global level template variables

teneightfive:
Using constant case makes it clear from the view that the variable
being used is defined in the global scope for all views.

This also clears out a few variables that were no longer being used in any views.

### DIFF
--- a/src/middleware/locals.js
+++ b/src/middleware/locals.js
@@ -1,8 +1,5 @@
 const logger = require('../../config/logger')
-const pjson = require('../../package.json')
 const config = require('../../config')
-
-const startTime = new Date().getTime()
 
 let webpackManifest = {}
 try {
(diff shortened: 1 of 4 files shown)


src/middleware/locals.js
```diff
@@ -12,19 +9,13 @@ try {
...
-    releaseVersion: pjson.version,
-    startTime: startTime,
-    referer: req.headers.referer,
-    env: config.env,
-    googleTagManagerKey: config.googleTagManagerKey,
-    query: req.query,
-    currentPath: req.path,
     BASE_URL: baseUrl,
     CANONICAL_URL: baseUrl + req.originalUrl,
+    CURRENT_PATH: req.path,
```
tyom: Do we even need this in locals? Can always get it from `req.path`.

src/templates/_layouts/datahub-base.njk
```diff
@@ -34,17 +34,17 @@
     <ul class="proposition-menu">
       {% if user %}
         <li class="proposition-menu__item">
-          <a href="/" class="proposition-menu__link{{ ' is-active' if currentPath == '/' }}">Dashboard</a>
+          <a href="/" class="proposition-menu__link{{ ' is-active' if CURRENT_PATH == '/' }}">Dashboard</a>
```
tyom: I think it's easier to read and manage if space is placed outside of Nunjucks variable in the attribute value. It doesn't affect the browser in any way and at this point there's probably not much use caring about whitespace control in the compiled HTML.

src/middleware/locals.js
teneightfive: It's so that it can be accessed in a view. We're not passing `req` to views so you wouldn't be able to access that property.

PR #308, opened 2017-07-05
Breadcrumbs

teneightfive:
This adds support for breadcrumbs in the application. It has been a key missing part which helps the user in knowing where they are and to navigate around.

## Note

This hasn't been implemented in the companies sub-app yet because the URL structure isn't neatly tied to the mountpath yet. Once URLs are fixed we can introduce a `mountpath` and `displayName` to the app.

### Before

screen shot 2017-07-07 at 12 48 32

### After

screen shot 2017-07-07 at 15 55 40


src/apps/investment-projects/router.js
```diff
@@ -18,6 +18,14 @@ const {
 router.use(shared.handleEmptyMiddleware)
 router.use(shared.getLocalNavMiddleware)
 
+router.use((req, res, next) => {
```
feedmypixel: Should this be a `addBaseBreadcrumb`?

src/apps/contacts/controllers/interactions.js
```diff
@@ -9,6 +9,7 @@ async function getInteractions (req, res, next) {
     res.locals.addInteractionUrl = `/interactions/create/1?contact=${res.locals.contact.id}`
 
     res.locals.title.unshift('Interactions')
```
feedmypixel: It feels like we maybe getting some common functionality that could be abstracted into middleware? Maybe we have a think about moving some locals such as `pathname`, `title` and `breadcrumb` creation up a level to the `routes` files? We could even abstract the `pathname` and `title` into configs (if needed)?

so:

#### current src/apps/contacts/router.js
```
router.get('/:contactId/interactions', contactController.getCommon, contactInteractionController.getInteractions)
```

could become:

#### src/apps/contacts/router.js
```
router.get('/:contactId/interactions', 
  viewDetails('Interactions', 'Interactions'),
  contactController.getCommon, 
  contactInteractionController.getInteractions
)

```

#### Middleware
```
function viewDetails(breadcrumb, title) {
  return function (req, res, next) {
    req.breadcrumbs(breadcrumb)
    res.locals.title.unshift(title)
    next()
  }
}
```


_caveat: this is psuedo code_

teneightfive: Yes, that was my next step. @tyom and I were actually discussing that we may use the breadcrumb array to create the contents of `<title />`. 

This would mean we remove the duplication of that content. I think the page heading will still be different but could also be abstracted at a later date.

The problem is that all the apps are still in different structures at the moment so would mean a lot of refactoring before comes completely consistent. 

I was also looking at trying to move the details from sub-app level breadcrumb middleware to the sub-app index loader so we could drop those but it meant it was loaded too late.

Will see if there is some more refactoring that can be done around the basic sub-app breadcrumbs.

feedmypixel: > The problem is that all the apps are still in different structures at the moment so would mean a lot of refactoring before comes completely consistent.

Agreed, sounds sensible to get the refactoring done and then build on top

teneightfive: @feedmypixel I found a nicer way to do it: add a `displayName` property to each sub-app that contains a `mountpath` and add the breadcrumb in the `routers.js` file.

Adding breadcrumb items to each view is still a bit manual at this stage though.

assets/stylesheets/components/_breadcrumb.scss
```diff
@@ -0,0 +1,44 @@
...
+  &:link,
+  &:active,
+  &:hover,
+  &:visited {
+    color: $black;
+    text-decoration: underline;
+  }
+
+  &.is-active {
+    @include bold-font(16);
```
feedmypixel: Is bold right for an active item? I would of though going the other way with a normal font weight and a grey? Seeing it in the screenshot has highlighted this.

src/apps/routers.js
```diff
@@ -10,7 +12,10 @@ subApps.forEach(subAppDir => {
     const subApp = require(`./${subAppDir}`)
 
     if (subApp.mountpath) {
-      appsRouters.push(router.use(subApp.mountpath, subApp.router))
+      router.use(subApp.mountpath, addBaseBreadcrumb(subApp.displayName))
+      router.use(subApp.mountpath, subApp.router)
```
feedmypixel: can these become?
```router.use(subApp.mountpath, addBaseBreadcrumb(subApp.displayName), subApp.router)```

File: src/templates/_components/breadcrumbs.njk [hunk 5]
```diff
@@ -0,0 +1,38 @@
...
+ #     }, {
+ #       name: 'Second level',
+ #       url: '/second-level'
+ #     }]
+ #   } %}
+ #}
+
+{% if items|length > 1 %}
+  <nav aria-label="Breadcrumbs">
+    <ul class="c-breadcrumb">
```
tyom: Should this have `iitemscope temtype="http://schema.org/BreadcrumbList"` attribute followed by `ListItem` for list items if you're using [Microdata schema](http://schema.org/BreadcrumbList)? Also, should you be using `schema.org` for references?

tyom:
Great work. Is this still WIP?

src/apps/contacts/controllers/interactions.js
feedmypixel: Nice - feels like the starts of a model in `src/apps/<appName>/index.js` files

assets/stylesheets/components/_breadcrumb.scss
teneightfive: I've was basing it on the style used for the GOV.UK components navigation:

screen shot 2017-07-07 at 14 35 59

But taking a look it seems inconsistent on GOV.UK too. Some pages use the same weight and also underlined:

screen shot 2017-07-07 at 14 34 55

And others use no underline and the same weight for the active state:

screen shot 2017-07-07 at 14 36 10

Personally I think the bold stands out slightly more making it clearer as to what item in the hierarchy you are currently on. It has more importance so I'm happy with it being a heavier weight.

src/apps/routers.js
teneightfive: Possibly, let me try.

teneightfive:
>Great work. Is this still WIP?

I suppose its less WIP now but potentially blocked by the collections work. We can decide if we just want to merge it to get it in or hold off.

The potential issue is that it will expose the gaps in our IA where we don't have the higher level pages for entities.

I'm leaning towards merging it rather than having branches hanging around for too long. And we can then also refactor page titles to use the breadcrumb to clean up some of the controller code.

assets/stylesheets/components/_breadcrumb.scss
teneightfive: Have dropped the bold

PR #309, opened 2017-07-05
Tweak main header styles

tyom:
- Add bottom border to help balance the page with multiple columns and plenty of white space
- Fix company name issue in Contacts page

## Before
image

## After
after-fullpage

### DIFF
--- a/assets/stylesheets/objects/base/_main-content.scss
+++ b/assets/stylesheets/objects/base/_main-content.scss
@@ -20,5 +20,16 @@
 }
 
 .main-content__header {
-  margin: 30px 0;
+  padding: ($default-spacing-unit * 2) 0;
+  border-bottom: 4px solid $grey-2;
+  margin-bottom: $default-spacing-unit * 2;
+
+  h1,
+  h2 {
+    margin: 0;
+  }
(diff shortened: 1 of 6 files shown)


assets/stylesheets/objects/base/_main-content.scss
```diff
@@ -20,5 +20,16 @@
 }
 
 .main-content__header {
-  margin: 30px 0;
+  padding: ($default-spacing-unit * 2) 0;
+  border-bottom: 4px solid $grey-2;
+  margin-bottom: $default-spacing-unit * 2;
+
+  h1,
```
teneightfive: Using element selectors doesn't feel quite BEM. If you use a class selector that can be used to set the spacing.

teneightfive:
You might want to check the consistency of this. 

For example, it looks strange on the sign in page when there is only the form content beneath the page title:

screen shot 2017-07-06 at 16 49 41

This style should probably be applied by a modifier and only to entity pages.

ztolley:
Have any of the designers seen this?

tyom:
Closing in favour of #345.

PR #315, opened 2017-07-06
Remove yellow background on logo focus

teneightfive:
## Before

screen shot 2017-07-06 at 19 03 26

## After

screen shot 2017-07-06 at 19 03 55


assets/stylesheets/objects/base/_global-header.scss
```diff
@@ -51,6 +51,10 @@
   &:focus {
     color: inherit;
   }
+
+  &:focus {
+    background: transparent;
```
feedmypixel: I'd question if changing the [global pattern](https://github.com/uktrade/data-hub-frontend/blob/develop/assets/stylesheets/elements/_base.scss#L32-L36) for link backgrounds used across the app, just for this link is right thing to do? 
I think the transparent `.png` is sensible, but why not just keep the [background-colour](https://github.com/uktrade/data-hub-frontend/blob/develop/assets/stylesheets/elements/_base.scss#L33)?

teneightfive: This was actually following the lead of what had been done on GOV.UK. I think the white on the yellow doesn't meet AA contrast requirements so to fix that we can either remove the yellow background colour or change the text colour to black on focus.

Personally I think dropping the background colour actually looks neater and is clearer. I believe @tyom has the view that we should drop the background colour on link focus altogether but seeing as we are using the GOV.UK brand as a base we should probably keep it for now.

tyom:
I'd also remove that line next to the crest.

teneightfive:
>I'd also remove that line next to the crest.

It's part of the logo unfortunately.

tyom:
> It's part of the logo unfortunately.

Maybe it should be `#CF102D` then. I can' find any examples of the department using the white line on black background.

feedmypixel:
> I'd also remove that line next to the crest.

yeah that does look funny on focus

feedmypixel:
> Maybe it should be #CF102D then. I can' find any examples of the department using the white line on black background.

feels like we should have contact with the branding department around this?

Issue #169, opened 2017-06-07
Use address formatter macro when displaying addresses


teneightfive:
## Problem

We are still manually displaying addresses in [some views](https://github.com/uktrade/data-hub-fe-beta2/blob/1a4e05d8316688e7cc9940d046c5bf98e3e8b77d/src/views/investment/start.njk#L77-L89). We should be consistent and use the [address formatter macro](https://github.com/uktrade/data-hub-fe-beta2/commit/0eff2670cbd6e5e5e62d9152d342d3d95535f26b) for all displayed addresses.

## Solution

Ensure we use a consistent macro/filter/component for displaying addresses.

teneightfive:
As part of this change the macro could also be moved to a filter as it normally returns a string.

PR #318, opened 2017-07-07
Investment search

teneightfive:
The main piece of work here is to add investment projects to the global search. It includes a refactor to search to make it easier to use in other sub-apps.

## What it looks like

image


src/templates/_layouts/datahub-base.njk
```diff
@@ -13,6 +13,7 @@
 
 {% from "_macros/form.njk" import EntitySearchForm, Form with context %}
 {% from "_macros/form.njk" import TextField, MultipleChoiceField %}
+{% from "_macros/address.njk" import addressFormatter %}
```
tyom: Should this be a macro or a filter? Perhaps a separate PR.

src/templates/_layouts/search.njk
```diff
@@ -0,0 +1,44 @@
...
+{% endblock %}
+
+
+{% block main_grid_left_column %}
+  &nbsp; {# // TODO filters #}
+{% endblock %}
+
+
+{% block main_grid_right_column %}
+  {{ super() }}
```
tyom: Is this needed? If so should it be before summary?

src/templates/_layouts/search.njk
```diff
@@ -0,0 +1,44 @@
...
+
+  {% if results %}
```
tyom: Should the block be called `search_results` to be more explicit about its context? Same for the others.

src/templates/_layouts/search.njk
```diff
@@ -0,0 +1,44 @@
...
+
+    {% block results_summary %}
+      {% component 'results-summary', {
+        resultType: searchEntity | replace('_', ' '),
+        total: results.count,
+        searchTerm: searchTerm
+      } %}
+    {% endblock %}
+
+    {% block results %}
```
tyom: Should the block be called `search_results` to be more explicit about its context? Same for the others.

src/templates/_layouts/datahub-base.njk
teneightfive: Could potentially be moved to a filter. Definitely a separate PR though.

teneightfive: There is already an issue linked to this which I've added a comment to: #169

PR #320, opened 2017-07-07
Entity Search component

tyom:
This adds a new EntitySearchForm component which can be used inline for doing company search or as a global company/contacts/investments search.

This will unify and replace several components and ad-hoc implementations we use across the site. It is based on new form components so it builds on top of existing components rather than a separate implementations.

- [x] Build new entity search component
- [x] Create examples of using this component on Components page
- [x] Replace existing searches with single Entity Search component

image

[Demo](https://datahub-beta2-pr-320.herokuapp.com/components/form)


assets/stylesheets/components/form/_entity-search.scss
```diff
@@ -0,0 +1,52 @@
...
+}
+
+.c-entity-search__button {
+  @include hide-text($search-button-width, 100%);
+  position: absolute;
+  top: 0;
+  right: 0;
+  border: 0;
+  padding: 0;
+  background: $govuk-blue url("../../../images/_deprecated/search-button.png") no-repeat 8px 50%; // TODO convert to SVG
```
teneightfive: Might be worth converting as part of this PR. Or at least support 2x png.

src/apps/components/views/form.njk
```diff
@@ -5,6 +5,24 @@
...
+
+  <h3 class="heading-small">Global</h3>
+  {{ entitySearch|log|safe }}
+  {{ EntitySearchForm({
+    inputName: 'term',
+    modifier: 'global',
+    inputLabel: 'Search for a company name or contact',
+    searchTerm: entitySearch.searchTerm,
+    aggregations: entitySearch.searchEntityResultsData
+  }) }}
```
teneightfive: Should we add a heading to break the other form elements up?

src/apps/components/views/form.njk
```diff
@@ -5,6 +5,24 @@
...
+
+  <h3 class="heading-small">Standard</h3>
+  {{ EntitySearchForm({
+    inputName: 'term',
+    inputLabel: 'Search for a company name or contact',
+    searchTerm: entitySearch.searchTerm
+  }) }}
+
+  <h3 class="heading-small">Global</h3>
+  {{ entitySearch|log|safe }}
```
teneightfive: Is this from debugging?

test/unit/components/search-bar.test.js
```diff
@@ -1,54 +0,0 @@
-const { renderComponentToDom } = require('../component-helper')
```
teneightfive: This is something to consider with the new macros...we don't have any tests for them...

assets/stylesheets/components/form/_entity-search.scss
feedmypixel: Definitely do the SVG work but I'd say as a separate PR as we would have to do IE8 fallback with it.

assets/stylesheets/components/form/_entity-search.scss
```diff
@@ -0,0 +1,52 @@
...
+    outline: 3px solid $focus-colour;
+  }
+}
+
+.c-entity-search--global {
+  background-color: $grey-3;
+  padding: $default-spacing-unit;
+
+  .c-form-group {
+    font-size: 1.1em;
```
feedmypixel: can we use the `core-font` mixin?

test/unit/components/search-bar.test.js
feedmypixel: yeah agreed we should get some tests in around these

assets/stylesheets/components/form/_entity-search.scss
tyom: No.

test/unit/components/search-bar.test.js
tyom: We do have a form of tests – our components page. It could be useful to have some sort of unit tests for those but I would argue for it to be a separate PR. Even though new form components are pretty stable there are still quite a few changes going in there every time I convert a new form. I'd wait until it stabilises a bit further to avoid wasting time constantly rewriting the tests.

tyom: Initial tests are in #326.

assets/stylesheets/components/form/_entity-search.scss
teneightfive: 😂

teneightfive: My thoughts: 

We should use one way or another. Not a mix of the two.

And as we're currently using `core-font` throughout I would be inclined to keep it for consistency. Once we no longer have a mix of code style for components we can do a PR to replace how we set font sizing.

We can review whether we want to set a different font size for components based on resolution and whether we want to do that at a component level or a global level.

I also think we should consider dropping the font-size as a whole and it could tie into the work of using the full width of the browser. Being an internal service we have some flexibility on this.

tyom: As discussed offline we should address typography globally after we remove more of deprecated stuff.

PR #321, opened 2017-07-07
Replace page titles with breadcrumbs

teneightfive:
The breadcrumb functionality is giving us the same thing we're doing
with page titles. To keep consistency this change switches out the use
of manually setting the page title within all the views and uses
the breadcrumb object to construct the page titles.

## Notes

The breadcrumb middleware uses the original dependency ([express-breadcrumbs](https://github.com/heroicyang/express-breadcrumbs)) as a starting point.

### DIFF
--- a/src/apps/contacts/controllers/details.js
+++ b/src/apps/contacts/controllers/details.js
@@ -15,17 +15,12 @@ async function getCommon (req, res, next) {
     const token = req.session.token
     const contact = res.locals.contact = await contactsRepository.getContact(token, req.params.contactId)
     const company = res.locals.company = await companyRepository.getDitCompany(token, contact.company.id)
-    const name = `${contact.first_name} ${contact.last_name}`
 
     res.locals.id = req.params.contactId
     res.locals.companyUrl = companyService.buildCompanyUrl(company)
     res.locals.reasonForArchiveOptions = reasonForArchiveOptions
-    res.locals.title = [name, 'Contacts']
 
-    req.breadcrumbs({
-      name,
-      url: `/contacts/${contact.id}`,
-    })
+    res.breadcrumb(`${contact.first_name} ${contact.last_name}`, `/contacts/${contact.id}`)
 
     next()
   } catch (error) {
--- a/src/middleware/locals.js
+++ b/src/middleware/locals.js
@@ -10,13 +10,30 @@ try {
 
 module.exports = function locals (req, res, next) {
   const baseUrl = `${(req.encrypted ? 'https' : req.protocol)}://${req.get('host')}`
+  const breadcrumbItems = res.breadcrumb()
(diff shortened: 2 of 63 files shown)


teneightfive:
After discussing with @feedmypixel a bit more it might be better to refactor how we add to page titles and breadcrumbs and also move the method from the request (`req`) object to the response (`res`) object. 

That allows titles and breadcrumb items to be set more like this:

```js
res
  .title('Custom page title')
  .breadcrumb('Page title')
  .render('page-template')
```

It clearly links build of these items to the final step of a chain of middleware which is normally to render a view.

src/apps/contacts/controllers/details.js
```diff
@@ -15,15 +15,13 @@ async function getCommon (req, res, next) {
...
 
     res.locals.id = req.params.contactId
     res.locals.companyUrl = companyService.buildCompanyUrl(company)
     res.locals.reasonForArchiveOptions = reasonForArchiveOptions
-    res.locals.title = [name, 'Contacts']
 
-    req.breadcrumbs({
-      name,
+    res.breadcrumb({
+      name: `${contact.first_name} ${contact.last_name}`,
```
tyom: Could be consistent with other places
```js
res.breadcrumb(`${contact.first_name} ${contact.last_name}`, `/contacts/${contact.id}`)
```

src/middleware/locals.js
```diff
@@ -10,13 +12,29 @@ try {
...
   res.locals = Object.assign({}, res.locals, {
     BASE_URL: baseUrl,
     CANONICAL_URL: baseUrl + req.originalUrl,
     CURRENT_PATH: req.path,
     GOOGLE_TAG_MANAGER_KEY: config.googleTagManagerKey,
-    BREADCRUMBS: req.breadcrumbs(),
+    BREADCRUMBS: breadcrumbItems,
+
+    getPageTitle: () => {
+      const items = map(breadcrumbItems, clone)
```
tyom: Could `breadcrumbItems.map(item => item.name)` here to avoid doing it further down and using two extra dependencies. `map` creates a new array.

src/middleware/locals.js
```diff
@@ -10,13 +12,29 @@ try {
...
+
+      if (title) {
+        if (items.length === 1) {
+          return title
+        }
+
+        items[items.length - 1] = { name: title }
+      }
+
+      return map(items, 'name').reverse().slice(0, -1)
```
tyom: Would this be easier to read?
```js
const items = breadcrumbItems.map(item => item.name)
const title = res.locals.title

if (title) {
  if (items.length === 1) {
    return title
  }

  items.pop()
  items.push(title)
}

return items.reverse().slice(0, -1)
```

PR #323, opened 2017-07-07
Fix contacts header company name

tyom:
## Before
image

## After
image


teneightfive:
This looks a bit rubbish but could be an interim solution before we sort out a consistent header for all pages.

PR #324, opened 2017-07-07
Use form components for company step 1 form

tyom:
image

### DIFF
--- a/src/apps/companies/controllers/add.js
+++ b/src/apps/companies/controllers/add.js
@@ -20,22 +20,24 @@ function getAddStepOne (req, res, next) {
 
 function postAddStepOne (req, res, next) {
   // validate, if bad then generate errors, and show form again
-  const errors = {}
+  const errorMessages = {}
 
   if (isBlank(req.body.business_type)) {
-    errors.business_type = ['You must select a company type']
+    errorMessages.business_type = ['You must select a company type']
   }
 
   if (req.body.business_type === 'ukother' && isBlank(req.body.business_type_uk_other)) {
-    errors.business_type_uk_other = ['You must select the type of business']
+    errorMessages.business_type_uk_other = ['You must select the type of business']
   }
 
   if (req.body.business_type === 'foreign' && isBlank(req.body.business_type_for_other)) {
-    errors.business_type_for_other = ['You must select the type of business']
+    errorMessages.business_type_for_other = ['You must select the type of business']
   }
 
-  if (Object.keys(errors).length > 0) {
-    res.locals.errors = errors
+  if (Object.keys(errorMessages).length > 0) {
+    res.locals.errors = {
+      messages: errorMessages,
+    }
     return getAddStepOne(req, res)
   }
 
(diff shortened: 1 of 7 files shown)


src/apps/companies/controllers/add.js
```diff
@@ -35,7 +35,9 @@ function postAddStepOne (req, res, next) {
   }
 
   if (Object.keys(errors).length > 0) {
-    res.locals.errors = errors
+    res.locals.errors = {
+      messages: errors,
```
feedmypixel: should these be called `errorMessages`?

src/apps/companies/views/add-step-1.njk
```diff
@@ -5,56 +5,57 @@
...
+      selected: company.business_type,
+      error: errors.messages.business_type,
+      options: [
+        {
+          value: 'ltd',
+          label: 'UK private or public limited company',
+          hint: 'Must be a company registered with Companies House'
+        }, {
+          value: 'ukother',
+          label: 'Other type of UK organisation',
```
feedmypixel: Feels like some of this work is not being done in the view in other apps. For example in [labels.js](https://github.com/uktrade/data-hub-frontend/blob/develop/src/apps/investment-projects/labels.js).
How awkward would it be to move this config work out of the view? It potentially feels like this doesn't belong here?

tyom: You could argue that it does belong here as those are just other fields and labels defined in the view. Just like the ones above and below. I think going forward it would be nice to have a form builder that automatically builds and validates forms from predefined schema. And I have done some exploration work on that but it's not ready yet.

@teneightfive your thoughts please.

teneightfive:
@tyom classic Chidgey :trollface: :

screen shot 2017-07-09 at 22 55 25

src/apps/companies/views/add-step-1.njk
teneightfive: We do seem to have a few different places where values and options are defined.

As @tyom said I think moving forward we'd like to remove the need for many of these views and have the form pages built from a schema. This schema could also potentially handle multi-step forms too.

I would like us to get to a state where we have one `form.njk` template in `src/templates/_layouts/` and it loops over an expected schema object to build out the view.

Will hopefully make maintenance easier so that we don't have to be dipping between controllers, middleware and views to update fields.

src/apps/companies/controllers/add.js
```diff
@@ -20,22 +20,24 @@ function getAddStepOne (req, res, next) {
...
   if (req.body.business_type === 'foreign' && isBlank(req.body.business_type_for_other)) {
-    errors.business_type_for_other = ['You must select the type of business']
+    errorMessages.business_type_for_other = ['You must select the type of business']
   }
 
-  if (Object.keys(errors).length > 0) {
-    res.locals.errors = errors
+  if (Object.keys(errorMessages).length > 0) {
+    res.locals.errors = {
+      messages: errorMessages,
```
teneightfive: Maybe consider renaming `messages` to `fields` in future as they need to map to fields anyway.

tyom: That'd have to be a different PR renaming it in all other apps and macro.

PR #325, opened 2017-07-08
Use new form components in edit/create contact form

tyom:
First pass at rebuilding contact form with new components.

There are a couple of TODOs left which are out of scope of this task and should be looked at later when other forms are replaced as some of the leftover code is required for other forms to function.

I've also removed "Cancel" button as I don't believe there is a need for it in non-single page application which preserves history (it's always easier to hit back button or one of the links, e.g. breadcrumb) and it often gets the return path wrong.

localhost-3001-contacts-216d1b1c-173e-4ef5-9508-5a1ab34a0e99-edit


src/templates/_macros/form.njk
```diff
@@ -30,6 +30,7 @@
  # @param {string} [props.method=POST] - Form method
  # @param {string} [props.action] - Form action url
  # @param {object} [props.errors] - Object containing form errors
+ # @param {object} [props.customFields] - Custom fields to be added as hidden inputs
```
teneightfive: `hiddenFields` might be a better name

PR #326, opened 2017-07-08
Initial ideas on how to test new form macros

tyom:
This adds a new helper for unit testing macros as well as initial tests for `TextField` component.

### DIFF
--- a/test/unit/component-helper.js
+++ b/test/unit/component-helper.js
@@ -11,7 +11,8 @@ const nunjucks = nunjucksConfig(null, {
 })
 
 const COMPONENTS_PATH = '_components/'
-const COMPONENT_EXT = 'njk'
+const MACROS_PATH = '_macros/'
+const EXT = 'njk'
 
 const normaliseHtml = (string) => {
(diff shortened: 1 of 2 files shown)


test/unit/component-helper.js
```diff
@@ -24,8 +25,25 @@ const normaliseHtml = (string) => {
...
+  return {
+    render (macroName, props) {
+      const importString = `{% from "${filePath}" import ${macroName} %}`
+      const macroOutput = nunjucks.renderString(`${importString} {{ ${macroName}(${JSON.stringify(props)}) }}`)
+      return normaliseHtml(macroOutput)
+    },
+
+    renderToDom (macroName, props) {
+      const macroOutput = this.render(macroName, props)
+      return (new JSDOM(macroOutput)).window.document.body.querySelector('body>:first-child')
```
feedmypixel: could do `document.documentElement.firstElementChild`

tyom: True that. Although `document.body.firstElementChild` is the one we want rather than `documentElement` as that refers to `html` element.

PR #327, opened 2017-07-10
Feature/invest form refactor

ztolley:
Refactored the investment project form middleware into separate files a the more forms are about to be added and the existing single file would become unmanageable.


ztolley:
Parking this while waiting for Ben to refactor some api calls

ztolley:
Reopen now that Ben is not refactoring this code.

PR #330, opened 2017-07-10
Updated test data and modified the changed CSS locators

venureddy95:
I have updated the Test data for company searches and Modified the changed CSS locators for Page objects. And even fixed the login issues (with help of Tyom) for feature level test runs. This should be a quick PR to move forward in using the Automation scripts by FEDs


test/acceptance/features/company-create.feature
```diff
@@ -10,6 +10,7 @@ Feature: Create a new company
     When I create a new “UK private or public limited company”
     Then I see the creation confirmation message
     And I verify that my newly created company is present in search results
+    And I logout of Data Hub website
```
tyom: Why do we need to log out every time?

venureddy95: I had introduced it when we had login page issues for each scenario. It can be removed for all and stay only for couple of scenarios. I will remove them in my next PR.

PR #331, opened 2017-07-11
Evaluation readout information screen

feedmypixel:
This work covers the evaluation details page:

screencapture-localhost-3000-investment-projects-32fe267d-34fe-4467-92be-20b34f3715eb-evaluation-1500290976239

### With Defaults
screencapture-localhost-3000-investment-projects-f22ae6ac-b269-4fe5-aeba-d6a605b9a7a7-evaluation-1500290961434

### DIFF
--- a/config/nunjucks/filters.js
+++ b/config/nunjucks/filters.js
@@ -36,7 +36,7 @@ const filters = {
     )
   },
 
-  provideDefault: (collection, defaultValue = 'Not found') => {
+  collectionDefault: (collection, defaultValue = 'Not found') => {
     const isEmptyObjOrArray = (value) => {
       return ((isPlainObject(value) || isArray(value)) && isEmpty(value))
     }
--- a/config/nunjucks/filters.js
+++ b/config/nunjucks/filters.js
@@ -26,6 +26,7 @@ const filters = {
   keys,
   flatten,
   map,
+  isArray,
 
   highlight: (string, searchTerm) => {
     const regEx = new RegExp(`(${searchTerm})`, 'gi')
@@ -33,6 +34,28 @@ const filters = {
     return new nunjucks.runtime.SafeString(
       string.replace(regEx, '<span class="u-highlight">$1</span>')
     )
+  },
+
+  provideDefault: (collection, defaultValue = 'Not found') => {
+    const isEmptyObjOrArray = (value) => {
+      return ((isPlainObject(value) || isArray(value)) && isEmpty(value))
+    }
+
+    if (isArray(collection)) {
+      return collection.slice().map((element) => {
+        return !element || isEmptyObjOrArray(element) ? defaultValue : element
+      })
+    }
+    if (isPlainObject(collection)) {
+      const obj = Object.assign({}, collection)
+
+      Object.keys(obj).forEach((key) => {
+        if (!obj[key] || isEmptyObjOrArray(obj[key])) {
+          obj[key] = defaultValue
+        }
+      })
+      return obj
+    }
   },
 
   removeNilAndEmpty: (collection) => {
--- a/src/apps/investment-projects/controllers/evaluation.js
+++ b/src/apps/investment-projects/controllers/evaluation.js
@@ -1,0 +1,36 @@
+const { get } = require('lodash')
+
+const {
+  evaluationValueLabels,
+  evaluationFdiLabels,
+  evaluationLandingLabels,
+} = require('../labels')
+const { getDataLabels } = require('../../../lib/controller-utils')
+const {
+  transformInvestmentValueForView,
+  transformInvestmentFDIForView,
+  transformInvestmentLandingForView,
+} = require('../services/formatting')
+
+function renderEvaluationPage (req, res, next) {
+  const transformedValue = transformInvestmentValueForView(res.locals.investmentData)
+  const transformedFDI = transformInvestmentFDIForView(res.locals.investmentData)
+  const transformedLanding = transformInvestmentLandingForView(res.locals.investmentData)
+
+  if (get(res, 'locals.investmentData')) {
+    return res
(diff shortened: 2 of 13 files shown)


src/apps/investment-projects/controllers/evaluation.js
```diff
@@ -0,0 +1,34 @@
...
+} = require('../services/formatting')
+
+function renderEvaluationPage (req, res, next) {
+  res.locals.title.unshift('Evaluation')
+
+  const evaluationValueViewData = transformInvestmentEvaluationValueForView(res.locals.investmentData)
+  const evaluationFdiViewData = transformInvestmentEvaluationFDIForView(res.locals.investmentData)
+  const evaluationLandingViewData = transformInvestmentEvaluationLandingForView(res.locals.investmentData)
+
+  if (get(res, 'locals.investmentData')) {
```
teneightfive: What happens if this condition doesn't get met? Won't the page hang?

src/apps/investment-projects/services/formatting.js
```diff
@@ -164,10 +165,111 @@ function transformInvestmentRequirementsForView (data) {
...
+    return
+  }
+
+// TODO how is research_and_development wanting to be displayed? currently
+// it is an array
+// TODO do we need defaults or just don't show items if they don't have a value?
+// TODO the text values for type_of_investment where going to come from
+// metadataRepo.investmentTypeOptions, but this only contains FDI,
+// None FDI and Commitment to Invest
+// TODO "exporting" - get a ticket in Jira about entering this value into the app
```
jim68000: There should be a Boolean value called 'export_oriented' in the API, but this has disappeared from the docs - it's the last form field on the value screen - I suspect that means it's in the API, but not my docs - I will chase

reupen: It's called `export_revenue` in the API (probably because that's what it's called on the summary screen).

src/apps/investment-projects/services/formatting.js
```diff
@@ -164,10 +170,73 @@ function transformInvestmentRequirementsForView (data) {
...
+function transformInvestmentLandingForView (data) {
+  if (!isPlainObject(data)) {
+    return
+  }
+
+  return Object.assign({}, {
+    uk_company: data.uk_company ? { name: data.uk_company.name, url: buildCompanyUrl(data.uk_company) } : null,
+    company_number: get(data, 'uk_company.company_number'),
+    registered_address: data.uk_company ? [
+      get(data, 'uk_company.registered_address_1'),
```
ztolley: What happens when address 2 is empty, you get ',,' ?
Don't we have a standard formatter some place for address?

feedmypixel: 👍 moved the logic to the view layer/component

src/apps/investment-projects/router.js
```diff
@@ -73,6 +74,10 @@ router
   )
 
 router
+  .route('/:id/evaluation')
```
teneightfive: Could this be a oneliner as its only using one method?

`router.get('/:id/evaluation', evaluation.renderEvaluationPage)`

config/nunjucks/filters.js
```diff
@@ -35,6 +36,28 @@ const filters = {
     )
   },
 
+  provideDefault: (collection, defaultValue = 'Not found') => {
```
teneightfive: Could this name maybe be more specific? `collectionDefault`?

src/apps/investment-projects/views/evaluation.njk
```diff
@@ -0,0 +1,29 @@
+{% extends "./_layout.njk" %}
+
+{% set verfied = investmentStatus.phaseName | lower == 'won' %}
```
reupen: `phase` was renamed to `stage`, could you check references here?

src/apps/investment-projects/services/formatting.js
```diff
@@ -164,10 +170,73 @@ function transformInvestmentRequirementsForView (data) {
...
+    return
+  }
+
+  const [
+    { name: commitmentToInvestName },
+    { name: fdiName },
+    { name: nonFdiName },
+  ] = metadataRepository.investmentTypeOptions
+
+  return Object.assign({}, {
```
teneightfive: What's the purpose of using `Object.assign()` in this instance?

src/apps/investment-projects/services/formatting.js
```diff
@@ -164,10 +170,73 @@ function transformInvestmentRequirementsForView (data) {
...
+      },
+      {
+        displayText: fdiName,
+        isApproved: data.approved_fdi,
+      },
+      {
+        displayText: nonFdiName,
+        isApproved: data.approved_non_fdi,
+      },
+    ].reduce((displayText, investmentDetail) => {
```
teneightfive: Could this logic be abstracted out so that it's not happening where the object is being returned. 

Feels harder to follow the way it is.

src/apps/investment-projects/services/formatting.js
teneightfive: What's the purpose of using all of these gets? Feels very repetitive.

If it's to filter out null values you could potentially use filter:

```js
filter([
  uk_company. registered_address_1,
  uk_company. registered_address_2,
  uk_company. registered_address_town,
  uk_company. registered_address_country.name,
  uk_company. registered_address_country,
  uk_company. registered_address_postcode,
])
```

src/apps/investment-projects/services/formatting.js
feedmypixel: ah sorry just seen this - yes your right no purpose here

src/apps/investment-projects/services/formatting.js
feedmypixel: yes will look into abstracting it out

PR #332, opened 2017-07-11
Reuse Form components In EntitySearch

tyom:
Ensure consistency by reusing the same form components within other form components.

Also allow configuring form groups and input fields from container components such as `Form`.


src/templates/_macros/form.njk
```diff
@@ -42,9 +44,15 @@
   <form
     method="{{ method }}"
     {% if props.action %}action="{{ props.action }}"{% endif -%}
+    {% if props.class %}class="{{ props.class }}"{% endif %}
+    {% if props.role %}role="{{ props.role }}"{% endif %}
   >
     {{ ErrorSummary(props.errors) }}
-    {{ CSRF(csrfToken) }}
+
+    {% if method|lower === 'post' %}
```
teneightfive: 👍

PR #335, opened 2017-07-11
Add more detailed documentation

teneightfive:
This starts to add some more documentation to explain why the app is structured as it is and to
document patterns that the application is starting to move to and expect.

This will help maintainability moving forward and ensure that there is a consistent experience when working on different parts of the codebase.

### DIFF
--- a/docs/folder-structure.md
+++ b/docs/folder-structure.md
@@ -1,0 +1,81 @@
+# Folder structure
+
+## Assets
+
+The assets folder is used to store client side source code. This code is
+compiled during a build process handled by Webpack to `.build/` which is
+served by the express app.
+
+### JavaScripts
+
+JavaScript source code written using ES2015 syntax and compiled to ES5 using [Babel](https://babeljs.io/).
+
+### Stylesheets
+
+[Sass](http://sass-lang.com/) source code compiled to CSS during build process.
+
+## Common
+
+JavaScript modules shared between both client and server.
+
+## Config
+
+App wide configuration as well as library specific configuration like [nujucks](https://github.com/uktrade/data-hub-frontend/tree/develop/config/nunjucks).
+
+## Docs
+
+Documentation to support product development and maintenance.
+
+## Public
+
+Files that are served directly by the server. Files in public do not go through
+any kind of build step. Normally for third-party libraries or assets that will
+rarely change.
+
+Assets referenced in CSS should not be placed here as Webpack will rewrite
+paths during the build process.
+
+## Src
+
+Server-side source code for the [Express](https://expressjs.com/) application.
+
+### Apps
+
+Collection of Express [sub-apps](http://expressjs.com/en/api.html#app) with
+self-contained logic.
+
+For more information on how these are structured see the [sub-app structure documentation](./sub-apps.md).
+
+### Lib
+
+Server-side only modules.
+
+### Middleware
+
+Application-level middleware that is loaded in [server.js](https://github.com/uktrade/data-hub-frontend/blob/develop/src/server.js).
+
+### Templates
+
+Contains layouts, includes, macros and components that are used throughout the
+views.
+
+#### Components
+
+Components snippets of template with specific context that are rendered using
+a [custom Nunjucks tag](https://mozilla.github.io/nunjucks/api.html#custom-tags).
+
+They resemble macros but do not have to be imported where they are needed.
+
+## Test
+
+Unit and acceptance test files.
+
+### Acceptance
+
+Acceptance tests written using the [Nightwatch](http://nightwatchjs.org/)
+framework to drive the browser through each scenario.
+
+### Unit
+
+Isolated unit tests for controllers, middleware, libraries.
+ Uses the [Mocha](https://mochajs.org/) testing framework.
--- a/docs/sub-apps.md
+++ b/docs/sub-apps.md
@@ -1,0 +1,112 @@
+# Sub-app structure
+
+Sub-apps form the basis of the application. They are self-contained Express
+apps that can be mounted at a particular url.
+
+- [Basic structure](#basic-structure)
+- [File structure](#file-structure)
+
+## Basic structure
+
+Each sub-app should contain a recognised structure. This structure can change
+slightly based on its size.
+
+Each should normally contain:
+
+- an index loader which exports `router`, `displayName` (optional), `mountpath` (optional)
+- a router
+- controllers
+- views
+
+It can also contain:
+
+- middleware
+- repos
+- services
+
+## Components
+
+### Index loader
+
+The loader is responsible for exporting the basic items needed for the sub-app.
+
+- `router` the express router that contains the routes to mount
+- `displayName` [optional] used to create a breadcrumb item for routes within this sub-app
+- `mountpath` [optional] the path all routes will be nested under
+
+### Router
+
+[Express documentation](https://expressjs.com/en/guide/using-middleware.html#middleware.router)
+
+This is where all routes are defined. It sets what middleware and controllers
+should be called for each route and also any sub-app level middleware that should
+be called.
+
+### Controllers
+
+Controllers should be thin and not contain much logic. They should be responsible
+for deciding what action should be taken at the end of a route, normally rendering
+a view but often redirecting to a different URL.
+
+#### Naming convention
+
+TBD
+
+### Views
+
+Views contain markup and components that are required for this route.
+
+### Middleware
+
+Sub-app specific middleware or logic that is required for multiple routes within
+the sub-app.
+
+#### Naming convention
+
+TBD
+
+### API
+
+Contain calls to the API and are responsible for sending/receiving data.
+
+### Transformers
+
+Handle logic required to transform data to and from the API.
+
+## File structure
+
+For small sub-apps it may not warrant splitting the code into folder and files
+but for larger apps it is suggested to help maintain the code.
+
+A small app should start using this folder structure:
+
+```
+views/
+controllers.js
+middleware.js
+index.js
+router.js
+```
+
+As the app grows `controllers`, `middlware`, `repos` and `services` can be
+expanded to a folder structure:
+
+```
+views/
+controllers/
+  details.js
+  edit.js
+  index.js
+  list.js
+middleware/
+  details.js
+  form.js
+  index.js
+index.js
+router.js
+```
+
+When a folder structure is added, index loaders should be added to each folder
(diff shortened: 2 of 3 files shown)


docs/sub-apps.md
```diff
@@ -0,0 +1,110 @@
...
+
+### Controllers
+
+Controllers should be thin and not contain much logic. They should be responsible
+for deciding what action should be taken at the end of a route, normally rendering
+a view but often redirecting to a different URL.
+
+#### Naming convention
+
+TBD
```
teneightfive: We should decide how we want to name controller methods...

docs/sub-apps.md
```diff
@@ -0,0 +1,110 @@
...
+
+### Repos
+
+Contain calls to the API and are responsible for sending/receiving data.
+
+### Services
```
teneightfive: Could these be renamed to `transformers` if that's what they're doing?

docs/folder-structure.md
```diff
@@ -0,0 +1,92 @@
+# Folder structure
+
+The high-level folder structure is:
+
+```
+assets/
```
tyom: Probably don't need to replicate the folder structure here to avoid getting out of sync.

docs/folder-structure.md
```diff
@@ -0,0 +1,92 @@
...
+
+## Assets
+
+The assets folder is used to store client side source code. This code is
+compiled during a build process handled by Webpack to `.build/` which is
+served by the express app.
+
+### Javascripts
+
+Javascript source code written using ES6 syntax and compiled to ES5 using babel.
```
tyom: Babel. And technically it's ES2015.

docs/folder-structure.md
```diff
@@ -0,0 +1,92 @@
...
+
+### Stylesheets
+
+CSS source code written using Sass syntax.
```
tyom: Sass is a language rather than syntax of CSS as implied in that sentence.

docs/folder-structure.md
```diff
@@ -0,0 +1,92 @@
...
+
+## Common
+
+Libraries shared between both client and server.
```
tyom: Not libraries per se, rather JS modules.

docs/folder-structure.md
```diff
@@ -0,0 +1,92 @@
...
+
+## Docs
+
+Documentation to support product development and maintenance.
+
+## Public
+
+Files that are served directly by the server. Files in public do not go through
+any kind of build step. Normally for third-party libraries or assets that will
+rarely change.
```
tyom: Might be worth adding that assets referenced from CSS are not meant to go here since Webpack automatically rewrites URLs in Sass files for those.

docs/folder-structure.md
```diff
@@ -0,0 +1,92 @@
...
+
+## Src
+
+Server-side source code for the express application.
```
tyom: Express

docs/folder-structure.md
```diff
@@ -0,0 +1,92 @@
...
+
+### Apps
+
+Collection of express [sub-apps](http://expressjs.com/en/api.html#app) that contain
```
tyom: Express

docs/folder-structure.md
```diff
@@ -0,0 +1,92 @@
...
+self-contained logic.
```
tyom: "contain self-contained logic" sounds a bit odd.

docs/folder-structure.md
```diff
@@ -0,0 +1,92 @@
...
+
+For more information on how these are structure see the [sub-app structure documentation](./sub-apps.md)
```
tyom: "these are structured"

docs/folder-structure.md
```diff
@@ -0,0 +1,92 @@
...
+
+### Lib
+
+Server-side only libraries.
```
tyom: Maybe drop "only" as not all are strictly server-side.

docs/folder-structure.md
```diff
@@ -0,0 +1,92 @@
...
+
+### Middleware
+
+Application level middleware that is loaded in [server.js](https://github.com/uktrade/data-hub-frontend/blob/develop/src/server.js).
```
tyom: Application-level?

docs/folder-structure.md
```diff
@@ -0,0 +1,92 @@
...
+
+### Templates
+
+Contains layouts, includes, macros and components that are used throughout the
```
tyom: Do we want to explain what components are? Unlike macros they are not Nunjucks terminology. They are a custom tag that we use to render bits of template with specific context. A bit like macros.

docs/sub-apps.md
```diff
@@ -0,0 +1,110 @@
+# Sub-app structure
+
+Sub-apps form the basis of the application. They are self-contained express
```
tyom: Express

docs/sub-apps.md
```diff
@@ -0,0 +1,110 @@
...
+- [File structure](#file-structure)
+
+## Basic structure
+
+Each sub-app should contain a recognised structure. This structure can change
+slightly based on its size.
+
+Each should should contain at least:
+
+- an index loader which exports `displayName`, `mountpath` and `router`
```
tyom: Only `router` is required. The rest are optional.

docs/sub-apps.md
```diff
@@ -0,0 +1,110 @@
...
+
+- [Basic structure](#basic-structure)
+- [File structure](#file-structure)
+
+## Basic structure
+
+Each sub-app should contain a recognised structure. This structure can change
+slightly based on its size.
+
+Each should should contain at least:
```
tyom: Don't think "at least" is appropriate here. Probably better say normally or usually. For example, it's perfectly normal to create a sub-app that groups middleware, or an sub-app that has no views.

docs/sub-apps.md
```diff
@@ -0,0 +1,110 @@
...
+
+## Components
+
+### Index loader
+
+The loader is responsible for exporting the basic items needed for the sub-app.
+
+- `displayName` used to create a breadcrumb item for routes within this sub-app
+- `mountpath` the path all routes will be nested under
+- `router` the express router that contains the routes to mount
```
tyom: Again, only `router` is required.

docs/sub-apps.md
```diff
@@ -0,0 +1,110 @@
...
+
+### Router
+
+This is where all routes are defined. It sets what middleware and controllers
```
tyom: Might be worth linking to [Express docs](https://expressjs.com/en/guide/using-middleware.html#middleware.router) that explains the concept of using sub-stack middleware using Express Router.

docs/sub-apps.md
```diff
@@ -0,0 +1,110 @@
...
+### Middleware
+
+Sub-app specific middleware or logic that is required for multiple routes within
+the sub-app.
+
+#### Naming convention
+
+TBD
+
+### Repos
```
tyom: I wondered if "repo" is the right term. Perhaps "remote" or "api" is more appropriate?

docs/folder-structure.md
teneightfive: These should be server-side only modules as shared modules should be in `common/`

docs/sub-apps.md
teneightfive: Yes, I think `repos` and `services` no longer feel the right terms.

Maybe `api` and `transformers`?

docs/folder-structure.md
```diff
@@ -29,15 +8,15 @@ served by the express app.
 
 ### Javascripts
 
-Javascript source code written using ES6 syntax and compiled to ES5 using babel.
+Javascript source code written using ES2015 syntax and compiled to ES5 using [Babel](https://babeljs.io/).
```
tyom: JavaScript is spelled using Pascal case.

docs/folder-structure.md
```diff
@@ -53,30 +32,40 @@ Files that are served directly by the server. Files in public do not go through
...
 
 Contains layouts, includes, macros and components that are used throughout the
 views.
 
+#### Components
+
+Components snippets of template with specific context that are rendered using
+a [custom Nunjucks tag](https://mozilla.github.io/nunjucks/api.html#custom-tags).
+
+There are a bit like macros but do not have to be imported where they are needed.
```
tyom: They are.

docs/sub-apps.md
teneightfive: @tyom Do we use the plural for api (`apis.js`) though?

teneightfive: Maybe `models` and `transformers`?

tyom: Probably plural, yes. Not sure if `models` work. I think we need to be quiet explicit that it's where network connections are defined.

feedmypixel: It feels like in the current app `repos` define the call to the api, `services` then use that `repo` call to do something. But sometimes that `repo` call is used elsewhere like `middleware` and occasionally `controllers`

How about just calling it `api` so then calling it would become:
- `api.getCompanyInvestmentProjects`

It can then be used in middleware|controllers|services (ideally not in controllers).

docs/sub-apps.md
feedmypixel: Are they always transforming data? `services`. Sometimes they fetch, sometimes they fetch and manipulate. To me feels like the parent of a `transformer`? and `services` is the more generic container.

teneightfive: I think we need to be clearer with what we name things and separate the concerns. At the moment the services mainly seem to transform data and the fetching happens in the _repos_.

I think some repos are also actually transforming data too which is where the confusion happens. Feels having transformers separate from services would be clearer. 

Also, to me services should be for doing more specific things like sending emails or notifications, rather than just having a catchall service for an entity.

docs/sub-apps.md
teneightfive: That's also why I was wondering if using a model would be more clear. So we could have a model for `companies` and we can standardise our methods to things like:

- `companies.findAll()`
- `companies.findById('id')`
- `companies.findOne({filterObject})`
- `companies.save({saveObject})`
- `companies.destroy('id')`

teneightfive: @tyom / @feedmypixel / @ztolley I think this got squashed after a fixup. 

What are the thoughts on using models to save/get data from the API with a consistent method naming convention?

teneightfive:
Being merged as a base for documenting app structure but will need a large refactor to make all sub-apps conform.

PR #337, opened 2017-07-12
New Pagination component

tyom:
Use new Pagination macro in place of old component.

Pagination macro handles truncation and allows to disable page numbers.

## Before
image

## After
image


assets/stylesheets/components/_pagination.scss
```diff
@@ -1,14 +1,32 @@
...
 
-.pagination__item {
-  @include core-font(20);
-  margin: ($gutter / 8);
-  padding: ($gutter / 8);
+.c-pagination__list {
+  display: inline-block;
+}
+
+.c-pagination__item {
```
teneightfive: `.c-pagination__list-item` if you're using `.c-pagination__list`?

assets/stylesheets/layout/_assemblies.scss
```diff
@@ -15,16 +17,9 @@
   }
 
   .c-form-fieldset--subfield,
-  .c-form-group--subfield {
+  .c-form-group--subfield,
+  .c-pagination {
     margin-top: $default-spacing-unit;
     margin-bottom: $default-spacing-unit;
```
teneightfive: Do these need a bottom margin?

teneightfive: https://alistapart.com/article/axiomatic-css-and-lobotomized-owls#snippet5 :trollface:

src/templates/_macros/common.njk
```diff
@@ -0,0 +1,51 @@
+{##
```
teneightfive: Are we moving completely away from custom tags and moving everything to macros?

I can see the benefit for form components as the custom tag doesn't have support for nesting but I don't see a reason this would need to be a macro if we continue to use components.

assets/stylesheets/layout/_assemblies.scss
tyom: They do currently as it's the easiest way to integrate them with non-lobotomised elements. Going forward we should be able to remove all bottom margins as more elements will have top margin defined.

image

src/templates/_macros/common.njk
tyom: I think it's better to have one way of doing things and I don't see the much point in custom tags that try to replicate what macros do.

teneightfive: I think they're neater to maintain and as we talked about the other day it is easier to see from a view that you're calling a component and not just a function. 

But not being able to call child components does mean that they aren't suitable so maybe we should get rid of them.

`common.njk` feels like it will just become a dumping ground though just like `lib/` becomes.

ztolley: So split them up into their own files, and just have a simple script that combines the file and import that, which is exactly how trade elements worked so that macros could be kept separate and easy to maintain whilst still giving the ease of use.

PR #340, opened 2017-07-13
bump node version for latest security release

feedmypixel:


teneightfive:
I know Heroku envs should pick this up but will this also be picked up where our dev/staging/production environments are being deployed? Or will we have to manually bump node?

cc @jim68000

feedmypixel:
> I know Heroku envs should pick this up but will this also be picked up where our dev/staging/production environments are being deployed? Or will we have to manually bump node?

Flynn takes it from the engines property in [package.json](https://flynn.io/docs/languages/nodejs#specifying-a-node.js-version)

PR #344, opened 2017-07-13
Rename investment phase to stage (internally)

reupen:
Phase for investment projects is being renamed in the API to stage to use one term consistently. This changes references in this repo to use stage.

Stage is already present in the API.


reupen:
Just spotting the old investment API endpoints were being used in a couple of places, so have updated those as well.

PR #345, opened 2017-07-13
Tweak layout and header appearance

tyom:
Make structural and visual changes to page header. By making the header visually stand out from from the rest of the content it helps the user to scan the page and contain certain global UI elements (such as search, page-level actions).

Additionally tweak local nav styles by tightening the buttons and removing side border.

Fix a number of issues with inconsistent or missing heading on some of the pages.

### 💥 [Demo](https://datahub-beta2-pr-345.herokuapp.com/) 💥

localhost-3001-search-companies-term samsung

localhost-3001-companies-view-ukother-dcdabbc9-1781-e411-8955-e4115bead28a

localhost-3001-contacts-64f7aa36-0334-4628-a7c3-682328d613f6

localhost-3001-components-local-header

### DIFF
--- a/assets/stylesheets/_deprecated/_errors.scss
+++ b/assets/stylesheets/_deprecated/_errors.scss
@@ -3,6 +3,7 @@
 
 .stack-trace {
   margin-bottom: $gutter;
+  max-width: 100%;
 }
 
 .stack-trace__heading {
--- a/assets/stylesheets/components/_local-header.scss
+++ b/assets/stylesheets/components/_local-header.scss
@@ -1,0 +1,79 @@
+@import "../tools/mixins";
+
+.c-local-header {
+  padding-top: $default-spacing-unit * 4;
+  position: relative;
+
+  &[class*="c-local-header--"] {
+    padding-bottom: $default-spacing-unit * 2;
+  }
+
+  .c-breadcrumb {
+    position: absolute;
+    top: $default-spacing-unit;
+  }
+
+  .c-messages {
+    margin-top: -$default-spacing-unit * 2;
+    margin-bottom: $default-spacing-unit * 2;
+  }
+
+  .status-bar {
+    margin-top: $default-spacing-unit * 2;
+    margin-bottom: -$default-spacing-unit;
+  }
+
+  .c-entity-search__aggregations {
+    margin-bottom: -$default-spacing-unit * 2;
+  }
+
+  &::before {
+    content: "";
+    position: absolute;
+    top: 0;
+    right: 0;
+    left: 0;
+    border-top: 10px solid $govuk-blue;
+    max-width: 100%;
+    width: $site-width;
+    margin: 0 auto;
+
+    @include media($max-width: $site-width) {
+      width: auto;
+      left: $default-spacing-unit;
+      right: $default-spacing-unit;
+    }
+  }
+}
+
+.c-local-header__heading-before {
+  @include core-font(20);
+  color: $grey-1;
+}
+
+.c-local-header__heading {
+  @include bold-font(36);
+  margin-top: 0;
+  margin-bottom: 0;
+}
+
+* + .c-local-header__heading-after {
+  margin-top: $default-spacing-unit / 2;
+  margin-bottom: 0;
+}
+
+.c-local-header--light-banner {
+  background-color: $grey-4;
+}
+
+.c-local-header--dark-banner {
+  background-color: $grey-3;
+
+  &.c-local-header--keyline {
+    border-bottom: 1px solid $grey-2;
+  }
+}
+
+.c-local-header--keyline {
+  border-bottom: 1px solid $grey-3;
+}
--- a/assets/stylesheets/components/_local-header.scss
+++ b/assets/stylesheets/components/_local-header.scss
@@ -26,24 +26,6 @@
   .c-entity-search__aggregations {
     margin-bottom: -$default-spacing-unit * 2;
   }
-
-  &::before {
-    content: "";
-    position: absolute;
-    top: 0;
-    right: 0;
-    left: 0;
-    border-top: 10px solid $govuk-blue;
-    max-width: 100%;
-    width: $site-width;
-    margin: 0 auto;
-
-    @include media($max-width: $site-width) {
-      width: auto;
-      left: $default-spacing-unit;
-      right: $default-spacing-unit;
-    }
-  }
 }
 
 .c-local-header__heading-before {
--- a/assets/stylesheets/components/_local-nav.scss
+++ b/assets/stylesheets/components/_local-nav.scss
@@ -27,7 +27,6 @@
 
   &:hover,
   &:focus {
-    background-color: $grey-4;
     color: $text-colour;
     background-color: $grey-4;
   }
--- a/assets/stylesheets/components/_local-nav.scss
+++ b/assets/stylesheets/components/_local-nav.scss
@@ -3,7 +3,7 @@
 @import "../tools/mixins/layout";
 @import "../tools/mixins/typography";
 
-.local-nav {
+.c-local-nav {
   @include core-font(20);
   margin-left: -($gutter / 2);
   margin-right: -($gutter / 2);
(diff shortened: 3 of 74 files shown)


assets/stylesheets/_deprecated/_errors.scss
```diff
@@ -3,6 +3,7 @@
 
 .stack-trace {
   margin-bottom: $gutter;
+  max-width: 100%;
```
feedmypixel: 👍

assets/stylesheets/components/_local-nav.scss
```diff
@@ -9,7 +9,6 @@
   margin-right: -($gutter / 2);
 
   @include media(tablet) {
-    border-right: 1px solid $grey-2;
```
feedmypixel: 👍

src/templates/_layouts/search.njk
```diff
@@ -13,11 +13,6 @@
 {% endblock %}
 
 
-{% block main_grid_left_column %}
-  &nbsp; {# // TODO filters #}
```
feedmypixel: 👍

assets/stylesheets/components/_local-nav.scss
```diff
@@ -18,16 +17,16 @@
...
   &:visited {
     color: $black;
   }
 
-  @include media(tablet) {
-    padding-left:  (($gutter / 3) * 2);
-    padding-right:  (($gutter / 3) * 2);
+  &:focus {
+    background-color: transparent;
+    text-decoration: underline;
```
teneightfive: Do we need to underline just on focus? Feels unnecessary. 

I wonder if we should consider a hover state on these too.

assets/stylesheets/components/_messages.scss
```diff
@@ -13,6 +15,7 @@
   border: 5px solid;
   padding: $default-spacing-unit 90px $default-spacing-unit $default-spacing-unit;
   position: relative;
+  background-color: $white;
```
teneightfive: 👍

teneightfive: Should we add a white background colour to labels/badges too?

assets/stylesheets/objects/base/_main-content.scss
```diff
@@ -9,16 +9,36 @@
...
+    margin-bottom: -$default-spacing-unit;
+  }
+
+  .c-entity-search__aggregations {
+    margin-bottom: -$default-spacing-unit * 2;
+  }
 
-  @include media(desktop) {
-    padding-bottom: $default-spacing-unit * 4;
+  h1,
```
teneightfive: We should consider using an element classname to stick to BEM to apply this style.

This feels a bit hacky to try and catch all possibilities. We may have a need to a genuine `h1` and `h2` but want to apply different visual styles to them.

assets/stylesheets/objects/base/_main-content.scss
```diff
@@ -9,16 +9,36 @@
   flex-grow: 1;
 }
 
-.main-content__inner {
-  @include site-width-container;
-  border-top: 10px solid $govuk-blue;
+.main-content__header {
```
teneightfive: As discussed the other day, this is starting to feel like its own block rather than an element of `main-content`. Maybe `.local-header`?

Which could have `.local-header__heading` and `.local-header__sub-heading` to apply style to headings.

The `local-header` could then also support a modifier to change the appearance of the background/keyline rather than applying it to `.main-content` or `.main-content__header`.

assets/stylesheets/trumps/_layout.scss
```diff
@@ -1,5 +1,18 @@
 @import "../tools/mixins/layout";
 
+.container {
```
teneightfive: What are your thoughts on using prefixes for layout like we are for components? `.l-`?

src/templates/_layouts/dit-base.njk
```diff
@@ -94,29 +94,34 @@
...
+                    </span>
+                  </div>
                 </div>
               {% endif %}
             {% endblock %}
-            {% block body_main_header %}
-              <header class="main-content__header">
+            <header class="main-content__header">
+              <div class="container">
+                {% block body_main_header_info %}{% endblock %}
```
teneightfive: Maybe better to call this `body_main_header_content_before` as the pattern can then be re-used elsewhere.

ztolley:
Has this been ok'd by design or anyone else yet?

assets/stylesheets/components/_local-header.scss
```diff
@@ -0,0 +1,75 @@
...
+  .status-bar {
+    margin-top: $default-spacing-unit * 2;
+    margin-bottom: -$default-spacing-unit;
+  }
+
+  .c-entity-search__aggregations {
+    margin-bottom: -$default-spacing-unit * 2;
+  }
+
+  &::before {
```
teneightfive: Should this be tied to the local header? The blue border feels like it is a different scope to the local header. It's more of the level of the global header and should still exist even if the local header isn't present.

assets/stylesheets/components/_local-header.scss
```diff
@@ -0,0 +1,75 @@
...
+    @include media($max-width: $site-width) {
+      width: auto;
+      left: $default-spacing-unit;
+      right: $default-spacing-unit;
+    }
+  }
+}
+
+.c-local-header__heading-before {
+  @include bold-font(16);
```
teneightfive: Should this be slightly larger? `16px` feels too small against the `36px` heading. `20` or `24` might be more appropriate.

assets/stylesheets/components/_proposition-menu.scss
```diff
@@ -34,28 +31,28 @@
 .proposition-menu__link {
   color: inherit;
   display: block;
-  padding: $baseline-grid-unit;
+  padding: ($default-spacing-unit / 2) $default-spacing-unit;
   text-decoration: none;
 
+  @include media(tablet) {
+    padding: 0 $baseline-grid-unit;
```
teneightfive: This means the last item doesn't line up with the edge of the blue border any more, right? Was that intentional?

assets/stylesheets/components/_proposition-menu.scss
```diff
@@ -34,28 +31,28 @@
 .proposition-menu__link {
   color: inherit;
   display: block;
-  padding: $baseline-grid-unit;
+  padding: ($default-spacing-unit / 2) $default-spacing-unit;
```
teneightfive: This seems to give them quite a lot more room off the border on smaller devices too. Might want to take another look.

src/apps/contacts/views/_layout.njk
```diff
@@ -1,23 +1,27 @@
...
-      <strong>Reason:</strong> {{contact.archived_reason}}<br>
-      <br>
-      <a href="/contacts/{{contact.id}}/unarchive">Unarchive</a>
-    </p>
-  {% endif %}
+  {% call LocalHeader({
+    headingBefore: headingBefore,
+    heading: contact.first_name + ' ' + contact.last_name,
+    headingSuffix: statusbadge('Primary', variation="fuschia") if contact.primary,
+    breadcrumbs: BREADCRUMBS,
```
teneightfive: This feels very repetitive. Do we need to pass them in every time we want to call the header with breadcrumbs? 

I think it would feel simpler if these 2 weren't tied to each other.

teneightfive: The same applies to messages. If I want to call a variation to the local header on a view I don't want to have to remember to pass in `messages` and `breadcrumbs` to still have those rendered.

src/apps/service-deliveries/controllers.js
```diff
@@ -57,7 +57,7 @@ async function getServiceDeliveryEdit (req, res, next) {
     res.locals.companyUrl = buildCompanyUrl(res.locals.serviceDelivery.company)
 
     res
-      .breadcrumb('Add service delivery')
+      .breadcrumb('Edit service delivery')
```
teneightfive: Will this not change the add route to edit?

src/apps/contacts/views/_layout.njk
tyom: I'm importing `LocalHeader` with context now and default breadcrumbs and messages to those found in locals.

PR #349, opened 2017-07-18
Fix truncation in pagination

tyom:
1. Fixes pagination truncation for pagination block start/end that is next to the first/last pages.

### After

2. Removes page numbers on smaller screens and makes previous/next buttons larger

### DIFF
--- a/assets/stylesheets/components/_pagination.scss
+++ b/assets/stylesheets/components/_pagination.scss
@@ -1,6 +1,24 @@
 @import "../settings";
+@import "../tools/mixins";
+
+@mixin _pageless {
+  .c-pagination__list {
+    display: none;
+  }
+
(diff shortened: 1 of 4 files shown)


assets/stylesheets/components/_pagination.scss
```diff
@@ -1,6 +1,32 @@
...
+@mixin _pageless {
+  .c-pagination__list {
+    display: none;
+  }
+
+  .c-pagination__label--prev,
+  .c-pagination__label--next {
+    float: left;
+    font-size: 24px;
+    background-color: transparent;
```
teneightfive: How come you've dropped the background colour?

tyom: I dropped it for 50% but I guess it could be added back for floated buttons.

PR #350, opened 2017-07-19
Form components tweaks

tyom:
### Before
image

### After
image

### DIFF
--- a/assets/stylesheets/components/form/_multiple-choice.scss
+++ b/assets/stylesheets/components/form/_multiple-choice.scss
@@ -60,6 +60,7 @@
 
     &::before {
       border: 2px solid $text-colour;
+      background-color: $white;
       top: 0;
       left: 0;
     }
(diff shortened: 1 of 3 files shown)


assets/stylesheets/components/form/_multiple-choice.scss
```diff
@@ -82,3 +94,21 @@
...
+
+.c-form-group--small {
+  font-size: 16px;
+
+  .c-multiple-choice__input {
+    @include _scale-input(16px);
+  }
+
+  .c-multiple-choice__label {
+    padding-left: $default-spacing-unit * 2.8;
```
teneightfive: Where has `2.8` come from?

tyom: It's a magic number that is based on relation of the ratio between font/input sizes to the label padding in order to remain consistent in how text is spaced out from the input button. I guess a better formula would be `padding-left: round($default-spacing-unit * 3 * (16 / 19))`. Shall I stick that in?

PR #352, opened 2017-07-19
expose javascript errors and make them human readable

feedmypixel:
We have been loosing node errors to a `404` and that makes it hard to debug. This work exposes the errors by removing the `try/catch` in the `src/apps/routers` file and makes the `uncaughtException` (handled by winston/churchill) human readable.

### DIFF
--- a/config/logger.js
+++ b/config/logger.js
@@ -10,6 +10,7 @@ loggingTransports.push(new (winston.transports.Console)({
   json: false,
   colorize: true,
   handleExceptions: true,
+  humanReadableUnhandledException: true,
 }))
 
 if (config.env === 'production') {
--- a/src/apps/adviser/index.js
+++ b/src/apps/adviser/index.js
@@ -1,0 +1,2 @@
+// TODO adviser will become an app soon, this is a stop gap to avoid routers.js reporting an error
+module.exports = {}
--- a/src/apps/routers.js
+++ b/src/apps/routers.js
@@ -2,24 +2,19 @@ const router = require('express').Router()
 const fs = require('fs')
 
 const { setHomeBreadcrumb } = require('./middleware')
-const logger = require('../../config/logger')
 
 const appsRouters = []
 const subApps = fs.readdirSync(__dirname)
 
 subApps.forEach(subAppDir => {
-  try {
-    const subApp = require(`./${subAppDir}`)
+  const subApp = require(`./${subAppDir}`)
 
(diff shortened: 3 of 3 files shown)


src/repos/adviser.js
```diff
@@ -1,6 +1,6 @@
-const config = require('../../../config')
-const logger = require('../../../config/logger')
-const authorisedRequest = require('../../lib/authorised-request')
+const config = require('../../config')
+const logger = require('../../config/logger')
+const authorisedRequest = require('../lib/authorised-request')
```
feedmypixel: I have moved adviser out of src/apps to src/repos this is a stop gap until a better solution can be found.
a) It isn't an app
b) apps have repos in them and I would not expect to see anything else go into src/repos

As we are discussing what these should be called anyway it felt like it is ok for the moment.

src/apps/companies/services/data.js
```diff
@@ -1,7 +1,7 @@
 /* eslint camelcase: 0 */
 const { get } = require('lodash')
 const logger = require('../../../../config/logger')
-const adviserRepository = require('../../adviser/repos')
+const adviserRepository = require('../../../repos/adviser')
```
teneightfive: Are these paths correct? I thought we removed the `repos` folder?

src/repos/adviser.js
teneightfive: The reason it was moved into an app is because shortly it will _actually_ be an app.

We will have a collection route at `/advisers` like we will for `/companies`, `/investment-projects`, etc.

I would say leave it within apps for now and make the check look for whether there is an index loader for an app as to whether to load the router or not.

teneightfive: I don't think we want to start up the type folders again if we can avoid it. We should find a good way of handling whether to load a feature app or not.

src/apps/companies/services/data.js
feedmypixel: See the comments above and in the PR description. I'm interested in thoughts around how to approach this?

I initially just returned a empty object in `src/apps/adviser/index.js` but that didn't feel right

src/repos/adviser.js
feedmypixel: > an app is because shortly it will actually be an app.

so with knowing that this now feels right:

>I initially just returned a empty object in src/apps/adviser/index.js but that didn't feel right

will amend

PR #355, opened 2017-07-19
Add initial OMIS creation journey

teneightfive:
This creates the basis for the OMIS order creation journey.

It currently supports the company name being passed in as part of the URL which then allows datasets like the contacts to be populated from it.

It support being able to go back to steps within the journey and update the fields.

There is currently no way to get to the journey by navigating pages so you will have to access using the URL: `/omis/create/:companyId`. There is a link to the journey on heroku in the _what it looks like_ section.

## Todo

- [x] Controller/middleware tests
- [x] Reset session after submission

## Not included in this change

- Field validation
- Order view (it will currently `404` after submission)
- Final content (includes placeholder copy)

## What it looks like

[Journey link](https://datahub-beta2-pr-355.herokuapp.com/omis/create/dcdabbc9-1781-e411-8955-e4115bead28a)

omis-create

### DIFF
--- a/src/apps/omis/apps/create/controllers/assign-ita.js
+++ b/src/apps/omis/apps/create/controllers/assign-ita.js
@@ -1,0 +1,17 @@
+const { sortBy } = require('lodash')
+
+const Controller = require('./base')
+const { getAdvisers } = require('../../../../adviser/repos')
+const { transformToOptions } = require('../../../../transformers')
+
+class AssignItaController extends Controller {
+  async configure (req, res, next) {
+    const advisers = await getAdvisers(req.session.token)
+    const options = transformToOptions(advisers.results)
+
+    req.form.options.fields.ita.options = sortBy(options, 'label')
+    super.configure(req, res, next)
+  }
+}
+
+module.exports = AssignItaController
--- a/src/apps/omis/apps/create/controllers/base.js
+++ b/src/apps/omis/apps/create/controllers/base.js
@@ -1,0 +1,25 @@
+const { get, last } = require('lodash')
+const { Controller } = require('hmpo-form-wizard')
+
+class CreateBaseController extends Controller {
+  locals (req, res) {
+    let locals = super.locals(req, res)
+    locals.pageHeading = 'Create order'
+    return locals
+  }
+
+  errorHandler (err, req, res, next) {
+    if (get(err, 'code') === 'MISSING_PREREQ') {
+      const lastStep = last(req.journeyModel.get('history'))
+
+      if (!lastStep) {
+        return res.redirect('/omis/create')
+      }
+      return res.redirect(lastStep.path)
+    }
+
+    super.errorHandler(err, req, res, next)
+  }
+}
+
+module.exports = CreateBaseController
--- a/src/apps/omis/apps/create/controllers/base.js
+++ b/src/apps/omis/apps/create/controllers/base.js
@@ -2,12 +2,6 @@ const { get, last } = require('lodash')
 const { Controller } = require('hmpo-form-wizard')
 
 class CreateBaseController extends Controller {
-  locals (req, res) {
-    let locals = super.locals(req, res)
-    locals.pageHeading = 'Create order'
-    return locals
-  }
-
   errorHandler (err, req, res, next) {
     if (get(err, 'code') === 'MISSING_PREREQ') {
       const lastStep = last(req.journeyModel.get('history'))
--- a/src/apps/omis/apps/create/controllers/company-details.js
+++ b/src/apps/omis/apps/create/controllers/company-details.js
@@ -1,0 +1,31 @@
+const { sortBy } = require('lodash')
+
+const Controller = require('./base')
+const { transformContactsToOptions } = require('../../../../transformers')
+
+class CompanyDetailsController extends Controller {
+  configure (req, res, next) {
+    const company = res.locals.company
+    let contacts = []
+
+    if (company) {
+      contacts = transformContactsToOptions(company.contacts)
+      contacts = sortBy(contacts, 'label')
+    }
+
+    req.form.options.fields.contact.options = contacts
+    super.configure(req, res, next)
+  }
+
+  getValues (req, res, next) {
+    const company = res.locals.company
+
+    super.getValues(req, res, (err, values) => {
+      values.company = company.id
+
+      next(err, values)
+    })
+  }
+}
+
+module.exports = CompanyDetailsController
--- a/src/apps/omis/apps/create/controllers/market.js
+++ b/src/apps/omis/apps/create/controllers/market.js
@@ -1,0 +1,12 @@
+const Controller = require('./base')
+const metadataRepo = require('../../../../../lib/metadata')
+const { transformToOptions } = require('../../../../transformers')
+
+class MarketController extends Controller {
+  configure (req, res, next) {
+    req.form.options.fields.primary_market.options = transformToOptions(metadataRepo.countryOptions)
+    super.configure(req, res, next)
+  }
+}
+
+module.exports = MarketController
--- a/src/apps/omis/apps/create/index.js
+++ b/src/apps/omis/apps/create/index.js
@@ -1,7 +1,7 @@
 const router = require('./router')
 
 module.exports = {
-  displayName: 'Create new order',
+  displayName: 'Create order',
   mountpath: '/create',
   router,
 }
--- a/src/apps/omis/apps/create/index.js
+++ b/src/apps/omis/apps/create/index.js
@@ -1,0 +1,7 @@
+const router = require('./router')
+
+module.exports = {
+  displayName: 'Create new order',
+  mountpath: '/create',
+  router,
+}
--- a/src/apps/omis/middleware/params.js
+++ b/src/apps/omis/middleware/params.js
@@ -1,0 +1,16 @@
+const logger = require('../../../../config/logger')
+const { getInflatedDitCompany } = require('../../companies/services/data')
+
+async function getCompany (req, res, next, companyId) {
+  try {
+    res.locals.company = await getInflatedDitCompany(req.session.token, companyId)
+    next()
+  } catch (e) {
+    logger.error(e)
+    res.redirect('/')
+  }
+}
+
+module.exports = {
+  getCompany,
+}
--- a/test/unit/apps/transformers.test.js
+++ b/test/unit/apps/transformers.test.js
@@ -1,0 +1,71 @@
+const transformers = require('~/src/apps/transformers')
+
+describe('Global transformers', () => {
+  describe('transformToOptions()', () => {
+    context('with default value iteratee', () => {
+      it('should map id and name to value and label', () => {
+        const options = transformers.transformToOptions([{
+          id: '1',
+          name: 'One',
+        }, {
+          id: '2',
+          name: 'Two',
+        }])
+
+        expect(options).to.deep.equal([{
+          value: '1',
+          label: 'One',
+        }, {
+          value: '2',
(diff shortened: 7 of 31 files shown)


assets/stylesheets/components/_summary-table.scss
```diff
@@ -0,0 +1,52 @@
+@import "../settings/colours";
+@import "../settings/layout";
+@import "../tools/mixins/conditionals";
+@import "../tools/mixins/typography";
+
+.c-summary-table {
```
ztolley: How is this different from the existing table styles available

src/apps/omis/apps/create/controllers/base.js
```diff
@@ -0,0 +1,11 @@
+const Controller = require('hmpo-form-wizard').Controller
```
ztolley: I think wee need top all discuss hmpo as this is completely different to how the other forms work and means the code is now inconsistent, also architectural changes like this should be agreed as a team, maybe after a spike and demo

src/apps/omis/apps/create/controllers/company-details.js
```diff
@@ -0,0 +1,33 @@
+const { sortBy } = require('lodash')
```
ztolley: The filename confused me and I thought it was adding a company

src/apps/omis/apps/create/controllers/market.js
```diff
@@ -0,0 +1,16 @@
+const Controller = require('./base')
+const metadataRepo = require('../../../../../lib/metadata')
+
+class MarketController extends Controller {
+  configure (req, res, next) {
+    req.form.options.fields.primary_market.options = metadataRepo.countryOptions.map((item) => {
+      return {
+        value: item.id,
```
ztolley: Always annoys me we have to do this

src/apps/omis/apps/create/controllers/summary.js
```diff
@@ -0,0 +1,41 @@
...
+      values.ita = find(advisers.results, { id: values.ita })
+
+      next(err, values)
+    })
+  }
+
+  saveValues (req, res, next) {
+    const data = req.sessionModel.toJSON()
+
+    // clean unnedeed properties
```
ztolley: unnedeed ?

src/apps/omis/apps/create/controllers/summary.js
```diff
@@ -0,0 +1,41 @@
...
+    unset(data, 'csrf-secret')
+    unset(data, 'errors')
+
+    Order.save(req.session.token, data)
+      .catch(next)
+      .then((order) => {
+        console.log('RESETTING SESSION')
```
ztolley: don't we have a logger?

src/apps/omis/apps/create/index.js
```diff
@@ -0,0 +1,7 @@
+const router = require('./router')
+
+module.exports = {
+  displayName: 'Create new order',
+  mountpath: '/create',
```
ztolley: mountPath?

src/apps/omis/middleware/params.js
```diff
@@ -0,0 +1,16 @@
+const logger = require('../../../../config/logger')
+const { getInflatedDitCompany } = require('../../companies/services/data')
+
+async function getCompany (req, res, next, companyId) {
+  try {
+    res.locals.company = await getInflatedDitCompany(req.session.token, companyId)
```
ztolley: Do you REALLY need the inflated version? We actually need to phase this method out as it's no longer needed and when you need details for contacts etc this should be a separate call.

src/apps/omis/middleware/params.js
```diff
@@ -0,0 +1,16 @@
...
+    next()
+  } catch (e) {
+    logger.error(e)
+    res.redirect('/')
```
ztolley: not next(e)?

src/apps/omis/views/create/start.njk
```diff
@@ -0,0 +1,15 @@
+{% extends "_layouts/form-wizard-step.njk" %}
+
+{% block form %}
+  {% call MultiStepForm({
+    buttonText: 'Start'
+  }) %}
+    <p>You will need to be aware of these things before you start this process:</p>
+
+    <ul class="list-disc">
```
ztolley: namespace the class?

teneightfive:
>We all need to discuss HMPO as a team

Agreed but it is moving the direction of building forms dynamically that we have discussed. It would be more efficient that manually creating fields in views.

>There is no way of knowing where you are in the process and knowing what back takes you to

This is out of scope for this change and we have a ticket to track the problem of journey navigation. There is also some [research](https://www.gov.uk/service-manual/design/progress-indicators) done around GOV.UK journeys that [progress indicators](https://paper.dropbox.com/doc/Progress-indicators-FKi7F1KBjks0jGG5vMdVg) don't actually add much value so maybe we start with it off and see if it causes users problems.

assets/stylesheets/components/_summary-table.scss
teneightfive: It specifically adjusts the layout to be stacked for smaller devices which the current table styling doesn't cater for.

It also applies style to allow the controls to be consistently placed and aligned.

src/apps/omis/apps/create/controllers/base.js
teneightfive: Yes but this is also an interim replacement and being developed and maintained by a separate team so we shouldn't be restricted completely. 

It also moves to a form builder architecture which will make maintaining journeys within this app more straightforward.

src/apps/omis/apps/create/controllers/company-details.js
teneightfive: Not sure what would be a better name? It's a controller for adding company details as part of this journey.

src/apps/omis/apps/create/controllers/market.js
teneightfive: Completely agreed. Would like to find a better way to format these.

src/apps/omis/apps/create/controllers/summary.js
teneightfive: Ah. Good spot. This needs addressing. And needs to actually reset the session.

src/apps/omis/apps/create/index.js
teneightfive: This matches the [naming convention](http://expressjs.com/en/api.html#app.mountpath) used by Express and also used in our other sub-apps.

src/apps/omis/middleware/params.js
teneightfive: Fair point. I was using the methods that have previously been used. Will see if there is an alternative.

src/apps/omis/middleware/params.js
teneightfive: We may need to specifically handle this but I didn't want to present the user with a `404` from the API when this catch happens.

This will be addressed as we explore other entry points into this journey.

src/apps/omis/views/create/start.njk
teneightfive: This is using the pattern we currently adopt and personally I'd rather we don't have to apply a class to re-apply the default list style to a list.

But when I applied that style it had a lot of knock-on affects. We should really have a typography and style review sooner rather than later.

src/apps/omis/apps/create/controllers/market.js
tyom: Maybe it can be moved into a helper function somewhere in the lib?

assets/stylesheets/components/_summary-table.scss
```diff
@@ -0,0 +1,53 @@
+@import "../settings/colours";
```
tyom: Do we need this granularity? You could just import `../settings` and `../tools` or even just `../tools` which imports settings.

src/apps/omis/apps/create/controllers/base.js
tyom: `const { Controller } = require('hmpo-form-wizard')`?

src/apps/omis/apps/create/controllers/assign-ita.js
```diff
@@ -0,0 +1,21 @@
+const { sortBy } = require('lodash')
+
+const Controller = require('./base')
+const { getAdvisers } = require('../../../../adviser/repos')
+
+class AssignItaController extends Controller {
+  async configure (req, res, next) {
+    const advisers = await getAdvisers(req.session.token)
+    const options = advisers.results.map((adviser) => {
```
tyom: I'd suggest extracting this function into a standalone one. Maybe in transformers?

src/apps/omis/apps/create/controllers/company-details.js
```diff
@@ -0,0 +1,36 @@
...
+
+class CompanyDetailsController extends Controller {
+  configure (req, res, next) {
+    const company = res.locals.company
+    let contacts = []
+
+    if (company) {
+      contacts = company.contacts.map((option) => {
+        return {
+          value: option.id,
```
tyom: This is popping up everywhere. I reckon we should move it into a global transformer. In a separate PR I guess.

src/apps/omis/middleware/params.js
```diff
@@ -0,0 +1,16 @@
+const logger = require('../../../../config/logger')
+const { getInflatedDitCompany } = require('../../companies/services/data')
+
+async function getCompany (req, res, next, companyId) {
```
tyom: Should this go in models?

src/apps/omis/apps/create/controllers/company-details.js
teneightfive: I have extracted out to an OMIS level transformer but maybe feels like it should live higher level and could be used in lots of other places. Maybe the changing existing behaviour could be a separate PR but will add it with this one now.

src/apps/omis/middleware/params.js
teneightfive: Maybe if we created a model within `companies` but I suppose we haven't firmed up that decision yet, right?

teneightfive: These param level functions that basically set things on `res.locals` feel like they could start living higher too and being used in multiple apps to avoid repeating setting these values.

Definitely different change though.

src/apps/omis/transformers.js
```diff
@@ -0,0 +1,23 @@
+const { isArray } = require('lodash')
+
+function transformToOptions (items, labelIteratee = (item) => item.name) {
```
tyom: Should we move it level up to global transformers to other apps can use it?

src/apps/omis/transformers.js
```diff
@@ -0,0 +1,23 @@
...
+  if (!isArray(items)) { return [] }
```
tyom: Should it return empty array if it's not an array to begin with? It could potentially swallow a user error.

teneightfive: Fair point. Will remove this.

src/apps/omis/middleware/params.js
tyom: Do we need this in `params` file? They seem rather small, could they live in one `middleware.js` file?

test/unit/apps/transformers.test.js
```diff
@@ -1,4 +1,4 @@
-const transformers = require('~/src/apps/omis/transformers')
+const transformers = require('~/src/apps/transformers')
 
 describe('OMIS transformers', () => {
```
tyom: They are not OMIS anymore.

tyom: Global data transformers?

teneightfive: Damn. Good spot.

PR #357, opened 2017-07-20
Form component improvements

tyom:
- Improve styling of select box (including fixes for IE9)
- Fix inline form-group styling for various use cases
- Remove top margin from form-group if it follows hidden input
- Allow multiple modifiers to be passed to form components (e.g. `modifier: 'inline'` and `modifier: ['inline', 'small']`)

### DIFF
--- a/assets/stylesheets/components/form/_form-control.scss
+++ b/assets/stylesheets/components/form/_form-control.scss
@@ -40,6 +40,7 @@ select.c-form-control {
   appearance: none;
   background: $white url("../../../images/icon-triangle.svg") no-repeat right 8px top 50%;
   background-size: .8em;
+  padding-right: $default-spacing-unit * 2;
 
   &:-moz-focusring {
     color: transparent;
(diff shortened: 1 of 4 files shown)


assets/stylesheets/components/form/_form-control.scss
```diff
@@ -48,6 +49,14 @@ select.c-form-control {
   &::-ms-expand {
     display: none;
   }
+
+  // IE9
+  @media all and (min-width:0\0) and (min-resolution:.001dpcm) {
```
ztolley: Wow what the heck is that syntax?

src/templates/_macros/form.njk
```diff
@@ -415,6 +426,7 @@
  # @param {string} [props.value] - Checked field value
  # @param {string} [props.children] - Main field subfields
  # @param {string} [props.options[].subfield] - Field options subfields
+ # @param {string} [props.initialOption] - Initial option label to use for empty value in select box
```
ztolley: 👍

assets/stylesheets/components/form/_form-control.scss
jim68000: min width is 0 plus one null byte?

tyom: That hack is supposed to only target IE9. Saves creating a separate stylesheet for it. It's only been needed here so far.

PR #358, opened 2017-07-20
Form group modifiers

tyom:
- Pass field modifier from EntitySearchForm
- Pass field modifier to `multiple-choice` field
- Render `initialOption` for MultipleChoiceField for `radio` type
- Add `light` and `smaller` modifier for form fields
- Add `actions` modifier for form-group container for submit button

image

### DIFF
--- a/assets/stylesheets/components/form/_form-group.scss
+++ b/assets/stylesheets/components/form/_form-group.scss
@@ -100,3 +100,7 @@
     border-color: $black;
   }
 }
+
+.c-form-group--small {
+  font-size: 16px;
+}
--- a/assets/stylesheets/components/form/_form-group.scss
+++ b/assets/stylesheets/components/form/_form-group.scss
@@ -104,3 +104,7 @@
 .c-form-group--small {
   font-size: 16px;
 }
+
+.c-form-group--smaller {
+  font-size: 14px;
+}
--- a/assets/stylesheets/components/form/_form-group.scss
+++ b/assets/stylesheets/components/form/_form-group.scss
@@ -87,3 +87,16 @@
     }
   }
 }
+
+.c-form-group--light {
+  .c-form-control,
+  .c-multiple-choice__input + .c-multiple-choice__label::before {
+    border: 1px solid $grey-1;
+  }
+
+  .c-form-control:focus,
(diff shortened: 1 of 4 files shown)


src/templates/_macros/form.njk
```diff
@@ -431,11 +438,28 @@
...
     {% set fieldChildren = props.children or [] %}
+    {% set modifier = props.modifier | concat('') | reverse | join(' c-multiple-choice--') if props.modifier %}
+
+    {% if props.type == 'radio' and props.initialOption %}
+      <div class="c-multiple-choice {{ modifier }}">
+        <input
+          class="c-multiple-choice__input"
+          type="{{ props.type }}"
+          name="{{ props.name }}"
+          id="{{ props.fieldId }}-0"
```
feedmypixel: potential for multiples of this on a page?

assets/stylesheets/components/form/_form-group.scss
```diff
@@ -87,3 +87,24 @@
...
+
+  .c-form-control:focus,
+  .c-multiple-choice__input:checked + .c-multiple-choice__label::before,
+  .c-multiple-choice__input:focus + .c-multiple-choice__label::before {
+    border-color: $black;
+  }
+}
+
+.c-form-group--small {
+  font-size: 16px;
```
feedmypixel: are we not using the `core-font` work?

assets/stylesheets/components/form/_form-group.scss
```diff
@@ -87,3 +87,24 @@
...
+}
+
+.c-form-group--smaller {
+  font-size: 14px;
```
feedmypixel: are we not using the `core-font` work?

assets/stylesheets/components/form/_form-group.scss
tyom: The plan is to update the font mixins to do less. They are doing to much, but they really should be just setting font size and line height for different screen sizes. I'm hardcoding them for now with the hope that we can address the typography globally very soon.

src/templates/_macros/form.njk
tyom: This is always one as it is outside of the loop and is used for showing `initialOption` (see select box).

PR #360, opened 2017-07-21
Fixes 404 in add Company House company flow

jim68000:
This is just to unblock the ability to create companies. It reveals two further issues which will be dealt with in future commits: no labels on the company detail panel, and no error messages when an invalid DH company is sent to the back end. But I have other blockers to clear first.

### DIFF
--- a/src/apps/companies/router.js
+++ b/src/apps/companies/router.js
@@ -42,6 +42,11 @@ router
   .post(foreignController.postDetails)
 
 router
+  .route('/add/ltd/:company_number')
+  .get(foreignController.editCommon, ltdController.addDetails)
+  .post(foreignController.postDetails)
+
+router
   .route('/add/ltd')
   .get(foreignController.editCommon, ltdController.addDetails)
   .post(foreignController.postDetails)
--- a/assets/stylesheets/components/form/_form-group.scss
+++ b/assets/stylesheets/components/form/_form-group.scss
@@ -87,3 +87,24 @@
     }
   }
 }
+
+.c-form-group--light {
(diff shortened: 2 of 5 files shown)


src/apps/companies/controllers/ltd.js
```diff
@@ -54,8 +54,9 @@ function editCommon (req, res, next) {
 }
 
 async function addDetails (req, res, next) {
+  const companyNumber = req.params.company_number || req.params.id
```
feedmypixel: is `req.params.company_number` still a thing here? its just `req.params.id` right?

src/apps/companies/router.js
```diff
@@ -42,6 +42,11 @@ router
   .post(foreignController.postDetails)
 
 router
+  .route('/add/ltd/:id')
```
feedmypixel: using `.route('/add/ltd/:company_number')` could give a little more clarity?

src/apps/companies/controllers/ltd.js
jim68000: There's another route that uses this without the :id path param, but I don't know if it actually gets called, so maybe I'm being  over cautious. Will investigate.

src/apps/companies/router.js
jim68000: Agreed, will change

src/apps/companies/controllers/ltd.js
jim68000: Actually, change below fixes this...on it

jim68000:
Mucked up. Sorry.

PR #361, opened 2017-07-21
fixes failure in add company flow

jim68000:
Trying again after I managed to mess up rebasing

### DIFF
--- a/src/apps/companies/router.js
+++ b/src/apps/companies/router.js
@@ -42,6 +42,11 @@ router
   .post(foreignController.postDetails)
 
 router
+  .route('/add/ltd/:company_number')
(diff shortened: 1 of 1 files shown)


src/apps/companies/router.js
```diff
@@ -42,6 +42,11 @@ router
   .post(foreignController.postDetails)
 
 router
+  .route('/add/ltd/:company_number')
+  .get(foreignController.editCommon, ltdController.addDetails)
+  .post(foreignController.postDetails)
+
+router
   .route('/add/ltd')
```
jim68000: Is this route used?

feedmypixel: doesn't look like it

PR #362, opened 2017-07-21
restores Company as option in add foreign co flow

jim68000:


jim68000:
Pull yourself together smith

PR #367, opened 2017-07-21
Updated docker config

ztolley:
Updated to the docker file to bring in the latest version of node and add yarn and use yarn to install dependencies inline with new build methods.

* Fixed incorrect syntax in .dockerignore due to simply copying .gitignore
file
* Changed docker file to get both an up to date version of node and also install yarn, then use yarn to get dependencies.
* Changed docker-compose to create separate processes for the app server and webpack watcher
* Changed how docker compose starts the app server to address issue with requiring use of poller to watch files
* Created a new webpack config for docker get around dockers need for polling and set to sensible polling interval.
* Updated volume mounts in docker-compose to reflect updated directory structure


package.json
```diff
@@ -87,7 +87,7 @@
     "develop": "run-p watch:js:*",
     "watch:test": "npm run test:unit -- -w",
     "watch:js:client": "webpack --watch --progress",
-    "watch:js:server": "nodemon $NODE_DEBUG_OPTION --inspect",
+    "watch:js:server": "nodemon $NODE_DEBUG_OPTION -L --inspect",
```
tyom: This legacy polling is Docker-specific. It probably should be the default watch.

From nodemon docs:
> Though this should be a last resort as it will poll every file it can find.

ztolley: How would you add it for docker support? 

I did wonder if we should just change the docker compose start command but wasn't sure how to do that so it starts up the two things in parallel. Another options is setup the docker compose file to use separate 2 containers to run the server and webpack watch. I have done that elsewhere.

I tested this config and it didn't result in high CPU on windows but can't say what Mac would do.

tyom: I don't know how but I don't want this sneaked in, likely degrading performance for other non-Docker users. Need to find another way.

ztolley: I tested and it didn't but I will find another way. Just wanted to get a feed if you prefer using 2 containers (1 server, 1 webpack watcher) or 1 container using some sort of parallel launcher

webpack.config.js
```diff
@@ -132,8 +134,30 @@ const prod = merge.smart(common, {
...
+        '.build/images/*',
+        'src/**/*.njk',
+      ],
+    }, {
+      reload: false,
+    }),
+  ],
+})
+
+const webpackEnv = process.env.WEBPACK_ENV || (config.isProd) ? 'prod' : 'develop'
```
teneightfive: Is `process.env.WEBPACK_ENV` a boolean or can it be a string? If a boolean might be better to update the name to something clearer like `IS_[...]`.

Also might need to update the list of environment variables in the readme to explain what this is used for.

ztolley: It's a string. If it is present it uses it, if not then it falls back to the previous method to pick the env. By default people wont set it.

It's just a mechanism used to pick up the fact we are using docker, users/developers wouldn't use it.

teneightfive: Do you need to wrap the ternary in brackets rather than the config value? 

`const webpackEnv = process.env.WEBPACK_ENV || (config.isProd ? 'prod' : 'develop')`

teneightfive: Do you want to add that option to the list of environment variables in the readme too

webpack.config.js
```diff
@@ -132,8 +134,30 @@ const prod = merge.smart(common, {
   ],
 })
 
-if (config.isProd) {
-  module.exports = prod
-} else {
-  module.exports = develop
-}
+webpackConfigs.docker = merge.smart(common, {
```
tyom: Could extend `develop` instead to avoid repeating the `output` and `plugins` prop.

README.md
```diff
@@ -61,6 +61,7 @@ This file expects the following environment variables:
 | ZEN_BROWSER | Zendesk browser ID |
 | ZEN_IMPACT | Zendesk impact ID |
 | ZEN_SERVICE | Zendesk service ID |
+| WEBPACK_ENV | Optionally select the webpack configuration variation to use, the default will correctly pick a production or development config based on NODE_ENV |
```
teneightfive: Should we mention the available options as by the looks of it only `develop`, `docker` and `prod` would be accepted.

README.md
```diff
@@ -61,6 +61,7 @@ This file expects the following environment variables:
 | ZEN_BROWSER | Zendesk browser ID |
 | ZEN_IMPACT | Zendesk impact ID |
 | ZEN_SERVICE | Zendesk service ID |
+| WEBPACK_ENV | Optionally select the webpack configuration variation to use, the default will correctly pick a production or development config based on NODE_ENV. Value values include `prod`, `develop` and `docker` |
```
teneightfive: Think you may have meant `Valid values`

PR #368, opened 2017-07-21
Add update method to breadcrumbs

feedmypixel:
## Problem
The `investment projects` breadcrumb does not link to the `investment projects` page when you are in an investment project.

## Solution
When you are in an Investment project we know the company that the investment project is associated with. This work enables us to update the breadcrumb with the `investor_company.id` so users can navigate back to the parent company via the breadcrumb.

### This work involves:

- add `update` method to breadcrumbs to be able to change a breadcrumbs entry,
- introduce `add` and `get` methods onto `res.breadcrumb` to provide prior functionality in a clearer fashion,
- update all tests,
- introduce a way of merging mock response information so `res.breadcrumb` methods can be chained in tests

### Old
old

### New
new

### DIFF
--- a/src/middleware/breadcrumbs.js
+++ b/src/middleware/breadcrumbs.js
@@ -27,6 +27,21 @@ function init () {
     return findIndex(breadcrumbs, breadcrumb) !== -1
   }
 
+  function updateBreadcrumb (existingName, { name, url }) {
+    const index = findIndex(breadcrumbs, (breadcrumb) => {
+      return breadcrumb.name.toLowerCase() === existingName.toLowerCase()
+    })
+
+    if (index >= 0) {
+      if (name) {
+        breadcrumbs[index].name = name
+      }
+      if (url) {
+        breadcrumbs[index].url = url
+      }
+    }
+  }
+
   function addBreadcrumbs (name, url) {
     if (arguments.length === 1) {
       if (isArray(name)) {
@@ -51,11 +66,11 @@ function init () {
           url: url,
         })
       }
-    } else {
-      return breadcrumbs
     }
+  }
 
-    return this
+  function getBreadcrumbs () {
+    return breadcrumbs
   }
 
   function cleanBreadcrumbs () {
@@ -64,7 +79,19 @@ function init () {
 
   return function (req, res, next) {
     cleanBreadcrumbs()
-    res.breadcrumb = addBreadcrumbs
+
+    res.breadcrumb = {
+      add: (...args) => {
+        addBreadcrumbs(...args)
+        return res
+      },
+      get: getBreadcrumbs,
+      update: (existingName, { name, url }) => {
+        updateBreadcrumb(existingName, { name, url })
+        return res
+      },
+    }
+
     next()
   }
 }
@@ -89,12 +116,12 @@ function setHome (options = {}) {
   const homeUrl = options.url || '/'
 
   return function (req, res, next) {
-    const homeBreadcrumb = find(res.breadcrumb(), (breadcrumb) => {
+    const homeBreadcrumb = find(res.breadcrumb.get(), (breadcrumb) => {
       return breadcrumb._home
     })
 
     if (!homeBreadcrumb) {
-      res.breadcrumb({
+      res.breadcrumb.add({
         name: homeName,
         url: homeUrl,
         _home: true,
--- a/test/unit/apps/companies/controllers/add.test.js
+++ b/test/unit/apps/companies/controllers/add.test.js
@@ -13,6 +13,13 @@ describe('Company add controller', function () {
   let companyAddController
 
   beforeEach(function () {
+    this.resMock = {
+      breadcrumb: {
+        add: () => this.resMock,
+        update: () => this.resMock,
+        get: () => [],
+      },
+    }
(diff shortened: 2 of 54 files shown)


feedmypixel:
will wait for @teneightfive 👀  👀  👀  👀  on this one, but could do with getting it in ASAP as it touches many files.

teneightfive:
I have a couple comments around the thought behind this PR. Not sure it feels right:

### Hierarchy

>On Investment projects we know the company that the investment project is associated with. This work enables us to update the breadcrumb so users can navigate back to the parent company if they wish to do so.

I'm not sure this feels right to me now as breaks the structure we use elsewhere around _collections_ being the top level breadcrumb item as these are the pots that an item effectively sits in. 

I think we have a link to the company within a project which provides a way to navigate to that company.

This now feels inconsistent between contacts and investment projects too as a contact can be attached to a company but we're not updating the breadcrumb for these.

For me it breaks the clear hierarchy we have around items being part of a collection. I see that they are related to a company but I'm not sure the breadcrumb should show that relation when you're in an item.

### Update method

>add update method to breadcrumbs to be able to change a breadcrumbs entry

What's the purpose of wanting to update an item? Are we keeping the original value anywhere else within the investment projects sub-app on certain routes but not others? Why not just change that value if we do decide to update that breadcrumb item.

### Company investment projects

This change seems to have broken the relationship for investment projects:

**Company details**
screen shot 2017-07-24 at 10 40 14


**Company investment projects**

screen shot 2017-07-24 at 10 40 32

**Investment project**

screen shot 2017-07-24 at 10 40 41

That jump feels stranger to me that keeping companies nested within companies and keeping investment projects nested within investment projects.

src/middleware/breadcrumbs.js
```diff
@@ -64,7 +74,19 @@ function init () {
 
   return function (req, res, next) {
     cleanBreadcrumbs()
-    res.breadcrumb = addBreadcrumbs
+
+    res.breadcrumb = {
+      add: (...args) => {
```
teneightfive: This approach of nesting methods feels a little strange. Might be worth chatting about this.

I think it would be good to understand the thoughts behind needing an update method. I'm not convinced we should be updating items only for certain routes which might mean we don't need this new style.

test/unit/apps/companies/controllers/add.test.js
```diff
@@ -13,6 +13,13 @@ describe('Company add controller', function () {
   let companyAddController
 
   beforeEach(function () {
+    this.resMock = {
```
teneightfive: We should probably try and standardise some of our testing approach. I've noticed as we've been updating files we're now using a mix between declaring variables outside of this scope and setting them in here or setting properties on `this`.

feedmypixel:
> I'm not sure this feels right to me now as breaks the structure we use elsewhere around collections being the top level breadcrumb item as these are the pots that an item effectively sits in.

Agreed, that pot is also the company

> I think we have a link to the company within a project which provides a way to navigate to that company.

I couldn't see one that is one of the reasons I made this change. The other is because the `investment projects` breadcrumb link was in need of an `id` and when adding this `id` it then felt sensible to change the name to the company's name.
 
> This now feels inconsistent between contacts and investment projects too as a contact can be attached to a company but we're not updating the breadcrumb for these.

We could update these to reflect this pattern to be consistent.

> For me it breaks the clear hierarchy we have around items being part of a collection. I see that they are related to a company but I'm not sure the breadcrumb should show that relation when you're in an item.

Agree, yeah I see what your saying here

> Update method
> 
> add update method to breadcrumbs to be able to change a breadcrumbs entry
> What's the purpose of wanting to update an item? Are we keeping the original value anywhere else within the investment projects sub-app on certain routes but not others? Why not just change that value if we do decide to update that breadcrumb item.

We don't know the investment project id in the `investment-projects` index.js

> Company investment projects
> 
> This change seems to have broken the relationship for investment projects:
> 

I can have a look at that

src/middleware/breadcrumbs.js
feedmypixel: It was to clearly show that `res` was being returned for chaining. Also due to the methods being `add`, `update`, `get` being in an object literal on `res.breadcrumb` the ability to return `this` in the actual methods and refer to `res` is no longer possible.

test/unit/apps/companies/controllers/add.test.js
feedmypixel: yeah agreed - it felt very much like we should have a global mock of `res.breadcrumb` for testing

feedmypixel:
Thanks for the feedback - lets catch up when we are both in the office. We need to add the `investment project` id into the breadcrumb so it correctly goes to the parent page. If that means we don't change the links name to the company name for the reasons you have explained above then thats fine with me.

teneightfive:
@feedmypixel yeah lets catch up face-to-face. Maybe we just need to fix them when within sections of an investment project. But to me, updating it to be the company breaks the idea of collections and that being the line of hierarchy for entities. 

There will always be relations so maybe we need a different way to show that. Could be something we use from the potential new local header styles in #345 and use the idea of a context subheading above the main record title.

feedmypixel:
@teneightfive I've updated the description to reflect the changes in this PR after discussion. Let me know when your free to talk through this work.

feedmypixel:
Closing after conversation with @teneightfive this work is not doing the correct thing.

PR #343, opened 2017-07-13
Investment collection (pipeline)

tyom:
This creates collection page for investment projects with basic filters.

TODO:
- [x] Add tests
- [x] Add client-side behaviour to improve filtering experience in #373
- [x] Add global navigation item
- [x] Add sorting

### 💥 [Demo](https://datahub-beta2-pr-343.herokuapp.com/investment-projects) 💥

localhost-3001-investment-projects-sortby total_investment 3adesc stage 8a320cc9-ae2e-443e-9d26-2f36452c2ced investment_type 3e143372-496c-4d1e-8278-6fdd3da9b48b estimated_land_date_before estimated_land_date_after

PR #372, opened 2017-07-22
LocalNav component

tyom:
- Add LocalNav macro
- Replace existing hard-coded local navs with macro on investment projects and contacts page

Depends on #343

### DIFF
--- a/src/apps/companies/views/_layout-view.njk
+++ b/src/apps/companies/views/_layout-view.njk
@@ -20,15 +20,15 @@
 
 
 {% block main_grid_left_column %}
-  <nav class="local-nav">
-    <a class="local-nav__link {% if tab == 'details' %}is-active{% endif %}" href="{{companyUrl}}">Company details</a>
+  <nav class="c-local-nav">
+    <a class="c-local-nav__link {% if tab == 'details' %}is-active{% endif %}" href="{{companyUrl}}">Company details</a>
     {% if company.id %}
-      {# <a class="local-nav__link {% if tab == 'hierarchy' %}is-active{% endif %}" href="">Company hierarchy</a> #}
-      <a class="local-nav__link {% if tab == 'contacts' %}is-active{% endif %}" href="/companies/{{company.id}}/contacts">Contacts</a>
-      <a class="local-nav__link {% if tab == 'interactions' %}is-active{% endif %}" href="/companies/{{company.id}}/interactions">Interactions</a>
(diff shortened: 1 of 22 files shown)


src/apps/companies/views/_layout-view.njk
```diff
@@ -21,15 +21,15 @@
 
 
 {% block main_grid_left_column %}
-  <nav class="local-nav">
-    <a class="local-nav__link {% if tab == 'details' %}is-active{% endif %}" href="{{companyUrl}}">Company details</a>
+  <nav class="c-local-nav">
```
teneightfive: Don't you want to use the macro for this?

src/middleware/local-nav.js
```diff
@@ -0,0 +1,16 @@
+const path = require('path')
+
+function localNavMiddleware (items = []) {
+  return function buildLocalNav (req, res, next) {
+    res.locals.localNavItems = items.map(item => {
+      const url = path.resolve(res.locals.CURRENT_PATH, item.path)
+      return Object.assign(item, {
+        url,
+        isActive: res.locals.CURRENT_PATH === url,
```
teneightfive: Should this not be done in the macro as it's only used for rendering purposes?

When setting a list of items to view locals it probably doesn't care about this.

src/apps/companies/views/_layout-view.njk
tyom: As I explained in hangouts it's not straightforward to use this macro for local nav in companies. It would have to be a separate PR.

src/middleware/local-nav.js
tyom: But this is a generic middleware which creates new `url` property from `path` and sets `isActive`. Otherwise macro would have to expose `path.resolve` to do this operation.

teneightfive: Yeah that bit makes sense, but was wondering whether you need to set the `isActive` property within the middleware or whether that can be set within the macro as that's the only place which is concerned with it.

src/apps/companies/views/_layout-view.njk
ztolley: We so need to refactor company. I'll dig out my plan and try Again to convince that we need to do it and figure out how I can do it a piece at a time

src/middleware/local-nav.js
tyom: Fair enough. Updated.

tyom: I changed my mind. I think it's better to avoid side effects in macros (relying on external properties) for the sake of its testability and encapsulation.

PR #373, opened 2017-07-23
Add client-side behaviour to investment projects filters

tyom:
Adds two new JS components:
- js-xhr: makes links with this class use XHR mechanism to update the page
- js-AutoSubmit - makes form with this class to automatically submit on change or enter

Adds new Nunjucks layout for XHR mechanism, partially updating the page instead of full reload.

Depends on #343.

### 💥 [Demo](https://datahub-beta2-pr-373.herokuapp.com/investment-projects) 💥

investment-filters


tyom:
> It's big and might help to have a run through of the thinking behind how it works.

The client-side is actually not that big on its own. It's exploded due to the fact that its underlying branch has gone out of sync (fixups). It will shrink loads when I rebase.

src/server.js
```diff
@@ -53,6 +53,11 @@ app.use(csrfToken())
 
 app.use(compression())
 
+app.use((req, res, next) => {
+  res.locals.isXhr = req.xhr
```
teneightfive: Why not add to the [locals middleware](https://github.com/uktrade/data-hub-frontend/blob/develop/src/middleware/locals.js)?

src/templates/_layouts/dit-base.njk
```diff
@@ -113,7 +113,9 @@
             {% endblock %}
           {% endblock %}
           <div class="main-content__inner l-container">
-            {% block body_main_content %}{% endblock %}
+            <div id="xhr-outlet">
```
teneightfive: Does this element get replaced or does it replace the contents? Could add the id to `main-content__inner` if it doesn't get replaced?

assets/stylesheets/trumps/_utility.scss
```diff
@@ -45,3 +45,30 @@
     float: right !important;
   }
 }
+
+.u-loading {
+
```
teneightfive: Remove line?

src/server.js
tyom: True.

src/middleware/locals.js
```diff
@@ -18,6 +18,7 @@ module.exports = function locals (req, res, next) {
     CURRENT_PATH: req.path,
     GOOGLE_TAG_MANAGER_KEY: config.googleTagManagerKey,
     BREADCRUMBS: breadcrumbItems,
+    isXhr: req.xhr,
```
teneightfive: Const case?

assets/javascripts/lib/xhr.js
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    return axios
+      .get(url, {
+        params,
+        headers: { 'X-Requested-With': 'XMLHttpRequest' },
+      })
+      .then(res => this.updateOutlet(res, params))
+      .catch(error => {
+        console.log(error)
```
teneightfive: `console.error`? And what about if `console` isn't support?

PR #374, opened 2017-07-23
Updated breadcrumb for investment project management editing.

ztolley:
The previous breadcrumb pattern used didn't allow the user to return to the previous screen, level back, and the breadcrumb jumped around and didn't make sense. This pattern is logical and allows proper navigation, and works for other sections too.

Before
before

After
after

### DIFF
--- a/src/apps/investment-projects/controllers/team/edit-project-management.js
+++ b/src/apps/investment-projects/controllers/team/edit-project-management.js
@@ -7,8 +7,8 @@ function getHandler (req, res, next) {
   const briefInvestmentSummary = getDataLabels(res.locals.briefInvestmentSummaryData, briefInvestmentSummaryLabels.view)
 
   res
+    .breadcrumb('Project team', 'team')
     .breadcrumb('Project management')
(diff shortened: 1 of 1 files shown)


feedmypixel:
FYI this will need updating before #368 goes in

ztolley:
@feedmypixel So approve it now and I'll merge it before yours :)

teneightfive:
@ztolley screenshots of before and after might help to see the difference on this change

src/apps/investment-projects/controllers/team/edit-project-management.js
```diff
@@ -7,8 +7,8 @@ function getHandler (req, res, next) {
   const briefInvestmentSummary = getDataLabels(res.locals.briefInvestmentSummaryData, briefInvestmentSummaryLabels.view)
 
   res
+    .breadcrumb('Project team', 'team')
```
teneightfive: This works better but realised the breadcrumb when editing client relationship and referral source feels weird too.

Not sure what we can do there though as its going back to an edit step of a different journey.

PR #375, opened 2017-07-23
Feature/relationship management

ztolley:
Added the client relationship management editing screens.

Made an assumption that editing the account manager for the project updates the investor company account manager.

Refactored the tests a little as some changes to project management code structure didn't get moved in the tests, so had to refactor controller tests into formatter tests.

Formatter is getting big, needs a separate PR now to break ti up.

Will do screen grabs in the morning once some assumptions have been confirmed, plus doesn't work right now due to an error in the updated adviser api calls.

We also need to go back and add tests for the shared middleware as Bens changes don't have test coverage there and I need to add some too.

### DIFF
--- a/src/apps/investment-projects/controllers/team/edit-client-relationship-management.js
+++ b/src/apps/investment-projects/controllers/team/edit-client-relationship-management.js
@@ -1,0 +1,22 @@
+const { isEmpty } = require('lodash')
+
+function getHandler (req, res, next) {
+  res
+    .breadcrumb('Project team', 'team')
+    .breadcrumb('Client relationship management')
+    .render('investment-projects/views/team/edit-client-relationship-management')
+}
+
+function postHandler (req, res, next) {
+  if (isEmpty(res.locals.form.errors)) {
+    req.flash('success', 'Updated investment details')
+    return res.redirect(`/investment-projects/${res.locals.investmentData.id}/team`)
+  }
+
+  return next()
+}
+
+module.exports = {
+  getHandler,
+  postHandler,
+}
--- a/src/apps/investment-projects/middleware/forms/client-relationship-management.js
+++ b/src/apps/investment-projects/middleware/forms/client-relationship-management.js
@@ -1,0 +1,63 @@
+const { get } = require('lodash')
+const { getAdvisers } = require('../../../adviser/repos')
+const { updateCompany } = require('../../../companies/repos')
+const { updateInvestment } = require('../../repos')
+const { clientRelationshipManagementLabels } = require('../../labels')
+
+async function populateForm (req, res, next) {
+  try {
+    const investmentData = res.locals.investmentData
+
+    const advisersResponse = await getAdvisers(req.session.token)
+    const advisers = advisersResponse.results.map((adviser) => {
+      return {
+        value: adviser.id,
+        label: `${adviser.first_name} ${adviser.last_name}`,
+      }
+    })
+
+    res.locals.form = Object.assign({}, res.locals.form, {
+      labels: clientRelationshipManagementLabels.edit,
+      state: {
+        client_relationship_manager: get(investmentData, 'client_relationship_manager.id', null),
+        account_manager: get(investmentData, 'investor_company.account_manager.id', null),
+      },
+      options: {
+        advisers,
+      },
+      hiddenFields: {
+        investor_company: get(investmentData, 'investor_company.id'),
+      },
+      buttonText: 'Save',
+      returnLink: `/investment-projects/${investmentData.id}/team`,
+    })
+
+    next()
+  } catch (error) {
+    next(error)
+  }
+}
+
+async function handleFormPost (req, res, next) {
+  try {
+    res.locals.projectId = req.params.id
+    await updateCompany(req.session.token, req.body.investor_company, { account_manager: req.body.account_manager })
+    await updateInvestment(req.session.token, res.locals.projectId, { client_relationship_manager: req.body.client_relationship_manager })
+    next()
+  } catch (err) {
+    if (err.statusCode === 400) {
+      res.locals.form = Object.assign({}, res.locals.form, {
+        errors: err.error,
+        state: req.body,
+      })
+      next()
+    } else {
+      next(err)
+    }
+  }
+}
+
+module.exports = {
+  populateForm,
+  handleFormPost,
+}
--- a/src/apps/investment-projects/router.js
+++ b/src/apps/investment-projects/router.js
@@ -21,6 +21,7 @@ const valueFormMiddleware = require('./middleware/forms/value')
 const requirementsFormMiddleware = require('./middleware/forms/requirements')
 const interactionsFormMiddleware = require('./middleware/forms/interactions')
 const projectManagementFormMiddleware = require('./middleware/forms/project-management')
+const clientRelationshipManagementFormMiddleware = require('./middleware/forms/client-relationship-management')
 const { renderInvestmentList } = require('./controllers/list')
 const { setDefaults, getInvestmentProjectsCollection } = require('./middleware/collection')
 
@@ -84,6 +85,19 @@ router
     projectManagementFormMiddleware.handleFormPost,
     team.editProjectManagement.postHandler,
     team.editProjectManagement.getHandler
+  )
+
+router
+  .route('/:id/edit-client-relationship-management')
+  .get(
+    clientRelationshipManagementFormMiddleware.populateForm,
+    team.editClientRelationshipManagement.getHandler
+  )
+  .post(
+    clientRelationshipManagementFormMiddleware.populateForm,
+    clientRelationshipManagementFormMiddleware.handleFormPost,
+    team.editClientRelationshipManagement.postHandler,
+    team.editClientRelationshipManagement.getHandler
   )
 
 router.get('/:id/interactions', interactions.list.indexGetHandler)
--- a/src/apps/investment-projects/services/formatting.js
+++ b/src/apps/investment-projects/services/formatting.js
@@ -274,6 +274,27 @@ function transformProjectManagementForView (investmentData) {
   return null
 }
 
+function transformClientRelationshipManagementForView (investmentData) {
+  const crm = investmentData.client_relationship_manager
+
+  const result = [{
+    role: 'Client relationship manager',
+    adviser: crm.first_name + ' ' + crm.last_name,
+    team: crm.dit_team.name,
+  }]
+
+  const accountManager = get(investmentData, 'investor_company.accountManager.id', null)
+  if (accountManager) {
+    result.push({
+      advisor: get(investmentData, 'investor_company.accountManager.name', null),
+      role: 'Account manager',
+      team: get(investmentData, 'investor_company.accountManager.dit_team.name', null),
+    })
+  }
+
+  return result
+}
+
 module.exports = {
   transformInvestmentDataForView,
   transformInvestmentValueForView,
@@ -284,4 +305,5 @@ module.exports = {
   transformFromApi,
   transformBriefInvestmentSummary,
   transformProjectManagementForView,
+  transformClientRelationshipManagementForView,
 }
--- a/src/apps/investment-projects/services/formatting.js
+++ b/src/apps/investment-projects/services/formatting.js
@@ -261,7 +261,7 @@ function getAdviserName (investmentData, key) {
 function transformProjectManagementForView (investmentData) {
   if (investmentData.project_manager || investmentData.project_assurance_adviser) {
     return [{
-      role: 'Project assuranace adviser',
+      role: 'Project assurance adviser',
       adviser: getAdviserName(investmentData, 'project_assurance_adviser'),
       team: get(investmentData, 'project_assurance_team.name', null),
     }, {
@@ -275,20 +275,18 @@ function transformProjectManagementForView (investmentData) {
 }
 
 function transformClientRelationshipManagementForView (investmentData) {
-  const crm = investmentData.client_relationship_manager
-
   const result = [{
     role: 'Client relationship manager',
-    adviser: crm.first_name + ' ' + crm.last_name,
-    team: crm.dit_team.name,
+    adviser: getAdviserName(investmentData, 'client_relationship_manager'),
+    team: get(investmentData, 'client_relationship_manager.dit_team.name', null),
   }]
 
-  const accountManager = get(investmentData, 'investor_company.accountManager.id', null)
+  const accountManager = get(investmentData, 'investor_company.account_manager.id', null)
   if (accountManager) {
     result.push({
-      advisor: get(investmentData, 'investor_company.accountManager.name', null),
+      adviser: get(investmentData, 'investor_company.account_manager.name', null),
       role: 'Account manager',
-      team: get(investmentData, 'investor_company.accountManager.dit_team.name', null),
+      team: get(investmentData, 'investor_company.account_manager.dit_team.name', null),
     })
   }
 
--- a/src/apps/investment-projects/views/_details-form.njk
+++ b/src/apps/investment-projects/views/_details-form.njk
@@ -8,27 +8,29 @@
     options=form.options.contacts)
   }}
 
-  {% component 'form-multiple-choice', {
-    variant: 'radio',
-    label: 'Are you the client relationship manager for this project?',
-    name: 'is-relationship-manager',
-    inline: true,
-    value: form.state['is-relationship-manager'],
-    children: [{
-      label: 'Yes',
-      value: user.id
-    }, {
-      label: 'No'
-    }]
-  } %}
+  {% if not investmentData.id %}
+    {% component 'form-multiple-choice', {
+      variant: 'radio',
+      label: 'Are you the client relationship manager for this project?',
+      name: 'is-relationship-manager',
+      inline: true,
+      value: form.state['is-relationship-manager'],
+      children: [{
+        label: 'Yes',
+        value: user.id
+      }, {
+        label: 'No'
+      }]
+    } %}
 
-  <div class="panel panel-border-narrow js-ConditionalSubfield" data-controlled-by="is-relationship-manager" data-control-value="No">
-    {{ autocomplete('client_relationship_manager',
-      label="Client relationship manager",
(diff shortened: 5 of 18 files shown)


src/apps/investment-projects/middleware/forms/client-relationship-management.js
```diff
@@ -0,0 +1,63 @@
...
+    const investmentData = res.locals.investmentData
+
+    const advisersResponse = await getAdvisers(req.session.token)
+    const advisers = advisersResponse.results.map((adviser) => {
+      return {
+        value: adviser.id,
+        label: `${adviser.first_name} ${adviser.last_name}`,
+      }
+    })
+    .filter(adviser => adviser.label.trim().length > 0)
```
teneightfive: Should this not be being handled at API level? Don't think it shouldn't be returning empty results.

src/apps/investment-projects/middleware/forms/client-relationship-management.js
```diff
@@ -0,0 +1,63 @@
...
+
+async function handleFormPost (req, res, next) {
+  try {
+    res.locals.projectId = req.params.id
+    await updateCompany(req.session.token, req.body.investor_company, { account_manager: req.body.account_manager })
+    await updateInvestment(req.session.token, res.locals.projectId, { client_relationship_manager: req.body.client_relationship_manager })
+    next()
+  } catch (err) {
+    if (err.statusCode === 400) {
+      res.locals.form = get(res, 'locals.form', {})
```
teneightfive: Could be:

```js
  errors: err.error,
  state: req.body,
})
```

We don't need to use `get()` on `res.locals` as it always exists on the response object.

src/apps/investment-projects/router.js
```diff
@@ -73,6 +74,18 @@ router
     team.editProjectManagement.getHandler
   )
 
+router
+  .route('/:id/edit-client-relationship-management')
+  .get(
+    clientRelationshipManagementform.populateForm,
+    team.editClientRelationshipManagement.getHandler)
```
teneightfive: Drop `)` on to new line to match pattern below and elsewhere

src/apps/investment-projects/services/formatting.js
```diff
@@ -261,7 +261,7 @@ function getAdviserName (investmentData, key) {
 function transformProjectManagementForView (investmentData) {
   if (investmentData.project_manager || investmentData.project_assurance_adviser) {
     return [{
-      role: 'Project assuranace adviser',
+      role: 'Project assurance adviser',
```
teneightfive: Good spot 👍

src/apps/investment-projects/services/formatting.js
```diff
@@ -274,6 +274,25 @@ function transformProjectManagementForView (investmentData) {
   return null
 }
 
+function transformClientRelationshipManagementForView (investmentData) {
+  const result = [{
+    role: 'Client relationship manager',
+    adviser: getAdviserName(investmentData, 'client_relationship_manager'),
+    team: get(investmentData, 'client_relationship_manager.dit_team.name', null),
```
teneightfive: Do we need to return `null` as the default value? It should return undefined by default. Is that enough to check against?

src/apps/investment-projects/views/team/edit-client-relationship-management.njk
```diff
@@ -0,0 +1,30 @@
+{% extends "../_layout.njk" %}
+
+{% set phase = investmentData.stage.name %}
+{% set crm = investmentData.client_relationship_manager %}
+{% set rsa = investmentData.referral_source_adviser %}
+
+{% block main_grid_right_column %}
+  <h2 class="heading-medium">Edit client relationship management</h2>
```
teneightfive: Might want to look at the [new format](https://github.com/uktrade/data-hub-frontend/pull/345) for headings that @tyom added.

src/apps/investment-projects/views/team/edit-client-relationship-management.njk
```diff
@@ -0,0 +1,30 @@
...
+
+  {% call Form(form) %}
+    {{ MultipleChoiceField({
+      name: 'client_relationship_manager',
+      label: form.labels.client_relationship_manager,
+      error: form.errors.client_relationship_manager,
+      options: form.options.advisers,
+      value: form.state.client_relationship_manager,
+      initialOption: 'Select an adviser'
```
teneightfive: We don't really have a set pattern for this I suppose but elsewhere we've been prefixing and suffixing the initial value with `--` to distinguish it from other results. 

So this would become `-- Select an adviser --`.

@tyom / @feedmypixel what are your thoughts on this? Happy either way but we should try and be consistent? Maybe if that's the pattern we want we could add the `--` in the macro rather than in each call?

test/unit/apps/investment-projects/controllers/team/edit-client-relationship-management.test.js
```diff
@@ -0,0 +1,126 @@
...
+          token: 'abcd',
+        },
+      }, {
+        locals: {
+          investmentData,
+        },
+        breadcrumb: this.breadcrumbStub,
+        render: (template) => {
+          try {
+            expect(this.breadcrumbs[0]).to.equal('Project team')
```
teneightfive: Not sure we need to test for the value of breadcrumb items. Feels like if we change copy we don't want tests to break. 

The breadcrumb middleware tests should be enough to know that it will set the correct value.

test/unit/apps/investment-projects/controllers/team/edit-client-relationship-management.test.js
```diff
@@ -0,0 +1,126 @@
...
+    this.nextStub = this.sandbox.stub()
+    this.flashStub = this.sandbox.stub()
+    this.getDataLabelsStub = this.sandbox.stub()
+
+    // The breadcrumb needs to have 'this' within it's own context so it can be chained
+    // But also needs to link to the 'this' in the tests to record the breadcrumb set
+    // Just using a regular stub wont work.
+    this.breadcrumbs = []
+    const that = this
+    this.breadcrumbStub = function (title) {
```
teneightfive: If we remove the tests for the value this can also become a simpler stub:

```js
this.breadcrumbStub = function () { return this }
```

test/unit/apps/investment-projects/middleware/forms/client-relationship-management.test.js
```diff
@@ -0,0 +1,275 @@
...
+            updateCompany: this.updateCompanyStub,
+          },
+        })
+      })
+
+      afterEach(() => {
+        this.sandbox.restore()
+      })
+
+      it('sets form error data for the following controllers if form error', (done) => {
```
teneightfive: I find it easier to read when using `describe()` blocks to set the scene. 

So this instance could use a describe block to set the scene:

```js
describe('when a form error occurs', () => {})
```

and then an it block inside that to say the expected outcome:

```js
it('sets form error data', () => {})
```

which would produce this result:

```js
describe('when a form error exists', () => {
  it('should set form error data', () => {
    ...
  })
})
```

It also allows you to set things up for similar tests within the describe blocks.

test/unit/apps/investment-projects/controllers/team/edit-client-relationship-management.test.js
ztolley: Don't you want to test the breadcrumb is correctly set though?

test/unit/apps/investment-projects/controllers/team/edit-client-relationship-management.test.js
ztolley: @tyom @feedmypixel What do you think. My thoughts are if the controller is setting the breadcrumb to display, it should be tested.

teneightfive: The purpose and function of the breadcrumb being set to the correct value if one is given to it is already tested within the breadcrumb middleware so this is just duplicating that test and feels unnecessary.

I'd also try and avoid checking the value of text set in the controller within tests as it means unnecessary updates if we simply want to change copy.

src/apps/investment-projects/middleware/forms/client-relationship-management.js
ztolley: We had an issue with it returning advisers wtih no name so I put this in. removingit

src/apps/investment-projects/services/formatting.js
ztolley: I had unpredictable behaviour with undefined, in some instances it shows undefined in the screen, so I put in null to keep it predictable.

src/apps/investment-projects/middleware/forms/client-relationship-management.js
teneightfive: Cool, ideally we should be confident in the API handling cleaning the data it returns.

src/apps/investment-projects/views/team/edit-client-relationship-management.njk
ztolley: This is the same style currently in the team edit page in develop. So unless that was missed by @tyom I'll leave it as is

test/unit/apps/investment-projects/controllers/team/edit-client-relationship-management.test.js
ztolley: the breadcrumb middleware tests it can set a breadcrumb, this tests the correct breadcrumb is displayed.

If we want to say this is outside of scope of these tests then thats ok as long as user acceptance tests cover it.

test/unit/apps/investment-projects/middleware/forms/client-relationship-management.test.js
teneightfive: Another option is also [context](https://mochajs.org/#bdd) which is an alias for `describe` but helps with readability.

test/unit/apps/investment-projects/middleware/forms/client-relationship-management.test.js
```diff
@@ -196,7 +196,7 @@ describe('Investment form middleware - client relationship management', () => {
       })
     })
 
-    describe('post with form errors', () => {
+    describe('When a form is posted with errors', () => {
```
teneightfive: Could stick will all lowercase like other tests for now.

test/unit/apps/investment-projects/middleware/forms/client-relationship-management.test.js
```diff
@@ -224,7 +224,7 @@ describe('Investment form middleware - client relationship management', () => {
         this.sandbox.restore()
       })
 
-      it('sets form error data for the following controllers if form error', (done) => {
+      it('should set form error data for the following controllers if form error', (done) => {
```
teneightfive: Can probably drop the `if from error` from the end now.

I posted in a different thread but [context](https://mochajs.org/#bdd) is another option to use to help with readability.

src/apps/investment-projects/views/team/edit-client-relationship-management.njk
teneightfive: Ah hold on. I misread the block. Think this is correct as its the heading for the panel, right?

test/unit/apps/investment-projects/controllers/team/edit-client-relationship-management.test.js
teneightfive: Yeah, testing the correct values for things like headings/breadcrumbs I think are better handled in acceptance tests that check the contents of the page.

Unit tests could check a method is called and that will probably be enough for this controller test.

src/apps/investment-projects/views/team/edit-client-relationship-management.njk
ztolley: Right

src/apps/investment-projects/controllers/team/edit-client-relationship-management.js
```diff
@@ -0,0 +1,24 @@
+const { isEmpty } = require('lodash')
+
+function getHandler (req, res, next) {
+  res
+    .breadcrumb('Project team', 'team')
+    .breadcrumb('Client relationship management')
+    .render('investment-projects/views/team/edit-client-relationship-management', {
+      currentNavItem: 'team',
```
tyom: You shouldn't need this any more.

src/apps/investment-projects/controllers/team/edit-client-relationship-management.js
```diff
@@ -0,0 +1,24 @@
...
+      variant: 'edit',
+    })
+}
+
+function postHandler (req, res, next) {
+  if (!isEmpty(res.locals.form.errors)) {
```
tyom: Perhaps swap returns to avoid negating. If errors empty, redirect, else call next.

src/apps/investment-projects/middleware/forms/client-relationship-management.js
```diff
@@ -0,0 +1,63 @@
...
+    const advisersResponse = await getAdvisers(req.session.token)
+    const advisers = advisersResponse.results.map((adviser) => {
+      return {
+        value: adviser.id,
+        label: `${adviser.first_name} ${adviser.last_name}`,
+      }
+    })
+
+    res.locals.form = Object.assign({}, res.locals.form, {
+      labels: clientRelationshipManagementLabels.edit,
```
tyom: I'd suggest to wrap just `await` bits in try/catch to keep the block small.

src/apps/investment-projects/middleware/forms/client-relationship-management.js
```diff
@@ -0,0 +1,63 @@
...
+
+async function handleFormPost (req, res, next) {
+  try {
+    res.locals.projectId = req.params.id
+    await updateCompany(req.session.token, req.body.investor_company, { account_manager: req.body.account_manager })
+    await updateInvestment(req.session.token, res.locals.projectId, { client_relationship_manager: req.body.client_relationship_manager })
+    next()
+  } catch (err) {
+    if (err.statusCode === 400) {
+      res.locals.form = Object.assign({}, res.locals.form, {
```
tyom: This seems to be a common pattern. Should we extract it into something reusable across different sub-apps?

src/apps/investment-projects/middleware/forms/client-relationship-management.js
```diff
@@ -0,0 +1,63 @@
...
+        errors: err.error,
+        state: req.body,
```
tyom: We need to standardise the names for form properties. In some places we use `data` in other `state`.

src/apps/investment-projects/views/_details-form.njk
```diff
@@ -8,27 +8,29 @@
...
-  } %}
-
-  <div class="panel panel-border-narrow js-ConditionalSubfield" data-controlled-by="is-relationship-manager" data-control-value="No">
-    {{ autocomplete('client_relationship_manager',
-      label="Client relationship manager",
-      value=form.state['client_relationship_manager'],
-      options=form.options.advisers)
-    }}
-  </div>
+  {% if not investmentData.id %}
```
tyom: I would suggest using new `MultipleChoiceField` macro instead of this one. This should be deprecated.

tyom: Although I realise you are just re-indenting it so perhaps a separate PR for form refactor. I'm doing them one at a time when I have time.

src/apps/investment-projects/controllers/team/edit-client-relationship-management.js
ztolley: Shouldn't need what exactly? and has this been updated in other screens for me to copy?

src/apps/investment-projects/controllers/team/edit-client-relationship-management.js
ztolley: copied the same approach other controllers took to be consitant

src/apps/investment-projects/middleware/forms/client-relationship-management.js
ztolley: Good idea, but maybe in a different ticket to keep this one focussed

src/apps/investment-projects/middleware/forms/client-relationship-management.js
ztolley: I know. I copied what did did in other forms in investment. Again, maybe one for a separate pr

src/apps/investment-projects/views/_details-form.njk
ztolley: I was just adding the conditional so out of scope for this ticket but yeah. And I just got told to switch to this one in the previous ticket. The form controls change... often (change is good)

src/apps/investment-projects/controllers/team/edit-client-relationship-management.js
tyom: That line. Local nav has a different mechanism finding out if the page is current. The change was in one of my recent PRs.

ztolley: Point me at some code

teneightfive: Think @tyom is referring to this: https://github.com/uktrade/data-hub-frontend/pull/372

src/apps/investment-projects/middleware/forms/client-relationship-management.js
ztolley: Something could go wrong somewhere other than just the await call so it is simpler to handle any errors this way.

src/apps/investment-projects/controllers/team/edit-client-relationship-management.js
ztolley: ah that's a problem then. 

the url for editing project management is /:id/edit-project-management, and for customer relationship management it is /:id/edit-client-relationship-management

This mean that the Team tab now doesn't highlight when editing these.

src/apps/investment-projects/middleware/forms/client-relationship-management.js
tyom: If something goes wrong somewhere should the whole thing fail? Besides most other bits just set properties from existing data.

ztolley: I would rather be defensive in case there is a bad property plus this pattern has been used alot in the app.

PR #377, opened 2017-07-24
fixed reference error in investment interactions

jim68000:
If an interaction is created on a contact as part of an investment project, then the company.id is not at the top level of the interaction object - this was causing interaction views to throw a 500 when you attempted to view the interaction as part of the contact's investment history.

We should probably have a sensible link to the investment project as part of the interaction display, but this commit is just to cure the 500

### DIFF
--- a/src/apps/interactions/services/data.js
+++ b/src/apps/interactions/services/data.js
@@ -17,7 +17,8 @@ function getHydratedInteraction (token, interactionId) {
   return new Promise(async (resolve, reject) => {
     try {
       const interaction = await interactionRepository.getInteraction(token, interactionId)
-      interaction.company = await companyRepository.getDitCompany(token, interaction.company.id)
+      const companyId = interaction.company ? interaction.company.id : interaction.contact.company.id
+      interaction.company = await companyRepository.getDitCompany(token, companyId)
       resolve(interaction)
     } catch (error) {
       reject(error)
--- a/test/unit/apps/interactions/services/data.service.test.js
+++ b/test/unit/apps/interactions/services/data.service.test.js
@@ -1,14 +1,15 @@
 /* eslint handle-callback-err: 0, camelcase: 0 */
 
+let interactionDataService
+let getContactStub
+let getDitCompanyStub
+let getInteractionStub
+let company
+let contact
+let dit_adviser
+let interaction_type
+
 describe('interaction data service', function () {
-  let interactionDataService
-  let getContactStub
-  let getDitCompanyStub
-  let getInteractionStub
-  let company
-  let contact
-  let dit_adviser
-  let interaction_type
   let interaction
   const token = '9876'
 
@@ -18,7 +19,6 @@ describe('interaction data service', function () {
     dit_adviser = { id: '4455', name: 'Fred Jones', first_name: 'Fred', last_name: 'Jones' }
     interaction_type = { id: '1234', name: 'Email' }
     interaction = { id: '999', company: { id: company.id } }
-
     getContactStub = sinon.stub().resolves(contact)
     getDitCompanyStub = sinon.stub().resolves(company)
     getInteractionStub = sinon.stub().resolves(interaction)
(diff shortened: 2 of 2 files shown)


src/apps/interactions/services/data.js
```diff
@@ -17,7 +17,8 @@ function getHydratedInteraction (token, interactionId) {
   return new Promise(async (resolve, reject) => {
     try {
       const interaction = await interactionRepository.getInteraction(token, interactionId)
-      interaction.company = await companyRepository.getDitCompany(token, interaction.company.id)
+      const company = interaction.company ? interaction.company.id : interaction.contact.company.id
```
teneightfive: Maybe `const companyId = ...` as its only setting the ID not a company object?

jim68000: Done

jim68000:
@ztolley The tests need a bit of refactoring to make them understand the different interactions I'm passing in. On it.

src/apps/interactions/services/data.js
teneightfive: Have you pushed the change? Can't see it here

jim68000:
Is there a brighter way of handling these tests? Basically I want to run all the tests once with a normal interaction, and once with an investment interaction. Can I abstract out the tests somehow? Just C&P feels laborious

test/unit/apps/interactions/services/data.service.test.js
```diff
@@ -197,3 +197,201 @@ describe('interaction data service', function () {
...
+        .catch((error) => {
+          done()
+        })
+    })
+  })
+  describe('create blank interaction for company', function () {
+    it('should return a valid blank interaction with company populated', function () {
+      return interactionDataService.createBlankInteractionForCompany(token, dit_adviser, interaction_type.id, company.id)
+        .then((interaction) => {
+          expect(interaction).to.not.be.null
```
tyom: Could probably use `to.eventually.*`

test/unit/apps/interactions/services/data.service.test.js
```diff
@@ -197,3 +197,201 @@ describe('interaction data service', function () {
...
+          expect(interaction).to.not.have.property('id')
+          expect(interaction.contact).to.be.null
+          expect(interaction.company).to.deep.equal(company)
+          expect(interaction.interaction_type).to.deep.equal(interaction_type)
+          expect(interaction.dit_adviser).to.deep.equal(dit_adviser)
+          expect(interaction).to.have.property('date')
+          expect(interaction.service).to.deep.equal({ id: null, name: null })
+          expect(interaction.dit_team).to.deep.equal({ id: null, name: null })
+          expect
```
tyom: Leftovers?

jim68000: Weird, that was there before I made any changes - it's just obvious now because of the copy and paste. nvm - fixed them both

test/unit/apps/interactions/services/data.service.test.js
jim68000: Tried that, but Circle didn't like it for some reason. I may have just got it wrong, but this is the exact same code as the non-interaction tests up there ⬆️ so I'm just going to keep it as is.

test/unit/apps/interactions/services/data.service.test.js
```diff
@@ -197,3 +195,201 @@ describe('interaction data service', function () {
     })
   })
 })
+
+describe('interaction data service with investment interaction', function () {
+  let interactionDataService
```
teneightfive: You could move these definitions outside of both describe blocks and place them at the top of the file to avoid having to duplicate them.

jim68000: Fair point. Still wish there was a neater abstraction here, but that will remove some duplkication

jim68000: Fair point. Still wish there was a neater abstraction here, but that will remove some duplication

jim68000: Done

test/unit/apps/interactions/services/data.service.test.js
teneightfive: Do we want to drop this empty `expect`?

jim68000: Hm. Should have gone. Try again.

PR #382, opened 2017-07-26
Entity component

tyom:
Adds `Entity` and `MetaItem` components. Use these components to render investment projects list.

### 💥 [Demo](https://datahub-beta2-pr-382.herokuapp.com/investment-projects) 💥

localhost-3000-investment-projects


assets/stylesheets/components/_entity.scss
```diff
@@ -0,0 +1,66 @@
+@import "../tools";
+
+.c-entity__title {
+  @include bold-font(20);
+
```
teneightfive: Unnecessary new line

assets/stylesheets/components/_entity.scss
```diff
@@ -0,0 +1,66 @@
...
+.c-entity__tags {
+  @include core-font(14);
+}
+
+.c-entity__content .c-meta-item {
+  display: block;
+}
+
+.c-entity__tags {
+  margin: 20 - 14px 0;
```
teneightfive: What's `14px` relate to?

assets/stylesheets/components/_entity.scss
```diff
@@ -0,0 +1,66 @@
...
+}
+
+.c-entity__code {
+  color: $grey-1;
+}
+
+
+// Assemblies
```
teneightfive: Did we not want to keep these in a consistent place?

assets/stylesheets/components/_tag.scss
```diff
@@ -0,0 +1,48 @@
...
+  text-decoration: none;
+  font-weight: 600;
+
+  &:link,
+  &:visited {
+    background-color: $white;
+  }
+}
+
+.c-tag--alt {
```
teneightfive: What if we have a third style? Might `.c-tag--secondary` then we can use `.c-tag--tertiary` if add a third.

assets/stylesheets/components/_entity.scss
tyom: It depends on the agreed style :) I tend to separate `include` from other lines, but happy to change we collectively agree.

assets/stylesheets/components/_entity.scss
tyom: Font sizes: 20 for title and 14 for tags. The difference is offset.

src/apps/investment-projects/views/list.njk
```diff
@@ -1,5 +1,11 @@
 {% extends "_layouts/two-column.njk" %}
 
+{% from "_macros/entities.njk" import Entity with context %}
+
+{% block body_main_header_content %}
+  <h1 class="page-heading">Investment projects</h1>
```
teneightfive: Should this not be using the new local header method?

assets/stylesheets/components/_entity.scss
tyom: These in-component assemblies. Different to global assemblies which should only contain component roots.

assets/stylesheets/components/_entity.scss
teneightfive: We haven't used a line anywhere else so I'd say drop it.

assets/stylesheets/_components.scss
```diff
@@ -3,19 +3,22 @@
...
 @import "components/data-table";
 @import "components/details-list";
+@import "components/entity";
+@import "components/entity-list";
 @import "components/filters";
 @import "components/form";
 @import "components/local-header";
 @import "components/local-nav";
 @import "components/messages";
+@import "components/meta-item";
```
teneightfive: Do we want to revisit `metadata-list` with this new component?

assets/stylesheets/_components.scss
```diff
@@ -3,19 +3,22 @@
...
 @import "components/metadata-list";
 @import "components/pagination";
 @import "components/phase-badge";
 @import "components/proposition-menu";
 @import "components/results";
 @import "components/results-summary";
-@import "components/entity-list";
 @import "components/status-bar";
+@import "components/tag";
```
teneightfive: Would `label` or `badge` be better?

assets/stylesheets/_components.scss
tyom: Definitely. But later as we convert more lists in different entity types.

assets/stylesheets/_components.scss
tyom: Maybe `badge` could work.

src/apps/investment-projects/middleware/collection.js
```diff
@@ -1,19 +1,15 @@
-const { get, pick, pickBy, omit } = require('lodash')
+const { get, pick, pickBy } = require('lodash')
 
 const { buildPagination } = require('../../../lib/pagination')
-const { buildQueryString } = require('../../../lib/url-helpers')
 const metadataRepo = require('../../../lib/metadata')
 const { collectionFilterLabels } = require('../labels')
+const {
+  transformMetadataIntoOption,
```
teneightfive: Might want to look at the app level `transformers` to see if that can be used now.

src/templates/_macros/entities.njk
```diff
@@ -0,0 +1,89 @@
...
+ # @param {string} props.type - entity type (e.g. 'date' to format dates)
+ # @param {array} [props.metaTags{}] - an array of metadata item objects
+ # @param {array} [props.metaItems{}] - an array of metadata item objects
+ #}
+{% macro Entity (props) %}
+  {% if props.name and props.id and props.type %}
+    {% set metaTags = props.meta | filter(['isTag', true]) %}
+    {% set metaItems = props.meta | filter(['isTag', false]) %}
+
+    <div class="c-entity c-entity--{{ props.type }}">
```
teneightfive: Could these be and `<article>`?

tyom: I'm not sure `<article>` as list item fits the purpose of the tag. I wonder if when taken out of context it still makes sense as a self-contained composition. And having 10 or 20 of them on the page might confuse users of screen readers.

assets/stylesheets/components/_badge.scss
```diff
@@ -0,0 +1,48 @@
...
+  color: inherit;
+  transition: 60ms;
+  cursor: default;
+
+  &:link,
+  &:visited {
+    color: $colour;
+    cursor: pointer;
+
+    &:hover {
```
teneightfive: Do these need to be nested within `&:link, &:visited`?

assets/stylesheets/components/_badge.scss
```diff
@@ -0,0 +1,48 @@
...
+      background-color: $hover;
+      border-color: $hover;
+      color: $white;
+    }
+  }
+}
+
+.c-badge {
+  @include _style-badge($green);
+
```
teneightfive: Drop the line

assets/stylesheets/components/_badge.scss
tyom: The do because the badge be any element and we only want to hover on links.

assets/stylesheets/components/_meta-item.scss
```diff
@@ -0,0 +1,32 @@
+@import "../settings";
+
+.c-meta-item {
+  display: inline-block;
+
+  & + &:nth-child(even) {
```
teneightfive: Do you need the first `&`?

src/apps/components/controllers.js
```diff
@@ -66,10 +69,19 @@ function renderPagination (req, res) {
 }
 
 async function renderEntityList (req, res) {
+  const investmentProjects = await authorisedRequest(req.session.token, `${config.apiRoot}/v3/investment?limit=10`)
+    .then(result => {
+      result.page = 1
```
teneightfive: How about:
```js
return Object.assign({}, result, {
  page: 1,
  limit: 10,
  pagination: buildPagination(req, result),
  items: result.results.map(transformInvestmentProjectToListItem),
})
```

tyom: In this case I don't think it matters to much. In fact I think it should go outside to some transformer instead. But it's out of scope for this PR.

src/templates/_macros/entities.njk
```diff
@@ -0,0 +1,89 @@
...
+    {% set metaItems = props.meta | filter(['isBadge', false]) %}
+
+    <div class="c-entity c-entity--{{ props.type }}">
+      {% if props.code %}
+        <div class="c-entity__code">{{ props.code }}</div>
+      {% endif %}
+
+      <div class="c-entity__header">
+        <h3 class="c-entity__title">
+          <a href="/{{ props.type }}s/{{ props.id }}">{{ props.name }}</a>
```
teneightfive: `/{{ props.type }}s/{{ props.id }}` that `s` will break on _companys_. Isn't there also a pluralise filter that would be better?

tyom: True. Updated to use `pluralise` filter but it will still require fixing in future as it doesn't handle companies correctly either.

PR #390, opened 2017-07-28
dh 420: adviser lookup fails in interactions

jim68000:
This recreates the Ajax-driven Adviser lookup used in Interactions. Possible need to change this to the pre-loaded list of advisers? Not sure I like that approach, but this has an obvious non-JS issue.

This is just a bugfix for something that is broken in production: maybe raise a ticket for more extensive changes?

### DIFF
--- a/src/apps/api/controllers.js
+++ b/src/apps/api/controllers.js
@@ -1,11 +1,20 @@
 const { lookupAddress } = require('./services')
 const metadata = require('../../lib/metadata')
+const { adviserSearch } = require('../adviser/repos')
+
+async function adviserSearchHandler (req, res, next) {
+  try {
+    const advisers = await adviserSearch(req.session.token, req.params.term)
+    res.json(advisers)
+  } catch (error) {
+    next(error)
+  }
+}
 
 async function postcodeLookupHandler (req, res) {
   try {
     const postcode = req.params.postcode
     const addresses = await lookupAddress(postcode)
-
     const augmentedAddresses = addresses.map(address => {
       return Object.assign({}, address, {
         country: metadata.countryOptions.find(country => country.name.toLowerCase() === 'united kingdom').id,
@@ -19,5 +28,6 @@ async function postcodeLookupHandler (req, res) {
 }
 
 module.exports = {
+  adviserSearchHandler,
   postcodeLookupHandler,
 }
--- a/src/apps/api/index.js
+++ b/src/apps/api/index.js
@@ -1,7 +1,10 @@
 const router = require('express').Router()
-const { postcodeLookupHandler } = require('./controllers')
+const { adviserSearchHandler, postcodeLookupHandler } = require('./controllers')
+
+router.get('/adviserlookup/:term', adviserSearchHandler)
+router.get('/postcodelookup/:postcode', postcodeLookupHandler)
 
 module.exports = {
   mountpath: '/api',
-  router: router.get('/postcodelookup/:postcode', postcodeLookupHandler),
+  router,
 }
--- a/src/apps/interactions/views/edit.njk
+++ b/src/apps/interactions/views/edit.njk
@@ -50,7 +50,7 @@
       hint="Please start typing to search for an adviser",
       error=errors.dit_adviser,
       autofocus=false,
-      url='/api/accountmanagerlookup?term=')
+      url='/api/adviserlookup/')
     }}
 
     {{ autocomplete("service",


teneightfive:
I would consider dropping the AJAX support for now and use the method we're using elsewhere which is the dropdown of advisers that we will replace with a typeahead before November. Thoughts?

jim68000:
I'd agree. This is a temporary fix while I work on the full adviser list thing - that has an issue with the back end only providing a subset of the full advisers _and_ I need to get BE to bring teams and advisers up to V3 API spec, so it's a bit more involved.

test/unit/config/nunjucks/filters.test.js
```diff
@@ -105,15 +105,9 @@ describe('nunjucks filters', () => {
 
   describe('#formatNumber', () => {
     it('should correctly format number for "en-GB" locale', () => {
-      const formattedNumber = filters.formatNumber(12345678)
+      const formattedNumber = filters.formatNumber(12345678.1)
```
teneightfive: Isn't this change from a different PR?

src/apps/api/controllers.js
```diff
@@ -1,11 +1,41 @@
 const { lookupAddress } = require('./services')
 const metadata = require('../../lib/metadata')
+const config = require('../../../config')
+const authorisedRequest = require('../../lib/authorised-request')
+
+async function adviserSearch (req, res) {
```
teneightfive: Could you not import this function instead of moving it?

test/unit/config/nunjucks/filters.test.js
jim68000: Yeah, didn't spot that. Will fix.

src/apps/api/controllers.js
jim68000: I moved it because it felt more correct to keep things used by Ajax in one place (for future removal), but I can see a world where we might want to reuse, so I'll move it back

src/apps/adviser/repos.js
```diff
@@ -10,47 +9,37 @@ function getAdviser (token, id) {
...
-      if (!term || term.trim().length === 0) {
-        return resolve(null)
-      }
-
-      const parts = term.trim().toLowerCase().split(' ')
-      let url = `${config.apiRoot}/adviser/?first_name__icontains=${parts[0]}`
-      if (parts.length > 1) url += `&last_name__icontains=${parts[1]}`
-
-      const data = await authorisedRequest(token, { url })
+async function adviserSearch (req, res) {
```
teneightfive: Could this not return the promise like previously and the controller use async/await on that function call?

src/apps/api/controllers.js
```diff
@@ -1,11 +1,11 @@
 const { lookupAddress } = require('./services')
 const metadata = require('../../lib/metadata')
+const { adviserSearch } = require('../adviser/repos')
```
teneightfive: So rather than exporting the repo function as a controller method and having that concerned with request and response, create a simple controller method like this:

```js
async function adviserSearchHandler (req, res, next) {
  try {
    const advisers = await adviserSearch(req.session.token, req.params.term)
    res.json(advisers)
  } catch (error) {
    next(error)
  }
}
```

jim68000: Done

src/apps/adviser/repos.js
```diff
@@ -11,46 +11,35 @@ function getAdviser (token, id) {
...
-
-      // API only supports contains, so filter out results that don't start with term
-      // Then reduce the result down to id and name
-      // And finally sort things
-      const filtered = data.results.filter((adviser) => {
-        if (parts.length === 1) {
-          return adviser.first_name.toLowerCase().startsWith(parts[0])
-        }
-        return adviser.first_name.toLowerCase().startsWith(parts[0]) && adviser.last_name.toLowerCase().startsWith(parts[1])
+  return new Promise((resolve) => {
```
teneightfive: Does this function need changing to fix this issue? If not it would be easier to leave this file alone and stick to just adding the controller to handle the route.

src/apps/adviser/repos.js
```diff
@@ -1,6 +1,6 @@
 const config = require('../../../config')
-const logger = require('../../../config/logger')
 const authorisedRequest = require('../../lib/authorised-request')
+const logger = require('../../../config/logger')
```
teneightfive: These changes just seem like noise so may be better just to revert this whole file if there's nothing major that needs changing in the `adviserSearch` function.

jim68000:
I removed the changes to repos.js

src/apps/adviser/repos.js
```diff
@@ -32,14 +32,14 @@ function adviserSearch (token, term) {
...
       })
-      .map((adviser) => {
-        return { id: adviser.id, name: `${adviser.first_name} ${adviser.last_name}` }
-      })
-      .sort((a, b) => {
-        if (a.name.toLowerCase() < b.name.toLowerCase()) return -1
-        if (a.name.toLowerCase() > b.name.toLowerCase()) return 1
-        return 0
-      })
+        .map((adviser) => {
```
teneightfive: Maybe try and kill these whitespaces changes for this fix too. It just adds noise.

PR #392, opened 2017-07-31
Results macro components

tyom:
- Create new Nunjucks macros to componetise results list, summary and filters
- Move macro unit tests into separate directory
- Tweaks and fixes for existing macros/components
- Document component usage on components page

localhost-3001-components-results


assets/stylesheets/components/_results-summary.scss
```diff
@@ -1,5 +1,6 @@
-@import "../settings/layout";
```
teneightfive: Don't we still need to import settings? Looks like we're using variables from `layout` and `colours`.

tyom: `tools` import various settings implicitly. I think that's useful as you basically import single toolbox with all the tools and setting. Although currently it's not obvious because each of the individual tools files (mixins/function) import bits of settings it uses and I think all of them combined might be importing all the settings.

assets/stylesheets/components/_entity-list.scss
```diff
@@ -6,10 +6,6 @@
   border: solid $grey-2;
   border-width: 0 0 1px;
   padding: $default-spacing-unit 0;
-
-  &:first-child {
```
teneightfive: Is this intentional? After this change the search results no longer have the top border? Is this because they're using the old results components?

assets/stylesheets/components/_results-summary.scss
teneightfive: I still think it would be clearer to explicitly define the dependencies need for this module rather than assuming it has everything it needs loaded at an earlier stage.

Otherwise we may as well strip these all together and use the fact that anything loaded earlier in the chain in sass is available.

assets/stylesheets/components/_results-summary.scss
```diff
@@ -13,3 +14,72 @@
...
+  }
+
+  &:hover {
+    color: $red;
+  }
+}
+
+.c-results-summary__filter-remove-all {
+  @include core-font(16);
+
```
teneightfive: Drop new line. We haven't been using this pattern elsewhere.

src/apps/investment-projects/middleware/index.js
```diff
@@ -1,7 +1,9 @@
 const forms = require('./forms')
 const shared = require('./shared')
+const collection = require('./collection')
```
teneightfive: Plural?

assets/stylesheets/components/_results-summary.scss
tyom: OK but I think just explicitly setting `settings` loader is enough. No need to pick individual settings files I don't think.

assets/stylesheets/components/_entity-list.scss
tyom: Yes, this is intentional. I think the top border on search results is not needed as they are meant to be separators between rows. The sort dropdown and pagination summary does introduce the border though but for results on their own I don't believe the top border is necessary.

src/apps/investment-projects/middleware/index.js
tyom: Not sure. I think the top level files such as `middleware`, `controllers`, `services` should be plural as they split out into multiple sub files. But here this middleware ware contains logic of how to deal with this particular collection.

assets/stylesheets/components/_results-summary.scss
teneightfive: Yep, that's fine.

src/apps/investment-projects/middleware/index.js
teneightfive: Only reason I mentioned is because 2 lines above we have `const forms = ...` why not `form`? We should try and be consistent.

tyom: Hmm.. Maybe I should rename `forms` to `form`.

tyom: Although one could argue that there are multiple forms in investment projects and only one type of collection.

assets/stylesheets/components/_entity-list.scss
```diff
@@ -3,9 +3,11 @@
 @import "../tools/functions/typography";
 
 .c-entity-list__item {
-  border: solid $grey-2;
-  border-width: 0 0 1px;
   padding: $default-spacing-unit 0;
+
+  & + & {
```
teneightfive: Can this not just be `+ & { ... }`?

PR #395, opened 2017-07-31
Investment project stages

feedmypixel:
**NOTE - the Investment Project flow itself is in discussion at present as we need some discovery to confirm we are meeting user needs. However the work here can be easily repurposed and is still valid**

This work involves the first part of the investment creation project flow. 

### Of Note (see commits for more detailed information)
- Rework investment project flow to work with CTI, FDI and Non-FDI types (see flow diagram below)
- Improve url structure for the create flow
- Add new form macros to Investment project creation flow
- Add project stage progress component to show current position in the flow, #423 
- Add move project stage component to move between stages, #422 
- Add middleware for storing information under a namespace in the session
- Add date component pattern #425 

### TODO
- There are a number of UX issues to address, this will be addressed in the next piece of work.
   - what you need to do to proceed to the next stage
   - position of stage info/message box
   - Requirements and Value forms asking to much
   - forms too big
- Please note stronger test coverage will be coming in a future PR.

### Create Investment project journey flow diagram
create-investment-porject-journey-flow

### Investment project journey (uk and foreign)
investment-project


assets/stylesheets/components/_progress.scss
```diff
@@ -0,0 +1,152 @@
...
+    background-color: $grey-3;
+    z-index: 10;
+  }
+
+  &::after {
+    background-color: $govuk-blue;
+    z-index: 20;
+  }
+
+  &.c-progress__bar--prospect-stage {
```
teneightfive: I'd be tempted not to use class names that relate to investment progress only. Maybe modifiers that relate to the percentage complete?

assets/stylesheets/components/_progress.scss
```diff
@@ -0,0 +1,152 @@
...
+}
+
+.c-progress__stage--reset {
+  &::before {
+    background-color: $white;
+    border: 2px solid $grey-3;
+  }
+}
+
+.c-progress__stage--active {
```
teneightfive: This feels more like a state class than a modifier. Maybe just `is-active`.

assets/stylesheets/components/_progress.scss
```diff
@@ -0,0 +1,152 @@
...
+
+.c-progress__stage--part-4 {
+  left: $section-width * 3;
+}
+
+.c-progress__stage--part-5 {
+  left: $section-width * 4;
+}
+
+@include media($min-width: 1060px) {
```
teneightfive: What does `1060px` relate to? Don't think we've used this breakpoint before.

teneightfive: Might also be worth looking at on smaller screens. Feels like the center aligned text for first and last items could get cut off at certain resolutions too.

assets/stylesheets/components/_progress.scss
```diff
@@ -0,0 +1,152 @@
...
+
+.c-progress__stage--reset {
+  &::before {
+    background-color: $white;
+    border: 2px solid $grey-3;
+  }
+}
+
+.c-progress__stage--active {
+  @include bold-font(16);
```
teneightfive: Not sure about the font size jump for current item. Think bold font weight might be enough.

tyom:
Wonder if the investor company and code should be flipped to be consistent with investment projects pipeline layout.

tyom:
Initial sorting selection seems to have gone
image

assets/stylesheets/components/form/_form-control.scss
```diff
@@ -66,3 +66,11 @@ select.c-form-control {
 .c-form-control--short {
   max-width: 10em;
 }
+
+.c-form-control--medium {
+  max-width: 4em;
+}
+
+.c-form-control--small {
```
teneightfive: Maybe we should tweak the current `.c-form-control--short` to be `.c-form-control--long`, add `.c-form-control--medium` and then set `.c-form-control--short` to be `2.6em`?

Avoids us mixing `short` and `small` then.

assets/stylesheets/components/form/_form-fieldset.scss
```diff
@@ -16,6 +17,12 @@
   margin: 0 $default-spacing-unit / 2;
 }
 
+.c-form-fieldset__hint {
```
teneightfive: What's this being used for? Couldn't find it in the views.

assets/stylesheets/components/form/_form-control.scss
tyom: We use `small` and `smaller` to adjust input size.

image

You should pick something different. We already have `short`. Maybe `shorter` and `shortest`?
See above for existing rule `.c-form-control--short`.

assets/stylesheets/components/_status-bar.scss
```diff
@@ -2,39 +2,28 @@
...
+    &:last-child {
+      border-right: 0;
+    }
   }
 }
 
 .status-bar__section-title {
-  font-weight: 700;
+  color: $grey-1;
+  font-size: 16px;
```
tyom: Apparently we're using `core-font` mixin until we do a review of our typography.

src/templates/_components/investment-status.njk
```diff
@@ -14,28 +13,18 @@
  #}
 
 {% if projectCode %}
-  <div class="status-bar grid-row">
-    <dl class="status-bar__section column-one-quarter">
+  <div class="status-bar l-container">
```
tyom: Wonder if we can reuse `c-meta-item` here which is used on investment collection page and do away with `status-bar` altogether?

src/templates/_layouts/datahub-base.njk
```diff
@@ -9,11 +9,13 @@
 {% from "_macros/_deprecated/trade-elements/textarea.njk" import textarea %}
 {% from "_macros/_deprecated/trade-elements/textbox.njk" import textbox %}
 
-{% from "_macros/form.njk" import EntitySearchForm, Form, MultiStepForm with context %}
+{% from "_macros/form.njk" import EntitySearchForm, Form, InlineForm, MultiStepForm with context %}
 {% from "_macros/form.njk" import Fieldset, TextField, MultipleChoiceField, HiddenField %}
+{% from "_macros/form.njk" import DateField %}
```
tyom: Merge it with the line above?

src/templates/_macros/elements.njk
```diff
@@ -0,0 +1,11 @@
+{##
+ # Render anchor markup
+ # @param {string} props.href - Anchor href
+ # @param {string} props.text - Anchor text
+ # @param {string} [props.target=_self] - Anchor target
+ #}
+{% macro Link(props) %}
```
tyom: Not sure how useful this macro is. It's just an HTML element with a couple of attributes that are already restricted to certain values. I think you could justify having it if it had custom classes or wrapper.

src/templates/_macros/form.njk
```diff
@@ -78,6 +78,32 @@
...
 {##
+ # Render simple form with inline button for posting hidden values
+ # @param {object} props - A form object
+ #}
+{% macro InlineForm(props) %}
+  {% set hiddenFields = props.hiddenFields | assign({'_csrf': csrfToken}) %}
+  {% set buttonText = props.buttonText or 'Submit' %}
+
+  {{ ErrorSummary(props.errors) }}
+  <form
```
tyom: Can you reuse existing `Form` macro here?

src/apps/investment-projects/views/create/investment-type.njk
```diff
@@ -0,0 +1,120 @@
...
+      </tr>
+    {% endif %}
+    </tbody>
+  </table>
+
+  {% set investmentOptions = [
+      {
+        value: form.options.investmentTypesObj.fdi.value,
+        label: form.options.investmentTypesObj.fdi.label,
+        hint: Link({
```
tyom: You could just save that link in a variable and pass it along to macro to avoid creating a new macro.

```nunjucks
{% set investmentTypeFdiLink %}
  <a href="/investment-projects/create/investment-type/info#fdi">Is this an FDI project?</a>
{% endset %}
```

tyom: Also, what is `info` target? The target is meant to be a frame name. I think you meant `_blank` to open in a new window. Maybe we can do an XHR to open inline? Perhaps a follow-up.

assets/stylesheets/components/_progress.scss
feedmypixel: `1060px` is when they can become centre aligned. The declaration on [line 88](https://github.com/uktrade/data-hub-frontend/pull/395/files/56479d71550351766c224c3c9fe0d6f56c3b7c33#diff-f8bf403629d8d51bfd4ce85105beedd7R88) controls how they behave on smaller screens (pulling them in)

assets/stylesheets/components/_progress.scss
feedmypixel: yeah was thinking about colourblind people and making it apparent enough that something has changed

src/templates/_layouts/datahub-base.njk
feedmypixel: was wondering if we need to denote a difference between form elements and form patterns. The DateField felt more "patterny" (for want of a real word) than the others

src/templates/_macros/elements.njk
feedmypixel: agreed - the reason it was built is the having to send it in to the hint for the radios on the `investment-type` page. I initially just sent in a html string with `| safe` but it felt really hacky. This felt a little cleaner. Open to changing if your good with that.

src/templates/_macros/form.njk
feedmypixel: So differences:
- Error summary outside of form
- Button displayed inside form (not in a form group)

It felt like adding flags to the existing form to handle this was the wrong approach.

src/templates/_layouts/datahub-base.njk
tyom: I'm not sure we need that kind of separation at this point. `MultipleChoiceField` is also quite "patterny" but to the end user (developer) it just a component with certain inputs.

src/apps/investment-projects/views/create/investment-type.njk
feedmypixel: `target: 'info'` is to open in new window named `info` so that it doesn't just open different blank pages when clicking different links

feedmypixel: > You could just save that link in a variable and pass it along to macro to avoid creating a new macro.
This answers the above - will do

src/templates/_macros/form.njk
tyom: Hmm. But you're forking the form here so every time there are global changes to form we must remember to update this one as well. Perhaps we should add a flag to disable `ErrorSummary`. You can already disable form submit button.

assets/stylesheets/components/_progress.scss
teneightfive: Maybe we need more than colour change on the icon. Could be filled for current state and empty for other states?

feedmypixel:
> Wonder if the investor company and code should be flipped to be consistent with investment projects pipeline layout.

is there a component that can be shared with this?

assets/stylesheets/components/form/_form-control.scss
feedmypixel: Do 
- `short`
- `shorter`
- `shortest` 

provide enough meaning when they are by themselves?

src/templates/_components/investment-status.njk
feedmypixel: sounds like a refactor/alignment for after this PR is in

assets/stylesheets/components/_status-bar.scss
feedmypixel: enforced  👮  🚨

feedmypixel:
> Initial sorting selection seems to have gone in investment pipeline.

@tyom where is this? I can't track it down

src/apps/investment-projects/views/_layout.njk
```diff
@@ -21,6 +25,31 @@
...
+    <p>
+      {% if investmentStatus.currentStage.isComplete %}
+        You have added all required information to complete this stage.
+      {% else %}
+        Complete this stage to move to the {{ investmentStatus.nextStage.name }} stage.
+      {% endif %}
+    </p>
+  {% endset %}
+
+  {% call InlineForm(form | assign({
```
ztolley: Just for me, can you explain why you use assign?

ztolley: And could this logic go in a controller instead of a view?

src/apps/investment-projects/controllers/create/equity-source.js
```diff
@@ -0,0 +1,80 @@
...
+
+      if (searchResult) {
+        searchResult.pagination = buildPagination(req, searchResult)
+      }
+
+      if (company && company.uk_based) {
+        showSearch = true
+      }
+
+      res.locals.company = company
```
ztolley: could you use 
```
res.locals = Object.assign({}, res.locals, {
  companyInvestments,
  searchTerm,
```

src/apps/investment-projects/controllers/create/project.js
```diff
@@ -0,0 +1,51 @@
...
+  const investmentDetails = req.store.get('investment_details')
+
+  if (investmentDetails) {
+    const investmentType = getMetaDataObj(metadata.investmentTypeOptions.map(transformObjectToOption), investmentDetails.investment_type)
+    const fdiType = getMetaDataObj(metadata.fdiOptions.map(transformObjectToOption), investmentDetails.fdi_type)
+    const nonFdiType = getMetaDataObj(metadata.nonFdiOptions.map(transformObjectToOption), investmentDetails.non_fdi_type)
+
+    res.locals.form = get(res, 'locals.form', { options: {} })
+    res.locals.form.options.investmentDetails = {
+      'investment_type': investmentType,
```
ztolley: does it need quotes?

src/apps/investment-projects/middleware/forms/details.js
```diff
@@ -103,7 +111,28 @@ function handleFormPost (req, res, next) {
     })
 }
 
+function validateForm (req, res, next) {
```
ztolley: Should the API be doing validation?

src/apps/investment-projects/middleware/index.js
```diff
@@ -1,9 +1,11 @@
+const collection = require('./collection')
 const forms = require('./forms')
 const shared = require('./shared')
-const collection = require('./collection')
+const team = require('./team')
```
ztolley: huh?

src/templates/_macros/form.njk
```diff
@@ -514,3 +521,43 @@
     {% endfor %}
   {% endif %}
 {% endmacro %}
+
```
ztolley: :)

assets/stylesheets/components/form/_form-group.scss
```diff
@@ -4,7 +4,8 @@
   max-width: 40em;
 
   [type="hidden"] + &,
-  .c-form-fieldset__legend + & {
+  .c-form-fieldset__legend + &,
```
ztolley: Some of this stuff should be split into a separate PR maybe as it affects things globally but then again I know that allot of effort for the same result so no worries.

src/apps/components/views/form.njk
```diff
@@ -180,6 +180,19 @@
...
       {% endcall %}
+
+      {{ DateFieldset({
+        name: 'date',
+        label: 'What is your favourite day?',
+        hint: 'A day you really like',
+        error: form.errors.messages.date,
+        inputName: 'date',
+        inputValue: {
+          year: form.state.dateYear,
```
ztolley: I wonder if you could have a field named date with sub properties or date in a set format and split with filters or just pass data in a set format and the control splits it.

How does it handle multiple dates in a form?

ztolley:
In a way this PR is actually 5 or 6 PR's in one. Things like the session storage, the steps support and the date control. But all good stuff.

feedmypixel:
> Just for me, can you explain why you use assign?

because I am merging options into the [form](https://github.com/uktrade/data-hub-frontend/pull/395/files#diff-11b25278c870c6bb98a48ece404c54a8R42). The form object can potentially hold error state. It feels more appropriate to have the properties in the object that is being assigned in the view layer.

feedmypixel:
> And could this logic go in a controller instead of a view?

Explained above

src/apps/investment-projects/middleware/index.js
feedmypixel: alphabeticalness

src/templates/_macros/form.njk
```diff
@@ -488,3 +527,43 @@
...
+{% macro DateFieldset(props) %}
+  {% set fieldId = 'field-' + props.name if props.name %}
+
+  {% call FormGroup(props | assign({ fieldId: fieldId, element: 'fieldset', modifier: 'inline' })) %}
+    {{ TextField({
+      modifier: ['inline', 'shortest'],
+      name: props.inputName + '_day',
+      value: props.inputValue.day,
+      label: 'Day',
+      hint: 'dd'
```
teneightfive: I would drop these hints from this component. We may well decide to support different formats, in fact, I think we should. We shouldn't require a use to enter `01` for the first of the month.

src/templates/_macros/form.njk
```diff
@@ -488,3 +527,43 @@
...
+ #}
+{% macro DateFieldset(props) %}
+  {% set fieldId = 'field-' + props.name if props.name %}
+
+  {% call FormGroup(props | assign({ fieldId: fieldId, element: 'fieldset', modifier: 'inline' })) %}
+    {{ TextField({
+      modifier: ['inline', 'shortest'],
+      name: props.inputName + '_day',
+      value: props.inputValue.day,
+      label: 'Day',
```
teneightfive: We also talked about removing the bold from these labels so may need a modifier somewhere.

src/templates/_macros/form.njk
feedmypixel: So the InlineForm macro has now gone, after a rethink there is no need for this macro

src/apps/investment-projects/middleware/forms/details.js
feedmypixel: This is validating the UI radio buttons, these button values are not sent to the API. They are just a UI pattern to make the entering `client_relationship_manager` and `referral_source_adviser` easier

src/apps/components/views/form.njk
feedmypixel: > How does it handle multiple dates in a form?

Its encapsulated in the macro. This is the `/components` example a better of example of how this can be handled can be seen [here](https://github.com/uktrade/data-hub-frontend/pull/395/commits/8147aea82d866c1a0751de1d9d90c5b6dfd968b8#diff-f6a833c01c454216b27811d03a17b2c4R304)

assets/stylesheets/components/_local-header.scss
```diff
@@ -14,18 +14,18 @@
...
-    margin-bottom: -$default-spacing-unit;
-  }
-
   .c-entity-search__aggregations {
     margin-bottom: -$default-spacing-unit * 2;
   }
+
+  .c-progress__stage {
+    &::before {
+      outline-color: $grey-4;
```
feedmypixel: This is needed as the progress is either within the header in `investment-projects` or on a white background in `/components`. It felt like adding it here and defaulting to white rather than providing a modifier was the cleaner way of doing it. This controls the gap around the circles and the progress line so `transparent won't work`

screen shot 2017-08-03 at 14 41 49

tyom:
> > Wonder if the investor company and code should be flipped to be consistent with investment projects pipeline layout.
>
> is there a component that can be shared with this?

No, there's no component. But we could match it stylistically. But that's something we could look at later on.

assets/stylesheets/components/_progress.scss
```diff
@@ -0,0 +1,152 @@
...
+    z-index: 20;
+  }
+
+  &.c-progress__bar--0-percent {
+    &::after {
+      width: 0;
+    }
+  }
+
+  &.c-progress__bar--25--percent {
```
feedmypixel: >  tries to get me to avoid modifiers on elements if possible

👍

assets/stylesheets/components/_progress.scss
```diff
@@ -0,0 +1,152 @@
...
+  &::before {
+    position: relative;
+    width: 14px;
+    height: 14px;
+    margin: -7px -7px 0;
+    content: "";
+    border-radius: 20px;
+    border: 2px solid $govuk-blue;
+    background-color: $govuk-blue;
+    outline: 2px solid $white;
```
feedmypixel: outline is providing the gap between the circle and the line. Mentioned above `transparent` won't work here.

screen shot 2017-08-03 at 14 41 49

assets/stylesheets/components/_progress.scss
```diff
@@ -0,0 +1,152 @@
...
+  }
+}
+
+.c-progress__stage-title {
+  position: relative;
+  left: -50%;
+}
+
+.c-progress__bar {
+  > li {
```
feedmypixel: with refactor these declarations have now gone

assets/stylesheets/components/_progress.scss
```diff
@@ -0,0 +1,152 @@
...
+      left: -85%;
+    }
+  }
+}
+
+@include media($min-width: 1060px) {
+  .c-progress__bar {
+    width: 100%;
+
+    > li {
```
feedmypixel: with refactor these declarations have now gone

assets/stylesheets/components/_progress.scss
```diff
@@ -0,0 +1,152 @@
...
+      &:last-child {
+        .c-progress__stage-title {
+          left: -50%;
+        }
+      }
+    }
+  }
+}
+
+.c-progress__stage--part-1 {
```
feedmypixel: I have avoided `:nth-child` so this work covers all browsers. I have updated the names to `c-progress__stage--one`, `c-progress__stage--two`, etc... This work is inside of a for loop so unlessI make `c-progress__stage` a block it would be tricky to add it to the block

assets/stylesheets/components/_status-bar.scss
```diff
@@ -2,39 +2,28 @@
 @import "../settings/layout";
 @import "../tools/mixins";
 
-.status-bar {
```
feedmypixel: 👍  have updated the markup and BEM names as `status-bar` styles have changend

assets/stylesheets/components/_progress.scss
tyom: I feel like the modifier should be applied to the `c-progress` and not to specifically to `__bar`. The modifier may affect other elements within the component. Say you wanted to turn to text colour to green when it's complete or add additional pseudo element. Conceptually the percentage can be applied to the whole progress not just the gauge measuring it.

src/apps/investment-projects/controllers/create/equity-source.js
```diff
@@ -0,0 +1,80 @@
...
+
+function renderEquitySourcePage (req, res, next) {
+  return res
+    .breadcrumb('Add investment project')
+    .render('investment-projects/views/create/equity-source')
+}
+
+function getHandler (req, res, next) {
+  const company = res.locals.company
+  const searchTerm = req.query['q']
```
feedmypixel: this was previously in `create-1`, will change to `term` to be the same as the home page search

src/apps/components/views/progress.njk
```diff
@@ -0,0 +1,11 @@
+{% extends "./_layout.njk" %}
+
+{% block body_main_content %}
+  <h2 class="page-heading">Progress Indicator</h2>
```
feedmypixel: yeah this really became apparent when adding to `/components`

src/apps/investment-projects/controllers/create/equity-source.js
```diff
@@ -0,0 +1,80 @@
...
+      res.locals.showSearch = showSearch
+
+      next()
+    })
+    .catch(next)
+}
+
+function postHandler (req, res, next) {
+  const isEquitySource = req.body['is_equity_source']
+  const clientCompanyId = req.body['company_id']
```
feedmypixel: yup - this page feels like it was doing a bit much in the first place.
- Radios for is equity source,
- search for equity search,
- search results for equity source

 But I decided to keep the changes to a minimum from the previous `create-1`. Once I understand more about how we are approaching flows next week I will look at this again and potentially split out.

src/apps/investment-projects/controllers/create/investment-type.js
```diff
@@ -0,0 +1,54 @@
...
+    'fdi_type': req.body.fdi_type,
+    'non_fdi_type': req.body.non_fdi_type,
+  })
+
+  if (
+    investmentTypeId === investmentTypeOptions.fdi.value ||
+    investmentTypeId === investmentTypeOptions.non_fdi.value
+  ) {
+    return res.redirect(`/investment-projects/create/equity-source/${companyId}`)
+  } else {
```
feedmypixel: yup - its error reporting if there has been no selection of a radio button

feedmypixel: have rejigged. This is clearer now. Previously a symptom of the Commitment to invest flow not being done yet

src/apps/investment-projects/middleware/forms/details.js
```diff
@@ -93,8 +99,10 @@ function handleFormPost (req, res, next) {
     .catch((err) => {
       if (err.statusCode === 400) {
         res.locals.form = get(res, 'locals.form', {})
```
feedmypixel: 👍  if `res.locals.form` is `undefined` `Object.assign` handles it appropriately

src/middleware/form.js
```diff
@@ -0,0 +1,23 @@
+/**
+ * Simple util to send form object when using a res.redirect. Mimics res.locals functionality
+ * @returns {form}
+ */
+module.exports = () => {
+  return function form (req, res, next) {
+    req.session.form = req.session.form || []
```
feedmypixel: this has gone now but no form was an `Array` which you pushed and popped form objects from onto `locals`

src/apps/transformers.js
```diff
@@ -21,7 +22,35 @@ function transformContactToOption ({ id, first_name, last_name }) {
   }
 }
 
+/**
+ * Utility to build an object from a transformed metadata array of objects so you can reference properties
+ * by key rather than array index. Helpful when the array length changes.
+ * @returns {{}}
+ */
+function buildMetaDataObj (collection) {
```
feedmypixel: `keyBy` has done the trick

src/middleware/form.js
```diff
@@ -0,0 +1,23 @@
...
+
+    req.form = (errors) => {
```
feedmypixel: this functionality has gone now

src/apps/investment-projects/views/_details-form.njk
```diff
@@ -1,170 +1,299 @@
...
+
+{% if investmentData.id %}
+  <div class="c-form-group">
+    {{ MultipleChoiceField({
+      type: 'radio',
+      name: 'investment_type',
+      label: 'Type of investment',
+      value: form.state['investment_type'],
+      error: form.errors.messages['investment_type'],
+      children:  [
```
tyom: You don't need `children`. You can just call that macro and nest stuff in it. We should avoid multiple ways of doing the same thing.

src/templates/_macros/form.njk
```diff
@@ -429,6 +438,10 @@
       </option>
     {% endfor %}
   </select>
+
+  {% for child in fieldChildren %}
```
tyom: I'm not convinced this is needed.

src/apps/investment-projects/views/_details-form.njk
tyom: [An example of call](https://github.com/uktrade/data-hub-frontend/blob/develop/src/apps/investment-projects/views/value-edit.njk#L12-L36).

feedmypixel: > You don't need children. 

do you mean use [caller](https://github.com/uktrade/data-hub-frontend/blob/dcc35618e8c93e603a714e1ef346a8557074c098/src/templates/_macros/form.njk#L329).

> We should avoid multiple ways of doing the same thing.

This `select` pattern is the [same pattern](https://github.com/uktrade/data-hub-frontend/blob/dcc35618e8c93e603a714e1ef346a8557074c098/src/templates/_macros/form.njk#L493-L495) used for `radio` and `checkbox` when using the [MultipleChoiceField macro](https://github.com/uktrade/data-hub-frontend/blob/dcc35618e8c93e603a714e1ef346a8557074c098/src/templates/_macros/form.njk#L320)

feedmypixel: Use case is on the create form

business

feedmypixel: will align this work to use `caller` rather than children

assets/stylesheets/components/_meta-list.scss
```diff
@@ -0,0 +1,11 @@
+@import "../settings";
+
+.c-meta-list {
+  .c-meta-item {
+    margin-right: $default-spacing-unit;
```
teneightfive: A left margin when there is a sibling might be better here. But could be addressed later during a spacing review.

```scss
.c-meta-item + .c-meta-item {
}
```

PR #396, opened 2017-07-31
elements macro namespace

feedmypixel:
A simple link macro that is passed into a form hint. Usage can be seen in #395 [here](https://github.com/uktrade/data-hub-frontend/pull/395/files#diff-ca07686579545e20ce7ab022b4d19452R39)

screen shot 2017-07-31 at 15 03 04

### DIFF
--- a/src/templates/_macros/elements.njk
+++ b/src/templates/_macros/elements.njk
@@ -1,0 +1,11 @@
+{##
+ # Render anchor markup
+ # @param {string} props.href - Anchor href
+ # @param {string} props.text - Anchor text
+ # @param {string} [props.target=_self] - Anchor target
+ #}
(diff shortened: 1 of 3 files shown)


src/templates/_macros/elements.njk
```diff
@@ -0,0 +1,11 @@
+{##
+ # Render anchor markup
+ # @param {string} props.href - Anchor href
+ # @param {string} props.text - Anchor text
+ # @param {string} [props.target=_self] - Anchor target
```
teneightfive: Could the target be left empty by default and not add the attribute if not present?

feedmypixel: This macro is going to be dropped after discussion with @tyom - closing PR

PR #400, opened 2017-07-31
Adjust highlight style

teneightfive:
This changes the bold text style to one using a background colour to
distinguish highlighted text.

It also adds the highlight term filter to data that is used for the
search.

### Inspiration

This is loosely based on a [highlight pattern](https://home-office-digital-patterns.herokuapp.com/patterns/highlighting-matches) currently being developed as part of the Home Office design system.

## Before

image
image
image
image

## After

image
image
image
image

### DIFF
--- a/config/nunjucks/filters.js
+++ b/config/nunjucks/filters.js
@@ -6,6 +6,7 @@ const {
   isArray,
   isPlainObject,
   isEmpty,
+  isString,
   pickBy,
   filter,
   isNil,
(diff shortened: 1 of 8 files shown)


config/nunjucks/filters.js
```diff
@@ -33,6 +33,10 @@ const filters = {
   highlight: (string, searchTerm) => {
     const regEx = new RegExp(`(${searchTerm})`, 'gi')
 
+    if (!string) {
```
tyom: Should you check for this first, before anything else? Perhaps ensure that it's a real string with Lodash's `isString`.

We've also been doing single line returns for simple checks like this. I think it's easier to read.

PR #401, opened 2017-08-01
Change dismiss text on messages components

teneightfive:
Noticed after seeing a notification on GitHub that the term _dismiss_ seems more appropriate than _close_.

## Before

image

## After

image

### DIFF
--- a/assets/javascripts/modules/messages.js
+++ b/assets/javascripts/modules/messages.js
@@ -26,7 +26,7 @@ const Messages = {
     if (element.className.indexOf('error') === -1) {
       let link = document.createElement('a')
 
-      link.innerHTML = 'Close'
+      link.innerHTML = 'Dismiss'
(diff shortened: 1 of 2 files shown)


assets/javascripts/modules/messages.js
```diff
@@ -26,7 +26,7 @@ const Messages = {
     if (element.className.indexOf('error') === -1) {
       let link = document.createElement('a')
 
-      link.innerHTML = 'Close'
+      link.innerHTML = 'Dismiss'
```
feedmypixel: do we need a capital here? I'd vote for all lowercase `dismiss`

teneightfive: I would say we do. It's the start of a sentence effectively and therefore should have a capital.

Plus, it looks strange lowercase. I would also use GitHub as a good guide and they have used a capital letter:

image

PR #402, opened 2017-08-01
Feature/invest team members

ztolley:
team

Allows a user to add, edit and remove team members.

Team members are displayed in the team page.
Clicking change goes to an edit page, more members can be added via the add another button or add one, save it and then return to add another.

Removal is done by clearing the team member to remove.

### DIFF
--- a/src/apps/investment-projects/controllers/team/details.js
+++ b/src/apps/investment-projects/controllers/team/details.js
@@ -1,17 +1,22 @@
 const {
   projectManagementLabels,
   clientRelationshipManagementLabels,
+  teamMembersLabels,
 } = require('../../labels')
 
 const {
   transformProjectManagementForView,
   transformClientRelationshipManagementForView,
+  transformTeamMembersForView,
 } = require('../../services/formatting')
 
 function getDetailsHandler (req, res, next) {
   try {
-    const projectManagementData = transformProjectManagementForView(res.locals.investmentData)
-    const clientRelationshipManagementData = transformClientRelationshipManagementForView(res.locals.investmentData)
+    const investmentData = res.locals.investmentData
+
+    const projectManagementData = transformProjectManagementForView(investmentData)
+    const clientRelationshipManagementData = transformClientRelationshipManagementForView(investmentData)
+    const teamMembersData = investmentData.team_members.map(transformTeamMembersForView)
 
     res
       .breadcrumb('Project team')
@@ -20,6 +25,8 @@ function getDetailsHandler (req, res, next) {
         projectManagementLabels,
         clientRelationshipManagementData,
         clientRelationshipManagementLabels,
+        teamMembersData,
+        teamMembersLabels,
       })
   } catch (error) {
     next(error)
--- a/src/apps/investment-projects/middleware/forms/team-members.js
+++ b/src/apps/investment-projects/middleware/forms/team-members.js
@@ -1,0 +1,81 @@
+const zipWith = require('lodash/zipWith')
+const { getAdvisers } = require('../../../adviser/repos')
+const { updateInvestmentTeamMembers } = require('../../repos')
+const { teamMembersLabels } = require('../../labels')
+const { transformObjectToOption } = require('../../../transformers')
+
+// Transform the form posted to an array of objects to save
+// and filter out any blanks
+function transformDataToTeamMemberArray (body) {
+  if (Array.isArray(body.adviser)) {
+    return zipWith(body.adviser, body.role, (adviser, role) => ({ adviser, role }))
+      .filter(member => member.adviser.length > 0)
+  } else if (body.adviser.length > 0) {
+    return [{
+      adviser: body.adviser,
+      role: body.role,
+    }]
+  }
+
+  return []
+}
+
+async function populateForm (req, res, next) {
+  try {
+    const investmentData = res.locals.investmentData
+    const advisersResponse = await getAdvisers(req.session.token)
+    const advisers = advisersResponse.results.map(transformObjectToOption)
+
+    const teamMembers = investmentData.team_members.map((teamMember) => ({
+      adviser: teamMember.adviser.id,
+      role: teamMember.role,
+    }))
+
+    // Add an extra blank record to allow user to add a team member
+    teamMembers.push({
+      adviser: null,
+      role: null,
+    })
+
+    res.locals.form = Object.assign({}, res.locals.form, {
+      labels: teamMembersLabels.edit,
+      state: {
+        teamMembers,
+      },
+      options: {
+        advisers,
+      },
+      buttonText: 'Save',
+      returnLink: `/investment-projects/${investmentData.id}/team`,
+    })
+
+    next()
+  } catch (error) {
+    next(error)
+  }
+}
+
+async function handleFormPost (req, res, next) {
+  try {
+    res.locals.projectId = req.params.id
+    const teamMembers = transformDataToTeamMemberArray(req.body)
+    await updateInvestmentTeamMembers(req.session.token, req.params.id, teamMembers)
+    next()
+  } catch (err) {
+    if (err.statusCode === 400) {
+      res.locals.form = Object.assign({}, res.locals.form, {
+        errors: err.error,
+        state: req.body,
+      })
+      next()
+    } else {
+      next(err)
+    }
+  }
+}
+
+module.exports = {
+  populateForm,
+  handleFormPost,
+  transformDataToTeamMemberArray,
+}
--- a/src/apps/investment-projects/middleware/team.js
+++ b/src/apps/investment-projects/middleware/team.js
@@ -1,4 +1,5 @@
 const { transformBriefInvestmentSummary } = require('../services/formatting')
+const { getAdviser } = require('../../adviser/repos')
 
 function getBriefInvestmentSummary (req, res, next) {
   try {
@@ -9,6 +10,29 @@ function getBriefInvestmentSummary (req, res, next) {
   }
 }
 
+async function expandTeamMembers (req, res, next) {
+  try {
+    // Fetch further details about any team members so their team can be displayed.
+    // Note - Looping through items and fetching data instead of using map
+    // as await cannot be used arrow functions
+    const teamMembers = []
+
+    for (const teamMember of res.locals.investmentData.team_members) {
+      const teamMemberAdvisor = await getAdviser(req.session.token, teamMember.adviser.id)
+      teamMembers.push({
+        adviser: teamMemberAdvisor,
+        role: teamMember.role,
+      })
+    }
+
+    res.locals.investmentData.team_members = teamMembers
+    next()
+  } catch (error) {
+    next(error)
+  }
+}
+
 module.exports = {
   getBriefInvestmentSummary,
+  expandTeamMembers,
 }
--- a/src/apps/investment-projects/repos.js
+++ b/src/apps/investment-projects/repos.js
@@ -66,14 +66,17 @@ function unarchiveInvestmentProject (token, investmentId) {
 }
 
 async function updateInvestmentTeamMembers (token, investmentId, investmentTeamMembers) {
-  // Todo - Delete all existing records before saving (in case an advisor was changes.)
+  // Delete all existing records before saving
+  await authorisedRequest(token, {
+    url: `${config.apiRoot}/v3/investment/${investmentId}/team-member`,
+    method: 'DELETE',
+  })
 
   // Loop through each of the team members, saving it
-  /*
   for (const investmentTeamMember of investmentTeamMembers) {
     await authorisedRequest(token, {
       url: `${config.apiRoot}/v3/investment/${investmentId}/team-member`,
-      mothod: 'POST',
+      method: 'POST',
       body: {
         investment_project: investmentId,
         adviser: investmentTeamMember.adviser,
@@ -81,7 +84,6 @@ async function updateInvestmentTeamMembers (token, investmentId, investmentTeamM
       },
     })
   }
-  */
 }
 
 module.exports = {
--- a/src/apps/investment-projects/repos.js
+++ b/src/apps/investment-projects/repos.js
@@ -65,6 +65,25 @@ function unarchiveInvestmentProject (token, investmentId) {
   })
 }
 
+async function updateInvestmentTeamMembers (token, investmentId, investmentTeamMembers) {
+  // Todo - Delete all existing records before saving (in case an advisor was changes.)
+
+  // Loop through each of the team members, saving it
+  /*
+  for (const investmentTeamMember of investmentTeamMembers) {
+    await authorisedRequest(token, {
+      url: `${config.apiRoot}/v3/investment/${investmentId}/team-member`,
+      mothod: 'POST',
+      body: {
+        investment_project: investmentId,
+        adviser: investmentTeamMember.adviser,
+        role: investmentTeamMember.role,
+      },
+    })
+  }
+  */
+}
+
 module.exports = {
   getCompanyInvestmentProjects,
   getInvestment,
@@ -74,4 +93,5 @@ module.exports = {
   getInvestmentProjectAuditLog,
   archiveInvestmentProject,
   unarchiveInvestmentProject,
+  updateInvestmentTeamMembers,
 }
--- a/src/apps/investment-projects/services/formatting.js
+++ b/src/apps/investment-projects/services/formatting.js
@@ -293,6 +293,14 @@ function transformClientRelationshipManagementForView (investmentData) {
   return result
 }
 
+function transformTeamMembersForView ({ adviser, role }) {
+  return {
+    adviser: adviser.name,
+    team: adviser.dit_team.name,
+    role: role,
+  }
+}
+
 module.exports = {
   transformInvestmentDataForView,
   transformInvestmentValueForView,
@@ -304,4 +312,5 @@ module.exports = {
   transformBriefInvestmentSummary,
   transformProjectManagementForView,
   transformClientRelationshipManagementForView,
+  transformTeamMembersForView,
 }
--- a/src/apps/investment-projects/views/team/edit-team-members.njk
+++ b/src/apps/investment-projects/views/team/edit-team-members.njk
@@ -1,0 +1,31 @@
+{% extends "../_layout.njk" %}
+
+{% block main_grid_right_column %}
+  <h2 class="heading-medium">Assign project team members</h2>
+
+  {% call Form(form) %}
+    {% if form.errors.team_members %}
+      <div class="error">{{ form.errors.team_members }}</div>
+    {% endif %}
+
+    {% for teamMember in form.state.teamMembers %}
+      {% call Fieldset({ legend: 'Team member' }) %}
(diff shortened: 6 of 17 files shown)


tyom:
Is this feature only expected to work with JavaScript?

src/apps/investment-projects/controllers/team/details.js
```diff
@@ -1,17 +1,20 @@
 const {
   projectManagementLabels,
   clientRelationshipManagementLabels,
+  teamMembersLabels,
 } = require('../../labels')
 
 const {
   transformProjectManagementForView,
   transformClientRelationshipManagementForView,
+  tranformTeamMembersForView,
```
jim68000: s/tranform/transform/

src/apps/investment-projects/middleware/forms/team-members.js
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  if (Array.isArray(body.adviser)) {
+    for (var pos = 0; pos < body.adviser.length; pos += 1) {
+      if (body.adviser[pos].length > 0) {
+        cleanBody.adviser.push(body.adviser[pos])
+        cleanBody.role.push(body.role[pos])
+      }
+    }
+  } else if (body.adviser.length > 0) {
```
jim68000: I think I'm being really dim here. Why are you testing to see something is an Array, and then testing to see if its length is longer than 0 in the else? Is there a case where body.adviser can be an object? Wouldn't a simple boolean else if (body.adviser) be clearer?

test/unit/apps/investment-projects/middleware/forms/team-members.test.js
```diff
@@ -0,0 +1,363 @@
...
+        const state = this.resMock.locals.form.state
+        expect(state.teamMembers[0].adviser).to.equal('1234')
+        expect(state.teamMembers[0].role).to.equal('manager')
+        expect(state.teamMembers[1].adviser).to.equal('4321')
+        expect(state.teamMembers[1].role).to.equal('director')
+        done()
+      })
+    })
+
+    it('should include a blank record to allow things to be added when there is existing members', (done) => {
```
jim68000: /s/is/are/

src/apps/investment-projects/middleware/forms/team-members.js
ztolley: Yes, if one is posted it's posted as a string

ztolley: Because a form posts a string for a single value and an array for multiple values, this method normalises the posted data into arrays.

ztolley:
@tyom The feature works without JS. A user saves a team member, is returned to the details screen then selects the edit option again and can add another, so it's 'functional'

src/apps/investment-projects/middleware/forms/team-members.js
```diff
@@ -0,0 +1,106 @@
...
+
+  return transformedBody
+}
+
+async function populateForm (req, res, next) {
+  try {
+    const investmentData = res.locals.investmentData
+    const advisersResponse = await getAdvisers(req.session.token)
+
+    const advisers = advisersResponse.results.map((adviser) => ({
```
teneightfive: There's a method in the base transformer that will do this. Can then just do:

```js
```

I noticed the adviser response comes back with a name property which combines `first_name` and `last_name` so we don't need to do anything different on the frontend with that response now.

src/apps/investment-projects/middleware/forms/team-members.js
```diff
@@ -0,0 +1,106 @@
...
+      options: {
+        advisers,
+      },
+      buttonText: 'Save',
+      returnLink: `/investment-projects/${investmentData.id}/team`,
+    })
+
+    next()
+  } catch (error) {
+    console.log(error)
```
teneightfive: Should we use `logger.error(error)` instead?

src/apps/investment-projects/middleware/forms/team-members.js
```diff
@@ -0,0 +1,106 @@
...
+    await updateInvestmentTeamMembers(req.session.token, req.params.id, teamMembers)
+    next()
+  } catch (err) {
+    if (err.statusCode === 400) {
+      res.locals.form = Object.assign({}, res.locals.form, {
+        errors: err.error,
+        state: req.body,
+      })
+      next()
+    } else {
```
teneightfive: Could drop the else and return the earlier `next()` if the block passes.

src/apps/investment-projects/middleware/team.js
```diff
@@ -9,6 +10,29 @@ function getBriefInvestmentSummary (req, res, next) {
   }
 }
 
+async function expandTeamMembers (req, res, next) {
+  try {
+    // Fetch further details about any team members so their team can be displayed.
+    // Note - Looping through items and fetching data instead of using map
```
teneightfive: Could potentially use `Promise.all`: https://stackoverflow.com/questions/36992590/asynchronous-map-function-that-awaits-returns-promise-instead-of-value

src/apps/investment-projects/services/formatting.js
```diff
@@ -293,6 +293,16 @@ function transformClientRelationshipManagementForView (investmentData) {
   return result
 }
 
+function transformTeamMembersForView (investmentData) {
```
teneightfive: Should these start to live in _transformers_? Might also be worth looking at the pattern used in the [base transformers](https://github.com/uktrade/data-hub-frontend/blob/develop/src/apps/transformers.js) where they expect an object with certain properties and return an object with new properties. Rather than doing maps inside these functions.

src/apps/investment-projects/views/team/edit-team-members.njk
```diff
@@ -0,0 +1,34 @@
...
+  <h2 class="heading-medium">Assign project team members</h2>
+
+  {% call Form(form) %}
+    <div class="js-add-another-fieldset">
+
+      {% for teamMember in form.state.teamMembers %}
+        <fieldset class="section">
+          <legend class="visually-hidden">Investment team member</legend>
+          {% if form.errors.team_members %}
+            <div class="error">{{ form.errors.team_members }}</div>
```
teneightfive: Could this not be adapted to use the current error macro which is included in the majority of templates already?

src/apps/investment-projects/middleware/forms/team-members.js
ztolley: indeed.. BAD console log

src/apps/investment-projects/middleware/forms/team-members.js
ztolley: We could but I think other places follow this pattern and it feels easier to read.

src/apps/investment-projects/middleware/team.js
ztolley: I find this easier to read then building the promises and then running promise all

src/apps/investment-projects/services/formatting.js
ztolley: I think that we need to refactor much that is in here, just like with Bens PR. Lets do it after this one.

src/apps/investment-projects/views/team/edit-team-members.njk
ztolley: can you point me at one?

src/apps/investment-projects/middleware/forms/team-members.js
teneightfive: I thought the linter picked this up anyway?

src/apps/investment-projects/services/formatting.js
teneightfive: Fair enough, but it would still be worth looking at the method used in transformers.

So accept an object with expected properties and return a transformed object rather than doing the map within this function and returning an array.

src/apps/investment-projects/views/team/edit-team-members.njk
teneightfive: https://github.com/uktrade/data-hub-frontend/blob/develop/src/templates/_macros/form.njk#L7-L25

teneightfive: You can pass an errors property to the form macro: https://github.com/uktrade/data-hub-frontend/blob/develop/src/templates/_layouts/form-wizard-step.njk#L14

ztolley: That's for the form isn't it, not an individual field, or in this case associated with 2 fields that can be thought of as one.

tyom:
Could tweak that flash message to better fit with other elements in header.

image

ztolley:
@tyom the flash message is incorrectly placed as when I took the screen grab there was a bug, but @feedmypixel has sinse fixed that

src/apps/investment-projects/views/team/edit-team-members.njk
```diff
@@ -0,0 +1,34 @@
...
+          {{ MultipleChoiceField({
+            name: 'adviser',
+            label: form.labels.adviser,
+            options: form.options.advisers,
+            value: teamMember.adviser,
+            initialOption: '-- Select an adviser --'
+          }) }}
+
+          {{ TextField({
+            name: 'role',
```
tyom: I noticed you have duplicate ids when there's more than one item in the loop. I've added an ability to provide an id suffix to form field macros so in this case you would add `idSuffix: loop.index` here and in `MultipleChoiceField` above.

tyom:
> @tyom the flash message is incorrectly placed as when I took the screen grab there was a bug, but @feedmypixel has sinse fixed that

Oh, I've also fixed that in #413.

src/apps/investment-projects/views/team/edit-team-members.njk
ztolley: excellent

src/apps/investment-projects/views/team/edit-team-members.njk
teneightfive: Ah right, yep, I see.

src/apps/investment-projects/middleware/forms/team-members.js
```diff
@@ -0,0 +1,102 @@
...
+// Strip out any blank lines
+// and normalise the posted data to be arrays of values.
+function cleanFormData (body) {
+  const cleanBody = {
+    adviser: [],
+    role: [],
+  }
+
+  if (Array.isArray(body.adviser)) {
+    for (var pos = 0; pos < body.adviser.length; pos += 1) {
```
tyom: `var`?

I think this reads better:
```js
function transformDataToTeamMemberArray (body) {
  return body.adviser.map((adviser, idx) => {
    return {
      adviser,
      role: body.role[idx],
    }
  })
}
```

src/apps/investment-projects/middleware/forms/team-members.js
```diff
@@ -0,0 +1,98 @@
...
+// Strip out any blank lines
+// and normalise the posted data to be arrays of values.
+function cleanFormData (body) {
+  const cleanBody = {
+    adviser: [],
+    role: [],
+  }
+
+  if (Array.isArray(body.adviser)) {
+    for (let pos = 0; pos < body.adviser.length; pos += 1) {
```
tyom: Could this also be refactored to use `map`?

ztolley: It's picking things from 2 arrays so old school seemed simplest

assets/javascripts/modules/add-another-fieldset.js
```diff
@@ -0,0 +1,90 @@
...
+  cacheEls () {
+    this.fieldSetCollections = this.document.querySelectorAll(this.selector)
+  },
+
+  addButton () {
+    // Add a button to the end of each fieldset
+    this.fieldSetCollections.forEach((fieldsetsWrapper) => {
+      const addButtonElement = this.document.createElement('a')
+      addButtonElement.href = '#'
+      addButtonElement.addEventListener('click', this.addClickHandler.bind(this), true)
```
tyom: Any reason you're using capturing phase?

test/unit/assets/javascripts/add-another-fielset.test.js
```diff
@@ -0,0 +1,91 @@
+const jsdom = require('jsdom')
+const { JSDOM } = jsdom
+const AddAnotherFieldset = require('~/assets/javascripts/modules/add-another-fieldset')
+
+const HTML = `
+  <html><body>
+  <div class="js-add-another-fieldset">
```
tyom: Wonder if you could use macro helper to construct this.

assets/javascripts/modules/add-another-fieldset.js
ztolley: huh?

test/unit/assets/javascripts/add-another-fielset.test.js
ztolley: if we did we should for all tests but for now this'll do

assets/javascripts/modules/add-another-fieldset.js
tyom: `true` as your third argument.

test/unit/assets/javascripts/add-another-fielset.test.js
tyom: If it works it would be good to use it so we don't need to convert as many in future.

assets/javascripts/modules/add-another-fieldset.js
ztolley: Oh cut and paste tastic. Well spotted

test/unit/assets/javascripts/add-another-fielset.test.js
ztolley: just wonder how long that would take and distract from the point of the card. do we have anything else doing this that could be used fast?

tyom: You could try this:
```js
const { getMacros } = require('~/test/unit/macro-helper')
const formMacros = getMacros('form')
const component = formMacros.renderWithCallerToDom('Fieldset')(
  formMacros.render('MultipleChoiceField', {
    label: 'Test label',
    name: 'adviser',
    value: 'wilma',
    options: [
      { label: 'Fred', value: 'fred' },
      { label: 'Wilma', value: 'wilma' },
    ],
  }),
  formMacros.render('TextField', {
    label: 'Test',
    name: 'role',
    value: 'Fred Smith',
  })
)
```
(as seen in `test/unit/macro/form.test.js`)

Which also reminds me that you should be using `Fieldset` macro rather than hardcode one (to enforce consistency). I'll add a comment in the relevant place.

src/apps/investment-projects/views/team/edit-team-members.njk
```diff
@@ -0,0 +1,36 @@
+{% extends "../_layout.njk" %}
+
+{% block main_grid_right_column %}
+  <h2 class="heading-medium">Assign project team members</h2>
+
+  {% call Form(form) %}
+    <div class="js-add-another-fieldset">
+
+      {% for teamMember in form.state.teamMembers %}
+        <fieldset class="section">
```
tyom: You should use `Fieldset` macro here.

test/unit/assets/javascripts/add-another-fielset.test.js
ztolley: Ok will do

src/apps/investment-projects/views/team/edit-team-members.njk
ztolley: would that support the need for errors at a fieldset, not an individual field level?

test/unit/assets/javascripts/add-another-fielset.test.js
ztolley: That just gives me the fragment for the fieldset but I need to pass a 'document' to the control and that includes a wrapper around one or more fieldsets.

tyom: You can render it to HTML string if you want. The helper has method for that.

src/apps/investment-projects/views/team/edit-team-members.njk
tyom: How do you mean errors at a fieldset? Surely errors are specific to fields.

tyom:
Should one be able to remove fields as well?

tyom:
I took a step back and thought what if there was another way to add fields (or fieldsets)  to page without having to do specific operations to added fields.

Your new client module is not really a module its name should reflect. It relies on a specific field (name) and doesn't feel like it could be applied to another fieldset on another page. This also splits that form configuration in multiple areas making the maintenance harder.

I think you would agree that having the form elements defined in once place and using the same form macros would be easier to maintain and reuse. We're already using a similar technique on investment projects page. Make XHR request and partially replace DOM. In this case we send just the fieldset and inject it with JavaScript on the client. Reuse is achieved by creating a macro in the template that wraps the fieldset and can receive different index depending where argument comes from - loop or XHR request.

I [created a branch](https://github.com/uktrade/data-hub-frontend/compare/feature/invest-team-members...xhr-fragment) as a proof of concept with a generic module that attempts to do that. I'd be glad to hear your thoughts on that approach. It could also replace another similar logic in deprecated directory. I think it could be generic enough to be used in a number of different scenarios. Could do with a few more tweaks such as loading indicator, more robustness and tests of course. This is just a prototype.

@feedmypixel @teneightfive your thoughts are welcomed.

ztolley:
The current code is generic, it could be applied to any fieldset and it's current form the screen works with or without JS.

Switching to something that uses XHR to get page fragments feels like overkill. All that is needed is something that takes a fieldset, duplicates it and resets the values.

I want to see what everyone else things in case I'm missing something, but I'm worried that even small cards like this drag on and on and we have a deadline

ztolley:
With respect to removing, the current version is far from ideal and yes we should iterate this design to look at removing. For now the user can delete something by selecting a blank value for the drop down.

Once we get into removal then things get more complex: what happens if I change a person AND delete in one action and how do I do that without JS?

ztolley:
After discussions, removed the JS element of this card and will merge that before discussions on XHR fragments vs node duplication

src/apps/investment-projects/repos.js
```diff
@@ -65,6 +65,27 @@ function unarchiveInvestmentProject (token, investmentId) {
   })
 }
 
+async function updateInvestmentTeamMembers (token, investmentId, investmentTeamMembers) {
+  // Delete all existing records before saving
+  await authorisedRequest(token, {
```
tyom: Should it by in `try/catch`?

src/apps/investment-projects/views/team/edit-team-members.njk
```diff
@@ -0,0 +1,35 @@
+{% extends "../_layout.njk" %}
+
+{% block main_grid_right_column %}
+  <h2 class="heading-medium">Assign project team members</h2>
+
+  {% call Form(form) %}
+    <div class="section">
```
tyom: Is this needed?

ztolley: The switch in styles for fieldsets does indeed mean this is no longer required.

src/apps/investment-projects/repos.js
ztolley: This would be caught by the caller.

PR #403, opened 2017-08-01
Fix various bugs with the support form

reupen:
This fixes a few bugs with the support form:

- Correctly displays a success message on successful submission rather than an error
- Fixes duplicate messages on the 'thank you' page
- Correctly sets the user's email address in the created ticket if an email address is provided

(Could get rid of the thank you template completely, though we might want to add something else to that page.)


src/apps/support/middleware.js
```diff
@@ -79,9 +79,13 @@ async function submitForm (req, res, next) {
     tags: [req.body.type],
   }
 
+  if (req.body.email) {
+    ticket.requester.email = req.body.email
```
teneightfive: Could set this property on the ticket object above with a fallback:

```js
const ticket = {
  ...
  requester: {
    email: req.body.email || undefined,
  }
  ...
}
```

PR #413, opened 2017-08-03
Fix flash message margins in local header

tyom:
### Before
image

### After
image


feedmypixel:
LGTM 👍  will remove from #395

PR #416, opened 2017-08-04
Add initial OMIS edit handling

teneightfive:
This adds a sub-app to handle editing the initial fields of an order and the route in via the work order view.

👉    [**Demo**](https://datahub-beta2-pr-416.herokuapp.com/omis/084b934e-9bde-4e2b-88d8-5fbeb17de54d) 👈 

## What it looks like

order omis orders dit data hub

### Note

It currently only supports editing the initial properties that can be edited and are on the order model.


src/apps/omis/apps/view/views/work-order.njk
```diff
@@ -5,4 +5,63 @@
...
+      <th class="c-answers-summary__title" scope="row">Company</th>
+      <td class="c-answers-summary__content">{{ values.company.name }}</td>
+    </tr>
+    <tr>
+      <th class="c-answers-summary__title" scope="row">Contact</th>
+      <td class="c-answers-summary__content">{{ values.contact.name }}</td>
+    </tr>
+  </table>
+
+  <table class="c-answers-summary">
```
tyom: Could avoid repetition by creating a local macro above and use it on this page.

src/apps/omis/middleware/params.js
```diff
@@ -6,9 +6,8 @@ async function getCompany (req, res, next, companyId) {
   try {
     res.locals.company = await getInflatedDitCompany(req.session.token, companyId)
     next()
-  } catch (e) {
-    logger.error(e)
-    res.redirect('/')
+  } catch (error) {
+    next(error)
```
feedmypixel: why is this file called `params` when it is getting an `InflatedDitCompany` and assigning it to `res.locals.company`? Feels like the wrong name?

assets/stylesheets/components/_answers-summary.scss
```diff
@@ -2,7 +2,30 @@
...
+  position: relative;
+}
+
+.c-answers-summary__heading-action {
+  @include bold-font(16);
+  display: block;
+
+  @include media(tablet) {
+    font-size: 20px;
+    float: right;
```
teneightfive: It's not really creating a grid as such though, it's just floating the edit action to the right of the heading on wider screens. Smaller screens it stacks below. We don't want to constrain both left and right to particular sizes like you would in a grid.

assets/stylesheets/components/_answers-summary.scss
```diff
@@ -2,7 +2,30 @@
 @import "../tools/mixins";
 
 .c-answers-summary {
-  border-top: 1px solid $grey-2;
+  margin: 0;
```
teneightfive: Possibly. Think this is trying to clear the margins applied to the global [table element]() which aren't very sensible in hind-sight. 

I think this could fall under our global typography and spacing overhaul but for now this is clearing the bottom margin. @tyom and I just had a discussion about margins and I think we should avoid bottom margins as much as possibly and apply top margins using sibling selectors:

https://alistapart.com/article/axiomatic-css-and-lobotomized-owls

Feels out of scope of this for now though.

src/apps/omis/apps/view/views/work-order.njk
```diff
@@ -5,4 +5,63 @@
 {% endblock %}
 
 {% block main_grid_right_column %}
+  <table class="c-answers-summary">
```
teneightfive: Could change to a `<dl>` instead?

src/apps/omis/middleware/params.js
teneightfive: Because it contains express [param middleware](http://expressjs.com/en/api.html#app.param) functions. The benefit of using these rather than just normal middleware is that they are only called once no matter how many times that param is used in the router.

[Example current usage](https://github.com/uktrade/data-hub-frontend/blob/develop/src/apps/omis/router.js#L8)

src/apps/omis/apps/view/views/work-order.njk
teneightfive: Although for now to support both this version and the existing version in the create flow I'm inclined to keep it as a table and look at changing the layout for both as a separate change. 

I've tried to use a minimal version for now just to play back grouped information and allow a route to edit. I'd like to do some testing to see how users get on it with and will iterate after that.

teneightfive:
>As discussed offline I think we're heading the wrong way with this whole classical inheritance emulation. And I think we agree on that and should look at finding a better alternative in near future.

Yep, I think given that we're seeing we could find ourselves quickly down a hole we should try and limit its use.

>As far as making it work to move forward this looks good to me.

But yes, it has provided an alternative angle on how we do form handling with some benefits but we should take those and try and find a nicer solution without using classical inheritance.

src/apps/omis/middleware/params.js
feedmypixel: sure but could it have a more meaningful name [like](https://github.com/uktrade/data-hub-frontend/blob/develop/src/apps/investment-projects/router.js#L42-L43)?

feedmypixel:
> As discussed offline I think we're heading the wrong way with this whole classical inheritance emulation. And I think we agree on that and should look at finding a better alternative in near future.

+ 1 on this. Lets gather what @teneightfive has learnt, see what our needs are and when we have time build something that fits those needs.

PR #417, opened 2017-08-05
Highlight filter

tyom:
- remove border from highlight (avoid interfering with potential highlighted element's existing styles
- Allow to restrict to full word match only as search can be configured to match words only
- add highlighters in searchable fields of results macro
- avoid highlighting empty strings

### Before
screen shot 2017-08-05 at 14 40 01
screen shot 2017-08-05 at 14 40 24

### After
screen shot 2017-08-05 at 14 41 26
screen shot 2017-08-05 at 14 42 12

### DIFF
--- a/assets/stylesheets/trumps/_utility.scss
+++ b/assets/stylesheets/trumps/_utility.scss
@@ -28,8 +28,8 @@
 
 .u-highlight {
   background-color: $yellow-25;
-  border-bottom: 1px solid $yellow;
-  padding: 2px 1px;
+  padding: 2px;
+  margin: -2px;
 }
 
 .u-clearfix {
--- a/config/nunjucks/filters.js
+++ b/config/nunjucks/filters.js
@@ -33,13 +33,13 @@ const filters = {
   mapValues,
   isArray,
 
-  highlight: (string, searchTerm) => {
-    if (!isString(string)) { return }
+  highlight (string, searchTerm, shouldMatchFullWord = false) {
+    if (!isString(string) || !isString(searchTerm) || !searchTerm.trim()) { return string }
 
-    const regEx = new RegExp(`(${searchTerm})`, 'gi')
-    return new nunjucks.runtime.SafeString(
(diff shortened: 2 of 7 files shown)


assets/stylesheets/trumps/_utility.scss
```diff
@@ -28,8 +28,8 @@
 
 .u-highlight {
   background-color: $yellow-25;
-  border-bottom: 1px solid $yellow;
```
teneightfive: 😢 I'm not convinced on removing the border.

config/nunjucks/filters.js
```diff
@@ -33,13 +33,18 @@ const filters = {
   mapValues,
   isArray,
 
-  highlight: (string, searchTerm) => {
-    if (!isString(string)) { return }
+  highlight: (string, searchTerm, shouldMatchFullWord = false) => {
+    let result
```
teneightfive: Could move after the check below.

config/nunjucks/filters.js
```diff
@@ -33,13 +33,18 @@ const filters = {
...
+    if (!isString(string) || !isString(searchTerm)) { return string }
 
-    const regEx = new RegExp(`(${searchTerm})`, 'gi')
-    return new nunjucks.runtime.SafeString(
-      string.replace(regEx, '<span class="u-highlight">$1</span>')
-    )
+    if (!searchTerm.trim()) {
```
teneightfive: Why not add this to the check above? Does this string need to be returned with `SafeString`?

assets/stylesheets/trumps/_utility.scss
tyom: I was not convinced on adding it in the first place 😉

config/nunjucks/filters.js
tyom: Good point.

assets/stylesheets/trumps/_utility.scss
teneightfive: You were never convinced the world was round 😂

PR #418, opened 2017-08-05
Use Results macro in investment search results

tyom:
Replaces hard investment search results template for Results macro. 

This is based on #417 for improved search term highlighting.

**Next:** Add filters from investment projects collection.

### Before
localhost-3000-search-investment-projects-term invest

### After
localhost-3000-search-investment-projects-term invest 1


src/apps/investment-projects/middleware/collection.js
```diff
@@ -53,7 +46,27 @@ async function getInvestmentProjectsCollection (req, res, next) {
   }
 }
 
+function getRequestBody (req, res, next) {
```
ztolley: Kind of reminds me of rails using separate functions to get a clean set of params.

src/apps/search/controllers.js
```diff
@@ -47,7 +51,42 @@ async function viewCompanyResult (req, res, next) {
   }
 }
 
+async function renderInvestmentProjects (req, res) {
+  const searchTerm = get(req, 'query.term', '').trim()
```
teneightfive: Do you need `get` in this instance? `query` always exists on `req` so could just do `(req.query.term || '').trim()`?

tyom: I feel like using one consistent way to set defaults is preferential to different value-specific ways.

PR #419, opened 2017-08-05
Add ability to select multiple filter values in investment projects

tyom:
- Multiple values are added under the same reset button in results summary
- Fixes partial dates in results summary

### Before
localhost-3000-investment-projects-custom true sortby estimated_land_date-asc stage 8a320cc9-ae2e-443e-9d26-2f36452c2ced c9864359-fb1a-4646-a4c1-97d10189fc03 7606cc19-20da-4b74-aba1-2cec0d753ad8 investment_type 3e143372-496c-4d1e-8278-6fdd3

### After
localhost-3000-investment-projects-custom true sortby estimated_land_date-asc estimated_land_date_before 2018 estimated_land_date_after 2017-04-05


assets/stylesheets/components/_results-summary.scss
```diff
@@ -34,9 +34,11 @@
 .c-results-summary__filter {
   color: $text-colour;
   display: inline-block;
+  position: relative;
   margin-top: $default-spacing-unit / 2;
-  padding: 0 ($default-spacing-unit / 2);
+  padding: 0 ($baseline-grid-unit * 6) 0 ($default-spacing-unit / 2);
```
ztolley: What is the difference between default spacing unit and baseline grid unit, and when do you use them?

src/apps/investment-projects/middleware/collection.js
```diff
@@ -35,7 +35,7 @@ async function getInvestmentProjectsCollection (req, res, next) {
 
   const requestBody = Object.assign({}, {
     sortby: req.query.sortby,
-  }, selectedFiltersQuery)
+  }, pickBy(selectedFiltersQuery, x => x))
```
ztolley: not immediately obvious that the pickby function with this syntax in lodash removed null properties.

maybe a quick comment ?

assets/javascripts/lib/xhr.js
```diff
@@ -30,7 +30,7 @@ const XHR = {
     this.injectResponseInHtml(res.data)
 
     if (params) {
-      history.push(buildQueryString(params), { data: res.data })
```
ztolley: Is this fixing query string or using a different tool to get the same result?

If this and  the other commits replace the use of buildQueryString throughout the application then that should be it's own PR.

assets/stylesheets/components/_results-summary.scss
tyom: It's all a bit arbitrary. `baseline-grid-unit` is a quarter of `default-spacing-unit` and equals to `4px`. We use them to try to enforce a form of snap grid unit.

We talked about reviewing our typography and grid soon. We should come up with a better way to maintain consistent spacing (function perhaps?).

teneightfive: Yeah they've become a bit meaningless to be honest. I've also seen the use of `$gutter` which is a unit carried over from _govuk-elements_.

We should decide on a consistent way of spacing elements both internally and how they stack next to each other in certain regions of the layout.

teneightfive: Smells like quite a rabbit hole though...

src/apps/investment-projects/middleware/collection.js
tyom: I'm trying to avoid adding comments in favour of self documenting code.

`pickBy` is a Lodash function that picks by predicate function which in this case is identity function which just returns the value which predicate evaluates being truthy of falsey.

It's a pretty common utility in FP toolbelt. Actually identity function is the default function in many Lodash utilities so it can be omitted altogether.

src/apps/companies/controllers/add.js
```diff
@@ -63,13 +63,13 @@ function postAddStepOne (req, res, next) {
       break
   }
 
-  const queryString = buildQueryString(params)
+  const qs = `?${queryString.stringify(params)}`
```
teneightfive: `qs` feels a bit like using `e` to reference an error. Granted you have now defined `queryString` as the library though.

src/apps/companies/controllers/add.js
```diff
@@ -1,6 +1,6 @@
+const queryString = require('query-string')
```
teneightfive: I tend to add a new line between external and internal dependencies so it's easy to scan the top of a file and see the separation.

assets/javascripts/lib/xhr.js
tyom: It is fixing the query string in this specific instance as `buildQueryString` doesn't handle multiple values correctly. It should create multiple items with the same key in query string as `query-string` does.

After I fixed it in a couple of places here I thought there is not point having both and I removed the one that doesn't work correctly.

As this PR is quite small I chose to add the commit removing it here instead of creating a new 
 PR.

src/apps/companies/controllers/add.js
tyom: Suggestions?

assets/javascripts/lib/xhr.js
ztolley: if it was me I'd still break it out to it's own PR to make it clear what it's purpose is.

src/templates/_macros/entities.njk
```diff
@@ -62,7 +62,8 @@
   {% set isLabelHidden = props.isLabelHidden | default(false) %}
   {% set itemValueClass = 'c-badge ' + badgeModifier if props.isBadge else 'c-meta-item__value'%}
   {% if props.name and 'date' in props.name %}
-    {% set metaItemValue = props.value | formatDate('MMMM D, YYYY')  %}
```
ztolley: This could be a PR too

assets/javascripts/lib/xhr.js
tyom: Extracted to #421.

src/templates/_macros/entities.njk
tyom: It's a single commit change in a relatively small PR. I could extract it but I don't see much point.

PR #422, opened 2017-08-07
move between investment project stages

feedmypixel:
Add the ability to move to the next Investment project stage once all qualifications have been met

### Project stage progress component
move-stage

### Error state
Of note this is a total edge case and in the gif I have had to manipulate the api flags.
If for whatever reason the flags are incorrect and the user tries to move stage and can't they will be met with this flash error. **very unlikely to happen but is covered if it does**

move-stage-error

### DIFF
--- a/src/apps/investment-projects/middleware/forms/project-stage.js
+++ b/src/apps/investment-projects/middleware/forms/project-stage.js
@@ -1,0 +1,26 @@
+const logger = require('../../../../../config/logger')
+const { updateInvestment } = require('../../repos')
+
+function handleFormPost (req, res, next, projectId = req.params.id) {
+  updateInvestment(req.session.token, projectId, {
+    stage: {
+      id: req.body.next_project_stage,
+    },
+  })
+    .then(() => {
+      req.flash('success', `Investment project moved to ${res.locals.investmentStatus.nextStage.name} stage`)
+      res.redirect(`/investment-projects/${res.locals.investmentData.id}/details`)
+    })
+    .catch((err) => {
+      if (err.statusCode === 400) {
+        logger.error(err)
+        req.flash('error', 'Something has gone wrong')
+        return res.redirect(`/investment-projects/${res.locals.investmentData.id}/details`)
+      }
+      next(err)
+    })
+}
+
+module.exports = {
+  handleFormPost,
+}
--- a/src/apps/investment-projects/router.js
+++ b/src/apps/investment-projects/router.js
@@ -23,6 +23,7 @@ const {
   interactionsFormMiddleware,
   investmentTypeFormMiddleware,
   projectManagementFormMiddleware,
+  projectStageFormMiddleware,
   requirementsFormMiddleware,
   valueFormMiddleware,
   teamMembersFormMiddleware,
@@ -170,5 +171,7 @@ router
     interactions.edit.editPostInteractionHandler,
     interactions.edit.editGetInteractionHandler
   )
+
+router.post('/:id/change-project-stage', projectStageFormMiddleware.handleFormPost)
 
 module.exports = router
--- a/src/apps/investment-projects/views/_layout.njk
+++ b/src/apps/investment-projects/views/_layout.njk
@@ -28,6 +28,23 @@
 {% endblock %}
 
 {% block main_grid_right_column %}
+  {% call Form({
+    buttonText: 'Move to ' + investmentStatus.nextStage.name + ' stage',
+    action: 'change-project-stage',
+    class: 'flash flash--info',
+    disableFormActions: true if not investmentStatus.currentStage.isComplete,
+    hiddenFields: {
+      next_project_stage: investmentStatus.nextStage.id
+    }
+  }) %}
+    <p>
+      {% if investmentStatus.currentStage.isComplete %}
+        You have added all required information to complete this stage.
+        {% else %}
+        Complete this stage to move to the {{ investmentStatus.nextStage.name }} stage.
+      {% endif %}
+    </p>
+  {% endcall %}
 {% endblock %}
 
 {% block main_grid_left_column %}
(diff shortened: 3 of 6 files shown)


src/apps/investment-projects/views/_layout.njk
```diff
@@ -25,6 +25,27 @@
...
+  {% call Form({
+    buttonText: 'Move to ' + investmentStatus.nextStage.name + ' stage',
+    action: 'change-project-stage',
+    class: 'flash flash--info',
+    disableFormActions: true if not investmentStatus.currentStage.isComplete,
+    hiddenFields: {
+      next_project_stage: investmentStatus.nextStage.id
+    }
+  }) %}
+    {{ formContent | safe }}
```
tyom: Why not set the content directly in the macro instead of passing this variable?

src/apps/investment-projects/middleware/forms/project-stage.js
```diff
@@ -0,0 +1,29 @@
+const logger = require('../../../../../config/logger')
+const { updateInvestment } = require('../../repos')
+
+function handleFormPost (req, res, next, projectId = req.params.id) {
+  const projectPage = req.get('Referrer').split('/').pop()
```
teneightfive: Maybe something like `returnPath` would be easier to read.

Could also be worth setting `req.path` as a hidden field in the form before it is posted as a way of redirecting back to the current view rather than trying to chop up the referrer.

src/apps/investment-projects/middleware/forms/project-stage.js
```diff
@@ -0,0 +1,29 @@
...
+    .then(() => {
+      req.flash('success', `Investment project moved to ${res.locals.investmentStatus.nextStage.name} stage`)
+      return res.redirect(`/investment-projects/${res.locals.investmentData.id}/${projectPage}`)
+    })
+    .catch((err) => {
+      if (err.statusCode === 400) {
+        logger.error(err)
+        req.flash('error', 'Something has gone wrong')
+        return res.redirect(`/investment-projects/${res.locals.investmentData.id}/${projectPage}`)
+      } else {
```
teneightfive: Don't need the `else` if you're returning in the statement above.

src/apps/investment-projects/middleware/forms/project-stage.js
```diff
@@ -0,0 +1,29 @@
...
+  const projectPage = req.get('Referrer').split('/').pop()
+
+  updateInvestment(req.session.token, projectId, {
+    stage: {
+      id: req.body.next_project_stage,
+    },
+  })
+    .then(() => {
+      req.flash('success', `Investment project moved to ${res.locals.investmentStatus.nextStage.name} stage`)
+      return res.redirect(`/investment-projects/${res.locals.investmentData.id}/${projectPage}`)
```
teneightfive: Don't need to return this statement.

src/apps/investment-projects/router.js
```diff
@@ -171,4 +172,8 @@ router
     interactions.edit.editGetInteractionHandler
   )
 
+router.post('/:id/change-project-stage',
+  projectStageFormMiddleware.handleFormPost,
```
teneightfive: One line if it's just one middleware?

src/apps/investment-projects/views/_layout.njk
```diff
@@ -25,6 +25,23 @@
 {% endblock %}
 
 {% block main_grid_right_column %}
+  {% call Form({
+    buttonText: 'Move to ' + investmentStatus.nextStage.name + ' stage',
```
teneightfive: These actions could do with a content review. Something like `Project has been won` might be better than `Move to Won stage`. It gives a clearer indication as to when the button should be used and what action should trigger it. Rather than a button that says you can arbitrarily move the thing to a different stage.

src/apps/investment-projects/views/_layout.njk
```diff
@@ -25,6 +25,23 @@
...
+    action: 'change-project-stage',
+    class: 'flash flash--info',
+    disableFormActions: true if not investmentStatus.currentStage.isComplete,
+    hiddenFields: {
+      next_project_stage: investmentStatus.nextStage.id
+    }
+  }) %}
+    <p>
+      {% if investmentStatus.currentStage.isComplete %}
+        You have added all required information to complete this stage.
```
teneightfive: Feels superfluous at this stage.

src/apps/investment-projects/views/_layout.njk
```diff
@@ -25,6 +25,23 @@
 {% endblock %}
 
 {% block main_grid_right_column %}
+  {% call Form({
+    buttonText: 'Move to ' + investmentStatus.nextStage.name + ' stage',
+    action: 'change-project-stage',
+    class: 'flash flash--info',
+    disableFormActions: true if not investmentStatus.currentStage.isComplete,
```
teneightfive: This property feels like it's not doing what it says. Isn't it hiding the form actions rather than disabling them?

It might also be better to always show the action that can be taken but show it as disabled before it can be used.

src/apps/investment-projects/middleware/forms/project-stage.js
feedmypixel: as discussed just going to send people to `/investment-projects/<id>/details`

src/apps/investment-projects/views/_layout.njk
feedmypixel: Agreed. As discussed all the UI patterns round here need work as mentioned in #395 
It would be good to give mills context for the copy when this work is in. Once we know what we are doing in this area more we can rework this.

src/apps/investment-projects/views/_layout.njk
feedmypixel: #439 created to approach this in a different PR

src/apps/investment-projects/views/_layout.njk
feedmypixel: As this area is all in flux lets leave this as is and return in the future. This will all have some rework with mills

src/apps/investment-projects/views/_layout.njk
tyom: @teneightfive Surely the action button label be a verb?

PR #423, opened 2017-08-07
progress component

feedmypixel:
Add a progress component to demonstrate where you are in the flow

### Progress component
progress

### Progress
screen shot 2017-08-10 at 11 37 54

### DIFF
--- a/assets/stylesheets/components/_progress.scss
+++ b/assets/stylesheets/components/_progress.scss
@@ -1,0 +1,86 @@
+@import "../settings";
+@import "../tools";
+
+$_section-width: 25%;
+
+.c-progress {
+  margin: $default-spacing-unit * 2 $default-spacing-unit / 2;
+
+  @include media(tablet) {
+    margin: $default-spacing-unit * 2;
+    padding-bottom: $default-spacing-unit / 2;
+  }
+}
+
+.c-progress-bar {
+  @include media(tablet) {
+    display: table;
+    width: 100%;
+    position: relative;
+  }
+}
+
+.c-progress__stage {
+  @include core-font(14);
+  position: relative;
+  border-color: $grey-3;
+  border-width: 0 0 0 2px;
+  border-style: solid;
+
+  @include media(tablet) {
+    display: table-cell;
+    width: $_section-width;
+    border-width: 2px 0 0;
+  }
+
+  &::before {
+    position: absolute;
+    left: -8px;
+    width: 14px;
+    height: 14px;
+    content: "";
+    border-radius: 20px;
+    background-color: $white;
+    border: 2px solid $grey-3;
+    outline: 2px solid $white;
+    display: block;
+    z-index: 30;
+
+    @include media(tablet) {
+      top: -8px;
+    }
+  }
+
+  &.is-complete {
+    border-color: $govuk-blue;
+
+    &::before {
+      background-color: $govuk-blue;
+      border: 2px solid $govuk-blue;
+    }
+  }
+
+  &.is-active {
+    &::before {
+      background-color: $white;
+      border: 2px solid $govuk-blue;
+    }
+  }
+
+  &:last-child {
+    border-color: transparent;
+  }
+}
+
+.c-progress__stage-title {
+  display: block;
+  padding: 0 0 $default-spacing-unit $default-spacing-unit;
+
+  @include media(tablet) {
+    position: absolute;
+    left: -50px;
+    width: 100px;
+    text-align: center;
+    padding: $default-spacing-unit 0 0;
+  }
+}
--- a/src/apps/components/router.js
+++ b/src/apps/components/router.js
@@ -9,6 +9,7 @@ const {
   renderLocalHeader,
   renderBreadcrumbs,
   renderPagination,
+  renderProgress,
   renderResults,
 } = require('./controllers')
 
@@ -24,6 +25,7 @@ router
   .get('/local-header', renderLocalHeader)
   .get('/pagination', renderPagination)
   .get('/results', getInvestmentProjectsCollection, renderResults)
+  .get('/progress', renderProgress)
   .get('/form', handleEntitySearch, renderFormElements)
   .post('/form', handleFormPost, renderFormElements)
 
--- a/src/apps/components/views/progress.njk
+++ b/src/apps/components/views/progress.njk
@@ -1,0 +1,10 @@
+{% extends "./_layout.njk" %}
+
+{% block body_main_content %}
+  {% call Example('Progress') %}
+    {{ Progress({
+      stageNames: stageNames,
+      currentStageName: currentStageName
+    }) }}
+  {% endcall %}
+{% endblock %}
--- a/src/templates/_macros/common.njk
+++ b/src/templates/_macros/common.njk
@@ -115,3 +115,31 @@
     </nav>
   {% endif %}
 {% endmacro %}
+
+
+{##
+ # Render progress bar
+ # @param {object} props
+ # @param {array} props.stageNames - stage names
+ # @param {string} props.currentStageName - the current stages name
+ #
+ #}
+{% macro Progress(props) %}
+  {% if props.stageNames | length %}
+    {% set currentStageIndex = props.stageNames.indexOf(props.currentStageName) %}
+    <div class="c-progress">
+      <ol class="c-progress-bar">
+        {% for stageName in props.stageNames %}
+          <li class="c-progress__stage
+          {{- ' is-active' if props.currentStageName == stageName -}}
+          {{- ' is-complete' if loop.index0 < currentStageIndex -}}"
+          >
+            <span class="c-progress__stage-title">{{ stageName }}</span>
+          </li>
+        {% endfor %}
+      </ol>
+    </div>
+  {% endif %}
+{% endmacro %}
+
+
--- a/assets/stylesheets/_components.scss
+++ b/assets/stylesheets/_components.scss
@@ -16,6 +16,7 @@
 @import "components/metadata-list";
 @import "components/pagination";
 @import "components/phase-badge";
+@import "components/progress";
 @import "components/proposition-menu";
 @import "components/results";
 @import "components/results-summary";
--- a/assets/stylesheets/components/_local-header.scss
+++ b/assets/stylesheets/components/_local-header.scss
@@ -21,6 +21,12 @@
   .c-entity-search__aggregations {
     margin-bottom: -$default-spacing-unit * 2;
   }
+
+  .c-progress__stage {
+    &::before {
+      outline-color: $grey-4;
+    }
+  }
 }
 
 .c-local-header__heading-before {
--- a/src/apps/components/controllers.js
+++ b/src/apps/components/controllers.js
@@ -94,6 +94,21 @@ async function renderEntityList (req, res) {
     })
 }
 
+function renderProgress (req, res) {
+  return res
+    .breadcrumb('Progress')
+    .render('components/views/progress', {
+      stageNames: [
+        'one',
+        'two',
+        'three',
+        'four',
+        'five',
+      ],
+      currentStageName: 'three',
+    })
+}
+
 module.exports = {
   renderEntityList,
   renderFormElements,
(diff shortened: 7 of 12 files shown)


assets/stylesheets/components/_progress.scss
```diff
@@ -0,0 +1,121 @@
+@import "../settings";
+@import "../tools";
+
+$section-width: 25%;
```
teneightfive: Sometimes it can be useful to use an underscore prefix for local component variables. Although sass doesn't support them it can help to understand that it is only being used in this component.

assets/stylesheets/components/_progress.scss
```diff
@@ -0,0 +1,121 @@
...
+    z-index: 10;
+  }
+
+  &::after {
+    background-color: $govuk-blue;
+    z-index: 20;
+    width: 0;
+  }
+
+  &.c-progress-bar--one-quarter {
```
teneightfive: You can move these outside of this nesting so make them easier to read. BEM nicely aides the ability to avoid lots of nesting in Sass.

src/apps/components/router.js
```diff
@@ -24,6 +25,7 @@ router
   .get('/local-header', renderLocalHeader)
   .get('/pagination', renderPagination)
   .get('/results', getInvestmentProjectsCollection, renderResults)
+  .get('/progress', renderProgress)
```
teneightfive: Wonder if we're reaching the point to find a more dynamic way to render components. Maybe just a view and a config file which sets the data for it? And then loop over directories to load the routes for those into this router.

Obviously a separate PR but some of components is starting to feel quite repetitive now.

teneightfive:
Looks good but I have some wider concerns around progress indicators. I think we really need to test this with users to see if it actually adds value. Might be worth having a read of the [progress indicators page](https://paper.dropbox.com/doc/Progress-indicators-FKi7F1KBjks0jGG5vMdVg) on hackpad again to understand a lot of the research that has been done.

I know some of them are a different context but some of the points are still valid:
>take up valuable space and don’t scale well on small screens

I think this may need some further effort to work well on small devices. It doesn't allow for many characters and feels very squashed:

image

If we do feel it is useful we could do something similar to the prison visits example on hackpad and possibly hide the non-active stages as the main thing we're trying to communicate is what stage the project is currently on.

feedmypixel:
> I think we really need to test this with users to see if it actually adds value. 
Agreed. I think we also need to work out whats going on with the wider context of Investment Projects.

> take up valuable space and don’t scale well on small screens
This has been designed to be minimal and doesn't scale well `< 360px` but that could be fixed potentially as we were saying by stacking.

assets/stylesheets/components/_progress.scss
feedmypixel: Doesn't the fact its in this file means it should only be being used here? Otherwise it should sit somewhere in `assets/stylesheets/settings`

ztolley:
I'd consider either hiding or creating a vertical variation for mobile

assets/stylesheets/components/_progress.scss
ztolley: I still wonder if there is some way to make this work for a variable number of steps. What happens when the next multi stage thing comes along?

feedmypixel: When that happens we can build it. YAGNI

teneightfive: It's just a convention I've used and found helpful. 

I think it also helps when looking at declarations. You can see that if a variable is prefixed with an underscore you know it is set in this file, otherwise you will need to look in global settings for its value.

teneightfive: I wonder if there's a way to make use of `display: table-cell` to avoid having to set fixed widths. That way you won't need to change this if the amount of items changes.

There was [one created](https://gist.github.com/matharden/e8c56e484f5a03de4750) for a service at MOJ that used table layouts. Not sure if any of it could be re-used?

tyom: The fact that it is in the file still makes it global. To make it truly local it must be defined in the scope of selector.

The underscore prefix helps to communicate that it is meant to be a local declaration due to Sass limitation.

teneightfive: Yep, it's doesn't actually _make_ it global, it's just a convention that helps to see from a declaration that a local variable is being used rather than a variable defined outside of this file.

feedmypixel: will do 👍

feedmypixel: > display: table-cell to avoid having to set fixed widths. That way you won't need to change this if the amount of items changes

FYI I've added the `table-cell` work. However we still need to use the `$_section-width: 25%` variable to define the widths of the `table-cells` and to extend the progress bar for:
* `c-progress-bar--one-quarter`
* `c-progress-bar--half`
* `c-progress-bar--three-quarter`
* `c-progress-bar--full`

And for:

`c-progress__stage`

feedmypixel:
@teneightfive FYI this has now been made to work on mobile, made into a macro and changed to `display: table` layout. Thanks for the pointers on this.

assets/stylesheets/components/_progress.scss
```diff
@@ -0,0 +1,84 @@
...
+  border-style: solid;
+
+  @include media(tablet) {
+    display: table-cell;
+    width: $_section-width;
+    border-width: 2px 0 0 0;
+  }
+
+  &::before {
+    position: relative;
```
teneightfive: What about positioning this absolutely so that you don't have to mess around with `positive: relative` and `top: -33px` on `.c-progress__stage-title`?

Then just use `top` and `left` to position rather than minus margins.

assets/stylesheets/components/_progress.scss
```diff
@@ -0,0 +1,84 @@
...
+
+  &.is-active {
+    &::before {
+      background-color: $white;
+      border: 2px solid $govuk-blue;
+    }
+  }
+
+  &:last-child {
+    border-width: 0;
```
teneightfive: Should probably be `border-color: transparent;` instead of removing the border otherwise this element sits 2px further left than the others.

assets/stylesheets/components/_progress.scss
```diff
@@ -0,0 +1,84 @@
...
+  position: relative;
+  display: block;
+  padding: $default-spacing-unit;
+  top: -33px;
+
+  @include media(tablet) {
+    position: absolute;
+    left: -50px;
+    top: 0;
+    width: 100px;
```
teneightfive: This is fine for now as we only have the one case but this feels a bit restrictive and might need adapting if we use it elsewhere.

src/apps/components/views/progress.njk
```diff
@@ -0,0 +1,12 @@
+{% extends "./_layout.njk" %}
+
+{% block body_main_content %}
+  <h2 class="page-heading">Progress Indicator</h2>
```
teneightfive: Are we calling it `Progress` or `Progress indicator`? Which ever we use we should be consistent everywhere we reference it.

teneightfive: Ah, just looked at this page and realised this is a second level heading under `Progress`. Suppose that makes sense but just a bit strange that we're using 2 terms for it. All the files and macros are just called Progress but we call the item Progress indicator here.

tyom: Should use `{% call Example('Progress') %}` to be consistent with other components pages. It also creates visual boundaries for the component.

assets/stylesheets/components/_progress.scss
feedmypixel: So with using `absolute` here instead of negative margins it means I will have to apply different `top` and `left` values in `> tablet` view to `small` screen view.

>positive: relative and top: -33px on .c-progress__stage-title?

This is to pull the stage text up to line up with the stage circles in `small` screen view

assets/stylesheets/components/_progress.scss
feedmypixel: > This is fine for now as we only have the one case but this feels a bit restrictive and might need adapting if we use it elsewhere.

agreed, will potentially need work with longer text

tyom:
I feel like the vertical spacing between points should be reduced on smaller screens.

src/templates/_macros/common.njk
```diff
@@ -115,3 +115,31 @@
...
   {% endif %}
 {% endmacro %}
+
+
+{##
+ # Render progress bar
+ # @param {object} props
+ # @param {array} props.stageNames - stage names
+ # @param {string} props.currentStageName - the current stages name
+ # @param {string} props.currentStageIndex - the current stages index
```
tyom: You could avoid this extra prop by finding the index from `props.currentStageName`.
```njk
{% set currentStageIndex = props.stageNames.indexOf(props.currentStageName) %}
```

ztolley:
Btw I looked at a table version... er.. not easy

assets/stylesheets/components/_progress.scss
```diff
@@ -0,0 +1,84 @@
...
+}
+
+.c-progress__stage-title {
+  position: relative;
+  display: block;
+  padding: $default-spacing-unit;
+  top: -33px;
+
+  @include media(tablet) {
+    position: absolute;
```
tyom: I think you could just keep it `relative` which would work the same with addition of expanding container accommodating longer names.

assets/stylesheets/components/_progress.scss
```diff
@@ -0,0 +1,84 @@
...
+    display: table-cell;
+    width: $_section-width;
+    border-width: 2px 0 0;
+  }
+
+  &::before {
+    position: relative;
+    width: 14px;
+    height: 14px;
+    margin: -7px -7px 0;
```
teneightfive: It looks like `-8px` would be better for the left adjust:

## -7px
image

## -8px

image

tyom: Accounting for 2px border width.

assets/stylesheets/components/_progress.scss
teneightfive: >This is to pull the stage text up to line up with the stage circles in `small` screen view

Yep, but if you remove the `::before` from the document flow you won't need to mess around with random numbers like `33` as much.

Could change `.c-progress__stage-title` to just be:

```scss
.c-progress__stage-title {
  display: block;
  padding: 0 $default-spacing-unit $default-spacing-unit;
  line-height: 1em;

  @include media(tablet) {
    padding: $default-spacing-unit 0 0;
  }
}
```

and `.c-progress__stage::before` to dropping the need for margins:

```scss
.c-progress__stage::before {
  ...
  position: absolute;
  left: -8px;
  ...

  @include media(tablet) {
    top: -8px;
  }
}
```

feedmypixel:
> Btw I looked at a table version... er.. not easy

@ztolley it `display: table` now

assets/stylesheets/components/_progress.scss
feedmypixel: 👍  on closer inspection you sir are correct 😁  Good 👀

ztolley:
BIG thumbs up

feedmypixel:
> BIG thumbs up

kudos to the team for the PR comments and help also 💥

assets/stylesheets/components/_progress.scss
feedmypixel: screen shot 2017-08-10 at 14 39 01

assets/stylesheets/components/_progress.scss
feedmypixel: so this needs to be `absolute` to remove the last `li` element from being used as part of the 100% width of `c-progress-bar`

assets/stylesheets/components/_progress.scss
feedmypixel: There are some issues with need to take the last `li` out of the flow hence using `absolute`. Can I suggest that you have a play with this and raise a PR after this is in? As it would be easier for you to play around it to see the various issues requiring the current CSS use.

assets/stylesheets/components/_progress.scss
tyom: Could you do something with the last element then?

PR #424, opened 2017-08-07
updates to investment status component

feedmypixel:
Updates to the markup and styles in the investment status component

screen shot 2017-08-07 at 20 39 54


tyom:
I wonder if we can use `meta-item` for this. We use it in investment project list items.

feedmypixel:
> I wonder if we can use meta-item for this. We use it in investment project list items.

I'd be tempted to say lets do that as part of another PR?

feedmypixel:
I have updated this component to use the `c-meta-item` styles as suggested by @tyom 

screen shot 2017-08-09 at 12 23 59

src/templates/_macros/investment-meta.njk
```diff
@@ -0,0 +1,20 @@
+{##
+ # Render Meta fields for Investment
+ # @param {object[]} props.meta - Array of meta objects
+ #}
+{% macro InvestmentMeta (props) %}
+  {% if props.meta | length %}
+    <dl>
+      {% for meta in props.meta %}
+        {% if meta.value %}
+          <div class="c-meta-item">
```
tyom: You can't have `div` as sibling of `dl`.

teneightfive: s/sibling/child

feedmypixel: > You can't have div as sibling of dl.

According to [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl) its permitted content

tyom: It says there:
> WHATWG HTML allows wrapping each name-value group in a `<dl>` element in a `<div>` element. 

As far as I can tell [there's still no consensus](https://github.com/whatwg/html/issues/1937) on this issue.

I wonder if the macro should deal with loops at all. The `isLabelHidden` is also a global setting and there seems to be no way of using that for a specific item.

tyom: What if there was a macro that creates a single key/value pair? And then maybe It could be used in a loop using `map` filter.

assets/stylesheets/components/_meta-item.scss
```diff
@@ -6,16 +6,20 @@
   + &:nth-child(even) {
     margin-left: $baseline-grid-unit;
   }
+
+  & + & {
+    margin-left: $default-spacing-unit;
```
teneightfive: This seems to affect it's current use on investment projects collection:

image

assets/stylesheets/components/_meta-item.scss
```diff
@@ -6,16 +6,20 @@
...
 .c-meta-item__label {
-  color: $grey-1;
   min-width: 4em;
   display: inline-block;
+  color: $text-colour;
 }
 
 .c-meta-item__value {
-  color: $text-colour;
+  color: $grey-1;
```
teneightfive: This also changes how it previously looked on investment projects:

### Before

image


### After

image

src/templates/_macros/investment-meta.njk
```diff
@@ -0,0 +1,20 @@
+{##
+ # Render Meta fields for Investment
+ # @param {object[]} props.meta - Array of meta objects
+ #}
+{% macro InvestmentMeta (props) %}
```
teneightfive: This name feels very heavily tied to its content but I wonder if it needs to. The macro is just rendering a list of meta-item and it doesn't need to care if they're in investment or not.

This macro could be `MetaData` and added to `_macros/common.njk` (as much as I hate its generic name)?

feedmypixel:
@tyom @teneightfive updates up for this
screen shot 2017-08-09 at 16 45 51

assets/stylesheets/components/_meta-bar.scss
```diff
@@ -0,0 +1,9 @@
+@import "../settings";
+
+.c-meta-bar {
+  margin-top: $default-spacing-unit;
```
tyom: This should be set in assemblies.

src/apps/investment-projects/views/_layout.njk
```diff
@@ -1,4 +1,5 @@
 {% extends "_layouts/two-column.njk" %}
+{% from "_macros/entities.njk" import MetaItem %}
```
tyom: Is this used here?

src/apps/investment-projects/views/_layout.njk
```diff
@@ -10,9 +11,21 @@
     heading: investmentData.name,
     modifier: 'light-banner'
   }) %}
-    {% component 'investment-status', investmentStatus %}
 
-    {% if investmentData.archived %}
+  {{ Meta({
```
tyom: You could also do this without creating a new macro:
```
{{
  [
    {
      label: 'Project code',
      value: investmentStatus.projectCode
    },
    {
      label: 'Valuation',
      value: investmentStatus.valuation
    }
  ] | map(callAsMacro('MetaItem')) | join | safe
}}
```
Or set the data it in controller and just use one liner.

src/templates/_macros/common.njk
```diff
@@ -115,3 +115,24 @@
...
   {% endif %}
 {% endmacro %}
+
+{##
+ # Render Meta fields
+ # @param {object[]} props.meta - Array of meta objects
+ #}
+{% macro Meta(props) %}
+  {% if props.meta | length %}
+    <dl class="c-meta-bar">
```
tyom: This component seems unique to one use case so far. Is it really a component? Could we avoid creating new macros for a single use case?

feedmypixel: > This component seems unique to one use case so far. Is it really a component? Could we avoid creating new macros for a single use case?

will do - ill add it in as described above

src/apps/investment-projects/views/_layout.njk
feedmypixel: no - added here so its globally available

src/apps/investment-projects/views/_layout.njk
feedmypixel: Over cornflakes I have been pondering this... The reason I didn't go with the `MetaItem` macro is because it is just using `div`s and it felt like there was a need for a `dl` here. Hence the creation of `Meta` component.

src/apps/investment-projects/views/_layout.njk
teneightfive: Maybe just import in `datahub-layout.njk` like all the other Macros. Then we don't have random imports scattered around. Also allows us to see which macros are available to the views at a glance.

assets/stylesheets/components/_meta-list.scss
```diff
@@ -0,0 +1,9 @@
+@import "../settings";
+
+.c-meta-list {
+  margin-top: $default-spacing-unit;
```
teneightfive: What if it appears as the first element? Could use a sibling selector to tell it only to apply a top margin if it has a preceding element:

```scss
.c-meta-list {
  * + & {
  }
}
```

assets/stylesheets/components/_meta-list.scss
```diff
@@ -0,0 +1,9 @@
...
+
+  .c-meta-item {
+    margin-right: $default-spacing-unit;
```
teneightfive: Maybe use a sibling selector to set a left margin:

```scss
.c-meta-item {
  * + & {
  }
}
```

src/apps/investment-projects/views/_layout.njk
```diff
@@ -10,9 +11,21 @@
     heading: investmentData.name,
     modifier: 'light-banner'
   }) %}
-    {% component 'investment-status', investmentStatus %}
 
-    {% if investmentData.archived %}
+  {{ MetaList({
+    meta: [
```
teneightfive: `items`?

src/templates/_macros/common.njk
```diff
@@ -115,3 +115,24 @@
...
+{##
+ # Render Meta fields in a list
+ # @param {object[]} props.meta - Array of meta objects
+ #}
+{% macro MetaList(props) %}
+  {% if props.meta | length %}
+    <dl class="c-meta-list">
+      {% for meta in props.meta %}
+        {% if meta.value %}
+          <div class="c-meta-item">
```
teneightfive: Should we have a `<div>` as a child element of a `<dl>`?

src/apps/investment-projects/views/_layout.njk
tyom: We could make it configurable like we do with other macros. e.g. pass `el` prop into macro.

src/templates/_macros/common.njk
tyom: I wonder if we should try to fit it into HTML semantics at this stage. We could use use non-semantic elements for now and refactor once we have a better idea how these are used across the apps (in groups or as individual elements).

src/apps/investment-projects/views/_layout.njk
feedmypixel: ah my bad! this is meant to be in with all the [others](https://github.com/uktrade/data-hub-frontend/blob/develop/src/templates/_layouts/datahub-base.njk#L1-L16)

src/templates/_macros/common.njk
feedmypixel: Agreed 👍  we can align later when we know what this should be doing

PR #425, opened 2017-08-07
Date Fieldset component

feedmypixel:
Add date component pattern

### Date component
screen shot 2017-08-07 at 20 36 24

### DIFF
--- a/assets/stylesheets/components/form/_form-group.scss
+++ b/assets/stylesheets/components/form/_form-group.scss
@@ -4,7 +4,8 @@
   max-width: 40em;
 
   [type="hidden"] + &,
-  .c-form-fieldset__legend + & {
+  .c-form-fieldset__legend + &,
+  .c-form-fieldset__hint + & {
     margin-top: 0;
   }
 }
@@ -67,7 +68,6 @@
 
     .c-form-control,
     .c-multiple-choice {
-      width: auto;
       display: inline-block;
       margin-top: 0;
     }
@@ -93,6 +93,12 @@
   .c-multiple-choice__input:checked + .c-multiple-choice__label::before,
   .c-multiple-choice__input:focus + .c-multiple-choice__label::before {
     border-color: $black;
+  }
+}
+
+.c-form-group--soft {
+  .c-form-group__label-text {
+    font-weight: normal;
   }
 }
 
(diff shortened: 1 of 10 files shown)


tyom:
Should it also have a field for date? Maybe make it configurable?

feedmypixel:
> Should it also have a field for date? Maybe make it configurable?

I decided to just have month and year initially as this is all we are using it for. I think the way this is used for this question needs looking at anyway. Which could be future work. Do we need to actually ask for a date? or maybe how many months/years until this is expected to land?

src/apps/investment-projects/services/formatting.js
```diff
@@ -80,19 +84,24 @@ function transformFromApi (body) {
...
     }
     return get(body, `${key}.id`)
   })
 
   const date = new Date(body['estimated_land_date'])
   if (date) {
-    formatted['land-date_year'] = date.getFullYear()
-    formatted['land-date_month'] = date.getMonth() + 1 // month is zero based index
+    formatted['estimated_land_date_year'] = date.getFullYear()
+    formatted['estimated_land_date_month'] = padStart(date.getMonth() + 1, 2, 0) // month is zero based index
```
tyom: We have `date-fns` module. Might as well use it.
```js
const format = require('date-fns/format')
format(date, 'MM') // 08
```

assets/stylesheets/components/form/_form-group.scss
```diff
@@ -96,6 +96,12 @@
   }
 }
 
+.c-form-group--soft {
+  .c-form-group__label-text {
+    font-weight: normal;
```
teneightfive: Might be better to do this in `.c-form-group__label-text` instead. Ties the change to the element being affected rather than having multiple places where styles are applied for that element.

feedmypixel: The only issue with that is when using the `FormGroup` macro (via `TextField` in `DateFieldset`) I would have to build custom functionality to add a specific class to the ]label element](https://github.com/uktrade/data-hub-frontend/blob/develop/src/templates/_macros/form.njk#L217) which felt too specific. Where as this way just uses the [modifier](https://github.com/uktrade/data-hub-frontend/blob/develop/src/templates/_macros/form.njk#L199) that is already available.

PR #428, opened 2017-08-08
Ability to set investment value to being not provideable

tyom:
For total investment and foreign equity investment values.
Remove checkboxes that were essentially supposed to be doing that.

DH-390

### Shown in view when client can't provide the information
image

### DIFF
--- a/src/apps/investment-projects/views/value-edit.njk
+++ b/src/apps/investment-projects/views/value-edit.njk
@@ -4,140 +4,155 @@
 {% block main_grid_right_column %}
   <h2 class="heading-medium">Edit value</h2>
 
-  {{ errorPanel(form.errors) }}
+  {% call Form({
+    buttonText: 'Save',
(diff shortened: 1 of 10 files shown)


src/apps/investment-projects/views/value-edit.njk
```diff
@@ -4,140 +4,155 @@
...
+      error: form.errors.messages.number_new_jobs
+    }) }}
+
+    {{ MultipleChoiceField({
+      name: 'average_salary',
+      label: form.labels.average_salary,
+      value: form.state.average_salary,
+      error: form.errors.messages.average_salary,
+      options: form.options.averageSalaryRange,
+      initialOption: '-- Choose an average salary --'
```
teneightfive: Change this to `type: 'radio'` then we don't need an initial option.

tyom: Updated

PR #430, opened 2017-08-09
Refactor investment filter forms to use a simple form builder

tyom:
- Create a new builder to make filters form from a JS configuration with combination of labels file
- Add new macro which expects the macro configuration build by the builder above
- Add tests for builders

### DIFF
--- a/config/nunjucks/filters.js
+++ b/config/nunjucks/filters.js
@@ -8,6 +8,7 @@ const {
   isEmpty,
   isNull,
   isString,
+  isFunction,
   pickBy,
   filter,
   isNil,
@@ -101,6 +102,23 @@ const filters = {
       })
     }
     return result
+  },
+
+  applyMacro (config) {
+    function renderMacro (macroConfig) {
+      return map(macroConfig, (props, name) => {
+        const macro = this.ctx[name]
+        if (isFunction(macro)) {
+          return macro(props)
+        }
+      })[0]
+    }
+
+    if (isArray(config)) {
+      return config.map(renderMacro.bind(this))
+    }
+
+    return renderMacro.call(this, config)
   },
 }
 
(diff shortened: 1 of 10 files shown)


src/apps/investment-projects/builders.js
```diff
@@ -66,7 +67,23 @@ function buildInvestmentSorting (originalQuery = {}) {
   }
 }
 
+function buildMacroMacroConfigFromFormFields (filtersQuery = {}) {
```
teneightfive: `MacroMacro`?

teneightfive: `FromForm`?

src/apps/investment-projects/builders.js
```diff
@@ -66,7 +67,23 @@ function buildInvestmentSorting (originalQuery = {}) {
...
+  if (!filterFields) { return null }
```
teneightfive: Where is `filterFields ` coming from?

config/nunjucks/filters.js
```diff
@@ -100,6 +101,23 @@ const filters = {
     }
     return result
   },
+
+  applyMacro (config) {
```
teneightfive: Could there be a way to make use of [callAsMacro](https://github.com/uktrade/data-hub-frontend/blob/67fa79ad1c0326274afd64166c5c331b4cac2ada/config/nunjucks/globals.js#L10-L16) from globals instead?

teneightfive: Could it be a function in global to keep these kind of actions in the same place? Might be easier for someone to see that there are 2 ways of calling macros using a variable if they're in the same place.

tyom: Moved it to globals.

PR #433, opened 2017-08-09
Avoid changing window location when not needed

tyom:
The module was changing window location every time any part of it was updated which caused page refresh on fragment update (e.g. focus on error field).


assets/javascripts/lib/xhr.js
```diff
@@ -8,7 +8,7 @@ history.listen((location, action) => {
   if (action === 'POP') {
     if (location.state) {
       XHR.injectResponseInHtml(location.state.data)
-    } else {
+    } else if (window.location.pathname !== location.pathname && window.location.search !== location.search) {
```
feedmypixel: 🤙

PR #436, opened 2017-08-09
Ability to add multiple ITAs to an OMIS order during creation

teneightfive:
This adds the ability to add multiple ITAs to an OMIS order.

It's tries to create a generic _add another_ functionality for the form wizard that could be replicated elsewhere.

👉  [DEMO](https://datahub-beta2-pr-436.herokuapp.com/omis/create/dcdabbc9-1781-e411-8955-e4115bead28a) 👈 

## What it looks like

add-another

### DIFF
--- a/src/apps/omis/apps/create/controllers/assign-ita.js
+++ b/src/apps/omis/apps/create/controllers/assign-ita.js
@@ -1,4 +1,4 @@
-const { sortBy } = require('lodash')
+const { filter, flatten, forEach, sortBy } = require('lodash')
 
 const { FormController } = require('../../../controllers')
 const { getAdvisers } = require('../../../../adviser/repos')
@@ -11,6 +11,44 @@ class AssignItaController extends FormController {
 
     req.form.options.fields.ita.options = sortBy(options, 'label')
     super.configure(req, res, next)
+  }
+
+  process (req, res, next) {
+    const addKey = req.body['add-item']
+    const removeKey = req.body['remove-item']
+
+    if (addKey || removeKey) {
+      if (addKey) {
+        req.form.values[addKey] = flatten([req.form.values[addKey]])
+        // add empty value to array to show add a new add adviser input
+        req.form.values[addKey].push('')
+      }
+
+      if (removeKey) {
+        const [ key, index ] = removeKey.split('::')
+        req.form.values[key].splice(index, 1)
+      }
+
+      return this.saveValues(req, res, () => {
+        res.redirect(req.baseUrl + req.path)
+      })
+    }
+
+    next()
+  }
+
+  saveValues (req, res, next) {
+    if (!req.body['add-item'] && !req.body['remove-item']) {
+      const fields = req.form.options.fields
+
+      forEach(fields, (options, key) => {
+        if (options.repeatable) {
+          req.form.values[key] = filter(flatten([req.form.values[key]]))
+        }
+      })
+    }
+
+    super.saveValues(req, res, next)
   }
 }
 
--- a/src/apps/omis/apps/create/controllers/confirm.js
+++ b/src/apps/omis/apps/create/controllers/confirm.js
@@ -1,6 +1,7 @@
-const { find, unset } = require('lodash')
+const { filter, find, get, unset } = require('lodash')
 
 const metadataRepo = require('../../../../../lib/metadata')
+const { transformIdToObject } = require('../../../../transformers')
 const { FormController } = require('../../../controllers')
 const { getAdvisers } = require('../../../../adviser/repos')
 const { Order } = require('../../../models')
@@ -16,7 +17,10 @@ class ConfirmController extends FormController {
       values.contact = `${contact.first_name} ${contact.last_name}`
       values.company = company
       values.primary_market = find(metadataRepo.countryOptions, { id: values.primary_market })
-      values.ita = find(advisers.results, { id: values.ita })
+      values.ita = filter(values.ita).map((id) => {
+        const adviser = find(advisers.results, { id })
+        return get(adviser, 'name')
+      })
 
       next(err, values)
     })
@@ -24,6 +28,7 @@ class ConfirmController extends FormController {
 
   async successHandler (req, res, next) {
     const data = req.sessionModel.toJSON()
+    const subscribers = data.ita.map(transformIdToObject)
 
     // clean un-needed properties
     unset(data, 'csrf-secret')
@@ -31,6 +36,8 @@ class ConfirmController extends FormController {
 
     try {
       const order = await Order.save(req.session.token, data)
+
+      await Order.saveSubscribers(req.session.token, order.id, subscribers)
 
       req.journeyModel.reset()
       req.journeyModel.destroy()
--- a/src/templates/_layouts/form-wizard-step.njk
+++ b/src/templates/_layouts/form-wizard-step.njk
@@ -15,15 +15,54 @@
     })) %}
 
       {% block fields %}
-        {% for key, props in options.fields %}
+        {% for key, fieldOptions in options.fields %}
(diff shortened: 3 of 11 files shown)


src/apps/omis/apps/create/controllers/assign-ita.js
```diff
@@ -12,6 +12,67 @@ class AssignItaController extends FormController {
...
     super.configure(req, res, next)
   }
+
+  process (req, res, next) {
+    const addKey = req.body['add-item']
+    const removeKey = req.body['remove-item']
+
+    if (addKey || removeKey) {
+      if (addKey) {
+        if (!isArray(req.form.values[addKey])) {
```
feedmypixel: Could just use [flatten](https://lodash.com/docs/4.17.4#flatten)
```
...
if (addKey) {
  req.form.values[addKey] = flatten([req.form.values[addKey]])
  req.form.values[addKey].push('') 
}
...
```

src/apps/omis/apps/create/controllers/assign-ita.js
```diff
@@ -12,6 +12,67 @@ class AssignItaController extends FormController {
...
+          req.form.values[addKey] = [req.form.values[addKey]]
+        }
+        req.form.values[addKey].push('')
```
feedmypixel: feels like this needs a comment, maybe:
`// add empty value to array to show add a new add adviser input`

src/apps/omis/apps/create/controllers/assign-ita.js
```diff
@@ -12,6 +12,67 @@ class AssignItaController extends FormController {
...
+      }
+
+      if (removeKey) {
```
feedmypixel: could do:
```
if (removeKey) {
  const [ key, index ] = removeKey.split('::')
  req.form.values[key].splice(index, 1)
}
```

src/apps/omis/apps/create/controllers/assign-ita.js
```diff
@@ -12,6 +12,67 @@ class AssignItaController extends FormController {
...
+      const fields = req.form.options.fields
+
+      forEach(fields, (fieldOptions, key) => {
+        const fieldValue = values[key]
+
+        if (fieldOptions.repeatable) {
+          if (!isArray(fieldValue)) {
+            values[key] = [fieldValue]
+          } else if (!fieldValue.length) {
+            values[key] = ['']
```
feedmypixel: feels like this could do with a comment `// add empty value to array to show add a new add adviser input when no advisers`

src/apps/omis/apps/create/controllers/assign-ita.js
```diff
@@ -12,6 +12,67 @@ class AssignItaController extends FormController {
...
+  getValues (req, res, next) {
+    super.getValues(req, res, (err, values) => {
+      const fields = req.form.options.fields
+
+      forEach(fields, (fieldOptions, key) => {
+        const fieldValue = values[key]
+
+        if (fieldOptions.repeatable) {
+          if (!isArray(fieldValue)) {
+            values[key] = [fieldValue]
```
feedmypixel: could maybe also use [flatten](https://lodash.com/docs/4.17.4#flatten) here 
```
values[key] = flatten([fieldValue.length ? fieldValue : ''])
```

src/templates/_layouts/form-wizard-step.njk
```diff
@@ -15,15 +15,54 @@
...
           {% if props.fieldType %}
-            {{ callAsMacro(props.fieldType)(props | assign({
-              name: key,
-              label: trans(props.label),
-              hint: trans(props.hint),
-              value: values[key],
-              error: errors[key].message
-            })) }}
+            {% if props.repeatable %}
+              {% set fieldValues = values[key] or [''] %}
```
feedmypixel: potentially this adding of an empty array element needs to be done in one place and explained why its being done

src/templates/_layouts/form-wizard-step.njk
```diff
@@ -15,15 +15,54 @@
     })) %}
 
       {% block fields %}
-        {% for key, props in options.fields %}
+        {% for key, fieldOptions in options.fields %}
+          {% set props = {} | assign(fieldOptions, {
+            name: key,
+            label: trans(fieldOptions.label),
```
feedmypixel: Not part of this work but can we call this `translate` - the shortened version `trans` IMO looses a lot of meaning when compared to `translate`?

teneightfive: Yep fair point, will pick this up as a separate change after as out of scope for this I feel.

src/templates/_layouts/form-wizard-step.njk
teneightfive: Yes, good point.

I may try and rework this to actually allow the configuration of whether it should start with an empty one or not. In this case of it being optional it may actually be easier to have none to start with and if the user wants to add some they can.

src/templates/_layouts/form-wizard-step.njk
teneightfive: Added #450 to change this. If we merge that one first I'll update this branch to use the new full method name.

src/templates/_layouts/form-wizard-step.njk
feedmypixel: If you will always need 1 adviser then you may be able to make your life easier and just always have the ability to add 1 or another one rather than adding configuration to it

teneightfive: That's the problem. OMIS has both cases. During the creation, you don't have to add any _ITAs_ (or what we're calling subscribers). 

Then, during the edit state and whilst you're completing the rest of the order you will need to add advisers from post where you will need at least one.

So in the first case it might be better to not have one there by default but the second case it would make sense to already have one empty on first load.

teneightfive:
>Unless I am missing something it feels like this extra input can just be "hardcoded" and added in the view layer and always be there, then this logic to add it in could go potentially?

@feedmypixel you were right, I've removed the logic from the controller which hopefully simplifies that.

I think I will leave this PR with the single state of there always being an empty one by default for now and if we need to configure that option in the future it will be better to do as a separate change.

src/apps/omis/apps/create/controllers/confirm.js
```diff
@@ -16,7 +16,9 @@ class ConfirmController extends FormController {
       values.contact = `${contact.first_name} ${contact.last_name}`
       values.company = company
       values.primary_market = find(metadataRepo.countryOptions, { id: values.primary_market })
-      values.ita = find(advisers.results, { id: values.ita })
+      values.ita = values.ita.map((ita) => {
+        return find(advisers.results, { id: ita }).name
+      })
```
tyom: Do you need to remove `undefined`s afterwards?

src/apps/omis/apps/create/controllers/confirm.js
```diff
@@ -32,6 +34,10 @@ class ConfirmController extends FormController {
     try {
       const order = await Order.save(req.session.token, data)
 
+      await Order.saveSubscribers(req.session.token, order.id, data.ita.map((id) => {
```
tyom: Could you extract third argument into a variable to make it clear that you're passing  request body?

src/apps/omis/apps/create/controllers/confirm.js
teneightfive: Empty and `undefined`s are removed during the [`saveValues`](https://github.com/uktrade/data-hub-frontend/pull/436/files#diff-0df20b7f65681a461e68876f362df584R40) call.

teneightfive: I have just made this slightly more robust though so it doesn't fall over.

PR #437, opened 2017-08-09
Contacts collection

tyom:
Initial implementation of contacts index page with sorting but no filters.

### TODO
- [x] Tests

localhost-3001-contacts-sortby created_on 3adesc


src/apps/builders.js
```diff
@@ -0,0 +1,20 @@
+const get = require('lodash/get')
```
teneightfive: `const { get } = require('lodash')` for consistency?

PR #438, opened 2017-08-10
[DH-419] Fix error displaying company interactions page for some companies

reupen:
Fixes the 500 error on the company interactions page when there are interactions with a null interaction_type.

### DIFF
--- a/src/apps/interactions/services/formatting.js
+++ b/src/apps/interactions/services/formatting.js
@@ -36,12 +36,13 @@ function getDisplayInteraction (interaction) {
  * @returns {Object} A formatted service delivery or interaction
  */
 function getDisplayCompanyInteraction (interaction) {
-  const type = (interaction.interaction_type.name === 'Service delivery') ? 'service-deliveries' : 'interactions'
+  const interactionType = get(interaction, 'interaction_type.name')
+  const entityType = (interactionType === 'Service delivery') ? 'service-deliveries' : 'interactions'
 
   const result = {
     id: interaction.id,
-    url: `/${type}/${interaction.id}`,
-    interaction_type: interaction.interaction_type.name,
+    url: `/${entityType}/${interaction.id}`,
+    interaction_type: interactionType,
     subject: interaction.subject,
     date: formatMediumDate(interaction.date),
     adviser: getPropertyName(interaction, 'dit_adviser'),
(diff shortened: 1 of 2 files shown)


src/apps/interactions/services/formatting.js
```diff
@@ -36,12 +36,13 @@ function getDisplayInteraction (interaction) {
  * @returns {Object} A formatted service delivery or interaction
  */
 function getDisplayCompanyInteraction (interaction) {
-  const type = (interaction.interaction_type.name === 'Service delivery') ? 'service-deliveries' : 'interactions'
+  const interactionType = getPropertyName(interaction, 'interaction_type')
```
tyom: Should just use Lodash's `get` here.
```js
get(interaction, 'interaction_type.name')
```
I don't see why we need a library function to get a `.name` from a property.

test/unit/apps/interactions/services/formatting.service.test.js
```diff
@@ -41,6 +41,23 @@ describe('Interaction formatting service', function () {
...
+        subject: 'Subject 1234',
+        date: '14 Feb 2017',
+        adviser: 'John Brown',
+        contact: '<a href="/contacts/444">Fred Smith</a>',
+        service: 'service name',
+        notes: 'Here are some notes<br/>line 2.',
+        dit_team: 'team name',
+      }
+      const actualDisplayInteraction = interactionFormattingService.getDisplayCompanyInteraction(interaction)
+      expect(actualDisplayInteraction).to.deep.equal(expectedDisplayInteraction)
```
tyom: Could do with some white space here in this block.

reupen: I added a blank line, but it was same as all the other tests in this file.

src/apps/interactions/services/formatting.js
reupen: That needs looking at across the app. Changing it here breaks tests as one returns undefined when a property is missing while the other returns null. Some of the tests have data in JSON files, and you can't put undefined in JSON.

tyom: You can return null if unset:
```js
get(interaction, 'interaction_type.name', null)
```

reupen: I know that, but that introduces more inconsistency. Looks like half the time in the code base `_.get()` is used used without a third argument, and the other half of the time with null as the default value. What is the correct approach needs to be decided separately and applied across the whole code base.

teneightfive: @reupen I think the instances where it uses the third argument set to `null` is because of what the API accepts. Other cases where it has no third argument it will default to `undefined` and that it often enough for us to check against that value and do something if it exists.

So in short, use the third argument set to `null` if needing to send `null` values to the API, otherwise try and handle with `undefined`.

PR #440, opened 2017-08-10
Fix focus outline with highlight utility

teneightfive:
When the highlight element was given `position: relative` it was causing the box-shadow outline on elements that have a focus state to jump out of position.

## Before

image

## After

image


tyom:
Good spot!

PR #441, opened 2017-08-10
Form macro/element updates

teneightfive:
This PR extracts a few updates that will be needed for the _add another_ functionality being added to the OMIS form wizard. In summary:

- Tweak link modifier on `button` component
- Add a [compact style](https://datahub-beta2-pr-441.herokuapp.com/components/form#Compact%20version) as a modifier to `.form-group`
- Support for an [action to be attached](https://datahub-beta2-pr-441.herokuapp.com/components/form#Fields%20with%20actions) to a `.form-group`

### DIFF
--- a/assets/stylesheets/_deprecated/components/_button.scss
+++ b/assets/stylesheets/_deprecated/components/_button.scss
@@ -22,3 +22,26 @@
   clear: both;
   margin: 30px 0;
 }
+
+.button--link {
+  background: transparent;
+  border: 0;
+  box-shadow: none;
+  color: $link-colour;
+  cursor: pointer;
+  padding: .4em 0;
+  text-align: left;
+  text-decoration: underline;
+
+  &:hover,
+  &:active,
+  &:focus {
+    background: transparent;
+    color: $link-colour;
+    top: 0;
+  }
+
+  &:hover {
+    color: $link-hover-colour;
+  }
+}
--- a/src/apps/components/views/form.njk
+++ b/src/apps/components/views/form.njk
@@ -313,4 +313,48 @@
     {% endcall %}
   {% endcall %}
 
+  {% call Example('Fields with actions') %}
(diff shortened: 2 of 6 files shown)


assets/stylesheets/_deprecated/components/_button.scss
```diff
@@ -22,3 +22,26 @@
...
 }
+
+.button--link {
+  background: transparent;
+  border: 0;
+  box-shadow: none;
+  color: $link-colour;
+  cursor: pointer;
+  padding: .4em 0;
+  text-align: left;
```
tyom: Is this necessary? The button is auto-width anyway.

src/apps/components/views/form.njk
```diff
@@ -278,4 +278,83 @@
...
+          error: form.errors.messages.averageSalaryLight,
+          selected: form.state.averageSalaryLight,
+          modifier: 'compact'
+      }) }}
+    {% endcall %}
+  {% endcall %}
+
+  {% call Example('Fields with actions') %}
+    {% set removeButton %}
+      <button class="button button--link c-form-group__action">Remove</button>
```
tyom: Keep in mind that some older IE browsers have issues with submitting `button` elements as they are all submitted rather than the one that is pressed. `input type="submit"` with `name` and `value` attributes is more robust in that case. I presume this style works well with inputs?

assets/stylesheets/_deprecated/components/_button.scss
teneightfive: It's actually for smaller screens. The button by default goes full width and aligns text in the middle but this looks odd for text variations that don't have the background container.

src/apps/components/views/form.njk
teneightfive: Yes, very good point.

PR #443, opened 2017-08-11
Minor tweaks to entity/results macros

tyom:
Generic tweaks for use across multiple entities.

### DIFF
--- a/src/apps/investment-projects/transformers.js
+++ b/src/apps/investment-projects/transformers.js
@@ -34,20 +34,20 @@ function transformInvestmentProjectToListItem ({
         name: 'investor_company',
         label: 'Investor',
         value: investor_company,
-        isBadge: false,
       }, {
         name: 'estimated_land_date',
         label: 'Estimated to land',
         value: estimated_land_date,
         isInert: true,
-        isBadge: false,
       }, {
         name: 'sector',
         label: 'Sector',
         value: sector,
-        isBadge: false,
       },
-    ],
+    ].map(item => {
+      item.isBadge = !!item.isBadge
+      return item
+    }),
   }
 }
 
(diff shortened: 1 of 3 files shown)


src/apps/investment-projects/transformers.js
```diff
@@ -34,20 +34,20 @@ function transformInvestmentProjectToListItem ({
...
-        isBadge: false,
       }, {
         name: 'sector',
         label: 'Sector',
         value: sector,
-        isBadge: false,
       },
-    ],
+    ].map(item => {
+      item.isBadge = !!item.isBadge
```
feedmypixel: I've become a fan of using `Boolean(item.isBadge)` - thoughts?

src/templates/_macros/entities.njk
```diff
@@ -35,7 +35,7 @@
       {% if metaItems | length %}
         <div class="c-entity__content">
           {% for metaItem in metaItems %}
-            {{ MetaItem(metaItem| assign({ highlightTerm: props.highlightTerm })) }}
+            {{ MetaItem(metaItem | assign({ highlightTerm: props.highlightTerm, isLabelHidden: metaItem.isLabelHidden })) }}
```
feedmypixel: easier to read on a newline?

src/apps/investment-projects/transformers.js
tyom: Apparently there's a significant difference in performance in some browsers. While Chrome appears to perform slightly better with `Boolean`, Safari, for example, is much worse.

feedmypixel: I'd air more on the clarity side of things, but not hung up about it 👍

PR #445, opened 2017-08-11
Use Sentry for error reporting in Node.js layer

reupen:
DHE-549

Adds Sentry to Express to report unhandled errors, along the lines of:
https://docs.sentry.io/clients/node/integrations/express/

Technically:
* `if (config.sentryDsn)` isn't required as Raven won't do anything if a DSN isn't configured
* Adding sentryDsn to config/index.js isn't needed as Raven will use the SENTRY_DSN environment variable automatically

Reporting JS errors in the browser should also be looked at, but that can be done separately and isn't as important right now.

### DIFF
--- a/src/server.js
+++ b/src/server.js
@@ -4,6 +4,7 @@ const bodyParser = require('body-parser')
 const compression = require('compression')
 const config = require('../config')
 const express = require('express')
+const raven = require('raven')
 const flash = require('connect-flash')
 const csrf = require('csurf')
 const slashify = require('slashify')
@@ -32,6 +33,19 @@ const routers = require('./apps/routers')
 const app = express()
 
 app.disable('x-powered-by')
+
+if (config.sentryDsn) {
+  logger.info('Sentry DSN detected. Raven will be enabled.')
+  // See https://docs.sentry.io/clients/node/config/
+  // and https://docs.sentry.io/clients/node/usage/
+  // for info on Raven config options
+  raven.config(config.sentryDsn, {
+    autoBreadcrumbs: true,
+    captureUnhandledRejections: true,
+  }).install()
+  // Raven request handler must be the first middleware
+  app.use(raven.requestHandler())
+}
 
 if (!config.ci) {
   app.use(churchill(logger))
@@ -79,6 +93,11 @@ app.use(store())
 // routing
 app.use(slashify())
 app.use(routers)
+
+// Raven error handler must come before other error middleware
+if (config.sentryDsn) {
+  app.use(raven.errorHandler())
+}
 
 app.use(errors.notFound)
 app.use(errors.catchAll)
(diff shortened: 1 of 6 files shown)


src/server.js
```diff
@@ -80,6 +94,11 @@ app.use(store())
 app.use(slashify())
 app.use(routers)
 
+// Raven error handler must come before other error middleware
+if (config.sentryDsn) {
+  app.use(raven.errorHandler())
```
teneightfive: We use [winston](https://github.com/winstonjs/winston) for error logging so wonder if something like [winston-sentry](https://github.com/guzru/winston-sentry) could handle the error logging?

reupen: You mean in addition to the exception reporting directly from Raven?

Though winston-sentry doesn't look well-maintained as it's using Raven 1.1.4. Current version is 2.1.1.

reupen:
Here's an example exception report:

page-shot-2017-8-11 typeerror_ cannot read property name of null

src/server.js
```diff
@@ -33,6 +34,19 @@ const app = express()
 
 app.disable('x-powered-by')
 
+if (config.sentryDsn) {
```
feedmypixel: we could maybe have a flag in config name `config.enableSentry` and that flag consists of `config.isProd && config.sentryDsn`

src/server.js
feedmypixel: @teneightfive I don't think we need to add `winston-sentry` here. If you have a look into the `raven.errorHandler` its setting the sentry Id to `res.sentry`, so we can then add that into any winston logging further down the error handling and link the two up if needed. Unless you had another reason for using `winston-sentry` here?

feedmypixel:
These looks interesting 
- https://docs.sentry.io/integrations/github/
- https://docs.sentry.io/integrations/jira/

src/server.js
reupen: > are wanting this only on prod?
I would stick it on all deployed environments (dev, staging, prod). The errors can be categorised according to environment, or sent to separate projects in Sentry.

> if so can we add config.isProd && config.sentryDsn just relying on an env feels a little loose
Not sure what the benefit of that is. The environment variables are the configuration.

feedmypixel: > Not sure what the benefit of that is. The environment variables are the configuration.

Just reads nicer but agreed there is a level of abstraction that isn't helpful. Happy to leave it

src/server.js
teneightfive: Nope, lets just stick with Raven.

src/server.js
```diff
@@ -33,6 +34,19 @@ const app = express()
...
+  logger.log('info', 'Configuring Raven...')
```
teneightfive: `logger.info('Configuring Raven')`? I'd also be tempted to drop the log level to `.debug()` too.

reupen: > I'd also be tempted to drop the log level to .debug() too.

Seems like info stuff to me, it's just some indication that Raven has been enabled. It will be pretty much useless on debug since no one will run with Sentry locally.

PR #446, opened 2017-08-11
[DH-417] Fix error editing companies without a business type

reupen:
Fixes a 500 error editing non-UK and non-registered UK companies with a null business type.

### DIFF
--- a/src/apps/companies/controllers/foreign.js
+++ b/src/apps/companies/controllers/foreign.js
@@ -1,3 +1,5 @@
+const { get } = require('lodash')
+
 const companyRepository = require('../repos')
 const companyFormService = require('../services/form')
 const companyService = require('../services/data')
@@ -66,7 +68,7 @@ async function editDetails (req, res, next) {
     } else {
       res.locals.formData = companyFormService.getForeignCompanyAsFormData(company)
     }
-    res.locals.businessTypeName = company.business_type.name
+    res.locals.businessTypeName = get(company, 'business_type.name')
     res.locals.showTradingAddress = !isBlank(res.locals.formData.trading_address_country)
 
     res
--- a/src/apps/companies/controllers/ukother.js
+++ b/src/apps/companies/controllers/ukother.js
@@ -1,3 +1,5 @@
+const { get } = require('lodash')
+
 const companyRepository = require('../repos')
 const companyFormService = require('../services/form')
 const companyService = require('../services/data')
@@ -67,7 +69,7 @@ async function editDetails (req, res, next) {
     } else {
       res.locals.formData = companyFormService.getUkOtherCompanyAsFormData(company)
     }
-    res.locals.businessTypeName = company.business_type.name
+    res.locals.businessTypeName = get(company, 'business_type.name')
     res.locals.showTradingAddress = !isBlank(res.locals.formData.trading_address_country)
 
     res
--- a/src/apps/companies/views/edit-foreign.njk
+++ b/src/apps/companies/views/edit-foreign.njk
@@ -9,7 +9,9 @@
       <div id="business-type" class="key-value__key form-label-bold">Business type
         <a href="/companies/add-step-1" class="button-link">Change</a>
       </div>
-      <div class="key-value__value" aria-labelledby="business-type">Foreign {{businessTypeName}}</div>
+      <div class="key-value__value" aria-labelledby="business-type">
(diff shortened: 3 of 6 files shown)


src/apps/companies/controllers/foreign.js
```diff
@@ -1,3 +1,5 @@
+const get = require('lodash/get')
```
teneightfive: This can be: 

```js
```

Which uses [destructuring](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) and keeps it consistent with how [we import methods](https://github.com/uktrade/data-hub-frontend/blob/1553d8547d02906b23a3da98b62762113c0c3a24/src/apps/auth/controllers/sign-in.js#L2) elsewhere.

src/apps/companies/controllers/ukother.js
```diff
@@ -1,3 +1,5 @@
+const get = require('lodash/get')
```
teneightfive: Same as above.

src/apps/companies/views/edit-foreign.njk
```diff
@@ -9,7 +9,9 @@
       <div id="business-type" class="key-value__key form-label-bold">Business type
         <a href="/companies/add-step-1" class="button-link">Change</a>
       </div>
-      <div class="key-value__value" aria-labelledby="business-type">Foreign {{businessTypeName}}</div>
+      <div class="key-value__value" aria-labelledby="business-type">
+        Foreign {{businessTypeName or '(type of organisation not known)'}}
```
teneightfive: Do we need to say _type of organisation not known_? Can we just use _organisation_ as a fallback?

test/unit/apps/companies/controllers/ukother.test.js
```diff
@@ -57,7 +59,7 @@ describe('Company controller, uk other', function () {
     saveCompanyFormStub = sinon.stub().returns(fakeCompanyForm)
     flashStub = sinon.stub()
 
-    this.breadcrumbStub = function () { return this }
+    breadcrumbStub = function () { return this }
```
teneightfive: Why was this changed? `breadcrumbStub` should exist on the `this` context for all the tests, no?

src/apps/companies/controllers/foreign.js
reupen: [Not everywhere](https://github.com/uktrade/data-hub-frontend/blob/3d12520a05e82d5bfe98554d28518ea3def3aa09/src/apps/investment-projects/middleware/forms/team-members.js#L1), but will change it :)

src/apps/companies/views/edit-foreign.njk
reupen: Wording was from Mills.

test/unit/apps/companies/controllers/ukother.test.js
reupen: Why should it be on `this`? I made it consistent with `foreign.test.js` so that it works with arrow functions.

teneightfive: I suppose it fits with the rest of this file for now. We have a few different methods of referencing variables at the moment but would be good to be consistent at some point.

teneightfive: My concern was that it was adding noise to this fix. The placement of `breadcrumbStub` isn't concerned with this change. 

If you feel it needs to be addressed possibly split it into a separate commit on this PR?

src/apps/companies/views/edit-ukother.njk
```diff
@@ -9,7 +9,9 @@
       <div id="business-type" class="key-value__key form-label-bold">Business type
         <a href="/companies/add-step-1" class="button-link">Change</a>
       </div>
-      <div class="key-value__value" aria-labelledby="business-type">{{businessTypeName}}</div>
+      <div class="key-value__value" aria-labelledby="business-type">
+        {{businessTypeName or 'Type of organisation not known'}}
```
feedmypixel: is this meant to be capitalised and without brackets? the one above reads:
```(type of organisation not known)```
Where as this read:
```Type of organisation not known```

reupen: The other one was prefixed with 'Foreign', while this one had no prefix, hence the difference. (Though that does suggest that this one should be prefixed with UK...)

PR #447, opened 2017-08-11
Further review the building blocks of collections

tyom:
- Improve meta filtering in `Entity` component. Drop the need for optional properties to be always set
- Improve `applyMacro` to allow it take optional shared properties which mix macro config properties 
- Remove the object/array handling logic in `applyMacro`. Always work with objects.
- Create `constants` module to contain static configuration objects
- Add global transformer `transformFieldsObjectToMacrosObject` which takes field object:
  ```js
  { label: 'City', value: 'London', macroName: 'MetaItem' }
  ```
  and transforms it into macro configuration object:
  ```js
  {
    MetaItem: { label: 'City', value: 'London' }
  }
  ```
  which makes it easier to preconfigure macros in middleware and simply render in views with `applyMacro(macroConfigObject)`.

This work paves the way for other entities and collections as they can share many components.


config/nunjucks/filters.js
```diff
@@ -29,6 +30,7 @@ const filters = {
   assign,
   concat,
   filter,
+  reject,
```
teneightfive: We may as well make the whole of lodash available in the views 😂

tyom: LOL.

`module.exports = assign(filters, _)`

PR #448, opened 2017-08-11
Generic Meta macro and CSS component

tyom:
- Add `c-meta-container` CSS component with modifiers to configure the display style of `c-meta-item`s.
- Replace `c-meta-list` for more generic component


src/apps/investment-projects/middleware/shared.js
```diff
@@ -51,7 +52,7 @@ async function getInvestmentDetails (req, res, next, id = req.params.id) {
 
     res.locals.investmentStatus = {
       id: investmentData.id,
-      meta: [
+      meta: transformFieldsObjectToMacrosObject([
```
feedmypixel: is this work missing?

tyom: Hmm.. Might have lost that one somewhere between branches.

assets/stylesheets/components/_meta-container.scss
```diff
@@ -0,0 +1,48 @@
+@import "../settings";
+@import "../tools";
+
+.c-meta-container {
```
teneightfive: I think we should combine `.c-meta-container` and `.c-meta-item` seeing as the container is always included when adding meta items now.

The modifiers in this file aren't modifying this block they are modifying a different block which doesn't feel right to me.

src/apps/investment-projects/middleware/shared.js
tyom: Just realised that the function was added in #447 (`simplify-macro-config` branch) which is what this branch depends on.

assets/stylesheets/components/_meta-list.scss
```diff
@@ -1,11 +1,75 @@
...
     margin-left: $default-spacing-unit;
   }
 
-  .c-meta-item__label {
+  .c-meta-list__item-label {
     display: block;
   }
 }
+
+.c-meta-list--alternate {
```
teneightfive: Wonder if there's a better name than `alternate`. Alternate to what? Maybe `--split` or something to say that they're floated right and left.

config/nunjucks/filters.js
```diff
@@ -139,6 +139,15 @@ const filters = {
     }
     return result
   },
+
+  classModifier (className, modifier) {
```
teneightfive: 👏 but what about `applyClassModifiers`? Bit more specific as to what its doing as its not actually modifying the class you give it.

PR #453, opened 2017-08-14
Feature/add another js

ztolley:
Component that scans markup to find elements that require a 'add another' button
The target element defines a selector to define what page fragment should be duplicated when the user presses add another. The contents of the cloned fragment will have it's fields reset. The component also inserts a 'remove' button that will allow values to also be removed.

### DIFF
--- a/assets/javascripts/_deprecated/lib/element-stuff.js
+++ b/assets/javascripts/_deprecated/lib/element-stuff.js
@@ -183,6 +183,29 @@ function regenIds (wrapper) {
     })
 }
 
+function closest (element, selector) {
+  if (!element) {
+    return
+  }
+
+  if (!element.matches) {
+    element.prototype.matches = element.prototype.msMatchesSelector || element.prototype.webkitMatchesSelector
+  }
+
+  let parent
+
+  // traverse parents
+  while (element) {
+    parent = element.parentElement
+    if (parent && parent.matches(selector)) {
+      return parent
+    }
+    element = parent
+  }
+
+  return null
+}
+
 module.exports = {
   addClass,
   removeClass,
@@ -199,4 +222,5 @@ module.exports = {
   toggleVisible,
   regenIds,
   resetFieldValues,
+  closest,
 }
--- a/assets/javascripts/_deprecated/lib/element-stuff.js
+++ b/assets/javascripts/_deprecated/lib/element-stuff.js
@@ -58,13 +58,18 @@ function toggleVisible (element) {
   }
 }
 
-function generateID () {
-  let d = new Date().getTime()
-  return 'xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx'.replace(/[xy]/g, function (c) {
-    const r = (d + Math.random() * 16) % 16 | 0
-    d = Math.floor(d / 16)
-    return (c === 'x' ? r : (r & 0x3 | 0x8)).toString(16)
-  })
+/**
+ * generateId
+ *
+ * Create an ID string unique enough to use as an element ID in a page.
+ * The prefix is required as ID's cannot start with a number or a hyphen
+ *
+ * @param {string} prefix
+ * @returns {string} a unique string to use to indentify an element
+ */
+function generateID (prefix) {
+  const _prefix = (prefix && prefix.length > 0) ? prefix : 'dh'
+  return `${_prefix}-${Math.floor(Math.random() * 1000000) + 1}`
 }
 
 function isNodeList (nodes) {
@@ -119,6 +124,65 @@ function removeElement (element) {
   element.parentNode.removeChild(element)
 }
 
+/**
+ * resetFieldValues
+ *
+ * Scans the content of a dom fragment for fields
+ * and resets their values back to nothing.
+ *
+ * Useful when copying fields to use as new fields
+ *
+ * @param {nodeElement} fragment
+ */
+function resetFieldValues (fragment) {
+  Array
+    .from(fragment.querySelectorAll('option:checked'))
+    .forEach(selectedElement => {
+      selectedElement.selected = false
+    })
+
+  Array
+    .from(fragment.querySelectorAll('input:checked'))
+    .forEach(checkedElement => {
+      checkedElement.checked = false
+    })
+
+  Array
+    .from(fragment.querySelectorAll('input[type="text"], textarea'))
+    .forEach(textField => {
+      textField.value = ''
+    })
+}
+
+/**
+ * regenIds
+ * *
+ * Get all the all the elements in a fragment that have an ID
+ * attribute, generate a new one to replace it and look for related
+ * label tags to update their 'for' attribute
+ *
+ * @param {nodeElement} wrapper
+ *
+ */
+function regenIds (wrapper) {
+  Array
+    .from(wrapper.querySelectorAll('*[id]'))
+    .forEach((element) => {
+      const oldId = element.id
+
+      // If the element has a name, use that as part of the new ID
+      const name = element.name || ''
+      const newId = generateID(name)
+
+      element.id = newId
+
+      const relatedLabel = wrapper.querySelector(`[for="${oldId}"]`)
+      if (relatedLabel) {
+        relatedLabel.setAttribute('for', newId)
+      }
+    })
+}
+
 module.exports = {
   addClass,
   removeClass,
@@ -133,4 +197,6 @@ module.exports = {
   createElementFromMarkup,
   removeElement,
   toggleVisible,
+  regenIds,
+  resetFieldValues,
 }
--- a/assets/javascripts/modules/add-another-fragment.js
+++ b/assets/javascripts/modules/add-another-fragment.js
@@ -1,0 +1,291 @@
+const {
+  insertAfter,
+  resetFieldValues,
+  regenIds,
+  closest,
+  hide,
+  show,
+} = require('../_deprecated/lib/element-stuff')
+
+const wrapperClass = 'js-addanother'
+const defaultAddButtonClass = 'js-addanother__add'
+const defaultRemoveButtonClass = 'js-addanother__remove'
+
+/**
+ * AddAnotherFragment
+ *
+ * Component that scans markup to find elements that require a 'add another' button
+ * The target element defines a selector to define what page fragment should be duplicated
+ * when the user preses add another. The contents of the cloned fragment will have it's fields reset.
+ * The component also inserts a 'remove' button that will allow values to also be removed.
+ *
+ * The selector used to activate the component is 'js-add-another-fragment'
+ *
+ * The element that includes the activation class must also provide data attributes to define
+ * it's behaviour
+ *
+ * data-fragment-selector: The css selector that defines the element to clone.
+ * data-fragment-allow-remove-all: (Optional) By default the component won't let the user delete the last instance
+ *                                 of the defined element. If the user needs to be able to delete all instances
+ *                                 then provide a value 'yes'
+ * data-fragment-add-button-selector: (Optional) By default the component will add a 'add another' button
+ *                                 to the end of the wrapper fragment. If an add button already exists
+ *                                 within the markup then the css selector to that button can be specified.
+ * data-fragment-add-button-text: (Optional) Allow the text for the 'add another' button to be specified
+ * data-fragment-remove-selector: (Optional) If the initial markup includes a remove button then this
+ *                                selector will attach the remove event handler to it intead of adding a button
+ * data-fragment-remove-button-text: (Optional) Allow the text for the 'add another' button to be specified
+ *
+ * e.g.
+ * <div class="js-add-another-fragment" data-fragment-selector=".js-adviser">
+ *   <div class="c-form-group js-adviser">
+ *     <label for="adviser">Adviser</label>
+ *     <select name="adviser">
+ *       <option value="">---Select Adviser--</option>
+ *       ...
+ *     </select>
+ *   </div>
+ * </div>
+ */
+const AddAnotherFragment = {}
+
+AddAnotherFragment.prototype = {
+  /**
+   * init
+   *
+   * Main setup method
+   * Gets params, generates a template for new fragments
+   * and adds and activates add and delete buttons
+   */
+  init () {
+    this.parseProperties()
+    this.lastFragment = this.wrapper.querySelector(`${this.fragmentSelector}:last-of-type`)
+    this.getTemplate()
+    this.addPlaceholder()
+    this.setupAddAndRemoveButtons()
+    this.bindEvents()
+  },
+
+  /**
+   * setupAddAndRemoveButtons
+   *
+   * Parse the content and render additional markup,
+   * or decorate existing markup if required
+   */
+  setupAddAndRemoveButtons () {
+    if (!this.overrideAddButton) {
+      this.insertAddButton()
+    }
+    this.decorateAddButton()
+
+    if (!this.overrideRemoveButton) {
+      this.insertRemoveButtons()
+    }
+    this.decorateRemoveButtons()
+    this.updateRemoveButtonVisibility()
+  },
+
+  /**
+   * parseProperties
+   *
+   * Look at the wrapper element for parameters and properties
+   * that describe the behaviour of the component
+   */
+  parseProperties () {
+    this.fragmentSelector = this.wrapper.getAttribute('data-selector')
+
+    this.allowRemoveAll = this.wrapper.hasAttribute('data-allow-remove-all')
+
+    this.overrideAddButton = this.wrapper.hasAttribute('data-add-button-selector')
+    this.addButtonSelector = this.wrapper.getAttribute('data-add-button-selector') || `.${defaultAddButtonClass} a`
+    this.addText = this.wrapper.getAttribute('data-add-button-text') || 'Add another'
+
+    this.overrideRemoveButton = this.wrapper.hasAttribute('data-remove-button-selector')
+    this.removeButtonSelector = this.wrapper.getAttribute('data-remove-button-selector') || `.${defaultRemoveButtonClass} a`
+    this.removeText = this.wrapper.getAttribute('data-remove-button-text') || 'Remove'
+  },
+
+  /**
+   * addEventListeners
+   *
+   * Attach an event listener to the wrapper
+   * element that waits for the use to click on
+   * the add and remove buttons, and intercept and redirects
+   * those events, so events don't have to be attached to each
+   * link
+   */
+  bindEvents () {
+    this.wrapper.addEventListener('click', this.clickHandler.bind(this))
+  },
+
+  /**
+   * getTemplate
+   *
+   * Takes a copy of a fragment and resets the field values
+   * so it can be used as a blueprint for a new fragment
+   *
+   */
+  getTemplate () {
+    this.templateFragment = this.lastFragment
+    resetFieldValues(this.templateFragment)
+  },
+
+  /**
+   * addPlaceholder
+   *
+   * Insert a placeholder under the last existing fragment,
+   * so new fragments are always inserted in to the corect place
+   * even when adding a new fragment when there are none.
+   */
+  addPlaceholder () {
+    this.placeHolder = this.document.createElement('div')
+    insertAfter(this.placeHolder, this.lastFragment)
+  },
+
+  /**
+   * decorateAddButton
+   *
+   * Adds a data attribute to the add button so the event handler
+   * recognises it is an add button for this component.
+   */
+  decorateAddButton () {
+    const addButton = this.wrapper.querySelector(this.addButtonSelector)
+    if (addButton) {
+      addButton.setAttribute('data-method', 'add')
+    }
+  },
+
+  /**
+   * decorateRemoveButton
+   *
+   * Adds a data attribute to all remove buttons so the event handler
+   * recognises them as a remove button for this component.
+   */
+  decorateRemoveButtons () {
+    Array
+      .from(this.wrapper.querySelectorAll(this.removeButtonSelector))
+      .forEach((element) => {
+        element.setAttribute('data-method', 'remove')
+      })
+  },
+
+  /**
+   * insertAddButton
+   *
+   * Creates a new button to 'add another' and appends it after the
+   * placeholder that indicates the end of the fragments
+   */
+  insertAddButton () {
+    const addButtonElement = this.document.createElement('a')
+    addButtonElement.href = '#'
+    addButtonElement.innerHTML = this.addText
+
+    const addButtonWrapper = this.document.createElement('p')
+    addButtonWrapper.className = `c-form-group c-form-group--actions ${defaultAddButtonClass}`
+    addButtonWrapper.appendChild(addButtonElement)
+
+    insertAfter(addButtonWrapper, this.placeHolder)
+  },
+
+  /**
+   * insertRemoveButtons
+   *
+   * Adds a 'remove' button to the each of each of the fragments.
+   */
+  insertRemoveButtons () {
+    Array.from(this.wrapper.querySelectorAll(this.fragmentSelector))
+      .forEach((fragment) => {
+        const removeButtonElement = this.document.createElement('a')
+        removeButtonElement.innerHTML = this.removeText
+        removeButtonElement.href = '#'
+
+        const p = this.document.createElement('p')
+        p.className = `c-form-group c-form-group--actions ${defaultRemoveButtonClass}`
+        p.appendChild(removeButtonElement)
+
+        fragment.appendChild(p)
+      })
+  },
+
+  /**
+   * updateRemoveButtonVisibility
+   *
+   * Determines if remove buttons can be seen based on rules around
+   * allowing a user to delete, or not delete all fragments
+   */
+  updateRemoveButtonVisibility () {
+    const removeButtons = Array.from(this.wrapper.querySelectorAll(this.removeButtonSelector))
+
+    // Don't allow the last fragment to be deleted unless it's the desired behaviour
+    if (removeButtons.length === 1 && !this.allowRemoveAll) {
+      hide(removeButtons[0])
+    } else {
+      removeButtons.forEach((fragment) => show(fragment))
+    }
+  },
+
+  // When the user presses 'add another', clone a template
+  // Then reset any fields id's within it
+  addClickHandler (event) {
+    event.preventDefault()
+    event.target.blur()
+
+    const clonedFragment = this.templateFragment.cloneNode(true)
+    regenIds(clonedFragment)
+
+    // Add the fragment below the last one, unless there are none left,
+    // in which case add it before the 'add another' button
+    this.wrapper.insertBefore(clonedFragment, this.placeHolder)
+
+    this.updateRemoveButtonVisibility()
+  },
+
+  // Handle when a user clicks on a 'remove' link
+  // Gets the selector for the fragment to remove and removes it.
+  // Then updates the last fragment in case we just deleted it.
+  removeClickHandler (event) {
+    event.preventDefault()
+    event.target.blur()
+
+    // Remove the fragment that contains the remove button
+    const fragment = closest(event.target, this.fragmentSelector)
+    this.wrapper.removeChild(fragment)
+
+    this.updateRemoveButtonVisibility()
+  },
+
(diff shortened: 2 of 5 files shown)


assets/javascripts/_deprecated/lib/element-stuff.js
```diff
@@ -119,6 +124,67 @@ function removeElement (element) {
...
+  Array
+    .from(fragment.querySelectorAll('option:checked'))
+    .forEach(selectedElement => {
+      selectedElement.selected = false
+    })
+
+  Array
+    .from(fragment.querySelectorAll('input:checked'))
+    .forEach(checkedElement => {
+      checkedElement.checked = false
```
teneightfive: Could you reuse the same function that resets for this as `option:checked`?

assets/javascripts/_deprecated/lib/element-stuff.js
```diff
@@ -119,6 +124,67 @@ function removeElement (element) {
...
+    })
+
+  Array
+    .from(fragment.querySelectorAll('input[type="text"]'))
```
teneightfive: Could the selector be `input[type="text"], textarea` and then you've covered off your `// todo` on line 156?

assets/javascripts/_deprecated/lib/element-stuff.js
teneightfive: Maybe even `fragment.querySelectorAll('input:checked, option:checked')`?

assets/javascripts/modules/add-another-fragment.js
```diff
@@ -0,0 +1,206 @@
+const { insertAfter, generateID, resetFieldValues, regenIds } = require('../_deprecated/lib/element-stuff')
+
+const AddAnotherFragment = {}
+/**
+ * AddAnotherFragment
+ *
+ * Component that scans markup to find elements that require a 'add another' button
+ * The target element defines a selector to define what page fragment should be duplicated
+ * when the user preses add another. The contents of the cloned fragment will have it's fields reset.
+ * The component also insers a 'remove' button that will allow values to also be removed.
```
teneightfive: `insers` -> `inserts`

assets/javascripts/modules/add-another-fragment.js
```diff
@@ -0,0 +1,206 @@
...
+
+    // Don't allow the last fragment to be deleted unless it's the desired behaviour
+    if (this.fragmentCollection.length === 1 && !this.allowRemoveAll) {
+      this.deleteRemoveButton(this.lastFragment)
+    }
+  },
+}
+
+AddAnotherFragment.init = function (_document = document) {
+  const stuff = Array.from(_document.querySelectorAll('.js-add-another-fragment'))
```
teneightfive: `stuff`? Really?!

assets/javascripts/modules/add-another-fragment.js
```diff
@@ -0,0 +1,206 @@
...
+ *     <label for="adviser">Adviser</label>
+ *     <select name="adviser">
+ *       <option value="">---Select Adviser--</option>
+ *       ...
+ *     </select>
+ *   </div>
+ * </div>
+ */
+AddAnotherFragment.prototype = {
+  startup: function () {
```
teneightfive: Wonder if there's a better name than `startup`? It's not really a common method name we've used for the init method of our client side components.

assets/javascripts/modules/add-another-fragment.js
```diff
@@ -0,0 +1,206 @@
...
+    this.parseProperties()
+    this.cacheElements()
+    this.insertAddButton()
+    this.updateRemoveButtons()
+  },
+
+  parseProperties: function () {
```
teneightfive: What about assigning settings within the _startup_ method and using [`.dataset`](https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes#JavaScript_access) to get the data attributes:

```js
startup: function (options) {
  const defaults = {
    ...
  }
  this.settings = Object.assign({}, defaults, options, this.wrapper.dataset)
}
```

assets/javascripts/modules/add-another-fragment.js
```diff
@@ -0,0 +1,206 @@
...
+   */
+  insertAddButton: function () {
+    let addButtonElement
+
+    if (this.addButtonSelector) {
+      addButtonElement = this.wrapper.querySelector(this.addButtonSelector)
+    } else {
+      addButtonElement = this.document.createElement('a')
+      addButtonElement.href = '#'
+      addButtonElement.innerHTML = 'Add another'
```
teneightfive: Consider making this an option that can be changed using data attributes. In my case it would be good to use `Add another adviser`.

assets/javascripts/modules/add-another-fragment.js
```diff
@@ -0,0 +1,206 @@
...
+   * insertAddButton
+   *
+   * Creates a new button to 'add another' and appends it after the
+   * fragments or adds an event handler to an existing button to duplicate fields
+   */
+  insertAddButton: function () {
+    let addButtonElement
+
+    if (this.addButtonSelector) {
+      addButtonElement = this.wrapper.querySelector(this.addButtonSelector)
```
teneightfive: It feels like this might be better placed in `cacheEls`. And if not there, it can call a function to `appendAddButton()`. 

It feels this function is doing more than its name suggests. It doesn't always insert the button if it exists.

assets/javascripts/modules/add-another-fragment.js
```diff
@@ -0,0 +1,206 @@
...
+      addButtonElement.innerHTML = 'Add another'
+
+      this.addButtonWrapper = this.document.createElement('p')
+      this.addButtonWrapper.className = 'c-form-group c-form-group--actions js-add-button'
+      this.addButtonWrapper.appendChild(addButtonElement)
+
+      insertAfter(this.addButtonWrapper, this.lastFragment)
+    }
+
+    addButtonElement.addEventListener('click', this.addClickHandler.bind(this))
```
teneightfive: This statement might be better placed in a function called `bindEvents()` which gets called from the `init()` method.

assets/javascripts/modules/add-another-fragment.js
```diff
@@ -0,0 +1,206 @@
...
+  },
+
+  // Handle when a user clicks on a 'remove' link
+  // Gets the selector for the fragment to remove and removes it.
+  // Then updates the last fragment in case we just deleted it.
+  removeClickHandler: function (event) {
+    event.preventDefault()
+    event.target.blur()
+
+    const selector = event.target.getAttribute('data-remove-selector')
```
teneightfive: What about using [dataset](https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes#JavaScript_access)?

assets/javascripts/modules/add-another-fragment.js
```diff
@@ -0,0 +1,206 @@
...
+    this.updateFragmentCollection()
+
+    // Don't allow the last fragment to be deleted unless it's the desired behaviour
+    if (this.fragmentCollection.length === 1 && !this.allowRemoveAll) {
+      this.deleteRemoveButton(this.lastFragment)
+    }
+  },
+}
+
+AddAnotherFragment.init = function (_document = document) {
```
teneightfive: Maybe `wrapper` rather than `_document`?

teneightfive:
It feels like it might be handy to do this in the components area first and mock up some example use cases that we know we already have:

- Add/remove starting from 0 and ability to remove all
- Add/remove starting from 1 and only able to remove to last item
- Add button already present
- Remove button(s) already present

teneightfive:
It also seems like the add/remove functionality on the team page doesn't work without JavaScript, was that always the case?

assets/javascripts/_deprecated/lib/element-stuff.js
ztolley: One uses selected, the other uses checked.

assets/javascripts/modules/add-another-fragment.js
ztolley: yeah... stuff

assets/javascripts/modules/add-another-fragment.js
ztolley: Init is taken and it's only run the first time through. Open to suggestions

assets/javascripts/modules/add-another-fragment.js
ztolley: JSDOM doesn't support dataset at this time so it would not be possible to test this

assets/javascripts/modules/add-another-fragment.js
ztolley: jsdom doesn't support it so I cant write unit tests that use dataset.

assets/javascripts/modules/add-another-fragment.js
ztolley: Already using wrapper now and it needs to be a document, as it needs to support the create element method. The only reason it's in there is to allow unit testing to pass a document in as document isn't in the global namespace by default.

assets/javascripts/_deprecated/lib/element-stuff.js
teneightfive: These are the reason why its better not to write your own library to do trivial things in the browser :(

assets/javascripts/modules/add-another-fragment.js
teneightfive: You could use 2 different objects? One that handles everything for 1 instance and one that is exported and whose `init` method loops over each element caught by the selector and attaches the instance object.

That way we're not introducing another term for a method which is doing the same sort of thing as `init` everywhere else.

assets/javascripts/modules/add-another-fragment.js
teneightfive: Wrapper is being used below but something like `element` might be better there if you rename `stuff` to `elements`. That would make way for the argument of this method to be `wrapper` rather than using `_document = document`.

It feels like the days of `var _this = this` or `var that = this`.

assets/javascripts/modules/add-another-fragment.js
teneightfive: `elements`.

assets/javascripts/modules/add-another-fragment.js
ztolley: thats a big change for the sake of a method name

assets/javascripts/modules/add-another-fragment.js
ztolley: I reworked it a little but I like the school of having a function do just one thing

ztolley:
with regards to the comment about test scenarios, those are handled in the unit tests already except existing remove button which I will look at.

As for none js, currently you can remove a line by de-selecting the adviser. Not ideal, but functional.
With the idea of the existing remove button this could be added but then there is the balance of the effort it requires vs the benefit to the user.

assets/javascripts/_deprecated/lib/element-stuff.js
ztolley: You wanna use jQuery?

ztolley:
In the process of refactoring to add the support for the re-rendered remove button.

Done the component, now have to update the tests to take into consideration I am hiding remove buttons and add the new tests to test the pre-rendered function.

Also add the ability to specify the remove text.

assets/javascripts/_deprecated/lib/element-stuff.js
teneightfive: Not necessarily something as heavy as that but using a small library for trivial interactions with the DOM could save us some headaches.

assets/javascripts/_deprecated/lib/element-stuff.js
```diff
@@ -58,13 +58,18 @@ function toggleVisible (element) {
...
-  return 'xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx'.replace(/[xy]/g, function (c) {
-    const r = (d + Math.random() * 16) % 16 | 0
-    d = Math.floor(d / 16)
-    return (c === 'x' ? r : (r & 0x3 | 0x8)).toString(16)
-  })
+/**
+ * generateId
+ *
+ * Create an ID string unique enough to use as an element ID in a page.
+ * The prefix is required as ID's cannot start with a number of a dash
```
teneightfive: `number of a dash` -> `number or a dash`

assets/javascripts/_deprecated/lib/element-stuff.js
ztolley: I'm starting to agree. Something to allow dom manipulation and also fill in some of the limits of some browsers, to give something like 'closest'.

But what? most people don't bother and just use jquery.

assets/javascripts/_deprecated/lib/element-stuff.js
teneightfive: Could this method not be smarter to ensure that the first character is never a number or dash?

assets/javascripts/_deprecated/lib/element-stuff.js
```diff
@@ -119,6 +124,88 @@ function removeElement (element) {
...
+ * regenIds
+ * *
+ * Get all the all the elements in a fragment that have an ID
+ * attribute, generate a new one to replace it and look for related
+ * label tags to update their 'for' attribute
+ *
+ * @param {nodeElement} fragment
+ *
+ */
+function regenIds (fragment) {
```
teneightfive: Would `element` or `wrapper` be a better name for the argument? Feels more consistent with what we've called things previously.

assets/javascripts/modules/add-another-fragment.js
```diff
@@ -0,0 +1,277 @@
+const {
+  insertAfter,
+  resetFieldValues,
+  regenIds,
+  closest,
+  hide,
+  show,
+} = require('../_deprecated/lib/element-stuff')
+
+const defaultAddButtonClass = 'js-add-button'
```
teneightfive: Wonder if using a classname that namespaces from this object would be clearer.

So maybe call the main js class `js-AddAnother` or `js-RepeatableContent` and then use an element style selector to show child selectors `js-AddAnother__add` and `js-AddAnother__remove`. This will be clear in the markup that these JavaScript selectors are then tied to this module.

assets/javascripts/modules/add-another-fragment.js
```diff
@@ -0,0 +1,277 @@
...
+
+AddAnotherFragment.prototype = {
+  /**
+   * init
+   *
+   * Main setup method
+   * Gets params, generates a template for new fragments
+   * and adds and activates add and delete buttons
+   */
+  init: function () {
```
teneightfive: Maybe use shorthand for these method declarations (or at least named functions rather than anonymous functions):

```js
init () {
  ...
}
```

assets/javascripts/modules/add-another-fragment.js
```diff
@@ -0,0 +1,277 @@
...
+    this.addEventListeners()
+  },
+
+  /**
+   * parseProperties
+   *
+   * Look at the wrapper element for parameters and properties
+   * that describe the behaviour of the component
+   */
+  parseProperties: function () {
```
teneightfive: It feels like this method could be split up a bit better to follow the existing pattern.

Maybe set the module settings in the `init()` method and then cache the elements in a `cacheEls()` method. This could have some logic to use selectors set within the module settings if they are truthy.

assets/javascripts/modules/add-another-fragment.js
```diff
@@ -0,0 +1,277 @@
...
+   * Main setup method
+   * Gets params, generates a template for new fragments
+   * and adds and activates add and delete buttons
+   */
+  init: function () {
+    this.parseProperties()
+    this.getTemplate()
+
+    if (!this.overrideAddButton) {
+      this.insertAddButton()
```
teneightfive: These methods could be combined into a `render()` method which could be called from this `init()` method and nicely extract the logic for the first render of the module state.

assets/javascripts/_deprecated/lib/element-stuff.js
reupen: That's a hyphen, not a dash :P

assets/javascripts/modules/add-another-fragment.js
```diff
@@ -0,0 +1,277 @@
...
+    }
+    this.decorateAddButton()
+
+    if (!this.overrideRemoveButton) {
+      this.insertRemoveButtons()
+    }
+    this.decorateRemoveButtons()
+    this.updateRemoveButtonVisibility()
+
+    this.addEventListeners()
```
teneightfive: `bindEvents()` to be consistent with [other](https://github.com/uktrade/data-hub-frontend/blob/5997da8670458bf4f58235dc1c5d009670d3f65c/assets/javascripts/modules/label-select.js#L15) [modules](https://github.com/uktrade/data-hub-frontend/blob/5997da8670458bf4f58235dc1c5d009670d3f65c/assets/javascripts/modules/conditional-subfields.js#L55) we've created?

assets/javascripts/modules/add-another-fragment.js
```diff
@@ -0,0 +1,277 @@
...
+
+  // Listen for button clicks on the wrapper and direct events to the correct
+  // handler if they are to add or remove
+  clickHandler: function (event) {
+    const target = event.target
+    if (!target.hasAttribute('data-method')) {
+      return
+    }
+
+    switch (target.getAttribute('data-method')) {
```
teneightfive: Rather than using `switch` it might be better to bind event handlers separately and set the listeners in the `bindEvents()` method using event delegation.

assets/javascripts/modules/add-another-fragment.js
```diff
@@ -0,0 +1,277 @@
...
+        break
+      case 'remove':
+        this.removeClickHandler(event)
+        break
+    }
+  },
+}
+
+module.exports = {
+  init (_document = document) {
```
teneightfive: Still think `wrapper = document` would be better than `_document`.

assets/javascripts/modules/add-another-fragment.js
ztolley: its much easier to listen to events at the wrapper level, as you are constantly adding and removing fragments and handling their events too gets messy.

src/apps/investment-projects/views/team/edit-team-members.njk
```diff
@@ -7,25 +7,28 @@
     {% if form.errors.team_members %}
       <div class="error">{{ form.errors.team_members }}</div>
     {% endif %}
+    <div class="js-add-another-fragment"
+      data-fragment-selector=".c-form-fieldset"
+      data-fragment-allow-remove-all="yes">
```
teneightfive: I wonder if you need to namespace the data attributes. And also whether this option could be decided based on its presence. So having an attribute of `data-allow-remove-all` might be enough. Or at least use `true` as the value rather than `yes`.

assets/javascripts/modules/add-another-fragment.js
ztolley: its not caching elements though. The the properties method is focussed on extrating parameters from the data attributes.

assets/javascripts/modules/add-another-fragment.js
ztolley: that wont work with the unit tests

src/apps/investment-projects/views/team/edit-team-members.njk
ztolley: not sure we need namespace but agree presence should be enough

ztolley: actually I'm namespacing it

assets/javascripts/modules/add-another-fragment.js
teneightfive: Surely changing the variable name shouldn't affect tests? I thought the only reason this argument was there was to support unit tests. 

I meant replace `init (_document = document) {` with `init (wrapper = document) {` to avoid having to use underscore prefixes.

src/apps/investment-projects/views/team/edit-team-members.njk
teneightfive: Yes, I was suggesting we remove the namespace.

ztolley: ah

assets/javascripts/modules/add-another-fragment.js
```diff
@@ -0,0 +1,285 @@
...
+ * when the user preses add another. The contents of the cloned fragment will have it's fields reset.
+ * The component also inserts a 'remove' button that will allow values to also be removed.
+ *
+ * The selector used to activate the component is 'js-add-another-fragment'
+ *
+ * The element that includes the activation class must also provide data attributes to define
+ * it's behaviour
+ *
+ * data-fragment-selector: The css selector that defines the element to clone.
+ * data-fragment-allow-remove-all: (Optional) By default the component won't let the user delete the last instance
```
teneightfive: Feels like this would be better described using JSDoc like we have elsewhere.

assets/javascripts/modules/add-another-fragment.js
```diff
@@ -0,0 +1,285 @@
...
+   * Look at the wrapper element for parameters and properties
+   * that describe the behaviour of the component
+   */
+  parseProperties () {
+    this.fragmentSelector = this.wrapper.getAttribute('data-selector')
+
+    this.allowRemoveAll = this.wrapper.hasAttribute('data-allow-remove-all')
+
+    this.overrideAddButton = this.wrapper.hasAttribute('data-add-button-selector')
+    this.addButtonSelector = this.wrapper.getAttribute('data-add-button-selector') || `.${defaultAddButtonClass} a`
```
teneightfive: Is this default selector correct? Should the class selector not be enough? Does it need `a`?

assets/javascripts/modules/add-another-fragment.js
```diff
@@ -0,0 +1,285 @@
...
+  },
+
+  /**
+   * parseProperties
+   *
+   * Look at the wrapper element for parameters and properties
+   * that describe the behaviour of the component
+   */
+  parseProperties () {
+    this.fragmentSelector = this.wrapper.getAttribute('data-selector')
```
teneightfive: I think there's a better way to handle these settings. Why not optionally pass `options` in to the `init()` method and set `this.settings` in the `init()` method:

```js
const defaults = {
  ...
}

AddAnotherFragment.prototype = {
  init (options) {
    this.settings = Object.assign({}, defaults, options)
  }
}
```

then in the loop where you call the object:

```js
const options = {
  ...
  selector: this.wrapper.getAttribute('data-selector'),
  allowRemoveAll: this.wrapper.hasAttribute('data-allow-remove-all'),
  ...
}

addAnotherFragment.init(options)
```

assets/javascripts/modules/add-another-fragment.js
```diff
@@ -0,0 +1,285 @@
...
+   * getTemplate
+   *
+   * Takes a copy of a fragment and resets the field values
+   * so it can be used as a blueprint for a new fragment
+   *
+   * Also inserts a placeholder under the last existing fragment,
+   * so new fragments are always inserted in to the corect place
+   * even when adding a new fragment when there are none.
+   */
+  getTemplate () {
```
teneightfive: This method seems to be doing more than what it says. It's appending content as well as getting the value for things.

Might be easier to read it you split this into storing of the template within `init()` or some kind of `cacheEls()` type method and then did the rendering in a `render()` method that is called from `init()`.

assets/javascripts/modules/add-another-fragment.js
```diff
@@ -0,0 +1,285 @@
...
+    insertAfter(this.placeHolder, lastFragment)
+  },
+
+  /**
+   * decorateAddButton
+   *
+   * Adds a data attribute to the add button so the event handler
+   * recognises it is an add button for this component.
+   */
+  decorateAddButton () {
```
teneightfive: These actions could also happen in a `render()` method called from `init()`.

assets/javascripts/modules/add-another-fragment.js
```diff
@@ -0,0 +1,285 @@
...
+  },
+
+  /**
+   * insertRemoveButtons
+   *
+   * Adds a 'remove' button to the each of each of the fragments.
+   */
+  insertRemoveButtons () {
+    Array
+      .from(this.wrapper.querySelectorAll(this.fragmentSelector))
```
teneightfive: Move this up a line? Don't think it needs its own line like the next statement.

assets/javascripts/modules/add-another-fragment.js
ztolley: Where else have we used jsdoc for data attribute documentation?

assets/javascripts/modules/add-another-fragment.js
ztolley: Yes, the default button has a wrapper with a link in it, though the overriding one that can be sent in could be something else.

assets/javascripts/modules/add-another-fragment.js
ztolley: It feels like that would be moving logic that belongs in the object into the bootstrap function that is only supposed to grab the elements on the page and initialise them

assets/javascripts/modules/add-another-fragment.js
ztolley: You would then have an init function that is doing many different things, plus render is often associated with a function continuously called to render state (like react).

It would be better to stick with the advice you have given elsewhere of making a function do just one thing.

assets/javascripts/modules/add-another-fragment.js
teneightfive: We've been using JSDoc style to document components. It just felt like it might easier to read the configuration using JSDoc syntax.

If we use the format I suggested below by changing the module to accept a set of options or configuration then JSDoc could be more appropriate too.

assets/javascripts/modules/add-another-fragment.js
teneightfive: Can't we attach the default selector to the actual element rather than needing a child selector too?

assets/javascripts/modules/add-another-fragment.js
teneightfive: I would say its keeping the module self contained. It expects a set of configuration just like it expects a wrapper.

If the module doesn't have to deal with getting of data attributes and setting their value to its options it makes it simpler to understand. 

In the instance where we initiate it we are using the option to override the defaults using data attributes but we could use a different method elsewhere or in future and if that logic isn't within the module it makes it easier to refactor.

assets/javascripts/modules/add-another-fragment.js
teneightfive: The `init()` method is basically responsible for calling the setup methods the module requires.

The `render()` method is then responsible for setting the initial (or subsequent) state of the DOM for this module to operate in. In this case, adding data attributes to the add/remove buttons. In that sense, it doesn't seem like the methods are doing too many different things.

Currently there is a setup method which is checking whether an element exists and adding if not, then always calling separate decorate methods for add remove. This does feel like it's doing different things. I think it would be clearer to group those in a `render()` method which is called from the initial `init()` method.

You could still separate the add methods too.

jim68000: Do we have a pattern doc somewhere that says what we expect from a fragment like this? If we're still discussing how to structure code after 16 days of review it's an indicator that we need to agree approaches in advance.

teneightfive: We have been trying to use more of a standard approach to modules which is what these comments are about. [Recent](https://github.com/uktrade/data-hub-frontend/blob/develop/assets/javascripts/modules/messages.js#L7) modules [have been](https://github.com/uktrade/data-hub-frontend/blob/develop/assets/javascripts/modules/conditional-subfields.js#L56) using the idea of `render()` to set initial state of module and common methods like `cacheEls` and `bindEvents()`.

If we structure modules in a similar way it makes them easier to come back and understand where things are happening. 

I think this is the first that allows configuration to be customised by data attributes but the suggestion was to keep the module simple with default settings that can be passed into the `init()` method to override and the method that sets each instance up should handle where it gets those from to pass to the module.

teneightfive: >after 16 days of review

To be fair, the PR was parked for a while because of a known bug elsewhere so it hasn't been under extensive review for this whole period, also not including weekends and bank holidays.

PR #454, opened 2017-08-14
Add remaining OMIS order fields to work order view

teneightfive:
This adds some of the legacy fields to the work order view.

The work order view now has logic to display fallback text if a value
is falsey. If not fallback text is supplied it will not render the
row if the value is falsey.

### DIFF
--- a/src/apps/omis/apps/view/views/work-order.njk
+++ b/src/apps/omis/apps/view/views/work-order.njk
@@ -11,10 +11,12 @@
       {% endif %}
     </caption>
     {% for item in props.items %}
-      <tr>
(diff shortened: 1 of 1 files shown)


src/apps/omis/apps/view/views/work-order.njk
```diff
@@ -69,14 +73,33 @@
     heading: 'Work description',
     editUrl: 'edit/work-description',
     items: [{
+      label: 'Delivery date',
+      value: values.delivery_date,
+      fallbackText: 'Not set'
```
feedmypixel: rather than passing this into the macro can it be done here? `value: values.delivery_date || 'Not set'`

PR #457, opened 2017-08-15
Update OMIS work order view 

teneightfive:
This updates the methods used to display some new and legacy fields.

It reduces an array of objects to an array of strings as well as
adding date formatting for displaying dates in a view.

### DIFF
--- a/config/nunjucks/filters.js
+++ b/config/nunjucks/filters.js
@@ -1,5 +1,5 @@
 const nunjucks = require('nunjucks')
-
+const dateFns = require('date-fns')
 const {
   assign,
   concat,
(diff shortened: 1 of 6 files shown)


config/nunjucks/filters.js
```diff
@@ -92,6 +94,14 @@ const filters = {
     return number.toLocaleString(locales)
   },
 
+  formatDate: (value, format = longDateFormat) => {
+    const parsedDate = dateFns.parse(value)
+
+    if (!dateFns.isValid(parsedDate)) { return }
```
tyom: Should return `value`.

tyom:
Perhaps remove the [old `dateFormat`](https://github.com/uktrade/data-hub-frontend/blob/develop/config/nunjucks/trade-elements-filters.js#L209-L225) filter from Trade Elements.

PR #458, opened 2017-08-15
Investment Project interactions form

feedmypixel:
- fix error with date input
- add service type and service provider inputs to investment projects interaction form (this is because the backend requires them)
- refactor form to use new form components
- update tests

add-interaction-investment

### DIFF
--- a/src/apps/interactions/services/formatting.js
+++ b/src/apps/interactions/services/formatting.js
@@ -102,11 +102,11 @@ function transformFromApi (body) {
     }
   })
 
-  const date = new Date(body['date'])
+  const date = new Date(body.date)
   if (date) {
-    formatted['date_year'] = date.getFullYear().toString()
-    formatted['date_month'] = format(date, 'MM')
-    formatted['date_day'] = format(date, 'DD')
+    formatted.date_year = date.getFullYear().toString()
+    formatted.date_month = format(date, 'MM')
+    formatted.date_day = format(date, 'DD')
   }
 
   return Object.assign({}, body, formatted)
@@ -137,7 +137,7 @@ function transformToApi (body) {
   // TODO please see JIRA-469 this needs to be converted to a Date not DateTime in the BE. When done the DateTime creation can be removed
   formatted['date'] = new Date(
     body.date_year,
-    body.date_month - 1,
+    body.date_month - 1, // month is zero based
     body.date_day,
   ).toISOString()
 
--- a/src/apps/interactions/services/formatting.js
+++ b/src/apps/interactions/services/formatting.js
@@ -1,3 +1,4 @@
+const format = require('date-fns/format')
 const { formatLongDate, formatMediumDate } = require('../../../../common/date')
 const { newlineToBr, getContactLink } = require('../../../lib/text-formatting')
 const { getPropertyName } = require('../../../lib/property-helpers')
@@ -101,6 +102,13 @@ function transformFromApi (body) {
     }
   })
 
+  const date = new Date(body['date'])
+  if (date) {
+    formatted['date_year'] = date.getFullYear().toString()
+    formatted['date_month'] = format(date, 'MM')
+    formatted['date_day'] = format(date, 'DD')
+  }
+
   return Object.assign({}, body, formatted)
 }
 
@@ -126,7 +134,12 @@ function transformToApi (body) {
     return value
   })
 
-  formatted['date'] = new Date(body['date_year'], (body['date_month'] - 1), body['date_day']).toISOString()
+  // TODO please see JIRA-469 this needs to be converted to a Date not DateTime in the BE. When done the DateTime creation can be removed
+  formatted['date'] = new Date(
+    body.date_year,
+    body.date_month - 1,
+    body.date_day,
+  ).toISOString()
 
   return Object.assign({}, body, formatted)
 }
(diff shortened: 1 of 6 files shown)


feedmypixel:
@venureddy95 FYI - fix for date validation errors

src/apps/investment-projects/views/interactions/_form.njk
```diff
@@ -1,42 +1,84 @@
...
+  <div class="c-form-group">
+    {{ MultipleChoiceField({
+      name: 'contact',
+      label: form.labels.contact,
+      optional: form.errors.messages.contact,
+      value: form.state.contact,
+      options: form.options.contacts,
+      initialOption: '-- Select contact --'
+    }) }}
+
```
ztolley: This uses a drop down which is different from company interactions I think and all the adviser dropdowns are broken

feedmypixel: @ztolley for the moment until autocomplete is fixed all autocompletes are being made dropdown's. When we can we can sort autocomplete out and then apply where needed?

feedmypixel:
@jim68000 @reupen FYI the `service` and `dit_team` inputs have been added to this form as the BE requires them for the form to be submitted. As spoke about this needs revisiting as it appears the user may not need to enter them

src/apps/investment-projects/views/interactions/_form.njk
ztolley: Talk to Jim to get the code he updated for Ajax based adviser autocomplete and use that

ztolley: Jim seems to think he fixed it for company interaction

ztolley: And drop downs are WAY more broken right making it impossible to add or edit stuff

feedmypixel: @ztolley @tyom is going to own this PR with me being away. The autocomplete discussion needs to happen elsewhere and the work needs to be played sooner rather than later. 

>And drop downs are WAY more broken right making it impossible to add or edit stuff

I wouldn't agree with that the dropdown is usable and a known pattern to users. The autocomplete has many problems, affordance, usability, functionality to name a few

ztolley: The comment about dropdowns for advisers isn't saying dropdowns don't work but right now ALL dropdowns for advisers don't work as we only fetch the first 100, so if you edit a record with an adviser it saves it back as null because the value wasn't in the option list

The company interaction done by jim uses Ajax autocomplete

This is an issue not with this ticket but site wide and it's a showstopper

src/apps/interactions/services/formatting.js
```diff
@@ -101,6 +102,13 @@ function transformFromApi (body) {
     }
   })
 
+  const date = new Date(body['date'])
```
teneightfive: `const date = new Date(body.date)`?

src/apps/interactions/services/formatting.js
```diff
@@ -101,6 +102,13 @@ function transformFromApi (body) {
...
+  if (date) {
+    formatted['date_year'] = date.getFullYear().toString()
```
teneightfive: Could use dot notation for all of these?

src/apps/interactions/services/formatting.js
```diff
@@ -126,7 +134,12 @@ function transformToApi (body) {
     return value
   })
 
-  formatted['date'] = new Date(body['date_year'], (body['date_month'] - 1), body['date_day']).toISOString()
+  // TODO please see JIRA-469 this needs to be converted to a Date not DateTime in the BE. When done the DateTime creation can be removed
+  formatted['date'] = new Date(
+    body.date_year,
+    body.date_month - 1,
```
teneightfive: Maybe add a comment just to remind people why you need to `- 1`.

src/apps/investment-projects/views/interactions/_form.njk
```diff
@@ -1,42 +1,79 @@
-<form method="post">
-  <input type="hidden" name="investment_project" value="{{ investmentStatus.id }}">
-  <input type="hidden" name="_csrf" value="{{ csrfToken }}">
+{% call Form(form | assign({
+  buttonText: 'Save',
```
teneightfive: Is `Save` not the default button text?

src/apps/investment-projects/views/interactions/_form.njk
```diff
@@ -1,42 +1,79 @@
...
-  {{ dropdown('interaction_type',
-    label=form.labels['interaction_type'],
-    emptyLabel='Select interaction',
-    value=form.state['interaction_type'],
-    options=form.options.interactionTypes)
-  }}
+  {{ TextField({
+    name: 'subject',
+    label: form.labels.subject,
+    placeholder: 'e.g a coffee and chat',
```
teneightfive: Do we need this placeholder? Feels a bit informal too...

src/apps/investment-projects/views/interactions/_form.njk
feedmypixel: nope its `Submit` https://github.com/uktrade/data-hub-frontend/blob/develop/src/templates/_macros/form.njk#L48

teneightfive: Ah yes. I wonder where we even use `Submit`. Maybe `Save` would be a more logical default...not part of this change though!

PR #459, opened 2017-08-15
Edit work description fields

teneightfive:
This adds work so that the remaining fields associated with the OMIS order can be edited.

👉 [**Demo**](https://datahub-beta2-pr-459.herokuapp.com/omis/084b934e-9bde-4e2b-88d8-5fbeb17de54d) 👈 

## What it looks like

localhost_3001_omis_084b934e-9bde-4e2b-88d8-5fbeb17de54d_edit_work-description

### DIFF
--- a/src/apps/omis/apps/edit/controllers/work-description.js
+++ b/src/apps/omis/apps/edit/controllers/work-description.js
@@ -1,0 +1,31 @@
+const { flatten } = require('lodash')
+const chrono = require('chrono-node')
+const dateFns = require('date-fns')
+
+const { EditController } = require('../../../controllers')
+const { transformObjectToOption } = require('../../../../transformers')
+const metadataRepo = require('../../../../../lib/metadata')
+
+class EditWorkDescriptionController extends EditController {
+  configure (req, res, next) {
+    req.form.options.fields.sector.options = metadataRepo.sectorOptions.map(transformObjectToOption)
+    req.form.options.fields.service_types.options = metadataRepo.orderServiceTypesOptions.map(transformObjectToOption)
+    next()
+  }
+
+  process (req, res, next) {
+    const deliveryDateStr = req.body.delivery_date
+    const parsedDeliveryDate = chrono.en_GB.parseDate(deliveryDateStr)
+
+    if (parsedDeliveryDate) {
+      req.form.values.delivery_date = dateFns.format(parsedDeliveryDate, 'YYYY-MM-DD')
+    } else {
+      req.form.values.delivery_date = deliveryDateStr
+    }
+
+    req.form.values.service_types = flatten([req.form.values.service_types])
+    next()
+  }
+}
+
+module.exports = EditWorkDescriptionController
--- a/src/apps/omis/apps/edit/controllers/work-description.js
+++ b/src/apps/omis/apps/edit/controllers/work-description.js
@@ -1,4 +1,4 @@
-const { flatten } = require('lodash')
+const { filter, find, flatten } = require('lodash')
 const chrono = require('chrono-node')
 const dateFns = require('date-fns')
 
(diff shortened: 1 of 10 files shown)


src/apps/omis/apps/edit/controllers/work-description.js
```diff
@@ -0,0 +1,57 @@
+const { filter, find, flatten } = require('lodash')
+const chrono = require('chrono-node')
+const dateFns = require('date-fns')
+
+const { EditController } = require('../../../controllers')
+const { transformObjectToOption } = require('../../../../transformers')
+const metadataRepo = require('../../../../../lib/metadata')
+
+class EditWorkDescriptionController extends EditController {
```
tyom: Deep inheritance will lead to problems in future. Mark my words.

src/apps/omis/apps/edit/controllers/work-description.js
```diff
@@ -0,0 +1,57 @@
...
+  configure (req, res, next) {
+    const filteredServiceTypes = filter(metadataRepo.orderServiceTypesOptions, (service) => {
+      if (!service.disabled_on) { return true }
+
+      if (Date.parse(service.disabled_on) > Date.parse(res.locals.order.created_on)) {
```
tyom: Might be worth storing these into variables and use those in conditions to improve readability and explain what's going on.

src/apps/omis/apps/edit/controllers/work-description.js
teneightfive: I know 😞 

Once you're new method is a bit further along I'll be able to swap this out for that though so not to worry 😬

src/apps/omis/controllers/edit.js
```diff
@@ -26,18 +39,49 @@ class EditController extends FormController {
...
+    let combinedValues
 
     delete values.errorValues
     delete values.errors
 
-    forEach(req.form.options.fields, (value, key) => {
-      values[key] = get(res.locals, `order.${key}.id`)
+    // set current values for order
+    forEach(req.form.options.fields, (fieldOptions, key) => {
+      if (!values[key]) {
```
tyom: Do you need to mutate `values`? Wouldn't using `map` to create a new array be better instead?

src/apps/omis/controllers/edit.js
```diff
@@ -26,18 +39,49 @@ class EditController extends FormController {
...
+        }
+
+        values[key] = newValue
+      }
     })
 
+    // get error values for step fields
     errorValues = pick(errorValues, Object.keys(req.form.options.fields))
-    errorValues = pickBy(errorValues, (e, k) => !errors || !errors[k] || !errors[k].url || errors[k].url === req.path)
+    errorValues = pickBy(errorValues, (e, key) => {
```
tyom: `error`?

Also,  seems like `errors` shouldn't be in `pickBy` predicate.

Could you express it as ```get(errors, `${key}.url`) === req.path)```?

src/apps/omis/controllers/edit.js
teneightfive: I think the problem was that map doesn't give you the value and the key. This loop is actually going through each field, getting the value of it from the order if it exists and then setting that on the `values`.

Effectively getting the default value from the server. If you see a neater/more readable way then shout. I did try and think of a better way but I tried a few until I got it working as expected.

tyom: Have you tried Lodash `map`?

teneightfive: Actually, what I might try is create a different object and combine them all using assign...

teneightfive: >Have you tried Lodash `map`? It can iterate arrays as well as objects.

That's what I ended up using.

Hopefully what its doing now is easier to read. It creates separate objects then merges them using assign.

PR #460, opened 2017-08-16
Use investment projects collection on company/investments page

tyom:
### Before
before

### After
localhost_3001_companies_0fb3379c-341c-4da4-b825-bf8d47b26baa_investments

### DIFF
--- a/src/apps/companies/views/investments.njk
+++ b/src/apps/companies/views/investments.njk
@@ -1,37 +1,19 @@
 {% extends "./_layout-view.njk" %}
 
+{% set addProjectButton %}
+  <a class="button button-secondary" href="/investment-projects/create/{{ company.id }}">Add investment project</a>
+{% endset %}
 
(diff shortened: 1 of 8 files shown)


src/apps/companies/views/investments.njk
```diff
@@ -1,37 +1,19 @@
...
-            <dd class="metadata-list__value">{{ project.sector.name }}</dd>
-          </dl>
-        </li>
-      {% endfor %}
-    </ol>
-  {% endif %}
+  <h2 class="heading-medium">Investment projects</h2>
+
+  {{
+    Results(results | assign({
```
teneightfive: `{} | assign(results, {...})` so that you don't mutate `results`?

tyom: That syntax is a bit odd. Maybe I'll use the new `assignCopy` from #452.

PR #461, opened 2017-08-16
Firefox issue with highlighter

tyom:
Firefox has issues with applying negative margin to inline elements. It forces them to drop to the new line. This fix means the small padding around the highlighter had to be removed.

### Before
image

### After
image

### DIFF
--- a/config/nunjucks/trade-elements-filters.js
+++ b/config/nunjucks/trade-elements-filters.js
@@ -51,17 +51,6 @@ filter.toHyphenated = function toHyphenated (string) {
   return string.trim().toLowerCase().replace(/\s+/g, '-')
 }
 
-/**
- * Highlights a phrase in a source piece of text
- * @param  {String} text    The original text to be updated
- * @param  {String} phrase  The phrase to highlight in the original text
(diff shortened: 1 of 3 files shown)


config/nunjucks/trade-elements-filters.js
```diff
@@ -51,17 +51,6 @@ filter.toHyphenated = function toHyphenated (string) {
   return string.trim().toLowerCase().replace(/\s+/g, '-')
 }
 
-/**
- * Highlights a phrase in a source piece of text
- * @param  {String} text    The original text to be updated
- * @param  {String} phrase  The phrase to highlight in the original text
- * @return {String}     The resulting text with highlights using <strong>
- */
-filter.highlight = function highlight (text, phrase) {
```
jim68000: PR looks good by I'm puzzled by this deletion. Is this just some leftover unused code?

tyom: Yes. This is left over from Trade Elements. We use new filter located in `filters.js`.

jim68000: Cool. Approved

PR #466, opened 2017-08-16
Only map and applyAsMacro when props.children is an array

tyom:
In some cases the macro passed to props.children is already pre-rendered so we can
just output it directly

### DIFF
--- a/src/templates/_macros/form.njk
+++ b/src/templates/_macros/form.njk
@@ -465,7 +465,7 @@
   {% if props.type in ['checkbox', 'radio'] %}
     {% set modifier = props.modifier | concat('') | reverse | join(' c-multiple-choice--') if props.modifier %}
     {% set options = props.options() if props.options | isFunction else props.options %}
-    {% set fieldChildren = props.children | map(callAsMacro) if props.children %}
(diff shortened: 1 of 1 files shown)


src/templates/_macros/form.njk
```diff
@@ -465,7 +465,7 @@
   {% if props.type in ['checkbox', 'radio'] %}
     {% set modifier = props.modifier | concat('') | reverse | join(' c-multiple-choice--') if props.modifier %}
     {% set options = props.options() if props.options | isFunction else props.options %}
-    {% set fieldChildren = props.children | map(callAsMacro) if props.children %}
+    {% set fieldChildren = props.children | map(callAsMacro) if props.children | isArray else props.children %}
```
teneightfive: Shouldn't `children` always be an array? The name implies it is a plural.

What are the other use cases?

tyom: Because we pass props down to children one of the parents (in this case `FormGroup`) might have already processed it as macro.

We should potentially take another look at when the macros should be applied, but this fixes the immediate issue.

teneightfive: Okay, maybe add an issue so we don't completely forget.

tyom: #473.

PR #467, opened 2017-08-16
Limit pagination results to 10,000 items paginated

tyom:
As ElasticSearch is only limited to provide up to 10,000 results and because there is not much value for user in showing pagination beyond 10,000 items we should limit the number of results that can be pagination to 10,000.

### DIFF
--- a/src/lib/pagination.js
+++ b/src/lib/pagination.js
@@ -1,5 +1,6 @@
 const { range, take, get, omitBy } = require('lodash')
 const queryString = require('query-string')
+const config = require('../../config')
 
 function getPageLink (page, query = {}) {
   const newQuery = Object.assign({}, query, {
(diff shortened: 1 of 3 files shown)


src/lib/pagination.js
```diff
@@ -53,18 +53,20 @@ function truncatePages (pagination, blockSize) {
 }
 
 function buildPagination (query = {}, results, truncate = 4) {
+  const RESULTS_HARD_LIMIT = 10000
```
teneightfive: Could this const be set at the top of the pagination library rather than within this function if it can't be overridden?

If we do need to change it, would it have any benefit to set it in our app config with 10000 as the default value?

marcofucci: +1

PR #469, opened 2017-08-17
E2e/investment stages

venureddy95:
Minor changes to feature files. Cleaning up before the new Sprint starts today.

### DIFF
--- a/test/acceptance/features/step_definitions/investment-stages-step-definitions.js
+++ b/test/acceptance/features/step_definitions/investment-stages-step-definitions.js
@@ -18,7 +18,6 @@ defineSupportCode(({ Given, Then, When }) => {
   let safeguardedJobsValue
   let projectAssurance
   let projectManager
-  let projectManager1
 
   Given(/^I am an authenticated Client relationship manager user on Data Hub website$/, async () => {
     await client.page.Login().navigate().authenticate()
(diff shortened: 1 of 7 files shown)


test/acceptance/features/step_definitions/investment-stages-step-definitions.js
```diff
@@ -0,0 +1,373 @@
...
+const { client } = require('nightwatch-cucumber')
+const { defineSupportCode } = require('cucumber')
+
+defineSupportCode(({ Given, Then, When }) => {
+  const Stages = client.page.InvestmentStages()
+  const Contact = client.page.Contact()
+  const Company = client.page.Company()
+  const Investment = client.page.InvestmentProject()
+  const foreignCompanyName = 'Lambda plc'
+  let projectName
```
tyom: Do you need these variables in this context? Looks like some of those variables are specific to the certain blocks. And if you change variables from one function call to another it produces side-effects which may lead to bugs.

test/acceptance/page_objects/InvestmentStages.js
```diff
@@ -0,0 +1,155 @@
...
+    editValueButton: {
+      selector: "//a[@class='button-secondary'][@href='edit-value']",
+      locateStrategy: 'xpath',
+    },
+    editPMButton: {
+      selector: "//a[@class='button button-secondary'][@href='edit-project-management']",
+      locateStrategy: 'xpath',
+    },
+    editSummaryButton: {
+      selector: "//a[@href='edit-details']",
```
tyom: I feel like these would be more readable using CSS selectors.

test/acceptance/features/step_definitions/investment-stages-step-definitions.js
venureddy95: Tyom, I need the variables to read the values from the page and verify whether they are correctly displayed which I entered while adding requirements/value/PM fields. Lets discuss if there is a better way once I am in office next week.

PR #470, opened 2017-08-17
Rename results macros to collection

tyom:
Rename macro to align with what we call the concept internally. Fix a few minor bugs and simplify the macro props interface.


src/apps/investment-projects/transformers.js
```diff
@@ -14,41 +15,55 @@ function transformInvestmentProjectToListItem ({
...
+
+  if (stage) {
+    meta.push({
+      name: 'stage',
+      label: 'Stage',
+      value: stage,
+      type: 'badge',
+    })
+  }
+  if (stage) {
```
teneightfive: `if (investment_type) { ... }`?

PR #471, opened 2017-08-17
Edit assignee time allocated

teneightfive:
This adds support for editing the time assignees spend on an order.

It creates the step, controller and custom view needed to edit this data.

👉 [**DEMO**](https://datahub-beta2-pr-471.herokuapp.com/omis/ebad867d-2527-4439-b5fb-d39fb7095392/edit/assignee-time) 👈

## What it looks like

localhost_3001_omis_5325ad09-0dd4-480e-94be-9a4a3cde0eb9_edit_assignee-time

### DIFF
--- a/config/nunjucks/filters.js
+++ b/config/nunjucks/filters.js
@@ -113,12 +113,16 @@ const filters = {
     return dateFns.format(parsedDate, format)
   },
 
-  formatDuration: (value, format = longDateFormat, measurement = 'minutes') => {
+  humanizeDuration: (value, measurement = 'minutes') => {
     const duration = moment.duration(value, measurement)
-    const hoursSuffix = pluralise('hour', duration.hours())
-    const minutesSuffix = pluralise('minute', duration.minutes())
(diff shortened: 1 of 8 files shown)


config/nunjucks/filters.js
```diff
@@ -1,6 +1,8 @@
 const nunjucks = require('nunjucks')
 const humanizeDuration = require('humanize-duration')
 const dateFns = require('date-fns')
+const moment = require('moment')
```
ztolley: Are we using 2 different date libraries?

teneightfive: Unfortunately so :(

Moment was added here for duration support. #475 has just been updated to use [moment's duration](https://momentjs.com/docs/#/durations/) support with the addition of a [plugin](https://github.com/jsmreese/moment-duration-format).

I think we started favouring [dateFns](https://date-fns.org/) over moment which is why it's already being used in this file but when scenarios like this crop up and it can't do what moment does maybe we should refer our preferred choice.

I have created an issue so it can be tracked: #497

PR #475, opened 2017-08-18
View assignee time allocated

teneightfive:
This change adds the time an assignee has allocated to an order to the work order view.

👉 [**DEMO**](https://datahub-beta2-pr-475.herokuapp.com/omis/ebad867d-2527-4439-b5fb-d39fb7095392/work-order) 👈

## What it looks like

localhost_3001_omis_5325ad09-0dd4-480e-94be-9a4a3cde0eb9_work-order 1

### DIFF
--- a/config/nunjucks/filters.js
+++ b/config/nunjucks/filters.js
@@ -1,4 +1,6 @@
 const nunjucks = require('nunjucks')
+const moment = require('moment')
+require('moment-duration-format')
 const dateFns = require('date-fns')
 const {
   assign,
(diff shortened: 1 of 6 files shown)


config/nunjucks/filters.js
```diff
@@ -110,6 +111,10 @@ const filters = {
     return dateFns.format(parsedDate, format)
   },
 
+  humaniseMinutes: (value, format = longDateFormat) => {
```
ztolley: Humanise and Humanize. maybe we use the US spelling to be consistent.

Also doesn't moment already contain this and the functions in humanize-duration ?

teneightfive: You're right. For some reason I missed it when I was looking for ways to support this. Will update.

config/nunjucks/filters.js
```diff
@@ -1,5 +1,7 @@
 const nunjucks = require('nunjucks')
-const humanizeDuration = require('humanize-duration')
+const moment = require('moment')
+require('moment-duration-format')
+// const humanizeDuration = require('humanize-duration')
```
ztolley: why comment it and not just delete?

teneightfive: My bad.

PR #477, opened 2017-08-18
Fix issue with Mutiple choice macros

ztolley:
Fix issue with Mutiple choice macros and form group macros not supporting groupClass due to mutating props

### DIFF
--- a/src/templates/_macros/form.njk
+++ b/src/templates/_macros/form.njk
@@ -201,6 +201,8 @@
   {% set isLabelHidden = props.isLabelHidden | default(false) %}
   {% set isConditional = props.condition %}
   {% set modifier = props.modifier | concat('') | reverse | join(' c-form-group--') if props.modifier %}
+
+
   {% set children = caller() if caller else renderAsMacro(props.children) %}
 
   {% if props.label and props.name -%}
@@ -211,8 +213,7 @@
         {{ 'has-error' if props.error and not isLabelHidden }}
         {{ 'js-ConditionalSubfield' if isConditional }}
         {{ modifier }}
-        {{ props.class if props.class }}
-      "
+        {{ props.class if props.class }}"
       {% if isConditional %}
         data-controlled-by="{{ props.condition.name }}"
         data-control-value="{{ props.condition.value }}"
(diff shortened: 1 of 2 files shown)


src/templates/_macros/form.njk
```diff
@@ -211,8 +212,7 @@
         {{ 'has-error' if props.error }}
         {{ 'js-ConditionalSubfield' if isConditional }}
         {{ modifier }}
-        {{ props.class if props.class }}
```
teneightfive: Not sure I fully understand this change. It seems you're just removing the `if props.class`?

Do you have an example of the error before and after? Might help explain it better.

teneightfive: It also looks like the docs are out of date as they don't mention `props.class`.

ztolley: Before, if you passed a value to props.class to the macro it ignored it. Something about the line

`{% set children = caller() if caller else renderAsMacro(props.children) %}`

stripped the props object of the class property.

The change here removes the use of props.class and instead just inserts class, which is taken from props while it is still there.

teneightfive: Hmm, that doesn't feel like it's fixing the real problem.

Maybe look at renderAsMacro and see if its mutating something which is removing the `class` property from `props`?

ztolley: its not render as macro then but one of the macros it calls in turn .

We need more tests as the macros become more complex.

src/templates/_macros/form.njk
```diff
@@ -201,6 +201,8 @@
   {% set isLabelHidden = props.isLabelHidden | default(false) %}
   {% set isConditional = props.condition %}
   {% set modifier = props.modifier | concat('') | reverse | join(' c-form-group--') if props.modifier %}
+
```
teneightfive: Maybe see if you can remove these noise changes?

src/templates/_macros/form.njk
```diff
@@ -211,8 +213,7 @@
         {{ 'has-error' if props.error and not isLabelHidden }}
         {{ 'js-ConditionalSubfield' if isConditional }}
         {{ modifier }}
-        {{ props.class if props.class }}
-      "
+        {{ props.class if props.class }}"
```
teneightfive: Plus the noise here? Revert back to original version with `"` on next line.

test/unit/macros/form.test.js
```diff
@@ -384,4 +384,21 @@ describe('Nunjucks form macros', () => {
       })
     })
   })
+
+  describe('Formgroup component', () => {
+    it('should allow a class to be specified for use in the form-grouo wrapper', () => {
```
teneightfive: `grouo` -> `group`

PR #478, opened 2017-08-18
Set lead adviser

teneightfive:
This change adds the ability to set who the lead adviser on an order is.

It adds an extra post handler to the edit sub-app which handles unsetting the previous lead and setting the new lead.

**Note:** Needs to be merged and rebased after #475 

👉 [**DEMO**](https://datahub-beta2-pr-478.herokuapp.com/omis/ebad867d-2527-4439-b5fb-d39fb7095392/work-order) 👈

## What it looks like

localhost_3001_omis_5325ad09-0dd4-480e-94be-9a4a3cde0eb9_work-order

### DIFF
--- a/assets/stylesheets/_deprecated/components/_button.scss
+++ b/assets/stylesheets/_deprecated/components/_button.scss
@@ -23,6 +23,7 @@
   margin: 30px 0;
 }
 
+// modifiers
 .button--link {
   background: transparent;
   border: 0;
@@ -44,4 +45,8 @@
   &:hover {
     color: $link-hover-colour;
   }
+}
+
+.button--compact {
+  padding: 0;
 }
--- a/assets/stylesheets/components/_answers-summary.scss
+++ b/assets/stylesheets/components/_answers-summary.scss
@@ -71,7 +71,7 @@
     text-align: right;
   }
 
-  .button {
+  .button--secondary {
     @include core-font(16);
     margin-bottom: $default-spacing-unit / 2;
 
--- a/assets/stylesheets/components/_badge.scss
+++ b/assets/stylesheets/components/_badge.scss
@@ -37,6 +37,10 @@
   text-decoration: none;
   font-weight: 600;
 
+  .c-answers-summary & {
+    font-size: 14px;
+  }
+
   &:link,
   &:visited {
     background-color: $white;
--- a/src/apps/omis/apps/edit/controllers/edit-lead-assignee.js
+++ b/src/apps/omis/apps/edit/controllers/edit-lead-assignee.js
@@ -1,0 +1,32 @@
+const { find, get } = require('lodash')
+
+const { Order } = require('../../../models')
+
+async function editLeadAssignee (req, res, next) {
+  const adviserId = req.body.adviserId
+  const orderId = req.body.orderId
+  const returnUrl = req.body.returnUrl || req.header('Referer')
+
+  if (!adviserId || !orderId) {
+    return res.redirect(returnUrl)
+  }
+
+  try {
+    const allAssignees = await Order.getAssignees(req.session.token, orderId)
+    const assignees = allAssignees.map(assignee => {
+      return Object.assign(assignee, {
+        is_lead: assignee.adviser.id === adviserId,
+      })
+    })
+    const leadAdviser = find(assignees, { adviser: { id: adviserId } })
+
+    await Order.saveAssignees(req.session.token, orderId, assignees)
+
+    req.flash('success', `Lead post adviser set to ${get(leadAdviser, 'adviser.name')}`)
+    res.redirect(returnUrl)
+  } catch (error) {
+    next(error)
+  }
+}
+
+module.exports = editLeadAssignee
--- a/src/apps/omis/apps/view/views/work-order.njk
+++ b/src/apps/omis/apps/view/views/work-order.njk
@@ -91,8 +91,25 @@
       {% for assignee in values.assignees %}
         <tr>
           <th class="c-answers-summary__title" scope="row">
-            {{ assignee.adviser.name }}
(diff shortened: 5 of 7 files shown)


assets/stylesheets/components/_answers-summary.scss
```diff
@@ -71,7 +71,7 @@
     text-align: right;
   }
 
-  .button {
+  .button--secondary {
```
ztolley: could this not be a generic button style, what is unique about it in this context?

assets/stylesheets/_deprecated/components/_button.scss
```diff
@@ -23,6 +23,7 @@
   margin: 30px 0;
 }
 
+// modifiers
```
ztolley: doesn't the -- explicitly demonstrate it's a modifier?

assets/stylesheets/components/_badge.scss
```diff
@@ -32,6 +32,7 @@
 }
 
 .c-badge {
+  @include core-font(14);
```
ztolley: I thought we had a modifier for badges for a small variation?

src/apps/omis/apps/edit/controllers/edit-lead-assignee.js
```diff
@@ -0,0 +1,32 @@
+const { find, get } = require('lodash')
+
+const { Order } = require('../../../models')
+
+async function editLeadAssignee (req, res, next) {
+  const adviserId = req.body.adviserId
```
ztolley: I guess you could

```
const { orderId, returnUrl } = req.body

if ...
  return res.redirect( req.body.returnUrl || req.header('Referer'))
```

Then again nah

src/apps/omis/apps/edit/controllers/edit-lead-assignee.js
```diff
@@ -0,0 +1,32 @@
...
+  const orderId = req.body.orderId
+  const returnUrl = req.body.returnUrl || req.header('Referer')
+
+  if (!adviserId || !orderId) {
+    return res.redirect(returnUrl)
+  }
+
+  try {
+    const allAssignees = await Order.getAssignees(req.session.token, orderId)
```
ztolley: I miss the old days when you use a capital for a class and a lower case for an instance

src/apps/omis/apps/view/views/work-order.njk
```diff
@@ -83,8 +83,25 @@
       <tbody>
         <tr>
           <th class="c-answers-summary__title" scope="row">
-            {{ assignee.adviser.name }}
+            {{ assignee.adviser.name }} {{ '(you)' if assignee.adviser.id === user.id }}
```
ztolley: so if would read     'Fred Smith you'?

src/apps/omis/apps/view/views/work-order.njk
```diff
@@ -95,7 +112,7 @@
     {% if values.estimatedTimeSum and values.assignees.length %}
     <tfoot>
       <tr>
-        <td class="c-answers-summary__footer" colspan="2">
+        <td class="c-answers-summary__footer" colspan="3">
```
ztolley: Tables?

assets/stylesheets/components/_answers-summary.scss
teneightfive: I only want the margin applied to secondary style buttons. I have an instance which uses _link style_ buttons where I don't want this spacing applied.

assets/stylesheets/components/_badge.scss
teneightfive: Well actually the only place we use it is in entities at the moment where it is set to `14px`. I was trying to set the base to `14` and if we need it scaled in other places it can be given a modifier.

Maybe if we do set it to `14px` everywhere I could remove the declaration from the entities or I set it to `14px` within the component that contains it but I prefer to amend the style of a component using modifiers.

Maybe one to check with @tyom as he wrote this component originally.

src/apps/omis/apps/edit/controllers/edit-lead-assignee.js
teneightfive: Could possibly do `const { orderId, adviserId } = req.body` but `returnUrl` uses a fallback if it's not set.

src/apps/omis/apps/edit/controllers/edit-lead-assignee.js
teneightfive: `Order` isn't a class actually so maybe it should be lowercase actually. Different change though...

It might be worth having the discussion about _models_ though. Models doesn't feel like its quite right but I think the idea of having a standard set of methods that exist on entities and a set pattern to create custom ones will be useful. Possibly something similar to [SailsJS Waterline](http://sailsjs.com/documentation/reference/waterline-orm/models).

src/apps/omis/apps/view/views/work-order.njk
teneightfive: If you were _John Rogers_ in this instance you would see this:

image

Very similar to the pattern Slack uses:

image

src/apps/omis/apps/view/views/work-order.njk
teneightfive: I know, but in this instance it is tabular data. A list of advisers with time associated to each one and a total time which is displayed in this footer.

assets/stylesheets/_deprecated/components/_button.scss
teneightfive: I suppose it does but it's a pattern we've used in a few places to make the separation clearer when scanning the file and for when people add new modifiers. Keeps things together I suppose.

assets/stylesheets/components/_badge.scss
tyom: I avoided setting the font size on badge originally because I wanted it to inherit the font-size from the text around the badge.

PR #481, opened 2017-08-20
Content updates to create journey and work order

teneightfive:
This change adds some content amends to the creation journey and enhances the work order by
adding links to entities.

👉 [**DEMO (Create journey)**](https://datahub-beta2-pr-481.herokuapp.com/omis/create/dcdabbc9-1781-e411-8955-e4115bead28a) 👈

👉 [**DEMO (Work order)**](https://datahub-beta2-pr-481.herokuapp.com/omis/ebad867d-2527-4439-b5fb-d39fb7095392/work-order) 👈

### Before

localhost_3001_omis_5325ad09-0dd4-480e-94be-9a4a3cde0eb9_work-order 2

### After

localhost_3001_omis_5325ad09-0dd4-480e-94be-9a4a3cde0eb9_work-order 1

### Before

localhost_3001_omis_create_a73efeba-8499-11e6-ae22-56b6b6499611_client-details

localhost_3001_omis_create_a73efeba-8499-11e6-ae22-56b6b6499611_market

localhost_3001_omis_create_a73efeba-8499-11e6-ae22-56b6b6499611_confirm

### After

localhost_3001_omis_create_a73efeba-8499-11e6-ae22-56b6b6499611_client-details_edit

localhost_3001_omis_create_a73efeba-8499-11e6-ae22-56b6b6499611_market_edit

localhost_3001_omis_create_a73efeba-8499-11e6-ae22-56b6b6499611_confirm 1

### DIFF
--- a/src/apps/omis/apps/create/views/client-details.njk
+++ b/src/apps/omis/apps/create/views/client-details.njk
@@ -1,0 +1,22 @@
+{% extends "_layouts/form-wizard-step.njk" %}
+
+{% block fields %}
+  <h2 class="heading-small">Company</h2>
+  <div>
+    {{ company.name }}
+    <br>
+    {{ [
+      company.registered_address_1,
+      company.registered_address_2,
+      company.registered_address_county,
+      company.registered_address_town,
+      company.registered_address_postcode
+    ] | removeNilAndEmpty | join(", ") }}
+  </div>
+
+  {{ super( )}}
+
+  <p class="infostrip">
+    If the contact you are looking for is not listed you can <a href="/contacts/create?company={{ company.id }}">create a new one</a>.
+  </p>
+{% endblock %}
--- a/src/apps/omis/apps/view/views/work-order.njk
+++ b/src/apps/omis/apps/view/views/work-order.njk
@@ -25,7 +25,7 @@
           {% if item.value %}
             <tr>
               <th class="c-answers-summary__title" scope="row">{{ item.label }}</th>
-              <td class="c-answers-summary__content">{{ item.value }}</td>
+              <td class="c-answers-summary__content">{{ item.value | safe }}</td>
             </tr>
           {% endif %}
         {% endif %}
@@ -39,6 +39,12 @@
 {% endblock %}
 
 {% block main_grid_right_column %}
+  {% set companyLink %}
+    <a href="/viewcompanyresult/{{ values.company.id }}">{{ values.company.name }}</a>
+  {% endset %}
+  {% set contactLink %}
+    <a href="/contacts/{{ values.contact.id }}">{{ values.contact.name }}</a>
+  {% endset %}
 
   {{ AnswersSummary({
     heading: 'Client details',
@@ -47,10 +53,10 @@
     }],
     items: [{
       label: 'Company',
-      value: values.company.name
+      value: companyLink
     }, {
       label: 'Contact',
-      value: values.contact.name
+      value: contactLink
     }]
   }) }}
 
--- a/assets/stylesheets/_deprecated/components/_infostrip.scss
+++ b/assets/stylesheets/_deprecated/components/_infostrip.scss
@@ -6,4 +6,8 @@ p.infostrip,
   border-left: 8px solid $border-colour;
   background: $panel-colour;
   padding: 15px 20px;
+
+  * + & {
+    margin-top: $default-spacing-unit * 2;
+  }
 }
--- a/src/apps/omis/apps/create/controllers/client-details.js
+++ b/src/apps/omis/apps/create/controllers/client-details.js
@@ -17,14 +17,9 @@ class ClientDetailsController extends FormController {
     super.configure(req, res, next)
   }
 
-  getValues (req, res, next) {
-    const company = res.locals.company
-
-    super.getValues(req, res, (err, values) => {
-      values.company = company.id
-
-      next(err, values)
-    })
+  process (req, res, next) {
+    req.form.values.company = res.locals.company.id
+    next()
   }
 }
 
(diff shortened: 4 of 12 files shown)


tyom:
Should "Edit time" be there when no advisers are given?

image
---
image

assets/stylesheets/components/form/_form-group.scss
```diff
@@ -27,6 +27,7 @@
 
 .c-form-group__label-text {
   font-weight: 600;
+  margin: 0;
```
tyom: What is this resetting? Where is this originally set?

assets/stylesheets/trumps/_utility.scss
```diff
@@ -46,6 +46,10 @@
   }
 }
 
+.u-block {
```
tyom: Hmm. Why is this useful?

src/apps/omis/apps/create/views/client-details.njk
```diff
@@ -0,0 +1,27 @@
+{% extends "_layouts/form-wizard-step.njk" %}
+
+{% block fields %}
+  <div class="c-form-group">
```
tyom: Should this be a `c-form-group`? It has nothing to do with the form. It looks like you're trying to reuse something the way it was not designed to be used.

assets/stylesheets/components/form/_form-group.scss
```diff
@@ -49,6 +50,11 @@
   }
 }
 
+.c-form-group__value {
```
tyom: This is here because of a specific use case you have to show company data in `form-group`. I question its purpose in CSS component designed for use with forms.

src/apps/omis/apps/create/views/client-details.njk
```diff
@@ -0,0 +1,27 @@
...
+    <h2 class="c-form-group__label-text">Company</h2>
+
+    <p class="c-form-group__value">
+      <span class="u-block">{{ company.name }}</span>
```
tyom: If you must use `p` above, could you use `br` to break the new line instead of adding new utility class?

src/apps/omis/apps/view/views/work-order.njk
```diff
@@ -23,7 +23,7 @@
           {% if item.value %}
             <tr>
               <th class="c-answers-summary__title" scope="row">{{ item.label }}</th>
-              <td class="c-answers-summary__content">{{ item.value }}</td>
+              <td class="c-answers-summary__content">{{ item.value | safe }}</td>
```
tyom: Would be nice to avoid these random `safe` filters.

assets/stylesheets/components/form/_form-group.scss
teneightfive: This resets the [margin](https://github.com/uktrade/data-hub-frontend/blob/49035780e9eb827248df6a4ccabef2cff507387c/assets/stylesheets/elements/_headings.scss#L10) applied to all headings.

assets/stylesheets/components/form/_form-group.scss
teneightfive: Am open to alternative suggestion?

assets/stylesheets/trumps/_utility.scss
teneightfive: Was being used to display an inline element as block to separate company name and address. 

Alternative is to kill the spacing so could have a `u-compact` helper to remove margins/padding?

src/apps/omis/apps/create/views/client-details.njk
teneightfive: Visually it is being represented as a form group. Would could we use as an alternative to show a predefined value that cannot be changed?

src/apps/omis/apps/create/views/client-details.njk
teneightfive: Possibly but a `br` didn't feel like a good solution.

teneightfive:
>Should "Edit time" be there when no advisers are given?

assets/stylesheets/components/form/_form-group.scss
tyom: Should this class be applied to a heading? I feel like we're introducing vague patterns here.

assets/stylesheets/components/form/_form-group.scss
tyom: I think that `20px` font-size is actually different from `10px` we use for body and forms. The margin reset is only needed due to how our current paragraphs work and we said we would like to address typography and spacing in future.

I wonder if this should just be a `div` rather than `p` which would inherit the font styles from parents. This class name is very specific to this use case, yet it is in a generic component.

assets/stylesheets/trumps/_utility.scss
tyom: I think that you could just do a carriage return `br` for that purpose. I'm struggling to see other use cases for this.

src/apps/omis/apps/create/views/client-details.njk
tyom: Disabled input?

src/apps/omis/apps/create/views/client-details.njk
tyom: It seems simpler than introducing new elements and a utility class. It's even [used as example for address formatting](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/br) on MDN.

src/apps/omis/apps/create/views/client-details.njk
teneightfive: It's currently displaying more than just a single value though.

We're displaying company name and address. This is likely to change once we start doing some research on these steps though.

PR #489, opened 2017-08-21
DH 489 audit for contact

ztolley:
View an audit trail of changes made to a contact.

* List summary of changes
* Allow user to drill into details of which fields changed.

### DIFF
--- a/config/index.js
+++ b/config/index.js
@@ -44,6 +44,7 @@ const config = {
   sentryDsn: process.env.SENTRY_DSN,
   longDateFormat: 'D MMMM YYYY',
   mediumDateFormat: 'D MMM YYYY',
+  mediumDateTimeFormat: 'D MMM YYYY, h:MMa',
   paginationMaxResults: 10000,
 }
 
--- a/src/apps/audit/transformers.js
+++ b/src/apps/audit/transformers.js
@@ -1,7 +1,8 @@
 /* eslint camelcase: 0 */
 const { isArray, isPlainObject } = require('lodash')
 const dateFns = require('date-fns')
-const { mediumDateFormat } = require('../../../config')
+const { mediumDateTimeFormat } = require('../../../config')
+const { buildPagination } = require('../../lib/pagination')
 
 function transformChanges (changes, labels) {
   return Object.keys(changes)
@@ -14,10 +15,10 @@ function transformAuditLogToListItem (logEntry, labels = {}) {
 
   return {
     type: 'audit',
-    name: dateFns.format(logEntry.timestamp, `${mediumDateFormat}, HH:MMa`),
+    name: dateFns.format(logEntry.timestamp, mediumDateTimeFormat),
     contentMetaModifier: 'stacked',
     meta: [{
-      label: 'Advisor',
+      label: 'Adviser',
       value: logEntry.user.name,
     }, {
       label: 'Change count',
@@ -30,7 +31,7 @@ function transformAuditLogToListItem (logEntry, labels = {}) {
   }
 }
 
-function transformAuditLogToCollection (auditData, labels = {}) {
+function transformAuditLogToCollection (auditData, labels = {}, options = {}) {
   if (!isPlainObject(auditData)) { return }
   const resultItems = auditData.results
   if (!isArray(resultItems)) { return }
@@ -40,10 +41,10 @@ function transformAuditLogToCollection (auditData, labels = {}) {
   })
 
   return {
-    count: resultItems.length,
     items,
-    pagination: null,
+    count: items.length,
     countLabel: 'log entry',
+    pagination: buildPagination(options.query, auditData),
   }
 }
 
--- a/src/apps/audit/transformers.js
+++ b/src/apps/audit/transformers.js
@@ -1,0 +1,52 @@
+/* eslint camelcase: 0 */
+const { isArray, isPlainObject } = require('lodash')
+const dateFns = require('date-fns')
+const { mediumDateFormat } = require('../../../config')
+
+function transformChanges (changes, labels) {
+  return Object.keys(changes)
+    .map(key => labels[key] || key)
+    .join(', ')
+}
+
+function transformAuditLogToListItem (logEntry, labels = {}) {
+  const changeCount = (logEntry.changes && Object.keys(logEntry.changes).length) || 0
+
+  return {
+    type: 'audit',
+    name: dateFns.format(logEntry.timestamp, `${mediumDateFormat}, HH:MMa`),
+    contentMetaModifier: 'stacked',
+    meta: [{
+      label: 'Advisor',
+      value: logEntry.user.name,
+    }, {
+      label: 'Change count',
+      type: 'badge',
+      value: changeCount > 0 ? `${changeCount} changes` : 'No changes saved',
+    }, {
+      label: 'Fields',
+      value: transformChanges(logEntry.changes, labels),
+    }],
+  }
+}
+
+function transformAuditLogToCollection (auditData, labels = {}) {
+  if (!isPlainObject(auditData)) { return }
+  const resultItems = auditData.results
+  if (!isArray(resultItems)) { return }
+
+  const items = resultItems.map((resultItem) => {
+    return transformAuditLogToListItem(resultItem, labels)
+  })
+
+  return {
+    count: resultItems.length,
+    items,
+    pagination: null,
+    countLabel: 'log entry',
+  }
+}
+
+module.exports = {
+  transformAuditLogToCollection,
+}
--- a/src/apps/audit/transformers.js
+++ b/src/apps/audit/transformers.js
@@ -10,6 +10,18 @@ function transformChanges (changes, labels) {
     .join(', ')
 }
 
+function getChangeCountDisplay (changeCount) {
+  if (changeCount === 0) {
+    return 'No changes saved'
+  }
+
+  if (changeCount === 1) {
+    return `${changeCount} change`
+  }
+
+  return `${changeCount} changes`
+}
+
 function transformAuditLogToListItem (logEntry, labels = {}) {
   const changeCount = (logEntry.changes && Object.keys(logEntry.changes).length) || 0
 
@@ -23,7 +35,7 @@ function transformAuditLogToListItem (logEntry, labels = {}) {
     }, {
       label: 'Change count',
       type: 'badge',
-      value: changeCount > 0 ? `${changeCount} changes` : 'No changes saved',
+      value: getChangeCountDisplay(changeCount),
     }, {
       label: 'Fields',
       value: transformChanges(logEntry.changes, labels),
--- a/src/apps/contacts/controllers/audit.js
+++ b/src/apps/contacts/controllers/audit.js
@@ -1,0 +1,30 @@
+const { formatLongDate } = require('../../../../common/date')
+const { getContactAuditLog } = require('../repos')
+
+function formatAuditLog (logEntry) {
+  return {
+    name: logEntry.user.name,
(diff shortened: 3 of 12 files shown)


src/apps/audit/transformers.js
```diff
@@ -0,0 +1,49 @@
...
+const dateFns = require('date-fns')
+const { mediumDateFormat } = require('../../../config')
+
+function transformChanges (changes, labels) {
+  return Object.keys(changes)
+    .map(key => labels[key] || key)
+    .join(', ')
+}
+
+function transformAuditLogItemToListItem (logEntry, labels = {}) {
```
tyom: Can probably drop the first `Item`.
`transformAuditLogToListItem` to be consistent with other transformers.

src/apps/audit/transformers.js
```diff
@@ -0,0 +1,49 @@
...
+    type: 'audit',
+    name: dateFns.format(logEntry.timestamp, `${mediumDateFormat}, HH:MMa`),
+    contentMetaModifier: 'stacked',
+    meta: [{
+      label: 'Advisor',
+      value: logEntry.user.name,
+    }, {
+      label: 'Change count',
+      type: 'badge',
+      value: (logEntry.changes && Object.keys(logEntry.changes).length) || 'No changes saved',
```
tyom: Maybe worth adding word "changes" to the count. The count on its own in badge is not clear in the UI.

src/apps/audit/transformers.js
```diff
@@ -0,0 +1,49 @@
...
+    }, {
+      label: 'Fields',
+      value: transformChanges(logEntry.changes, labels),
+    }],
+  }
+}
+
+function transformAuditLogToCollection (auditData, query = {}, hasItemFilterLinks = false, labels = {}) {
```
tyom: You can remove arguments you're not using.

src/apps/audit/transformers.js
```diff
@@ -0,0 +1,49 @@
...
+    meta: [{
+      label: 'Advisor',
+      value: logEntry.user.name,
+    }, {
+      label: 'Change count',
+      type: 'badge',
+      value: (logEntry.changes && Object.keys(logEntry.changes).length) || 'No changes saved',
+    }, {
+      label: 'Fields',
+      value: transformChanges(logEntry.changes, labels),
```
tyom: Looks like the labels can be imported directly inside module as this transformer is specific to audit.

src/apps/contacts/views/audit.njk
```diff
@@ -0,0 +1,10 @@
+{% extends "./_layout.njk" %}
+
+{% block main_grid_right_column %}
+  <h2 class="heading-medium">Audit history</h2>
+  {{
+    Collection(auditLog | assignCopy({
+      countLabel: 'log entry'
```
tyom: You could set this label in the controller or even in transformer.

src/apps/audit/transformers.js
ztolley: The labels are used to show what fields changed in an entity, so they are different for different audit logs, i.e. fields changed for contact are fields changed for company

test/unit/apps/contacts/controllers/audit.test.js
```diff
@@ -0,0 +1,89 @@
...
+
+  afterEach(() => {
+    this.sandbox.restore()
+  })
+
+  it('should call the contact audit repository', (done) => {
+    this.controller.getAudit(this.req, {
+      breadcrumb: this.breadcrumbStub,
+      render: (template, data) => {
+        expect(this.getContactAuditLogStub).to.be.calledWith(this.req.session.token, this.req.params.contactId, 1)
```
tyom: Why not spy `render` instead?

ztolley: because that is the exit point for the code and it's where I want to see if other stubs have been called.

src/apps/contacts/controllers/audit.js
```diff
@@ -0,0 +1,29 @@
...
+async function getAudit (req, res, next) {
+  try {
+    const token = req.session.token
+    const contactId = req.params.contactId
+    const page = req.query.page || 1
+
+    const auditLog = await getContactAuditLog(token, contactId, page)
+      .then(result => transformAuditLogToCollection(result, contactDetailsLabels))
+
+    auditLog.pagination = buildPagination(req.query, auditLog)
```
tyom: Why not build pagination in `transformAuditLogToCollection`? Pagination is the concern of collection not `getAudit` controller which should just get the collection and pass it to view.

test/unit/apps/contacts/controllers/audit.test.js
```diff
@@ -0,0 +1,89 @@
...
+
+describe('Contact audit controller', () => {
+  beforeEach(() => {
+    this.sandbox = sinon.sandbox.create()
+    this.next = function (error) {
+      console.log(error)
+    }
+
+    this.transformed = {}
+    this.getContactAuditLogStub = this.sandbox.stub().resolves(auditLog)
```
tyom: Why not `nock` network requests?

src/apps/contacts/controllers/audit.js
ztolley: I followed other uses of pagination which were added in the controller, maybe because it's tied to the request to generate the URL

test/unit/apps/contacts/controllers/audit.test.js
ztolley: We've used this approach in other places too and I find it simpler than nock, it simply replaces immediate dependencies to make it more unit test like.

src/apps/contacts/controllers/audit.js
tyom: Which uses are those? We're trying to make controllers simpler. Would be good to know what else can be improved.

In other places (e.g. company investments) we use `transformResultsToCollection` which is currently part of search but I think can be extracted further up to be more generic and can be used to transform results from API to collection that macro expects.

Needless to say, I always prefer having a consistent way of solving similar problems. That way refactoring in future is less painful and error-prone.

test/unit/apps/contacts/controllers/audit.test.js
```diff
@@ -0,0 +1,89 @@
...
+      breadcrumb: this.breadcrumbStub,
+      render: (template, data) => {
+        expect(this.getContactAuditLogStub).to.be.calledWith(this.req.session.token, this.req.params.contactId, 1)
+        done()
+      },
+    }, this.next)
+  })
+
+  it('should pass on any page number specified, for pagination', (done) => {
+    const req = Object.assign({}, this.req, {
```
tyom: Why not `this.req.query.page = 3`? It's being reset for each test.

src/apps/audit/transformers.js
```diff
@@ -0,0 +1,51 @@
...
+    .map(key => labels[key] || key)
+    .join(', ')
+}
+
+function transformAuditLogToListItem (logEntry, labels = {}) {
+  const changeCount = (logEntry.changes && Object.keys(logEntry.changes).length) || 0
+
+  return {
+    type: 'audit',
+    name: dateFns.format(logEntry.timestamp, `${mediumDateFormat}, HH:MMa`),
```
teneightfive: Should we add a datetime format to config? Feels like it could be useful in a few places.

src/apps/audit/transformers.js
```diff
@@ -0,0 +1,51 @@
...
+    contentMetaModifier: 'stacked',
+    meta: [{
+      label: 'Advisor',
```
teneightfive: `Adviser`

test/unit/apps/contacts/controllers/audit.test.js
tyom: I think we prefer to use `nock` for API responses and stub smaller responses that can be mocked inline in the same file. You have a fairly large JSON that emulates API response and I feel that `nock` is more appropriate to use in this case.

src/apps/contacts/controllers/audit.js
ztolley: I'll keep this for now so it doesn't use results to collection, otherwise it'll get held up and its already holding up company audit.

Will move the call to pagination into the transformer and pass options in the same as search

config/index.js
```diff
@@ -44,6 +44,7 @@ const config = {
   sentryDsn: process.env.SENTRY_DSN,
   longDateFormat: 'D MMMM YYYY',
   mediumDateFormat: 'D MMM YYYY',
+  mediumDateTimeFormat: 'D MMM YYYY, HH:MMa',
```
teneightfive: Shouldn't it be `h:mma` for time? Current format could end up with as `23:30pm`

ztolley: I'll update a test to be in the afternoon to check.

ztolley: you are correct, I had a test that was checking for 16:06pm... DOH!

src/apps/audit/transformers.js
```diff
@@ -0,0 +1,63 @@
...
+function transformChanges (changes, labels) {
+  return Object.keys(changes)
+    .map(key => labels[key] || key)
+    .join(', ')
+}
+
+function getChangeCountDisplay (changeCount) {
+  if (changeCount === 0) {
+    return 'No changes saved'
+  } else if (changeCount === 1) {
```
tyom: `else if` is not necessary here as you return from this function. I think it would also improve readability via format: each if statement starting on new line.

src/apps/audit/transformers.js
```diff
@@ -0,0 +1,63 @@
...
+      type: 'badge',
+      value: getChangeCountDisplay(changeCount),
+    }, {
+      label: 'Fields',
+      value: transformChanges(logEntry.changes, labels),
+    }],
+  }
+}
+
+function transformAuditLogToCollection (auditData, labels = {}, options = {}) {
```
tyom: Still not sure why we need to pass labels down from collection container all the way to transform changes. Aren't the labels part of audit sub-app anyway?

ztolley: The label it a human readable label for a entity field name. The field names differ from entity to entity, so when I show a list of change for a company it will list different field names to when I create a list of changes for a contact.

tyom: > The label it a human readable label for a entity field name

Thanks.

My question is whether these labels could be defined/referenced under audit sub-app so that the field names in `changes` match those from within the static labels predefined in the app?

ztolley: no, the labels belong in the sub app of the entity you are showing the audit log of

tyom: Why can't they be referenced from audit instead of being passed down via multiple functions? Do they ever change on their way down to transformer?

src/apps/contacts/router.js
```diff
@@ -7,10 +7,12 @@ const { renderContactList } = require('./controllers/list')
 const { postDetails, editDetails } = require('./controllers/edit')
 const { archiveContact, unarchiveContact } = require('./controllers/archive')
 const { getInteractions } = require('./controllers/interactions')
+const { getAudit } = require('./controllers/audit')
 
 const LOCAL_NAV = [
   { path: 'details', label: 'Details' },
   { path: 'interactions', label: 'Interactions' },
+  { path: 'audit', label: 'Audit log' },
```
teneightfive: `Audit log` feels quite a technical term for this section. Like _server log_. 

What about `Activity` (think Facebook uses this) or `History` or even `Changes`?

jim68000: I'll raise that with our content guru as that's a good point. This is what it's called in the story though, so it shouldn't be a blocker for now.

ztolley: Activity could get confused with interactions

The card says Audit log, so I'm sticking with that

PR #491, opened 2017-08-22
Refactor investment collections

tyom:
- Use macro config objects to create filters and sort fields
- Simplify logic for creating selected collection filters summary


src/apps/investment-projects/builders.js
```diff
@@ -1,65 +1,32 @@
...
-function buildInvestmentSorting (filtersQuery = {}) {
-  const query = Object.assign(
-    {},
-    { sortby: SORT_OPTIONS[0].value },
-    filtersQuery
-  )
+      if (isFunction(field.options)) {
+        const selectedValues = at(keyBy(field.options(), 'value'), fieldValue).filter(x => x)
+        fieldsObj[field.name].valueLabel = selectedValues.map(x => x.label).join(', ')
+      }
```
marcofucci: You probably want to do the same when `field.options` is not a function but an array already.
Maybe you need:
```javascript
const fieldOptions = isFunction(field.options) ? field.options() : field.options
...
```

tyom: Nice spot! Thanks.

I've updated the code and added test to ensure both cases are handled.

PR #494, opened 2017-08-24
End to End features for contact collections

venureddy95:
Adding contact collection features and refactored the CSS changes across all the features.


test/acceptance/features/collections-contact.feature
```diff
@@ -0,0 +1,112 @@
+@collections-contact
+Feature: View Collection of contacts
+  As an existing user
+  I would like to view all the Contacts in one place
+  And be able to read the contact details as expected
+
+  @verify-contact-name
```
tyom: Can I suggest using the same suffix for tags so the whole feature can be run by dropping the suffix of the tag? e.g. `collections-contact` as feature and `collections-contact--name`, `collections-contact--company` as scenarios.

test/acceptance/features/contacts-audit-log.feature
```diff
@@ -0,0 +1,65 @@
+Feature: View Audit history of a contact
```
tyom: Should we add tags here as well?

PR #496, opened 2017-08-24
Allow entity macro titles without links and override metalist modifiers

ztolley:
Change the entity macro so that it only makes the entity title a link
if an id is passed to it, otherwise make it a plain title.

Add an optional parameter to the entity macro 'contentMetaModifier' to
allow an alternative modifier to applied to the metalist macro call
to allow alernative layouts to be requested when generating entity
collections.

### DIFF
--- a/assets/stylesheets/components/_meta-list.scss
+++ b/assets/stylesheets/components/_meta-list.scss
@@ -72,3 +72,15 @@
     margin-right: 0;
   }
 }
+
+.c-meta-list--stacked {
+  display: block;
+
+  .c-meta-list__item {
+    display: block;
+  }
+
+  .c-meta-list__item  + .c-meta-list__item {
+    margin-left: 0;
+  }
+}
--- a/assets/stylesheets/components/_meta-list.scss
+++ b/assets/stylesheets/components/_meta-list.scss
@@ -72,15 +72,3 @@
     margin-right: 0;
   }
 }
-
-.c-meta-list--stacked {
-  display: block;
-
-  .c-meta-list__item {
-    display: block;
-  }
-
-  .c-meta-list__item  + .c-meta-list__item {
-    margin-left: 0;
-  }
-}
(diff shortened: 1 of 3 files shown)


src/templates/_macros/entities.njk
```diff
@@ -24,7 +26,11 @@
 
       <div class="c-entity__header">
         <h3 class="c-entity__title">
-          <a href="/{{ props.type | pluralise }}/{{ props.id }}">{{ props.name | highlight(props.highlightTerm) }}</a>
+          {% if props.id %}
+            <a href="/{{ props.type | pluralise }}/{{ props.id }}">{{ props.name | highlight(props.highlightTerm) }}</a>
```
tyom: Can store `props.name ...` in a variable above to avoid repeating.

assets/stylesheets/components/_meta-list.scss
```diff
@@ -72,3 +72,15 @@
     margin-right: 0;
   }
 }
+
+.c-meta-list--stacked {
```
tyom: Does this change belong here? Also what is the use case for it?

ztolley: This change is to allow fields to be stacked instead of the left right layout we have.
If it doesn't go here then it would just be a follow up PR , I just thought they are both making a minor change to presentation for entities so could be in the same pr

assets/stylesheets/components/_meta-list.scss
```diff
@@ -72,3 +72,15 @@
...
+  display: block;
+
+  .c-meta-list__item {
+    display: block;
+  }
+
+  .c-meta-list__item  + .c-meta-list__item {
```
tyom: Would be more consistent to `& + &` inside `c-meta-list__item`.

assets/stylesheets/components/_meta-list.scss
```diff
@@ -72,3 +72,15 @@
     margin-right: 0;
   }
 }
+
+.c-meta-list--stacked {
+  display: block;
+
+  .c-meta-list__item {
+    display: block;
```
tyom: What if I wanted to stack items with inlined key/value?

There's already a stacked modifier on list item. It might be better to combine those from the view/macro.

PR #499, opened 2017-08-25
Contact collection filters

tyom:
Add four filters for contacts collections:
- company name (partial match)
- company sector
- company country
- company UK region

### TODO
- [x] Add controller tests
- [x] Fix issue with conditional subfield (UK region)


src/apps/contacts/macros.js
```diff
@@ -1,21 +1,24 @@
-const metadataRepo = require('../../lib/metadata')
-const { transformObjectToOption } = require('../transformers')
+const { globalFields } = require('../macros')
```
ztolley: macros just feels like the wrong name for that file

src/apps/contacts/views/list.njk
```diff
@@ -1,10 +1,21 @@
 {% extends "_layouts/two-column.njk" %}
 
+{% block main_grid_left_column %}
+  {{
+    CollectionFilters({
+      query: QUERY,
```
ztolley: Curious why we have starting putting in Capitals for some vars to indicate they are constants but this isn't something we've done across the board. I've tried it before, and just felt easier to keep them lower case as so much is a constant

src/apps/builders.js
```diff
@@ -76,10 +76,38 @@ function buildFormWithStateAndErrors (form, requestBody, errorsObject) {
   return buildFormWithErrors(formWithState, errorsObject)
 }
 
+function buildSelectedFiltersSummary (fields, query = {}) {
```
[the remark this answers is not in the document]
tyom: Currently builder is kind of a dumping ground for stuff that does not fit into other concepts (transformers, middleware, controllers etc). I try to avoid builders but here they are sufficiently reusable to go in the top level builders file to be used across all sub-apps.

src/apps/contacts/macros.js
tyom: Do suggest an alternative please.

src/apps/contacts/views/list.njk
tyom: All top level locals properties are all caps. I think we introduced them a while ago to differentiate from other locals properties added by different middleware.

PR #500, opened 2017-08-26
Refactor auth pages

tyom:
- Use new form patterns
- Add missing unit tests


src/apps/auth/macros.js
```diff
@@ -0,0 +1,21 @@
+const signInForm = {
```
ztolley: Why is it called macros?

Maybe call it something to do with form definitions?

tyom: It's called macros because it's not specific to forms. We use them here for forms, but you can define anything that is composed of macros.

ztolley: I associate macros with nunjucks

tyom: This is the bridge between controller and view layer.

ztolley: hmm well I just feel the name isn't clear

src/apps/auth/middleware.js
```diff
@@ -60,7 +50,15 @@ function postHandler (req, res, next) {
     })
 }
 
+function handleSignOut (req, res, next) {
```
teneightfive: Should this be in controllers?

The `next()` method is never called like other middleware. This method feels like more of an endpoint than a piece of middleware.

src/apps/auth/macros.js
tyom: May be so, but at least it's consistent and used in other sub-apps. If we find a better name we can change it in one go.

PR #501, opened 2017-08-28
Created a stacked style for metalists

ztolley:
capture

### DIFF
--- a/assets/stylesheets/components/_meta-list.scss
+++ b/assets/stylesheets/components/_meta-list.scss
@@ -72,3 +72,15 @@
     margin-right: 0;
   }
 }
+
+.c-meta-list--stacked {
+  display: block;
+
+  .c-meta-list__item {
+    display: block;
+  }
+
+  .c-meta-list__item + .c-meta-list__item {
+    margin-left: 0;
+  }
+}
--- a/src/apps/components/controllers.js
+++ b/src/apps/components/controllers.js
@@ -51,10 +51,31 @@ async function renderEntityList (req, res) {
       })
     })
 
+  const auditLog = {
+    items: [{
+      type: 'audit',
+      name: '20 December 2016 9:00am',
+      contentMetaModifier: 'stacked',
+      meta: [{
+        label: 'Advisor',
+        value: 'Fred Smith',
+      }, {
+        label: 'Change count',
+        type: 'badge',
+        value: '2 changes',
+      }, {
+        label: 'Fields',
+        value: 'Name',
+      }],
+    }],
+    count: 1,
+  }
+
   return res
     .breadcrumb('Entity list')
     .render('components/views/entity-list', {
       investmentProjects,
+      auditLog,
       companiesSearch: await authorisedRequest(req.session.token, `${config.apiRoot}/v3/search?term=samsung&entity=company&limit=10`),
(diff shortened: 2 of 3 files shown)


assets/stylesheets/components/_meta-list.scss
```diff
@@ -72,3 +72,15 @@
     margin-right: 0;
   }
 }
+
+.c-meta-list--stacked {
+  display: block;
+
+  > .c-meta-list__item {
```
teneightfive: Do you need the immediate child selector here? Isn't it being nested within the parent class selector enough?

src/apps/components/views/entity-list.njk
```diff
@@ -87,4 +87,9 @@
     </div>
   {% endcall %}
 
+    {% call Example('Audit log') %}
```
teneightfive: Needs outdenting by 2 spaces.

assets/stylesheets/components/_meta-list.scss
ztolley: probably. Had it there initially to stop the style affecting sub-components but then that's not going to happen

src/apps/components/controllers.js
```diff
@@ -51,10 +51,31 @@ async function renderEntityList (req, res) {
       })
     })
 
+  const auditLog = {
+    items: [{
+      type: 'audit',
+      name: '20th December 2016 09:00am',
```
teneightfive: This is fine because its an example but we should make sure we [write dates](https://www.gov.uk/service-manual/design/dates#how-to-write-dates) as 20 December 2016 for consistency.

teneightfive: Also, it looks like the time is using a combination of 24 hour and am/pm.

tyom:
Is it worth taking the opportunity to document all other variants of `MetaList`?

ztolley:
@tyom I'd humbly say that's beyond the scope of this PR and I don't want it to hold up the audit card

assets/stylesheets/components/_meta-list.scss
teneightfive: Correct. One of the points of using BEM.

src/apps/components/controllers.js
```diff
@@ -51,10 +51,31 @@ async function renderEntityList (req, res) {
       })
     })
 
+  const auditLog = {
+    items: [{
+      type: 'audit',
+      name: '20 December 2016 09:00am',
```
teneightfive: `9:00am`?

src/apps/components/views/entity-list.njk
```diff
@@ -21,7 +21,7 @@
           {% for company in companiesSearch.companies %}
             <li class="c-entity-list__item">
               <h3 class="c-entity-list__item-header">
-                <a href="/viewcompanyresult/{{ company.id }}">{{ company.name}}</a>
+              <a href="/viewcompanyresult/{{ company.id }}">{{ company.name}}</a>
```
teneightfive: Unnecessary change?

src/apps/components/controllers.js
ztolley: I have missed your attention to detail :)

jim68000:
🤘

src/apps/components/controllers.js
teneightfive: It's all about the polish :)

PR #502, opened 2017-08-29
DH-484 Content Review

yeahfro:
- Wording tweaks

### DIFF
--- a/src/apps/companies/views/_layout-view.njk
+++ b/src/apps/companies/views/_layout-view.njk
@@ -27,8 +27,8 @@
       <a class="c-local-nav__link {% if tab == 'contacts' %}is-active{% endif %}" href="/companies/{{company.id}}/contacts">Contacts</a>
       <a class="c-local-nav__link {% if tab == 'interactions' %}is-active{% endif %}" href="/companies/{{company.id}}/interactions">Interactions</a>
       {# <a class="c-local-nav__link {% if tab == 'documents' %}is-active{% endif %}" href="">Documents</a> #}
-      <a class="c-local-nav__link {% if tab == 'exports' %}is-active{% endif %}" href="/companies/{{company.id}}/exports">Exports</a>
-      <a class="c-local-nav__link {% if tab == 'investments' %}is-active{% endif %}" href="/companies/{{ company.id }}/investments">Investments</a>
+      <a class="c-local-nav__link {% if tab == 'exports' %}is-active{% endif %}" href="/companies/{{company.id}}/exports">Export</a>
+      <a class="c-local-nav__link {% if tab == 'investments' %}is-active{% endif %}" href="/companies/{{ company.id }}/investments">Investment</a>
     {% endif %}
   </nav>
 {% endblock %}
--- a/src/apps/companies/views/exports-edit.njk
+++ b/src/apps/companies/views/exports-edit.njk
@@ -3,7 +3,7 @@
 {% block main_grid_right_column %}
   {{ super() }}
 
-  <h2 class="heading-medium">Add export markets(s)</h2>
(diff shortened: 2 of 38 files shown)


test/unit/apps/contacts/controllers/details.test.js
```diff
@@ -137,6 +137,47 @@ describe('Contact controller', function () {
...
+          'Does not want to be contacted',
+          'Changed role/responsibility',
+        ]
+        expect(res.locals.reasonForArchiveOptions).to.deep.equal(expected)
+        done()
+      }
+      contactController.getCommon(req, res, next)
+    })
+    it('should store reason for archive options prefix in locals', (done) => {
+      const req = {
```
tyom: Might be better to define those once in `beforeEach` at the top and spy `render` and `next` instead of redefining them for each case.

src/apps/companies/views/_layout-view.njk
```diff
@@ -27,8 +27,8 @@
       <a class="c-local-nav__link {% if tab == 'contacts' %}is-active{% endif %}" href="/companies/{{company.id}}/contacts">Contacts</a>
       <a class="c-local-nav__link {% if tab == 'interactions' %}is-active{% endif %}" href="/companies/{{company.id}}/interactions">Interactions</a>
       {# <a class="c-local-nav__link {% if tab == 'documents' %}is-active{% endif %}" href="">Documents</a> #}
-      <a class="c-local-nav__link {% if tab == 'exports' %}is-active{% endif %}" href="/companies/{{company.id}}/exports">Exports</a>
-      <a class="c-local-nav__link {% if tab == 'investments' %}is-active{% endif %}" href="/companies/{{ company.id }}/investments">Investments</a>
+      <a class="c-local-nav__link {% if tab == 'exports' %}is-active{% endif %}" href="/companies/{{company.id}}/exports">Export</a>
+      <a class="c-local-nav__link {% if tab == 'investments' %}is-active{% endif %}" href="/companies/{{ company.id }}/investments">Investment</a>
```
teneightfive: Should we be mixing singular and plural in this navigation? We're saying _Contacts_ plural but then _Investment_ singular.

We might need to address the consistency of how we refer to investment too as elsewhere we say _Investment projects_.

src/apps/companies/views/exports-edit.njk
```diff
@@ -3,7 +3,7 @@
 {% block main_grid_right_column %}
   {{ super() }}
 
-  <h2 class="heading-medium">An an export summary</h2>
+  <h2 class="heading-medium">Add export markets(s)</h2>
```
teneightfive: Should it be `market(s)` if we're using that format of plural in brackets?

Or what about always using the plural in this instance?

src/apps/companies/views/_layout-view.njk
yeahfro: I will discuss this with Mills to see how she wants to proceed.

src/apps/companies/views/exports-edit.njk
yeahfro: I have removed the redundant s and will discuss this further with Mills regarding the brackets.

test/unit/apps/contacts/controllers/details.test.js
```diff
@@ -12,143 +12,99 @@ describe('Contact controller', function () {
...
-        buildCompanyUrl: buildCompanyUrlStub,
+
+    contactController = createContactController(
+      getContactStub,
+      getDisplayContactStub,
+      getDitCompanyStub,
+      buildCompanyUrlStub
+    )
+
+    req = {
```
tyom: In many other places we just set these on `this`.

PR #504, opened 2017-08-29
OMIS Dashboard (collection)

teneightfive:
This PR adds support for a simple list of orders, or as they are being named, a collection.

👉 [**DEMO**](https://datahub-beta2-pr-504.herokuapp.com/omis) 👈 

## What it looks like

localhost_3001_omis_page 1 sortby created_on 3adesc

### DIFF
--- a/src/apps/omis/apps/list/controllers.js
+++ b/src/apps/omis/apps/list/controllers.js
@@ -1,0 +1,7 @@
+function renderList (req, res) {
+  res.render('omis/apps/list/views/list')
+}
+
+module.exports = {
+  renderList,
+}
--- a/src/apps/omis/apps/list/views/list.njk
+++ b/src/apps/omis/apps/list/views/list.njk
@@ -1,0 +1,11 @@
+{% extends "_layouts/two-column.njk" %}
+
+{% block main_grid_right_column %}
+  {{
+    Collection(results | assign({
+      countLabel: 'order',
+      sortForm: sortForm,
+      query: QUERY
+    }))
+  }}
+{% endblock %}
--- a/src/apps/dashboard/views/dashboard.njk
+++ b/src/apps/dashboard/views/dashboard.njk
@@ -6,7 +6,7 @@
       inputName: 'term',
       modifier: 'global',
       action: '/search/companies',
-      inputLabel: 'Search for company, contact or investment project'
+      inputLabel: 'Search for company, contact, investment project or OMIS order'
     }) }}
   {% endcall %}
 {% endblock %}
--- a/src/apps/omis/apps/list/index.js
+++ b/src/apps/omis/apps/list/index.js
@@ -1,0 +1,7 @@
+const router = require('./router')
+
+module.exports = {
+  displayName: 'Orders',
+  mountpath: '/',
+  router,
+}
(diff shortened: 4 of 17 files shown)


tyom:
Do you think it's worth making country a badge as we do for contacts?
image

src/templates/_macros/entities.njk
```diff
@@ -28,7 +30,7 @@
       <div class="c-entity__header">
         <h3 class="c-entity__title">
           {% if props.id %}
-            <a href="/{{ props.type | pluralise }}/{{ props.id }}">{{ highlightedName }}</a>
+            <a href="/{{ url }}{{ props.id }}">{{ highlightedName }}</a>
```
tyom: Should `url` be full so it can be predefined in a transformer? I think it might be necessary for companies anyway as URLs are tricky there.

teneightfive:
>Do you think it's worth making country a badge as we do for contacts?

👍

src/templates/_macros/entities.njk
teneightfive: Won't it always need to link the item though? Felt like I would be repeating that logic in the transformer so I started with only replacing the bit that was different.

If more needs to be different in future we can always refactor again.

src/apps/omis/apps/list/controllers.js
```diff
@@ -0,0 +1,7 @@
+function renderList (req, res) {
```
ztolley: getList?

src/apps/omis/apps/list/transformers.js
```diff
@@ -0,0 +1,49 @@
...
+        type: 'badge',
+        value: get(primary_market, 'name'),
+      },
+      {
+        label: 'Company',
+        value: get(company, 'name'),
+      },
+      {
+        label: 'Last modified',
+        value: modified_on ? format(modified_on, mediumDateFormat) : null,
```
ztolley: would it need time?

src/apps/omis/apps/list/views/list.njk
```diff
@@ -0,0 +1,11 @@
+{% extends "_layouts/two-column.njk" %}
+
+{% block main_grid_right_column %}
+  {{
+    Collection(results | assign({
```
ztolley: could those be added in the controller?

src/apps/omis/apps/list/controllers.js
teneightfive: This is the controller method which follows the pattern of trying to describe what the end method is doing. So in this case it is rendering the list view.

src/apps/omis/apps/list/transformers.js
teneightfive: Good shout. We may trial doing a time ago approach in some testing but for now will use datetime.

src/apps/omis/apps/list/views/list.njk
teneightfive: Not sure which bits you mean? I used the same approach that was used within [contacts view](https://github.com/uktrade/data-hub-frontend/blob/2d08c5410346ca18986c8bba6635e4632a46f04e/src/apps/contacts/views/list.njk#L14-L19).

src/templates/_macros/entities.njk
tyom: Maybe call it `urlPrefix` then. Because it's only part of the URL.

PR #505, opened 2017-08-29
Extract collection transformers to me more generic

tyom:
- create generic collection transformer to transform API response to collection with pagination, and result count.
- create search transformer to transform API response to collection with highlight term and aggregations

**TODO**
- [x] Update tests

### DIFF
--- a/src/apps/transformers.js
+++ b/src/apps/transformers.js
@@ -1,5 +1,7 @@
 /* eslint-disable camelcase */
-const { keyBy, snakeCase } = require('lodash')
+const { keyBy, snakeCase, isPlainObject, isFunction } = require('lodash')
+
+const { buildPagination } = require('../lib/pagination')
 
 function transformObjectToOption ({ id, name }) {
   return {
@@ -39,10 +41,45 @@ function buildMetaDataObj (collection) {
   })
 }
 
+/**
+ * @param {object} [options] {object}
+ * @param {string} [options.entityType] - API entity type
+ * @param {string} [options.searchTerm] - search term used for highlighting words in collection macro
+ * @param {object} [options.query] - URL query object used in pagination
+ * @param {...function} [itemTransformers] - an array of transformer functions to apply for each item in the list
+ * @returns {object, undefined}
+ */
+function transformApiResponseToCollection (options = {}, ...itemTransformers) {
+  /**
+   * @param {object} response - API response object
+   * @returns {function}
+   */
+  return function transformResponseToCollection (response) {
+    if (!isPlainObject(response) || !options.entityType) { return }
+
+    const pluralisedEntity = options.entityType === 'company' ? 'companies' : `${options.entityType}s`
+    let items = response[pluralisedEntity] || response.items || response.results
+
+    if (!items) { return }
+
+    itemTransformers.forEach(transformer => {
+      if (!isFunction(transformer)) { return }
+      items = items.map(transformer)
+    })
+
+    return Object.assign({}, {
+      items,
+      count: response.count,
+      pagination: buildPagination(options.query, response),
+    })
+  }
+}
+
 module.exports = {
   buildMetaDataObj,
   transformObjectToOption,
   transformStringToOption,
   transformContactToOption,
   transformIdToObject,
+  transformApiResponseToCollection,
 }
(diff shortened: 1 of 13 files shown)


src/apps/transformers.js
```diff
@@ -39,10 +43,41 @@ function buildMetaDataObj (collection) {
...
+ */
+function transformApiResponseToCollection (response, entityType, options = {}) {
+  const pluralisedEntity = entityType === 'company' ? 'companies' : `${entityType}s`
+  const resultsItems = response[pluralisedEntity] || response.items || response.results
+
+  if (!resultsItems) { return null }
+
+  let items = resultsItems.map(item => Object.assign({}, item, { type: entityType }))
+
+  if (entityType === 'investment_project') {
```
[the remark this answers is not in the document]
tyom: Not sure what you mean. Can you give an example of how a transformer would know which list item transformer dependency it should apply to a specific entity type?

src/apps/transformers.js
```diff
@@ -39,10 +43,43 @@ function buildMetaDataObj (collection) {
   })
 }
 
+/**
+ * @param response {object}
```
ztolley: Maybe describe what each of the parameters is and how to use it.

src/apps/transformers.js
```diff
@@ -39,10 +43,43 @@ function buildMetaDataObj (collection) {
...
+  if (!isPlainObject(response) || !entityType) { return }
+
+  const pluralisedEntity = entityType === 'company' ? 'companies' : `${entityType}s`
+  const resultsItems = response[pluralisedEntity] || response.items || response.results
+
+  if (!resultsItems) { return }
+
+  let items = resultsItems.map(item => Object.assign({}, item, { type: entityType }))
+
+  if (entityType === 'investment_project') {
```
ztolley: why the conditional here and what happens if you don't provide a type or a new unknown type?

ztolley: Also it feels wrong that a piece of logic unique to a particular entity type is in a  common module

src/apps/transformers.js
tyom: I try to avoid obsolete comments and rely on code documenting the purpose. Comments get outdated very quickly because people inevitably forget to update them to reflect the code which in turn can become a major point of confusion for users.

In this case I documented the types of arguments and option properties, keeping it to the minimum to avoid them becoming quickly outdated. Do you think it's not clear what those parameters are from their names and their context (function name)?

src/apps/transformers.js
tyom: I started off with a generic collection that would only apply pagination and add entity types on list items. Then as I started using it I realise I was repeating the same thing in different places to make give collections entity-specific item configuration. At which point I decided to do it in the body of this function because I couldn't think of a use case when we would want to have a bare API response in collection.

Potentially that could be extracted into a separate function which would be called only if entity type is provided.

tyom:
> Not sure I get what benefit this provides or how to use it. In the case of audit logs, how would you apply this and what benefits would it provide over the current code in it's PR? Audit log has a simple transformer to convert the api response format to the presentation format but needs to be unique to audit log.

The main benefit is to avoid the repetition of definition of common collection properties such as pagination. Of course your audit-specific collection can then further enhance the collection to fit the purpose of audit.

src/apps/transformers.js
ztolley: If there is something unique, and you want to keep this generic then why not have a parameter to pass in a function or object that entity can pass to be called

ztolley:
I think I'd need to see in practice how audit log would use this to understand the benefits over the current approach

tyom:
> I think I'd need to see in practice how audit log would use this to understand the benefits over the current approach

Would you like me to branch off your branch to demonstrate?

src/apps/transformers.js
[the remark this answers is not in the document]
teneightfive: What about passing the transformer in to the search method?

src/apps/transformers.js
ztolley: it makes it quicker to pick things up if there is clear comments about the expected params instead of deducing it from the code is all I'm thinking.

ztolley:
Maybe a branch that would demonstrate how audit would work , or if's just a small change to one file then a gist or just message on slack. But this is lower priority than other stuff in the sprint.

PR #506, opened 2017-08-30
Only list markets where OMIS is supplied

teneightfive:
Previously we were using all countries in the market choice step
but OMIS is only supplied in selected markets so this changes that
list to the correct options.

### DIFF
--- a/src/apps/omis/apps/create/controllers/market.js
+++ b/src/apps/omis/apps/create/controllers/market.js
@@ -1,10 +1,14 @@
+const { filter } = require('lodash')
+
 const metadataRepo = require('../../../../../lib/metadata')
 const { FormController } = require('../../../controllers')
 const { transformObjectToOption } = require('../../../../transformers')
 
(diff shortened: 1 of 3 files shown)


src/apps/omis/apps/create/controllers/market.js
```diff
@@ -1,10 +1,16 @@
...
+
 const metadataRepo = require('../../../../../lib/metadata')
 const { FormController } = require('../../../controllers')
 const { transformObjectToOption } = require('../../../../transformers')
 
 class MarketController extends FormController {
   configure (req, res, next) {
-    req.form.options.fields.primary_market.options = metadataRepo.countryOptions.map(transformObjectToOption)
+    const filterMarkets = filter(metadataRepo.omisMarketOptions, (market) => {
+      return !market.disabled_on
```
tyom: I'd prefer to have these kinds of predicates inline with ES6 auto-return.

PR #508, opened 2017-08-30
Make contact audit history titles/links consistent with investment

ztolley:
Realised that although the requirements for contact audit log had 'log' as the title, the existing screens use the words 'Audit history'.


jim68000:
As mentioned - I don't like 'audit history' as it doesn't make a lot of sense as a title, buuuuut I'd rather we put something in with the wrong but consistent wording and review that once we have the feature live.

PR #513, opened 2017-08-30
Support custom URL within entity macro

teneightfive:
In some cases we don't want to just use a plural of the type so this
adds support for a custom URL to be passed to the macro.

### DIFF
--- a/src/templates/_macros/entities.njk
+++ b/src/templates/_macros/entities.njk
@@ -4,6 +4,7 @@
  # @param {string} props.name - entity name
  # @param {string} props.type - entity type (e.g. 'date' to format dates)
  # @param {string} [props.id] - entity id, used to create a link to the entity if provided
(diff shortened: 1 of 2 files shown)


src/templates/_macros/entities.njk
```diff
@@ -28,7 +30,7 @@
       <div class="c-entity__header">
         <h3 class="c-entity__title">
           {% if props.id %}
-            <a href="/{{ props.type | pluralise }}/{{ props.id }}">{{ highlightedName }}</a>
+            <a href="/{{ urlPrefix }}{{ props.id }}">{{ highlightedName }}</a>
```
ztolley: Are there any situations where you want to give a fully custom URL? say /thing/324234/details ?

teneightfive: @tyom did potentially suggest that but as we don't have those cases at the moment I didn't cater for them. If we do, we can always change it to support `urlSuffix`.

PR #514, opened 2017-08-30
Update audit logs to use generic collection

tyom:
Updates the audit transformer to be compatible with new generic collection handlers.

Changed company and contact audit controllers to use new generic collections

### DIFF
--- a/test/unit/apps/companies/controllers/audit.test.js
+++ b/test/unit/apps/companies/controllers/audit.test.js
@@ -9,9 +9,13 @@ describe('Company audit controller', () => {
     this.transformed = {}
     this.getDitCompanyStub = this.sandbox.stub().resolves(companyData)
     this.getCompanyAuditLogStub = this.sandbox.stub().resolves(auditLog)
-    this.transformAuditLogToCollectionStub = this.sandbox.stub().returns(this.transformed)
     this.getCommonTitlesAndlinksStub = this.sandbox.stub()
     this.next = this.sandbox.stub()
+
+    this.transformApiResponseToCollectionInnerStub = this.sandbox.stub().returns()
+    this.transformApiResponseToCollectionStub = this.sandbox.stub().returns(this.transformApiResponseToCollectionInnerStub)
+    this.generatedTransformer = this.sandbox.stub()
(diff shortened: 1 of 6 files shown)


test/unit/apps/companies/controllers/audit.test.js
```diff
@@ -9,9 +9,13 @@ describe('Company audit controller', () => {
     this.transformed = {}
     this.getDitCompanyStub = this.sandbox.stub().resolves(companyData)
     this.getCompanyAuditLogStub = this.sandbox.stub().resolves(auditLog)
-    this.transformAuditLogToCollectionStub = this.sandbox.stub().returns(this.transformed)
     this.getCommonTitlesAndlinksStub = this.sandbox.stub()
-    this.next = this.sandbox.stub()
+    this.next = function (error) { console.log(error) }
```
teneightfive: Do we need to be logging errors in tests?

ztolley: When we just spy then tests timeout and don't report what the error was.

test/unit/apps/contacts/controllers/audit.test.js
```diff
@@ -59,11 +64,14 @@ describe('Contact audit controller', () => {
...
 
-  it('should transform the results returns', (done) => {
+  it('should transform the results returned', (done) => {
+    const options = { entityType: 'audit' }
+
     this.controller.getAudit(this.req, {
       breadcrumb: this.breadcrumbStub,
       render: (template, data) => {
-        expect(this.transformAuditLogToCollectionStub).to.be.calledWith(auditLog, contactDetailsLabels)
+        expect(this.transformApiResponseToCollectionStub).to.be.calledWith(options, this.generatedTransformer)
```
teneightfive: Useful to wrap async tests in a try/catch so that failures are returned rather than a timout:

```js
render: (template, data) => {
  try {
    ...
    done()
  } catch (error) {
    done(error)
  }
}
```

ztolley: Using try catch and replaced next with done as errors will call done and pass the error to it.

PR #516, opened 2017-08-31
Remove entityType from apps transformer

tyom:
Back-end now always returns `results` so there is no need to pass `entityType`
into transformer options.

### DIFF
--- a/src/apps/contacts/middleware.js
+++ b/src/apps/contacts/middleware.js
@@ -5,22 +5,17 @@ const { transformApiResponseToSearchCollection } = require('../search/transforme
 const { transformContactToListItem } = require('./transformers')
 
 async function getContactsCollection (req, res, next) {
-  const searchEntity = 'contact'
-
   try {
     res.locals.results = await search({
       searchTerm: '',
-      searchEntity,
+      searchEntity: 'contact',
       requestBody: req.body,
       token: req.session.token,
       page: req.query.page,
       isAggregation: false,
     })
       .then(transformApiResponseToSearchCollection(
-        {
-          entityType: searchEntity,
-          query: req.query,
-        },
+        { query: req.query },
         transformContactToListItem,
       ))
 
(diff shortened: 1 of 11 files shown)


src/apps/contacts/middleware.js
```diff
@@ -17,10 +17,7 @@ async function getContactsCollection (req, res, next) {
       isAggregation: false,
     })
       .then(transformApiResponseToSearchCollection(
-        {
-          entityType: searchEntity,
```
teneightfive: You can pass the `searchEntity` string directly into `search()` now as its only used once.

tyom:
> EntityType doesn't appear to be in all endpoints so this card me be premature.

`entityType` was used in collections to work around problem where API was giving different property for different entities. e.g. `companies` for global search with company entity, `contacts` for global search with contact entity and `results` for entity search endpoint.

What should I be looking for in the URL you quoted?

src/apps/investment-projects/middleware/collection.js
```diff
@@ -8,22 +8,17 @@ const {
 } = require('../transformers')
 
 async function getInvestmentProjectsCollection (req, res, next) {
-  const searchEntity = 'investment_project'
-
   try {
     res.locals.results = await search({
       searchTerm: '',
```
teneightfive: Do you need to pass an empty string here by the way? Option for more 🔴?

tyom: You're really good at 🔴

ztolley:
If you goto that URL you see the data returned for audit log does not include an entity type in the response. Maybe audit will be ok as it doesn't need a link. Just need you to confirm it works

tyom:
> If you goto that URL you see the data returned for audit log does not include an entity type in the response. Maybe audit will be ok as it doesn't need a link. Just need you to confirm it works

Entity type should be set in item transformers.

PR #517, opened 2017-08-31
Tweak stacked meta list variation

teneightfive:
Reduces the font size of the meta items when stacked and fixes
their position on smaller resolutions.

Also removes the colon when stacked.

### Before

image

### After

image

### Before

image

### After

image


assets/stylesheets/components/_meta-list.scss
```diff
@@ -48,8 +48,21 @@
 }
 
 .c-meta-list__item--stacked {
+  @include core-font(16);
```
tyom: Should `stacked` be less prescriptive in terms of its font size? What if it's used in another context where font size is set by a container?

teneightfive: Yeah, I was thinking that. Maybe the font-size should be set if it's within `c-local-header`?

PR #519, opened 2017-08-31
Local header actions

teneightfive:
This change adds support for an _actions_ area on the right hand side of the local header component.

## Example

image

### DIFF
--- a/assets/stylesheets/_deprecated/components/_button.scss
+++ b/assets/stylesheets/_deprecated/components/_button.scss
@@ -50,3 +50,8 @@
 .button--compact {
   padding: 0;
 }
+
+.button--small {
+  @include core-font(16);
+  padding: .3em .6em .15em;
+}
--- a/assets/stylesheets/components/_local-header.scss
+++ b/assets/stylesheets/components/_local-header.scss
@@ -1,6 +1,7 @@
 @import "../tools/mixins";
 
 .c-local-header {
+  @include clearfix;
   padding-top: $default-spacing-unit * 4;
   position: relative;
 
@@ -47,6 +48,38 @@
 * + .c-local-header__heading-after {
   margin-top: $default-spacing-unit / 2;
   margin-bottom: 0;
+}
+
+.c-local-header__content {
+  @include media(tablet) {
+    float: left;
+    width: 70%;
+  }
+}
+
+.c-local-header__actions {
+  * + & {
+    margin-top: $default-spacing-unit;
+  }
+
+  @include media(tablet) {
+    float: right;
+    width: 30%;
+
+    * + & {
+      margin-top: 0;
+    }
+  }
+}
+
+.c-local-header__action {
+  @include core-font(16);
+  display: block;
+  margin: 0;
+
+  & + & {
+    margin-top: $default-spacing-unit / 2;
+  }
 }
 
 .c-local-header--light-banner {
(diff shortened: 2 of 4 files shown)


assets/stylesheets/_deprecated/components/_button.scss
```diff
@@ -50,3 +50,8 @@
 .button--compact {
   padding: 0;
 }
+
+.button--small {
+  @include core-font(16);
```
feedmypixel: why's this in deprecated? can it not be placed in non deprecated file?

assets/stylesheets/components/_local-header.scss
```diff
@@ -49,6 +50,26 @@
   margin-bottom: 0;
 }
 
+.c-local-header__content {
+  @include media(tablet) {
+    float: left;
+    width: 70%;
```
feedmypixel: not a massive fan of floats for layout but I feel that just personal preference. I prefer `display: table` and `display:table-cell`

src/templates/_macros/common.njk
```diff
@@ -60,6 +60,7 @@
  # @param {string} [props.heading] - page heading
  # @param {string} [props.headingBefore] - a string before heading (safe to render HTML)
  # @param {string} [props.headingSuffix] - a string suffix for heading (safe to render HTML)
+ # @param {object} [props.actions] - a set of actions that will appear on the right hand side of the component
```
feedmypixel: maybe worth removing "that will appear on the right hand side of the component" as this is only true `> tablet`

assets/stylesheets/_deprecated/components/_button.scss
teneightfive: My guess is its in deprecated when we pulled the tradeelements dependency in. This however hasn't been refactored yet but there have been cases where we've needed to extend the button styles slightly.

I opted for this rather than refactor all of the buttons across the app for now. We could move it out and use BEM syntax as a separate change though.

assets/stylesheets/components/_local-header.scss
teneightfive: I gave it a go with tables and it feels like you need to mess around with more properties to get the same effect. You would need to apply `display: table` to the current `.l-container` `div` which would probably mean creating another element selector (`.c-local-header__container`) to apply that style.

Then you also have to ensure the element with display table has `width: 100%` and it started to feel more overloaded.

assets/stylesheets/_deprecated/components/_button.scss
feedmypixel: > We could move it out and use BEM syntax as a separate change though.

sounds good 👍

teneightfive: I took a brief look into this at the weekend and realised it's a pretty big change 😞 

Basically because the current method is tied in to both the govuk_frontend_toolkit and govuk_elements so it would mean quite a bit of re-writing to remove both those deps. 

Or alternatively we could still keep the govuk_frontend_toolkit dependency but it felt like a half-way fix so didn't warrant the effort right now.

So in summary, I think we leave that until we get time to properly address the structure of the sass across the board.

feedmypixel: Totally agreed - its one of the reasons that the button file is still in here. A lot of `govuk_frontend_toolkit` and `govuk_elements` are wound up with each other.

Something post November I feel

PR #522, opened 2017-09-01
Primary navigation

teneightfive:
This adds an initial piece of primary navigation to navigate to key areas or collections within the site.

It uses a [similar pattern](https://home-office-digital-patterns.herokuapp.com/patterns/navigation) being developer by the Home Office as a base but using the code style set out in this project.

It creates a basic stacked version on smaller resolutions but a better solution for smaller screens which reduces the navigation might need to be developed further down the line.

**Note:** No _home_ or _dashboard_ link has been added for 2 reasons at the moment. 1) the jump when it was highlight on the homepage to not being highlighted when searching seemed confusing and 2) the breadcrumb provides a clear route back to _Home_ and having the same link duplicated seemed unnecessary.

👉 [**DEMO**](http://datahub-beta2-pr-522.herokuapp.com/investment-projects) 👈 

## What it looks like

image

---

image

---

image

---

image

---

image

---

image

### DIFF
--- a/src/middleware/locals.js
+++ b/src/middleware/locals.js
@@ -1,3 +1,5 @@
+const path = require('path')
+
 const logger = require('../../config/logger')
 const config = require('../../config')
 
@@ -7,6 +9,13 @@ try {
 } catch (err) {
   logger.error('Manifest file is not found. Ensure assets are built.')
 }
+
+const globalNavItems = [
+  { path: '/companies', label: 'Companies' },
+  { path: '/contacts', label: 'Contacts' },
+  { path: '/investment-projects', label: 'Investment projects' },
+  { path: '/omis', label: 'OMIS Orders' },
+]
 
 module.exports = function locals (req, res, next) {
   const baseUrl = `${(req.encrypted ? 'https' : req.protocol)}://${req.get('host')}`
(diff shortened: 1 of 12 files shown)


feedmypixel:
feels like the bottom border isn't needed. Maybe remove it and add a touch more `margin-bottom`?

src/middleware/locals.js
```diff
@@ -20,6 +29,21 @@ module.exports = function locals (req, res, next) {
     BREADCRUMBS: breadcrumbItems,
     IS_XHR: req.xhr,
     QUERY: req.query,
+    GLOBAL_NAV_ITEMS: globalNavItems.map(item => {
+      const url = path.resolve(req.baseUrl, item.path)
+      let isActive
+
+      if (url === '/') {
```
feedmypixel: is the url ever `/` here?

src/templates/_layouts/datahub-base.njk
```diff
@@ -73,6 +63,13 @@
 
 {% block body_main_header %}
   {{ super() }}
+
+  {% if user %}
+    {{ GlobalNav({
+      items: GLOBAL_NAV_ITEMS
```
feedmypixel: do we need to pass this in? Is this for testing purposes?

src/middleware/locals.js
teneightfive: Good spot. That was catering for when the _Home_ link was in the nav. Maybe we can remove this seeing as we're not including it at this stage.

PR #525, opened 2017-09-01
Initial quote preview route

teneightfive:
This adds a placeholder view and route for the preview and generate
quote logic that will follow.

It makes a few amends to the Order model and existing macros to allow this to happen too.

### DIFF
--- a/assets/stylesheets/_deprecated/components/_button.scss
+++ b/assets/stylesheets/_deprecated/components/_button.scss
@@ -10,6 +10,10 @@
   }
 }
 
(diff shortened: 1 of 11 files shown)


assets/stylesheets/_deprecated/components/_button.scss
```diff
@@ -10,6 +10,10 @@
   }
 }
 
+.button--destructive {
+  @include button ($red);
```
feedmypixel: is there meant to be an extra space here?

PR #526, opened 2017-09-04
DH-537 Add event sub app

yeahfro:
This change adds the event sub app.

screen shot 2017-09-04 at 09 30 19

### DIFF
--- a/src/apps/events/router.js
+++ b/src/apps/events/router.js
@@ -1,0 +1,7 @@
+const router = require('express').Router()
+
+const { renderEventPage } = require('./controllers/edit')
+
+router.get('/create', renderEventPage)
+
+module.exports = router
--- a/src/apps/events/views/edit.njk
+++ b/src/apps/events/views/edit.njk
@@ -1,0 +1,9 @@
+{% extends "_layouts/two-column.njk" %}
+
+{% block main_grid_left_column %}{% endblock %}
+
+{% block main_grid_right_column %}
+
+
+  
+{% endblock %}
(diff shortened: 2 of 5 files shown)


src/apps/events/router.js
```diff
@@ -0,0 +1,7 @@
+const router = require('express').Router()
+
+const { renderEventPage } = require('./controllers/edit')
+
+router.get('/create', renderEventPage)
```
ztolley: I know this has no real stuff in it yet, but ..

Maybe copy the naming convention used in contacts, like

router
  .route('/create')
  .get(editDetails)
  .post(postDetails)

yeahfro: I took the naming convention from investment-projects and the list controller. I agree it could be changed but are we all in agreement that it should be `editDetails` over `renderEditEventPage`?

feedmypixel: @ztolley IMO `renderEventPage` is fine. There has been a move to `render<something>` when it is the last thing in the chain and `get`/`post` handler when it is one of those. [example](https://github.com/uktrade/data-hub-frontend/blob/develop/src/apps/investment-projects/controllers/create/investment-type.js)

src/apps/events/views/edit.njk
```diff
@@ -0,0 +1,9 @@
+{% extends "_layouts/two-column.njk" %}
+
+{% block main_grid_left_column %}{% endblock %}
+
+{% block main_grid_right_column %}
+
+
```
feedmypixel: whitespace can be removed

feedmypixel:
@yeahfro may also be worth adding a test for this PR, so the skeleton for that is also setup

src/apps/events/router.js
ztolley: Fair enough

test/unit/apps/events/controllers/edit.test.js
```diff
@@ -0,0 +1,33 @@
+describe('Event edit controller', () => {
```
feedmypixel: 👍

PR #530, opened 2017-09-04
add primary contact badge to contacts

feedmypixel:
Add the primary contact badge to contacts search results and collections list.

### Contacts search Results
screen shot 2017-09-04 at 18 12 11

### Contacts collection list
screen shot 2017-09-04 at 18 11 45

### DIFF
--- a/src/apps/contacts/transformers.js
+++ b/src/apps/contacts/transformers.js
@@ -13,6 +13,7 @@ function transformContactToListItem ({
   archived_on,
   archived_reason,
   company_sector,
+  primary,
 } = {}) {
   if (!id || !first_name || !last_name) { return }
 
@@ -35,13 +36,25 @@ function transformContactToListItem ({
         label: 'Sector',
         value: get(company_sector, 'name'),
       },
-      {
-        label: 'Country',
-        type: 'badge',
-        value: get(address_country, 'name'),
-      },
     ],
   }
+
+  // Add Contact type as first badge to be displayed
+  if (primary) {
+    item.meta.push({
+      label: 'Contact type',
+      value: 'Primary',
+      type: 'badge',
+      badgeModifier: 'secondary',
+    })
+  }
+
+  item.meta.push(
+    {
+      label: 'Country',
+      type: 'badge',
+      value: get(address_country, 'name'),
+    })
 
   if (archived_reason) {
     item.meta.push({
--- a/test/unit/apps/companies/controllers/contacts.test.js
+++ b/test/unit/apps/companies/controllers/contacts.test.js
@@ -221,7 +221,7 @@ describe('Company contacts controller', function () {
       expect(contact).to.have.property('id')
       expect(contact).to.have.property('type', 'contact')
       expect(contact).to.have.property('name', 'Fred Smith')
-      expect(contact).to.have.property('meta').to.be.an('array').to.have.length(3)
+      expect(contact).to.have.property('meta').to.be.an('array').to.have.length(4)
     })
 
     it('should return the required fields to list archived contacts', function () {
@@ -230,7 +230,7 @@ describe('Company contacts controller', function () {
       expect(contact).to.have.property('type', 'contact')
       expect(contact).to.have.property('name', 'Jane Smith')
       expect(contact).to.have.property('isArchived').to.be.true
-      expect(contact).to.have.property('meta').to.be.an('array').to.have.length(5)
+      expect(contact).to.have.property('meta').to.be.an('array').to.have.length(6)
     })
 
     it('should return a link to add a new contact', function () {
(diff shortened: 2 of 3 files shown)


assets/stylesheets/components/_badge.scss
```diff
@@ -3,7 +3,7 @@
 @mixin _style-badge($colour: $text-colour, $hover: $colour) {
   border: 2px solid rgba($text-colour, .15);
   border-radius: 4px;
-  color: inherit;
+  color: $colour;
```
teneightfive: I would hesitate in changing this. I believe the colour was to signify that it can be clicked to narrow down the filter on a collection.

In search results (at least for the moment) we don't have filtering so the colour would be confusing.

teneightfive: I would hesitate in changing this. I believe the colour was to signify that it can be clicked to narrow down the filter on a collection.

In search results (at least for the moment) we don't have filtering so the colour would be confusing.

assets/stylesheets/components/_badge.scss
```diff
@@ -50,3 +50,7 @@
 .c-badge--secondary {
   @include _style-badge($purple);
 }
+
+.c-badge--info {
```
teneightfive: Maybe stick with the current theme of `primary` (implicit) and `secondary` (explicit) and use `--tertiary`?

It doesn't look like we need a third colour yet though from the screenshot?

src/apps/contacts/transformers.js
```diff
@@ -36,6 +37,12 @@ function transformContactToListItem ({
         value: get(company_sector, 'name'),
       },
       {
+        label: 'Primary Contact',
+        type: 'badge',
+        value: primary ? 'Primary' : '',
```
teneightfive: Maybe use the method used below for adding meta items if a value is present:

```js
if (primary) {
  item.meta.push({
    label: 'Primary contact',
    value: 'Primary',
    type: 'badge',
    badgeModifier: 'secondary',
  })
}
```

teneightfive: A bigger question though is for the need of this badge? I questioned this originally when Tyom added the initial listing and I don't see the relevance at this level as seeing lots of _Primary_ labels surely voids the whole purpose of having a single primary contact?

It makes sense during a listing of contacts just for one company but in search of the default collection filter I think it causes more confusion that benefit. Is there a specific user need to see primary contacts when searching or filtering?

ztolley: Primary contacts are important in search results so that you can see and select the most important contact for a company. This is existing behaviour for a reason, if we want to question if it should be there then that should not be in the scope of this card.

assets/stylesheets/components/_badge.scss
teneightfive: I would look at how the investment projects collection handles these badges and allows them to be clickable to narrow down the search to that particular value rather than this change.

src/apps/contacts/transformers.js
ztolley: I take it back. It was never in the search results.

jim68000: Primary has been requested by PM & DM for this ticket. If we have issues around whether the user needs exist for this can we a> add it so users can tell us and b> talk to Ian and/or Nandu if we're unhappy about that.

assets/stylesheets/components/_badge.scss
feedmypixel: have reverted this work as I agree with the above 👍

assets/stylesheets/components/_badge.scss
feedmypixel: Have removed this modifier, no need for it

src/apps/contacts/transformers.js
```diff
@@ -43,6 +44,15 @@ function transformContactToListItem ({
     ],
   }
 
+  if (primary) {
```
teneightfive: Just looking at the screenshot I wonder if it might be worth playing with the order a little so that the country doesn't jump around when this is/isn't present?

src/apps/contacts/transformers.js
```diff
@@ -43,6 +44,15 @@ function transformContactToListItem ({
...
+    item.meta.push({
+      label: 'Primary contact',
```
teneightfive: Probably not a massive deal but `Primary contact: Primary` feels like it reads funny 🤔 

What about `Contact type` or just `Type` for the label?

src/apps/contacts/transformers.js
feedmypixel: - Primary contact doesn't really mean anything
- You can have multiple Primary contacts on a company
- People use it in different ways
- It adds unnecessary noise to the UI
- You cannot search for a Primary contact

as @madgoose mentioned and as @teneightfive has touched on if we don't include it we can then learn and see if the users require it. Then build it.

src/apps/contacts/transformers.js
```diff
@@ -21,28 +22,39 @@ function transformContactToListItem ({
...
+    item.meta.push({
+      label: 'Contact type',
+      value: 'Primary',
+      type: 'badge',
+      badgeModifier: 'secondary',
+    })
+  }
+
+  item.meta.push(
+    {
```
teneightfive: Wonder if you need to pull all of these out of the previous initial definition. It only really needs to sit in font of the `country` badge. 

So `company`, `archived_on` and `sector` could all remain in the initial definition.

PR #531, opened 2017-09-04
Generate a quote

teneightfive:
This adds the controller and middleware logic to handle generating
a quote from an order.

Has some basic error handling on API responses but more detailed
user errors will come later.

### DIFF
--- a/src/apps/omis/apps/view/middleware.js
+++ b/src/apps/omis/apps/view/middleware.js
@@ -21,7 +21,43 @@ async function getQuote (req, res, next) {
   }
 }
 
+async function generateQuote (req, res, next) {
+  const orderId = res.locals.order.id
+
+  try {
+    await Order.createQuote(req.session.token, orderId)
+
+    req.flash('success', 'Quote successfully generated.')
+    res.redirect(`/omis/${orderId}`)
+  } catch (error) {
+    const errorCode = error.statusCode
+
+    if (errorCode === 400) {
+      req.flash('error', 'Quote could not be generated. Some fields were missing.')
+      return res.redirect(`/omis/${orderId}`)
+    }
+
+    if (errorCode === 409) {
+      req.flash('error', 'Quote could not be generated. A valid quote already exists.')
+      return res.redirect(`/omis/${orderId}`)
+    }
+    next(error)
+  }
+}
+
+function setQuoteForm (req, res, next) {
+  res.locals.quoteForm = {
+    disableFormAction: false,
+    buttonText: 'Generate and send quote',
+    returnText: 'Return to order',
+    returnLink: `/omis/${res.locals.order.id}`,
+  }
+  next()
+}
+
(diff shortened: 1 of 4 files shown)


src/apps/omis/apps/view/controllers.js
```diff
@@ -8,12 +8,18 @@ function renderWorkOrder (req, res) {
...
-    .breadcrumb('Work order')
     .render('omis/apps/view/views/work-order', {
       values,
     })
 }
 
+function renderQuote (req, res) {
+  res
+    .breadcrumb('Quote')
+    .render('omis/apps/view/views/quote')
```
feedmypixel: can the naming for this folder structure `/view/views` be improved?

src/apps/omis/apps/view/middleware.js
```diff
@@ -1,11 +1,59 @@
...
+      return next()
+    }
+
+    next(error)
+  }
+}
+
+async function generateQuote (req, res, next) {
+  try {
+    await Order.createQuote(req.session.token, res.locals.order.id)
```
feedmypixel: `res.locals.order.id` could be put into a local const as its used twice

src/apps/omis/apps/view/views/_layouts/view.njk
```diff
@@ -18,9 +15,8 @@
 
     {{ MetaList({
       items: [
-        { "label": "Status", "value": "Draft" },
-        { "label": "Created on", "value": order.created_on | formatDate },
-        { "label": "Created by", "value": order.created_by }
+        { "label": "Created on", "value": order.created_on | formatDateTime },
```
feedmypixel: I presumed the quotes in nunjucks for non html work were single? (html is double) I may be wrong

teneightfive: I'm not sure to be honest. I probably haven't been very good at using a standard set within nunjucks files. Will try and use this pattern moving forward.

assets/stylesheets/_deprecated/components/_button.scss
```diff
@@ -10,6 +10,10 @@
   }
 }
 
+.button--destructive {
+  @include button ($red);
```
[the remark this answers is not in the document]
teneightfive: This was a hangover from the branch is was rebased on top of (#525). This change was fixed there and isn't present here any more :)

src/apps/omis/apps/view/controllers.js
teneightfive: It's not the best but `view` seemed to best describe the sub-app state. Happy for any alternative suggestions for the _view_ app.

Definitely outside of scope for this change though.

feedmypixel: > Definitely outside of scope for this change though.

agreed - something a little more semantic may help explain what it is (as discussed)

PR #532, opened 2017-09-04
breadcrumb and global nav tweaks

feedmypixel:
👉  [Demo here](https://datahub-beta2-pr-532.herokuapp.com/sign-in) 👈 

A few suggestions/tweaks to the styles and markup associated with the breadcrumb and global nav. This removes the need for sending in modifier flags to the macro.

- move setting of background colour for header to wrapper
- make font size for global nav a bit bigger
- remove border underline

header


feedmypixel:
After discussion with @teneightfive closing this

PR #535, opened 2017-09-05
Cancel quote

teneightfive:
This change adds the initial journey for cancelling a quote that has been sent for approval.

It can be cancelled as long as it hasn't been accepted or previously cancelled. There are 2 slightly different variations to the cancellation too: 

1) **Not expired**: Can be cancelled but a warning is provided that it will notify the client and they will no longer be able to accept that quote
2) **Expired**: Can be cancelled without major consequence as client can no longer accept anyway

**Note:** The API doesn't currently return the `expires_on` property of the quote object so the functionality won't work on the demo link at this point.

### Not expired

localhost_3001_omis_6c4f3d1b-b9ad-4d27-a2ff-94a53167f86f_quote

### Expired

localhost_3001_omis_6c4f3d1b-b9ad-4d27-a2ff-94a53167f86f_quote 1

### DIFF
--- a/src/apps/omis/apps/view/middleware.js
+++ b/src/apps/omis/apps/view/middleware.js
@@ -45,13 +45,53 @@ async function generateQuote (req, res, next) {
   }
 }
 
+async function cancelQuote (req, res, next) {
+  try {
+    await Order.cancelQuote(req.session.token, res.locals.order.id)
+
+    req.flash('success', 'Quote successfully cancelled.')
+    res.redirect(`/omis/${res.locals.order.id}`)
+  } catch (error) {
+    if (error.statusCode === 404) {
+      req.flash('error', 'The quote does not exist so could not be cancelled.')
+      return res.redirect(`/omis/${res.locals.order.id}/quote`)
+    }
+
+    if (error.statusCode === 409) {
+      req.flash('error', 'Quote could not be cancelled. It has already been accepted.')
+      return res.redirect(`/omis/${res.locals.order.id}/quote`)
+    }
+
+    next(error)
+  }
+}
+
 function setQuoteForm (req, res, next) {
-  res.locals.quoteForm = {
-    disableFormAction: false,
+  const quote = res.locals.quote
+  const orderId = res.locals.order.id
+  const form = {
     buttonText: 'Generate and send quote',
     returnText: 'Return to order',
-    returnLink: `/omis/${res.locals.order.id}`,
+    returnLink: `/omis/${orderId}`,
   }
+
+  if (quote.expires_on) {
+    form.action = `/omis/${orderId}/quote/cancel`
+    form.buttonText = 'Cancel quote'
+    form.buttonModifiers = 'button-secondary'
+
+    if (quote.accepted_on || quote.cancelled_on) {
+      form.disableFormAction = true
+    }
+
+    if (new Date(quote.expires_on) > new Date()) {
+      form.buttonModifiers = 'button--destructive'
+
+      res.locals.destructive = true
+    }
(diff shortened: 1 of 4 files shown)


src/apps/omis/apps/view/middleware.js
```diff
@@ -45,19 +45,60 @@ async function generateQuote (req, res, next) {
...
+  if (quote.expires_on) {
+    form.action = `/omis/${orderId}/quote/cancel`
+    form.buttonText = 'Cancel quote'
+    form.buttonModifiers = 'button-secondary'
+
+    if (quote.accepted_on || quote.cancelled_on) {
+      form.disableFormAction = true
+    }
+
+    if (new Date(quote.expires_on) > new Date()) {
```
feedmypixel: this could go in `const` and be sent to the view `expiresInTheFuture = new Date(quote.expires_on) > new Date()` and then the 'button--destructive' modifier can be added in the view?

src/apps/omis/apps/view/middleware.js
```diff
@@ -45,19 +45,60 @@ async function generateQuote (req, res, next) {
...
+      form.buttonModifiers = 'button--destructive'
+
+      res.locals.destructive = true
```
feedmypixel: you could use `expiresInTheFuture` const for this also

src/apps/omis/apps/view/middleware.js
```diff
@@ -45,19 +45,60 @@ async function generateQuote (req, res, next) {
...
 
+async function cancelQuote (req, res, next) {
+  try {
+    await Order.cancelQuote(req.session.token, res.locals.order.id)
+
+    req.flash('success', 'Quote successfully cancelled.')
+    res.redirect(`/omis/${res.locals.order.id}`)
+  } catch (error) {
+    if (error.statusCode === 404) {
+      req.flash('error', 'The quote does not exist so could not be cancelled.')
```
feedmypixel: should this have a comma? `The quote does not exist, so could not be cancelled.`

teneightfive: Yes, possibly. I have a feeling these messages will be changing shortly though.

src/apps/omis/apps/view/middleware.js
teneightfive: This piece of middleware is actually building the form object that is passed in to the `Form()` macro so I've tried to keep all the configuration of that in this middleware. 

@tyom and I have been discussing reducing the configuration of macros within views so this is trying to carry on down that road.

src/apps/omis/apps/view/middleware.js
teneightfive: I used the name `destructive` because it may be used for different scenarios and now just when it is expired.

src/apps/omis/apps/view/middleware.js
feedmypixel: There maybe a sweet spot here, it feels that adding modifiers would be better in the view and potentially having strings hidden in `js` files doesn't feel quite right. Something worth revisiting.

src/apps/omis/apps/view/middleware.js
feedmypixel: _temporary is never temporary_

src/apps/omis/apps/view/middleware.js
teneightfive: I suppose the idea is that we move away from multiple views. And the view just receives a macros object which it renders.

It would be nicer in future to only have 5/6 templates rather than 3/4 templates and hundreds of views with lots of macro repetition.

PR #536, opened 2017-09-05
Added ability for pluralize to handle y -> ies

jim68000:
Before: 
image

After:
image

### DIFF
--- a/config/nunjucks/filters.js
+++ b/config/nunjucks/filters.js
@@ -34,6 +34,8 @@ function pluralise (string, count, pluralisedWord) {
   if (parseInt(count, 10) !== 1) {
     if (pluralisedWord) {
       string = pluralisedWord
+    } else if (string.match(/[^aeiou]y$/)) {
+      string = string.replace(/y$/, 'ies')
     } else {
       string += 's'
     }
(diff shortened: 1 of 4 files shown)


feedmypixel:
@jim68000 sozz you can already do this via the [pluralisedWord](https://github.com/uktrade/data-hub-frontend/blob/develop/config/nunjucks/filters.js#L35)

To see the pluralisation of `company` to `companies` see this [test](https://github.com/uktrade/data-hub-frontend/blob/develop/test/unit/config/nunjucks/filters.test.js#L124-L125)

jim68000:
Ah yeah, but that relies on someone actually remembering to do that and we have a surprising number of things that end in Y. If it's not useful, it's not useful, but perhaps people could spellcheck the strings in their code before committing?

image

teneightfive:
@jim68000 I like this 👍 

@feedmypixel I'm wondering whether we add this support and maybe revert the cases where we've had to specifically pass the word as an argument. It would be more robust and we can still support the `pluralisedWord` for edge cases?

feedmypixel:
> I'm wondering whether we add this support and maybe revert the cases where we've had to specifically pass the word as an argument.

Sounds good. 

@jim68000 can we also include the removal of 

> the cases where we've had to specifically pass the word as an argument

in this PR as a commit?

PR #539, opened 2017-09-06
Add support for markdown in nunjucks

teneightfive:
This adds a custom tag for supporting markdown in nunjucks views.

It also adds some very basic style overrides for content if surrounded
by a markdown class.


assets/stylesheets/trumps/_layout.scss
```diff
@@ -24,3 +24,36 @@
...
+  h1 {
+    @include bold-font(24);
+  }
+
+  h2 {
+    @include bold-font(20);
+  }
+
+  h3 {
+    @include bold-font(20);
```
yeahfro: Slight hesitation around h2 and h3 being the same.

teneightfive: Have amended so that `h1`s and `h2`s use the next font sizes up.

PR #543, opened 2017-09-06
Update OMIS model methods

teneightfive:
Contains some updates to the OMIS order model:

- Use correct method for getting quote preview (`POST` instead of `GET`)
- Add method on OMIS order model to support full quote

### DIFF
--- a/src/apps/omis/models.js
+++ b/src/apps/omis/models.js
@@ -65,6 +65,10 @@ const Order = {
     return authorisedRequest(token, `${config.apiRoot}/v3/omis/order/${id}/quote`)
   },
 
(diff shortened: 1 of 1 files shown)


src/apps/omis/models.js
```diff
@@ -55,7 +55,10 @@ const Order = {
   },
 
   getQuotePreview (token, id) {
```
yeahfro: is it still correct for this to be `getQuotePreview`?

teneightfive: Yep, it will be used in #544. There are endpoints for both the `quote` and the `preview` of the quote.

PR #544, opened 2017-09-06
Preview order quote

teneightfive:
This PR includes changes to the quote preview view.

It adds support for previewing the contents of the quote before it is sent and seeing the contents after it has been sent.

### Quote cannot be generated - missing information

localhost_3001_omis_3456238e-e94f-4db0-889f-437eb6ac8377_quote

### Quote can be generated

localhost_3001_omis_3456238e-e94f-4db0-889f-437eb6ac8377_quote 1

### Cancel quote (not expired)

localhost_3001_omis_3456238e-e94f-4db0-889f-437eb6ac8377_quote 2

### Cancel quote (expired)

localhost_3001_omis_3456238e-e94f-4db0-889f-437eb6ac8377_quote 3

### DIFF
--- a/src/apps/omis/apps/view/middleware.js
+++ b/src/apps/omis/apps/view/middleware.js
@@ -23,17 +23,30 @@ function setTranslation (req, res, next) {
 }
 
 async function getQuote (req, res, next) {
+  const orderId = get(res.locals, 'order.id')
+
   try {
-    const orderId = get(res.locals, 'order.id')
-    res.locals.quote = await Order.getQuote(req.session.token, orderId)
-    next()
+    res.locals.quote = await Order.previewQuote(req.session.token, orderId)
+    return next()
   } catch (error) {
-    if (error.statusCode === 404) {
-      res.locals.quote = {}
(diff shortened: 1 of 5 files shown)


src/apps/omis/apps/view/middleware.js
```diff
@@ -1,26 +1,52 @@
...
+  const orderId = get(res.locals, 'order.id')
+
   try {
-    const orderId = get(res.locals, 'order.id')
-    res.locals.quote = await Order.getQuote(req.session.token, orderId)
-    next()
+    res.locals.quote = await Order.previewQuote(req.session.token, orderId)
+    return next()
   } catch (error) {
-    if (error.statusCode === 404) {
```
feedmypixel: This is starting to feel a little confused and feels like its jumping around a little in the code. 
Current code flow:
- try and get a preview
- if it errors in a certain way try and get a full quote
- if it errors in another way pass missing data to view

Something to have a think about, could it become:
- does preview exist?
  - get preview?
  - handle errors
- does quote exist?
  - get quote
  - handle errors

PR #553, opened 2017-09-07
Add items javascript enhancement

teneightfive:
This makes some updates to the way the add items module is structured to bring it more in line with the existing modules.

## Possible future improvements

- Cater for there being no starting template
- Check where keyboard focus lands after updating
- Update screenreaders using ARIA API


assets/javascripts/_deprecated/lib/element-stuff.js
```diff
@@ -134,24 +134,29 @@ function removeElement (element) {
...
-  Array
-    .from(fragment.querySelectorAll('input[type="text"], textarea'))
+  Array.from(element.querySelectorAll('input[type="text"], textarea'))
     .forEach(textField => {
       textField.value = ''
     })
+
+  Array.from(element.querySelectorAll('select'))
+    .forEach(field => {
+      console.log('resetting')
```
feedmypixel: 👮 🚨

assets/javascripts/_deprecated/lib/element-stuff.js
```diff
@@ -134,24 +134,29 @@ function removeElement (element) {
...
+      field.selectedIndex = 0
```
feedmypixel: this could be a 1 liner

assets/javascripts/_deprecated/lib/element-stuff.js
```diff
@@ -181,6 +186,8 @@ function regenIds (wrapper) {
         relatedLabel.setAttribute('for', newId)
       }
     })
+
+  return wrapper
```
feedmypixel: can you update the JSDOC to say it now returns the wrapper

assets/javascripts/_deprecated/lib/element-stuff.js
teneightfive: Doh 🤦‍♂️

assets/javascripts/modules/add-items.js
```diff
@@ -0,0 +1,197 @@
+const { pickBy } = require('lodash')
+const {
+  insertAfter,
+  resetFieldValues,
+  regenIds,
+  closest,
+  removeElement,
+} = require('../_deprecated/lib/element-stuff')
+
+const addButtonClass = 'js-AddItems__add'
```
feedmypixel: not sure we need to be using `BEM` syntax for `js` hooks wouldn't the following work?
- `js-add-items`
- `js-remove-items`

feedmypixel: I'd also keep things kebab-case and therefore lowercase

assets/javascripts/modules/add-items.js
```diff
@@ -0,0 +1,197 @@
...
+ *   <div class="c-form-group js-adviser">
+ *     <label for="adviser">Adviser</label>
+ *     <select name="adviser">
+ *       <option value="">---Select Adviser--</option>
+ *       ...
+ *     </select>
+ *   </div>
+ * </div>
+ */
+const AddItems = {
```
feedmypixel: is this not `AddItem`? as it only adds one thing at a time

assets/javascripts/modules/add-items.js
```diff
@@ -0,0 +1,197 @@
...
+    this.wrapper = wrapper
+    this.settings = Object.assign({}, this.defaults, options)
+
+    this.cacheEls()
+    this.bindEvents()
+    this.render()
+  },
+
+  cacheEls () {
+    const lastItem = this.wrapper.querySelector(`${this.settings.itemSelector}`)
```
feedmypixel: doesn't need to be a template string

assets/javascripts/modules/add-items.js
```diff
@@ -0,0 +1,197 @@
...
+  render () {
+    this.insertAddButton()
+    this.insertRemoveButtons()
+
+    this.decorateButtons()
+    this.updateButtonState()
+  },
+
+  decorateButtons () {
+    const removeButtons = Array.from(this.wrapper.querySelectorAll(this.settings.removeButtonSelector))
```
feedmypixel: technically `.length` and `.forEach` are available on a `NodeList` so no need for `Array.from` here

assets/javascripts/modules/add-items.js
```diff
@@ -0,0 +1,197 @@
...
+
+        const removeButtonElement = this.document.createElement('button')
+        removeButtonElement.className = `button button--link ${removeButtonClass}`
+
+        item.appendChild(removeButtonElement)
+      })
+  },
+
+  updateButtonState () {
+    const removeButtons = Array.from(this.wrapper.querySelectorAll(this.settings.removeButtonSelector))
```
feedmypixel: no need for `Array.from` here

assets/javascripts/modules/add-items.js
```diff
@@ -0,0 +1,197 @@
...
+  },
+
+  removeItem (item) {
+    item.parentNode.removeChild(item)
+  },
+}
+
+module.exports = {
+  init (page = document) {
+    const elements = Array.from(page.querySelectorAll('.js-AddItems'))
```
feedmypixel: no need for `Array.from` here

assets/javascripts/modules/add-items.js
teneightfive: The thought behind it was that it clearly ties it to only this module.

I'm not concerned with kebab-case for `js-` prefixed classes, in fact I prefer pascal case as it maps directly to the module definition. Using BEM syntax was a way to show it was a child selector of the main module.

Something like `js-remove-items` could have a similar use in different modules so prefixing with the module name separates them more clearly. Also, if you have multiple `js-` hooks on some markup it allows you to easily distinguish which is tied to which module.

assets/javascripts/modules/add-items.js
teneightfive: But the possibly to add multiple items at a time. The module doesn't only allow the adding of one item.

assets/javascripts/modules/add-items.js
teneightfive: Good spot.

assets/javascripts/modules/add-items.js
teneightfive: IE doesn't support `.forEach()` on `NodeList`. `NodeList` isn't a true `Array` its an Array-like object so that's why we need to wrap it in the `Array.from()` so that we can always expect those methods.

assets/javascripts/modules/add-items.js
teneightfive: Same comment as above about `NodeList` being an Array-like object.

assets/javascripts/modules/add-items.js
teneightfive: Same comment as above about `NodeList` being an Array-like object.

assets/javascripts/_deprecated/lib/element-stuff.js
teneightfive: I'm tempted to leave as is as follows the pattern of the rest of the method which doesn't use one-liners. Also, I've been trying to avoid them as a pattern in this app so far as they're sometimes harder to read.

assets/javascripts/modules/add-items.js
feedmypixel: it adds 1 thing that could contain anything right? unless I'm misunderstanding?

assets/javascripts/modules/add-items.js
feedmypixel: ah yeah forgot about `IE` 😞

assets/javascripts/modules/add-items.js
teneightfive: The idea of the module is to add _items_ to a container. It adds one thing each click but the module as a whole is responsible for adding or removing items from a list.

Maybe something like `RepeatableItems` might be better? Or `AddRemoveItems`?

feedmypixel: > the module as a whole is responsible for adding or removing multiple items to/from a list rather than just one.

AddItems is fine 👍

PR #554, opened 2017-09-08
Return all advisers from API in OMIS

teneightfive:
This uses a different method to return all the advisers rather than
a limited amount for OMIS.

This is not an ideal solution as it can add a lot of time to API
requests but it avoids advisers not being found in the list for
existing orders that are being migrated over.

## Problems this adds

- Long page request times when returning all advisers

## Problems it solves

- Being able to select advisers that aren't in the first 100 results
- Migrated orders not having the correct adviser selected as it is not in the list, then on save it removes the advisers from the order

## What hasn't been done

This has only been done in OMIS for now to test legacy orders so won't effect other parts of data hub.

### DIFF
--- a/src/apps/adviser/repos.js
+++ b/src/apps/adviser/repos.js
@@ -6,6 +6,12 @@ function getAdvisers (token) {
   return authorisedRequest(token, `${config.apiRoot}/adviser/`)
 }
 
+// TODO: Make sure this is removed and replaced with a better way to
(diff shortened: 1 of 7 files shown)


teneightfive:
>@teneightfive can you stick an issue and a //TODO in the code please so we can fix this before live.

@feedmypixel 👍 

See #558.

PR #555, opened 2017-09-08
Add interaction transformer and add interactions to search results

ztolley:
screenshot from 2017-09-08 14-04-29

Add interactions to search results
Created a transformer for interactions to use in generic lists.

### DIFF
--- a/src/apps/interactions/transformers.js
+++ b/src/apps/interactions/transformers.js
@@ -1,6 +1,4 @@
 /* eslint-disable camelcase */
-const dateFns = require('date-fns')
-const { mediumDateTimeFormat } = require('../../../config')
 
 function transformInteractionToListItem (interaction) {
   return {
@@ -9,12 +7,12 @@ function transformInteractionToListItem (interaction) {
     name: interaction.subject,
     meta: [
       {
-        label: 'Type:',
+        label: 'Type',
         type: 'badge',
         value: interaction.interaction_type,
       },
       {
-        label: 'Contact:',
+        label: 'Contact',
         value: interaction.contact,
       },
       {
@@ -23,10 +21,11 @@ function transformInteractionToListItem (interaction) {
       },
       {
         label: 'Date',
-        value: dateFns.format(interaction.date, mediumDateTimeFormat),
+        value: interaction.date,
+        type: 'date',
       },
       {
-        label: 'Adviser:',
+        label: 'Adviser',
         value: interaction.dit_adviser,
       },
     ],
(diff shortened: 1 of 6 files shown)


src/apps/interactions/transformers.js
```diff
@@ -0,0 +1,38 @@
...
+        label: 'Contact:',
+        value: interaction.contact,
+      },
+      {
+        label: 'Company',
+        value: interaction.company,
+      },
+      {
+        label: 'Date',
+        value: dateFns.format(interaction.date, mediumDateTimeFormat),
```
teneightfive: The macro supports different formats from the view layer. You can change this to:

```js
{
  label: 'Date',
  value: interaction.date,
  type: 'date',
}
```

This avoids the date being formatted as `September 1970` if interaction.date doesn't have a value.

src/apps/interactions/transformers.js
```diff
@@ -0,0 +1,38 @@
...
+const { mediumDateTimeFormat } = require('../../../config')
+
+function transformInteractionToListItem (interaction) {
+  return {
+    id: interaction.id,
+    type: 'interaction',
+    name: interaction.subject,
+    meta: [
+      {
+        label: 'Type:',
```
teneightfive: Drop `:`

src/apps/interactions/transformers.js
```diff
@@ -0,0 +1,38 @@
...
+        type: 'badge',
+        value: interaction.interaction_type,
+      },
+      {
+        label: 'Contact:',
```
teneightfive: Drop `:`

src/apps/interactions/transformers.js
```diff
@@ -0,0 +1,38 @@
...
+      {
+        label: 'Company',
+        value: interaction.company,
+      },
+      {
+        label: 'Date',
+        value: dateFns.format(interaction.date, mediumDateTimeFormat),
+      },
+      {
+        label: 'Adviser:',
```
teneightfive: Drop `:`

src/apps/interactions/transformers.js
```diff
@@ -0,0 +1,37 @@
+/* eslint-disable camelcase */
+
+function transformInteractionToListItem (interaction) {
```
teneightfive: Just realised this uses a different format to other transformers.

Elsewhere we've defined the properties used as an object for the arg so its clear what is expected. So this would become:

```js
function transformInteractionToListItem ({
  id,
  subject,
  interaction_type,
  contact,
  company,
  date,
  dit_adviser,
}) {
  ...
}
```

PR #556, opened 2017-09-08
DH-545 Add another to event shared teams and programmes

yeahfro:
screen shot 2017-09-08 at 16 13 10

Changes give the ability to add another for fields:
- Event shared teams
- Programmes

Note this is a no JS version.

I was feeling confident, so special attention needs to be given to:
- `MultipleChoiceFieldAddAnother`
- My change to `conditional-subfields.js` - this was required because there is a conditional input that was losing its value each time visibility was toggled

### DIFF
--- a/assets/javascripts/modules/conditional-subfields.js
+++ b/assets/javascripts/modules/conditional-subfields.js
@@ -136,8 +136,10 @@ const ConditionalSubfields = {
       const children = subField.querySelectorAll('input, select, checkbox, textarea')
 
       Array.from(children).forEach((field) => {
-        field.value = ''
-        field.checked = false
+        if (!field.getAttribute('data-persist-values')) {
+          field.value = ''
+          field.checked = false
+        }
 
         const event = this.wrapper.createEvent('HTMLEvents')
         event.initEvent('change', true, false)
--- a/src/apps/components/views/form.njk
+++ b/src/apps/components/views/form.njk
@@ -196,4 +196,33 @@
     {% endcall %}
   {% endcall %}
 
+  {% call Example('Add another (macro)') %}
+    {% call Form(form) %}
+
+      {{ AddAnother({
+
+        buttonName: 'add_country',
+        name: 'country',
+        label: 'Country',
+        children: [
+          {
+            macroName: 'MultipleChoiceField',
+            name: 'country',
+            label: 'Country',
+            isLabelHidden: true,
+            optional: true,
+            initialOption: '-- Select country --',
+            options: [
+              { value: 1, label: 'country 1' },
+              { value: 2, label: 'country 2' },
+              { value: 3, label: 'country 3' }
+            ]
+          }
+        ]
+
+      }) }}
+
+    {% endcall %}
+  {% endcall %}
+
 {% endblock %}
--- a/src/apps/events/controllers/edit.js
+++ b/src/apps/events/controllers/edit.js
@@ -31,8 +31,9 @@ async function renderEventPage (req, res, next) {
 function postHandler (req, res, next) {
   const setAddAnotherField = (value) => compact(castArray(value))
   req.body.event_shared_teams = setAddAnotherField(req.body.event_shared_teams)
+  req.body.event_programmes = setAddAnotherField(req.body.event_programmes)
 
-  if (req.body.add_event_shared_team) {
+  if (req.body.add_event_shared_team || req.body.add_event_programme) {
     return next()
   }
 }
--- a/src/apps/events/controllers/edit.js
+++ b/src/apps/events/controllers/edit.js
@@ -2,7 +2,7 @@ const { eventFormConfig } = require('../macros')
 const { getAdvisers } = require('../../adviser/repos')
 const { transformObjectToOption } = require('../../transformers')
 const { buildFormWithState } = require('../../builders')
-const { castArray, compact } = require('lodash')
+const { get, castArray, compact } = require('lodash')
 
 async function renderEventPage (req, res, next) {
   try {
@@ -10,12 +10,12 @@ async function renderEventPage (req, res, next) {
     const advisers = advisersResponse.results.map(transformObjectToOption)
     const eventForm = eventFormConfig(advisers)
     const emptyDate = { year: '', month: '', day: '' }
-    const requestBody = {
-      'event-start-date': emptyDate,
-      'event-end-date': emptyDate,
-      'event-team-hosting': req.session.user.dit_team.id,
-    }
-    const eventFormWithState = buildFormWithState(eventForm, requestBody)
+    const body = Object.assign(req.body, {
+      event_start_date: emptyDate,
+      event_end_date: emptyDate,
+      event_team_hosting: get(req, 'session.user.dit_team.id', null),
+    })
+    const eventFormWithState = buildFormWithState(eventForm, body)
 
     res
       .breadcrumb('Add event')
--- a/src/apps/events/controllers/edit.js
+++ b/src/apps/events/controllers/edit.js
@@ -2,6 +2,7 @@ const { eventFormConfig } = require('../macros')
 const { getAdvisers } = require('../../adviser/repos')
 const { transformObjectToOption } = require('../../transformers')
 const { buildFormWithState } = require('../../builders')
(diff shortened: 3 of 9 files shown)


src/apps/events/controllers/edit.js
```diff
@@ -27,6 +27,21 @@ async function renderEventPage (req, res, next) {
   }
 }
 
+function postHandler (req, res, next) {
+  const setAddAnotherField = (value) => compact(castArray(value))
+  req.body['event-shared-teams'] = setAddAnotherField(req.body['event-shared-teams'])
+  req.body['event-programmes'] = setAddAnotherField(req.body['event-programmes'])
```
yeahfro: Tempted to move these lines up into renderEventPage - thoughts?

assets/javascripts/modules/conditional-subfields.js
```diff
@@ -136,8 +136,10 @@ const ConditionalSubfields = {
       const children = subField.querySelectorAll('input, select, checkbox, textarea')
 
       Array.from(children).forEach((field) => {
-        field.value = ''
-        field.checked = false
+        if (!children.getAttribute('data-persist-values')) {
```
ztolley: I discovered a thing for this. @teneightfive  if you set a fields disabled flag then it doesn't send it to the server. So you can show and hide a field and set disable to make sure it's value isn't set. The bonus is that you can hide and reveal and it would not loose the value you entered.

feedmypixel: is this check needed? its already being performed on line [#L135](https://github.com/uktrade/data-hub-frontend/blob/179fdc97c155bf74d6d6b00149d2f40465874263/assets/javascripts/modules/conditional-subfields.js#L135). So anything thats here already has had to pass the check `!subField.getAttribute('data-persist-values')`

src/apps/events/controllers/edit.js
```diff
@@ -2,19 +2,19 @@ const { eventFormConfig } = require('../macros')
...
     const advisers = advisersResponse.results.map(transformObjectToOption)
     const eventForm = eventFormConfig(advisers)
     const emptyDate = { year: '', month: '', day: '' }
-    const requestBody = {
-      'event-start-date': emptyDate,
-      'event-end-date': emptyDate,
-      'event-team-hosting': req.session.user.dit_team.id,
-    }
-    const eventFormWithState = buildFormWithState(eventForm, requestBody)
+    req.body['event-start-date'] = emptyDate
```
feedmypixel: snakeCase is used for name values, you also don't need the brackets in that case `req.body.event_start_date`

src/apps/events/controllers/edit.js
```diff
@@ -2,19 +2,19 @@ const { eventFormConfig } = require('../macros')
...
+    req.body['event-end-date'] = emptyDate
+    req.body['event-team-hosting'] = req.session.user.dit_team ? req.session.user.dit_team.id : null
```
feedmypixel: feels like you could use lodash `get` here `get(req, 'session.user.dit_team.id', null)`

src/templates/_macros/form.njk
```diff
@@ -352,6 +352,33 @@
 {% endmacro %}
 
 {##
+ # Render a collection for multiple fields for add another functionality
+ # @param {object} props - An object containing field properties
+ # @param {array} [props.value] - Input values
+ #}
+{% macro MultipleChoiceFieldAddAnother(props) %}
```
ztolley: My gut says if the multiple choice part of it is identical to the existing one, use the existing one and add the markup into the view for a wrapper around it and add the add button. The add another code allows you to specify the add button

src/apps/events/controllers/edit.js
```diff
@@ -2,19 +2,19 @@ const { eventFormConfig } = require('../macros')
...
+
+    const eventFormWithState = buildFormWithState(eventForm, req.body)
```
feedmypixel: the mutating of `req.body` feels a bit odd here. Can I suggest you `Object.assign` it

```
const body = Object.assign(req.body, {
  event_team_hosting: get(req, 'session.user.dit_team.id', null),
  event_start_date: emptyDate,
   event_end_date: emptyDate
})
```

`buildFormWithState(eventForm, body)`

src/apps/events/controllers/edit.js
feedmypixel: Where are they needed?

src/apps/events/controllers/edit.js
```diff
@@ -27,6 +27,21 @@ async function renderEventPage (req, res, next) {
   }
 }
 
+function postHandler (req, res, next) {
+  const setAddAnotherField = (value) => compact(castArray(value))
+  req.body['event-shared-teams'] = setAddAnotherField(req.body['event-shared-teams'])
+  req.body['event-programmes'] = setAddAnotherField(req.body['event-programmes'])
+
+  if (req.body.addEventSharedTeam) {
+    return renderEventPage(req, res, next)
```
feedmypixel: you could just drop through with `next()` and then add `renderEventPage` on the end of the array in the middleware

src/templates/_macros/form.njk
feedmypixel: I don't feel we need to lock this down to `MultipleChoice` as mentioned can we not just have a wrapper for `addAnother` and then send in what ever we like via `caller`?

src/apps/events/controllers/edit.js
yeahfro: On saving so I guess this is the right place

assets/javascripts/modules/conditional-subfields.js
yeahfro: This check is required because in this situation subField is the div and children contains the button which needs to retain its value.

src/apps/events/router.js
```diff
@@ -1,7 +1,7 @@
 const router = require('express').Router()
 
-const { renderEventPage } = require('./controllers/edit')
+const { renderEventPage, postHandler } = require('./controllers/edit')
 
-router.get('/create', renderEventPage)
+router.route('/create').get(renderEventPage).post(postHandler, renderEventPage)
```
feedmypixel: the `.get` and `.post` can go on newlines for better readability

src/apps/components/views/form.njk
```diff
@@ -196,4 +196,33 @@
...
+      {{ AddAnother({
+
+        buttonName: 'add_another',
+        name: 'add_another',
+        label: 'Another',
+        children: [
+          {
+            macroName: 'MultipleChoiceField',
+            name: 'add_another',
+            label: 'Another',
```
feedmypixel: Probs stick a better example label in. Wee bit generic

test/unit/macros/form.test.js
```diff
@@ -407,4 +407,49 @@ describe('Nunjucks form macros', () => {
...
+              { value: 3, label: 'another 3' },
+            ],
+          },
+        ],
+        value: [ 1, 2 ],
+      })
+    })
+
+    it('should render a component with 2 preselected drop downs', () => {
+      const actual = this.component.getElementsByTagName('select')
```
feedmypixel: could do with a better name - what does `actual` mean here? Maybe `selectElement`

test/unit/macros/form.test.js
```diff
@@ -407,4 +407,49 @@ describe('Nunjucks form macros', () => {
...
+
+      expect(actual[0].value).to.equal('1')
+      expect(actual[1].value).to.equal('2')
+    })
+
+    it('should render a component with 1 unselected drop down', () => {
+      const actual = this.component.getElementsByTagName('select')
```
feedmypixel: could do with a better name - what does actual mean here? Maybe selectElement

test/unit/macros/form.test.js
```diff
@@ -407,4 +407,49 @@ describe('Nunjucks form macros', () => {
...
+
+      expect(actual[2].value).to.equal('')
+    })
+
+    it('should render a component with an "Add another" button', () => {
+      const actual = this.component.querySelectorAll('input[type=submit]')
```
feedmypixel: could do with a better name - what does actual mean here? Maybe selectElement

PR #557, opened 2017-09-08
Feature/dh 547 list interactions

ztolley:
Provides a top level link to list interactions, and a list of interactions using collection pattern.

No filters as back end doesn't support this yet.

screenshot from 2017-09-11 12-14-47

### DIFF
--- a/src/apps/interactions/router.js
+++ b/src/apps/interactions/router.js
@@ -1,9 +1,16 @@
 const router = require('express').Router()
 
-const detailsController = require('./controllers/details')
-const editController = require('./controllers/edit')
+const { setDefaultQuery } = require('../middleware')
+const { getInteractionsCollection, getRequestBody } = require('./middleware')
+const { detailsController, listController, editController } = require('./controllers')
+
+const DEFAULT_COLLECTION_QUERY = {
+  sortby: 'date:desc',
+}
 
 router.param('interactionId', detailsController.getCommon)
+
+router.get('/', setDefaultQuery(DEFAULT_COLLECTION_QUERY), getRequestBody, getInteractionsCollection, listController.renderInteractionList)
 
 router.get('/create', (req, res) => { res.redirect('create/1') })
 router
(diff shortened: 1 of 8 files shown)


test/unit/apps/interactions/middleware.test.js
```diff
@@ -0,0 +1,69 @@
...
+      })
+    })
+    it('should call the search response transformer with the interaction entity transformer', async () => {
+      await this.middleware.getInteractionsCollection(this.req, this.res, this.next)
+
+      expect(this.transformApiResponseToSearchCollectionStub).to.be.calledWith(
+        { query: this.req.query },
+        transformInteractionToListItem
+      )
+    })
```
yeahfro: I see what is happening here. My feelings around this style of tests is that it is testing implementation details which can be costly if the code ever changes. My preferred approach would be to do similar to what you have done with setting up the mocks and then just asserting what comes back out.. I cannot see any assertions on `res.locals` and that is what we actually care about on this controller function.

ztolley: What I was worried about was duplication of testing and would end up re-testing the item transformer, but I'll just mock that out and test the result

src/apps/interactions/router.js
```diff
@@ -1,10 +1,19 @@
 const router = require('express').Router()
 
+const { setDefaultQuery } = require('../middleware')
+const { getInteractionsCollection, getRequestBody } = require('./middleware')
+const { renderInteractionList } = require('./controllers/list')
```
feedmypixel: can these controllers can be merged?

ztolley: ??

feedmypixel: you can use `const { detailsController, listController, editController } = require('./controllers')`

ztolley: One thing that bugs me is some places we use module.method and other places we just use method

PR #561, opened 2017-09-08
Add a medium modifier for form groups

teneightfive:
Adds another modifier in same style as currently exists to support medium width fields.

### DIFF
--- a/assets/stylesheets/components/form/_form-control.scss
+++ b/assets/stylesheets/components/form/_form-control.scss
@@ -59,6 +59,10 @@ select.c-form-control {
   border-color: $error-colour;
 }
 
+.c-form-control--medium {
+  max-width: 15em;
+}
+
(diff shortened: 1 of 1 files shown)


assets/stylesheets/components/form/_form-control.scss
```diff
@@ -59,6 +59,10 @@ select.c-form-control {
   border-color: $error-colour;
 }
 
+.c-form-control--medium {
```
ztolley: whats driving the need for this? Why not size the container of the form?

teneightfive: Because it's not a form wide modifier. This gives the ability to reduce the size of particular inputs rather than all inputs.

In some instances, like a PO number, I don't want a full width field like for something like Name.

PR #564, opened 2017-09-09
Removed use of Path library when generating URL's

ztolley:
The path library is designed to generate file system paths,
not HTTP URL's. When the service is run on windows the urls
generated come out as 'c:\'.

### DIFF
--- a/src/middleware/locals.js
+++ b/src/middleware/locals.js
@@ -1,5 +1,3 @@
-const path = require('path')
-
 const logger = require('../../config/logger')
 const config = require('../../config')
 
@@ -30,10 +28,10 @@ module.exports = function locals (req, res, next) {
     BREADCRUMBS: breadcrumbItems,
     IS_XHR: req.xhr,
     QUERY: req.query,
-    GLOBAL_NAV_ITEMS: globalNavItems.map(item => {
-      const url = path.resolve(req.baseUrl, item.path)
+    GLOBAL_NAV_ITEMS: globalNavItems.map(globalNavItem => {
+      const url = globalNavItem.path
 
-      return Object.assign(item, {
+      return Object.assign(globalNavItem, {
         url,
         isActive: req.path.startsWith(url),
       })
--- a/src/apps/middleware.js
+++ b/src/apps/middleware.js
@@ -1,4 +1,3 @@
-const path = require('path')
 const { isEmpty } = require('lodash')
 const queryString = require('query-string')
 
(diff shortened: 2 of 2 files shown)


src/middleware/locals.js
```diff
@@ -30,11 +28,9 @@ module.exports = function locals (req, res, next) {
     IS_XHR: req.xhr,
     QUERY: req.query,
     GLOBAL_NAV_ITEMS: globalNavItems.map(item => {
-      const url = path.resolve(req.baseUrl, item.path)
```
teneightfive: We still need to include the `req.baseUrl` property as is used in sub-apps.

Maybe still cache the url as its also used in multiple places.

ztolley: Can you give me an example screen to goto so I can test it and confirm the changes I make work?

ztolley: done

src/middleware/locals.js
```diff
@@ -30,11 +28,9 @@ module.exports = function locals (req, res, next) {
     IS_XHR: req.xhr,
     QUERY: req.query,
     GLOBAL_NAV_ITEMS: globalNavItems.map(item => {
```
feedmypixel: can we name `item` -> `globalNavItem`? feels like we are loosing something calling it `item`

src/middleware/locals.js
teneightfive: >Maybe still cache the url as its also used in multiple places.

teneightfive:
```js
const url = `${req.baseUrl}/${item.path}`
```

ztolley: ??

teneightfive: Replace 

```js
path.resolve(req.baseUrl, item.path)
```

with

```js
const url = `${req.baseUrl}/${item.path}
```

rather than removing this line and using the same value twice.

src/middleware/locals.js
```diff
@@ -30,12 +28,10 @@ module.exports = function locals (req, res, next) {
...
     QUERY: req.query,
-    GLOBAL_NAV_ITEMS: globalNavItems.map(item => {
-      const url = path.resolve(req.baseUrl, item.path)
-
-      return Object.assign(item, {
-        url,
-        isActive: req.path.startsWith(url),
+    GLOBAL_NAV_ITEMS: globalNavItems.map(globalNavItem => {
+      return Object.assign(globalNavItem, {
+        url: baseUrl + globalNavItem.path,
```
teneightfive: This is using the wrong baseUrl. It was originally using `req.baseUrl`.

But having looked at it, seeing as this is being used at the root of the app we can probably drop it.

So this method can become:

```js
GLOBAL_NAV_ITEMS: globalNavItems.map(globalNavItem => {
  const url = globalNavItem.path

  return Object.assign(globalNavItem, {
    url,
    isActive: req.path.startsWith(url),
  })
}
```

Which would reduce the diff to the only bit you're concerned with changing, the `url` value.

ztolley: ah. was confused because I was sure the startsWith would fail, but what do you know it works.

PR #565, opened 2017-09-09
Feature/DH-599 update company interactions

ztolley:
screenshot from 2017-09-11 18-28-45

* Updates company to use collections to list related interactions.
* Switches interaction data to be collected from new V3 endpoint
* Introduce the start of moving company to middleware pattern by using middleware to get company details, labels and interaction data before render.
* Remove code no longer needed in getExpandedDitCompany


teneightfive:
Once thing I just noticed on the screenshot - should we move the _Add interaction_ button into the same location it is in for investment projects:

image

src/apps/companies/middleware/details.js
```diff
@@ -0,0 +1,45 @@
...
+
+  if (company.companies_house_data && company.companies_house_data !== null) {
+    return getFormattedAddress(company.companies_house_data, 'registered')
+  }
+
+  return getFormattedAddress(company, 'registered')
+}
+
+function getHeadingName (company) {
+  if (company.trading_name && company.trading_name.length > 0) {
```
teneightfive: This feels like something that could just be handled in the view as the only place it is used is when being passed to `res.locals`.

src/apps/companies/middleware/details.js
```diff
@@ -0,0 +1,45 @@
...
+      next(error)
+    }
+  }
+
+  next()
+}
+
+function getHeadingAddress (company) {
+  // If this is a CDMS company
+  const cdmsTradingAddress = getFormattedAddress(company, 'trading')
```
teneightfive: I'd be tempted to also consider handling this in the view as its only for being used for rendering.

It then also allows us to use the view filters rather than having separate library methods for it.

src/apps/companies/views/interactions.njk
```diff
@@ -1,84 +1,25 @@
...
-  {% elseif (not company.contacts or company.contacts.length == 0) %}
+  <div class="split-header">
     <h2 class="heading-medium">Company interactions</h2>
-    <p>
-      An interaction could be a meeting, call, email or another activity where you have been in touch with a company.
-    </p>
-    <p id="no-contact-warning" class="infostrip">
-      You currently have no contacts for this company. To add an interaction you must first <a href="{{addContact}}">add a contact</a>.
+    {% if (company.contacts and company.contacts.length > 0) %}
+      <a id="add-interaction-link" class="button" href="{{addInteractionUrl}}">Add interaction</a>
```
teneightfive: Maybe move this to the same location as on investment projects.

src/apps/interactions/transformers.js
```diff
@@ -40,6 +40,42 @@ function transformInteractionToListItem ({
   }
 }
 
+function transformInteractionToCompanyListItem ({
```
teneightfive: This looks like quite a repetition of the other transformer in this file. Is the only different to remove company from the meta?

Maybe consider chaining another transformer to remove that meta item rather than duplicating it.

If the transformer in search/collection changed you'd have to make the change twice or risk it being missed in this listing.

src/apps/companies/middleware/details.js
ztolley: Good point though not trivial

src/apps/companies/middleware/details.js
ztolley: Easier to test in the middleware

src/apps/interactions/transformers.js
ztolley: Yeah I kept thinking that, bugged me. will do

src/apps/companies/middleware/details.js
ztolley: Can we agree that I'll create a filter for it later and refactor this to use it?

src/apps/companies/middleware/details.js
teneightfive: I don't think we need to unit test views.

What is rendered is something that can be picked up by end to end tests.

This feels like the wrong place to have this logic though.

ztolley: Well we could argue that the view should not contain logic, but simply show values and loop.

I was simply trying to start to introduce middleware into company as part of chipping away at the current way it's structured. If I move this into the view layer then it's more risky and I then have to go back and test it with companies house vs cdms records

Can we discuss this with the other Fed's and if we want to do this, and the address suggestion, we can do it as part of a separate PR that would focus on that.

src/apps/companies/middleware/details.js
teneightfive: Yes, want to create an issue so we don't forget?

src/apps/interactions/transformers.js
```diff
@@ -40,6 +40,24 @@ function transformInteractionToListItem ({
...
+ *
+ * @param {object} iteraction action (id, type, name, meta)
+ *
+ * Accepts an iteraction collection obect and reduces it to exclude company data
+ */
+function transformInteractionToCompanyListItem (interaction) {
+  const { id, type, name, meta } = transformInteractionToListItem(interaction)
+
+  return Object.assign({}, {
+    id,
```
teneightfive: I'm wondering if this is the best implementation as it feels like it's still repeating some things.

What about chaining the `.then()` in the company controller to strip out the unwanted value.

Whilst writing this and looking at investment project listing...why don't we just leave this in? We have done for investment projects. That way the other meta items will also still appear in the same location therefore making them easier to scan.

test/unit/apps/companies/controllers/interactions.test.js
```diff
@@ -1,232 +1,65 @@
-const { render } = require('~/test/unit/nunjucks')
-const next = function (error) {
-  throw Error(error)
-}
-
 describe('Company interactions controller', function () {
-  let company
-  let companyinteractioncontroller
+  beforeEach(() => {
+    this.sandbox = sinon.sandbox.create()
```
teneightfive: Have been wondering whether to move this out to a global level so that the sandbox is available in all tests. What are the thoughts on that?

Definitely not part of this change but seeing the diff reminded me. @tyom and I were discussing that we should define the pattern for our tests too as there's a lot of different methods going on at the moment.

src/apps/interactions/transformers.js
ztolley: The stripping should happen in the transformer.

The reason for having it, and in the PR for contacts, is that they would all just say this company. 

I tried chaining them but that didn't work plus the controller or middleware needs to know to do that, so poor abstraction.

src/apps/companies/views/interactions.njk
```diff
@@ -1,84 +1,25 @@
 {% extends "./_layout-view.njk" %}
 
+{% set addButton %}
+  <a id="add-interaction-link" class="button" href="{{addInteractionUrl}}">Add interaction</a>
```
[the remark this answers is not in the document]
ztolley: Also updated investment interactions to be consistent.

Also no longer generate the url in the controller, just create it in the view.

https://github.com/uktrade/data-hub-frontend/pull/571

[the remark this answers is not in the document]
ztolley: The id is used for tests

test/unit/apps/companies/controllers/interactions.test.js
ztolley: yes and yes

Tests need patterns too (some use nock, some stub repos)

src/apps/interactions/transformers.js
ztolley: That all said, ask the others, put it to the vote. it's a quick change

ztolley:
Made a few changes tonight to make names more obvious and improve consistency between this, contact interactions and investment interactions.

src/apps/interactions/transformers.js
teneightfive: >The reason for having it, and in the PR for contacts, is that they would all just say this company.

Yeah it's not ideal but I noticed whilst writing that comment that investment projects does the same, it still includes the company. I'd also be tempted to say that for now it would also keep the other meta items in the same location as when they appear on search and collections therefore benefitting scanability. 

Was more just a thought to do less work for now and save us some time. Maybe put an issue in GitHub to see if we want to improve it later on, but if we do, we should do the same in investment and probably.

Although looking at contacts when in a company it does also filter out the company name but uses a [different method](https://github.com/uktrade/data-hub-frontend/blob/5b85355be71a47d7f7ab7df0f80974f599a9c391/src/apps/companies/controllers/contacts.js#L18-L23) to reject that meta item. I would say go with this method now for consistency.

src/apps/companies/views/interactions.njk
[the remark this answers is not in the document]
teneightfive: >Also updated investment interactions to be consistent.

I think this is wrong way to update. We've currently used secondary buttons for adding a contact in all of the collections so I would update this PR for consistency rather than investment projects.

## Companies collection

image

## Company search results

image

## Company contacts listing

image

## Company investment project listing

image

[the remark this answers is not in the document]
teneightfive: >The id is used for tests

@feedmypixel and I were discussing the use of IDs purely for tests yesterday and I feel it's a bad habit to get into. If we need specific hooks we should use prefixed classes (`qa-`) rather than IDs.

IDs should be kept available in case we wish to use them for anchors or for interacting with the ARIA API. Using prefixed classes also better communicates what is going on in the markup and if we wanted we could optionally only show them in certain environments that we run tests on so that we don't pollute the final markup more than we need to.

src/apps/interactions/transformers.js
ztolley: Thats the old way of doing collections. We should migrate the company contacts to this method instead, not the other way around

ztolley: Changing to use the common transformer just to make it simpler and get this done but contact collections in company needs to be re-done

src/apps/companies/views/interactions.njk
```diff
@@ -1,84 +1,25 @@
...
-    </ul>
-  {% elseif (not company.contacts or company.contacts.length == 0) %}
-    <h2 class="heading-medium">Company interactions</h2>
-    <p>
-      An interaction could be a meeting, call, email or another activity where you have been in touch with a company.
-    </p>
-    <p id="no-contact-warning" class="infostrip">
-      You currently have no contacts for this company. To add an interaction you must first <a href="{{addContact}}">add a contact</a>.
-    </p>
+  {% if (company.contacts and company.contacts.length > 0) %}
```
teneightfive: Checking for just `company.contacts.length` should be enough:

```nunjucks
{% if company.contacts.length %}
  ...
{% endif %}
```

ztolley: it was being defensive in case there is no property there, and if there isn't it breaks things. tried it before and it did. That said we should be able to trust the api to return contacts no matter what.

teneightfive: Don't think you need to be in nunjucks. It won't error even if the properties don't exist. It will just fail the condition.

Just tried in a view with a variable and property that didn't exist and it was fine.

ztolley: done

src/apps/companies/middleware/interactions.js
```diff
@@ -0,0 +1,25 @@
...
+    const page = req.query.page || '1'
+    const companyId = req.params.companyId
+
+    res.locals.interactions = await getInteractionsForCompany(token, companyId, page)
+      .then(transformApiResponseToCollection(
+        { entityType: 'interaction' },
+        transformInteractionToListItem
+      ))
+  } catch (error) {
+    next(error)
```
teneightfive: Does this need to be returned so that it doesn't call the `next()` without an error on line 20?

Or it might be better to move the `next()` from line 20 up to line 16 within the `try` block.

src/apps/interactions/repos.js
```diff
@@ -49,19 +49,13 @@ function getInteractionsForContact (token, contactId) {
...
-    qs: {
-      company_id: companyId,
-      limit,
-    },
-  })
-  return response.results
+function getInteractionsForCompany (token, companyId, page = 1) {
+  const limit = 10
+  const offset = limit * (page - 1)
+  return authorisedRequest(token, `${config.apiRoot}/v3/interaction?company_id=${companyId}&limit=${limit}&offset=${offset}`)
```
teneightfive: Just looking at the diff here, why did you change it from passing query strings as an object using a template literal?

Looks easier to read as a querystring object.

ztolley: All the other methods in the module use this format so I was being consistent

src/apps/companies/middleware/interactions.js
ztolley: moving it up would mean that a failure further down the chain would come back to this catch, so I'll return the error

ztolley: btw we have this in other places.

PR #570, opened 2017-09-11
Feature/DH-600 Contact interactions

ztolley:
screenshot from 2017-09-11 21-11-40

* Use new v3 API endpoint to get interaction data.
* Add transformer to trim down interaction data for display under company contact
* splits interaction list into middleware to get data and controller to render


src/apps/contacts/controllers/interactions.js
```diff
@@ -1,14 +1,11 @@
...
+ */
+function renderInteractions (req, res, next) {
   try {
-    res.locals.tab = 'interactions'
-    const interactions = await contactDataService.getContactInteractionsAndServiceDeliveries(req.session.token, req.params.contactId)
-    res.locals.interactions = interactions.map(interaction => interactionFormattingService.getDisplayContactInteraction(interaction))
-    res.locals.addInteractionUrl = `/interactions/create/1?contact=${res.locals.contact.id}`
-
     res
+      .breadcrumb(res.locals.contact.name, `/contact/${res.locals.contact.id}`)
```
teneightfive: What about:

```js
const { name, id } = res.locals.contact

res
  .breadcrumb(name, `/contact/${id}`)
```

src/apps/contacts/views/interactions.njk
```diff
@@ -1,60 +1,18 @@
 {% extends "./_layout.njk" %}
 
+{% set addButton %}
+  <a id="add-interaction-link" class="button" href="/interactions/create/1?contact=${contact.id}">Add interaction</a>
```
teneightfive: As [mentioned](https://github.com/uktrade/data-hub-frontend/pull/565#discussion_r138269430) in #565 we should use secondary button style for now for consistency.

src/apps/interactions/transformers.js
```diff
@@ -40,6 +40,24 @@ function transformInteractionToListItem ({
   }
 }
 
+/**
+ *
+ * @param {object} iteraction action (id, type, name, meta)
+ *
+ * Accepts an iteraction collection obect and reduces it to exclude contact data
+ */
+function transformInteractionToContactListItem (item) {
```
teneightfive: Have a look at [the comment](https://github.com/uktrade/data-hub-frontend/pull/565#discussion_r138268583) on #565 around how to filter out a specific meta item.

Doesn't feel like this method is behaving in the same way as our existing transformers. The name is slightly misleading as we don't provide arguments in the same way as the other method. This is doing something more specific around removing a particular item.

PR #571, opened 2017-09-11
DH-605 Investment interactions

ztolley:
screenshot from 2017-09-11 21-16-13

Updated the investment interactions screen to use the v3 interaction API and collection pattern for layout.

### DIFF
--- a/src/apps/investment-projects/views/interactions/index.njk
+++ b/src/apps/investment-projects/views/interactions/index.njk
@@ -1,63 +1,19 @@
 {% extends "../_layout.njk" %}
 
+{% set addButton %}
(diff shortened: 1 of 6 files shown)


src/apps/investment-projects/views/interactions/index.njk
```diff
@@ -1,63 +1,19 @@
 {% extends "../_layout.njk" %}
 
+{% set addButton %}
+  <a href="interactions/create" class="button">Add interaction</a>
```
teneightfive: As [mentioned](https://github.com/uktrade/data-hub-frontend/pull/565#discussion_r138269430) in #565 I feel we should use secondary button style here to be consistent with the rest of the app for now.

PR #573, opened 2017-09-12
Editing an Event failing e2e tests

venureddy95:
Includes feature file and step definitions. Page objects will follow once dev task is completed.


test/acceptance/features/event-edit.feature
```diff
@@ -0,0 +1,127 @@
+@edit-event @ignore
```
feedmypixel: I wonder if `@todo` might be better? then I can skip then via `--tags ~@todo`?

PR #574, opened 2017-09-12
Acceptance tests improvements to create company feature

feedmypixel:
- fix bug with create UK ltd company where `regionOptions` was not in the form
- improve step definition reuse
- improve variable names
- remove duplicate work
- Start to think about `global/shared` scenarios (W.I.P)

screen shot 2017-09-12 at 11 50 41


test/acceptance/features/create-company.feature
```diff
@@ -0,0 +1,28 @@
...
+    Given I am an authenticated user on the Datahub website
+
+  @create-uk-private-or-public-ltd-company
+  Scenario: Create a UK private or public limited company
+
+    When I create a "UK private or public limited company"
+    Then I see the "Company record updated" success message
+    And The company is present in the search results
+
+  @create-uk-non-private-or-non-public-ltd-company
```
venureddy95: Tag names should be small, so that we can, at time, run each scenario using these tags. Lengthy tag name leads to misspelling. The Scenario name will describe the tag in detailed. @tyom was the one who was against the Lengthy tag names.

PR #586, opened 2017-09-13
OMIS content updates

teneightfive:
This change is centered around the content for the OMIS create and edit journeys.

It uses consistent headings and titles for edit and create steps and adds some better supporting texts to explain what pieces of information are required.

### DIFF
--- a/src/apps/omis/apps/edit/fields.js
+++ b/src/apps/omis/apps/edit/fields.js
@@ -1,4 +1,4 @@
-const { isArray } = require('lodash')
+const { isArray, merge } = require('lodash')
 
 const globalFields = require('../../fields')
 
@@ -23,7 +23,11 @@ function isDuration (value) {
   return valid
 }
 
-const editFields = Object.assign({}, globalFields, {
+const editFields = merge({}, globalFields, {
+  subscribers: {
+    hint: 'fields.subscribers.hint.edit',
+    optional: false,
+  },
   service_types: {
     fieldType: 'MultipleChoiceField',
     type: 'checkbox',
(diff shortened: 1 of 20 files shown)


src/apps/omis/apps/edit/fields.js
```diff
@@ -23,7 +23,11 @@ function isDuration (value) {
   return valid
 }
 
-const editFields = Object.assign({}, globalFields, {
+const editFields = merge({}, globalFields, {
```
yeahfro: I like the use of `merge` here. Should we just switch to using merge everywhere? (over `Object.assign`)

yeahfro: In addition, the first empty object parameter is not required.

yeahfro: Actually I am wrong

teneightfive: >In addition, the first empty object parameter is not required.

If you don't include the first empty object it will mutate the original object I believe rather than create a copy.

>I like the use of `merge` here. Should we just switch to using merge everywhere? (over `Object.assign`)

I think it depends on the scenario. I feel we should stick to using the native `Object.assign()` when we want to override full properties but we need to make sure we understand the difference. `merge` is maybe a little harder to understand so should only use it in the cases where we don't want to override all deep properties.

yeahfro: Yep - after playing around with it during the day and rereading my comments I realise my naivety!

tyom: `merge` is like `assign` except it recursively merges own and inherited enumerable string keyed properties.

`assign` only works with own enumerable properties and is flat (doesn't recurse in properties that are objects).

PR #587, opened 2017-09-13
Update form actions to give more breathing space to secondary action

teneightfive:
This uses a sass component to provide some better spacing to the action
within a form.

## Before

image

### mobile

image

### > tablet

image

### DIFF
--- a/assets/stylesheets/components/form/_form-actions.scss
+++ b/assets/stylesheets/components/form/_form-actions.scss
@@ -1,0 +1,23 @@
+@import "../../settings";
+
+.c-form-actions {
+  * {
+    display: block;
+  }
+
+  * + * {
+    margin-top: $default-spacing-unit;
+  }
+
+  @include media(tablet) {
+    * {
+      display: inline-block;
+      vertical-align: baseline;
+    }
+
+    * + * {
+      margin-top: 0;
+      margin-left: $default-spacing-unit;
+    }
+  }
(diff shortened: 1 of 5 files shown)


assets/stylesheets/components/form/_form-actions.scss
```diff
@@ -0,0 +1,20 @@
+@import "../../settings";
+
+.c-form-actions {
+  * + * {
```
ztolley: If this is for mobile, why aren't just making all buttons in form actions a block? and just have it so they all have a top margin except the first?

teneightfive: Actually it should be more like:

```scss
.c-form-actions {
  * {
    display: block;
  }

  * + * {
  }
}
```

>and just have it so they all have a top margin except the first?

The [lobotomized owl](https://alistapart.com/article/axiomatic-css-and-lobotomized-owls) selector (`* + *`) is doing exactly this. It only sets a top margin when there is a preceeding sibling. It's a nicer than using something like this:

```scss
.selector {
  margin-top: 20px;
}

.selector:first-child {
  margin-top: 0;
}
```

If you haven't read that article before I highly recommend it. It shows some interesting tactics for making spacing easier to handle and more consistent. I'd like to use that approach for our elements when we have some more time. It avoids you having to declare margins and the majority of components.

PR #588, opened 2017-09-13
Add link to edit advisers from OMIS time edit

teneightfive:
This allows a direct link back to changing the advisers when a
user is editing the estimated time for advisers.

## Before

localhost_3001_omis_698f449e-9c86-428f-b674-f86174d67897_edit_assignee-time 1

## After

localhost_3001_omis_698f449e-9c86-428f-b674-f86174d67897_edit_assignee-time

### DIFF
--- a/assets/stylesheets/components/_answers-summary.scss
+++ b/assets/stylesheets/components/_answers-summary.scss
@@ -116,3 +116,7 @@
     text-align: right;
   }
 }
+
+.c-answers-summary__title--unstyled {
+  border: 0;
+}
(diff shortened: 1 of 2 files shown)


teneightfive:
>I wonder if we want to split to 2 fileds, hh and mm, like dates.

I think splitting fields in a system where the user has to enter a lot of data often is actually not very productive. The 3 date fields works for one of transactions like entering your date of birth when registering to vote but for a system people use day-to-day we should adapt these patterns for speed.

This is something we will pay attention to in testing though but it is out of scope for this change. This change involves add a deep link in to editing advisers if the adviser somebody is looking for is missing.

PR #593, opened 2017-09-14
Fix an OMIS edit controller so that it doesn't treat true as a date

teneightfive:
If true was passed to the datefns parse function it would treat it
as a date and therefore parse the value as a date.

### DIFF
--- a/src/apps/omis/controllers/edit.js
+++ b/src/apps/omis/controllers/edit.js
@@ -2,7 +2,9 @@ const {
   find,
   flatten,
   get,
+  isNull,
   isPlainObject,
+  isUndefined,
   map,
   mapValues,
   pick,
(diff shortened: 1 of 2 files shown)


src/apps/omis/controllers/edit.js
```diff
@@ -65,10 +65,13 @@ class EditController extends FormController {
...
     let combinedValues = Object.assign({}, orderValues, sessionValues, errorValues)
     // convert dates to default format
-    combinedValues = mapValues(combinedValues, (value) => {
-      const parsedDate = dateFns.parse(value)
+    combinedValues = mapValues(combinedValues, (value, key) => {
+      if (!value) {
+        return value
+      }
 
-      if (value && dateFns.isValid(parsedDate)) {
```
feedmypixel: could you just do a string check?
```
if (typeof value === 'string') {
 const parsedDate = dateFns.parse(value)
  if (dateFns.isValid(parsedDate)) {
    return dateFns.format(parsedDate, longDateFormat)
  }
}
```

src/apps/omis/controllers/edit.js
```diff
@@ -65,10 +65,13 @@ class EditController extends FormController {
     // combine order values and error values
     let combinedValues = Object.assign({}, orderValues, sessionValues, errorValues)
     // convert dates to default format
-    combinedValues = mapValues(combinedValues, (value) => {
-      const parsedDate = dateFns.parse(value)
+    combinedValues = mapValues(combinedValues, (value, key) => {
+      if (!value) {
```
feedmypixel: > If true was passed to the datefns parse function it would treat it as a date and therefore parse the value as a date.

this doesn't cover `true` though right?

PR #595, opened 2017-09-14
Fix the checking of the right value in multiple choice field

teneightfive:
There was an error where for radio type it would check for current
value in the sent value rather than matching exactly.

This also now supports checking the option correctly if a value is `false`.

### DIFF
--- a/src/templates/_macros/form.njk
+++ b/src/templates/_macros/form.njk
@@ -512,8 +512,13 @@
     {% endif %}
 
     {% for option in options %}
-      {% set selected = props.value|default([]) %}
       {% set children = option.children | map(callAsMacro) if options.children else renderAsMacro(option.children) %}
+      {% if props.type == 'checkbox' %}
+        {% set isSelected = (option.value in props.value | default([])) %}
+      {% else %}
+        {% set isSelected = (option.value | string == props.value | default(false) | string) %}
+      {% endif %}
+
       <div class="c-multiple-choice {{ modifier }}">
         <input
           class="c-multiple-choice__input"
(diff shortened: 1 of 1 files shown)


src/templates/_macros/form.njk
```diff
@@ -512,16 +512,21 @@
     {% endif %}
 
     {% for option in options %}
-      {% set selected = props.value|default([]) %}
       {% set children = option.children | map(callAsMacro) if options.children else renderAsMacro(option.children) %}
+      {% if props.type == 'checkbox' %}
+        {% set isSelected = (option.value in props.value | default([])) %}
```
feedmypixel: this is  `boolean` or an `array`?

src/templates/_macros/form.njk
```diff
@@ -512,16 +512,21 @@
...
+      {% else %}
+        {% set isSelected = (option.value | string == props.value | default(false) | string) %}
```
feedmypixel: same here this is a `boolean` or a `string`? feels like theres possible some funny types somewhere? Is this work a symptom of something else?

src/templates/_macros/form.njk
teneightfive: When the type is `checkbox` we should be checking for the current value to be in an array. Wonder if we need to cast to an array in case there is only 1 value and it comes back as a string...

src/templates/_macros/form.njk
teneightfive: The else catches a `string`. When the type is `radio` is should just be expecting a string for `props.value`. If it's a `checkbox` type then it should expect an array.

The string cast is being used for boolean values as the value coming in for option.value will always be a string.

PR #596, opened 2017-09-15
DH-524 Set teams and related programmes before POST to API

yeahfro:
This changes moves the setting of the `teams` and `related_programmes` to before POST to create event. If the user is attempting to add another then the POST will not happen.

In future PRs:
- Tidy up edit controller to be consistent with new patterns
- Check presentation of errors (fingers crossed this should just work)

### DIFF
--- a/src/apps/events/middleware/details.js
+++ b/src/apps/events/middleware/details.js
@@ -1,8 +1,21 @@
 const { transformToApi } = require('../services/formatting')
 const { createEvent } = require('../repos')
+const { includes, assign, castArray, compact } = require('lodash')
 
 async function handleFormPost (req, res, next) {
-  const formattedBody = transformToApi(Object.assign({}, req.body))
+  const castToArrayAndRemoveEmpty = (value) => compact(castArray(value))
+  req.body.teams = castToArrayAndRemoveEmpty(req.body.teams)
+  req.body.related_programmes = castToArrayAndRemoveEmpty(req.body.related_programmes)
+
+  if (!includes(req.body.teams, req.body.lead_team)) {
+    req.body.teams.push(req.body.lead_team)
+  }
+
+  if (req.body.add_team || req.body.add_related_programme) {
+    return next()
+  }
+
+  const formattedBody = transformToApi(assign({}, req.body))
 
   try {
     const result = await createEvent(req.session.token, formattedBody)
@@ -12,7 +25,7 @@ async function handleFormPost (req, res, next) {
     next()
   } catch (err) {
     if (err.statusCode === 400) {
-      res.locals.form = Object.assign({}, res.locals.form, {
+      res.locals.form = assign({}, res.locals.form, {
         state: req.body,
         errors: {
           messages: err.error,
--- a/src/apps/events/services/formatting.js
+++ b/src/apps/events/services/formatting.js
@@ -1,4 +1,4 @@
-const { mapValues, isPlainObject } = require('lodash')
+const { mapValues, isPlainObject, some, assign } = require('lodash')
 
 function transformToApi (body) {
   if (!isPlainObject(body)) { return }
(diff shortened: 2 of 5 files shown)


src/apps/events/middleware/details.js
```diff
@@ -1,8 +1,21 @@
 const { transformToApi } = require('../services/formatting')
 const { createEvent } = require('../repos')
+const { includes, assign, castArray, compact } = require('lodash')
 
 async function handleFormPost (req, res, next) {
-  const formattedBody = transformToApi(Object.assign({}, req.body))
+  const setAddAnotherField = (value) => compact(castArray(value))
+  req.body.teams = setAddAnotherField(req.body.teams)
```
feedmypixel: - feels like you maybe loosing a little bit of clarity here with the `setAddAnotherField` abstraction. 
- does clearing out false values here mean that there is an ordering issue when you have errors?
-feels like maybe `req.body.team` should always be delivered as an array?

src/apps/events/services/formatting.js
```diff
@@ -33,17 +33,21 @@ function transformToApi (body) {
...
+    const fields = [
       body[`${key}_year`],
       body[`${key}_month`],
       body[`${key}_day`],
-    ].join('-')
+    ]
+
+    if (sumBy(fields, isEmpty) === 0) {
+      formatted[key] = [ ...fields ].join('-')
+    }
```
feedmypixel: Am I reading this right?
if all the `fields[]` have a value then `.join('-')` them and assign to `formatted[key]`

Why not just always provide all the fields and do a `.join('-')` on the array?

src/apps/events/services/formatting.js
```diff
@@ -33,17 +33,21 @@ function transformToApi (body) {
...
   }
 
   setDate('start_date')
   setDate('end_date')
 
-  return Object.assign({}, formatted)
+  return assign({}, formatted)
```
feedmypixel: what was the consensus are we slowly changing these as we go?

src/apps/events/middleware/details.js
yeahfro: On reflection the naming of this is not great so I have changed this

src/apps/events/services/formatting.js
yeahfro: Discussion started on slack channel

src/apps/events/services/formatting.js
yeahfro: Ben at this stage start date and end date are optional fields. Moving forward only end date will be an optional field.

Alternative code will be as follows:

<pre><code>
  formatted['start_date'] = [
    body['start_date_year'],
    body['start_date_month'],
    body['start_date_day'],
  ].join('-')

  if (!isEmpty(body['end_date_year']) &&
      !isEmpty(body['end_date_month']) &&
      !isEmpty(body['end_date_day'])) {
    formatted['end_date'] = [
      body['end_date_year'],
      body['end_date_month'],
      body['end_date_day'],
    ].join('-')
  }</code></pre>

My original code is more fun but I appreciate this might be easier to read.

feedmypixel: It feels like what we need to be doing is:

- If a `month`, `year` or `day` has been entered build `formatted['start_date']` and let it through for the API to validate
- if all three (`month`, `year` or `day`) have *NOT* been entered, skip and do not send `formatted['start_date']`


With this we are validating in the `transformToApi` and then on the backend

Does the api not give you the error back if you do the following without all three values?
```
formatted[key] = [
  body[`${key}_year`],
  body[`${key}_month`],
  body[`${key}_day`],
].join('-')
```

src/apps/events/middleware/details.js
```diff
@@ -1,8 +1,21 @@
 const { transformToApi } = require('../services/formatting')
 const { createEvent } = require('../repos')
+const { includes, assign, castArray, compact } = require('lodash')
 
 async function handleFormPost (req, res, next) {
-  const formattedBody = transformToApi(Object.assign({}, req.body))
+  const castToArrayAndRemoveEmpty = (value) => compact(castArray(value))
```
feedmypixel: 👍

src/apps/events/services/formatting.js
yeahfro: As discussed I have addressed concerns

src/apps/events/services/formatting.js
```diff
@@ -33,17 +33,21 @@ function transformToApi (body) {
...
   const setDate = (key) => {
-    formatted[key] = [
+    const dateParts = [
       body[`${key}_year`],
       body[`${key}_month`],
       body[`${key}_day`],
-    ].join('-')
+    ]
+
+    if (some(dateParts)) {
```
feedmypixel: 👍

PR #600, opened 2017-09-18
Fix cast to string bug in multiple choice macro

teneightfive:
This moves the check to deciding to select an option or not inside a
check for an expected value. It then casts those values both to a
string so options are correctly selected.


tyom:
Would be good to add tests for this but there are no tests for this macro at all. Is it a good time to add some?

PR #605, opened 2017-09-18
bump node version to 8.5.0

feedmypixel:
_does what it says on the tin_


feedmypixel:
> What about >= 8.5.0 ?

I feel we should be explicit with the node version. It changes when we change it, rather than "magical" things happening when we release.

PR #599, opened 2017-09-18
Run acceptance tests CircleCI

feedmypixel:
This work covers:

- [x] Running acceptance tests on `CircleCi`
- [x] Migrating to `CircleCi 2.0` workflows so we now have the following workflows:
  - `unit_tests`
  - `lint_code`
  - `acceptance_tests`
- [x] Providing html reports from `Nightwatch.js`
- [x] Move ~[feedmypixel/java-node-chrome](https://hub.docker.com/r/feedmypixel/java-node-chrome/)~ to `DIT` [docker-nightwatch-base](https://hub.docker.com/r/ukti/docker-nightwatch-base/) dockerhub
- [x] Move ~[feedmypixel/java-node-chrome-docker](https://github.com/feedmypixel/java-node-chrome-docker)~ to `DIT` [docker-nightwatch-base](https://github.com/uktrade/docker-nightwatch-base) GitHub
- [x] Comment out `acceptance_tests` job in CircleCI until backend test capacity is addressed.

## TODO
turn on required status checks for new jobs in GitHub

## Of Note
- As discussed I will comment out the `acceptance_tests` job until we can address the backend capacity issue.

- Currently Acceptance tests are not green so this job will not be a required status check until we have got on top of the tests and they are in a much more stable state

- GitHub repo [docker-nightwatch-base](https://github.com/uktrade/docker-nightwatch-base) created

- Docker Hub repo [docker-nightwatch-base](https://hub.docker.com/r/ukti/docker-nightwatch-base/) created

### Without Acceptance tests
screen shot 2017-09-18 at 13 08 31

### With Acceptance tests
screen shot 2017-09-18 at 08 28 33

PR #606, opened 2017-09-18
update npm deps

feedmypixel:
- updated dependencies
- fix test errors due to updating dependencies
- fix linting errors due to updating dependencies

Bit weird that we have had errors in tests appear due to updating dependencies. But as you can see they are real errors.

Linting also appears to of got stricter with updating dependencies. Although it may have something to do with [#3ee0fd6c](https://github.com/uktrade/data-hub-frontend/pull/599/commits/3ee0fd6cf5c175ccc1059e4fdeb1fb36ab463e01)

I have ran unit tests and linting and logged in. Everything seems ok!! (looking forward to the acceptance test reassurance)

## Of Note
I have a feeling we have a few dependencies that are not in use. I will do a clean up in a separate PR.


test/unit/macros/form.test.js
```diff
@@ -261,8 +261,8 @@ describe('Nunjucks form macros', () => {
             })).parentElement
             const options = component.querySelectorAll('.c-multiple-choice__input')
 
-            expect(options[0].getAttribute('checked')).to.be.mull
-            expect(options[1].getAttribute('checked')).to.be.mull
```
yeahfro: Revert these changes in December!

PR #608, opened 2017-09-19
add pingdom xml route to healthcheck

feedmypixel:
Turns out pingdom does need a `ping.xml` file so this work adds that back in.

### DIFF
--- a/test/unit/apps/healthcheck/controllers/healthcheck.test.js
+++ b/test/unit/apps/healthcheck/controllers/healthcheck.test.js
@@ -29,15 +29,30 @@ describe('Health check controller', () => {
       this.sandbox.restore()
     })
 
-    it('should render the healthcheck page with OK response', async () => {
+    it('should set cache control', async () => {
       await this.controller.getHandler(this.req, this.res, this.next)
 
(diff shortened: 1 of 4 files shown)


test/unit/apps/healthcheck/controllers/healthcheck.test.js
```diff
@@ -89,4 +89,103 @@ describe('Health check controller', () => {
...
+        }),
+      }
+      this.next = this.sandbox.spy()
+    })
+
+    afterEach(() => {
+      this.sandbox.restore()
+    })
+
+    it('should render the healthcheck page with OK response', async () => {
```
teneightfive: I've started to try and break my `its` out a little more to describe what it's actually checking. I find it easier to find when something fails and also easier to read what behaviour is being expected.

So rather than one large `should render the healthcheck page with OK response` it could be broken down into smaller chunks:

- `it('should set content type and cache control')`
- `it('should return a 200 status code')`
- `it('should return OK')`
- `it('should not call the logger')`

Not necessarily suggesting you need to spend time changing this test now but more food for thought for the future.

feedmypixel: agreed - have split them apart

PR #610, opened 2017-09-19
change redis client to ioredis and add sentinel support

jim68000:
This unblocks Bowen in the creation of our GD PaaS installation. 

There are two issues with this PR:

1. There's no way to instantiate the ioredis client without using 'new' as it's OO all the way down. The damage is limited to one file however.

2. I had hoped to replace connect-redis-crypto with this commit, but it's still broken; currently the version we use, 3.1.1, causes node 8.5 to emit a warning for every request. I have a plan to replace this but it's bigger in imapct, so I've removed it from this PR.

### DIFF
--- a/config/redis-store.js
+++ b/config/redis-store.js
@@ -1,4 +1,4 @@
-const redis = require('redis')
+const Redis = require('ioredis')
 const redisCrypto = require('connect-redis-crypto')
 const session = require('express-session')
 const url = require('url')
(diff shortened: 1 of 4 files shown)


config/redis-store.js
```diff
@@ -10,12 +10,20 @@ const RedisStore = redisCrypto(session)
 
 let client
 
-if (config.redis.url) {
+if (config.redis.url && config.redis.sentinel) {
   const redisURL = url.parse(config.redis.url)
-  client = redis.createClient(redisURL.port, redisURL.hostname, { no_ready_check: true })
+
+  client = new Redis({
```
tyom: Would it be better to change configuration for each condition and instantiate Redis once at the end.

jim68000: It would be tidier, yes. Let me do that.

config/redis-store.js
```diff
@@ -9,15 +9,34 @@ const config = require('./')
...
+  }
 } else {
-  client = redis.createClient(config.redis.port, config.redis.host)
+  redisConfig = {
+    port: config.redis.port,
+    host: config.redis.host,
+  }
 }
 
+client = new Redis(redisConfig)
```
tyom: You coul `const` right here instead of `let` above. You could also set default `redisConfig` object and just change the properties for specific conditions.

jim68000: Done. I did mean to do that, and then forgot...

PR #611, opened 2017-09-19
Extract each macro into individual unit test suite

tyom:
Managing unit tests for macro was becoming more difficult as typically contain macros within collection files and testing individual macros within those collection was resulting in huge test files.

This change extracts each macro as a separate test suite under directories representing macro collection files.


tyom:
> Let's stay consistent and use kebab case for file names. On the second thought...

They are PascalCase in order to tie them to macro names rather than files. Macro container files are expressed as directories and macro names are as files.

The macros, being "components" are very different from all other file types so I felt that it makes sense to distinguish them in unit tests as such and avoid making the developer doing that conversion in their head.

Unfortunately it seems impossible to split macros into separate file and load them into the container (like we can in `index.js`, for example). Having them in one container files allows us to include private helper macros within those containers so they can be used to support macros with shared functionality. It also makes it easier to import several macros in one go.

The alternative is to split each macro into individual file but I'm not sure how we'd avoid losing the benefits I mentioned above.

tyom:
After some investigation I discovered a way to create container files which can import individual macros. Since those will use kebab-case for file names  I've updated the tests to match.

PR #612, opened 2017-09-19
OMIS Work order amends

teneightfive:
This amends one of the work description fields to always be displayed on the work order whether or not it has value and to use a fallback if it doesn't.

It also includes changes to how fallback text is output to allow the component to control it. This means it can have a different visual style applied to allow it to be distinguished from user generated content.

## Before

localhost_3001_omis_10613433-3b9f-4e2d-9617-02769a84e876_work-order 1

localhost_3001_omis_10613433-3b9f-4e2d-9617-02769a84e876_work-order 3

## After

localhost_3001_omis_10613433-3b9f-4e2d-9617-02769a84e876_work-order

localhost_3001_omis_10613433-3b9f-4e2d-9617-02769a84e876_work-order 2

### DIFF
--- a/src/templates/_macros/common.njk
+++ b/src/templates/_macros/common.njk
@@ -196,9 +196,11 @@
  # Render answers summary table
  #
  # @param {object} props
- # @param {object|string[]} props.items - array of ojects or strings to display in table body
+ # @param {object[]|string[]} props.items - array of ojects or strings to display in table body
  # @param {string} props.items.value - item value
  # @param {string} props.items.label - item label
+ # @param {string} [props.items.fallbackText] - fallback text if value is empty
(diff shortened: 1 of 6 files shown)


src/templates/_macros/common.njk
```diff
@@ -196,9 +196,11 @@
  # Render answers summary table
  #
  # @param {object} props
- # @param {object|string[]} props.items - array of ojects or strings to display in table body
+ # @param {object[]|string[]} props.items - array of ojects or strings to display in table body
  # @param {string} props.items.value - item value
  # @param {string} props.items.label - item label
+ # @param {string} [props.items.fallbackText] - fallback text if value is empty
+ # @param {string} [props.fallbackText] - component level fallback when items is a list of string
```
tyom: `strings`

PR #613, opened 2017-09-19
DH-524 Save event acceptance test clean up

yeahfro:
These changes include:
- Updating event field names to match recently changed field names
- Using assert rather than verify

Next:
- Passing tests for saving events


test/acceptance/features/step_definitions/event-create.js
```diff
@@ -13,87 +13,92 @@ defineSupportCode(({ Given, Then, When }) => {
 
   Then(/^I verify the event name field is displayed$/, async () => {
     await Events
-      .verify.visible('@eventName')
+      .assert.visible('@eventName')
```
tyom: Do we need to hard fail? Would it be more useful to continue with the rest of the tests to get the summary of all errors?

feedmypixel: thats good point about the summary, but if running on CI it could be expensive to wait for all tests to pass to get a summary.

Fail fast would mean more running of tests but then dealing with one fail at a time, which would mean:
- it fails on CI quick
- then you fix it locally and failing fast locally means dealing with one thing at a time

feedmypixel: Maybe a symptom of wanting to fail early is because we have so many failures at present... another for the testing discussion

PR #614, opened 2017-09-19
OMIS edit return url support

teneightfive:
This uses the `returnURL` query string param and sets the backLink and
next properties on the form options object so that they can be used
to change the destinations.


src/apps/omis/apps/edit/controllers/assignee-time.js
```diff
@@ -22,14 +22,15 @@ class EditAssigneeHoursController extends EditController {
 
     try {
       await Order.saveAssignees(req.session.token, res.locals.order.id, filter(assignees))
+      const next = req.form.options.next || `/omis/${res.locals.order.id}`
```
tyom: Name class with middleware parameter can be confusing.

PR #616, opened 2017-09-19
Events list page

feedmypixel:
I wanted to put this up for discussion around points 2 and 3 below. Once done I can get on with the tests

- [x] Events link in navigation
- [x] Events listing page
- [x] Tests (only unit - will do acceptance tests once #631 is in)

## Of Note
Sorting is reliant on https://github.com/uktrade/data-hub-leeloo/pull/478

screencapture-localhost-3000-events-1505896734884

### DIFF
--- a/src/apps/events/controllers/list.js
+++ b/src/apps/events/controllers/list.js
@@ -1,0 +1,33 @@
+const { get, omit, merge } = require('lodash')
+const { eventFiltersFields: filtersFields, eventSortForm } = require('../macros')
+const { buildSelectedFiltersSummary } = require('../../builders')
+
+function renderEventList (req, res) {
+  const sortForm = merge({}, eventSortForm, {
+    hiddenFields: Object.assign({}, omit(req.query, 'sortby')),
+    children: [
+      { value: req.query.sortby },
+    ],
+  })
+
+  const selectedFilters = buildSelectedFiltersSummary(filtersFields, req.query)
+  const isUKSelected = get(selectedFilters, 'address_country.valueLabel') === 'United Kingdom'
+
+  const visibleFiltersFields = filtersFields.filter(field => {
+    if (field.name === 'uk_region') {
+      return isUKSelected
+    }
+    return true
+  })
+
+  res.render('events/views/list', {
+    title: 'Events',
+    sortForm,
+    filtersFields: visibleFiltersFields,
+    selectedFilters,
+  })
+}
+
+module.exports = {
+  renderEventList,
+}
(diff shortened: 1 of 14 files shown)


src/apps/events/router.js
```diff
@@ -1,10 +1,27 @@
 const router = require('express').Router()
 
+const { setDefaultQuery } = require('../middleware')
+const { getRequestBody, getEventsCollection } = require('./middleware/collection')
+const { renderEventList } = require('./controllers/list')
+
 const {
   details,
```
feedmypixel: @yeahfro as mentioned I feel we may be loosing some clarity here using `index.js` files. I quite like the approach taken with variable names a good example can be seen in [contacts router](https://github.com/uktrade/data-hub-frontend/blob/develop/src/apps/contacts/router.js)

ztolley:
Also needs an add button

src/apps/events/controllers/list.js
```diff
@@ -0,0 +1,33 @@
+const { get, omit, merge } = require('lodash')
+const { eventFiltersFields: filtersFields, eventSortForm } = require('../macros')
+const { buildSelectedFiltersSummary } = require('../../builders')
+
+function renderEventList (req, res) {
+  const sortForm = merge({}, eventSortForm, {
+    hiddenFields: Object.assign({}, omit(req.query, 'sortby')),
```
yeahfro: How about `assign` over `Object.assign`?

src/apps/events/labels.js
```diff
@@ -0,0 +1,14 @@
+const labels = {
+  collectionFilterLabels: {
```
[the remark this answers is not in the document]
yeahfro: Thank you!

test/unit/data/events/event-data.json
```diff
@@ -0,0 +1,56 @@
+{
```
feedmypixel: @yeahfro FYI I have dropped in a fuller `event-data.json` mock. I didn't want to remove the current [event.json](https://github.com/uktrade/data-hub-frontend/blob/develop/test/unit/data/events/event.json) as it shouldn't be part of this PR. But it feels like maybe we should just have one?

src/apps/events/transformers.js
```diff
@@ -0,0 +1,90 @@
...
+  address_1,
+  address_2,
+  address_town,
+  address_county,
+  address_postcode,
+  notes,
+  organiser,
+  lead_team,
+  uk_region,
+} = {}) {
```
tyom: I think this is unnecessary. We expect the first argument to be an object.

PR #617, opened 2017-09-19
Dashboard links

teneightfive:
This change removes the need for client side code to show/hide links.

## Before

localhost_3001_ 1

## After

localhost_3001_


src/apps/dashboard/controllers.js
```diff
@@ -6,12 +6,15 @@ function getHandler (req, res) {
...
     .then((data) => {
+      // TODO: Remove manual limit on interactions and contacts once API supports it
       res
         .title('Dashboard')
         .render('dashboard/views/dashboard', {
           totalDays: days,
-          interactions: data.interactions,
-          contacts: data.contacts,
+          interactionsCount: data.interactions.length,
+          interactions: data.interactions.slice(0, 5),
```
feedmypixel: are we always going to have data to `slice`?

src/apps/dashboard/controllers.js
```diff
@@ -6,12 +6,15 @@ function getHandler (req, res) {
...
+          contactsCount: data.contacts.length,
+          contacts: data.contacts.slice(0, 5),
```
feedmypixel: same as above

src/apps/dashboard/controllers.js
```diff
@@ -6,12 +6,15 @@ function getHandler (req, res) {
   dashboardService
     .getHomepageData(req.session.token, days)
```
feedmypixel: feels like we could do with a `limit=10` rather `days` on this

feedmypixel: @reupen FYI ☝️

teneightfive: I would agree as the response to the dashboard is currently very slow as it's returning ~200 items.

I would say for now we should limit to last 5 and we can revisit properly when we look at what the dashboard should actually contain.

feedmypixel: > I would say for now we should limit to last 5 and we can revisit properly when we look at what the dashboard should actually contain.

👍

PR #622, opened 2017-09-20
Removal of more deprecated code

teneightfive:
This cleans up a few bits of deprecated code that were moved over when dependencies were brought into this app.

The interface is now making more use of newer macros which has removed the need for some of these scripts.


tyom:
LGTM

PR #624, opened 2017-09-20
Use new form macros in investment requirements form

tyom:
The form build in old style. This refactor moves it to new form macros as well as fixing some form state bugs found in the process. It also removes unused `form-multiple-choice` legacy component.


src/apps/investment-projects/views/requirements-edit.njk
```diff
@@ -4,89 +4,5 @@
...
-        value: 'true'
-      }, {
-        label: 'No',
-        value: 'false'
-      }]
-    } %}
-
-    {{ save(backUrl='/investment-projects/' + investmentData.id) }}
-  </form>
+  {{ Form(requirementsForm | assign({ returnLink: '/investment-projects/' + investmentData.id })) }}
```
teneightfive: `| assign` or `| assignCopy`?

teneightfive: Also, why not build that where `requirementsForm` is being built?

tyom: Good point.

PR #625, opened 2017-09-20
Merge company details templates into one.

ztolley:
Replace the 4 separate view templates used for company details into a single nunjucks template.

Foreign:
https://datahub-beta2-pr-625.herokuapp.com/companies/view/foreign/b2c34b41-1d5a-4b4b-9249-7c53ff2868dd

ukother:
https://datahub-beta2-pr-625.herokuapp.com/companies/view/ukother/dcdabbc9-1781-e411-8955-e4115bead28a

ch only:
https://datahub-beta2-pr-625.herokuapp.com/companies/view/ch/03588574

ltd:
https://datahub-beta2-pr-625.herokuapp.com/companies/view/ltd/725c2b53-ee88-43f2-b584-672171ca8538


README.md
```diff
@@ -267,7 +267,7 @@ layouts/
 | |                            |
 | +------+ _layout.view.njk <-+|
 |                             ||
-|           details-ltd.njk +-+|
+|           details.njk +-+|
```
teneightfive: Not sure you need to update this as such as I think it was showing example structures. It feels little out of date now too.

src/apps/companies/controllers/ch.js
```diff
@@ -17,9 +17,11 @@ async function getDetails (req, res, next) {
     res.locals.chDetailsLabels = chDetailsLabels
     res.locals.addUrl = `/companies/add/ltd/${company.company_number}`
 
+    res.locals.companyType = 'ltd'
```
teneightfive: Should this be `ch`?

README.md
ztolley: It may be out of date just didn't want to leave behind a reference to a file that doesn't exist

src/apps/companies/controllers/ch.js
ztolley: ends up not being used so I'll removeit

src/apps/companies/controllers/foreign.js
```diff
@@ -24,9 +24,10 @@ async function getDetails (req, res, next) {
     }
     res.locals.accountManagementDisplayLabels = accountManagementDisplayLabels
 
+    res.locals.companyType = 'foreign'
```
teneightfive: I wonder whether these should actually be passed into the `render()` method as the second parameter so it's easier to read.

When there's lots of repetition of things like `res.locals` its good to try and refactor it.

teneightfive: Although, thinking about it, that should be a separate change. 

Will approve this one. If you have time maybe look at the controllers next?

ztolley: There's the whole let's refactor and stop using res.locals in a controller thing. This is just merging templates, after all, I was told only do tiny bits at a time.

PR #629, opened 2017-09-20
Move the global nav outside of the main container

teneightfive:
The global nav shouldn't sit within the main content area. This moves
it to outside in the DOM.


teneightfive:
Duplicate of #630.

PR #631, opened 2017-09-20
Restructure acceptance test folders

yeahfro:
This change restructure acceptance tests which will assist us with finding a repeatable pattern.

screen shot 2017-09-20 at 15 56 07


test/acceptance/nightwatch.conf.js
```diff
@@ -1,18 +1,24 @@
 require('dotenv').config()
 const seleniumServer = require('selenium-server')
 const chromeDriver = require('chromedriver')
+const { readdirSync } = require('fs')
+const { join } = require('path')
+const { without } = require('lodash')
+
+const features = readdirSync(join(__dirname, 'features'))
```
feedmypixel: could you just have `readdirSync('./features')` ?

test/acceptance/nightwatch.conf.js
```diff
@@ -1,18 +1,24 @@
...
+const apps = without(features, 'step-definitions')
```
feedmypixel: not sure if `apps` is the right name here? maybe `pageObjects`?

feedmypixel: could you be explicit in getting `pageObject`? by using a filter and only gathering `page-objects` paths?

test/acceptance/nightwatch.conf.js
```diff
@@ -1,18 +1,24 @@
...
-    '--require', 'test/acceptance/features/step_definitions',
+    '--require', 'test/acceptance/features',
     '--format', 'json:cucumber/reports/tests.json',
     process.env.FEATURES_FOLDER || 'test/acceptance/features',
   ],
 })
 
 module.exports = {
-  page_objects_path: 'test/acceptance/page_objects',
+  page_objects_path: apps.map((app) => { return `test/acceptance/features/${app}/page-objects` }),
```
feedmypixel: if you create an array of `page-objects` paths above then you shouldn't have to do this

test/acceptance/nightwatch.conf.js
```diff
@@ -1,18 +1,24 @@
 require('dotenv').config()
 const seleniumServer = require('selenium-server')
 const chromeDriver = require('chromedriver')
+const { readdirSync } = require('fs')
+
+const pageObjects =
+  readdirSync('./features')
```
feedmypixel: so this needs to be `test/acceptance/features` as it currently fails when running `yarn test:acceptance`

test/acceptance/nightwatch.conf.js
```diff
@@ -1,18 +1,24 @@
 require('dotenv').config()
 const seleniumServer = require('selenium-server')
 const chromeDriver = require('chromedriver')
+const { readdirSync } = require('fs')
+
+const pageObjects =
```
feedmypixel: sorry my bad after checking it out feels like these are `pageObjectPaths`?

test/acceptance/nightwatch.conf.js
```diff
@@ -1,18 +1,24 @@
...
+  readdirSync('./features')
+    .filter((featuresFolderName) => { return featuresFolderName !== 'step-definitions' })
```
feedmypixel: formatting is a little funky. Would suggest:
```
const pageObjectPaths = readdirSync('test/acceptance/features')
  .filter((featuresFolderName) => {
    return featuresFolderName !== 'step-definitions'
  })
  .map((appName) => {
    return `test/acceptance/features/${appName}/page-objects`
  })
```
or

```
const pageObjectPaths = readdirSync('test/acceptance/features')
  .filter((featuresFolderName) => featuresFolderName !== 'step-definitions')
  .map((appName) => `test/acceptance/features/${appName}/page-objects`)
```

test/acceptance/nightwatch.conf.js
tyom: Could just do `const pageObjects = glob.sync('test/acceptance/features/**/page-objects')` with `glob`.

PR #634, opened 2017-09-20
Remove deprecated flash

teneightfive:
This change removes the deprecated flash component and replaces its last use with the messages style.

## Before

localhost_3001_investment-projects_b75f2a3b-df25-4e56-b526-33268ef1c854_details 1

## After

localhost_3001_investment-projects_b75f2a3b-df25-4e56-b526-33268ef1c854_details

### DIFF
--- a/assets/stylesheets/components/_messages.scss
+++ b/assets/stylesheets/components/_messages.scss
@@ -36,7 +36,7 @@
     }
   }
 
-  > * + * {
+  > *:not([type="hidden"]) + * {
     margin-top: $default-spacing-unit;
     margin-bottom: 0;
   }
--- a/assets/stylesheets/_deprecated/_flash.scss
+++ b/assets/stylesheets/_deprecated/_flash.scss
@@ -1,8 +1,0 @@
-.flash {
-  padding: 15px 20px;
-
-  &.flash--info {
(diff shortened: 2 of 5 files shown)


assets/stylesheets/components/_messages.scss
```diff
@@ -36,7 +36,7 @@
     }
   }
 
-  > * + * {
+  > *:not([type="hidden"]) + * {
```
tyom: Can you explain why we need this in messages component?

tyom: I see. Because we sometimes use it as form.

teneightfive:
>I'm not so sold on the design TBH

Good job you're not the user :trollface: 

I can kind of see what you mean but we can work on those later now we know where their use is. Maybe it doesn't fit for the thing you're trying to do in this instance any way and it might need a different component all together.

The style of these messages was actually taken from the [digital marketplace](http://alphagov.github.io/digitalmarketplace-frontend-toolkit/notification-banner.html) and they've done a lot of research around this kind of thing I imagine. And to be blunt, they're not there to look pretty, they're there to do a job of notifying the user of something so standing out is a good thing in my opinion.

PR #636, opened 2017-09-20
Remove deprecated statusbadge

tyom:
This was only used in one place - local header for contact (primary).

I added matching styles directly in local header for badge. Didn't want to bring those into generic `c-badge` component as I'm not sure we'll be keeping it there.

### Before
image

### After
image


assets/stylesheets/components/_local-header.scss
```diff
@@ -55,6 +55,14 @@ $_first-element-spacing: $default-spacing-unit * 3;
   .c-meta-list {
     @include core-font(16);
   }
+
+  .c-badge {
```
teneightfive: I know you didn't want to add them to the component but I think these would be better there as a modifier rather than randomly in this component.

We should avoid lots of modifying components when they're in others. It feels anti-BEM.

tyom: This is quite specific to this context. I don't feel comfortable adding this in generic component even as modifier. I think this will probably go away sooner rather than later.

teneightfive: Could it not be 2 modifiers in that component? One for a large style and then add a `.c-badge--tertiary` modifier for the colour? 

Or even use the `$purple` instead of `$fuschia`.

tyom: Why add all that unneeded stuff to a generic component. We can add it all there but then we wouldn't know if it's used any more. Be wary of removing it and it will stay there for ever.

At least this way we know that only in the context of local header do we adjust the styles of the badge. And when we remove it we know we can remove it from local header as we can check the macro for its existence.

teneightfive: I think it's actually harder to know at a later date if its being used or not with this implementation. 

You're just using the `.c-badge` selector. Searching for it will return results, and you'll have to start looking up parents.

At least if its implemented as a clear modifier, a search later on will reveal if its being used anywhere or not much more easily.

teneightfive:
A screenshot of before/after might also help with reviewing this change as it looks like the visual style will have changed slightly.

assets/stylesheets/components/_local-header.scss
tyom: The modifier is also meant to be generic. In this instance it's basically replacing like for like. I don't think we should be introducing a modifier for a temporary solution. I'd much rather leave the specific styles in the context they are set and review/remove them at a later stage.

Having this modifier there would suggest to other people that it's OK to use it.

teneightfive: If that's what you think is best I'll approve it. Maybe we can review the use of badges as a separate thing as I've started to notice that colour isn't a good identifier that a badge is a link.

This implementation breaks that in particular.

tyom: I agree. The badges need a review. But we should do it as a separate task.

PR #640, opened 2017-09-20
Replace use of infostrip

teneightfive:
This change replaces one of the deprecated sass components in favour of the newer messages component.

#### Things out of scope for future change

- Refactor `.c-messages` component so that it doesn't require a container and the selector can be `.c-message` instead and therefore be applied to individual items as well as in a list

### Before
localhost_3001_investment-projects_create_equity-source_82fd0dab-ee6b-e411-a72b-e4115bead28a_term something 1
### After
localhost_3001_investment-projects_create_equity-source_82fd0dab-ee6b-e411-a72b-e4115bead28a_term something

### Before
localhost_3001_investment-projects_f322a0f8-818d-41c8-bd39-ed861383a739_details
### After
localhost_3001_investment-projects_f322a0f8-818d-41c8-bd39-ed861383a739_details 1

### Before
localhost_3001_investment-projects_f322a0f8-818d-41c8-bd39-ed861383a739_team
### After
localhost_3001_investment-projects_f322a0f8-818d-41c8-bd39-ed861383a739_team 1

### Before
localhost_3001_companies_view_ukother_82fd0dab-ee6b-e411-a72b-e4115bead28a 1
### After
localhost_3001_companies_view_ukother_82fd0dab-ee6b-e411-a72b-e4115bead28a

### Before
localhost_3001_companies_add-step-2_business_type ltd country uk term samsung 1
### After
localhost_3001_companies_add-step-2_business_type ltd country uk term samsung

## Company no contacts
The logic on the panel actually meant it wasn't being displayed so the screenshot here shows what it would have looked like.

The panel has been removed in this update though as the page already clearly shows that there are 0 contacts and has a _Add contact_ call to action.

### Before
localhost_3001_companies_82fd0dab-ee6b-e411-a72b-e4115bead28a_contacts 3
### After
localhost_3001_companies_82fd0dab-ee6b-e411-a72b-e4115bead28a_contacts 2

### Before
localhost_3001_companies_82fd0dab-ee6b-e411-a72b-e4115bead28a_interactions
### After
localhost_3001_companies_82fd0dab-ee6b-e411-a72b-e4115bead28a_interactions 1

### Before
localhost_3001_omis_96c8576e-fa76-4202-84c2-77d646d1d909_work-order 1
### After
localhost_3001_omis_96c8576e-fa76-4202-84c2-77d646d1d909_work-order


src/apps/companies/views/_layout-view.njk
```diff
@@ -8,7 +8,7 @@
   %}
     <p class="c-local-header__heading-after">{{ headingAddress }}</p>
     {% if company.archived %}
-      <p class="infostrip">
+      <p class="c-messages__item c-messages__item--info">
```
tyom: Surely using element outside of block is not proper BEM.

teneightfive: Correct. But this pattern has no evolved. I mentioned in the PR description that this is something that needs to be addressed but feels out of scope of this PR because we are using this child element class elsewhere.

As we discussed yesterday I think it might be better suited as a `c-message`.

tyom: If it's out of scope, let's get that in first and then update this PR with new component.

assets/stylesheets/components/_messages.scss
```diff
@@ -23,7 +23,7 @@
     }
   }
 
-  + * {
+  * + & {
```
tyom: Is this right? Shouldn't it be `& + &`?

Otherwise it compiles to
```
* + .c-message-list { ... }
```
which is what assemblies are for.

teneightfive: Maybe it could be moved to assemblies.

PR #646, opened 2017-09-21
Add detail view to event sub app

ztolley:
Adds the detail view to the events section to view an event.

capture
(updated capture to show almost fully populated and using updated component)

Transforms raw event data into the key/value style for viewing, with some simple modifications to handle dates and 'other teams'.

- [x] Refactor tests to use context
- [x] Switch to use component for key-value-table
- [x] Update key-value-table to support date type, and format it, e.g. 12 January 2017 (details have long dates compared to lists
- [x] Refine transformer to use ideas from Tyom and not format date
- [x] Expand parameters like collections transformers

### DIFF
--- a/src/apps/events/controllers/details.js
+++ b/src/apps/events/controllers/details.js
@@ -1,9 +1,20 @@
-async function renderDetailsPage (req, res) {
-  res
-    .breadcrumb('Event details')
-    .render('events/views/details', {
-      title: 'Event details',
-    })
+const { fetchEvent } = require('../repos')
+const { transformEventResponseToViewRecord } = require('../transformers')
+
+async function renderDetailsPage (req, res, next) {
+  try {
+    const event = await fetchEvent(req.session.token, req.params.id)
+    const eventViewRecord = transformEventResponseToViewRecord(event)
+
+    res
+      .breadcrumb(event.name)
+      .render('events/views/details', {
+        eventViewRecord,
+        eventId: req.params.id,
+      })
+  } catch (error) {
+    next(error)
+  }
 }
 
 module.exports = {
--- a/src/apps/events/repos.js
+++ b/src/apps/events/repos.js
@@ -9,6 +9,11 @@ function createEvent (token, body) {
   })
 }
 
+function fetchEvent (token, id) {
+  return authorisedRequest(token, `${config.apiRoot}/v3/event/${id}`)
+}
+
 module.exports = {
   createEvent,
+  fetchEvent,
 }
--- a/src/apps/events/transformers.js
+++ b/src/apps/events/transformers.js
@@ -1,6 +1,8 @@
 /* eslint-disable camelcase */
 const { get } = require('lodash')
 
+const { getFormattedAddress } = require('../../lib/address')
+
 function transformEventToListItem ({
   id,
   name,
@@ -85,6 +87,72 @@ function transformEventToListItem ({
   return item
 }
 
+function transformEventResponseToViewRecord ({
+  event_type,
+  start_date,
+  end_date,
+  location_type,
+  address_1,
+  address_2,
+  address_town,
+  address_county,
+  address_postcode,
+  address_country,
+  uk_region,
+  notes,
+  lead_team,
+  organiser,
+  teams,
+  related_programmes,
+  service,
+}) {
+  teams = teams || []
+  related_programmes = related_programmes || []
+
+  const transformedEvent = {
+    'Type of event': event_type,
+  }
+
+  if (start_date === end_date) {
+    transformedEvent['Event date'] = {
+      type: 'date',
+      name: start_date,
+    }
+  } else {
+    transformedEvent['Event start date'] = {
+      type: 'date',
+      name: start_date,
+    }
+    transformedEvent['Event end date'] = {
+      type: 'date',
+      name: end_date,
+    }
+  }
+
+  return Object.assign(transformedEvent, {
+    'Event location type': location_type,
+    'Address': getFormattedAddress({
+      address_1,
+      address_2,
+      address_town,
+      address_county,
+      address_postcode,
+      address_country,
+    }),
+    'Region': uk_region,
+    'Notes': notes,
+    'Lead team': lead_team,
+    'Organiser': organiser,
+    'Other teams': teams
+      .filter(team => team.id !== lead_team.id)
+      .map(item => item.name),
+    'Related programmes': related_programmes
+      .map(item => item.name),
+    'Service': service,
+  })
+}
+
 module.exports = {
   transformEventToListItem,
+  transformEventResponseToViewRecord,
 }
--- a/src/apps/events/views/details.njk
+++ b/src/apps/events/views/details.njk
@@ -1,7 +1,6 @@
 {% extends "_layouts/two-column.njk" %}
 
-{% block main_grid_left_column %}{% endblock %}
-
 {% block main_grid_right_column %}
-
+  {% component 'key-value-table', variant='striped', items=eventViewRecord %}
+  <a class="button button-secondary" href="{{ eventId }}/edit">Edit Event</a>
 {% endblock %}
--- a/test/unit/apps/events/controllers/details.test.js
+++ b/test/unit/apps/events/controllers/details.test.js
@@ -1,41 +1,85 @@
+const eventData = require('~/test/unit/data/events/event-data')
+
 describe('Event details controller', () => {
   beforeEach(() => {
-    this.controller = require('~/src/apps/events/controllers/details')
-
+    this.transformedData = {}
     this.sandbox = sinon.sandbox.create()
 
-    this.req = { }
+    this.fetchEventStub = this.sandbox
+      .stub()
+      .withArgs('4321', '1234')
+      .resolves(eventData)
+
+    this.controller = proxyquire('~/src/apps/events/controllers/details', {
+      '../transformers': {
+        transformEventResponseToViewRecord: this.sandbox
+          .stub()
+          .withArgs(eventData)
+          .returns(this.transformedData),
+      },
+      '../repos': {
+        fetchEvent: this.fetchEventStub,
+      },
+    })
+
+    this.req = {
+      params: {
+        id: '1234',
+      },
+      session: {
+        token: '4321',
+      },
+    }
+
     this.res = {
       breadcrumb: this.sandbox.stub().returnsThis(),
       render: this.sandbox.spy(),
     }
+
+    this.next = this.sandbox.spy()
   })
 
   afterEach(() => {
     this.sandbox.restore()
   })
 
-  describe('#renderDetailsPage', () => {
-    it('should add a breadcrumb', async () => {
-      this.controller.renderDetailsPage(this.req, this.res)
+  describe('#renderDetailsPage', async () => {
+    context('when there are no errors', () => {
+      beforeEach(async () => {
+        await this.controller.renderDetailsPage(this.req, this.res, this.next)
+      })
 
-      expect(this.res.breadcrumb).to.be.calledWith('Event details')
-      expect(this.res.breadcrumb).to.have.been.calledOnce
+      it('should add a breadcrumb', () => {
+        expect(this.res.breadcrumb).to.be.calledWith(eventData.name)
+        expect(this.res.breadcrumb).to.have.been.calledOnce
+      })
+
+      it('should render the event details template', () => {
+        expect(this.res.render).to.be.calledWith('events/views/details')
+        expect(this.res.render).to.have.been.calledOnce
+      })
+
+      it('should return transformed events data', () => {
+        const options = this.res.render.firstCall.args[1]
+        expect(options.eventViewRecord).to.deep.equal(this.transformedData)
+      })
+
+      it('should return the event id', () => {
+        const options = this.res.render.firstCall.args[1]
+        expect(options.eventId).to.equal('1234')
+      })
     })
 
-    it('should render the event details page', async () => {
-      await this.controller.renderDetailsPage(this.req, this.res)
+    context('when there are errors fetching data', async () => {
+      beforeEach(async () => {
+        this.error = new Error()
+        this.fetchEventStub.rejects(this.error)
+        await this.controller.renderDetailsPage(this.req, this.res, this.next)
+      })
 
-      expect(this.res.render).to.be.calledWith('events/views/details')
-      expect(this.res.render).to.have.been.calledOnce
-    })
-
-    it('should render the event details page with a title', async () => {
-      await this.controller.renderDetailsPage(this.req, this.res, this.next)
-
-      const actual = this.res.render.getCall(0).args[1].title
-
-      expect(actual).to.equal('Event details')
+      it('should pass the error to the error page', () => {
+        expect(this.next).to.be.calledWith(this.error)
+      })
     })
   })
 })
--- a/test/unit/apps/events/transformers.test.js
+++ b/test/unit/apps/events/transformers.test.js
@@ -1,11 +1,14 @@
+/* eslint camelcase: 0 */
+
 const {
   transformEventToListItem,
+  transformEventResponseToViewRecord,
 } = require('~/src/apps/events/transformers')
 
 describe('Event transformers', function () {
-  describe('#transformEventToListItem', () => {
-    const mockEvent = require('~/test/unit/data/events/event-data')
+  const mockEvent = require('~/test/unit/data/events/event-data')
 
+  describe('#transformEventToListItem', () => {
     it('should return undefined when no arguments', () => {
       expect(transformEventToListItem({})).to.be.undefined
     })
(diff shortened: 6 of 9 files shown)


src/apps/events/repos.js
```diff
@@ -9,6 +9,11 @@ function createEvent (token, body) {
   })
 }
 
+function getEvent (token, id) {
```
tyom: Perhaps `fetchEvent` would be more appropriate to communicate that it's a network operation.

ztolley: agreed

src/apps/events/transformers.js
```diff
@@ -85,6 +89,42 @@ function transformEventToListItem ({
   return item
 }
 
+/**
+ * Transform a raw event into a display format
```
tyom: I feel like these comments are unnecessary. We should highlight these concepts in README instead of adding comments in code. The problem with comments in code is that they become outdated quick, can be misleading and confusing. The code should be written in such way as to explain its purpose.

src/apps/events/transformers.js
```diff
@@ -85,6 +89,42 @@ function transformEventToListItem ({
...
+
+  if (event.start_date === event.end_date) {
+    transformedEvent.date = formatLongDate(event.start_date) || ''
+  } else {
+    transformedEvent.start_date = formatLongDate(event.start_date) || ''
+    transformedEvent.end_date = formatLongDate(event.end_date) || ''
+  }
+
+  return Object.assign(transformedEvent, {
+    location_type: getPropertyName(event, 'location_type', ''),
```
tyom: Use Lodash `get`.

src/apps/events/transformers.js
```diff
@@ -85,6 +89,42 @@ function transformEventToListItem ({
...
+    address: getFormattedAddress(event),
+    uk_region: getPropertyName(event, 'uk_region', ''),
+    notes: event.notes,
+    lead_team: getPropertyName(event, 'lead_team', ''),
+    organiser: getPropertyName(event, 'organiser', ''),
+    other_teams: event.teams
```
tyom: What if `teams` is not an array?

src/lib/property-helpers.js
```diff
@@ -35,13 +35,14 @@ function deleteNulls (data) {
  *
  * @param {Object} object
  * @param {string} key
+ * @param {*} defaultValue
  * @returns {string|null} the name for the object
  */
-function getPropertyName (object, key) {
+function getPropertyName (object, key, defaultValue = null) {
```
teneightfive: This looks a lot like Lodash's [get](https://lodash.com/docs/4.17.4#get) method. Maybe consider swapping it for that.

src/apps/events/transformers.js
ztolley: it always is, even when nothing has been provided. I checked with @reupen

src/apps/events/transformers.js
```diff
@@ -85,6 +88,41 @@ function transformEventToListItem ({
...
+    notes: event.notes,
+    lead_team: get(event.lead_team, 'name', ''),
+    organiser: get(event.organiser, 'name', ''),
+    other_teams: Array.isArray(event.teams)
+      ? event.teams
+        .filter(team => team.id !== event.lead_team.id)
+        .map(item => item.name)
+        .join(', ')
+      : '',
+    related_programmes: Array.isArray(event.teams)
```
yeahfro: This doesn't look quite right `event.teams`

test/unit/data/events/minimal-event.json
```diff
@@ -0,0 +1,29 @@
+{
```
yeahfro: We have multiple event data files doing very similar things:
- `event.json`
- `event-data.json`
- `minimal-event.json`

I think this will be confusing in the future. Can we consolidate? I was planning to get right of my `event.json`

test/unit/apps/events/transformers.test.js
```diff
@@ -52,4 +57,135 @@ describe('Event transformers', function () {
...
+        uk_region: 'FDI Hub',
+        service: 'Account Management',
+      }
+
+      const transformedEvent = transformEventToDisplayEvent(this.mockEvent)
+
+      expect(transformedEvent).to.deep.equal(expected)
+    })
+
+    it('should handle when there is only an end date', () => {
```
yeahfro: I am becoming a fan of using the mocha `Context()` .. just a thought for the future

test/unit/data/events/minimal-event.json
ztolley: I could rename event-data.json something like full-event.json

test/unit/apps/events/transformers.test.js
teneightfive: `context()` is an alias for `describe()` but it adds an extra layer of readability. I've really started to prefer that way of writing unit tests.

Basically, any time you find yourself using the word `if` in an `it` block, see if you can wrap the test in a `describe()` or `context()` and use that context as the description for the context.

test/unit/apps/events/controllers/details.test.js
```diff
@@ -1,14 +1,37 @@
...
+      },
+    }
 
-    this.req = { }
     this.res = {
       breadcrumb: this.sandbox.stub().returnsThis(),
       render: this.sandbox.spy(),
     }
+
+    this.next = this.sandbox.stub()
```
teneightfive: You can probably use a `spy()` rather than a `stub()` here as you're not stubbing what the function returns.

Also, I find using names like `this.nextSpy` to signify what a spy/stub is doing helps when reading individual tests.

test/unit/apps/events/transformers.test.js
ztolley: And I'll spend some time looking into it as an alternative, though I'd like to merge this and then if I do switch to that syntax I can do a follow up pr

test/unit/apps/events/controllers/details.test.js
```diff
@@ -17,25 +40,41 @@ describe('Event details controller', () => {
...
       expect(this.res.render).to.be.calledWith('events/views/details')
       expect(this.res.render).to.have.been.calledOnce
     })
 
-    it('should render the event details page with a title', async () => {
+    it('should call the repo to get data', async () => {
+      await this.controller.renderDetailsPage(this.req, this.res, this.next)
+      expect(this.fetchEventStub).to.be.calledWith(this.req.session.token, this.req.params.id)
+
+      const params = this.res.render.firstCall.args[1]
```
teneightfive: This feels like it could be a separate `it` block that describes this assertion.

test/unit/apps/events/transformers.test.js
ztolley: If you have tests you have written in this style, point me at them so I can make sure I'm being consistent.

test/unit/apps/events/controllers/details.test.js
```diff
@@ -17,25 +40,41 @@ describe('Event details controller', () => {
...
+      expect(params.event).to.deep.equal(eventData)
+    })
+
+    it('should transform data and send it to the view with labels', async () => {
```
teneightfive: Rather than using `and` in it description use another it block to describe them separately.

test/unit/apps/events/transformers.test.js
```diff
@@ -1,10 +1,15 @@
...
   transformEventToListItem,
+  transformEventToDisplayEvent,
 } = require('~/src/apps/events/transformers')
 
+const eventData = require('~/test/unit/data/events/event-data')
+
 describe('Event transformers', function () {
   describe('#transformEventToListItem', () => {
-    const mockEvent = require('~/test/unit/data/events/event-data')
+    const mockEvent = cloneDeep(eventData)
```
teneightfive: What's the purpose of needing to clone this?

We should be avoiding mutating things within functions so if its being passed to a transformer it shouldn't be mutating it.

test/unit/apps/events/transformers.test.js
```diff
@@ -52,4 +57,135 @@ describe('Event transformers', function () {
...
+        id: '3123123123123123',
+        name: 'The dog foundation',
+      })
+
+      const transformedEvent = transformEventToDisplayEvent(this.mockEvent)
+
+      expect(transformedEvent.other_teams).to.equal('The dog foundation')
+    })
+
+    it('should include other team details when there are multiple other teams', () => {
```
teneightfive: As @yeahfro mentioned, it's nicer to use a `context()` block to wrap the current state and keep `it` descriptions short and specific.

src/apps/events/repos.js
```diff
@@ -9,6 +9,11 @@ function createEvent (token, body) {
   })
 }
 
+function fetchEvent (token, id) {
```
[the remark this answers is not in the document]
ztolley: @tyom requested we change it to fetchEvent

src/apps/events/transformers.js
```diff
@@ -85,6 +88,41 @@ function transformEventToListItem ({
   return item
 }
 
+function transformEventToDisplayEvent (event) {
```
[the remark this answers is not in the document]
ztolley: Not sure what you mean, and we do use this in other places. It accepts an event and transforms it.

src/apps/events/transformers.js
```diff
@@ -85,6 +88,41 @@ function transformEventToListItem ({
...
+  const transformedEvent = {
+    event_type: get(event.event_type, 'name', ''),
```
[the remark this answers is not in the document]
ztolley: I put a default of '' so we get consistent results when there is no property. Some helpers like date return '', this would return undefined, something else would return something else. It keeps it consistent and easier to test.

src/apps/events/transformers.js
```diff
@@ -85,6 +88,41 @@ function transformEventToListItem ({
...
+  }
+
+  if (event.start_date === event.end_date) {
+    transformedEvent.date = formatLongDate(event.start_date) || ''
```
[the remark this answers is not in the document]
ztolley: The display is done using the key/value display macros so it has to be done here

src/apps/events/transformers.js
```diff
@@ -85,6 +88,41 @@ function transformEventToListItem ({
...
+  } else {
+    transformedEvent.start_date = formatLongDate(event.start_date) || ''
+    transformedEvent.end_date = formatLongDate(event.end_date) || ''
+  }
+
+  return Object.assign(transformedEvent, {
```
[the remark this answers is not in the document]
ztolley: Yup, and that's fine as that's what I want to do, I don't need to copy it to make a new thing to avoid side effects.

test/unit/apps/events/transformers.test.js
teneightfive: [Example in OMIS middleware](https://github.com/uktrade/data-hub-frontend/blob/develop/test/unit/apps/omis/middleware.test.js#L52-L102)

src/apps/events/repos.js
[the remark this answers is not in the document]
teneightfive: Why? Doesn't seem like we've used the term `fetch` anywhere else in the app when returning things from the API. We've always used `getThing()`.

src/apps/events/transformers.js
```diff
@@ -85,6 +88,41 @@ function transformEventToListItem ({
...
+    address: getFormattedAddress(event),
+    uk_region: get(event.uk_region, 'name', ''),
+    notes: event.notes,
+    lead_team: get(event.lead_team, 'name', ''),
+    organiser: get(event.organiser, 'name', ''),
+    other_teams: Array.isArray(event.teams)
+      ? event.teams
+        .filter(team => team.id !== event.lead_team.id)
+        .map(item => item.name)
+        .join(', ')
```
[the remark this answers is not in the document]
ztolley: As mentioned, the view component used to display this doesn't allow that.

src/apps/events/views/details.njk
```diff
@@ -1,7 +1,6 @@
 {% extends "_layouts/two-column.njk" %}
 
-{% block main_grid_left_column %}{% endblock %}
-
 {% block main_grid_right_column %}
-
+  {{ keyvaluetable(displayEventData, stripey=true, labels=displayEventLabels) }}
```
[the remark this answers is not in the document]
ztolley: I will take a look

test/unit/apps/events/transformers.test.js
ztolley: For some tests I change some of the data in the test data, without cloning it meant one test was breaking another.

test/unit/apps/events/controllers/details.test.js
ztolley: just copying what we've done in other tests

src/apps/events/transformers.js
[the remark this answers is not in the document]
teneightfive: Look at the arguments in the method above. It doesn't just accept a single object as the argument, it lays out the properties of that argument that it expects using destructuring.

Something more like this:

```js
const { assign, get, isArray } = require('lodash')

function transformEventToDisplayEvent ({
  event_type,
  end_date,
  start_date,
  location_type,
  uk_region,
  notes,
  lead_team,
  organiser,
  teams,
  notes,
  related_programmes,
  service,
}) {
  const event = {
    event_type: get(event_type, 'name'),
  }

  if (start_date === end_date) {
    transformedEvent.date = formatLongDate(start_date)
  } else {
    transformedEvent.start_date = formatLongDate(start_date)
    transformedEvent.end_date = formatLongDate(end_date)
  }

  return assign(event, {
    notes,
    location_type: get(location_type, 'name'),
    address: getFormattedAddress(event),
    uk_region: get(uk_region, 'name'),
    lead_team: get(lead_team, 'name'),
    organiser: get(organiser, 'name'),
    other_teams: isArray(teams)
      ? teams
        .filter(team => team.id !== lead_team.id)
        .map(item => item.name)
        .join(', ')
      : null,
    related_programmes: isArray(related_programmes)
      ? related_programmes
        .map(item => item.name)
        .join(', ')
      : null,
    service: get(service, 'name'),
  })
}
```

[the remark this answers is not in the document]
teneightfive: I'm still concerned about using the old format of macros that requires us to transform markup and such in places like this.

Might be better to change the macro we use that renders key/value table to support an array of items with optional `type` property to handle its display for things like dates/address.

Then we can remove the need for `getFormattedAddress()` and `formatLongDate()` from the generic _lib_ and the transformation is done using view filters in nunjucks.

src/apps/events/transformers.js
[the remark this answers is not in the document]
tyom: I think normally `get` is used like this: `get(source, 'path.within.source', default)`. I also think source should be just `event`.

src/apps/events/transformers.js
```diff
@@ -85,6 +88,41 @@ function transformEventToListItem ({
...
+
+  return Object.assign(transformedEvent, {
+    location_type: get(event.location_type, 'name', ''),
+    address: getFormattedAddress(event),
+    uk_region: get(event.uk_region, 'name', ''),
+    notes: event.notes,
+    lead_team: get(event.lead_team, 'name', ''),
+    organiser: get(event.organiser, 'name', ''),
+    other_teams: Array.isArray(event.teams)
+      ? event.teams
```
tyom: Can this be done above `return` to avoid ternary and make it easier to read?

tyom: Also this can be done like this:
```
get(event, 'teams', [])
  .filter(team => team.id !== event.lead_team.id)
  .map(item => item.name)
  .join(', ')
```

src/apps/events/transformers.js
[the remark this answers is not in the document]
teneightfive: I'd be inclined to not set everything to an empty string as default. If its to work with values in the view the view layer should know what to do if the value is defined rather than having to check for an empty string.

src/apps/events/transformers.js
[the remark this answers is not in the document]
teneightfive: As mentioned in slack. We are phasing out old macros so we should look at adapting new macros to suit the use cases we have rather than adding more deprecated code that we will have to remove later.

src/apps/events/transformers.js
ztolley: The reason it's in the return like this is order is important but I'll find a way to make it nicer to read. Can't remember if the macros allow you to pick a display order if lodash `pick` could extract them in the desired order. None of the solutions it particularly pleasant.

src/apps/events/repos.js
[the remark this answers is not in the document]
tyom: I suggested it because `fetch` implies network operation. We use `get` to get property or resource. Perhaps as the new convention we should have `fetch` prefix for all network operations.

src/apps/events/transformers.js
ztolley: using the expanded parameters method means I can't use this 

`function x = ({ param, param,`

src/apps/events/controllers/details.js
```diff
@@ -1,18 +1,16 @@
 const { fetchEvent } = require('../repos')
 const { transformEventToDisplayEvent } = require('../transformers')
-const { displayEventLabels } = require('../labels')
 
 async function renderDetailsPage (req, res, next) {
   try {
     const event = await fetchEvent(req.session.token, req.params.id)
-    const displayEventData = transformEventToDisplayEvent(event)
+    const displayEvent = transformEventToDisplayEvent(event)
```
tyom: Should this variable be a noun? It returns object. Maybe `eventData` is more appropriate?

src/apps/events/transformers.js
```diff
@@ -88,37 +88,64 @@ function transformEventToListItem ({
   return item
 }
 
-function transformEventToDisplayEvent (event) {
+function transformEventToDisplayEvent ({
```
tyom: In other places we're trying to use a pattern for naming transformers. `transformSomethingToSomething`
This could be called something like `transformEventResponseToViewRecord` to make it obvious what it does.

src/apps/events/transformers.js
```diff
@@ -85,6 +88,68 @@ function transformEventToListItem ({
...
+  uk_region,
+  notes,
+  lead_team,
+  organiser,
+  teams,
+  related_programmes,
+  service,
+}) {
+  if (!teams) {
+    teams = []
```
tyom: Just initialise the parameters above.
```js
{
  // …
  teams = [],
  related_programmes = [],
  // …
}
```

src/apps/events/repos.js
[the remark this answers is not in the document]
ztolley: There's been times that multiple files have getEntity and fetch helps clarify things. I'm sticking with fetch

src/apps/events/transformers.js
ztolley: Tried that but will try again

ztolley: Yup tried it, no joy

tyom: It only doesn't work if value is set to `null`. Does API give array or `null`? If so, should it?

If you just want to ensure it's array (not falsey) you can do so with one-liners:
```js
  teams = teams || []
  related_programmes = related_programmes || []
```

src/templates/_components/key-value-table.njk
```diff
@@ -27,6 +27,12 @@
                 <li>{{ item }}{{ "," if not loop.last }}</li>
                {% endfor %}
               </ul>
+            {% elif data.name %}
```
teneightfive: I would look at changing the structure of this component to accept an array of objects like the entity list does.

And each item could then have an optional `type` property which handles how it should be displayed and apply filters if necessary.

src/apps/events/controllers/details.js
```diff
@@ -1,9 +1,20 @@
...
+async function renderDetailsPage (req, res, next) {
+  try {
+    const event = await fetchEvent(req.session.token, req.params.id)
+    const displayEvent = transformEventToDisplayEvent(event)
+
+    res
+      .breadcrumb(event.name)
+      .render('events/views/details', {
+        id: req.params.id,
+        displayEvent,
```
teneightfive: It's a good pattern to put shorthand properties first. Makes it easier to scan as more get added.

src/apps/events/views/details.njk
```diff
@@ -1,7 +1,6 @@
 {% extends "_layouts/two-column.njk" %}
 
-{% block main_grid_left_column %}{% endblock %}
-
 {% block main_grid_right_column %}
-
+  {% component 'key-value-table', variant='striped', items=displayEvent %}
+  <a class="button button-secondary" href="{{id}}/edit">Edit Event</a>
```
teneightfive: Spacing around `{{ id }}`.

src/apps/events/views/details.njk
```diff
@@ -1,7 +1,6 @@
 {% extends "_layouts/two-column.njk" %}
 
-{% block main_grid_left_column %}{% endblock %}
-
 {% block main_grid_right_column %}
-
+  {% component 'key-value-table', variant='striped', items=displayEvent %}
```
teneightfive: As @tyom mentioned, this variable name isn't very meaningful in this case. `items` = `displayEvent`?

test/unit/apps/events/controllers/details.test.js
```diff
@@ -1,41 +1,85 @@
...
 
   afterEach(() => {
     this.sandbox.restore()
   })
 
-  describe('#renderDetailsPage', () => {
-    it('should add a breadcrumb', async () => {
-      this.controller.renderDetailsPage(this.req, this.res)
+  describe('#renderDetailsPage', async () => {
+    context('when a page request is made and there are no errors', () => {
```
teneightfive: Do you need to say `when a page request is made`? It's also duplicated in the context below so if you do need it, wrap both of them in another `context()` with this as the description.

When you see duplication, that's when it's good to refactor.

test/unit/apps/events/transformers.test.js
```diff
@@ -52,4 +55,290 @@ describe('Event transformers', function () {
...
+
+          this.transformedEvent = transformEventToDisplayEvent(eventWithNoProgrammes)
+        })
+
+        it('should indicate there are no related programmes', () => {
+          expect(this.transformedEvent['Related programmes']).to.deep.equal([])
+        })
+      })
+
+      context('when related programmes is null', () => {
```
teneightfive: This feels pretty smelly having to check if something is `null` and then setting it to an empty array.

ztolley: in reality, it will never be passed a null, it will always be passed an empty array, @tyom wanted to put in a check to make sure it was an array before mapping over in.

I could just remove the test

src/apps/events/transformers.js
ztolley: API always gives you [] when it's an array so this won't be used but will put int he one liners for now

src/apps/events/views/details.njk
ztolley: Suggest a name

src/apps/events/controllers/details.js
ztolley: eventData is too close to event. is it raw, is it formatted?

To go with the function name suggested I am using eventViewRecord just so it's easy to associate the transformation name to the data in the view

config/nunjucks/filters.js
```diff
@@ -115,6 +116,9 @@ const filters = {
   },
 
   formatDate: (value, format = longDateFormat) => {
+    if (!value) {
+      return
```
tyom: Should return `value`.

ztolley: value is null or undefined so why?

ztolley: Plus this is actually going to be removed once the other PR ha been merged.

tyom: It can be an empty string and you're converting it to `undefined`.

src/apps/events/controllers/details.js
```diff
@@ -1,9 +1,20 @@
...
+async function renderDetailsPage (req, res, next) {
+  try {
+    const event = await fetchEvent(req.session.token, req.params.id)
+    const eventViewRecord = transformEventResponseToViewRecord(event)
+
+    res
+      .breadcrumb(event.name)
+      .render('events/views/details', {
+        eventViewRecord,
+        id: req.params.id,
```
tyom: I'd suggest `eventId`.

test/unit/apps/events/transformers.test.js
```diff
@@ -1,11 +1,14 @@
...
+
 const {
   transformEventToListItem,
+  transformEventResponseToViewRecord,
 } = require('~/src/apps/events/transformers')
 
 describe('Event transformers', function () {
-  describe('#transformEventToListItem', () => {
-    const mockEvent = require('~/test/unit/data/events/event-data')
+  const mockEvent = require('~/test/unit/data/events/event-data')
```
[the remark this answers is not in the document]
ztolley: it is, but that was there before, I just moved it. I wanted to change as little as possible in the existing tests.

PR #647, opened 2017-09-21
Updated CSS changes

venureddy95:
Updated changed CSS, Changed verify to assert and Updated feature lines to match the correct ones

### DIFF
--- a/test/acceptance/features/interactions/create-a-investment-project.feature
+++ b/test/acceptance/features/interactions/create-a-investment-project.feature
@@ -17,69 +17,62 @@ Feature: Create a new Investment project
   Scenario: Create a new Investment project as a source of foreign equity investment
 
     When I create a new Investment project as a source of foreign equity investment
-    Then I see the Investment project created confirmation message
+    Then I see the "Investment project created" success message
     And I verify my newly created Investment project in company profile
     And I verify Type of Investment is shown as "FDI, Acquisition"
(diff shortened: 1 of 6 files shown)


test/acceptance/features/interactions/create-a-investment-project.feature
```diff
@@ -17,81 +17,74 @@ Feature: Create a new Investment project
   Scenario: Create a new Investment project as a source of foreign equity investment
 
     When I create a new Investment project as a source of foreign equity investment
-    Then I see the Investment project created confirmation message
+    Then I see the "Investment project created" success message
```
tyom: This means we'll have to update these features whenever we change the message text. I think a more generic "creation confirmation" message would be better.

venureddy95: @tyom : You mean, to display a generic success/creation message for all the posts in front end?

tyom: No @venureddy95, a generic message in tests so they are not coupled with implementation.

venureddy95: @tyom : That is fine. I can make it a generic message and assert only on the CSS for the Success flash message. This needs revisiting all the feature files across creating and editing. Would rather do it in a separate PR soon.

PR #649, opened 2017-09-21
renamed step definitions folders with underscore to recognize feature lines

venureddy95:
Feature files are not recognizing the step definitions as the folder name has dash instead of an underscore. So these changes....


tyom:
Can you expand on the issue of using underscores in folder names?

yeahfro:
Unfortunately WebStorm/JetBrains does not recognise step definitions unless they are in a step_definitions folder.

feedmypixel:
> @tyom @feedmypixel Unfortunately WebStorm/JetBrains does not recognise step definitions unless they are in a step_definitions folder. This allows the step definition to be clickable from the cucumber file.

yup confirmed, and you can't change it in `IntelliJ`

feedmypixel:
screen shot 2017-09-21 at 15 55 59
@tyom you don't get the `ctrl` click Jazz with the folder `-` name change

PR #656, opened 2017-09-21
Refactor company edit views

teneightfive:
This change reduces the duplicated views used to edit different types of companies.

The differences are relatively minimal and have been handled using `if/else` blocks in one shared view for now.

### Future changes (out of scope for this PR)

- Switch to new form building method and new form macros
- Reduce duplication in controllers and middleware


src/apps/companies/controllers/ltd.js
```diff
@@ -66,11 +50,15 @@ async function addDetails (req, res, next) {
     } else {
       res.locals.formData = companyFormService.getDefaultLtdFormForCH(res.locals.chCompany)
     }
+    res.locals.chDetailsLabels = chDetailsLabels
```
ztolley: move into own middleware as this is used in multiple plaves

src/apps/companies/middleware/form.js
```diff
@@ -0,0 +1,47 @@
...
+
+  next()
+}
+
+async function handleFormPost (req, res, next) {
+  try {
+    const savedCompany = await companyFormService.saveCompanyForm(req.session.token, req.body)
+
+    req.flash('success', 'Company record updated')
+    res.redirect(`/companies/view/foreign/${savedCompany.id}`)
```
ztolley: carries over the old bug to redirect to foreign. After this, we merge the company details view too

teneightfive: Yes, was aware of this. Hadn't finished with this method yet. Was planning to fix it with a follow up commit but it was getting late last night...

src/apps/companies/controllers/ltd.js
teneightfive: Yeah, haven't quite nailed all of this yet and was kind of a temporary measure until I found a better place for it.

teneightfive: Actually, I'm not sure what you mean here. Do you mean the labels are used in multiple views or that this middleware is used in multiple other places?

src/apps/companies/views/edit.njk
```diff
@@ -0,0 +1,147 @@
+{% extends "./_layout-edit.njk" %}
+
+{% block main_grid_right_column %}
+  {{ super() }}
```
ztolley: Do we need super?

teneightfive: Nope. Good spot.

PR #659, opened 2017-09-22
Removes the use of getFullQuote for OMIS orders

teneightfive:
The API now returns the full quote regardless so don't need to pass
an extra query string.


marcofucci:
might be worth removing the `getFullQuote` from `models.js` as well.

PR #663, opened 2017-09-22
Company edit refactor

teneightfive:
This work tries to remove more duplication from the controllers and middleware used to add and edit companies.

This work is still not complete and needs more refinement to bring the view in line with newer form macros.


src/apps/companies/middleware/params.js
```diff
@@ -0,0 +1,24 @@
+const { getDitCompany, getCHCompany } = require('../repos')
+
+async function getCompany (req, res, next, id) {
```
ztolley: There's already a middleware named getCompanyDetails that is doing this and extracting the company name and address for the heading but it is only designed to work for non ch and only being used for interactions. We will need to remove that from the router (line 77) and replace it with something that is called after this to calculate the heading title and address only.

tyom:
[Companies House URL](http://localhost:3000/companies/view/ch/05588682) seems to be broken.

teneightfive:
>Companies House URL seems to be broken.

I did spot that too. It's being addressed in a [follow up PR](https://github.com/uktrade/data-hub-frontend/pull/679) to refactor companies view routes.

src/apps/companies/router.js
```diff
@@ -20,11 +20,16 @@ const {
...
 const { getRequestBody, getCompanyCollection } = require('./middleware/collection')
 const { populateForm, handleFormPost } = require('./middleware/form')
+const { getCompany, getCompaniesHouseRecord } = require('./middleware/params')
 const { setDefaultQuery } = require('../middleware')
 
 const DEFAULT_COLLECTION_QUERY = {
   sortby: 'modified_on:desc',
 }
+
+router.param('id', getCompany)
```
tyom: `getCompanyRecord` to match the line below?

src/apps/companies/transformers.js
```diff
@@ -103,7 +105,43 @@ function transformCompaniesHouseCompanyToListItem (props) {
   })
 }
 
+function transformCompanyFromApi (body) {
```
tyom: I don't think this belongs here in this form.

PR #665, opened 2017-09-22
Collections in add company flow

tyom:
Use embedded collection for add company flow search results.

### Before
localhost_3001_companies_add-step-2_business_type ltd term ltd

### After
localhost_3001_companies_add-step-2_business_type ltd term ltd 1

### DIFF
--- a/assets/stylesheets/components/_entity.scss
+++ b/assets/stylesheets/components/_entity.scss
@@ -11,6 +11,10 @@
       text-decoration: underline;
     }
   }
+
+  @include media(tablet) {
+    max-width: 80%;
+  }
 }
 
 @include media(tablet) {
--- a/src/apps/companies/controllers/add.js
+++ b/src/apps/companies/controllers/add.js
@@ -1,4 +1,5 @@
 const { map } = require('asyncro')
+const { assign } = require('lodash')
 
 const logger = require('../../../../config/logger')
 const { ukOtherCompanyOptions, foreignOtherCompanyOptions } = require('../options')
@@ -6,7 +7,8 @@ const { getCHCompany } = require('../repos')
 const { buildQueryString } = require('../../../lib/url-helpers')
 const { isBlank } = require('../../../lib/controller-utils')
 const { searchLimitedCompanies } = require('../../search/services')
-const { buildPagination } = require('../../../lib/pagination')
+const { transformApiResponseToSearchCollection } = require('../../search/transformers')
+const { transformCompaniesHouseCompanyToListItem } = require('../../companies/transformers')
 const { companyDetailsLabels, companyTypeOptions } = require('../labels')
 
 function getAddStepOne (req, res, next) {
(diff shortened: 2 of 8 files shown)


assets/stylesheets/components/_entity.scss
```diff
@@ -11,6 +11,10 @@
       text-decoration: underline;
     }
   }
+
+  @include media(tablet) {
+    max-width: 80%;
```
teneightfive: Do we want a max-width on it or should this be determined by the container?

tyom: This is an attempt to avoid strange float breaks by wrapping earlier.

ztolley: What was the strange behaviour and how can I reproduce it?

tyom: Nothing major, just title going across the whole row and in places leaving [widow](https://www.fonts.com/content/learning/fontology/level-2/text-typography/rags-widows-orphans) which doesn't look great. Although come to think of it there will be cases when this specific width restriction will cause that. Unfortunately on the web it's harder to control text flow.

Still I think restricting the full width of the title on bigger screens makes improves readability and compacts longer titles.

src/apps/companies/controllers/add.js
```diff
@@ -88,33 +89,49 @@ async function getAddStepTwo (req, res, next) {
...
-      results,
-      pagination: buildPagination(req.query, results),
-    }
+      })
+      .then(
+        transformApiResponseToSearchCollection(
+          { query: req.query },
+          transformCompaniesHouseCompanyToListItem,
+          (item) => {
+            return Object.assign({}, item, {
```
teneightfive: Should we start using lodash's `assign`?

src/apps/companies/transformers.js
```diff
@@ -74,6 +74,37 @@ function transformCompanyToListItem ({
   }
 }
 
+function transformCompaniesHouseCompanyToListItem (props) {
+  console.log(props.companies_house_data)
```
teneightfive: 🚨

src/apps/companies/transformers.js
```diff
@@ -74,6 +74,37 @@ function transformCompanyToListItem ({
   }
 }
 
+function transformCompaniesHouseCompanyToListItem (props) {
```
teneightfive: Haven't we previously been passing a destructured object in as the argument rather than just a single name which we reference properties from? Makes it easier to see what properties are expected for this transformation.

tyom: We have indeed. But in this case it's slightly different as we're extending `transformCompanyToListItem` inside `transformCompaniesHouseCompanyToListItem`. So it's easier to just pass the `props` and then reset `meta` to Companies House specific than redo everything `transformCompanyToListItem` does by listing out individual props.

PR #666, opened 2017-09-22
Collection acceptance test patterns

feedmypixel:
A test suite for `events-list` and a pattern for `lists/collections` that can be used across all  `lists/collections` tests.

## Tests

- `@events-list`
  - `@events-list-add`
  - `@events-list-view`
  - `@events-list-filter`
  - `@events-list-sort-a-z`
  - `@events-list-sort-z-a`

## Xpath for choosing siblings
A pretty awesome xpath (if I do say so) for picking out sibling [metadata values](https://github.com/uktrade/data-hub-frontend/pull/666/files#diff-88e8584036bdd16773e333c38feaab40R18) of named meta data elements.

## Sections
Demonstrates how to use `sections` in page objects.

## State
Introduces state objects on pageObjects to record information between steps. Also adds `Before` and `After` blocks to setup and tear down these objects.

## Custom commands
Brings in custom `commands`, specifically `getListOption.js` and `wait.js`

## Naming conventions
Starts to form some naming conventions. IMO feature tags should use the folder names at the start this will enable us to use the `--name` flag with regexes. So all features in `/events` should be `@events-<description>` and then ran with the flag `--name ^events-*`

## Pausing/Waiting
I've found a few use cases of needing to pause tests, waiting for xhr, waiting for ES to sync so I have added a global variable `pauseDuration` and a convenience command `wait`. I feel it would be good practice to put why the `wait` is needed in a comment next to it.

## Screenshot
screen shot 2017-09-24 at 11 25 31


feedmypixel:
Base branch https://github.com/uktrade/data-hub-frontend/pull/667

tyom:
Regarding naming conventions, I was wondering if it would make it easier to scan, predict and run tags for specific scenario. Since we're using kebab case to separate words and feature tags are often multi-word should we use underscore or double hyphen or else to separate scenarios from features?

feedmypixel:
> Regarding naming conventions, I was wondering if it would make it easier to scan, predict and run tags for specific scenario. Since we're using kebab case to separate words and feature tags are often multi-word should we use underscore or double hyphen or else to separate scenarios from features?

yeah lets catch up next week - I liked your `BEM` thoughts on this

ztolley:
The details view for even has a few new fields to add, including services and region.

Still getting my head around this as it's clever but a lot to take in.

All Great stuff. My question is, how can we make it generic and reusable across all entity types, so instead of saying check the fields specific to an event collection list we test a collection list and pass a list of the field we expect to see?

test/acceptance/features/events/page-objects/Events.js
```diff
@@ -70,6 +72,63 @@ module.exports = {
         return this
           .assert.visible('#group-field-related_programmes #group-field-related_programmes:nth-child(' + listNumber + ') select')
       },
+      populateCreateEventForm () {
+        const today = new Date()
+        const futureDate = addWeeks(today, 1)
+
+        // Store created event details
+        this.state.eventDetails = {
```
ztolley: Need to add service and uk region

test/acceptance/features/events/page-objects/EventsList.js
```diff
@@ -0,0 +1,54 @@
...
+        },
+        eventEnd: {
+          selector: '//span[contains(@class,"c-meta-list__item-label")][text()[normalize-space()="Ends"]]/following-sibling::span',
+          locateStrategy: 'xpath',
+        },
+        organiser: {
+          selector: '//span[contains(@class,"c-meta-list__item-label")][text()[normalize-space()="Organiser"]]/following-sibling::span',
+          locateStrategy: 'xpath',
+        },
+        leadTeam: {
```
ztolley: Need to add service and uk region

test/acceptance/features/events/page-objects/Events.js
feedmypixel: yup this wasn't in at the weekend - it appeared yesterday so will add in to tests

test/acceptance/features/events/page-objects/EventsList.js
feedmypixel: yup this wasn't in at the weekend - it appeared yesterday so will add in to tests

ztolley: Yeah, I know. I was going to write the tests when I did the card but then noticed you already had done the hard work 🥇

PR #668, opened 2017-09-22
Add support to key value table for dates and properties with name

ztolley:
The key value component can show text and links, this pr adds 2 features to that:

- Support date 'type'
- Pass properties that contain a name and display that name

This is really usefully for showing entity details without having to put presentation logic or a lot of boilerplate in the transformer.

### DIFF
--- a/config/nunjucks/filters.js
+++ b/config/nunjucks/filters.js
@@ -62,6 +62,7 @@ const filters = {
   isArray,
   isFunction,
   isNull,
+  isPlainObject,
   isString,
   pluralise,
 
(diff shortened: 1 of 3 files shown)


config/nunjucks/filters.js
```diff
@@ -115,14 +116,20 @@ const filters = {
   },
 
   formatDate: (value, format = longDateFormat) => {
+    if (!value) {
+      return null
```
tyom: I think that this being a filter should leave `value` intact, i.e. return `value` so if it's an empty string at remains it.

PR #670, opened 2017-09-23
Refactor service deliveries to use new forms

tyom:


src/apps/service-deliveries/controllers/edit.js
```diff
@@ -0,0 +1,57 @@
...
+        }),
+        recordAndFormState,
+        res.locals.formErrors,
+      )
+  } catch (error) {
+    next(error)
+  }
+
+  res
+    .breadcrumb(serviceDelivery ? `${get(serviceDelivery, 'subject')}` : null)
```
feedmypixel: can you not use the default in `get` ?

src/apps/service-deliveries/controllers/edit.js
```diff
@@ -0,0 +1,57 @@
...
+    .breadcrumb(`${serviceDelivery ? 'Edit' : 'Create'} service delivery`)
+    .render('service-deliveries/views/edit', {
+      serviceDeliveryForm: assign(serviceDeliveryForm, {
+        returnLink,
+        errors: {
+          summary: get(res.locals, 'formErrors.[0].detail'),
```
feedmypixel: does this need a `get`?

src/apps/service-deliveries/transformers.js
```diff
@@ -0,0 +1,43 @@
...
+  contact,
+  date,
+  event,
+  status,
+  dit_adviser,
+  service,
+  uk_region,
+  sector,
+  country_of_interest,
+} = {}) {
```
feedmypixel: is this default needed?

src/apps/service-deliveries/controllers/edit.js
tyom: Good shout.

src/apps/service-deliveries/controllers/edit.js
tyom: This is V2 error summary. I don't understand it well enough to trust the structure. Also I think it's deprecated so I decided use `get`.

src/apps/service-deliveries/transformers.js
tyom: We do in this case as we use the same transformer when creating new form in which `serviceDelivery` object doesn't exist. So we either handle it in the controller of just initialise with empty object.

src/apps/service-deliveries/controllers/edit.js
ztolley: The v2 error structure has bugs, hence handling code that says things like errors.errors.errors

ztolley:
As mentioned on other PR, Interactions and Service deliveries are about to change

tyom:
> As mentioned on other PR, Interactions and Service deliveries are about to change

And it'll be easier to change them without having to worry about the deprecated code.

src/apps/service-deliveries/controllers/details.js
```diff
@@ -0,0 +1,32 @@
...
+  'dit_adviser',
+  'uk_region',
+  'sector',
+  'contact',
+  'country_of_interest',
+]
+
+function renderDetailsPage (req, res) {
+  res
+    .breadcrumb(`${res.locals.serviceDelivery.subject}`)
```
ztolley: why use string interpolation when you don't need to?

src/apps/service-deliveries/controllers/edit.js
```diff
@@ -0,0 +1,57 @@
+const { assign, get } = require('lodash')
```
ztolley: You've imported assign and used it once and used Object.assign as well. Why not stick with one or the other to keep in consistent in this file.

src/apps/service-deliveries/macros.js
```diff
@@ -0,0 +1,84 @@
+const { assign, get } = require('lodash')
+const { globalFields } = require('../macros')
+const formLabels = require('./labels')
+const metadata = require('../../lib/metadata')
+const { transformContactToOption, transformObjectToOption } = require('../transformers')
+
+const serviceDeliveryEditFormConfig = function ({
+  services = [],
+  contacts = [],
+  advisers = [],
```
ztolley: Have you tested this works? I tried it in my PR and didn't get the results I expected (I think it always set the value to [] so I had to result to those one liners in the function

src/apps/service-deliveries/mixins.js
```diff
@@ -0,0 +1,32 @@
+const { saveServiceDelivery } = require('./repos')
+const { convertServiceDeliveryFormToApiFormat } = require('./services/data')
+const { getHydratedServiceDelivery } = require('./services/data')
+
+async function fetchDetails (req, res, next) {
+  try {
+    res.locals.serviceDelivery = await getHydratedServiceDelivery(req.session.token, req.params.serviceDeliveryId)
+    next()
```
ztolley: if there is an error in whatever is in next, so a function later down the chain, it can sometimes get caught by this catch instead. Moving the next outside of this and returning next(error) can fix that.

src/apps/service-deliveries/controllers/details.js
tyom: Good point. Leftovers from previous use with a string.

src/apps/service-deliveries/controllers/edit.js
tyom: I'll use `assign`.

src/apps/service-deliveries/macros.js
tyom: I'm not sure why it didn't work in your PR. This is a [default value assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Default_values_2) and it will only work when value is `undefined`, not `null` or another value.

ztolley: My issue was they were null, could that happen here?

src/apps/service-deliveries/mixins.js
tyom: Hmm.. I tried throwing in the middleware further down the line and it behaved correctly. Just curious, do you have a specific example?

ztolley: If the next step was a controller, and an error is thrown outside of that controllers own try catch then this one would catch it.

It's an unlikely scenario, and I think I only ever came across it in unit tests, but it could happen. @feedmypixel also mentioned it in a recent PR.

src/apps/service-deliveries/macros.js
tyom: That's a fair point. I've added additional protection against unexpected and a few test cases with it.

src/apps/service-deliveries/controllers/details.js
```diff
@@ -0,0 +1,32 @@
...
+  'uk_region',
+  'sector',
+  'contact',
+  'country_of_interest',
+]
+
+function renderDetailsPage (req, res) {
+  res
+    .breadcrumb(res.locals.serviceDelivery.subject)
+    .title('Service delivery details')
```
teneightfive: Should title not match breadcrumb like elsewhere?

src/apps/service-deliveries/middleware.js
```diff
@@ -0,0 +1,32 @@
+const { saveServiceDelivery } = require('./repos')
+const { convertServiceDeliveryFormToApiFormat } = require('./services/data')
+const { getHydratedServiceDelivery } = require('./services/data')
+
+async function fetchDetails (req, res, next) {
+  try {
+    res.locals.serviceDelivery = await getHydratedServiceDelivery(req.session.token, req.params.serviceDeliveryId)
```
teneightfive: You can pass `serviceDeliveryId` in as the fourth parameter in [express param middleware](http://expressjs.com/en/api.html#router.param).

Also, are we using a new pattern for naming these methods now? We've used `get` elsewhere.

src/apps/service-deliveries/middleware.js
```diff
@@ -0,0 +1,32 @@
...
+    next()
+  } catch (error) {
+    next(error)
+  }
+}
+
+async function postDetails (req, res, next) {
+  try {
+    const deliveryToSave = await convertServiceDeliveryFormToApiFormat(req.body)
```
teneightfive: Just had a look at this method. I don't think it needs `await` as its not async.

src/apps/service-deliveries/middleware.js
```diff
@@ -0,0 +1,32 @@
...
+    const result = await saveServiceDelivery(req.session.token, deliveryToSave)
+    res.redirect(`/service-deliveries/${result.data.id}`)
+  } catch (error) {
+    if (error.error) {
+      console.log(error.error)
```
teneightfive: Should you be logging errors just to the console? Either use `logger.error()` if we want the error logged but not to interrupt the request. Or use `next(error)`.

src/apps/service-deliveries/transformers.js
```diff
@@ -0,0 +1,43 @@
...
+  if (!id) { return null }
+
+  return {
+    subject,
+    notes,
+    date: {
+      day: format(date, 'DD'),
```
teneightfive: What if `date` is not valid?

src/apps/service-deliveries/views/edit.njk
```diff
@@ -1,138 +1,23 @@
...
-      {% set returnUrl = '/companies/' + QUERY.company + '/interactions' %}
-    {% else %}
-      {% set returnUrl = backUrl %}
-    {% endif %}
-
-    {{ save(returnUrl) }}
-  </form>
+  {% if serviceDelivery.company.name %}
+    {% set companyLink %}
+      <a href="{{companyUrl}}">{{ serviceDelivery.company.name }}</a>
```
teneightfive: Breathing space around `{{ companyUrl }}`

src/apps/service-deliveries/middleware.js
tyom: Nice.

tyom: I think "fetch" better communicates the network operation that's happening. Maybe we could review these later to standardise across all apps?

src/apps/service-deliveries/middleware.js
tyom: Good catch.

src/apps/service-deliveries/middleware.js
tyom: I'll remove it.

src/apps/service-deliveries/middleware.js
ztolley: Didn't we agree that over the weekend?

tyom: > Didn't we agree that over the weekend?

We did, but looking further into it, the `fetchDetails` is not appropriate name either as it's not just fetching but also setting the result onto locals. Hence the need for a global review later.

src/apps/service-deliveries/transformers.js
tyom: Fair point. Added checks.

PR #673, opened 2017-09-23
Refactor new interaction step 1

tyom:
Replace HTML form with new form builder and simplify logic in controller.

Also fix the absence of error summary when form is submitted with nothing selected.

### Before
localhost_3001_interactions_create_1_company 31d9b4dc-cc52-e411-985c-e4115bead28a 1

### After
localhost_3001_interactions_create_1_company 31d9b4dc-cc52-e411-985c-e4115bead28a

### DIFF
--- a/assets/stylesheets/components/form/_form-group.scss
+++ b/assets/stylesheets/components/form/_form-group.scss
@@ -117,6 +117,12 @@
   font-size: 16px;
 }
 
+.c-form-group--columns {
+  .c-form-group__inner {
+    column-count: 2;
+  }
+}
+
 @keyframes targetHighlight {
   0% {
     border-color: $black;
--- a/assets/stylesheets/components/form/_multiple-choice.scss
+++ b/assets/stylesheets/components/form/_multiple-choice.scss
@@ -4,6 +4,7 @@
 .c-multiple-choice {
   position: relative;
   clear: left;
+  transform: translateZ(0); // Enforcing GPU acceleration fixes overflow cropping caused by CSS columns
 }
 
 @mixin _scale-input($font-size) {
--- a/src/apps/interactions/controllers/details.js
+++ b/src/apps/interactions/controllers/details.js
@@ -1,28 +1,24 @@
-/* eslint camelcase: 0 */
 const { sortBy, pick, omit } = require('lodash')
 const queryString = require('query-string')
 
 const interactionLabels = require('../labels')
 const metadataRepository = require('../../../lib/metadata')
-const interactionDataService = require('../services/data')
 const { getDisplayInteraction } = require('../services/formatting')
 const { transformObjectToOption } = require('../../transformers')
 
-const interactonDisplayOrder = ['company', 'interaction_type', 'subject', 'notes', 'contact', 'date', 'dit_adviser', 'service', 'dit_team']
+const interactonDisplayOrder = [
+  'company',
+  'interaction_type',
+  'subject',
+  'notes',
+  'contact',
+  'date',
+  'dit_adviser',
+  'service',
+  'dit_team',
+]
 
-async function getCommon (req, res, next) {
-  try {
-    const token = req.session.token
-    if (req.params.interactionId && req.params.interactionId !== 'add') {
-      res.locals.interaction = await interactionDataService.getHydratedInteraction(token, req.params.interactionId)
-    }
-    next()
-  } catch (error) {
-    next(error)
-  }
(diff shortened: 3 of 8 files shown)


ztolley:
FYI I think these screens are due to change due to changes to service deliveries. Speak to Ian and Mills

tyom:
> FYI I think these screens are due to change due to changes to service deliveries. Speak to Ian and Mills

Sure. In the meantime I'm getting rid of old HTML forms so we can remove deprecated code.

assets/stylesheets/components/form/_multiple-choice.scss
```diff
@@ -4,6 +4,7 @@
 .c-multiple-choice {
   position: relative;
   clear: left;
+  transform: translateZ(0)
```
ztolley: out of curiosity, why?

assets/stylesheets/components/form/_form-group.scss
```diff
@@ -117,6 +117,12 @@
   font-size: 16px;
 }
 
+.c-form-group--columns {
+  .c-form-group__inner {
+    column-count: 2;
```
ztolley: neat

src/apps/interactions/controllers/details.js
```diff
@@ -20,57 +23,36 @@ async function getCommon (req, res, next) {
...
+      contactId: req.query.contact,
+      companyId: req.query.company,
+      interactionTypeOptions: sortBy(interactionTypeOptions, 'label'),
     })
 }
 
 function postAddStep1 (req, res) {
-  // error if no selection
   if (!req.body.interaction_type) {
     res.locals.errors = {
```
ztolley: Are there any new patterns for form post handling that supersede calling the get when a post fails? using a chain of middleware?

ztolley: Like some others could it not say if it is successful, redirect to the next step, otherwise continue onto render step 1 again.

assets/stylesheets/components/form/_multiple-choice.scss
tyom: I explained it in [commit message](https://github.com/uktrade/data-hub-frontend/pull/673/commits/51904034d1c34e2c61bf88a1319e3ac531786ec4).

ztolley: Maybe put a comment in the code to explain hacks/fixes as people won't look for the commit message that added it

src/apps/interactions/controllers/details.js
tyom: I don't think we have a good pattern for it yet. But we pass the errors into `buildFormWithStateAndErrors` along with data and form props to pass it down to macro.

We should explore some options on making it easier to handle forms in views. Should have a FED chat.

ztolley: But for this, rather than call the GET variation, which is what I did way back, wouldn't it be better to do as I describe, redirect if OK, otherwise pass errors to the next step, the controller that will render the step 1 form again?

src/apps/interactions/controllers/details.js
```diff
@@ -1,8 +1,11 @@
 /* eslint camelcase: 0 */
+const { sortBy, pick, omit } = require('lodash')
+const queryString = require('query-string')
 const interactionLabels = require('../labels')
```
teneightfive: I find it a nice pattern to separate external and internal deps with a new line. Helps break them up and keep a clear separation when people add new ones.

assets/stylesheets/components/form/_multiple-choice.scss
tyom: People should look in commit messages. If commits are granular they can give a lot more insight than comments in the code, which become outdated.

src/apps/interactions/controllers/details.js
tyom: I see what you mean. I wasn't going to dive too deep in this refactor in favour of doing multiple passes to make PRs smaller but I guess I can squeeze this in.

src/apps/interactions/controllers/details.js
```diff
@@ -1,79 +1,57 @@
-/* eslint camelcase: 0 */
+const { sortBy, pick, omit } = require('lodash')
+const queryString = require('query-string')
 const interactionLabels = require('../labels')
```
teneightfive: Same here with a new line to break external vs internal

PR #674, opened 2017-09-23
Refactor feedback page to use new form builder

tyom:
Align the structure with establishing patterns and get the benefit of improved error summary for free.

### Before
localhost_3001_support

### After
localhost_3001_support 2

### DIFF
--- a/src/apps/builders.js
+++ b/src/apps/builders.js
@@ -47,7 +47,7 @@ function buildFormWithErrors (form = {}, errorMessages = {}) {
   const messages = pick(pickBy(errorMessages), formFieldNames)
 
   const fieldLabels = form.children
-    .filter(x => Object.keys(errorMessages).includes(x.name))
+    .filter(x => Object.keys(messages).includes(x.name))
     .reduce((obj, item) => {
       obj[item.name] = item.label || item.name
       return obj
--- a/src/apps/support/middleware.js
+++ b/src/apps/support/middleware.js
@@ -1,92 +1,37 @@
-const { get, isEmpty } = require('lodash')
-const { title } = require('case')
-const Sniffr = require('sniffr')
+const { isEmpty, pickBy } = require('lodash')
 
-const config = require('../../../config')
 const logger = require('../../../config/logger')
-const { postToZenDesk } = require('./services')
+const { createZenDeskMessage, postToZenDesk } = require('./services')
 
-function populateFormData (req, res, next) {
-  const sniffr = new Sniffr()
-  sniffr.sniff(req.headers['user-agent'])
+async function postFeedback (req, res, next) {
+  const { title, feedbackType, email } = req.body
 
-  res.locals.form = Object.assign({}, res.locals.form, {
-    data: {
-      title: req.body.title,
-      feedbackType: req.body.feedbackType,
-      description: req.body.description,
-      email: req.body.email,
-      browser: req.body.browser || `${title(sniffr.browser.name)} ${sniffr.browser.versionString}, ${title(sniffr.os.name)} ${sniffr.os.versionString}`,
-    },
-    options: {
-      feedbackType: [
-        {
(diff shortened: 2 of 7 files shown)


src/apps/builders.js
```diff
@@ -47,7 +47,7 @@ function buildFormWithErrors (form = {}, errorMessages = {}) {
   const messages = pick(pickBy(errorMessages), formFieldNames)
 
   const fieldLabels = form.children
-    .filter(x => Object.keys(errorMessages).includes(x.name))
+    .filter(x => Object.keys(messages).includes(x.name))
```
ztolley: I know you didn't name it but something more descriptive than x maybe?

tyom: I think in cases like this `x` as the name of the variable is somewhat irrelevant. In functional programming `x` is [often used as a variable name](http://blog.ploeh.dk/2015/08/17/when-x-y-and-z-are-great-variable-names/). It's a single generic parameter, a thing that is being operated on.

ztolley: cant hurt to have a sensible name rather than x

tyom: Did you read that article I linked? The scope is so small here spelling out full name of what it is unnecessary. I'm all for good variable names but in this case I think `x` works well.

ztolley: Yup I read it and yes I'm seen examples that just use x. Not going to let it stop this, just I like to use names.

src/apps/support/middleware.js
```diff
@@ -1,100 +1,44 @@
...
-
-  res.locals.form = Object.assign({}, res.locals.form, {
-    errors,
+  res.locals.formErrors = Object.assign({}, res.locals.formErrors, {
+    messages,
   })
 
-  next()
-}
+  console.log(res.locals.formErrors)
```
teneightfive: 🚨

src/apps/support/middleware.js
```diff
@@ -1,100 +1,44 @@
...
 
-async function submitForm (req, res, next) {
-  const errorMessages = get(res.locals, 'form.errors.messages')
+  const hasErrors = !isEmpty(res.locals.formErrors.messages) || res.locals.formErrors.summary
```
teneightfive: Should we use lodash's `get` to find these deep properties?

src/apps/support/middleware.js
```diff
@@ -1,100 +1,44 @@
...
-      { id: config.zenBrowser, value: req.body.browser },
-      { id: config.zenService, value: 'datahub_export' },
-    ],
-    tags: [req.body.type],
-  }
+  const ticket = createZenDeskMessage(req.body)
 
   try {
     const response = await postToZenDesk(ticket)
     req.flash('success', `Created new report, reference number ${response.data.ticket.id}`)
```
teneightfive: Can you stick a todo in here to display this in a better fashion on the complete page.

Would be better to use a pattern like [confirmation pages](https://www.gov.uk/service-manual/design/confirmation-pages) to play the reference back to the user.

src/apps/support/services.js
```diff
@@ -2,6 +2,24 @@ const axios = require('axios')
...
 
+function createZenDeskMessage (body) {
+  return {
+    requester: {
+      name: 'Data Hub user',
+      email: body.email,
+    },
+    subject: body.title,
+    comment: {
+      body: body.description || 'N/A',
```
teneightfive: Can we not just leave description empty? `N/A` feels like a hack to get round badly required fields.

src/apps/support/middleware.js
tyom: We're ensuring that `formErrors` object exists just a few lines above so we're confident to check properties on it.

src/apps/builders.js
teneightfive: `stuff` :trollface:

PR #678, opened 2017-09-24
CSS: Move layout up the import order

tyom:
As components may have overrides for sibling margins it makes sense to set
layout defaults (specifically assemblies) to order before the overrides.

Example below. See margin between label and radio buttons.
### Before
image

### After
image

### DIFF
--- a/assets/stylesheets/application.scss
+++ b/assets/stylesheets/application.scss
@@ -5,10 +5,10 @@ $images-dir: "/images/";
 @import "settings";
 @import "tools";
 @import "generic";
+@import "layout";
 @import "elements";
 @import "objects";
 @import "components";
-@import "layout";
 @import "trumps";
 
 //// application specific (deprecated)


assets/stylesheets/application.scss
```diff
@@ -5,10 +5,10 @@ $images-dir: "/images/";
 @import "settings";
 @import "tools";
 @import "generic";
+@import "layout";
```
ztolley: Shouldn't CSS be specified so that order is irrelevant?

tyom: Ideally it should. However I think there are cases when specificity dimension (load order) is also valid.

For example typography and generic styles would be loaded before everything else. One could argue the same should be with layout.

We can make the `.c-multiple-choice--inline` selector stronger but should we? It will only make it harder further down the line when we want to override it again for another purpose.

ztolley: I know we think of things like type and layout as the foundations, so tend to put them first but, in reality, it's all about the selector score and shouldn't matter. The browsers not going to parse them until the whole CSS file is loaded is it? It feels a bit hacky to put something after to make sure it is more likely to be used, the same way that !important used to be used.

I tend to go with the making the selector stronger and through components and BEM we shouldn't need to override it for another purpose should we?

I also load dependants alphabetically to force myself to make selectors correct and not to depend on load order, plus easier to scan the list to check an import is there.

tyom: Adding specificity to components will push the problems further down the line when those components would need to be overridden. I'd rather have components and modifiers with consistent specificity score than having layout importable in any order.

PR #680, opened 2017-09-25
Enable breadcrumbs only on the third level of hirachy

tyom:
I'd like to propose to show breadcrumbs starting from the third level of depth.

I think that the first two levels (Home and global sections) are easily accessible from the header and make it obvious enough to the user where they are which makes the breadcrumb somewhat redundant, adding to UI noise.

### Before

image
image

### After
image
image

I encourage people to check out the 💥 [Demo](https://datahub-beta2-pr-680.herokuapp.com/) 💥 to get the feel for how it works.

### DIFF
--- a/assets/stylesheets/components/_local-header.scss
+++ b/assets/stylesheets/components/_local-header.scss
@@ -1,6 +1,6 @@
 @import "../tools/mixins";
 
-$_first-element-spacing: $default-spacing-unit * 3;
+$_first-element-spacing: $default-spacing-unit * 4;
 
 .c-local-header {
   @include clearfix;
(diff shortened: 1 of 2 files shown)


feedmypixel:
> Also, what about going full hog and removing the whole of the local-header when on collections. Not sure it actually provides that much use now:

This will cause some jumping around no?

feedmypixel:
I like this approach

teneightfive:
>This will cause some jumping around no?

@feedmypixel what jumping do you mean?

feedmypixel:
> @feedmypixel what jumping do you mean?
If the space was removed when drilling down and moving back up through pages the header with h1 would go up and down.

teneightfive:
@feedmypixel I suppose my thought was that it's a different type of layout so that wouldn't really matter. You'd be clearly going into a different type of thing.

There's jumping happening at the moment when you go from search results to entity. The local header switches background too.

PR #682, opened 2017-09-25
DH-594 Add service to event form

yeahfro:
This change adds the service field to the create event form.

## New service field
screen shot 2017-09-25 at 15 06 33

## Validation on service field
screen shot 2017-09-25 at 15 07 00

### DIFF
--- a/src/apps/events/macros/event-edit-form.js
+++ b/src/apps/events/macros/event-edit-form.js
@@ -108,6 +108,7 @@ const eventFormConfig = ({ advisers }) => {
           return metadataRepo.teams.map(transformObjectToOption)
         },
       },
+      globalFields.serviceDeliveryServices,
       {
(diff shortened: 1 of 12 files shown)


src/apps/events/macros/event-edit-form.js
```diff
@@ -103,6 +103,15 @@ const eventFormConfig = ({ advisers }) => {
       },
       {
         macroName: 'MultipleChoiceField',
+        name: 'service',
+        label: 'Service',
+        initialOption: '-- Select service --',
+        options () {
+          return metadataRepo.serviceDeliveryServiceOptions.map(transformObjectToOption)
```
tyom: This could be moved to global fields [macros.js](https://github.com/uktrade/data-hub-frontend/blob/develop/src/apps/macros.js).

PR #683, opened 2017-09-25
TLS Support for Redis

jim68000:
Our new Redis set up has TLS and encryption built in, so this commit removes the buggy connect-redis-crypto in favour of the standard connect-redis module. 

It also introduces a new environment variable - REDIS_USE_TLS - that must be set to true in an environment where TLS is enabled.

I've included one tiny comment in the redis-store file, as I don't think what `rejectUnauthorized` does is easy to understand from its name.


jim68000:
Ugh

PR #686, opened 2017-09-26
Company router refactor

teneightfive:
This refactors the remaining child routes in company to use the shared param middleware rather than having to manually get the company in each route.


src/apps/companies/router.js
```diff
@@ -1,21 +1,17 @@
...
 const { renderCompanyList } = require('./controllers/list')
 const { renderForm } = require('./controllers/edit')
 const { renderDetails } = require('./controllers/view')
+const { renderInvestments } = require('./controllers/investments')
+const { renderAuditLog } = require('./controllers/audit')
+const { renderInteractions } = require('./controllers/interactions')
 const { renderCompaniesHouseCompany } = require('./controllers/companies-house')
+const { archiveCompany, unarchiveCompany } = require('./controllers/archive')
+const { renderContacts } = require('./controllers/contacts')
+const { renderExports, populateExportForm, renderExportEdit, handleEditFormPost } = require('./controllers/exports')
```
ztolley: Can we set a rule for how many functions to expand in a single line before we put each on a new line?

src/apps/companies/router.js
```diff
@@ -37,35 +33,32 @@ router
...
-    '/add/ukother',
-    '/add/ltd',
-    '/add/ltd/:companyNumber',
+    '/add',
+    '/add/:companyNumber',
     '/:companyId/edit',
   ])
-  .get(populateForm, renderForm)
-  .post(populateForm, handleFormPost, renderForm)
+  .all(populateForm)
```
ztolley: never knew All was a thing... how does it differ to USE and where else can we use it?

src/apps/companies/router.js
```diff
@@ -37,35 +33,32 @@ router
...
-  .post(expController.post)
-
-router.get('/view/ch/:companyNumber', renderCompaniesHouseCompany)
-router.get('/:companyId', renderDetails)
+  .route('/:companyId/exports/edit')
+  .all(populateExportForm)
+  .post(handleEditFormPost)
+  .all(renderExportEdit)
 
 router
```
ztolley: Possibly the most exciting piece of code in the entire datahub app...

src/apps/companies/router.js
```diff
@@ -37,35 +33,32 @@ router
...
+  .all(renderForm)
 
-router.get('/:id/investments', investmentsController.getAction)
-router.get('/:id/audit', auditController.getAudit)
+router.post('/:companyId/archive', archiveCompany)
+router.get('/:companyId/unarchive', unarchiveCompany)
 
+router.get('/:companyId/contacts', renderContacts)
 router.get('/:companyId/interactions', getInteractionCollection, renderInteractions)
+router.get('/:companyId/exports', renderExports)
```
ztolley: 👍

src/apps/companies/router.js
```diff
@@ -37,35 +33,32 @@ router
...
+router.get('/:companyId/investments', renderInvestments)
+router.get('/:companyId/audit', renderAuditLog)
+
+router.get('/view/ch/:companyNumber', renderCompaniesHouseCompany)
```
ztolley: maybe /:companyNumber/companieshouse ??

src/apps/companies/controllers/archive.js
```diff
@@ -1,38 +1,49 @@
-const { archiveCompany, unarchiveCompany, getDitCompany } = require('../repos')
-const { buildCompanyUrl } = require('../services/data')
+/* eslint-disable camelcase */
+const { archiveCompany: archive, unarchiveCompany: unarchive } = require('../repos')
```
ztolley: 👍

src/apps/companies/controllers/audit.js
```diff
@@ -1,36 +1,32 @@
...
 
-    return res
+    res
+      .breadcrumb(name, `/companies/${id}`)
       .breadcrumb('Audit history')
       .render('companies/views/audit', {
-        tab: 'audit',
-        company,
         auditLog,
+        tab: 'audit',
```
ztolley: we are getting rid of this soon, right?

src/apps/companies/controllers/contacts.js
```diff
@@ -1,43 +1,30 @@
...
 
-    const transformedContacts = company.contacts
-      .map(transformContactToListItem)
-      .map(contact => {
-        const meta = reject(contact.meta, ['label', 'Company'])
-        return Object.assign({}, contact, { meta })
-      })
-    const activeContacts = transformedContacts.filter(contact => !contact.isArchived)
-    const archivedContacts = transformedContacts.filter(contact => contact.isArchived)
+  const transformedContacts = contacts
```
ztolley: This should be in a transformer

src/apps/companies/router.js
teneightfive: Don't think there is a hard and fast rule, it often just comes down to line length for me. For imports if it starts to creep over 80 characters I'll sometimes drop them down.

src/apps/companies/router.js
teneightfive: Yeah, @tyom pointed me at it. It [matches all HTTP verbs](http://expressjs.com/en/api.html#router.use). I think it's slightly cleaner than chaining the middleware on HTTP verb methods like `post` and `get`.

So instead of this:

```js
router
  .get(renderMethod)
  .post(postHandler, renderMethod)
```

you can do this to save duplicating the call to `renderMethod`:

```js
router
  .post(postHandler)
  .all(renderMethod)
```

I think `.use()` is used for router level middleware rather than route middleware. Potentially a pattern we could also explore. I think we're always evolving these patterns and @tyom and I mentioned it might be worth the FEDs getting together and starting to agree and document the patterns that are coming out.

src/apps/companies/router.js
teneightfive: Yeah, I was trying not to change too much. Will do that as a follow up PR I think if that's okay?

src/apps/companies/controllers/audit.js
teneightfive: We are. Again, a follow up PR.

src/apps/companies/controllers/contacts.js
teneightfive: I was trying not to change the functionality of the actual controllers too much. This feels like it could also be a separate PR to change how we get and format contacts for this view but don't want to pollute this change even more. As you pointed out, it's already quite a lot to review.

src/apps/companies/router.js
teneightfive: ??

src/apps/companies/controllers/add.js
```diff
@@ -70,7 +70,7 @@ function postAddStepOne (req, res, next) {
   const queryString = buildQueryString(params)
 
   if (req.body.business_type === 'ukother' || req.body.business_type === 'foreign') {
-    return res.redirect(`/companies/add/${req.body.business_type + queryString}`)
+    return res.redirect(`/companies/add${queryString}`)
```
tyom: Maybe use `query-string` package we have. Our own `buildQueryString` is too simplistic. For example, it doesn't support multiple values.

src/apps/companies/controllers/archive.js
feedmypixel: Always makes me think if we are changing the variable name like this should we just rename the canonical implementation? Maybe in another PR?

src/apps/companies/controllers/archive.js
```diff
@@ -1,38 +1,49 @@
...
+const logger = require('../../../../config/logger')
+
+async function archiveCompany (req, res) {
+  const { id } = res.locals.company
```
feedmypixel: what id? could this be `companyId` for clarity later in the file?

src/apps/companies/controllers/archive.js
```diff
@@ -1,38 +1,49 @@
...
+  const { archived_reason, archived_reason_other } = req.body
+  const reason = archived_reason !== 'Other' ? archived_reason : archived_reason_other
```
tyom: Might be easier to read if flipped to non-negating condition check.

src/apps/companies/controllers/archive.js
```diff
@@ -1,38 +1,49 @@
...
+    logger.error(error)
+
+    req.flash('error', 'Company could not be archived')
+    res.redirect(returnUrl)
   }
 }
 
-async function getUnarchiveCompany (req, res, next) {
+async function unarchiveCompany (req, res) {
+  const { id } = res.locals.company
```
feedmypixel: what id? could this be companyId for clarity later in the file?

src/apps/companies/controllers/audit.js
```diff
@@ -1,36 +1,32 @@
...
-    const token = req.session.token
-    const companyId = req.params.id
-    const page = req.query.page || 1
-
-    const company = await getDitCompany(token, companyId)
-    getCommonTitlesAndlinks(req, res, company)
+async function renderAuditLog (req, res, next) {
+  const token = req.session.token
+  const page = req.query.page || 1
+  const { id, name } = res.locals.company
```
feedmypixel: what id? what name? could this be companyId for clarity later in the file?

src/apps/companies/controllers/contacts.js
```diff
@@ -1,43 +1,30 @@
...
- *  Gets a list of contact associated with a company and splits them into curent and archived contacts
- *  Furthermore the controller extracts only the fields required and pre-formats them to make layout easier.
- */
-async function getContacts (req, res, next) {
-  try {
-    res.locals.tab = 'contacts'
-    const company = res.locals.company = await getDitCompany(req.session.token, req.params.id)
-    getCommonTitlesAndlinks(req, res, company)
+function renderContacts (req, res, next) {
+  const { id, name, contacts } = res.locals.company
```
feedmypixel: variable naming feels very generic

src/apps/companies/controllers/exports.js
```diff
@@ -0,0 +1,84 @@
...
+
+  res
+    .breadcrumb(name, `/companies/${id}`)
+    .breadcrumb('Exports')
+    .render('companies/views/exports-view', {
+      exportDetailsLabels,
+      tab: 'exports',
+      exportDetailsDisplayOrder: ['exportToCountries', 'futureInterestCountries'],
+      exportDetails: {
+        exportToCountries: map(export_to_countries, 'name').join('<br>'),
```
feedmypixel: feels like maybe this html work should be done in the view in a loop?

src/apps/companies/controllers/exports.js
```diff
@@ -0,0 +1,84 @@
...
+        futureInterestCountries: map(future_interest_countries, 'name').join('<br>'),
+      },
+    })
+}
+
+function populateExportForm (req, res, next) {
+  const { export_to_countries, future_interest_countries } = res.locals.company
```
feedmypixel: separate lines?

src/apps/companies/controllers/exports.js
```diff
@@ -0,0 +1,84 @@
...
+    future_interest_countries: future_interest_countries.map(country => country.id),
+  }
+
+  res.locals.formData = assign({}, savedValues, req.body)
+
+  next()
+}
+
+function renderExportEdit (req, res) {
+  const { id, name } = res.locals.company
```
feedmypixel: names feel very generic, easy to understand here, potentially less so later down the file

src/apps/companies/controllers/investments.js
```diff
@@ -1,34 +1,31 @@
...
-const { getCommonTitlesAndlinks } = require('../services/data')
 const { transformInvestmentProjectToListItem } = require('../../investment-projects/transformers')
 const { transformApiResponseToCollection } = require('../../transformers')
 
-async function getAction (req, res, next) {
+async function renderInvestments (req, res, next) {
   const token = req.session.token
-  const companyId = req.params.id
   const page = req.query.page || 1
+  const { id, name } = res.locals.company
```
feedmypixel: generic variable names

src/apps/companies/router.js
feedmypixel: I'd say this would be clearer to read on separate lines

src/apps/companies/views/details.njk
```diff
@@ -37,7 +37,7 @@
       options: ['Company is dissolved'],
       csrfToken: csrfToken,
       error: form.errors.reason,
-      url: '/companies/archive/' + company.id
+      url: '/companies/' + company.id + '/archive'
```
tyom: `/companies/${company.id}/archive`?

src/apps/companies/router.js
feedmypixel: I do feel we loose a little bit of clarity with using `all`, maybe thats just me needing to adjust to the understanding.

Is there any side affects for adding this for all verbs? 
Feels like this is more intended for `auth`
You may also come across problems when sometimes you want middleware from `all()` and for some verbs you want specifics. This would then involve an adding up of the middleware used in verb by looking at `<verb>()` and `all()`

teneightfive: I suppose one side affects could be that if people do a `PATCH` or `PUT` the server would respond with these functions.

Maybe we drop this pattern for now and go with consistency.

src/apps/companies/controllers/audit.js
teneightfive: I'm going to add some `// TODOs`...

src/apps/companies/router.js
tyom: I think that it's not a bad thing to render a view for those verbs. Particularly that those verbs are generally not used in client (only API). We can always add specific verbs before `all`. It's meant to be a catch-all.

src/apps/companies/controllers/add.js
teneightfive: Have added a follow up TODO as don't want to bloat this change. The purpose of this change was to remove the type from the url path.

src/apps/companies/views/details.njk
teneightfive: It's a nunjucks file. Doesn't support template strings...we've all made that mistake when reviewing 😂

src/apps/companies/controllers/archive.js
teneightfive: Potentially. I wanted to make the controller methods more inline with the rest of the app for now but it clashed with the repo.

We've started floating a pattern for method names within repos so doing a global change will solve the need to change the name here.

src/apps/companies/controllers/archive.js
teneightfive: It's destructuring the id property from the `res.locals.company` object. Changing the name within this method would mean doing what you pointed out above is not ideal:

```js
const { id: companyId } = res.locals.company
```

Thoughts?

src/apps/companies/controllers/exports.js
teneightfive: Absolutely, but I'd say that's a separate change.

This is all because we're using the legacy `keyvaluetable`. @tyom and I discussed implementing a macro similar to entity list to display tables if we really need them but that they accept an array of items and each time can have an optional `type` property to determine how it is rendered.

I'd like to reduce the use of any kind of HTML in JS files to only nunjucks filters. And remove it from any controllers or services.

src/apps/companies/controllers/exports.js
teneightfive: It's only 2 properties so didn't feel it was necessary.

src/apps/companies/controllers/exports.js
teneightfive: Most of these controllers are small enough that I don't think it causes a huge problem. As mentioned above, these are being destructured from the original company object. Changing the name would mean mapping this to a different name which could causes a similar cognitive load.

src/apps/companies/controllers/archive.js
feedmypixel: I would approach this without destructuring `companyId = res.locals.company.id`. Always been a fan of verbose and declarative variables. But as mentioned these are not huge files so should be ok.

src/apps/companies/router.js
feedmypixel: Its more the clarity of the code that feels like it potentially may get a little confusing. `get`, `post`, etc feel more explicit.

One for a f2f natter I suspect

PR #689, opened 2017-09-26
Deprecated code

teneightfive:
Remove unused code from companies services.


src/apps/companies/controllers/contacts.js
```diff
@@ -1,43 +1,30 @@
...
 
-    const transformedContacts = company.contacts
-      .map(transformContactToListItem)
-      .map(contact => {
-        const meta = reject(contact.meta, ['label', 'Company'])
-        return Object.assign({}, contact, { meta })
-      })
-    const activeContacts = transformedContacts.filter(contact => !contact.isArchived)
-    const archivedContacts = transformedContacts.filter(contact => contact.isArchived)
+  const transformedContacts = contacts
```
ztolley: shouldn't this be in contact transformer, like interactions?

src/apps/companies/controllers/exports.js
```diff
@@ -0,0 +1,84 @@
...
+    .breadcrumb('Exports', `/companies/${id}/exports`)
+    .breadcrumb('Edit')
+    .render('companies/views/exports-edit', {
+      exportDetailsLabels,
+      tab: 'exports',
+      countryOptions: metadataRepo.countryOptions.map(transformObjectToOption),
+    })
+}
+
+async function handleEditFormPost (req, res, next) {
```
ztolley: Will this get refactored to middleware like other edit flows?

I guess it runs into the same issue as investment in the way it has, in effect, sub apps

src/apps/companies/controllers/contacts.js
teneightfive: Potentially, but the main aim for this change was to remove the need for `getDitCompany` in all of these controllers and remove the need for some to be async functions.

We could do another change to move how we get the contacts in a separate change after.

ztolley: Yup. This is why I'd like a PR for each Sub app, maybe after this, to clean each up separately and let us concentrate on making the best that it can be.

PR #692, opened 2017-09-27
e2e test for UK region display

venureddy95:
E2E acceptance tests for selecting country as UK and setting a uk-region, then verifying it in the event list.

### DIFF
--- a/test/acceptance/features/events/page-objects/Event.js
+++ b/test/acceptance/features/events/page-objects/Event.js
@@ -150,6 +150,25 @@ module.exports = {
             }
           })
       },
+
+      populateCountryAndRegion () {
+        this.state.countryDetails = {
+          address_country: 'United Kingdom',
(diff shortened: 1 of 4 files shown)


test/acceptance/features/events/page-objects/Event.js
```diff
@@ -150,6 +150,25 @@ module.exports = {
             }
           })
       },
+
+      populateCountryAndRegion () {
```
feedmypixel: It feels like we should add to `populateCreateEventForm` to accept values then when we initially call it we can pass through values. Anything with a `null` will then go and retrive a value from a select. Rather than creating a whole new method for it. If ok lets get this in and then if you don't mind I will have a little refactor when I am near this area later or tomorrow and show you.

feedmypixel:
@venureddy95 linting is failing

PR #693, opened 2017-09-27
Replace use of internal query string builder with query-string

teneightfive:
This removes the need to write our own query string builder method
and uses a more comprehensive solution.


teneightfive:
Closing as duplicate of #694

PR #695, opened 2017-09-27
Companies local nav

teneightfive:
Replace companies nav with newer dynamic macro.

This removes the manual markup for the nav and the need to send the tab to each view in order to highlight it.

### DIFF
--- a/src/apps/companies/controllers/audit.js
+++ b/src/apps/companies/controllers/audit.js
@@ -20,7 +20,6 @@ async function renderAuditLog (req, res, next) {
       .breadcrumb('Audit history')
       .render('companies/views/audit', {
         auditLog,
-        tab: 'audit', // TODO: Use newer local nav macro to remove need for this
       })
   } catch (error) {
     next(error)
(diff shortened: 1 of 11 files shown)


tyom:
We've been calling details middleware `details.js` in other places. I'm not against `params.js` but we should make it consistent.

src/apps/companies/middleware/params.js
```diff
@@ -17,9 +18,13 @@ function getHeadingAddress (company) {
   return getFormattedAddress(company, 'registered')
 }
 
-async function getCompany (req, res, next, id) {
+async function getCompany (req, res, next, companyId) {
+  if (!isValidGuid(companyId)) {
```
tyom: Problem is your next controller in the chain expects a company and we get `500`. I think we should let `gitDitCompany` throw and let our error middleware handle `404` which is more appropriate for invalid id.

teneightfive:
>We've been calling details middleware details.js in other places. I'm not against params.js but we should make it consistent.

It's not specific to details though. It's middleware to house param functions. I've used this pattern in OMIS so it's not different to what exists. I feel `details` doesn't really say that much and makes me think its only for the details/overview route, not shared across many routes.

src/apps/companies/middleware/params.js
teneightfive: Will remove commit.

PR #696, opened 2017-09-27
Remove use of buildCompanyUrl and viewcompanyresult

teneightfive:
These helpers were used to solve a problem with splitting company
views into different routes.

These have been consolidated so these helpers are no longer needed.


src/apps/dashboard/services.js
```diff
@@ -31,7 +30,7 @@ function mapInteractions (interactions) {
         subject: interaction.subject,
         company: {
           name: company ? company.name : null,
-          url: company ? buildCompanyUrl(company) : null,
+          url: company ? `/companies/${company.id}` : null,
```
tyom: Should `company` property be `null` instead?

teneightfive: Outside of scope for this PR. Could be addressed when we remove the need for these services.

PR #698, opened 2017-09-27
Refactored dashboard and updated dashboard content

ztolley:
Requested to update the copy to reflect the dashboard now shows last 5 entries.
Simplified the code.

capture

### DIFF
--- a/assets/stylesheets/_deprecated/_dashboard.scss
+++ b/assets/stylesheets/_deprecated/_dashboard.scss
@@ -8,16 +8,9 @@
 
   .dashboard-section-title {
     @include bold-font(24);
-    margin: 0;
-  }
-
-  .dashboard-section-intro {
-    margin-top: $default-spacing-unit / 2;
   }
 
   .dashboard-list {
-    margin-bottom: $default-spacing-unit;
-
     li + li {
       padding-top: $default-spacing-unit / 2;
     }
--- a/src/apps/dashboard/controllers.js
+++ b/src/apps/dashboard/controllers.js
@@ -1,24 +1,23 @@
-const dashboardService = require('./services')
+const { fetchHomepageData } = require('./repos')
 
-function getHandler (req, res) {
-  const days = 15
+async function renderDashboard (req, res, next) {
+  try {
+    const {
+      contacts,
+      interactions,
+    } = await fetchHomepageData(req.session.token)
 
-  dashboardService
-    .getHomepageData(req.session.token, days)
-    .then((data) => {
-      // TODO: Remove manual limit on interactions and contacts once API supports it
-      res
-        .title('Dashboard')
-        .render('dashboard/views/dashboard', {
-          totalDays: days,
-          interactionsCount: data.interactions.length,
-          interactions: data.interactions && data.interactions.length ? data.interactions.slice(0, 5) : [],
-          contactsCount: data.contacts.length,
-          contacts: data.interactions && data.contacts.length ? data.contacts.slice(0, 5) : [],
-        })
-    })
+    res
+      .title('Dashboard')
+      .render('dashboard/views/dashboard', {
+        contacts,
+        interactions,
+      })
+  } catch (error) {
+    next(error)
+  }
 }
 
 module.exports = {
-  getHandler,
+  renderDashboard,
 }
--- a/src/apps/dashboard/views/dashboard.njk
+++ b/src/apps/dashboard/views/dashboard.njk
@@ -17,27 +17,20 @@
       <div class="dashboard-section">
         <h2 class="dashboard-section-title">Interactions</h2>
         {% if interactions.length %}
-          <p class="dashboard-section-intro">
-            You have added {{ interactionsCount }} {{ 'interaction' | pluralise(interactions.length) }} in the last {{ totalDays }} {{ 'day' | pluralise(totalDays) }}
-          </p>
-
           <ol class="dashboard-list">
             {% for interaction in interactions %}
               <li>
                 {% if interaction.company.name %}
-                  <a href="{{ interaction.company.url }}">{{ interaction.company.name }}</a>
+                  <a href="/companies/{{ interaction.company.id }}">{{ interaction.company.name }}</a>
                   &ndash;
                 {% endif %}
-                <a href="{{ interaction.url }}">{{ interaction.subject }}</a>
+                <a href="/interactions/{{ interaction.id }}">{{ interaction.subject }}</a>
               </li>
             {% endfor %}
           </ol>
         {% else %}
-          <p class="dashboard-section-intro">
-            No interactions have been added in the last {{ totalDays }} {{ 'day' | pluralise(totalDays) }}
-          </p>
+          <p>No interactions have been added.</p>
         {% endif %}
-
         <p><a href="/interactions">See all interactions</a></p>
       </div>
     </div>
@@ -47,21 +40,15 @@
         <h2 class="dashboard-section-title">Contacts</h2>
 
         {% if contacts.length %}
-          <p class="dashboard-section-intro">
-            You have added {{ contactsCount }} {{ 'contact' | pluralise(contacts.length) }} in the last {{ totalDays }} {{ 'day' | pluralise(totalDays) }}
-          </p>
-
           <ol class="dashboard-list">
             {% for contact in contacts %}
               <li>
-                <a href="{{ contact.url }}">{{ contact.name }}</a> from <a href="{{ contact.company.url }}">{{ contact.company.name }}</a>
+                <a href="/contacts/{{ contact.id }}">{{ contact.first_name }} {{ contact.last_name }}</a> from <a href="/companies/{{ contact.company.id }}">{{ contact.company.name }}</a>
               </li>
             {% endfor %}
           </ol>
         {% else %}
-          <p class="dashboard-section-intro">
-            No contacts have been added in the last {{ totalDays }} {{ 'day' | pluralise(totalDays) }}
-          </p>
+          <p>No contacts have been added.</p>
         {% endif %}
 
         <p><a href="/contacts">See all contacts</a></p>
--- a/test/unit/apps/dashboard/controllers.test.js
+++ b/test/unit/apps/dashboard/controllers.test.js
@@ -1,0 +1,81 @@
+describe('dashboard controller', () => {
+  beforeEach(() => {
+    this.sandbox = sinon.sandbox.create()
+
+    this.reqMock = Object.assign({}, globalReq, {
+      session: {
+        token: 'abcd',
+      },
+    })
+
+    this.resMock = {
+      locals: {},
+      render: this.sandbox.spy(),
+      title: this.sandbox.stub().returnsThis(),
+    }
+
+    this.nextSpy = this.sandbox.spy()
+
+    this.fetchHomepageDataStub = this.sandbox.stub()
+
+    this.controllers = proxyquire('~/src/apps/dashboard/controllers', {
+      './repos': {
+        fetchHomepageData: this.fetchHomepageDataStub,
+      },
+    })
+  })
+
+  context('when there are no errors calling the API', () => {
+    beforeEach(async () => {
+      this.dashData = {
+        contacts: [{
+          id: '1234',
+        }],
+        interactions: [{
+          id: '4321',
+        }],
+      }
+
+      this.fetchHomepageDataStub.resolves(this.dashData)
+      await this.controllers.renderDashboard(this.reqMock, this.resMock, this.nextSpy)
+    })
+
+    it('should render the dashboard template', () => {
+      expect(this.resMock.render).to.be.calledWith('dashboard/views/dashboard')
+    })
+
+    it('should render the page with contacts', () => {
+      const renderOptions = this.resMock.render.firstCall.args[1]
+      expect(renderOptions.contacts).to.deep.equal(this.dashData.contacts)
+    })
+
+    it('sould render the page with interactions', () => {
+      const renderOptions = this.resMock.render.firstCall.args[1]
+      expect(renderOptions.interactions).to.deep.equal(this.dashData.interactions)
+    })
+
+    it('should not call next', () => {
+      expect(this.nextSpy).to.not.have.been.called
+    })
+
+    it('should set a page title', () => {
+      expect(this.resMock.title).to.be.calledWith('Dashboard')
+    })
+  })
+
+  context('when there is a problem calling the API', () => {
+    beforeEach(async () => {
+      this.error = { status: 500 }
+      this.fetchHomepageDataStub.rejects(this.error)
+      await this.controllers.renderDashboard(this.reqMock, this.resMock, this.nextSpy)
+    })
+
+    it('should show an error', () => {
+      expect(this.nextSpy).to.have.been.calledWith(this.error)
+    })
+
+    it('should not render the page', () => {
+      expect(this.resMock.render).to.not.have.been.called
+    })
+  })
+})
--- a/src/apps/dashboard/index.js
+++ b/src/apps/dashboard/index.js
@@ -1,6 +1,6 @@
 const router = require('express').Router()
-const { getHandler } = require('./controllers')
+const { renderDashboard } = require('./controllers')
 
 module.exports = {
-  router: router.get('/', getHandler),
+  router: router.get('/', renderDashboard),
 }
--- a/src/apps/dashboard/repos.js
+++ b/src/apps/dashboard/repos.js
@@ -1,0 +1,10 @@
+const authorisedRequest = require('../../lib/authorised-request')
+const config = require('../../../config')
+
+function fetchHomepageData (token) {
+  return authorisedRequest(token, `${config.apiRoot}/dashboard/homepage/`)
+}
+
+module.exports = {
+  fetchHomepageData,
+}


tyom:
> No contacts have been added
> [See all contacts](#)

That looks odd. Should it say "no new contacts..."?

ztolley:
its last 5 contacts you have added, so there could be other contacts. The stuff on this screen is things allocated to you.

assets/stylesheets/_deprecated/_dashboard.scss
```diff
@@ -8,7 +8,7 @@
 
   .dashboard-section-title {
     @include bold-font(24);
-    margin: 0;
+    margin: 0 0 $default-spacing-unit / 2 0;
```
tyom: Do you need to set margin? I think default `h2` works fine
image

Also, we typically set top margin for spacing. This allow us to control first item margin to avoid space inconsistencies.

There's `assemblies.scss` for component stacking. And each component should set its elements spacing rules.
That was we can reuse components in different contexts.

ztolley: Without the margin, the spacing is off because the styles expected a paragraph

src/apps/dashboard/controllers.js
```diff
@@ -1,24 +1,23 @@
-const dashboardService = require('./services')
+const { fetchHomepageData } = require('./repos')
```
feedmypixel: should this be `get`? we don't use `fetch` anywhere else do we?

ztolley: Didn't you get the Memo? we've started using fetch

assets/stylesheets/_deprecated/_dashboard.scss
ztolley: Updated to be top margin based

test/unit/apps/dashboard/controllers.test.js
```diff
@@ -0,0 +1,73 @@
...
+
+    this.next = this.sandbox.spy()
+  })
+
+  context('when there are no errors calling the API', () => {
+    beforeEach(async () => {
+      this.contacts = []
+      this.interactions = []
+
+      nock(config.apiRoot)
```
tyom: You could move this outside of `beforeEach` into module scope. It's not needed to be called before each test.

test/unit/apps/dashboard/controllers.test.js
```diff
@@ -0,0 +1,73 @@
...
+      render: this.sandbox.spy(),
+      title: this.sandbox.stub().returnsThis(),
+    }
+
+    this.next = this.sandbox.spy()
+  })
+
+  context('when there are no errors calling the API', () => {
+    beforeEach(async () => {
+      this.contacts = []
```
tyom: What's the point of these? They're just noise. You don't test the data so just initialise them with `[]`.

test/unit/apps/dashboard/controllers.test.js
```diff
@@ -0,0 +1,73 @@
...
+      await getDashboard(this.req, this.res, this.next)
+    })
+
+    it('should render the dashboard template', () => {
+      expect(this.res.render).to.be.calledWith('dashboard/views/dashboard')
+    })
+
+    it('should render the page with contacts', async () => {
+      const options = this.res.render.firstCall.args[1]
+      expect(options.contacts).to.deep.equal(this.contacts)
```
tyom: I'd suggest `expect(option).to.have.property('contacts')`. You're not really checking data.

test/unit/apps/dashboard/controllers.test.js
```diff
@@ -0,0 +1,73 @@
...
+    })
+
+    it('sould render the page with interactions', () => {
+      const options = this.res.render.firstCall.args[1]
+      expect(options.interactions).to.deep.equal(this.interactions)
```
tyom: Same here. I'd check for `options` to have property `interactions`.

test/unit/apps/dashboard/controllers.test.js
```diff
@@ -0,0 +1,73 @@
...
+
+      await getDashboard(this.req, this.res, this.next)
+    })
+
+    it('should render the dashboard template', () => {
+      expect(this.res.render).to.be.calledWith('dashboard/views/dashboard')
+    })
+
+    it('should render the page with contacts', async () => {
+      const options = this.res.render.firstCall.args[1]
```
tyom: Better call it `actual`. `options` is ambiguous.

test/unit/apps/dashboard/controllers.test.js
```diff
@@ -0,0 +1,73 @@
...
+    })
+
+    it('should not call next', () => {
+      expect(this.next.callCount).to.equal(0)
+    })
+  })
+
+  context('when there is a problem calling the API', () => {
+    beforeEach(async () => {
+      nock(config.apiRoot)
```
tyom: This can be moved into module scope.

test/unit/apps/dashboard/controllers.test.js
```diff
@@ -0,0 +1,73 @@
...
+        .get(`/dashboard/homepage/`)
+        .reply(500)
+
+      await getDashboard(this.req, this.res, this.next)
+    })
+
+    it('should show an error', () => {
+      expect(this.next.calledOnce).to.eq(true)
```
tyom: `expect(this.next).to.have.been.calledOnce`

Try to remain consistent with new tests we're all writing.

test/unit/apps/dashboard/controllers.test.js
```diff
@@ -0,0 +1,73 @@
...
+    })
+
+    it('should not render the page', () => {
+      expect(this.res.render.callCount).to.equal(0)
```
tyom: `expect(this.res.render).to.not.have.been.called`

assets/stylesheets/_deprecated/_dashboard.scss
```diff
@@ -8,15 +8,14 @@
...
-    margin: 0;
   }
 
   .dashboard-section-intro {
     margin-top: $default-spacing-unit / 2;
   }
 
   .dashboard-list {
-    margin-bottom: $default-spacing-unit;
+    margin-top: $default-spacing-unit;
```
tyom: Not quite. You need to do it with adjacent sibling combinator. And it being a top-level component would need to be in assemblies.

Although, as it's a deprecated style I'll let you off. It will have to be removed anyway.

src/apps/dashboard/controllers.js
```diff
@@ -1,24 +1,23 @@
...
 
-function getHandler (req, res) {
-  const days = 15
+async function getDashboard (req, res, next) {
```
teneightfive: `renderDashboard`?

test/unit/apps/dashboard/controllers.test.js
ztolley: You test the data that is returned by the fetch gets sent to the page.

src/apps/dashboard/views/dashboard.njk
```diff
@@ -17,27 +17,20 @@
...
-                <a href="{{ interaction.url }}">{{ interaction.subject }}</a>
+                <a href="/interactions/{{ interaction.id }}">{{ interaction.subject }}</a>
               </li>
             {% endfor %}
           </ol>
         {% else %}
-          <p class="dashboard-section-intro">
-            No interactions have been added in the last {{ totalDays }} {{ 'day' | pluralise(totalDays) }}
-          </p>
+          <p class="dashboard-section-intro">No interactions have been added.</p>
```
teneightfive: Do we still need these classes?

src/apps/dashboard/views/dashboard.njk
```diff
@@ -47,21 +40,15 @@
...
         {% if contacts.length %}
-          <p class="dashboard-section-intro">
-            You have added {{ contactsCount }} {{ 'contact' | pluralise(contacts.length) }} in the last {{ totalDays }} {{ 'day' | pluralise(totalDays) }}
-          </p>
-
           <ol class="dashboard-list">
             {% for contact in contacts %}
               <li>
-                <a href="{{ contact.url }}">{{ contact.name }}</a> from <a href="{{ contact.company.url }}">{{ contact.company.name }}</a>
+                <a href="/contacts/{{ contact.id }}">{{ contact.first_name }} {{ contact.last_name }}</a> from <a href="/companies/{{ contact.company.id }}">{{ contact.company.name }}</a>
```
teneightfive: Doesn't `contact` have a `name` property to use rather than joining `first_name` and `last_name`?

test/unit/apps/dashboard/controllers.test.js
ztolley: Then you aren't testing that the property contains the data returned from the server, only that it has the property

tyom:
> its last 5 contacts you have added, so there could be other contacts. The stuff on this screen is things allocated to you.

I understand. I'm just saying that it reads oddly. It doesn't explain that it's the contacts user added.

test/unit/apps/dashboard/controllers.test.js
ztolley: that too seems ambigious

test/unit/apps/dashboard/controllers.test.js
tyom: You don't really need to set it on the object. You can just check that it's `[]`. Less typing, less noise.

test/unit/apps/dashboard/controllers.test.js
tyom: Well, it's not options, is it? It's context if anything.

test/unit/apps/dashboard/controllers.test.js
tyom: OK then: `expect(options).to.have.property('interactions').deep.equal([])`
At least it reads better. Tells you exactly what to expect.

test/unit/apps/dashboard/controllers.test.js
ztolley: Moving nock outside of the before each results in 

`{ RequestError: Error: Nock: No match for request {`

tyom: You should move it to module scope (top level).

test/unit/apps/dashboard/controllers.test.js
ztolley: I'll just use fake data from a json file and check against that

test/unit/apps/dashboard/controllers.test.js
ztolley: You can't have it at the top level as the contexts need different results.

If you move it outside the beforeEach it doesn't work.

This works.

test/unit/apps/dashboard/controllers.test.js
tyom: But is it important? You're not testing nock. You're testing that data is what you expect it to be (mocked).

test/unit/apps/dashboard/controllers.test.js
ztolley: I'll call it render options... it's whatever gets send to render.

test/unit/apps/dashboard/controllers.test.js
ztolley: As discussed, I need to test that it sets the property correctly

src/apps/dashboard/views/dashboard.njk
ztolley: no, thats just adviser, I checked

src/apps/dashboard/controllers.js
ztolley: it doesn't just render.. it gets things too

test/unit/apps/dashboard/controllers.test.js
tyom: The contexts are your URLs. Why do you need to nock before each test?

test/unit/apps/dashboard/controllers.test.js
tyom: It doesn't make sense. `locals` are sent into `render`. Why `options`?

test/unit/apps/dashboard/controllers.test.js
ztolley: I'm testing that it extracts the values from the network response and puts them into the properties sent to the page.

tyom: Cool. Just mock them inline at the top of the file then: `const events = [{ id: 'e1' }]`. Along with `nock` config.

ztolley:
@tyom it reads odd, but it's how it works now and out of scope for the card.

src/apps/dashboard/views/dashboard.njk
ztolley: removed

test/unit/apps/dashboard/controllers.test.js
ztolley: breaks the tests if you do that.

test/unit/apps/dashboard/controllers.test.js
ztolley: it breaks

test/unit/apps/dashboard/controllers.test.js
tyom: You're right, does.

Do you need `nock` though? I think `nock` is better suited for integration tests where one mocks network request and then tests how parts of application respond to it. You can be quick focused in unit test by stubbing `fetchHomepageData`.

src/apps/dashboard/controllers.js
```diff
@@ -1,24 +1,23 @@
...
+  try {
+    const {
+      contacts,
```
tyom: Not sure destructuring here helps only to set the same properties on render. Why not `const locals = fetchHomepageData` and then `render('...', locals)`?

test/unit/apps/dashboard/controllers.test.js
ztolley: Yup. .. this whole integration test approach needs to be settled once and for all.

Setting that will still go in before each

test/unit/apps/dashboard/controllers.test.js
```diff
@@ -0,0 +1,70 @@
...
+
+      await renderDashboard(this.req, this.res, this.next)
+    })
+
+    it('should render the dashboard template', async () => {
+      expect(this.res.render).to.be.calledWith('dashboard/views/dashboard')
+    })
+
+    it('should render the page with contacts', async () => {
+      const renderOptions = this.res.render.firstCall.args[1]
```
tyom: Why `renderOptions`? It's called [locals](http://expressjs.com/en/api.html#app.render).

tyom:
Do we need such a massive [dashdata.json](dashdata.json) file? Can't we strip it out to make it easier to scan? It's bound to become outdated as API changes.

I'd argue we might need it at all. Some basic sample data can be defined in test file.

test/unit/apps/dashboard/controllers.test.js
```diff
@@ -0,0 +1,77 @@
...
+      render: this.sandbox.spy(),
+      title: this.sandbox.stub().returnsThis(),
+    }
+
+    this.next = this.sandbox.spy()
+  })
+
+  context('when there are no errors calling the API', () => {
+    beforeEach(async () => {
+      const { renderDashboard } = proxyquire('~/src/apps/dashboard/controllers', {
```
tyom: Can avoid repeating proxyquire by declaring it once in main `beforeEach`
```js
this.fetchHomepageDataStub = this.sandbox.stub()

this.controllers = proxyquire('~/src/apps/dashboard/controllers', {
  './repos': { fetchHomepageData: this.fetchHomepageDataStub },
})
```
then just resolve or reject for each context
```js
this.fetchHomepageDataStub.resolves(dashData)

await this.controllers.renderDashboard(this.req, this.res, this.next)
```
```js
this.error = { status: 500 }
this.fetchHomepageDataStub.rejects(this.error)

await this.controllers.renderDashboard(this.req, this.res, this.next)
```

test/unit/apps/dashboard/controllers.test.js
```diff
@@ -0,0 +1,77 @@
...
+      },
+    })
+
+    this.res = {
+      locals: {},
+      render: this.sandbox.spy(),
+      title: this.sandbox.stub().returnsThis(),
+    }
+
+    this.next = this.sandbox.spy()
```
tyom: I think we call these `functionType` to make it obvious throughout the code what it is. So `this.nextSpy` would read better.

ztolley: all the other tests I have done and seen others do have been using this.res and this next recently

ztolley:
i put test data into the file before and was told it distracted

test/unit/apps/dashboard/controllers.test.js
ztolley: Then let's make this consistent across all tests because right now it looks like it's 50/50

tyom:
> i put test data into the file before and was told it distracted

It's where you put it. You buried it deep inside `context` `beforeEach`. It's much nicer to declare it at the top (just like you did with JSON import) so it's visible. Now you put it back inside. You're not changing the data between tests. Why do you need to declare it on each test iteration?

test/unit/apps/dashboard/controllers.test.js
tyom: I agree, let's. When we find free time.

ztolley:
in case something in the test changes it if you don't reset the data for each test it can make tests fragile.

tyom:
> in case something in the test changes it if you don't reset the data for each test it can make tests fragile.

And nothing should change it, right? Would be a good way to catch unwanted mutations.

ztolley:
It's simply good practice to have each test isolated and not allow one test to affect another.

src/apps/dashboard/views/dashboard.njk
```diff
@@ -17,27 +17,20 @@
       <div class="dashboard-section">
         <h2 class="dashboard-section-title">Interactions</h2>
         {% if interactions.length %}
-          <p class="dashboard-section-intro">
```
teneightfive: Wonder whether removing this means the users loses context of what these items are? Are they recent interactions I've added? Are they recent interactions anyone has added? Are they interactions I've recently edited?

src/apps/dashboard/views/dashboard.njk
teneightfive: Maybe we should look at adding it to the API to make it consistent?

It's strange that some entities have a combined `name` property and others don't. Makes them harder to work with on the frontend.

test/unit/apps/dashboard/controllers.test.js
```diff
@@ -0,0 +1,81 @@
...
+      const renderOptions = this.resMock.render.firstCall.args[1]
+      expect(renderOptions.interactions).to.deep.equal(this.dashData.interactions)
+    })
+
+    it('should not call next', () => {
+      expect(this.nextSpy).to.not.have.been.called
+    })
+
+    it('should set a page title', () => {
+      expect(this.resMock.title).to.be.calledWith('Dashboard')
```
teneightfive: Wonder whether we need to test the value? Can mean making content changes mean updating tests which is annoying.

src/apps/dashboard/views/dashboard.njk
ztolley: The card says to remove them.

test/unit/apps/dashboard/controllers.test.js
ztolley: checks the page is created with the correct title. We test that we set the breadcrumb in other places.

PR #700, opened 2017-09-27
add organiser to filters and tidy up UAT tests a little

feedmypixel:
- Add organiser to `/event` list filters
- UAT tests
- Unit tests
- Move selectListOptionByText over to commands for global use and rename to clickListOption

screen shot 2017-09-27 at 17 07 45

organiser

### DIFF
--- a/src/apps/events/controllers/edit.js
+++ b/src/apps/events/controllers/edit.js
@@ -1,40 +1,32 @@
 /* eslint camelcase: 0 */
 const { eventFormConfig } = require('../macros')
-const { getAdvisers } = require('../../adviser/repos')
 const { transformEventResponseToFormBody } = require('../transformers')
 const { buildFormWithStateAndErrors } = require('../../builders')
 const { assign, merge, get, pickBy } = require('lodash')
 
-async function renderEditPage (req, res, next) {
+function renderEditPage (req, res) {
   const eventData = transformEventResponseToFormBody(res.locals.event)
+  const advisers = get(res.locals, 'advisers.results')
   const eventId = get(eventData, 'id', '')
   const eventName = get(eventData, 'name')
   const lead_team = eventData.lead_team || get(req, 'session.user.dit_team.id')
+  const mergedData = pickBy(merge({}, eventData, { lead_team }, res.locals.requestBody))
+  const eventForm =
+    buildFormWithStateAndErrors(
+      eventFormConfig({ eventId, advisers }),
+      mergedData,
+      get(res.locals, 'form.errors.messages'),
+    )
 
-  try {
-    const advisersResponse = await getAdvisers(req.session.token)
-    const advisers = advisersResponse.results
-    const mergedData = pickBy(merge({}, eventData, { lead_team }, res.locals.requestBody))
-
-    const eventForm =
-      buildFormWithStateAndErrors(
-        eventFormConfig({ eventId, advisers }),
-        mergedData,
-        get(res.locals, 'form.errors.messages'),
-      )
-
-    res
-      .breadcrumb(eventName, `/events/${eventId}`)
-      .breadcrumb(`${eventId ? 'Edit' : 'Add'} event`)
-      .render('events/views/edit', {
-        eventForm: assign(eventForm, {
-          returnLink: `/events/${eventId}`,
-          returnText: 'Cancel',
-        }),
-      })
-  } catch (e) {
-    next(e)
-  }
+  res
+    .breadcrumb(eventName, `/events/${eventId}`)
+    .breadcrumb(`${eventId ? 'Edit' : 'Add'} event`)
+    .render('events/views/edit', {
+      eventForm: assign(eventForm, {
+        returnLink: `/events/${eventId}`,
+        returnText: 'Cancel',
+      }),
+    })
 }
 
 module.exports = {
--- a/src/apps/events/controllers/list.js
+++ b/src/apps/events/controllers/list.js
@@ -1,15 +1,16 @@
 const { get, omit, merge, assign } = require('lodash')
-const { eventFiltersFields: filtersFields, eventSortForm } = require('../macros')
+const { eventFiltersFields, eventSortForm } = require('../macros')
 const { buildSelectedFiltersSummary } = require('../../builders')
 
 function renderEventList (req, res) {
+  const advisers = get(res.locals, 'advisers.results')
   const sortForm = merge({}, eventSortForm, {
     hiddenFields: assign({}, omit(req.query, 'sortby')),
     children: [
       { value: req.query.sortby },
     ],
   })
-
+  const filtersFields = eventFiltersFields({ advisers })
   const selectedFilters = buildSelectedFiltersSummary(filtersFields, req.query)
   const isUKSelected = get(selectedFilters, 'address_country.valueLabel') === 'United Kingdom'
 
--- a/src/apps/events/middleware/details.js
+++ b/src/apps/events/middleware/details.js
@@ -1,8 +1,8 @@
 const { assign } = require('lodash')
 
-const { fetchEvent } = require('../repos')
 const { transformEventResponseToViewRecord, transformEventFormBodyToApiRequest } = require('../transformers')
-const { saveEvent } = require('../repos')
+const { fetchEvent, saveEvent } = require('../repos')
+const { getAdvisers } = require('../../adviser/repos')
 
 async function postDetails (req, res, next) {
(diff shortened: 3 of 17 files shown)


src/apps/events/router.js
```diff
@@ -15,14 +15,15 @@ router.param('id', getEventDetails)
...
   getRequestBody,
   getEventsCollection,
   renderEventList,
 )
 
 router.route(['/create', '/:id/edit'])
-  .post(postDetails)
-  .all(renderEditPage)
+  .get(getDetails, getEventDetails, renderEditPage)
+  .post(getDetails, getEventDetails, postDetails, renderEditPage)
```
tyom: Why?

feedmypixel: Feels easy to read (although a bit of variable duplication)
```
.get(getDetails, getEventDetails, renderEditPage)
.post(getDetails, getEventDetails, postDetails, renderEditPage)
```

Feels quite a little bit harder to work out whats doing what
```
.all(getDetails, getEventDetails)
  .get(renderEditPage)
  .post(postDetails, renderEditPage)
```

feedmypixel: actually I'm not overly sold on my comment, lets chat f2f

src/apps/events/middleware/details.js
```diff
@@ -37,12 +37,22 @@ async function getEventDetails (req, res, next, eventId) {
...
     res.locals.eventViewRecord = transformEventResponseToViewRecord(res.locals.event)
     next()
-  } catch (error) {
-    next(error)
+  } catch (err) {
+    next(err)
+  }
+}
+
+async function getDetails (req, res, next) {
```
tyom: Surely this should be called `getAdvisers`.

src/apps/events/router.js
tyom: ~`getDetails`~ `getAdvisers` should go on `param` method as it's needed on each event. Then you're left with `postDetails` and `renderEditPage` which should work with previous code.

ztolley: Agree that 'All' makes it harder to read

feedmypixel: > getDetails getAdvisers should go on param method as it's needed on each event. Then you're left with postDetails and renderEditPage which should work with previous code.

@tyom my ponder and thought about it whilst 🏊  I am in agreement with this

src/apps/events/router.js
```diff
@@ -15,14 +15,16 @@ router.param('id', getEventDetails)
...
+  getAdviserDetails,
   getRequestBody,
   getEventsCollection,
   renderEventList,
 )
 
 router.route(['/create', '/:id/edit'])
-  .post(postDetails)
-  .all(renderEditPage)
+  .all(getAdviserDetails, getEventDetails)
```
tyom: Please review the order. Terminal middleware without `next` (controllers) should be at the bottom.

test/acceptance/features/events/page-objects/Event.js
```diff
@@ -41,7 +41,7 @@ module.exports = {
     relatedProgrammes: '#field-related_programmes',
     addAnotherProgramme: 'input[name="add_related_programme"]',
     saveButton: {
-      selector: '//button[text() = \'Save and Continue\']',
+      selector: '//button[text() = \'Save\']',
```
tyom: Double quote?

test/unit/apps/events/controllers/list.test.js
```diff
@@ -50,24 +57,26 @@ describe('Event list controller', () => {
...
-            {
-              macroName: 'MultipleChoiceField',
-              name: 'uk_region',
-              options: [
-                { value: '1', label: 'England' },
-                { value: '2', label: 'Scotland' },
-              ],
-            },
-          ],
+          eventFiltersFields: () => {
```
tyom: Maybe define this at the top of the file if you're not changing it between tests so it's more visible.

src/apps/events/router.js
```diff
@@ -22,9 +22,9 @@ router.get('/',
 )
 
 router.route(['/create', '/:id/edit'])
-  .all(getAdviserDetails, getEventDetails)
-  .get(renderEditPage)
+  .all(getAdviserDetails)
   .post(postDetails, renderEditPage)
+  .get(renderEditPage)
```
feedmypixel: @tyom this ordering with `all` works fine.

test/unit/apps/events/controllers/list.test.js
feedmypixel: So we need 2 `proxyquires` because the `proxyquire` with overrides needs to be added again. You can't override a stub that has been `proxyquired` in this way.
I have moved the code around so it makes more sense though.

test/acceptance/commands/clickListOption.js
```diff
@@ -0,0 +1,9 @@
+/**
+ * Utility to click a specific option by textContent in a named select element element
```
yeahfro: Extra "element"

src/apps/events/controllers/edit.js
```diff
@@ -1,40 +1,33 @@
 /* eslint camelcase: 0 */
 const { eventFormConfig } = require('../macros')
-const { getAdvisers } = require('../../adviser/repos')
 const { transformEventResponseToFormBody } = require('../transformers')
 const { buildFormWithStateAndErrors } = require('../../builders')
 const { assign, merge, get, pickBy } = require('lodash')
 
-async function renderEditPage (req, res, next) {
+function renderEditPage (req, res) {
```
ztolley: Could there be an error thrown requiring Next to be put back in?

test/acceptance/commands/clickListOption.js
feedmypixel: oh yes! ta

test/unit/apps/events/controllers/list.test.js
tyom: Can't you set the stub on `this` and the return different results depending on the context?

src/apps/events/controllers/edit.js
tyom: Do we need to redirect to `500` error page or can we do something in-page? e.g. flash a message?

I think we should try and recover from errors in a more user-friendly way rather than always fall back to scary server error page.

src/apps/events/controllers/list.js
```diff
@@ -1,15 +1,17 @@
 const { get, omit, merge, assign } = require('lodash')
-const { eventFiltersFields: filtersFields, eventSortForm } = require('../macros')
+const { eventFiltersFields, eventSortForm } = require('../macros')
 const { buildSelectedFiltersSummary } = require('../../builders')
 
 function renderEventList (req, res) {
+  const adviserData = res.locals.advisers
+  const advisers = adviserData.results
```
tyom: Could just `const advisers = get(res.locals, 'advisers.results')`. Do you need the redundant `adviserData`?

src/apps/events/controllers/edit.js
feedmypixel: Feels like this is a global question for another PR?

I did original have an if wrapped around it to check for `res.locals.advisers` and then a `next` to error out if `res.locals.advisers` was not there but the middleware has a `next` if the call to `getAdvisers` fails so this covers that situation. It is also expected that you get advisers in the middleware before this method.

tyom: I think in this case, as the function is not async we shouldn't expect any unexpected errors.

test/unit/apps/events/controllers/list.test.js
```diff
@@ -11,19 +36,15 @@ describe('Event list controller', () => {
     this.res = {
       render: this.sandbox.spy(),
```
tyom: Could change to `resMock` and `renderSpy` whilst you're in this file. It would be easier to follow further down in the code.

tyom: Also `nextSpy` etc. I think all mocked functions should have a suffix of what they are: spy, stub etc.

tyom: Sorry, obviously `renderSpy` won't work, doh. But the others still apply 😄

test/unit/apps/events/controllers/list.test.js
```diff
@@ -11,19 +36,15 @@ describe('Event list controller', () => {
...
       query: {},
+      locals: {
+        advisers: advisersData,
+      },
     }
-
+    this.eventFiltersFieldsStub = this.sandbox.stub()
     this.buildSelectedFiltersSummaryStub = this.sandbox.spy()
```
tyom: That's a `Spy` not a `Stub`.

PR #701, opened 2017-09-27
Break three quarter column at same breakpoint as one column

teneightfive:
This is a quick fix to only float the three quarter column class
when the one column class floats. This stops an overlap in columns
between tablet and desktop.

## Before
localhost_3001_contacts_sortby modified_on 3adesc 1

## After
localhost_3001_contacts_sortby modified_on 3adesc


teneightfive:
You're welcome 😉

PR #711, opened 2017-09-30
Fix DateFieldset

tyom:
DateFieldset was not great with handling partial dates so when a form error occurred because the date was not in the right format the missing fields would be prefilled to unexpected values.

### DIFF
--- a/src/apps/transformers.js
+++ b/src/apps/transformers.js
@@ -36,10 +36,11 @@ function transformDateObjectToDateString (key) {
     throw Error('date object key is required to transform date')
   }
   return function transformDateObjectToStringWithKey (props = {}) {
-    return ['year', 'month', 'day']
+    const dateString = ['year', 'month', 'day']
       .map(x => props[`${key}_${x}`])
-      .filter(x => x)
-      .join('-') || null
+      .join('-')
+
+    return dateString === '--' ? null : dateString
   }
 }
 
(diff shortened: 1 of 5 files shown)


src/apps/transformers.js
```diff
@@ -36,10 +36,11 @@ function transformDateObjectToDateString (key) {
...
   }
   return function transformDateObjectToStringWithKey (props = {}) {
-    return ['year', 'month', 'day']
+    const dateString = ['year', 'month', 'day']
       .map(x => props[`${key}_${x}`])
-      .filter(x => x)
-      .join('-') || null
+      .join('-')
+
+    return dateString !== '--' ? dateString : null
```
yeahfro: wouldn't `dateString === '--' ? null : dateString` read better here?

tyom: You know, that's how I originally wrote it but then for some reason I thought that `dateString`  being first makes more sense. Now I'm starting to think it doesn't.

src/templates/_macros/form/date-fieldset.njk
```diff
@@ -12,17 +12,19 @@
...
-  {% set value = props.value|default('') %}
-  {% set formattedDate = props.value | formatDate('DD-MM-YYYY') %}
+  {% set value = props.value | default('') %}
 
   {% if value | isPlainObject  %}
     {% set dateObj = props.value %}
-  {% else %}
-    {% set splitDate = formattedDate.split('-') if formattedDate %}
+  {% elif value | isString %}
+    {# Expects YYYY-MM-DD #}
```
teneightfive: Seems it also accepts a datetime?

PR #713, opened 2017-10-02
Add acceptance tests for event edit feature

tyom:
This adds 10 new scenarios for editing an event.

### Before
image

### After
image


ztolley:
There seems to be a change in the level of verbosity in the tests which makes me worry they would end up being fragile. Interested to hear what others views are as to what level of abstraction we are aiming for with UAT

test/acceptance/features/events/step_definitions/edit.js
```diff
@@ -1,80 +1,112 @@
...
-  When(/^I change event dates field to a new value$/, async () => {
-  })
+  When(/^I change start date to decrease year by one$/, async () => {
+    await Form.getState()
 
-  Then(/^I verify the event dates is updated with new value$/, async () => {
-  })
+    await Event
+      .api.perform(() => {
+        const start_date_year = Form.state.start_date_year || '2017'
```
feedmypixel: could we just use `current` year? `(new Date).getFullYear()`

test/acceptance/features/events/step_definitions/edit.js
```diff
@@ -1,80 +1,112 @@
...
 
-  When(/^I change event location type field to a new value$/, async () => {
-  })
+        Form.state = {
+          '@startDateYear': parseInt(start_date_year, 10) - 1,
+          '@startDateMonth': Form.state.start_date_month || '10',
+          '@startDateDay': Form.state.start_date_day || '01',
```
feedmypixel: same could we just use the current of each?

test/acceptance/features/events/step_definitions/edit.js
```diff
@@ -1,80 +1,112 @@
...
+  When(/^I change dropdown (.+) to be (.+)/, async (fieldName, newValue) => {
+    Form.state[fieldName] = newValue
 
-  When(/^I change event notes field to a new value$/, async () => {
+    await Event
+      .clickListOption(fieldName, Form.state[fieldName])
   })
 
-  Then(/^I verify the event notes is updated with new value$/, async () => {
+  // Thens
```
feedmypixel: is the comment needed? its explicit with the name?

tyom: I only added it because I started grouping them by their name, so it's easier to find a group of thens when scanning the file.

test/acceptance/features/events/step_definitions/edit.js
```diff
@@ -48,16 +49,18 @@ defineSupportCode(({ Before, Then, When }) => {
...
 
   When(/^I change start date to decrease year by one$/, async () => {
+    const currentDate = new Date()
+
     await Form.getState()
 
     await Event
       .api.perform(() => {
-        const start_date_year = Form.state.start_date_year || '2017'
+        const start_date_year = Form.state.start_date_year || format(currentDate, 'YYYY')
```
feedmypixel: 👍

PR #716, opened 2017-10-02
Remove interactions nav and search

ztolley:
Interactions should only be visible within the context of a contact, company or investment project, not from the top level navigation or search results.

* Removed interactions from top-level nav
* Removed interactions tab from search
* Removed code related to interaction lists no longer used
* Removed tests no longer used


tyom:
Could we use this sub-app for embedding it into other apps rather than completely removing?

ztolley:
The common code that is used for company/contact/investment views remains and is not removed, this is just to remove the code for the top level view and remove the nav tabs

tyom:
This breaks the breadcrumb and `/interactions` URL. Do we have a plan on how we're going to support those?

ztolley:
The point is we remove the /interactions endpoint. but if this is a link in the breadcrumb I need to correct that

PR #718, opened 2017-10-02
bump the pause duration 

feedmypixel:
due to ES taking a little longer than usual to sync
also added `retryAssertionTimeout`


tyom:
Might be worth also setting `retryAssertionTimeout` to a second or two.

PR #723, opened 2017-10-03
Update components pages

tyom:
- Remove old and unused components
- Fix broken Collection component

### DIFF
--- a/src/apps/components/views/collection.njk
+++ b/src/apps/components/views/collection.njk
@@ -1,13 +1,14 @@
 {% extends "./_layout.njk" %}
 
-{% from "_macros/results.njk" import ResultsSummary %}
+{% from "_macros/collection/collection-header.njk" import CollectionHeader %}
 
 {% block body_main_content %}
-  {% call Example('ResultsFilters') %}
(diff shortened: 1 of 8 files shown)


src/apps/components/views/collection.njk
```diff
@@ -33,9 +34,9 @@
             ],
             initialOption: 'All types',
             modifier: ['smaller', 'light']
-          }),
-
-          MultipleChoiceField({
+          },
+          {
+            type: 'MultipleChoiceField',
```
teneightfive: `macroName`?

src/apps/components/views/entity-list.njk
```diff
@@ -3,92 +3,13 @@
 {% from "_macros/entity.njk" import Entity with context %}
 
 {% block body_main_content %}
-  {% call Example('Investment projects') %}
+  {% call Example('entity-list of Entity') %}
```
teneightfive: `entity-list of Entity` doesn't seem to read correctly. What about `List of projects (entities)`?

tyom: The idea is to use the actual component names.

PR #729, opened 2017-10-04
Refactor AT for event edit to utilise multiple features

tyom:
- Use better language for feature files, while retaining parameterisation
- Split page into separate features which can be reused across different contexts (Details to enable testing key/value tables and Form to contain form-specific features)
- Add `World` for sharing state between tests.

screen shot 2017-10-04 at 11 36 08

### DIFF
--- a/test/acceptance/features/details/step_definitions/details.js
+++ b/test/acceptance/features/details/step_definitions/details.js
@@ -1,0 +1,38 @@
+const { client } = require('nightwatch-cucumber')
+const { defineSupportCode } = require('cucumber')
+
+const World = require('../../../common/world')
+
+defineSupportCode(({ Then }) => {
+  const Details = client.page.Details()
+
+  // Then
+
+  Then(/^details heading should contain what I entered for "(.+)" field$/, async (fieldName) => {
+    await Details
+      .assert.containsText('@heading', World.state[fieldName])
+  })
+
+  Then(/^details heading should contain "(.+)"$/, async (value) => {
+    await Details
+      .assert.containsText('@heading', value)
+  })
+
+  Then(/^details view data for "(.+)" should contain what I entered for "(.+)" field$/, async (detailsItemName, fieldName) => {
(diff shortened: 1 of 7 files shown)


test/acceptance/features/events/step_definitions/edit.js
```diff
@@ -1,20 +1,21 @@
...
+const World = require('../../../common/world')
+
 defineSupportCode(({ Before, Then, When }) => {
   const Form = client.page.Form()
   const Event = client.page.Event()
   const EventList = client.page.EventList()
 
   Before(() => {
-    Form.state = {}
+    World.resetState()
```
feedmypixel: I wonder if we could just use [hooks.js](http://mucsi96.github.io/nightwatch-cucumber/#hooks) and on all tests reset `World.resetState()`. This would be something to think of in another PR as it may be overkill running it for all.

test/acceptance/features/events/step_definitions/edit.js
```diff
@@ -1,20 +1,21 @@
...
   })
 
-  // Whens
+  // When
```
yeahfro: do we really need this?

test/acceptance/features/details/step_definitions/details.js
```diff
@@ -0,0 +1,38 @@
+const { client } = require('nightwatch-cucumber')
+const { defineSupportCode } = require('cucumber')
+
+const World = require('../../../common/world')
+
+defineSupportCode(({ Then }) => {
+  const Details = client.page.Details()
+
+  // Then
```
yeahfro: do we really need this?

test/acceptance/features/events/step_definitions/edit.js
tyom: I added those to be able to create visual hooks for groups of Whens and Thens when looking through a longer file. I find it easier to scan as they serve as titles.

PR #734, opened 2017-10-04
Add sort to contact interactions.

ztolley:
- [x] Refactor interactions collection middleware so it can be used across multiple sub-apps.
- [x] Update contact interactions route to use interactions middleware and strip out code no longer needed

By Refactoring the interactions collection it can be used across sub apps and they will inherit the ability to sort for free.

Also found there was a /middleware and a middleware.js in interactions so tidied that up.

int


src/apps/interactions/middleware/collection.js
```diff
@@ -0,0 +1,40 @@
...
+async function getInteractionCollection (req, res, next) {
+  try {
+    res.locals.sortForm = merge({}, interactionSortForm, {
+      hiddenFields: assign({}, omit(req.query, 'sortby')),
+      children: [
+        { value: req.query.sortby },
+      ],
+    })
+
+    const selectedSortBy = req.query.sortby ? { sortby: req.query.sortby } : null
```
teneightfive: I find it sometimes easier to read when variable declarations are at the top of the block, maybe move these lines up?

You could then declare `sortForm` and `interactions` and only set `res.locals` once at the bottom of the block. Might need an `assign` though to only override those properties.

src/apps/contacts/router.js
```diff
@@ -39,7 +40,11 @@ router
 router.post('/:id/archive', archiveContact)
 router.get('/:id/unarchive', unarchiveContact)
 
-router.get('/:contactId/interactions', getCommon, getInteractionCollection, renderInteractions)
+router.get('/:contactId/interactions',
+  getCommon,
+  getInteractionCollection,
+  renderInteractions)
```
teneightfive: Maybe drop the closing bracket on to a new line if you're going to put each middleware on a new line. That's [the pattern](https://github.com/uktrade/data-hub-frontend/blob/develop/src/apps/investment-projects/router.js#L88-L92) that has been used elsewhere when splitting middleware calls onto separate lines.

src/apps/interactions/repos.js
```diff
@@ -58,6 +52,5 @@ module.exports = {
   saveInteraction,
   fetchInteraction,
   getInteractionsForCompany,
-  getInteractionsForContact,
```
teneightfive: Do we no longer associate interactions with a contact?

teneightfive:
Just noticed our heroku review apps aren't working any more...is that something to do with moving to PaaS? Be good to find a way to have demo apps for PRs again as they're very useful for viewing changes before they get merged.

src/apps/interactions/repos.js
ztolley: We do but we use search now instead of the v3 API, and in the rest of the code we call search from middleware

PR #735, opened 2017-10-05
Re-introduce interactions navigation to top level

ztolley:
Some confusion over if this should be taken out or not so it's going back in until told otherwise.


teneightfive:
@ztolley for future reference this could have been done as revert commit to the original change and would've made a clear link to that code.

PR #737, opened 2017-10-06
Refactor interactions render edit page

yeahfro:
This changes focuses around refactor `renderEditPage` for interactions. We ultimately want to mount interactions as a sub app within other apps.

Notes:
- a temporary route has been created within `router.js` of interactions. This demonstrates the setting of `res.locals.companies` `res.locals.returnLink` depending on the app
- I feel uncomfortable about `getInteractionTypeAndService` - any thoughts?

screen shot 2017-10-06 at 09 18 41


src/apps/interactions/middleware/details.js
```diff
@@ -29,6 +35,48 @@ async function postDetails (req, res, next) {
...
+  try {
+    res.locals.interaction = await fetchInteraction(req.session.token, interactionId)
+    res.locals.interactionViewRecord = transformInteractionResponseToViewRecord(res.locals.interaction)
+    next()
+  } catch (err) {
+    next(err)
+  }
+}
+
+async function getContactsForCompanyDetails (req, res, next) {
```
ztolley: Feels like a weird name, Is there a middleware that could be added to companies to get contacts and then call it here?

src/apps/interactions/middleware/details.js
```diff
@@ -29,6 +35,48 @@ async function postDetails (req, res, next) {
...
+  try {
+    res.locals.contacts = await getContactsForCompany(req.session.token, res.locals.company.id)
+    next()
+  } catch (err) {
+    next(err)
+  }
+}
+
+async function getAdviserDetails (req, res, next) {
```
ztolley: maybe call it just getAdvisers and also maybe as a middleware under advisers app?

src/apps/interactions/middleware/details.js
```diff
@@ -29,6 +35,48 @@ async function postDetails (req, res, next) {
...
+  try {
+    res.locals.advisers = await getAdvisers(req.session.token)
+    next()
+  } catch (err) {
+    next(err)
+  }
+}
+
+async function getInteractionTypeAndService (req, res, next) {
```
ztolley: getServicesForInteractionType? dunno

src/apps/interactions/router.js
```diff
@@ -20,9 +26,26 @@ router
...
-router.route(['/create/2', '/:id/edit'])
-  .post(postDetails, renderEditPage)
-  .get(renderEditPage)
+router.route(['/create/2', '/:interactionId/edit'])
+  .post(
+    postDetails,
+    renderEditPage,
+  )
+  .get(
+    async function (req, res, next) {
```
ztolley: You're moving this into a separate file right?

src/apps/interactions/transformers.js
```diff
@@ -72,6 +66,29 @@ function transformInteractionToListItem ({
   }
 }
 
+function transformInteractionResponseToViewRecord ({
```
ztolley: I take this will be replaced by the view transformer I'm about to merge?

src/apps/interactions/middleware/details.js
yeahfro: @ztolley that's what I did yesterday and we talked about premature optimisation!

src/apps/interactions/middleware/details.js
yeahfro: @ztolley that would imply that the `services` are retrieved based upon the `interactionType`

src/apps/interactions/router.js
yeahfro: @ztolley this implementation will be deleted as interactions will no longer be mounted as a root app. This was just an example.

src/apps/interactions/transformers.js
yeahfro: @ztolley yeah I will sort out the merge conflicts shortly

src/apps/interactions/middleware/details.js
yeahfro: @ztolley yes I feel uncomfortable with this and `getAdviserDetails` but I would anticipate we can spot patterns and refactor over time.

src/apps/interactions/router.js
ztolley: Now it's going back into the top level nav then we need to support it both as a sub and root app (though wont have add in root list)

src/apps/interactions/router.js
```diff
@@ -20,9 +26,26 @@ router
...
+      // set company
+      const companyRepo = require('../companies/repos')
+      res.locals.company = await companyRepo.getDitCompany(req.session.token, req.query.company)
+
+      // set return link
+      res.locals.returnLink = `/companies/${req.query.company}/interactions`
```
teneightfive: `returnUrl`? I've used that variable name within OMIS and seems clearer to me.

src/apps/interactions/router.js
teneightfive: I've been thinking about how we handle returning to a particular spot and wonder if we could have a higher level middleware across all entities that accepts a `returnUrl` query string param and returns a successful submit back there and sets the cancel/back url.

Would solve problems like adding a contact in the middle of creating an OMIS order for example.

src/apps/interactions/transformers.js
```diff
@@ -4,32 +4,26 @@ const dateFns = require('date-fns')
...
-    interaction_type: get(interaction_type, 'id'),
-    company: get(company, 'id'),
-    subject,
-    notes,
-    contact: get(contact, 'id'),
-    date: {
-      day: dateFns.format(date, 'DD'),
-      month: dateFns.format(date, 'MM'),
-      year: dateFns.format(date, 'YYYY'),
+function transformInteractionResponseToForm (props = {}) {
```
teneightfive: What's the reason for changing this format? Are we only overriding certain properties with this transformer?

src/apps/interactions/middleware/details.js
jim68000: YAGNI,Y

src/apps/interactions/router.js
jim68000: 👍

yeahfro: @teneightfive not sure, the form builder expects a `returnLink`. Shall we change it everywhere?

src/apps/interactions/controllers/edit.js
```diff
@@ -1,57 +1,43 @@
...
-            interaction_type: formData.interaction_type,
-            company: formData.company,
-          },
-        }
-      ),
-      formData,
-      get(res.locals, 'form.errors.messages'),
-    )
+    const interactionData = transformInteractionResponseToForm(res.locals.interaction)
+    const dit_adviser = interactionData.dit_adviser || get(req, 'session.user')
```
feedmypixel: can this const be camelCase like the otehrs?

src/apps/interactions/controllers/edit.js
```diff
@@ -1,57 +1,43 @@
...
+    const date = interactionData.date || transformDateStringToDateObject(new Date())
```
feedmypixel: I'd give this a better name `interactionDate` maybe?

src/apps/interactions/controllers/edit.js
```diff
@@ -1,57 +1,43 @@
...
+    const mergedData = pickBy(merge({}, interactionData, { dit_adviser, date }, res.locals.requestBody))
+    const advisers = get(res.locals, 'advisers.results')
```
feedmypixel: do these need to be `const`? they are only used in one place Lines `26`, `27`, `28`

src/apps/interactions/controllers/edit.js
```diff
@@ -1,57 +1,43 @@
...
-          },
-        }
-      ),
-      formData,
-      get(res.locals, 'form.errors.messages'),
-    )
+    const interactionData = transformInteractionResponseToForm(res.locals.interaction)
+    const dit_adviser = interactionData.dit_adviser || get(req, 'session.user')
+    const date = interactionData.date || transformDateStringToDateObject(new Date())
+    const mergedData = pickBy(merge({}, interactionData, { dit_adviser, date }, res.locals.requestBody))
```
feedmypixel: rather than naming this by the action (`mergedData`) can it be named by what it actually is?

src/apps/interactions/controllers/edit.js
```diff
@@ -1,57 +1,43 @@
...
+        get(res.locals, 'form.errors.messages'),
+      )
 
     res
-      .breadcrumb(pageTitle)
+      .breadcrumb(`${interactionId ? 'Edit' : 'Add'} interaction for ${res.locals.company.name}`)
       .render('interactions/views/edit', {
-        editInteractionForm,
-        interactionTypeLabel: interaction.interaction_type.name,
+        interactionForm: assign(interactionForm, {
```
feedmypixel: it ok that `interactionForm` is being muted here? Feels a little more functional to add `{}` as first argument. More preference though, so your call.

src/apps/interactions/middleware/details.js
```diff
@@ -1,7 +1,13 @@
...
-
-const { transformInteractionFormBodyToApiRequest } = require('../transformers')
-const { saveInteraction } = require('../repos')
+const {
+  transformInteractionResponseToViewRecord,
+  transformInteractionFormBodyToApiRequest,
+} = require('../transformers')
+const { fetchInteraction, saveInteraction } = require('../repos')
+const metaDataRepository = require('../../../lib/metadata')
+const { find } = require('lodash')
```
feedmypixel: can be combined with the above `require('lodash')`

src/apps/interactions/middleware/details.js
feedmypixel: @ztolley `getAdvisers` is already a method (used within this method) hence it was originally called `getAdviserDetails`

src/apps/interactions/middleware/details.js
feedmypixel: surely `getCompanyDetails`?

src/apps/interactions/middleware/details.js
feedmypixel: `getInteractionDetails`?

src/apps/interactions/repos.js
```diff
@@ -20,6 +20,10 @@ function saveInteraction (token, interaction) {
   return authorisedRequest(token, options)
 }
 
+function fetchInteraction (token, id) {
+  return authorisedRequest(token, `${config.apiRoot}/v3/interaction/${id}`)
```
feedmypixel: the use of `fetch` feels wrong here. We use `get` all over the application. `fetch` to me implies a http call

src/apps/interactions/transformers.js
feedmypixel: I'd also stick with the format we have elsewhere for transformers

src/apps/interactions/transformers.js
```diff
@@ -84,4 +101,5 @@ module.exports = {
   transformInteractionResponseToForm,
   transformInteractionToListItem,
   transformInteractionFormBodyToApiRequest,
+  transformInteractionResponseToViewRecord,
```
feedmypixel: is `Record` needed at the end here?

test/unit/apps/interactions/controllers/edit.test.js
```diff
@@ -1,166 +1,127 @@
...
 
-    context('when success', () => {
-      it('should call createBlankInteractionForCompany for company', async () => {
-        this.req.query.company = 'Hooli Inc'
-        const expectedArgs = ['abcd', { name: 'Spencer' }, 'tel', 'Hooli Inc']
+    it('should render the interaction page with an interaction type label', async () => {
+      await this.controller.renderEditPage(this.req, this.res, this.next)
 
-        await this.edit.renderEditPage(this.req, this.res, this.nextSpy)
+      const actual = this.res.render.getCall(0).args[1].interactionForm.interactionTypeLabel
```
feedmypixel: I don't feel testing the labels here is the right thing to be testing in a unit test. 

IMO you should unit test the transformer to make sure it gives you back what is expected. You should unit test the controller to make sure it goes in the right direction. I don't feel unit testing the controllers and the values that the controllers passes on gives us much value here. It does give us a lot of maintenance.

- Controllers should just be "this goes in the direction I expect it to" by testing `next()`/`render()`
- Transformers can be tested by saying "I give the transformer this and I expect this"

This means we don't then test controllers for all the different values being sent through the one transformer.

As always open to discussion on this

feedmypixel:
@yeahfro There have a been a lot of tests removed in this PR. Can we wait to remove tests until we have had the chat on Thursday. I want to make sure we are all on the same page before we start removing swathes of the tests.

test/unit/apps/interactions/controllers/edit.test.js
ztolley: But I've always thought we should test that the controller returns the result of a transformation, to make sure it calls it and that it sends it back.

I think if we move to a integration approach then this would an end to this discussion :)

src/apps/interactions/transformers.js
ztolley: I think that's the naming convention we've followed recently after being suggested by Tyom

src/apps/interactions/middleware/details.js
yeahfro: @feedmypixel this is only getting the contacts for the company

src/apps/interactions/router.js
ztolley: Consistency is good

src/apps/interactions/middleware/details.js
yeahfro: @feedmypixel there is already a `getInteractionDetails` which is used by the router for param `interactionId`. This function needs to be added as middleware even when there is not an `interactionId`

feedmypixel:
> @yeahfro There have a been a lot of tests removed in this PR. Can we wait to remove tests until we have had the chat on Thursday. I want to make sure we are all on the same page before we start removing swathes of the tests.

sorry to clarify, if they need removing as part of this work go for it. If they are being removed re: integration/ testing chat lets hold off till then

test/unit/apps/interactions/controllers/edit.test.js
feedmypixel: > But I've always thought we should test that the controller returns the result of a transformation, to make sure it calls it and that it sends it back.

And that sir is an integration test. Which yes we should be doing but in integration tests not in unit tests. TBH (more) 😉

ztolley: I test I call something and that I used the correct thing. I know we disagree on this but it's fine, we'll move to integration tests and agree on the pattern there.

src/apps/interactions/router.js
ztolley: I'm still torn between going to 
`/interactions/:interactionId` and `/contact/:contactId/interaction/:interactionId`

Are we trying to make it harder than it needs to be?

Should we not just go with the former, and use something like return URL for a back button? Even add a returnTitle so you would have a link 'Return to contact' ?

feedmypixel: @ztolley think of it as a signpost:
`/interactions/:interactionId`
`/contact/:contactId/interaction/:interactionId` (edited)
which 1 out of the above 2 tells you what it is and where it is? which 1 just tells you what it is?

ztolley: Users don't care about url's though

feedmypixel: > Users don't care about url's though
@ztolley agreed, but that is a different issue to the one we are discussing here.

src/apps/interactions/repos.js
```diff
@@ -1,7 +1,7 @@
 const config = require('../../../config')
 const authorisedRequest = require('../../lib/authorised-request')
 
-function fetchInteraction (token, interactionId) {
+function getInteraction (token, interactionId) {
```
teneightfive: Maybe revert this change for now until we come up with a pattern that we replicate across the codebase. Feels like its adding noise to this change.

PR #743, opened 2017-10-09
Add new step1 for interactions

ztolley:
step1

Changes interaction creation so that Step 1 selects the kind and the edit page asks for the communication type.

- [x] Create interstitial page to select 'kind' of interaction

-- Next Step will be to update the interaction form to support kind & interaction type now it has been removed from this screen


src/apps/interactions/controllers/step1.js
```diff
@@ -0,0 +1,38 @@
+const { pick, lowerCase } = require('lodash')
+const queryString = require('query-string')
+const { selectKindFormConfig } = require('../macros')
+
+function postStep1 (req, res, next) {
```
teneightfive: I think this file and maybe this controller could be more specific to the create journey. `step1` doesn't say what journey it is the first step of. 

Maybe `create-1` for filename?

src/apps/interactions/controllers/step1.js
```diff
@@ -0,0 +1,38 @@
...
+  const kind = req.body.kind || null
+
+  if (!kind) {
+    res.locals.errors = {
+      messages: {
+        kind: ['You must select an interaction type'],
```
teneightfive: `kind` feels really weird. I know its coming from the API but I would have thought `type` would have been more appropriate.

src/apps/interactions/controllers/step1.js
```diff
@@ -0,0 +1,38 @@
...
+}
+
+function renderStep1 (req, res) {
+  const selectKindForm = selectKindFormConfig({
+    errors: res.locals.errors || [],
+  })
+
+  res
+    .breadcrumb('Add interaction/service')
+    .render('interactions/views/add-step-1.njk', {
```
teneightfive: Feels like a lack of consistency in our naming for this. `step1`, `add-step-1`, `/create`.

Maybe we use `create` which we've used elsewhere in the app for the create steps and maybe the steps more specific like `create-step-1`.

src/apps/interactions/controllers/step1.js
```diff
@@ -0,0 +1,38 @@
...
+  return res.redirect(`/interactions/create/${lowerCase(kind)}?${createQueryString}`)
+}
+
+function renderStep1 (req, res) {
+  const selectKindForm = selectKindFormConfig({
+    errors: res.locals.errors || [],
+  })
+
+  res
+    .breadcrumb('Add interaction/service')
```
teneightfive: Can we not handle this better? Putting `interaction/service` feels lazy. It's like saying `Add/edit`.

src/apps/interactions/router.js
```diff
@@ -22,11 +23,11 @@ router.param('interactionId', getInteractionDetails)
 router.get('/', setDefaultQuery(DEFAULT_COLLECTION_QUERY), getInteractionCollection, renderInteractionList)
 
 router
-  .route(/^\/create(\/1)?$/)
-  .post(postAddStep1)
-  .all(renderCreatePage)
+  .route('/create')
```
teneightfive: Do we not have multiple steps any more?

src/apps/interactions/controllers/step1.js
ztolley: maybe type would get confused with the old interaction type, which is being removed. I'd like to stick with kind for now and raise the question the datahub dev channel

src/apps/interactions/controllers/step1.js
ztolley: Suggestions?

src/apps/interactions/controllers/step1.js
ztolley: I agree create-step-1

src/apps/interactions/router.js
ztolley: No. You'd go from step 1 right into the edit form for interaction or service delivery, and communications channel becomes radio or dropdown in that form.

ztolley: also, if coming from investment projects you bypass step 1 as investment projects can't create service deliveries

src/apps/interactions/macros.js
```diff
@@ -29,6 +29,33 @@ const interactionSortForm = {
   ],
 }
 
+const selectKindFormConfig = function ({
+  returnLink,
+  errors = [],
+}) {
+  return {
+    buttonText: 'Continue',
+    errors: errors,
```
teneightfive: Can use shorthand here: `errors,`

Also move to first line as its easier to scan when shorthand declarations are first.

src/apps/interactions/controllers/create.js
```diff
@@ -0,0 +1,38 @@
...
+  return res.redirect(`/interactions/create/${lowerCase(kind)}?${createQueryString}`)
+}
+
+function renderCreate (req, res) {
+  const selectKindForm = selectKindFormConfig({
+    errors: res.locals.errors || [],
+  })
+
+  res
+    .breadcrumb('Add interaction/service')
```
teneightfive: Is there no way we know the difference between when it is an interaction or a service (delivery)?

teneightfive: If we really don't it would be better to use the word `or` rather than a `/`.

ztolley: This the screen that picks which to create and the options describe the options in more detail. I'll change it to 'or'

PR #745, opened 2017-10-09
Update investment project interactions

ztolley:
Switched investment project to use interaction middleware, like company and contact, and enabled sorting.


src/apps/investment-projects/router.js
```diff
@@ -31,6 +31,7 @@ const {
 const { renderInvestmentList } = require('./controllers/list')
 const { getInvestmentProjectsCollection, getRequestBody } = require('./middleware/collection')
 
+const { getInteractionCollection } = require('../interactions/middleware/collection')
```
teneightfive: As mentioned previously I think this might be extra work and will need to be changed when we mount the interactions app within this app.

We can mount that app using `router.use('/interactions', interactionsApp)`

And we can potentially use app specific middleware to determine the list like @yeahfro started to demonstrate.

ztolley: so so we put this on hold or merge this and then replace with the routing when @yeahfro does it?

PR #746, opened 2017-10-09
Add interactions create routes to companies

yeahfro:
The temporary routes in `/interactions/router.js` are becoming increasingly difficult to manage.

This change continues the move towards interactions being a sub app by adding the create routes to `/companies/router.js`


ztolley:
give me a shout on slack, refactoring interaction edit form to fix some bugs and simplify, I want to show you what I have

src/apps/interactions/subApp.js
```diff
@@ -0,0 +1,34 @@
+const { renderEditPage } = require('../interactions/controllers/edit')
+const { renderCreatePage, postAddStep1 } = require('./controllers/details')
+const { postDetails, getInteractionOptions } = require('../interactions/middleware/details')
+
+function configRouter (router, routeBase, setReturnUrl) {
```
ztolley: mixed case filenames?

src/apps/companies/router.js
```diff
@@ -77,4 +79,6 @@ router.get('/:companyId/exports', renderExports)
 router.get('/:companyId/investments', renderInvestments)
 router.get('/:companyId/audit', renderAuditLog)
 
+configInteractionsRouter(router, '/:companyId', setInteractionsReturnUrl)
```
teneightfive: This could be `app.use('/:companyId', setInteractionsReturnUrl, interactionsSubApp)` instead.

Then just format the file at `interactionsSubApp` as a normal router. It should nest the URLs correctly.

src/apps/interactions/subApp.js
teneightfive: Don't think we need to create a function that accepts the router. This file could be formatted as the other routers which exports the `router` and declares the nested URLs it needs on that.

src/apps/companies/router.js
```diff
@@ -77,4 +79,6 @@ router.get('/:companyId/exports', renderExports)
 router.get('/:companyId/investments', renderInvestments)
 router.get('/:companyId/audit', renderAuditLog)
 
+router.use(interactionsSubApp('/:companyId', setInteractionsReturnUrl))
```
teneightfive: I think this is still more complicated than it needs to be. It could be something like:

```
const interactionsRouter = require('../interactions/router.sub-app')
...
router.use('/:companyId/interactions', setInteractionsReturnUrl, interactionsRouter)
```

src/apps/companies/middleware/details.js
```diff
@@ -0,0 +1,14 @@
+function setInteractionsReturnUrl (req, res, next) {
```
ztolley: is details the best name?

src/apps/companies/router.js
```diff
@@ -72,9 +75,11 @@ router.use('/:companyId', setLocalNav(LOCAL_NAV))
 router.get('/:companyId', redirectToFirstNavItem)
 router.get('/:companyId/details', renderDetails)
 router.get('/:companyId/contacts', renderContacts)
-router.get('/:companyId/interactions', getInteractionCollection, renderInteractions)
+router.get('/:companyId/interactions', setInteractionsReturnUrl, getInteractionCollection, renderInteractions)
```
ztolley: can this be in the sub-app router?

src/apps/interactions/subApp.js
```diff
@@ -0,0 +1,37 @@
...
+const { renderEditPage } = require('./controllers/edit')
+const { renderDetailsPage } = require('./controllers/details')
+const { postCreate, renderCreate } = require('./controllers/create')
+
+const { postDetails, getInteractionOptions, getInteractionDetails } = require('./middleware/details')
+
+router.param('interactionId', getInteractionDetails)
+
+router
+  .route('/interactions/create/1')
```
ztolley: There is no create 1 and 2, there just `/create` then `create/interaction` `create/service-delivery`. In the new flow, you select the `kind` in step 1 and then the communication channel is a drop-down in the edit page. Alternatively investment projects will only go straight to `/create/interaction` as investment projects can't create service deliveries.

src/apps/companies/middleware/details.js
yeahfro: @ztolley No, open to suggestions..

src/apps/companies/router.js
yeahfro: @ztolley Yes I thought so too but didn't want to make this PR any bigger

ztolley: ok. The sorted investment interaction list PR got turned down because Dom assumed this would replace it. I still have the branch if we want to do it that way or we change the lists in a follow up PR to be consistent with details and edit.

src/apps/companies/middleware/details.js
ztolley: The only suggestion I had was 'options', otherwise, you get stuck in lots of little middlewares. or I like the idea you mentioned before.

src/apps/companies/middleware/details.js
```diff
@@ -0,0 +1,14 @@
...
+  res.locals.returnLink = `/companies/${req.params.companyId}/interactions/`
+  next()
+}
+
+function setInteractionsEntityName (req, res, next) {
+  res.locals.entityName = res.locals.company.name
```
teneightfive: Could the view not just use `company.name` if it is already being set on `res.locals`?

src/apps/companies/router.js
teneightfive: Yeah I had thought this would be in the sub-app and middleware at company/contact router level determines the list of items. But happy for this to be in a follow up PR.

src/apps/companies/middleware/details.js
teneightfive: Ah. I see why now after seeing the contacts changes...

src/apps/interactions/controllers/edit.js
```diff
@@ -28,7 +28,7 @@ function renderEditPage (req, res) {
 
   res
     .breadcrumb(`${interactionData ? 'Edit' : 'Add'} interaction`)
-    .title(`${interactionData ? 'Edit' : 'Add'} interaction for ${res.locals.company.name}`)
+    .title(`${interactionData ? 'Edit' : 'Add'} interaction for ${res.locals.entityName}`)
```
teneightfive: I wonder if we need `entityName` in the title now that it will be nested. Essentially the breadcrumb item before this will say the company/contact name won't it so it might not be needed?

src/apps/interactions/transformers.js
```diff
@@ -137,9 +137,16 @@ function transformInteractionFormBodyToApiRequest ({ props, company, communicati
   })
 }
 
+function transformInteractionListItemToHaveUrlPrefix (returnLink) {
```
teneightfive: Not sure I fully understand the need for this. Maybe we can go through it tomorrow morning in person, might be easier to show in person.

src/apps/interactions/middleware/details.js
```diff
@@ -8,15 +8,15 @@ const { getAdvisers } = require('../../adviser/repos')
 async function postDetails (req, res, next) {
   res.locals.requestBody = transformInteractionFormBodyToApiRequest({
     props: req.body,
-    company: req.query.company,
-    communicationChannel: req.query.interaction_type,
+    company: res.locals.company.id,
```
teneightfive: Should this be `companyId` instead of `company`?

src/apps/companies/middleware/details.js
```diff
@@ -0,0 +1,14 @@
+function setInteractionsReturnUrl (req, res, next) {
+  res.locals.returnLink = `/companies/${req.params.companyId}/interactions/`
```
teneightfive: I wonder if we can make use of `req.baseUrl` instead of having to manually set this...maybe one for a follow up PR though.

src/apps/interactions/middleware/details.js
```diff
@@ -8,15 +8,15 @@ const { getAdvisers } = require('../../adviser/repos')
...
+    communicationChannel: res.locals.interactionType.id,
   })
 
   try {
     const result = await saveInteraction(req.session.token, res.locals.requestBody)
 
     req.flash('success', `Interaction ${res.locals.interaction ? 'updated' : 'created'}`)
-    return res.redirect(`/interactions/${result.id}`)
+    return res.redirect(`${res.locals.returnLink}${result.id}`)
```
teneightfive: Could this just be a relative redirect that doesn't need to include the `returnLink` local?

src/apps/companies/router.js
ztolley: So can we use mine for now and then replace with the new code later?

ztolley: Just so I know what to do with the Jira card :/

src/apps/interactions/middleware/details.js
ztolley: It's company in the model, but I think we need to completely remove this and add company and project as hidden fields in the form as this version doesn't work if editing an interaction from anywhere other than the company app, so it's broken.

src/apps/contacts/middleware/details.js
```diff
@@ -0,0 +1,27 @@
+const { getContact } = require('../../contacts/repos')
+
+function setInteractionsReturnUrl (req, res, next) {
+  res.locals.returnLink = `/contacts/${req.params.contactId}/interactions/`
+  next()
+}
+
+function setInteractionsEntityName (req, res, next) {
+  res.locals.entityName = `${res.locals.contact.first_name} ${res.locals.contact.last_name}`
```
feedmypixel: do we need a `get` here?

src/apps/contacts/middleware/details.js
```diff
@@ -0,0 +1,27 @@
+const { getContact } = require('../../contacts/repos')
```
feedmypixel: feels like there is a lot of abstraction in this file, is there a reason for this abstraction? Could it not all sit in one middleware method `getDetails()`? Essentially you are setting up a load of `res.locals`

src/apps/interactions/middleware/collection.js
```diff
@@ -28,6 +28,7 @@ async function getInteractionCollection (req, res, next) {
       .then(transformApiResponseToSearchCollection(
         { query: req.query },
         transformInteractionToListItem,
+        transformInteractionListItemToHaveUrlPrefix(res.locals.returnLink),
```
feedmypixel: not sure why this is being done, would the `returnLink` not already be a fully working url?

src/apps/interactions/middleware/details.js
feedmypixel: does this need to be a template literal?

src/apps/interactions/middleware/details.js
```diff
@@ -40,28 +40,14 @@ async function getInteractionDetails (req, res, next, interactionId) {
...
-  try {
-    res.locals.contacts = await getContactsForCompany(req.session.token, res.locals.company.id)
-    next()
-  } catch (err) {
-    next(err)
-  }
-}
-
-async function getAdviserDetails (req, res, next) {
+async function getInteractionOptions (req, res, next) {
```
feedmypixel: We should stick to one convention `Options` || `Details`

src/apps/interactions/controllers/edit.js
teneightfive: Maybe do this a follow up as it would involve moving where the breadcrumb is set for companies.

src/apps/interactions/transformers.js
feedmypixel: agreed - not sure if there is a need for this?

src/apps/investment-projects/middleware/details.js
```diff
@@ -0,0 +1,27 @@
+const { getInvestment } = require('../repos')
+
+function setInteractionsReturnUrl (req, res, next) {
```
feedmypixel: Same comment above on the contacts details. Feels like there is a lot of abstraction here maybe just a `getDetails()` method would suffice? Probably will be clearer once you can get a 1000ft view on it.

src/apps/interactions/transformers.js
teneightfive: Makes more sense now we've discussed. But maybe change `returnLink` to `urlPrefix` and check the first character is a slash before removing it.

src/apps/contacts/middleware/details.js
yeahfro: There will be other PRs, it's likely that `setInteractionsReturnUrl` and `setInteractionsEntityName` will become redundant.

src/apps/investment-projects/middleware/details.js
yeahfro: There will be other PRs, it's likely that `setInteractionsReturnUrl` and `setInteractionsEntityName` will become redundant.

src/apps/contacts/middleware/details.js
yeahfro: The `contact` is always populated before this middleware runs

PR #748, opened 2017-10-09
Introduce service delivery form config

yeahfro:
This changes introduces the service delivery form config which displays an edit form when `kind` is `service_delivery`.

screen shot 2017-10-09 at 22 24 11

### DIFF
--- a/src/apps/interactions/controllers/edit.js
+++ b/src/apps/interactions/controllers/edit.js
@@ -1,10 +1,14 @@
 /* eslint camelcase: 0 */
-const { get, merge, pickBy } = require('lodash')
+const { get, merge, pickBy, lowerCase, snakeCase } = require('lodash')
 
 const { transformInteractionResponseToForm } = require('../transformers')
 const { transformDateStringToDateObject } = require('../../transformers')
-const { interactionEditFormConfig } = require('../macros')
+const { interactionFormConfig, serviceDeliveryFormConfig } = require('../macros')
 const { buildFormWithStateAndErrors } = require('../../builders')
+const formConfigs = {
+  'interaction': interactionFormConfig,
+  'service-delivery': serviceDeliveryFormConfig,
+}
 
 function renderEditPage (req, res) {
   const interactionData = transformInteractionResponseToForm(res.locals.interaction)
@@ -16,16 +20,18 @@ function renderEditPage (req, res) {
   const mergedInteractionData = pickBy(merge({}, interactionDefaults, interactionData, res.locals.requestBody))
   const interactionForm =
     buildFormWithStateAndErrors(
-      interactionEditFormConfig(
+      formConfigs[req.params.kind](
         {
           returnLink: res.locals.returnLink,
           advisers: get(res.locals, 'advisers.results'),
           contacts: res.locals.contacts,
           services: res.locals.services,
+          events: get(res.locals, 'events.results'),
           hiddenFields: {
             id: get(res.locals, 'interaction.id'),
             company: res.locals.company.id,
             investment_project: get(res.locals, 'investmentData.id'),
+            kind: snakeCase(req.params.kind),
           },
         }),
       mergedInteractionData,
@@ -35,8 +41,8 @@ function renderEditPage (req, res) {
   const forEntityName = res.locals.entityName ? ` for ${res.locals.entityName}` : ''
 
   res
-    .breadcrumb(`${interactionData ? 'Edit' : 'Add'} interaction`)
-    .title(`${interactionData ? 'Edit' : 'Add'} interaction${forEntityName}`)
+    .breadcrumb(`${interactionData ? 'Edit' : 'Add'} ${lowerCase(req.params.kind)}`)
+    .title(`${interactionData ? 'Edit' : 'Add'} ${lowerCase(req.params.kind) + forEntityName}`)
     .render('interactions/views/edit', {
       interactionForm,
     })
--- a/src/apps/interactions/macros.js
+++ b/src/apps/interactions/macros.js
@@ -58,7 +58,69 @@ const selectKindFormConfig = function ({
   }
 }
 
-const interactionEditFormConfig = function ({
+const interactionFields = {
+  contact (contacts) {
+    return {
+      macroName: 'MultipleChoiceField',
+      name: 'contact',
+      initialOption: '-- Select contact --',
+      options () {
+        return contacts.map(transformContactToOption)
+      },
+    }
+  },
+  provider: {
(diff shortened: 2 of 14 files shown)


src/apps/interactions/macros.js
```diff
@@ -29,7 +30,68 @@ const interactionSortForm = {
...
+  notes: {
+    macroName: 'TextField',
+    type: 'textarea',
+    name: 'notes',
+    label: formLabels.notes,
+  },
+  date: {
+    macroName: 'DateFieldset',
+    name: 'date',
+    label: formLabels.date,
```
feedmypixel: label wise we could follow the map pattern shown [here](https://github.com/uktrade/data-hub-frontend/blob/b3f8c3afe960cdfa13adb71f978ce2139198aaa4/src/apps/events/macros/event-filters-fields.js#L48)

src/apps/interactions/macros.js
```diff
@@ -40,64 +102,76 @@ const interactionEditFormConfig = function ({
...
+          },
+        ],
       },
       {
         macroName: 'MultipleChoiceField',
-        name: 'service',
-        label: formLabels.service,
-        initialOption: '-- Select service --',
+        name: 'event',
+        label: 'Event',
```
feedmypixel: can this go into a `labels.js` file?

src/apps/interactions/macros.js
```diff
@@ -40,64 +102,76 @@ const interactionEditFormConfig = function ({
...
+        condition: {
+          name: 'is_event',
+          value: 'true',
         },
       },
+      interactionFields.service(services),
+      interactionFields.subject,
+      interactionFields.notes,
+      assign({}, interactionFields.date, {
+        label: 'Date of service delivery',
```
feedmypixel: can this go into a `labels.js` file somewhere?

feedmypixel: label wise we could follow the map pattern shown [here](https://github.com/uktrade/data-hub-frontend/blob/b3f8c3afe960cdfa13adb71f978ce2139198aaa4/src/apps/events/macros/event-filters-fields.js#L48)

src/apps/interactions/router.js
```diff
@@ -34,16 +35,27 @@ router.route(['/create/2', '/:interactionId/edit'])
   .get(
     async function (req, res, next) {
       // set company
-      const companyRepo = require('../companies/repos')
-      res.locals.company = await companyRepo.getDitCompany(req.session.token, req.query.company)
-
-      // set return link
-      res.locals.returnLink = `/companies/${req.query.company}/interactions`
+      if (req.query.company) {
+        const companyRepo = require('../companies/repos')
```
feedmypixel: do we need the requires here? No benefit to requiring them in an if, just require them at the top and use them as needed.

src/apps/interactions/router.js
```diff
@@ -34,16 +35,27 @@ router.route(['/create/2', '/:interactionId/edit'])
   .get(
     async function (req, res, next) {
       // set company
-      const companyRepo = require('../companies/repos')
-      res.locals.company = await companyRepo.getDitCompany(req.session.token, req.query.company)
-
-      // set return link
-      res.locals.returnLink = `/companies/${req.query.company}/interactions`
+      if (req.query.company) {
```
feedmypixel: Feels like there is some tidy up to do here once everything has been done. It also feels like this may be a good use case to abstract out to a few methods like:

`res.locals.returnLink = getReturnLink(req, res)`
`res.locals.company = getCompany(req, res)`

Then we can test them in isolation. Maybe not for this PR though.

test/unit/apps/interactions/macros.test.js
```diff
@@ -22,41 +22,49 @@ describe('Interaction macros', () => {
     })
   })
 
-  describe('#interactionEditFormConfig', () => {
+  describe('#interactionFormConfig', () => {
     it('should return object with specified returnLink present', () => {
-      const actual = this.macros.interactionEditFormConfig({
-        returnLink: 'xxx.yyy.zzz',
-      })
+      const returnLink = 'xxx.yyy.zzz'
```
feedmypixel: would be good to have test data in the format that is closer to actual data `/example/return/link` for instance

test/unit/apps/interactions/macros.test.js
```diff
@@ -22,41 +22,49 @@ describe('Interaction macros', () => {
...
       actual.children.find(x => x.name === 'dit_team').options()
 
       expect(this.transformObjectToOptionSpy).to.have.callCount(2)
     })
   })
+
+  describe('#serviceDeliveryFormConfig', () => {
+    it('should return object with transformed event options', () => {
+      const actual = this.macros.serviceDeliveryFormConfig({ events: sampleOptions })
+      actual.children.find(x => x.name === 'event').options()
```
feedmypixel: whats this doing?

src/apps/interactions/macros.js
```diff
@@ -40,64 +102,76 @@ const interactionEditFormConfig = function ({
...
-      },
-      {
-        macroName: 'MultipleChoiceField',
-        name: 'dit_team',
-        label: formLabels.dit_team,
-        initialOption: '-- Select provider --',
-        options () {
-          return metaData.teams.map(transformObjectToOption)
-        },
+        type: 'radio',
```
feedmypixel: Might be worth adding a `TODO` that this will be going once interactions are within events.

feedmypixel:
Re-firing CircleCi tests

feedmypixel:
💥

src/apps/interactions/controllers/edit.js
```diff
@@ -1,10 +1,14 @@
...
-const { get, merge, pickBy } = require('lodash')
+const { get, merge, pickBy, lowerCase, snakeCase } = require('lodash')
 
 const { transformInteractionResponseToForm } = require('../transformers')
 const { transformDateStringToDateObject } = require('../../transformers')
-const { interactionEditFormConfig } = require('../macros')
+const { interactionFormConfig, serviceDeliveryFormConfig } = require('../macros')
 const { buildFormWithStateAndErrors } = require('../../builders')
+const formConfigs = {
+  'interaction': interactionFormConfig,
```
feedmypixel: doesn't stricktly need quotes

src/apps/interactions/middleware/details.js
```diff
@@ -1,18 +1,20 @@
 const { assign } = require('lodash')
+const Case = require('case')
```
feedmypixel: would this work? `const { sentence } = require('case')`

PR #753, opened 2017-10-10
Add communication channel to interaction

ztolley:
- Adds the 'communication_channel' to the interaction form
- Removes the use of interaction type when posting an interaction
- No longer fetches interaction types for the edit form

### DIFF
--- a/src/apps/interactions/macros.js
+++ b/src/apps/interactions/macros.js
@@ -1,3 +1,5 @@
+const { assign } = require('lodash')
+
 const formLabels = require('./labels')
 const metaData = require('../../lib/metadata')
 const { transformContactToOption, transformObjectToOption } = require('../transformers')
@@ -70,7 +72,6 @@ const interactionEditFormConfig = function ({
       {
         macroName: 'MultipleChoiceField',
         name: 'contact',
-        label: formLabels.contact,
         initialOption: '-- Select contact --',
         options () {
           return contacts.map(transformContactToOption)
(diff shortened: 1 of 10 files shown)


src/apps/interactions/transformers.js
```diff
@@ -129,11 +129,10 @@ function transformInteractionResponseToViewRecord ({
   return viewRecord
 }
 
-function transformInteractionFormBodyToApiRequest ({ props, company, communicationChannel }) {
+function transformInteractionFormBodyToApiRequest ({ props, company }) {
   return assign({}, props, {
     date: transformDateObjectToDateString('date')(props),
-    company: company,
-    communication_channel: communicationChannel,
+    company,
```
teneightfive: Move up one line? Easier to read when shorthand declarations are first.

src/apps/interactions/macros.js
```diff
@@ -120,6 +120,15 @@ const interactionEditFormConfig = function ({
           return advisers.map(transformContactToOption)
         },
       },
+      {
+        macroName: 'MultipleChoiceField',
+        name: 'communication_channel',
+        label: formLabels.communication_channel,
```
feedmypixel: we can just add labels for all in a map like [here](https://github.com/uktrade/data-hub-frontend/blob/b3f8c3afe960cdfa13adb71f978ce2139198aaa4/src/apps/events/macros/event-filters-fields.js#L48)?

src/apps/interactions/middleware/details.js
```diff
@@ -1,4 +1,4 @@
-const { assign, find } = require('lodash')
+const { assign } = require('lodash')
```
feedmypixel: This is something we have elsewhere, can we keep a hard return here to define between external requires and internal ones.

src/apps/interactions/macros.js
ztolley: was trying to keep the pr small and focussed just on adding this field.

PR #756, opened 2017-10-11
Fix issues around editing interactions

yeahfro:
This change will fix the majority of issues around editing interactions, including:
- Using data set on `res.locals` rather than depending on the query string
- Displaying all data when editing
- Adding investment project ID for POST to API when editing an investment project interaction
- Selecting contact when adding an interaction for contact

Not covered:
- Editing interactions from the top-level `Interactions` list view

### DIFF
--- a/src/apps/interactions/transformers.js
+++ b/src/apps/interactions/transformers.js
@@ -134,9 +134,8 @@ function transformInteractionResponseToViewRecord ({
   return viewRecord
 }
 
-function transformInteractionFormBodyToApiRequest ({ props, company }) {
+function transformInteractionFormBodyToApiRequest (props) {
   return assign({}, props, {
-    company,
     date: transformDateObjectToDateString('date')(props),
   })
 }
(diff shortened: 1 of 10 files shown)


src/apps/interactions/transformers.js
```diff
@@ -129,10 +134,15 @@ function transformInteractionResponseToViewRecord ({
   return viewRecord
 }
 
-function transformInteractionFormBodyToApiRequest ({ props, company }) {
+function transformInteractionFormBodyToApiRequest ({
+  props,
```
ztolley: doesn't it get the company and investment from the form body? what about contact?

yeahfro: `contact` is on the form body but `company` and `investment` is not

ztolley: why not? wouldn't it be easier and more consistent to add them when creating the initial default values for the investment form?

ztolley:
Needs to default the Adviser based on the current user to replicate current functionality

PR #757, opened 2017-10-11
Some protection against displaying legacy OMIS orders

teneightfive:
This makes a few tweaks to work order and quote views to protect against legacy orders and certain states.

See list of commits for changes and explanations.


src/apps/omis/apps/view/middleware.js
```diff
@@ -59,6 +59,8 @@ async function getQuote (req, res, next) {
       return next()
     }
 
+    // console.log(error.statusCode)
```
yeahfro: Really? This is worse than my `console.log`

teneightfive: *&@$ 😳

PR #761, opened 2017-10-12
Refactor quote setting logic for OMIS orders

teneightfive:
This makes the quote logic more logical so that the app only makes
requests when it needs to.

This change also makes handling legacy orders more robust but not
throwing an error when expected cases occur.

### DIFF
--- a/src/apps/omis/apps/view/middleware.js
+++ b/src/apps/omis/apps/view/middleware.js
@@ -1,7 +1,8 @@
-const { get, filter, mapValues, pickBy } = require('lodash')
+const { assign, get, filter, mapValues, pickBy } = require('lodash')
 const path = require('path')
 const i18nFuture = require('i18n-future')
 
+const logger = require('../../../../../config/logger')
 const { Order } = require('../../models')
 const { getCompany } = require('../../middleware')
 const editSteps = require('../edit/steps')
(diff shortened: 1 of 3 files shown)


src/apps/omis/apps/view/middleware.js
```diff
@@ -16,30 +17,37 @@ function setTranslation (req, res, next) {
...
+      })
+    } catch (error) {
+      logger.error(error)
+    }
+  }
+
+  next()
+}
+
+async function setQuote (req, res, next) {
```
yeahfro: My initial thoughts on this function are that it's quite hairy

ztolley: Maybe break up the function following the rule 'do one thing and do it well'

src/apps/omis/apps/view/router.js
```diff
@@ -18,19 +20,20 @@ const LOCAL_NAV = [
...
 
-router.use(setLocalNav(LOCAL_NAV))
-router.use(setTranslation)
-router.use(setOrderBreadcrumb)
-
 router.use((req, res, next) => {
   getCompany(req, res, next, res.locals.order.company.id)
 })
 
+router.use(setLocalNav(LOCAL_NAV))
```
ztolley: lots of middlewares. Do they get used in other places?

src/apps/omis/apps/view/middleware.js
teneightfive: I split quote and quote preview so they are now more decoupled after a chat with Lee. Not sure it can be split much more :/

src/apps/omis/apps/view/router.js
teneightfive: They're all used within this sub-app and shared across all the routes within the view state.

PR #767, opened 2017-10-13
Delete redundant service deliveries app

yeahfro:
Now that service deliveries are covered as a `kind` of interaction, we no longer need the `service-deliveries` app. This change simply deletes the app and related tests.


reupen:
Need to get rid of service-delivery-status and event from the metadata repo as well. They were just used for the old service deliveries.

PR #768, opened 2017-10-14
Use investmentId as parameter name instead of id

ztolley:
In order to add a field to the investment summary form I first need to refactor the investment project code to use the form macros and the middleware pattern found in events and interactions.

This is the first step in that direction, renaming the parameter used in the investment project to `investmentId` instead of the generic name `id`

There are still places in investment that have inconsistent naming for the parameter, these will slow be refactored out in upcoming PR's

### DIFF
--- a/src/apps/investment-projects/middleware/shared.js
+++ b/src/apps/investment-projects/middleware/shared.js
@@ -22,13 +22,13 @@ function getCompanyDetails (req, res, next) {
     .catch(next)
 }
 
-async function getInvestmentDetails (req, res, next, id = req.params.id) {
-  if (!isValidGuid(id)) {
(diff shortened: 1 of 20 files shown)


src/apps/investment-projects/middleware/shared.js
```diff
@@ -22,13 +22,13 @@ function getCompanyDetails (req, res, next) {
     .catch(next)
 }
 
-async function getInvestmentDetails (req, res, next, id = req.params.id) {
-  if (!isValidGuid(id)) {
+async function getInvestmentDetails (req, res, next, investmentId = req.params.investmentId) {
```
teneightfive: I don't think you actually need to do `investmentId = req.params.investmentId`. I think it will work like this:

```js
async function getInvestmentDetails (req, res, next, investmentId) {
  ...
}
```

PR #769, opened 2017-10-16
DH-911  bug fixes

yeahfro:
This pull request addresses three issues:

1. Get all events rather than the first 100
2. Display "No" for editing an event when an event has not been selected
3. Set `is_event` to `true` or `false` strings depending if an event has been selected or not

### DIFF
--- a/src/apps/events/repos.js
+++ b/src/apps/events/repos.js
@@ -24,8 +24,13 @@ function getEvents (token) {
   return authorisedRequest(token, `${config.apiRoot}/v3/event`)
 }
 
+function getAllEvents (token) {
+  return authorisedRequest(token, `${config.apiRoot}/v3/event?limit=100000&offset=0`)
+}
+
 module.exports = {
   saveEvent,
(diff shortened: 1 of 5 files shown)


src/apps/events/repos.js
```diff
@@ -24,8 +24,13 @@ function getEvents (token) {
   return authorisedRequest(token, `${config.apiRoot}/v3/event`)
 }
 
+function getAllEvents (token) {
+  return authorisedRequest(token, `${config.apiRoot}/v3/event?limit=100000&offset=0`)
```
jim68000: Can we add a debt ticket somewhere that says we need an unbounded list of events from the API? (Also, as there are 26300 events in the DH migration data 10000 might not be enough!)

yeahfro: Note that the `limit` parameter is set to `100000`.

[I have created a technical debt issue](https://github.com/uktrade/data-hub-frontend/issues/770).

PR #773, opened 2017-10-16
Remove references to interaction type

yeahfro:
This change removes all references to interaction type as it's no longer available from the API.


src/apps/investment-projects/labels.js
```diff
@@ -58,7 +58,6 @@ const labels = {
   },
   interactionsLabels: {
```
ztolley: Think you can get rid of the entire section here

yeahfro: I have done that - unit tests still pass so I assume it is all good.

PR #775, opened 2017-10-17
Add contact to summary

ztolley:
add contact

Add the ability to add multiple client contacts to an investment project through editing the summary.

Also, fixes an issue not showing the investment contact value on the details page.

### DIFF
--- a/src/apps/investment-projects/controllers/edit.js
+++ b/src/apps/investment-projects/controllers/edit.js
@@ -16,9 +16,9 @@ function editRequirementsGet (req, res) {
   res.render('investment-projects/views/requirements-edit')
 }
 
-function editDetailsPost (req, res) {
-  if (res.locals.form.errors) {
-    return res.render('investment-projects/views/details-edit')
+function editDetailsPost (req, res, next) {
+  if (res.locals.form.errors || req.body.add_contact) {
+    return next()
   }
   req.flash('success', 'Investment details updated')
   return res.redirect(`/investment-projects/${res.locals.resultId}/details`)
--- a/src/apps/investment-projects/controllers/edit.js
+++ b/src/apps/investment-projects/controllers/edit.js
@@ -17,7 +17,7 @@ function editRequirementsGet (req, res) {
 }
 
 function editDetailsPost (req, res, next) {
-  if (res.locals.form.errors || req.body.add_contact) {
+  if (res.locals.form.errors || req.body['add-item']) {
     return next()
   }
   req.flash('success', 'Investment details updated')
--- a/src/apps/investment-projects/middleware/forms/details.js
+++ b/src/apps/investment-projects/middleware/forms/details.js
@@ -1,3 +1,4 @@
+/* eslint-disable camelcase */
 const { get, isEmpty } = require('lodash')
 
 const { transformToApi, transformFromApi } = require('../../services/formatting')
@@ -84,6 +85,24 @@ function handleFormPost (req, res, next) {
   const formattedBody = transformToApi(Object.assign({}, req.body))
   const projectId = res.locals.projectId || req.params.investmentId
   let saveMethod
+
+  if (req.body.add_contact) {
+    res.locals.form = Object.assign({}, res.locals.form, {
+      state: req.body,
+    })
+
+    let client_contacts = res.locals.form.state.client_contacts
+
+    if (Array.isArray(client_contacts)) {
+      client_contacts.push('')
+    } else {
+      client_contacts = [client_contacts, '']
+    }
+
+    res.locals.form.state.client_contacts = client_contacts
+
+    return next()
+  }
 
   if (projectId) {
     saveMethod = updateInvestment(req.session.token, projectId, formattedBody)
--- a/src/apps/investment-projects/middleware/forms/details.js
+++ b/src/apps/investment-projects/middleware/forms/details.js
@@ -1,5 +1,4 @@
-/* eslint-disable camelcase */
-const { get, isEmpty } = require('lodash')
+const { flatten, get, isEmpty } = require('lodash')
 
 const { transformToApi, transformFromApi } = require('../../services/formatting')
 const { isValidGuid } = require('../../../../lib/controller-utils')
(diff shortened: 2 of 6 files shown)


src/apps/investment-projects/middleware/forms/details.js
```diff
@@ -85,6 +86,24 @@ function handleFormPost (req, res, next) {
   const projectId = res.locals.projectId || req.params.id
   let saveMethod
 
+  if (req.body.add_contact) {
```
teneightfive: Are we only supporting adding a new contact without JS for now and not removing one too? 

I suppose its fine for now but maybe we should make a note of it or add a `// TODO` to fix in the future.

src/apps/investment-projects/middleware/forms/details.js
```diff
@@ -85,6 +86,24 @@ function handleFormPost (req, res, next) {
...
+    res.locals.form = Object.assign({}, res.locals.form, {
+      state: req.body,
+    })
+
+    let client_contacts = res.locals.form.state.client_contacts
```
teneightfive: See comments on add another button about changing to `button` element.

Then this block could become:

```js
const addKey = req.body['add-item']

req.body[addKey] = flatten([req.body[addKey]])
req.body[addKey].push('')

  state: req.body,
})

return next()
```

I think it makes it easier to read and also supports add another in any other part of the form if we need it in the future.

src/apps/investment-projects/views/_details-form.njk
```diff
@@ -112,14 +112,33 @@
...
+        isLabelHidden: true,
+        initialOption: '-- Pick a contact --',
+        options: form.options.contacts,
+        value: client_contact
+      }) }}
+    {% endfor %}
+
+
+    <p class="c-form-fieldset__add-another">
+      <input class="button button-secondary js-AddItems__add" type="submit" name="add_contact" value="Add another" data-persist-values="true" data-method="add">
```
teneightfive: Maybe change this to a button with the field to add as the value:

```html
<button class="button button-secondary js-AddItems__add" type="button" name="add-item" value="client_contacts" data-persist-values="true">Add another</button>
```

Also, the `data-method` attribute you have added is added by the JS module I believe so don't need to add in the markup.

src/apps/investment-projects/middleware/forms/details.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable camelcase */
```
teneightfive: We don't need to add this if we use the code below.

src/apps/investment-projects/services/formatting.js
```diff
@@ -51,7 +51,13 @@ function transformToApi (body) {
     }
 
     if (type === Array) {
-      return [{ id: value }]
+      if (Array.isArray(value)) {
+        return value.map(item => {
+          return { id: item }
+        })
+      } else {
```
teneightfive: We don't need the else as we're returning the value above anyway. It's easier to read if you don't add the `else` too.

src/apps/investment-projects/views/_details-form.njk
ztolley: I copied the forms example, hence using the input instead of button

src/apps/investment-projects/middleware/forms/details.js
```diff
@@ -84,26 +83,22 @@ async function populateForm (req, res, next) {
 function handleFormPost (req, res, next) {
   const formattedBody = transformToApi(Object.assign({}, req.body))
   const projectId = res.locals.projectId || req.params.id
-  let saveMethod
 
-  if (req.body.add_contact) {
+  // Todo - currently only supports add with none js,
```
teneightfive: Do you mean non-JS?

src/apps/investment-projects/middleware/forms/details.js
```diff
@@ -84,26 +83,22 @@ async function populateForm (req, res, next) {
...
-    } else {
-      client_contacts = [client_contacts, '']
-    }
-
-    res.locals.form.state.client_contacts = client_contacts
-
     return next()
   }
 
+  let saveMethod
```
teneightfive: Why has this moved? It's easier to understand when you try and keep the variable declarations in one place at the top of the function.

src/apps/investment-projects/views/_details-form.njk
```diff
@@ -90,27 +90,28 @@
     value: form.state['sector']
   }) }}
 
+
```
teneightfive: Unnecessary new line?

src/apps/investment-projects/views/_details-form.njk
```diff
@@ -90,27 +90,28 @@
...
-    error: form.errors.messages['other_business_activity'],
-    value: form.state['other_business_activity'],
-    condition: {
-      name: 'business_activities',
-      value: form.options.businessActivities[8].value
-    },
-    modifier: 'subfield'
-  }) }}
+  {{ TextField({
+      name: 'other_business_activity',
```
teneightfive: Is all this extra formatting noise necessary?

ztolley: ?

src/apps/investment-projects/middleware/forms/details.js
ztolley: I tend to declare variables when they are needed as cont and let are not hoisted.

src/apps/investment-projects/views/_details-form.njk
ztolley: It's correcting broken formatting in the existing form.

src/apps/investment-projects/middleware/forms/details.js
```diff
@@ -83,10 +83,11 @@ async function populateForm (req, res, next) {
 function handleFormPost (req, res, next) {
   const formattedBody = transformToApi(Object.assign({}, req.body))
   const projectId = res.locals.projectId || req.params.id
+  const addKey = req.body['add-item']
+  let saveMethod
 
-  // Todo - currently only supports add with none js,
+  // Todo - currently only supports add with non-js,
   // add remove action for none js later, for now the user just sets the value to nothing.
```
teneightfive: Missed another `none js` -> `non-js`. Sorry, should've put a comment on last time.

src/apps/investment-projects/controllers/edit.js
```diff
@@ -16,9 +16,9 @@ function editRequirementsGet (req, res) {
   res.render('investment-projects/views/requirements-edit')
 }
 
-function editDetailsPost (req, res) {
-  if (res.locals.form.errors) {
-    return res.render('investment-projects/views/details-edit')
+function editDetailsPost (req, res, next) {
+  if (res.locals.form.errors || req.body.add_contact) {
```
teneightfive: Just spotted this might need changing as you're not sending `add_contact` on the request body.

src/apps/investment-projects/views/_details-form.njk
```diff
@@ -96,30 +96,47 @@
...
+        isLabelHidden: true,
+        initialOption: '-- Pick a contact --',
+        options: form.options.contacts,
+        value: client_contact
+      }) }}
+    {% endfor %}
+
+
+    <p class="c-form-fieldset__add-another">
+      <button class="button button-secondary js-AddItems__add" type="button" name="add-item" value="client_contacts" data-persist-values="true">Add another</button>
```
teneightfive: I think I got this wrong originally, the `type` attribute should be `submit` for it to work without JavaScript.

PR #779, opened 2017-10-17
Reconciliation listing

teneightfive:
This change adds an initial listing page for orders that require reconciliation.

It makes use of a lot of the collections code and macros but uses a custom display for the items.

## What it looks like

localhost_3001_omis_reconciliation_sortby created_on 3adesc


src/apps/omis/transformers.js
```diff
@@ -57,6 +57,27 @@ function transformOrderToListItem ({
...
 
+function transformOrderToTableItem ({
+  id,
+  reference,
+  status,
+  company,
+  net_cost,
+  total_cost,
+} = {}) {
+  if (!id || !reference) { return }
```
yeahfro: are we okay with silently failing here and not throwing an exception?

teneightfive: Yes, it's what we've done in other transformers. Basically it just won't return this item if it doesn't have the minimum properties.

We usually use it to loop over an array of results from the API so a certain result just wouldn't show in the list.

teneightfive: We could log a warning, but if that's our approach we should probably do that across all collection style transformers. I'm okay with the current approach for now.

PR #784, opened 2017-10-19
Feature/fix circleci

feedmypixel:
- bump `selenium-server` and `chromeDriver` dependencies
- add `--no-sandbox` arg
- source `nvm.sh` before each `run` command via `BASH_ENV`

Associated `docker-data-hub-base` PR [#3](https://github.com/uktrade/docker-data-hub-base/pull/3)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -64,6 +64,8 @@ jobs:
       - store_artifacts:
           path: coverage
   user_acceptance_tests: &user_acceptance_tests
+    environment:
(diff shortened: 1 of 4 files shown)


ztolley:
Ever thought of pointing it at browser stack so you don’t need to maintain the chrome drivers, which will become a pain?

feedmypixel:
> Ever thought of pointing it at browser stack so you don’t need to maintain the chrome drivers, which will become a pain?

Yup we spoke about browser stack a little while ago. We would still need chrome driver to run locally and debug whilst creating tests.

PR #785, opened 2017-10-19
Project team labels

ztolley:
Updated the labels in the investment team page for https://uktrade.atlassian.net/browse/DH-803

While there also fixed a minor issue on the details page that caused it to crash due to missing data in the DB.  The page should be more robust.

cap1
cap2

### DIFF
--- a/src/apps/investment-projects/services/formatting.js
+++ b/src/apps/investment-projects/services/formatting.js
@@ -327,7 +327,7 @@ function transformClientRelationshipManagementForView (investmentData) {
 function transformTeamMembersForView ({ adviser, role }) {
   return {
     adviser: get(adviser, 'name'),
-    team: get(adviser.dit_team, 'name'),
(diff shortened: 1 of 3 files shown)


src/apps/investment-projects/services/formatting.js
```diff
@@ -326,8 +326,8 @@ function transformClientRelationshipManagementForView (investmentData) {
 
 function transformTeamMembersForView ({ adviser, role }) {
   return {
-    adviser: adviser.name,
-    team: adviser.dit_team.name,
+    adviser: get(adviser, 'name'),
+    team: get(adviser.dit_team, 'name'),
```
teneightfive: Change this to `get(adviser, 'dit_team.name')` to be safer.

ztolley: Done.. yes I rebased and squashed though. Obviously excited I had one approval :)

PR #788, opened 2017-10-19
Fixed show investment details errors

ztolley:
Previously if a user saved investment details form with a single contact or business activity, and that form had an error, then the form redraw would show multiple contacts and activities, all empty. This was caused because the form expects contacts and activities to be an array and if the form submits a single value then that value was passed back to the form to redraw.

The fix, for now, is to force the contact and activities fields to an array when the form is posted.

fix

### DIFF
--- a/src/apps/investment-projects/controllers/edit.js
+++ b/src/apps/investment-projects/controllers/edit.js
@@ -36,7 +36,7 @@ function editValuePost (req, res) {
 
 function editRequirementsPost (req, res, next) {
   if (get(res.locals, 'requirementsForm.errors')) {
-    return next()
+    return res.render('investment-projects/views/requirements-edit')
   }
   req.flash('success', 'Investment requirements updated')
   return res.redirect(`/investment-projects/${res.locals.projectId}/details`)
--- a/src/apps/investment-projects/controllers/edit.js
+++ b/src/apps/investment-projects/controllers/edit.js
@@ -34,7 +34,7 @@ function editValuePost (req, res) {
   return res.redirect(`/investment-projects/${res.locals.projectId}/details`)
 }
 
-function editRequirementsPost (req, res, next) {
+function editRequirementsPost (req, res) {
   if (get(res.locals, 'requirementsForm.errors')) {
     return res.render('investment-projects/views/requirements-edit')
   }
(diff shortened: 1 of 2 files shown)


src/apps/investment-projects/middleware/forms/details.js
```diff
@@ -81,6 +81,16 @@ async function populateForm (req, res, next) {
 }
 
 function handleFormPost (req, res, next) {
+  // force contacts and actitivies to be arrays to make handling them consistent,
+  // and avoid issues when errors happen and form is re-built
+  if (!Array.isArray(req.body['client_contacts'])) {
```
teneightfive: This could be neater if you keep flatten:

```js
req.body.client_contacts = flatten([req.body.client_contacts])
req.body.business_activities = flatten([req.body.client_contacts])
```

src/apps/investment-projects/controllers/edit.js
```diff
@@ -34,9 +34,9 @@ function editValuePost (req, res) {
   return res.redirect(`/investment-projects/${res.locals.projectId}/details`)
 }
 
-function editRequirementsPost (req, res) {
+function editRequirementsPost (req, res, next) {
   if (get(res.locals, 'requirementsForm.errors')) {
-    return res.render('investment-projects/views/requirements-edit')
+    return next()
```
teneightfive: I don't think you need to change this. This can be left to render the view.

src/apps/investment-projects/middleware/forms/details.js
```diff
@@ -81,6 +81,11 @@ async function populateForm (req, res, next) {
 }
 
 function handleFormPost (req, res, next) {
+  // force contacts and actitivies to be arrays to make handling them consistent,
+  // and avoid issues when errors happen and form is re-built
+  req.body.client_contacts = flatten([req.body.client_contacts])
```
teneightfive: Just had a look at this and I don't think they need to be so prominently placed in this function.

This case only happens during validation so these lines can be moved to the if block in the catch statement (currently on line 118).

src/apps/investment-projects/controllers/edit.js
ztolley: I had changed the router previously to support this but forgot to put in the next. Just thought I'd do it while here. If this is blocking I'll remove it

src/apps/investment-projects/controllers/edit.js
```diff
@@ -34,7 +34,7 @@ function editValuePost (req, res) {
   return res.redirect(`/investment-projects/${res.locals.projectId}/details`)
 }
 
-function editRequirementsPost (req, res) {
+function editRequirementsPost (req, res, next) {
```
teneightfive: This can also be removed seeing as its not being called any more.

PR #791, opened 2017-10-19
Use lodash's get for finding first and last name of contact

teneightfive:
This protects against the contact object not existing.

### DIFF
--- a/src/apps/omis/apps/create/controllers/confirm.js
+++ b/src/apps/omis/apps/create/controllers/confirm.js
@@ -14,8 +14,8 @@ class ConfirmController extends FormController {
       const company = res.locals.company
       const contact = find(company.contacts, { id: values.contact })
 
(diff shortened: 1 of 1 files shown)


src/apps/omis/apps/create/controllers/confirm.js
```diff
@@ -14,8 +14,8 @@ class ConfirmController extends FormController {
       const company = res.locals.company
       const contact = find(company.contacts, { id: values.contact })
 
-      values.contact = `${contact.first_name} ${contact.last_name}`
       values.company = company
+      values.contact = `${get(contact, 'first_name')} ${get(contact, 'last_name')}`
```
ztolley: Silly question, why?

teneightfive: I found an error when there is no contact. Receives `cannot find property first_name of undefined`.

This just guards against that and stops the request hanging.

PR #793, opened 2017-10-20
Fixed issue with empty contact and activity on add invest

ztolley:
The add investment screen did not show a contact or business activity option due to it expecting an array with at least one entry, so default entries needed to be added for the add flow.

### DIFF
--- a/src/apps/investment-projects/middleware/forms/details.js
+++ b/src/apps/investment-projects/middleware/forms/details.js
@@ -1,4 +1,4 @@
-const { flatten, get, isEmpty } = require('lodash')
+const { assign, flatten, get, isEmpty } = require('lodash')
 
 const { transformToApi, transformFromApi } = require('../../services/formatting')
 const { isValidGuid } = require('../../../../lib/controller-utils')
@@ -41,7 +41,10 @@ async function populateForm (req, res, next) {
     res.locals.equityCompany = equityCompany
     res.locals.equityCompanyInvestment = equityCompanyInvestment
     res.locals.form = res.locals.form || {}
-    res.locals.form.state = investmentData
+    res.locals.form.state = assign({}, {
+      client_contacts: [''],
+      business_activities: [''],
+    }, investmentData)
 
     res.locals.form.options = {
       advisers,
@@ -57,7 +60,7 @@ async function populateForm (req, res, next) {
       businessActivities: metadata.businessActivityOptions.map(transformObjectToOption),
     }
 
-    if (res.locals.form.state) {
+    if (investmentData) {
       // TODO: This is to support the leading question of whether current
       // user is the CRM or adviser - this journey will be changed in the
       // future but until then this supports the settings of that data


src/apps/investment-projects/middleware/forms/details.js
```diff
@@ -72,6 +72,16 @@ async function populateForm (req, res, next) {
       } else {
         res.locals.form.state.is_referral_source = 'false'
       }
+    } else {
+      res.locals.form.state = {}
```
teneightfive: What about doing this on line 44 instead to ensure its always an object:

```js
res.locals.form.state = investmentData || {}
```

src/apps/investment-projects/middleware/forms/details.js
```diff
@@ -72,6 +72,16 @@ async function populateForm (req, res, next) {
...
+    }
+
+    if (!res.locals.form.state.client_contacts) {
```
teneightfive: If fact, line 44 could become this to avoid these nasty looking if blocks:

```js
res.locals.form.state = assign({
  client_contacts: [''],
  business_activities: [''],
}, investmentData)
```

teneightfive: Although that may cause issues with the block above. If you change line 60 to this it fixes it: 

```js
if (investmentData) {
  ...
}
```

src/apps/investment-projects/middleware/forms/details.js
ztolley: If you do that the the next part fails as it's checking to see if there is state or not.

src/apps/investment-projects/middleware/forms/details.js
ztolley: .. ok I will 'revisit' this

src/apps/investment-projects/middleware/forms/details.js
```diff
@@ -41,7 +41,10 @@ async function populateForm (req, res, next) {
     res.locals.equityCompany = equityCompany
     res.locals.equityCompanyInvestment = equityCompanyInvestment
     res.locals.form = res.locals.form || {}
-    res.locals.form.state = investmentData
+    res.locals.form.state = assign({
```
feedmypixel: Best practice, worth having a {} as the first argument here to avoid mutation.

src/apps/investment-projects/middleware/forms/details.js
```diff
@@ -41,7 +41,10 @@ async function populateForm (req, res, next) {
...
+      client_contacts: [''],
```
feedmypixel: do these have to be arrays with an empty value in? can they not just be empty arrays? or `null`

ztolley: the whole point of the change it to put in an empty value in there otherwise no drop-down fields are shown. This is a quick and dirty fix to buy time to refactor the flow later.

PR #802, opened 2017-10-23
Fix invest errors

ztolley:
Fixed 2 errors:

1: Fixed repopulating the contact and business activity fields when saving an investment summary with errors. Added unit tests to confirm.
2: Fixed error not showing error messages for business activities and contacts when saving

errors

### DIFF
--- a/src/apps/investment-projects/views/_details-form.njk
+++ b/src/apps/investment-projects/views/_details-form.njk
@@ -90,12 +90,7 @@
     value: form.state['sector']
   }) }}
 
-  <fieldset class="c-form-fieldset {{ 'has-error' if form.errors.messages['business_activities'] }}">
-
-    <legend class="c-form-fieldset__legend">
-      <span class="c-form-fieldset__legend-text">Business activities</span>
-    </legend>
-
+  {% call Fieldset({ legend: 'Business activities' }) %}
     {% if form.errors.messages['business_activities'] %}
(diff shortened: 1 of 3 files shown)


src/apps/investment-projects/views/_details-form.njk
```diff
@@ -90,12 +90,16 @@
     value: form.state['sector']
   }) }}
 
-  <fieldset class="c-form-fieldset">
+  <fieldset class="c-form-fieldset {{ 'has-error' if form.errors.messages['business_activities'] }}">
```
teneightfive: Wonder if would be easier to use the `FormGroup` macro here with `fieldset` as the element and would mean we don't have this code duplicated?

ztolley: what would that look like?

feedmypixel:
@ztolley why can't you just use the form errors work we have?

src/apps/investment-projects/views/_details-form.njk
ztolley: Tried using formgroup macro but ran into issues with styling due to errors being shown within the legend. Any changes to formgroup would have an adverse effect on Omis.

Re. use of dot notation, the whole template uses [''] so sticking with that for consistency.

ztolley:
@feedmypixel the form errors get rendered in the legend which totally throw the layout.

feedmypixel:
@ztolley can you adda gif? Its hard to work out what the actual problem is here

PR #807, opened 2017-10-23
doublecheck whether company is *really* a CH company or not before assigning CH URL

jim68000:
This addresses the issue described here https://uktrade.atlassian.net/browse/DH-966

Is this the smallest PR yet at 7 characters?

### DIFF
--- a/src/apps/companies/transformers.js
+++ b/src/apps/companies/transformers.js
@@ -65,7 +65,7 @@ function transformCompanyToListItem ({
     })
   }
 
-  const url = companies_house_data ? `/companies/view/ch/${companies_house_data.company_number}` : `/companies/${id}`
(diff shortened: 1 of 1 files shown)


src/apps/companies/transformers.js
```diff
@@ -65,7 +65,7 @@ function transformCompanyToListItem ({
     })
   }
 
-  const url = companies_house_data ? `/companies/view/ch/${companies_house_data.company_number}` : `/companies/${id}`
+  const url = companies_house_data && !id ? `/companies/view/ch/${companies_house_data.company_number}` : `/companies/${id}`
```
teneightfive: Isn't this already being handled on line 26? This transformer should just `return` when there is no `id` property.

jim68000: Yeah, this is what happens when you think you've found a clever hack. Redoing

PR #810, opened 2017-10-23
Add a listing for orders for a specific company

teneightfive:
This change adds the listing of orders filtered by company id
and also adds a place to begin a new order for this company.

## What it looks like

localhost_3001_companies_dcdabbc9-1781-e411-8955-e4115bead28a_orders

### DIFF
--- a/src/apps/companies/controllers/orders.js
+++ b/src/apps/companies/controllers/orders.js
@@ -1,0 +1,38 @@
+const { search } = require('../../search/services')
+const { transformApiResponseToCollection } = require('../../transformers')
+const { transformOrderToListItem } = require('../../omis/transformers')
+
+async function renderOrders (req, res, next) {
+  const token = req.session.token
+  const page = req.query.page || 1
+  const { id, name } = res.locals.company
+
+  try {
+    const results = await search({
+      token,
+      page,
+      searchEntity: 'order',
+      requestBody: {
+        company: id,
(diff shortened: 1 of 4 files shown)


src/apps/companies/controllers/orders.js
```diff
@@ -0,0 +1,38 @@
...
+const { transformApiResponseToCollection } = require('../../transformers')
+const { transformOrderToListItem } = require('../../omis/transformers')
+
+async function renderOrders (req, res, next) {
+  const token = req.session.token
+  const page = req.query.page || 1
+  const { id, name } = res.locals.company
+
+  try {
+    const results = await search({
```
ztolley: Should this go into a repos?

ztolley: Also no tests, could test to make sure the search params are correct or maybe talk to Ben about UAT

teneightfive: >Should this go into a repos?

OMIS doesn't have the same concept of _repos_. It was actually something we discussed removing as wasn't very clear what they do. And the logic in them or services wasn't very consistent across the codebase.

OMIS uses the idea of models to make calls to the API to retrieve data. But it doesn't have the same of listings like companies or contacts and they can only be retrieved using search so made more sense to use the search service.

>Also no tests

Good shout. Will add some now.

PR #815, opened 2017-10-24
Add Performance dashboards global nav link

yeahfro:
Add "Performance dashboards"

screen shot 2017-10-24 at 10 55 05

### DIFF
--- a/src/middleware/locals.js
+++ b/src/middleware/locals.js
@@ -15,7 +15,7 @@ const globalNavItems = [
   { path: '/interactions', label: 'Interactions' },
   { path: '/investment-projects', label: 'Investment projects' },
   { path: '/omis', label: 'Orders (OMIS)' },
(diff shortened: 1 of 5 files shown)


feedmypixel:
Can you add some UAT around this change also?

src/middleware/locals.js
```diff
@@ -15,6 +15,7 @@ const globalNavItems = [
   { path: '/interactions', label: 'Interactions' },
   { path: '/investment-projects', label: 'Investment projects' },
   { path: '/omis', label: 'Orders (OMIS)' },
+  { path: 'https://mi.exportwins.service.trade.gov.uk', label: 'Performance dashboards' },
```
teneightfive: What about making this use an environment variable? So we can link to their development/staging environments when our systems are in that environment.

Avoids people accidentally ending up on production via our dev environment and changing something.

PR #816, opened 2017-10-24
Add project status

ztolley:
Add a status indicator for an investment project to the investment project layout, and allow the user to change the status.

As part of the change introduced a new tabular style for metadata and support in meta-list items to have a link separate from the value.

metalistitems

status

### DIFF
--- a/src/apps/investment-projects/controllers/status.js
+++ b/src/apps/investment-projects/controllers/status.js
@@ -1,0 +1,41 @@
+const { assign } = require('lodash')
+
+const { buildFormWithStateAndErrors } = require('../../builders')
+const { statusFormConfig } = require('../macros')
+const { updateInvestment } = require('../repos')
+
+function renderStatusPage (req, res, next) {
+  const status = req.body.status || res.locals.investmentData.status
+
+  const statusForm = assign(
+    buildFormWithStateAndErrors(statusFormConfig, { status }, res.locals.errors),
+    { returnLink: `/investment-projects/${res.locals.investmentData.id}/details` },
+  )
+
+  res
+    .breadcrumb('Change project status')
+    .render('investment-projects/views/status', { statusForm })
+}
+
+async function postStatus (req, res, next) {
+  try {
+    await updateInvestment(req.session.token, req.params.investmentId, {
+      status: req.body.status,
+    })
+
+    req.flash('success', 'Investment details updated')
+
+    return res.redirect(`/investment-projects/${res.locals.investmentData.id}/details`)
+  } catch (error) {
+    if (error.statusCode === 400) {
+      res.locals.errors = error.error
+      return next()
+    }
+    next(error)
+  }
+}
+
+module.exports = {
+  renderStatusPage,
+  postStatus,
+}
--- a/src/apps/investment-projects/middleware/shared.js
+++ b/src/apps/investment-projects/middleware/shared.js
@@ -1,4 +1,4 @@
-const { get } = require('lodash')
+const { get, upperFirst } = require('lodash')
 
 const metadata = require('../../../lib/metadata')
 const { isValidGuid } = require('../../../lib/controller-utils')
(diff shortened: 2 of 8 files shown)


assets/stylesheets/components/_meta-list.scss
```diff
@@ -104,3 +104,20 @@
...
 }
+
+.c-meta-list--tabular {
+  display: table;
+
+  .c-meta-list__item {
+    display: table-row;
+
+    .c-meta-list__item-label {
+      padding-right: 20px;
```
teneightfive: Maybe use `$default-spacing-unit` to keep consistency.

src/apps/investment-projects/controllers/status.js
```diff
@@ -0,0 +1,41 @@
...
+
+    req.flash('success', 'Investment details updated')
+
+    return res.redirect(`/investment-projects/${res.locals.investmentData.id}/details`)
+  } catch (error) {
+    if (error.statusCode === 400) {
+      res.locals.errors = error.error
+      return next()
+    }
+    return next(error)
```
teneightfive: You can drop the `return` on this line.

src/apps/investment-projects/middleware/shared.js
```diff
@@ -1,4 +1,7 @@
-const { get } = require('lodash')
+const {
```
teneightfive: I'd say this is still small enough to keep on one line.

src/apps/investment-projects/middleware/shared.js
```diff
@@ -52,6 +55,12 @@ async function getInvestmentDetails (req, res, next) {
       id: investmentData.id,
       meta: [
         {
+          label: 'Status',
+          value: upperFirst(investmentData.status),
```
teneightfive: Maybe try and avoid the old style of formatting values. This could be done using the `type` property and extending the macro.

src/templates/_macros/entity/meta-item.njk
```diff
@@ -44,15 +45,20 @@
...
           class="js-xhr {{ itemValueClass }} {{ 'is-selected' if props.isSelected }}"
           href="{{ props.url }}"
         >
           {{- safeValue -}}
         </a>
       {% else %}
-        <span class="{{ itemValueClass }}">{{ safeValue }}</span>
+        <span class="{{ itemValueClass }}">
+          {{ safeValue }}
+          {% if props.url and not props.isInert and props.urlLabel %}
```
teneightfive: I don't think `isInert` should apply here. It only applies to when using the value as the label.

Even if that prop exists in this format I imagine this link should be displayed.

assets/stylesheets/components/_meta-list.scss
```diff
@@ -104,3 +104,20 @@
     margin-left: 0;
   }
 }
+
+.c-meta-list--tabular {
```
teneightfive: I'm also wondering if we really need a _tabular_ modifier for this. I think if this is something we really want we should consider how it affects other parts of the site as we'd then be moving away from consistency again.

What about keeping the consistency for now and doing this as part of reviewing the header meta items as a whole. We should avoid introducing inconsistencies if we can at this stage.

ztolley: I actually like this style and I think that this style will spread. None of the other meta lists in headers allow for this right now, they don't give you the alignment you need. Plus I can't wait to review the meta items as a whole, I need to deliver this now.

src/apps/investment-projects/middleware/shared.js
ztolley: Not sure we have a 'type' that requires just the first letter to be transformed to a capital and that feels like overkill, this is simple.

src/templates/_macros/entity/meta-item.njk
ztolley: ok, the definition of what Inert is is not clear then.

assets/stylesheets/components/_meta-list.scss
ztolley: As agreed, removing this and switching to the existing styles.

ztolley: updatedstatus

PR #817, opened 2017-10-24
Feedback page: removed optional from email field

jim68000:


src/apps/support/macros.js
```diff
@@ -34,7 +34,7 @@ const feedbackFormConfig = (browserInfo) => ({
     },
     {
       macroName: 'TextField',
-      optional: true,
+      optional: false,
```
feedmypixel: @jim68000 it defaults to false so you can just remove the `optional`compeltely

jim68000: Pfft, now I have to a fixup

PR #823, opened 2017-10-25
Add switch for self signed certs to allow dev debugging of API calls

jim68000:


src/lib/authorised-request.js
```diff
@@ -54,6 +54,9 @@ module.exports = (token, opts) => {
 
   if (process.env.PROXY) {
     requestOptions.proxy = process.env.PROXY
+    if (process.env.NODE_ENV !== 'production') {
```
feedmypixel: you can use `isDev` found in [config](https://github.com/uktrade/data-hub-frontend/blob/develop/config/index.js#L11)

jim68000: Ah yeah, I didn't spot it was exported. Changing

PR #824, opened 2017-10-25
business_type errors out on Add company form halting the ability to add a company

feedmypixel:
## Problem
When creating a new company **Other type of UK organisation** or **Foreign organisation** and choosing **Government department** as **Type of organisation** you receive a form error on the **Add company** page and would be unable to proceed.

### Steps to reproduce
- Go to company list page
- Press add company button
- Choose Other type of UK organisation
- Select Government department as Type of organisation
- Click Continue
- Enter all relevant information in Add company form
- Click Save and Create
- Note form error

## Info
This work:
- Brings in **all** of the correct `businessTypes` from the metaData repo
- Uses the `business_type.id` rather than `business_type.name`

### DIFF
--- a/src/apps/companies/controllers/add.js
+++ b/src/apps/companies/controllers/add.js
@@ -2,7 +2,7 @@ const { assign } = require('lodash')
 const queryString = require('query-string')
 
 const logger = require('../../../../config/logger')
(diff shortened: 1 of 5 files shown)


feedmypixel:
> What is ltd child and are we sure it's safe to remove it? Is this not something in the db?

Just double checking - have remove this from this PR and will raise seperately

PR #830, opened 2017-10-25
Payment receipt

teneightfive:
This change adds a new route to view the payment receipt after an order has been paid.

There will be some follow-up changes around the flow but didn't want to bloat this PR too much to keep it easier to review.

## What it looks like

localhost_3001_omis_419af569-db12-4d41-8f00-a8c764faf578_payment-receipt

### DIFF
--- a/assets/stylesheets/elements/_table.scss
+++ b/assets/stylesheets/elements/_table.scss
@@ -35,3 +35,8 @@ table {
     text-align: right;
   }
 }
+
+
+.table__cell--plain {
+  border-bottom: 0;
+}
--- a/src/apps/omis/apps/view/middleware.js
+++ b/src/apps/omis/apps/view/middleware.js
@@ -112,6 +112,16 @@ async function setInvoice (req, res, next) {
   next()
 }
 
+async function setPayments (req, res, next) {
+  try {
+    res.locals.payments = await Order.getPayments(req.session.token, res.locals.order.id)
+  } catch (error) {
+    logger.error(error)
+  }
+
+  next()
+}
+
 async function generateQuote (req, res, next) {
   const orderId = get(res.locals, 'order.id')
 
@@ -200,6 +210,7 @@ module.exports = {
   setQuotePreview,
   setQuote,
   setInvoice,
+  setPayments,
   generateQuote,
   cancelQuote,
   setQuoteForm,
--- a/src/apps/omis/apps/view/middleware.js
+++ b/src/apps/omis/apps/view/middleware.js
@@ -102,6 +102,16 @@ async function setQuote (req, res, next) {
   next()
 }
 
+async function setInvoice (req, res, next) {
+  try {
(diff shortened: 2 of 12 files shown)


src/apps/omis/apps/view/middleware.js
```diff
@@ -102,6 +102,26 @@ async function setQuote (req, res, next) {
   next()
 }
 
+async function setInvoice (req, res, next) {
```
feedmypixel: are these not `getInvoice` and `getPayments`?

src/apps/omis/apps/view/router.js
```diff
@@ -29,6 +35,7 @@ router.use(setQuoteSummary)
 
 router.get('/', redirectToFirstNavItem)
 router.get('/work-order', renderWorkOrder)
+router.get('/payment-receipt', setInvoice, setPayments, renderPaymentReceipt)
```
feedmypixel: separate lines?

src/apps/omis/apps/view/views/payment-receipt.njk
```diff
@@ -0,0 +1,123 @@
...
+  </div>
+
+  <table>
+    <thead>
+      <tr>
+        <th>Order number</th>
+        <th>Service description</th>
+        <th>Market (country)</th>
+        <th></th>
+        <th align="right">Net amount</th>
```
feedmypixel: could this be a style?

assets/stylesheets/elements/_table.scss
```diff
@@ -35,3 +35,8 @@ table {
     text-align: right;
   }
 }
+
+
+.table__cell--unstyled {
+  border-bottom: 0;
+}
```
feedmypixel: maybe `table__cell--plain`/`table__cell--simple`/`table__cell--light` ? as it does still have styles

teneightfive: Good shout 👍

src/apps/omis/apps/view/middleware.js
teneightfive: Negative. I've been using the pattern of `set` because it better describers what these functions are doing.

The function doesn't return anything so `get` is the wrong verb. The purpose of it is to set a variable on locals to be available to the view.

src/apps/omis/apps/view/router.js
teneightfive: I find it doesn't really make it easier to read with function args because and this still doesn't break the 80 character limit which is what I normally break lines manually on.

src/apps/omis/apps/view/views/payment-receipt.njk
teneightfive: It felt like it made more sense to use the html attribute so that when styles aren't applied (like our current print style status) it still displays in a sensible way of aligning numbers to the right. Similar to what tools like Excel do when dealing with numbers.

PR #834, opened 2017-10-26
Update investment requirements to add another

ztolley:
Adds 'add another' support to the investment requirements form forstrategic drivers, competitor countries and possible uk regions.

screencapture-localhost-3000-investment-projects-d886a660-5268-4149-823b-5805029e0639-edit-requirements-1509030249644

screencapture-localhost-3000-investment-projects-d886a660-5268-4149-823b-5805029e0639-details-1509030274533

### DIFF
--- a/src/apps/investment-projects/labels.js
+++ b/src/apps/investment-projects/labels.js
@@ -45,16 +45,16 @@ const labels = {
   },
   requirementsLabels: {
     view: {
-      strategic_drivers: 'Main strategic driver',
+      strategic_drivers: 'Strategic drivers',
       client_requirements: 'Client requirements',
       competitor_countries: 'Competitor countries',
       uk_region_locations: 'Possible UK locations',
       uk_company: 'UK recipient company',
     },
     edit: {
-      strategic_drivers: 'Main strategic driver behind this investment',
+      strategic_drivers: 'Strategic drivers behind this investment',
       client_considering_other_countries: 'Is the client considering other countries?',
-      uk_region_locations: 'Possible UK location for this investment',
+      uk_region_locations: 'Possible UK locations for this investment',
       site_decided: 'Has the UK location (site address) for this investment been decided yet?',
     },
   },
--- a/src/apps/investment-projects/middleware/forms/requirements.js
+++ b/src/apps/investment-projects/middleware/forms/requirements.js
@@ -1,14 +1,13 @@
-const { assign, get, flatten } = require('lodash')
+const { assign, flatten } = require('lodash')
 const { updateInvestment } = require('../../repos')
 const { requirementsFormConfig } = require('../../macros')
-const { buildFormWithStateAndErrors } = require('../../../builders')
+const {
+  buildFormWithStateAndErrors,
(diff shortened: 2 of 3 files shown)


src/apps/investment-projects/labels.js
```diff
@@ -45,16 +45,16 @@ const labels = {
...
       competitor_countries: 'Competitor countries',
       uk_region_locations: 'Possible UK locations',
       uk_company: 'UK recipient company',
     },
     edit: {
-      strategic_drivers: 'Main strategic driver behind this investment',
+      strategic_drivers: 'Strategic drivers behind this investment',
       client_considering_other_countries: 'Is the client considering other countries?',
-      uk_region_locations: 'Possible UK location for this investment',
+      uk_region_locations: 'Possible UK locations for this investment',
```
feedmypixel: should this be `Possible UK location(s)` or `Possible UK location/s`

src/apps/investment-projects/labels.js
```diff
@@ -45,16 +45,16 @@ const labels = {
...
-      strategic_drivers: 'Main strategic driver',
+      strategic_drivers: 'Strategic drivers',
       client_requirements: 'Client requirements',
       competitor_countries: 'Competitor countries',
       uk_region_locations: 'Possible UK locations',
       uk_company: 'UK recipient company',
     },
     edit: {
-      strategic_drivers: 'Main strategic driver behind this investment',
+      strategic_drivers: 'Strategic drivers behind this investment',
```
feedmypixel: should this be `Strategic driver(s)` or `Strategic driver/s`

src/apps/investment-projects/middleware/forms/requirements.js
```diff
@@ -18,20 +17,32 @@ function populateForm (req, res, next) {
   next()
 }
 
+// Strips out empty entries so they are not posted (which result in errors)
+// and when an error is thrown it doesn't mean yet another is added
+function cleanArray (values) {
+  return flatten([values])
+    .filter(item => item)
+}
```
feedmypixel: feels like `values` could become `array`?

feedmypixel:
@ztolley this form is getting huge now. Does everything here still feel like it should be under the *Edit requirements and location* title?

ztolley:
Yes the form is big but yes it should be in this section

src/apps/investment-projects/middleware/forms/requirements.js
ztolley: It can either be an array or a single string, so array would be incorrect

MillsWest:
Hello. My suggestion was to keep it as plural then you don't need to do the (s) thing. 

Strategic drivers behind this investment 
then help text underneath in grey is
You can add more than one

ztolley:
We don't do that anywhere else though, so it may be inconsistent.

ztolley:
And currently, the styling doesn't support hints for 'add another', so it looks bad.

capture

So for consistency can we not use a hint and keep the wording used? It's obvious you can can have more than one because you have the 'add another' button

PR #835, opened 2017-10-26
Show link to archived files and documents for legacy orders

teneightfive:
For legacy orders we need to provide a link to the supporting files
and documents for the order. This shows that link in the actions
header.

## What it looks like

image

### DIFF
--- a/src/apps/omis/apps/view/views/_layouts/view.njk
+++ b/src/apps/omis/apps/view/views/_layouts/view.njk
@@ -2,13 +2,21 @@
 
 {% block local_header %}
   {% set actions %}
-    <p class="c-local-header__action">
-      {% if order.status == 'draft' %}
-        <a href="quote" class="button">Preview quote</a>
-      {% else %}
-        <a href="quote">View quote</a>
-      {% endif %}
-    </p>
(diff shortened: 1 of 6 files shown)


src/apps/omis/apps/view/views/_layouts/view.njk
```diff
@@ -2,13 +2,21 @@
...
-      {% if order.status == 'draft' %}
-        <a href="quote" class="button">Preview quote</a>
-      {% else %}
-        <a href="quote">View quote</a>
-      {% endif %}
-    </p>
+    {% if order.archived_documents_url_path %}
+      <p class="c-local-header__action">
+        <a href="{{ archivedDocumentsBaseUrl }}{{ order.archived_documents_url_path }}">View files and documents</a>
+        <br>
```
feedmypixel: first `break` I have seen in a while. Could this be solved via styling?

teneightfive: Felt like it could be overkill to introduce another style for this one instance. It would almost be this sort of thing:

```scss
.u-block {
  display: block;
}
```

```html
<span class="u-block">...</span>
```

which feels more dirty...

PR #836, opened 2017-10-27
Add support for links to key value table

ztolley:
Add the ability to action links to appear next to a value in a key-value table.

table

<a href="https://datahub-beta2-pr-836.herokuapp.com/components/keyvaluetables">:eyes: Demo</a>

### DIFF
--- a/src/apps/components/controllers.js
+++ b/src/apps/components/controllers.js
@@ -96,6 +96,41 @@ function renderProgress (req, res) {
     })
 }
 
+function renderKeyValueTables (req, res) {
+  const tableData = {
+    'Company': 'Acme corp',
+    'Something': {
+      name: 'Else',
+      url: '/components/',
+    },
+    'Date': {
+      type: 'date',
+      name: '20170107',
+    },
+    'Region': {
+      id: '1234',
+      name: 'South',
+    },
+    'Flavours': ['Chocolate', 'Strawberry', 'Melon'],
+    'Related Company': {
+      name: 'Freds Bananas',
+      actions: [{
+        url: '/components/',
+        label: 'Add company',
+      }, {
+        url: '/components/',
+        label: 'Remove company',
+      }],
+    },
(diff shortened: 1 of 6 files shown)


src/apps/components/controllers.js
```diff
@@ -96,6 +96,41 @@ function renderProgress (req, res) {
     })
 }
 
+function renderKeyValueTables (req, res) {
```
teneightfive: I wouldn't worry about adding a deprecated component to here to be honest.

src/templates/_components/key-value-table.njk
```diff
@@ -33,6 +33,19 @@
               {% else %}
                 {{ data.name }}
               {% endif %}
+
+              {% if data.links %}
```
teneightfive: `actions` instead of `links`?

src/templates/_components/key-value-table.njk
```diff
@@ -33,6 +33,19 @@
...
+                <span class="table__links">
+                  {% if data.links | isArray %}
```
teneightfive: I would have it always expect an array. Cleaner and easier to understand that needing to support too many different values at this stage. Keep it simple to start.

src/apps/components/controllers.js
ztolley: well it's done, I needed to be able to test the code.

src/templates/_components/key-value-table.njk
```diff
@@ -33,6 +33,15 @@
               {% else %}
                 {{ data.name }}
               {% endif %}
+
+              {% if data.actions %}
+                <span class="table__actions">
+                  {% for item in data.actions %}
+                    <a class="table__action" href="{{ item.url }}">{{ item.name }}</a>
```
teneightfive: `item.label`? `name` doesn't feel right as you're not giving the thing a name.

PR #848, opened 2017-10-30
removed duplicate client relationship manager dropdown

jim68000:
Guys, I'm not 100% sure about our favoured nunjucks code style, so please rip apart

### DIFF
--- a/src/apps/investment-projects/views/_details-form.njk
+++ b/src/apps/investment-projects/views/_details-form.njk
@@ -180,24 +180,24 @@
         value: 'false',
         label: 'No'
       }
-    ],
-    children:  [
-      {
-        macroName: 'MultipleChoiceField',
-        name: 'client_relationship_manager',
-        label: 'Client relationship manager',
-        error: form.errors.messages.client_relationship_manager,
-        value: form.state.client_relationship_manager,
-        options: form.options.advisers,
-        initialOption: '-- Select client relationship manager --',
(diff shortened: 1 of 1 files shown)


src/apps/investment-projects/views/_details-form.njk
```diff
@@ -180,24 +180,24 @@
...
-        label: 'Client relationship manager',
-        error: form.errors.messages.client_relationship_manager,
-        value: form.state.client_relationship_manager,
-        options: form.options.advisers,
-        initialOption: '-- Select client relationship manager --',
-        condition: {
+    ]
+  }) }}
+
+  {{ MultipleChoiceField({
```
teneightfive: Not quite sure what this is doing. It seems to be moving this field out of the previous form group and out on its own. What's the purpose? The title suggests removing a duplicate but this code looks like it just moves it?

jim68000: Both this dropdown, and the one immediately below it for referral source have the same UI requirements - if the user clicks 'no', we display a dropdown of advisers to pick from. 

However in the original code, the client relationship manager dropdown was expressed as a child of the radio formgroup, while the referral source has the radios and the dropdown as separate elements. 

I suspect the root problem is something screwy with the .children attribute of the component, but I thought it would be better to make the client relationship manager part of the form match the referral source part.

PR #852, opened 2017-10-31
Front end updates for mandatory event start and end date

yeahfro:
Event start and end date are now mandatory fields. This PR will:
- remove optional flag from form macro for these fields
- update acceptance tests
- some tidy up around step definitions

screen shot 2017-10-31 at 11 24 49


test/acceptance/features/events/step_definitions/create.js
```diff
@@ -189,9 +190,18 @@ defineSupportCode(function ({ Then, When }) {
   })
 
   When(/^I enter all mandatory fields related to the event$/, async () => {
+    const startDate = faker.date.future()
+    const endDate = addDays(startDate, Math.floor(Math.random() * 20))
```
feedmypixel: this could be `0` right? needs to be `Math.floor(( Math.random() * 20 ) + 1 )`

PR #856, opened 2017-10-31
Associate investment with related non-fdi R&D project

ztolley:
If a user indicates that an investment project is related to a non-fdi r&d project then the investment details summary offers a link to lookup the related project.

A new search function was required for investment as it needs to use original_query property, like company collection search.

associated


src/apps/investment-projects/controllers/associated.js
```diff
@@ -0,0 +1,89 @@
...
+    return next()
+  }
+
+  try {
+    await updateInvestment(req.session.token, req.params.investmentId, {
+      associated_non_fdi_r_and_d_project: req.query.project,
+    })
+
+    req.flash('success', 'Investment details updated')
+    res.redirect(`/investment-projects/${req.params.investmentId}/details`)
```
feedmypixel: `req.params.investmentId` could go in local `const`? its used a few times

src/apps/investment-projects/controllers/associated.js
```diff
@@ -0,0 +1,89 @@
...
+  const page = req.query.page || '1'
+  const token = req.session.token
+
+  if (!searchTerm) {
+    return next()
+  }
+
+  try {
+    const investmentTypes = metadata.investmentTypeOptions
+    const nonFDI = find(investmentTypes, item => item.name.toLowerCase() === 'non-fdi')
```
feedmypixel: can you add brackets and drop it onto a second line. To align with the rest of the codebase please?

src/apps/investment-projects/controllers/associated.js
```diff
@@ -0,0 +1,89 @@
...
+
+    res.locals.searchTerm = searchTerm
+
+    res.locals.results = await searchInvestments({
+      token,
+      searchTerm,
+      page,
+      filters: {
+        investment_type: nonFDI.id,
```
feedmypixel: feels like the `id` could be assigned above if the rest of `nonFDI` isn't being used?

src/apps/investment-projects/services/formatting.js
```diff
@@ -181,7 +177,39 @@ function transformInvestmentValueForView (data) {
     business_activities: data.business_activities.filter((activity) => {
       return /^(european|global) headquarters$/i.test(activity.name)
     }).length ? 'Yes' : 'No',
-  })
+  }
+
+  if (data.non_fdi_r_and_d_budget) {
+    value.associated_non_fdi_r_and_d_project = transformAssociatedProject(data)
+  } else {
+    value.associated_non_fdi_r_and_d_project = 'Not linked to a non-FDI R&D project'
```
feedmypixel: can this be done in the view? `data || Not linked to a non-FDI R&D project`

src/apps/investment-projects/views/associated.njk
```diff
@@ -0,0 +1,25 @@
...
+
+  {% if results | length %}
+    {{
+      Collection(results | assign({
+        highlightTerm: searchTerm,
+        countLabel: 'project',
+        query: QUERY
+      }))
+    }}
+
```
feedmypixel: random newline

src/apps/investment-projects/views/details.njk
```diff
@@ -32,7 +32,7 @@
   {% endif %}
 
   <h2 class="heading-medium">Value</h2>
-  {% if values | length %}
+  {% if values | length > 1 %}
```
feedmypixel: is this needed? `length` filter is not 0 based

src/apps/investment-projects/views/details.njk
```diff
@@ -20,7 +20,7 @@
   {{ renderButton('details', 'Edit summary', true) }}
 
   <h2 class="heading-medium">Requirements and location</h2>
-  {% if requirements | removeNilAndEmpty | length %}
+  {% if requirements | removeNilAndEmpty | length > 1 %}
```
feedmypixel: is this needed? `length` filter is not 0 based

src/apps/search/services.js
```diff
@@ -99,14 +101,38 @@ function searchCompanies ({ token, searchTerm, isUkBased, page = 1, limit = 10 }
...
+  })
+
+  const options = {
+    url: `${config.apiRoot}/v3/search/investment_project?${queryString.stringify(queryParams)}`,
+    method: 'POST',
+    body,
+  }
+
+  return authorisedRequest(token, options)
+    .then(result => {
```
feedmypixel: braces missing

src/apps/investment-projects/views/details.njk
ztolley: It’s needed as there will now always be at least 1 entry now, and if there is then the info message is not displayed so it’s there to keep the behaviour consistent with what it does now.

src/apps/investment-projects/views/details.njk
ztolley: Yes needed

src/apps/investment-projects/controllers/associated.js
ztolley: Not sure how and that would just be doing the same thing on a different line.

src/apps/investment-projects/services/formatting.js
ztolley: The view just calls the key value table to render this, so it can't be done in the view.

`{% component 'key-value-table', items=values, variant='striped' %}`

src/apps/search/services.js
ztolley: Direct copy of the syntax used on line 97 above

ztolley:
I todo would get lost and has no visibility in planning. Let's get a card in the backlog and use user feedback to get product management to make it a priority to review investment as a whole.

src/apps/search/services.js
feedmypixel: > Direct copy of the syntax used on line 97 above

Ok, can align both then? As an aside we should have braces around arguments on arrow functions in eslint really. Maybe thats part of another PR.

Always feel free to update things as you see them.

feedmypixel:
> A todo would get lost and has no visibility in planning. Let's get a card in the backlog and use user feedback to get product management to make it a priority to review investment as a whole.

We could do both of those things? As either could get lost?

ztolley:
OK

src/apps/investment-projects/views/details.njk
feedmypixel: hmmmm something feels a little funky here

src/apps/investment-projects/controllers/associated.js
feedmypixel:
```
const nonFDIId = get(find(investmentTypes, (item) => {
  return item.name.toLowerCase() === 'non-fdi'
}), 'id', null)
```

ztolley: I find that harder to read when scanning code

src/apps/investment-projects/views/details.njk
ztolley: Will revisit and see if I can de-funk it but can’t promise

src/apps/investment-projects/controllers/associated.js
feedmypixel:
```
const { id: nonFDIId } = find(investmentTypes, (item) => {
  return item.name.toLowerCase() === 'non-fdi'
})
```

ztolley: better

src/apps/investment-projects/views/details.njk
ztolley: Had a look, afraid it will have to stay this way. This ticket and the previous one mean there is always at least one, but we only want to display if there is > 1. Will have to stay like this for now.

src/apps/investment-projects/controllers/associated.js
feedmypixel: 🙇

src/apps/investment-projects/views/associated.njk
feedmypixel: @ztolley has this been done? I'm not seeing it?

feedmypixel:
> A todo would get lost and has no visibility in planning. Let's get a card in the backlog and use user feedback to get product management to make it a priority to review investment as a whole.

@ztolley has this TODO and card been done? can you drop the Jira number on please?

ztolley:
DH-1030

src/apps/investment-projects/views/details.njk
feedmypixel: Sorry I meant `{% if requirements | removeNilAndEmpty | length > 1 %}` this feels like it's grown or `requirements` could be santized before and just length is checked here?

Just this code feels like somewhere along the chain it could do with a tidy up. Thats just a hunch and something for a future PR

feedmypixel:
> DH-1030

For future reference this is the Jira ticket around the tidy up work needed around this functionality

src/apps/investment-projects/views/details.njk
ztolley: it could be sanitised to remove the nil in the controller and still check the length in view

ztolley: Though to do that I’d actially call the same function, just in the controller instead of the view, and yes you can argue that’s the right place to do it though starting to creep beyond the scope of the card

ztolley: Yeah had a look, it's in a nujucks filter in config.... it should be moved and thats a big enough thing to be it's own pr.

PR #857, opened 2017-10-31
Fixed missing page param searching company

ztolley:
When a user searches for a UK company receiving investment the search call was not passing the page number parameter from the request query, so when the user selected page 2 it still called the server with page 1.

Pass the req.query.page to the search call

### DIFF
--- a/src/apps/investment-projects/controllers/ukcompany.js
+++ b/src/apps/investment-projects/controllers/ukcompany.js
@@ -34,6 +34,7 @@ async function searchForUKCompany (req, res, next) {
     res.locals.results = await searchCompanies({
       token,
       searchTerm,
+      page: req.query.page,
       isUkBased: true,
     }).then(
       transformApiResponseToSearchCollection(
--- a/test/unit/apps/investment-projects/controllers/ukcompany.test.js
+++ b/test/unit/apps/investment-projects/controllers/ukcompany.test.js
@@ -107,40 +107,64 @@ describe('investment uk company', () => {
 
   describe('#searchForUKCompany', () => {
     context('when a search term is provided', () => {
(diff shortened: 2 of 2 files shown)


feedmypixel:
Not sure why this fix/PR has been added? is this functionality not covered in the method your calling? [searchCompanies](https://github.com/uktrade/data-hub-frontend/blob/a34defff41600fc8ad8a71317835a9696ada84a0/src/apps/search/services.js#L51) by the default for the `page` argument? `page = 1 `

ztolley:
Yes, there was a bug in which the page param wasn't being passed to the search call.

feedmypixel:
> Yes, there was a bug in which the page param wasn't being passed to the search call.

Ok got yah. Tired eyes on this last night! The only thing really then is you don't need to do the [default](https://github.com/uktrade/data-hub-frontend/pull/857/files#diff-e849dde18e71d5741bd7871fe6a67719R26) as it is being handled in the [searchCompanies](https://github.com/uktrade/data-hub-frontend/blob/04223fe1ee23a62e99f5f3889ea4490ebda3157d/src/apps/search/services.js#L79)  method?

ztolley:
Ok. Would rather leave it in just in case. Will rebase and merge later this morning

feedmypixel:
> Ok. Would rather leave it in just in case. Will rebase and merge later this morning

Why its covered in the `searchCompanies` method? That doesn't really make any sense? Why have 2 defaults?

teneightfive:
>Would rather leave it in just in case.

@ztolley I kind of agree with @feedmypixel on this. There's no point adding the default in 2 places.

Just passing the `req.query.page` through to the `searchCompanies` service should be enough.

PR #858, opened 2017-10-31
Changed the name of our Zen service

jim68000:
...and create yet another env var so we can change it without code in the future

### DIFF
--- a/config/index.js
+++ b/config/index.js
@@ -42,6 +42,7 @@ const config = {
   zenBrowser: process.env.ZEN_BROWSER,
   zenImpact: process.env.ZEN_IMPACT,
   zenService: process.env.ZEN_SERVICE,
+  zenServiceNickname: process.env.ZEN_SERVICE_CHANNEL || 'datahub',
   sentryDsn: process.env.SENTRY_DSN,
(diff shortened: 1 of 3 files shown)


config/index.js
```diff
@@ -42,6 +42,7 @@ const config = {
   zenBrowser: process.env.ZEN_BROWSER,
   zenImpact: process.env.ZEN_IMPACT,
   zenService: process.env.ZEN_SERVICE,
+  zenServiceNickname: process.env.ZEN_SERVICE_NICKNAME || 'datahub',
```
feedmypixel: what does `nickname` mean? can you give it something more semantic? if not worth dropping a comment in?

teneightfive: Should also update the [README](https://github.com/uktrade/data-hub-frontend#docker-composeyml) which gives an explanation of our envs.

jim68000: How about CHANNEL? The var gets used here:

`      { id: config.zenService, value: config.zenServiceNickname },
`

So it's the value of the zenService var, which is a number? 

And good spot on the README, that might be a bit out of date - I will update for this var, and then do a separate docs PR for the other vars that are missing.

PR #862, opened 2017-11-01
Added missing env vars and defaults

jim68000:


feedmypixel:
@jim am merging this as I need to add env details around SSO work

PR #867, opened 2017-11-01
Format telephone number to be displayed as entered

yeahfro:
This was discovered when we have intermittent failures with acceptance tests. The formatting of telephone numbers is assuming that a leading zero should be removed.

Reason for removing:
- code is causing edge case problems
- not all cases are being covered

This should be revisited with Miles.


feedmypixel:
LGTM - personally I see no need for the [phone.js](https://github.com/uktrade/data-hub-frontend/pull/867/files#diff-deaddbe494f326ebb93bc7a718266b97) file anymore. I'd get rid and move the tests into the [contacts/services/formatting.test.js](https://github.com/uktrade/data-hub-frontend/blob/develop/test/unit/apps/contacts/services/formatting.test.js) file

feedmypixel:
> Do we know what data from cdms looks like?

What data exactly?

PR #871, opened 2017-11-02
Order cancellation

teneightfive:
Add the ability to cancel and OMIS order and give a reason for the cancellation.

## What it looks like

localhost_3001_omis_f986e5e3-b0b7-4831-a066-a6eb1f1bb902_edit_cancel-order

localhost_3001_omis_f986e5e3-b0b7-4831-a066-a6eb1f1bb902_work-order 1

### Todo

- [x] Ensure API returns correct set of reasons
- [x] Ensure correct body is sent to cancellation API endpoint

### DIFF
--- a/src/apps/omis/apps/edit/controllers/cancel-order.js
+++ b/src/apps/omis/apps/edit/controllers/cancel-order.js
@@ -1,0 +1,57 @@
+const { get, filter, merge, pick } = require('lodash')
+
+const { EditController } = require('../../../controllers')
+const { transformObjectToOption } = require('../../../../transformers')
+const { Order } = require('../../../models')
+const metadataRepo = require('../../../../../lib/metadata')
+
+class CancelOrderController extends EditController {
+  async configure (req, res, next) {
+    const orderStatus = get(res.locals, 'order.status')
+
+    if (!['draft', 'quote_awaiting_acceptance'].includes(orderStatus)) {
+      return res.redirect(`/omis/${res.locals.order.id}`)
+    }
+
+    const filteredReasons = filter(metadataRepo.orderCancellationReasons, (reason) => {
+      return !reason.disabled_on
+    })
+
+    req.form.options = merge({}, req.form.options, {
+      disableFormAction: false,
+      returnText: 'Return without cancelling',
+      buttonText: 'Cancel order',
+      buttonModifiers: 'button--destructive',
+      fields: {
+        cancellation_reason: {
+          options: filteredReasons.map(transformObjectToOption),
+        },
+      },
+    })
+
+    super.configure(req, res, next)
+  }
+
+  async successHandler (req, res, next) {
+    const token = req.session.token
+    const orderId = get(res.locals, 'order.id')
+    const data = pick(req.sessionModel.toJSON(), Object.keys(req.form.options.fields))
+
+    try {
+      await Order.cancel(token, orderId, data)
+      const nextUrl = req.form.options.next || `/omis/${orderId}`
+
+      req.journeyModel.reset()
+      req.journeyModel.destroy()
+      req.sessionModel.reset()
+      req.sessionModel.destroy()
+
+      req.flash('success', 'Order cancelled')
+      res.redirect(nextUrl)
+    } catch (error) {
+      next(error)
+    }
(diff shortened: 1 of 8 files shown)


src/apps/omis/apps/edit/controllers/cancel-order.js
```diff
@@ -0,0 +1,58 @@
...
+          options: filteredReasons.map(transformObjectToOption),
+        },
+      },
+    })
+
+    super.configure(req, res, next)
+  }
+
+  async successHandler (req, res, next) {
+    const token = get(req.session, 'token')
```
ztolley: isn't a given there is a token to reach here?

src/apps/omis/apps/edit/controllers/cancel-order.js
```diff
@@ -0,0 +1,58 @@
...
+    const orderId = get(res.locals, 'order.id')
```
ztolley: and if there is no order id?

src/apps/omis/apps/edit/fields.js
```diff
@@ -205,6 +205,13 @@ const editFields = merge({}, globalFields, {
     optional: true,
     modifier: 'medium',
   },
+  cancellation_reason: {
+    fieldType: 'MultipleChoiceField',
+    label: 'fields.cancellation_reason.label',
+    type: 'radio',
+    options: [],
```
ztolley: How do the options get populated?

src/apps/omis/apps/edit/controllers/cancel-order.js
teneightfive: Nothing is a given with this app ;)

src/apps/omis/apps/edit/controllers/cancel-order.js
teneightfive: Then the try block would fail and `catch` would catch the error.

src/apps/omis/apps/edit/fields.js
teneightfive: It gets set in the `configure` lifecycle method: https://github.com/uktrade/data-hub-frontend/pull/871/files#diff-89d5773cfbe397bce0dcd610daefbd4fR28

src/apps/omis/apps/edit/controllers/cancel-order.js
ztolley: I know I'm just saying, we've not used this anywhere else.

teneightfive: Yeah I know. I'm probably just being overprotective. Could actually just use normal dot notation.

src/apps/omis/apps/edit/controllers/cancel-order.js
```diff
@@ -0,0 +1,58 @@
...
+      },
+    })
+
+    super.configure(req, res, next)
+  }
+
+  async successHandler (req, res, next) {
+    const token = req.session.token
+    const orderId = get(res.locals, 'order.id')
+    const data = pick(req.sessionModel.toJSON(), Object.keys(req.form.options.fields))
```
yeahfro: There is a lodash `keys`

src/apps/omis/apps/edit/controllers/cancel-order.js
```diff
@@ -0,0 +1,58 @@
...
+class CancelOrderController extends EditController {
+  async configure (req, res, next) {
+    const orderStatus = get(res.locals, 'order.status')
+
+    if (!['draft', 'quote_awaiting_acceptance'].includes(orderStatus)) {
+      return res.redirect(`/omis/${res.locals.order.id}`)
+    }
+
+    const filteredReasons = filter(metadataRepo.orderCancellationReasons, (reason) => {
+      if (reason.disabled_on) { return false }
```
yeahfro: `return !reason.disabled_on` would be nice

src/apps/omis/apps/edit/controllers/cancel-order.js
teneightfive: Using similar statements without `keys` elsewhere so going to leave this one for now.

PR #878, opened 2017-11-03
Investment investor

ztolley:
Adds 3 new fields to the investment summary form and details view: 'Investor type', 'Level of involvement' and 'Specific investment programme'.

screencapture-localhost-3000-investment-projects-d886a660-5268-4149-823b-5805029e0639-edit-details-1509714659896

screencapture-localhost-3000-investment-projects-d886a660-5268-4149-823b-5805029e0639-details-1509716007152

### DIFF
--- a/src/apps/investment-projects/labels.js
+++ b/src/apps/investment-projects/labels.js
@@ -15,6 +15,9 @@ const labels = {
       description: 'Project description',
       anonymous_description: 'Anonymised description',
       estimated_land_date: 'Estimated land date',
+      investor_type: 'New or existing investor',
+      level_of_involvement: 'Level of involvement',
+      specific_programme: 'Specific investment programme',
     },
   },
   valueLabels: {
--- a/src/apps/investment-projects/middleware/forms/details.js
+++ b/src/apps/investment-projects/middleware/forms/details.js
@@ -9,6 +9,7 @@ const {
   transformObjectToOption,
   transformContactToOption,
 } = require('../../../transformers')
+const { filterDisabledOption } = require('../../../filters')
 const {
   createInvestmentProject,
   getEquityCompanyDetails,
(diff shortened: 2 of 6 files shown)


src/apps/investment-projects/labels.js
```diff
@@ -15,6 +15,9 @@ const labels = {
       description: 'Project description',
       anonymous_description: 'Anonymised description',
       estimated_land_date: 'Estimated land date',
+      investor_type: 'New / Existing Investor',
```
teneightfive: `New or existing investor`? Doesn't need caps and slashes are bad for being read out, ie screenreaders.

src/apps/investment-projects/labels.js
```diff
@@ -15,6 +15,9 @@ const labels = {
...
+      level_of_involvement: 'Level of Involvement',
```
teneightfive: No need for random capitalised words.

src/apps/investment-projects/labels.js
```diff
@@ -15,6 +15,9 @@ const labels = {
...
+      specific_programme: 'Specific Investment Programme',
```
teneightfive: No need for random capitalised words.

src/apps/investment-projects/middleware/forms/details.js
```diff
@@ -41,7 +42,7 @@ async function populateForm (req, res, next) {
     res.locals.equityCompany = equityCompany
     res.locals.equityCompanyInvestment = equityCompanyInvestment
     res.locals.form = res.locals.form || {}
-    res.locals.form.state = assign({}, {
+    const state = res.locals.form.state = assign({}, {
```
teneightfive: Maybe move outside of where `res.locals` is defined? Consts at the top is easier to read.

src/apps/investment-projects/views/_details-form.njk
```diff
@@ -297,4 +297,34 @@
     }
   }) }}
 
+  {{ MultipleChoiceField({
+    name: 'investor_type',
+    label: 'New / Existing Investor',
```
teneightfive: Why isn't this using the label defined in labels?

src/apps/transformers.js
```diff
@@ -97,6 +97,18 @@ function transformApiResponseToCollection (options = {}, ...itemTransformers) {
...
 
+/**
+ *
+ * @param {string} currentValue - The current value for a field so that the options filtered include the current
+ *                                even if invalid
+ * @returns {function} - a filter function that will filter out inactive or non-current values
+ */
+function filterInactive (currentValue) {
+  return function (item) {
+    return !item.disabled_on || item.id === currentValue
```
teneightfive: Should `disabled_on` also not be checking date? Items could be disabled for a date in the future?

src/apps/transformers.js
```diff
@@ -97,6 +97,18 @@ function transformApiResponseToCollection (options = {}, ...itemTransformers) {
   }
 }
 
+/**
+ *
+ * @param {string} currentValue - The current value for a field so that the options filtered include the current
+ *                                even if invalid
+ * @returns {function} - a filter function that will filter out inactive or non-current values
+ */
+function filterInactive (currentValue) {
```
teneightfive: Maybe call this what it is doing, `filterDisabledItems`. It also doesn't necessarily feel like a transformer?

src/apps/investment-projects/labels.js
ztolley: yeah, my cut and paste bad.

src/apps/investment-projects/views/_details-form.njk
ztolley: To keep it consistent with the rest of the form which in turn needs to be refactored to use the form builder and switch to use them then.

src/apps/transformers.js
ztolley: You are right it's not a transformer.

src/apps/transformers.js
jim68000: I'd rather not implement that functionality - the date is just for the admins to see how old a piece of disabled metadata is so they can make decisions about purging records.

teneightfive: That's fair. Was more of a question as to whether that's what it should be doing.

src/apps/investment-projects/middleware/forms/details.js
```diff
@@ -37,14 +38,16 @@ async function populateForm (req, res, next) {
...
+
     res.locals.clientCompanyId = clientCompanyId || equityCompanyId
     res.locals.equityCompany = equityCompany
     res.locals.equityCompanyInvestment = equityCompanyInvestment
     res.locals.form = res.locals.form || {}
-    res.locals.form.state = assign({}, {
-      client_contacts: [''],
-      business_activities: [''],
-    }, investmentData)
+    res.locals.form.state = state
```
teneightfive: Wonder if this could be combined with line above to:

```js
  state,
})
```

Would reduce repetition of `res.locals` now that `state` is a const.

teneightfive: Maybe even one step neater and bring line 52 in as well: 

```js
  state,
  options: {
    ...
  },
})
```

ztolley: I prefer the way it is now, easier to read the code

teneightfive: I would disagree in that there's no a lot of repetition on adjacent lines. Repetition is harder to read.

teneightfive: Maybe options could be harder so what about defining options as a const like state then the setting of form would just be:

```js
  state,
  options,
})
```

Can't deny that's easier to read than three lines all starting with `res.locals.form`?

ztolley: I'll have to disagree and say it's easier to read x = y rather than looking through an assign statement.

Will change it for this PR but worry we overuse assign.

PR #818, opened 2017-10-24
Fix/company issues

feedmypixel:
A number of issues around companies have been fixed in this PR

- Hardcode "United Kingdom" for nonForeign companies
- Turn on and update `@companies-create` UAT tests
- Fix naming used on address fields
- Make `(optional)` from mandatory inputs
- Only show PostCode Lookup when JavaScript is available
- Replace `name` with `trading_name` in company results when it exists
- Added a few `// TODO` around work that can be addressed separately

And lots of other little tweaks around company creation.

PR #880, opened 2017-11-04
Refactor: form group and stop using it in wrong situation

feedmypixel:
I have been meaning to get back to this since I changed it in [#818](https://github.com/uktrade/data-hub-frontend/pull/818)

Essentially I was using the `FormGroup` macro for its styling.

This work:
-  uses `MetaList` and vanilla markup. There is potential for a new macro if we see a pattern of `Heading` -> `Text` -> `hidden input`.
- reinstates a `text-field.test.js`
- reinstates the fact you can only use `FormGroup` when it has a containing `input` element
- removes `FormGroup` from the import in `datahub-base.njk` essentially saying this should only be used by proxying through a higher level macro.

### Business type on add/edit company flow
screen shot 2017-11-04 at 12 31 05

### United Kingdom value on add/edit company flow
screen shot 2017-11-04 at 12 31 11

### DIFF
--- a/src/apps/companies/views/edit.njk
+++ b/src/apps/companies/views/edit.njk
@@ -4,18 +4,19 @@
   <button class="button button-secondary postcode-lookup-button" type="button">Find UK address</button>
 {% endset %}
 
+{% set businessType %}
+  {{ 'Foreign' if isForeign else 'UK' }}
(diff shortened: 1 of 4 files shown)


feedmypixel:
Not sure why this has closed? apparently I did an hour ago (I was asleep)? hmmmm

PR #884, opened 2017-11-06
Interaction filters

ztolley:
Minor refactor for interaction collections middleware to split out the 3 common jobs it needs to do:
- Parse request body and parameters
- Generate form for sorting
- Get collection data

Added support to the interaction list controller and view to introduce a filter form.

interactions

### DIFF
--- a/src/apps/interactions/macros.js
+++ b/src/apps/interactions/macros.js
@@ -4,6 +4,8 @@ const formLabels = require('./labels')
 const metaData = require('../../lib/metadata')
 const { transformContactToOption, transformObjectToOption } = require('../transformers')
 
+const currentYear = (new Date()).getFullYear()
+
 const interactionSortForm = {
   method: 'get',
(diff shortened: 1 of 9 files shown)


src/apps/interactions/macros.js
```diff
@@ -120,6 +122,47 @@ const interactionFields = {
...
+      placeholder: `e.g. ${currentYear}-07-18`,
+    },
+    {
+      macroName: 'TextField',
+      name: 'date_before',
+      hint: 'YYYY-MM-DD',
+      placeholder: `e.g. ${currentYear}-07-21`,
+    },
+  ].map(filter => {
+    return assign(filter, {
```
yeahfro: `assign` to empty object

yeahfro: `assign` to empty object

ztolley: all the other uses follow this pattern as it's just adding properties to a new object.

PR #891, opened 2017-11-08
Fix issue generating company url in collections

ztolley:
ltd

If a company contains companies house data the collections transformer
was giving a url for the ch screen instead of the full company details
screen. The CH link should only happen when the record is ch data only.

### DIFF
--- a/test/unit/apps/companies/transformers.test.js
+++ b/test/unit/apps/companies/transformers.test.js
@@ -6,11 +6,14 @@ describe('Company transformers', function () {
   describe('#transformCompanyToListItem', () => {
     const companyData = require('~/test/unit/data/company')
 
-    it('should return undefined for unqualified result', () => {
+    it('should return undefined if there is no companies house data or datahub data', () => {
       expect(transformCompanyToListItem()).to.be.undefined
       expect(transformCompanyToListItem({ a: 'b' })).to.be.undefined
-      expect(transformCompanyToListItem({ id: 'abcd' })).to.be.undefined
       expect(transformCompanyToListItem({ first_name: 'Peter', last_name: 'Great' })).to.be.undefined
+    })
+
+    it('should return undefined if companies house is incomplete', () => {
+      expect(transformCompanyToListItem({ companies_house_data: {} })).to.be.undefined
     })
 
     it('should return an object with data for company list item', () => {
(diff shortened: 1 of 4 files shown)


test/unit/apps/companies/transformers.test.js
```diff
@@ -6,13 +6,6 @@ describe('Company transformers', function () {
   describe('#transformCompanyToListItem', () => {
     const companyData = require('~/test/unit/data/company')
 
-    it('should return undefined for unqualified result', () => {
```
teneightfive: Do we no longer want any kind of check to filter out duff results? Surely at least still a name? Otherwise they'd be an empty row in collections.

ztolley: Thre would be no name for ch results, and the tricky bit is also there are is no way to see what a ch only result looks like because we apparently don't search ch data.

I will add a check to make sure the record has either id or companies_housedata

teneightfive: So what gets returned as the `name` property to Companies House companies?

teneightfive: Yes maybe just add a check for `companies_house_data` to the initial check. So it either needs an ID, a name or some data from companies house.

ztolley: No idea what gets returned as there is no way to get a ch only entry right now

teneightfive: Ah. Well that's a slight issue. Maybe just check for those things for now then?

ztolley: yup. I'm kinda more worried about what it has revealed

test/unit/apps/companies/transformers.test.js
```diff
@@ -9,7 +9,7 @@ describe('Company transformers', function () {
     it('should return undefined for unqualified result', () => {
       expect(transformCompanyToListItem()).to.be.undefined
       expect(transformCompanyToListItem({ a: 'b' })).to.be.undefined
-      expect(transformCompanyToListItem({ id: 'abcd' })).to.be.undefined
+      expect(transformCompanyToListItem({ companies_house_data: {} }))
```
teneightfive: `.to.be.undefined`?

teneightfive: No, it should be undefined because nothing is returned because it doesn't pass valid props.

This change doesn't test anything.

test/unit/apps/companies/transformers.test.js
```diff
@@ -6,13 +6,16 @@ describe('Company transformers', function () {
...
-    it('should return undefined for unqualified result', () => {
+    it('should return undefined if there is no companies house data or datahub data', () => {
       expect(transformCompanyToListItem()).to.be.undefined
       expect(transformCompanyToListItem({ a: 'b' })).to.be.undefined
-      expect(transformCompanyToListItem({ id: 'abcd' })).to.be.undefined
       expect(transformCompanyToListItem({ first_name: 'Peter', last_name: 'Great' })).to.be.undefined
     })
 
+    it('should return undefined if companies house is incomplete', () => {
+      expect(transformCompanyToListItem({ companies_house_data: {} }))
```
teneightfive: This doesn't test anything still...you need to give it an assertion to test, no?

PR #896, opened 2017-11-08
Hidden text (progressive disclosure) component

teneightfive:
This change adds support for the [hidden text (progressive disclosure)](https://govuk-elements.herokuapp.com/typography/#typography-hidden-text) component to the app.

It adds the examples shown below to the component overview pages.

## What it looks like 

localhost_3000_components_hidden-text 2

### DIFF
--- a/assets/javascripts/vendor/details.polyfill.js
+++ b/assets/javascripts/vendor/details.polyfill.js
@@ -1,0 +1,190 @@
+// <details> polyfill
+// http://caniuse.com/#feat=details
+
+// FF Support for HTML5's <details> and <summary>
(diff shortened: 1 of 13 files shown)


teneightfive:
I question around this work that would be good to get thoughts on:

- **Naming**
  - Styles I've added to `elements` called `details.scss` as its a `<details>` element.
  - Macro is called `Reveal`
  - Example is hidden text (taken from govuk elements example pages)

I'm sure whether this is confusing using lots of names or if it makes sense. It sort of made sense at the time of adding but be good to get other thoughts on it.

yeahfro:
I am not sure about the name `Reveal` as it's a verb and we have been using nouns. Could we use `Revealer`?

PR #898, opened 2017-11-08
Display quote terms and conditions

teneightfive:
This adds work around showing the full quote terms and conditions for OMIS orders.

## What it looks like

localhost_3000_omis_712d3118-fe8d-41bd-ae49-1a10eaf5c64e_quote

localhost_3000_omis_712d3118-fe8d-41bd-ae49-1a10eaf5c64e_quote 1

### DIFF
--- a/assets/stylesheets/trumps/_layout.scss
+++ b/assets/stylesheets/trumps/_layout.scss
@@ -56,4 +56,21 @@
   * + * {
     margin-top: $default-spacing-unit;
   }
+
+  ul, ol {
+    padding-left: 1em;
+  }
+
+  ul {
+    list-style-type: disc;
+  }
+
+  ol {
+    list-style-type: decimal;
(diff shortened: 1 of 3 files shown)


assets/stylesheets/trumps/_layout.scss
```diff
@@ -56,4 +56,24 @@
   * + * {
     margin-top: $default-spacing-unit;
   }
+
+  ul, ol {
```
ztolley: Have you checked the rest of the site to see if this style affects anything else? never a fan of global changes.

teneightfive: It's not a global change. This sits within the `.l-markdown` selector. (which was a bit of a hack because our styles are so fragile).

I needed to apply better spacing and style to formatted markdown as we don't have as much control of adding classes etc like we have to elsewhere at the moment.

PR #899, opened 2017-11-08
Turn off workers in nightwatch

feedmypixel:
I have a sneaky feeling this may be causing us some race condition issues that we are seeing with data being manipulated across different tests.

Tests expect 2 results but there are actually 3 - as seen here:
https://10463-83823675-gh.circle-artifacts.com/1/root/data-hub-frontend/cucumber/report.html


feedmypixel:
@yeahfro @teneightfive FYI for this PR

PR #900, opened 2017-11-08
Add acceptance tests around contact sorting

yeahfro:
This change will add acceptance tests around contact sorting.

Not covered (next PR):
- newest/oldest
- last name (potential bug)

### DIFF
--- a/test/acceptance/features/contacts/page-objects/ContactList.js
+++ b/test/acceptance/features/contacts/page-objects/ContactList.js
@@ -17,6 +17,15 @@ const getMetaListItemValueSelector = (text) => getSelectorForElementWithText(
   }
 )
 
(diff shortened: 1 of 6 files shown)


test/acceptance/features/contacts/page-objects/ContactList.js
```diff
@@ -17,6 +17,15 @@ const getMetaListItemValueSelector = (text) => getSelectorForElementWithText(
   }
 )
 
+const getCountryBadgeSelectorWithText = () => getSelectorForElementWithText(
```
ztolley: What makes this unique to country badge? Could it be getBadgeWithText and be reusable?

yeahfro: Good point I think I got a little carried away with this one!

PR #904, opened 2017-11-10
Set getAdvisers to get all advisers

ztolley:
Previously we had 2 methods to get a list of advisers when only one is needed and that should supply all the advisers.

### DIFF
--- a/src/apps/adviser/repos.js
+++ b/src/apps/adviser/repos.js
@@ -2,13 +2,9 @@ const config = require('../../../config')
 const logger = require('../../../config/logger')
 const authorisedRequest = require('../../lib/authorised-request')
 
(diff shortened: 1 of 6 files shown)


src/apps/adviser/repos.js
```diff
@@ -3,12 +3,6 @@ const logger = require('../../../config/logger')
 const authorisedRequest = require('../../lib/authorised-request')
 
 function getAdvisers (token) {
-  return authorisedRequest(token, `${config.apiRoot}/adviser/`)
-}
-
-// TODO: Make sure this is removed and replaced with a better way to
```
feedmypixel: please keep the `TODO` in here, the `/?limit=100000&offset=0` is a temporary fix

PR #909, opened 2017-11-10
Fix companies filtering on Firefox

richtier:
[this task](https://uktrade.atlassian.net/browse/DH-974)

todo:
 - [x] make forms explicitly recieve action prop

## The cause

The urls generated upon filtering are different:

browser| url
------------ | -------------
Chrome | http://localhost:3000/companies?custom=true&name=cool&sector=af22c9d2-5f95-e211-a939-e4115bead28a&sortby=modified_on%3Aasc
Firefox | http://localhost:3000/companies?custom=true&name=cool&sector=af22c9d2-5f95-e211-a939-e4115bead28a&sortby=modified_on%3Aasc?custom=true&name=cool&sector=af22c9d2-5f95-e211-a939-e4115bead28a&sortby=modified_on%3Aasc

The cause of this different behaviour is the HTML `<form>` elements were being rendered without the `action` attribute. Firefox and Chrome do different things in this situation:

browser| behaviour
------------ | -------------
Chrome | treat the form element's action attribute as an empty string
Firefox | treat the form element's action attribute as the current url (including the qureystring)

This affect's how the [`auto-submit`](https://github.com/uktrade/data-hub-frontend/blob/develop/assets/javascripts/modules/auto-submit.js#L37) will generate the url to submit the form:

browser| behaviour
------------ | -------------
Chrome | concatenate the empty string with the query, resulting in a valid url
Firefox | concatenate the current url (including the querystring) with the query, resulting in an invalid url with duplicate query fields, which the server does not comprehend.

## The fix

Setting the form element's action attribute to the current url (without the querystring).

### DIFF
--- a/src/templates/_macros/form/form.njk
+++ b/src/templates/_macros/form/form.njk
@@ -32,7 +32,7 @@
 
   <form
     method="{{ method }}"
-    {% if props.action %}action="{{ props.action }}"{% endif -%}
+    action="{{ props.action }}"
     {% if props.class %}class="{{ props.class }}"{% endif %}
     {% if props.role %}role="{{ props.role }}"{% endif %}
   >
--- a/src/apps/companies/views/interactions.njk
+++ b/src/apps/companies/views/interactions.njk
@@ -16,7 +16,7 @@
       Collection(interactions | assignCopy({
         countLabel: 'interaction',
         summaryActionsHTML: addButton,
-        sortForm: sortForm,
+        sortForm: assign({}, sortForm, {action: CURRENT_PATH}),
         query: QUERY
       }))
     }}
--- a/src/apps/companies/views/list.njk
+++ b/src/apps/companies/views/list.njk
@@ -4,7 +4,8 @@
   {{
     CollectionFilters({
       query: QUERY,
-      filtersFields: filtersFields
+      filtersFields: filtersFields,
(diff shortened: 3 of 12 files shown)


src/templates/_macros/form/form.njk
```diff
@@ -24,6 +24,7 @@
  #}
 {% macro Form(props) %}
   {% set method = props.method or 'POST' -%}
+  {% set action = props.action or '?' -%}
```
teneightfive: I'm not sure about always setting every form to `?` by default if action is empty. What about setting the action to `?` for the collection forms which are causing the errors?

richtier: to be clear: it's not one specific form causing the error: it's *all forms that do not have an `action` in firefox*.

we could update all forms that don't have `action` set, but then that invites reintroducing the error in future when someone forgets the firefox behaviour and does not set `action` for a form.

This change ensures consistent handling of "empty" `action` attributes across browsers.

feedmypixel: Rather than setting it to `?` can we set the action to the current page? which would be the explicit version of not having an action

feedmypixel: if we then make `action` required this wouldn't happen again

feedmypixel:
👍

PR #915, opened 2017-11-13
add copy to help page,

feedmypixel:
Add copy to the no scope service page to redirect users to the support form.

### Use case
A user who is logged in on SSO but doesn't have the scope set for `x` service and is trying to access it. 

### Screenshot
screen shot 2017-11-13 at 13 34 48

### Support page radio button
screen shot 2017-11-13 at 15 34 55

## Flow
contactliveservives

### DIFF
--- a/src/apps/oauth/views/help-page.njk
+++ b/src/apps/oauth/views/help-page.njk
@@ -1,5 +1,5 @@
 {% extends "_layouts/datahub-base.njk" %}
 
 {% block body_main_content %}
-  <p>If you think you should have access or need to sign up to Data Hub then get in touch <a href="/support">using this support form</a>.</p>
+  <p>If you think you should have access or need to sign up to Data Hub then get in touch <a href="/support">using the support form</a>.</p>
 {% endblock %}
--- a/src/apps/oauth/views/help-page.njk
+++ b/src/apps/oauth/views/help-page.njk
@@ -1,12 +1,5 @@
 {% extends "_layouts/datahub-base.njk" %}
 
 {% block body_main_content %}
-  {# TODO add live service contact details once supplied from content and design #}
-
-  <h2 class="heading-medium">{{ errorMessage }}</h2>
-  {{ MetaList({
-    items: [
-      { label: 'Error Code', value: errorCode }
-    ]
-  }) }}
+  <p>If you think you should have access or need to sign up to Data Hub then get in touch <a href="/support">using this support form</a>.</p>
 {% endblock %}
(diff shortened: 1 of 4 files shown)


src/apps/oauth/views/help-page.njk
```diff
@@ -1,12 +1,10 @@
...
-
-  <h2 class="heading-medium">{{ errorMessage }}</h2>
-  {{ MetaList({
-    items: [
-      { label: 'Error Code', value: errorCode }
-    ]
-  }) }}
+  <h2 class="heading-medium">You don't have permission to access this service</h2>
+  <p>If you think you should have access or need to sign up to Data Hub then get in touch.</p>
+  Email <a href="{{  'mailto:login.support@uktrade.zendesk.com
```
teneightfive: Can we not link to the support page? 

And extend the support page to include a `I do not have access` in the type option?

Feels like we already have a platform to handle this and `mailto` with a body feels worse. It also doesn't say which page so if this request came in it would be meaningless.

feedmypixel: Yeah I like that idea. Will speak with Mills about this.

feedmypixel:
@teneightfive changes as suggested implemented, thanks for the input

src/apps/oauth/views/help-page.njk
```diff
@@ -1,12 +1,5 @@
...
 {% block body_main_content %}
-  {# TODO add live service contact details once supplied from content and design #}
-
-  <h2 class="heading-medium">{{ errorMessage }}</h2>
-  {{ MetaList({
-    items: [
-      { label: 'Error Code', value: errorCode }
-    ]
-  }) }}
+  <p>If you think you should have access or need to sign up to Data Hub then get in touch <a href="/support">using this support form</a>.</p>
```
teneightfive: `the support form`? We don't have multiple ones.

src/apps/support/macros.js
```diff
@@ -11,16 +11,19 @@ const feedbackFormConfig = (browserInfo) => ({
       macroName: 'MultipleChoiceField',
       type: 'radio',
       name: 'feedbackType',
-      modifier: 'inline',
-      label: 'Problem or feedback?',
+      label: 'What do you want to tell us?',
```
teneightfive: The _tell_ part feels little strange here. They're not telling us anything in this bit. Maybe something more like asking what their problem is related to?

src/apps/oauth/views/help-page.njk
feedmypixel: yup agreed I originally had 
> with support
this a request from Mills

teneightfive:
Only other thing is how do we know what page they were trying to access? Is there a way to capture it and pass to the support form?

feedmypixel:
> Only other thing is how do we know what page they were trying to access? 

This is them trying to access the service and not having the appropriate scope

> Is there a way to capture it and pass to the support form?

This feels like its something for another PR?

PR #917, opened 2017-11-13
Add company number to company edit form for limited companies

richtier:
ch-number

todo:
 - [ ] Add tests

### DIFF
--- a/src/apps/companies/views/edit.njk
+++ b/src/apps/companies/views/edit.njk
@@ -39,9 +39,19 @@
       id: company.id
     }
   }) %}
+
+    {% if isCompaniesHouse %}
+      {{ TextField({
+        name: 'company_number',
+        label: 'Company number',
+        optional: false,
+        value: formData.company_number,
+        error: errors.company_number
+      }) }}
+    {% endif %}
+
     {% if companiesHouseRecord %}
       <input type="hidden" name="name" value="{{ formData.name }}">
-      <input type="hidden" name="company_number" value="{{ formData.company_number }}">
       <input type="hidden" name="registered_address_1" value="{{ formData.registered_address_1 }}">
       <input type="hidden" name="registered_address_2" value="{{ formData.registered_address_2 }}">
       <input type="hidden" name="registered_address_town" value="{{ formData.registered_address_town }}">
@@ -56,6 +66,7 @@
         value: formData.trading_name,
         error: errors.trading_name
       }) }}
+
     {% else %}
       {{ TextField({
         name: 'name',


ztolley:
Can I take a company that doesn't have a companise house number already and isn't marked as a ltd company and add the company number?

Also, there's alot of companies in the db that dont say they are a ltd company but are and you'd want to set their companies house to indicate this.

Check how it figures out if it's a companies house company or not, I suspect isCompaniesHouse may not be the best way to do this.

teneightfive:
I'm not convinced by this...it feels like going against the whole point of data hub vs CDMS. Feels like this just goes back to manually entering the data for everything which we know people don't do very well. Manual entry also leads to mistakes.

If they have a company number they will be present within companies house data, no? Is there not a better way to actually achieve whatever the purpose of this is? Look up company number from companies house, match up data from there if its not already present within data?

So make this a different journey which links the two rather than allows for duplicate entry of the same value.

cc/ @jim68000 this is probably more of a product question but surely we should be starting to think smarter longer term than just flooding these interfaces with manual data entry fields?

src/apps/companies/views/edit.njk
```diff
@@ -39,9 +39,19 @@
       id: company.id
     }
   }) %}
+
+    {% if isCompaniesHouse %}
```
teneightfive: As mentioned in the main set of comments I'm concerned by what this opens up.

A company can't easily update its company number (if at all?) so why do we allow it. If they're matched to a companies house record should we not be treating that company number as correct?

jim68000: So what is *supposed* to happen here is that when a user adds a CH number the DH API attempts to reconcile CH & DH data - eg, by updating registered name and address & adding SIC codes. Is this not happening? That may be another bug.

I see the issue around giving users permission to *change* CH numbers, but I'm not sure how we'd want to lock this down. If the user types the wrong number, what happens?

reupen: > So what is supposed to happen here is that when a user adds a CH number the DH API attempts to reconcile CH & DH data - eg, by updating registered name and address & adding SIC codes. Is this not happening? That may be another bug.

No, there has never proper logic around that.^ And if anything like that were to happen, there would be at least need to be a preview shown to the user before it happens (and probably an opportunity to make some choices about what happens).

^If a valid company number is added to a company, the API just starts returning the name and registered address from the CH record instead of what was there before (which is the logic we talked about getting rid of this morning, because it could lead to data loss).

reupen: > A company doesn't have the ability to change their company number

Although it's possible the wrong CH record was selected (where companies have hierarchies in there legal structure – like J Sainsbury Plc vs Sainsbury's Supermarkets Ltd).

ztolley: Left comments in Slack.. I think that may be the best place to discuss

PR #918, opened 2017-11-13
Update assignee time to only support hours

teneightfive:
Time previously support both hours and minutes. After agreeing with the steering group the OMIS service will now only support time to be recorded in hours. This simplifies some of the processing being done on the frontend. 

This change removes this and tidies up the language used to refer to entering estimated and actual time.

### Order overview
localhost_3000_omis_f804c606-fcf7-4eb7-bdd4-14fe7ecb7406_work-order 1

### Editing estimated time
localhost_3000_omis_f804c606-fcf7-4eb7-bdd4-14fe7ecb7406_edit_assignee-time 1

### Completing an order
localhost_3000_omis_6d3c9849-8dd4-4406-8336-2d8b8970e792_edit_complete-order 1

### Order overview
localhost_3000_omis_f804c606-fcf7-4eb7-bdd4-14fe7ecb7406_work-order

### Editing estimated time
localhost_3000_omis_f804c606-fcf7-4eb7-bdd4-14fe7ecb7406_edit_assignee-time

### Completing an order
localhost_3000_omis_6d3c9849-8dd4-4406-8336-2d8b8970e792_edit_complete-order

### DIFF
--- a/config/nunjucks/filters.js
+++ b/config/nunjucks/filters.js
@@ -162,10 +162,9 @@ const filters = {
 
   humanizeDuration: (value, measurement = 'minutes') => {
     const duration = moment.duration(value, measurement)
-    const hrsSuffix = pluralise('hour', duration.hours())
-    const minsSuffix = pluralise('minute', duration.minutes())
+    const hoursSuffix = pluralise('hour', duration.hours())
 
-    return duration.format(`h [${hrsSuffix}], m [${minsSuffix}]`)
+    return duration.format(`h [${hoursSuffix}]`)
   },
 
   formatDuration: (value, format = 'hh:mm', measurement = 'minutes') => {
--- a/src/apps/omis/apps/edit/fields.js
+++ b/src/apps/omis/apps/edit/fields.js
@@ -3,7 +3,7 @@ const { isArray, merge } = require('lodash')
 const globalFields = require('../../fields')
 
 function isDuration (value) {
-  const regex = /^\d+:\d{2}$/
+  const regex = /^\d*$/
 
   if (!isArray(value)) {
     if (!value) { return true }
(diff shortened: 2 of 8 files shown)


src/apps/omis/apps/view/views/work-order.njk
```diff
@@ -1,36 +1,37 @@
...
 
-{% set companyLink %}
-  <a href="/companies/{{ values.company.id }}">{{ values.company.name }}</a>
-{% endset %}
-
 {% set contactName %}
   <a href="/contacts/{{ values.contact.id }}">{{ values.contact.name }}</a>
 {% endset %}
 
 {% set contactEmail %}
```
ztolley: What do this have to do with hours?

teneightfive: Well spotted 😉 

I'll pull those amends out to a different PR...

teneightfive: Moved to #919

config/nunjucks/filters.js
```diff
@@ -163,9 +163,8 @@ const filters = {
   humanizeDuration: (value, measurement = 'minutes') => {
     const duration = moment.duration(value, measurement)
     const hrsSuffix = pluralise('hour', duration.hours())
-    const minsSuffix = pluralise('minute', duration.minutes())
 
-    return duration.format(`h [${hrsSuffix}], m [${minsSuffix}]`)
+    return duration.format(`h [${hrsSuffix}]`)
```
feedmypixel: not sure if `hrsSuffix` reads as well as `hoursSuffix` would?

src/apps/omis/apps/edit/fields.js
```diff
@@ -3,7 +3,7 @@ const { isArray, merge } = require('lodash')
 const globalFields = require('../../fields')
 
 function isDuration (value) {
-  const regex = /^\d+:\d{2}$/
+  const regex = /^\d*$/
```
feedmypixel: matching `0-9` zero or more times? is that correct? if you always want something then `\d+` is 1 or more times

teneightfive: As discussed on slack I think this is what we're after in this case so leaving as `*` for now.

PR #922, opened 2017-11-13
Add dormant investment status

ztolley:
Adds an option to the investment status form and the filter.

### DIFF
--- a/src/lib/metadata.js
+++ b/src/lib/metadata.js
@@ -160,11 +160,13 @@ module.exports.REASONS_FOR_ARCHIVE = [
   'Other',
 ]
 
+// Todo - Move to metadata repo in back end
 module.exports.investmentStatusOptions = [
   { label: 'Ongoing', value: 'ongoing' },
   { label: 'Delayed', value: 'delayed' },
   { label: 'Abandoned', value: 'abandoned' },
   { label: 'Lost', value: 'lost' },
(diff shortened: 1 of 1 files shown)


feedmypixel:
any tests around this work?

ztolley:
There are none around the existing status change form, I was waiting for the base set of collection tests before going and writing one. For this change we just add another option

feedmypixel:
> There are none around the existing status change form, I was waiting for the base set of collection tests before going and writing one. For this change we just add another option

sorry I meant unit tests

feedmypixel:
Ok we don't have any tests around metadata!

Also how come this is in metadata like this? feels like this should be in the backend?

ztolley:
The test would just say does the metadata have a list of options that include this, so not really a test of logic and there are no tests in metadata right now.

This will get moved to the backend soon but it seems that isn't going to happen before launch.

feedmypixel:
Can you add a `TODO` please that this will be moved to the backend? I don't want future devs to see this and just copy and paste the example. Really it shouldn't be in here in the first place.

ztolley:
https://uktrade.atlassian.net/browse/DH-1170

feedmypixel:
Grudgingly approved - we really shouldn't be adding things in to the frontend as a "quick" way of doing things. This should be in the backend like all the other things in this file...

It really is a false economy in the long run

PR #925, opened 2017-11-13
Show validation errors if role is missing from advisor

richtier:
# after this fix

image

# before this fix

image

## the cause

The form's state was being overridden with the request body, which was in a structure the template was not expecting.

## the fix

format the request body to be in a structure that the template expects.

### DIFF
--- a/src/apps/investment-projects/middleware/forms/team-members.js
+++ b/src/apps/investment-projects/middleware/forms/team-members.js
@@ -1,4 +1,4 @@
-const zipWith = require('lodash/zipWith')
+const { zipWith, get } = require('lodash')
 const { getAdvisers } = require('../../../adviser/repos')
 const { updateInvestmentTeamMembers } = require('../../repos')
 const { teamMembersLabels } = require('../../labels')
@@ -63,10 +63,21 @@ async function handleFormPost (req, res, next) {
     next()
   } catch (err) {
     if (err.statusCode === 400) {
+      const teamMembers = transformDataToTeamMemberArray(req.body)
+      const roleError = get(err, 'error.role')
+      let messages = {}
+      if (roleError) {
+        teamMembers.forEach((item, i) => {
+          if (item.role === '') {
+            messages['role-' + i] = roleError
+          }
+        })
+      }
       res.locals.form = Object.assign({}, res.locals.form, {
-        errors: err.error,
-        state: req.body,
+        errors: { messages },
+        state: { teamMembers },
       })
+
       next()
     } else {
       next(err)
--- a/src/apps/investment-projects/views/team/edit-team-members.njk
+++ b/src/apps/investment-projects/views/team/edit-team-members.njk
@@ -4,9 +4,6 @@
   <h2 class="heading-medium">Assign project specialist and team members</h2>
 
   {% call Form(form) %}
-    {% if form.errors.team_members %}
-      <div class="error">{{ form.errors.team_members }}</div>
-    {% endif %}
     <div class="js-AddItems"
       data-item-selector=".c-form-fieldset"
       data-item-name="team member"
@@ -28,7 +25,8 @@
               name: 'role',
               label: form.labels.role,
               value: teamMember.role,
-              idSuffix: loop.index
+              idSuffix: loop.index0,
+              error: form.errors.messages['role-' + loop.index0]
             }) }}
           {% endcall %}
         {% endfor %}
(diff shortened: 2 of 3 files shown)


src/apps/investment-projects/middleware/forms/team-members.js
```diff
@@ -63,10 +63,8 @@ async function handleFormPost (req, res, next) {
     next()
   } catch (err) {
     if (err.statusCode === 400) {
-      res.locals.form = Object.assign({}, res.locals.form, {
-        errors: err.error,
-        state: req.body,
-      })
+      res.locals.form.errors = err.error
```
teneightfive: Why has the assign been removed? Do we not need to keep any of the original state of this object?

src/apps/investment-projects/views/team/edit-team-members.njk
```diff
@@ -4,8 +4,20 @@
   <h2 class="heading-medium">Assign project specialist and team members</h2>
 
   {% call Form(form) %}
-    {% if form.errors.team_members %}
-      <div class="error">{{ form.errors.team_members }}</div>
+    {% if form.errors.role %}
+      <div class="c-error-summary js-ErrorSummary" role="alert" tabindex="-1">
```
teneightfive: Maybe take a look at the [ErrorSummary](https://github.com/uktrade/data-hub-frontend/blob/develop/src/templates/_macros/form/error-summary.njk) macro we have so that we don't duplicate this markup.

Errors can also be [passed directly](https://github.com/uktrade/data-hub-frontend/blob/develop/src/apps/contacts/views/edit.njk#L22) to the `Form` macro to handle display of them too.

Maybe take a look at where the form object on line 6 is built and add an `errors` property to that in the correct format.

src/apps/investment-projects/middleware/forms/team-members.js
richtier: ### before the change
None of the original state was being kept - `res.locals.form.state` was, in fact, **replaced** with the contents of `req.body` because `assign` does not merge with existing keys, it simply replaces. The outcome was a state that it looked like this:
```
{
    teamMembers: [<uuid>, <uuid>],
    roles: [<role>, <role>]
}
```
That caused the problem because [the template](https://github.com/uktrade/data-hub-frontend/pull/925/files#diff-463ef6da111e294fb7eed1a6e4ea4e61R28) is expecting `teamMembers` that look like this:
```
[
  {
    adviser: <uuuid>,
    role: <role>
  },
  {
    adviser: <uuuid>,
    role: <role>
  }
]
```
### after the change
`res.locals.form.state.teamMembers` is replaced with correctly formatted array of objects - in the above expected format.

Note we only replace `teamMembers` and keep other state.

src/apps/investment-projects/middleware/forms/team-members.js
```diff
@@ -63,10 +63,17 @@ async function handleFormPost (req, res, next) {
...
   } catch (err) {
     if (err.statusCode === 400) {
-      res.locals.form = Object.assign({}, res.locals.form, {
-        errors: err.error,
-        state: req.body,
-      })
+      const teamMembers = transformDataToTeamMemberArray(req.body)
+      if (err.error.role) {
+        let errorMessages = {}
+        for (var i = 0; i < teamMembers.length; i++) {
```
teneightfive: Just looking at this and think we can avoid having to use a `for` loop if we use `forEach` instead.

Also think we can keep a similar patter to before using `Object.assign` as its also a pattern we're using elsewhere in the codebase so helps us keep consistency. 

Thinking something like this might work:

```js
const teamMembers = transformDataToTeamMemberArray(req.body)
const roleError = get(err, 'error.role')
let messages = {}

if (roleError) {
  teamMembers.forEach((item, i) => {
    if (item.role === '') {
      messages['role-' + i] = roleError
    }
  })
}

  errors: { messages },
  state: { teamMembers },
})
```

src/apps/investment-projects/views/team/edit-team-members.njk
```diff
@@ -2,11 +2,7 @@
 
 {% block main_grid_right_column %}
   <h2 class="heading-medium">Assign project specialist and team members</h2>
-
```
teneightfive: Don't think we need to drop this line. Just adding extra noise.

richtier: unintentional. this is the outcome of me adding html then removing it. I removed an extra line.

PR #926, opened 2017-11-13
Correctly validate a list as required

teneightfive:
The default required validation doesn't support validating an array
of values and failing if any of them don't contain a value.

This validation is required for use on the order complete page so
a new validator has been added.

### DIFF
--- a/src/apps/omis/apps/edit/fields.js
+++ b/src/apps/omis/apps/edit/fields.js
@@ -1,4 +1,4 @@
-const { isArray, merge } = require('lodash')
+const { isArray, flatten, merge } = require('lodash')
 
 const globalFields = require('../../fields')
 
@@ -21,6 +21,14 @@ function isDuration (value) {
   })
 
   return valid
+}
+
+function arrayRequired (value) {
+  const invalidItems = flatten([value]).filter(itemValue => {
+    return itemValue === undefined || itemValue === ''
+  })
+
+  return invalidItems.length === 0
 }
 
 const editFields = merge({}, globalFields, {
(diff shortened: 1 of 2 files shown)


src/apps/omis/apps/edit/fields.js
```diff
@@ -23,6 +23,18 @@ function isDuration (value) {
   return valid
 }
 
+function arrayRequired (value) {
+  let valid = true
```
ztolley: Alternatively:

```
  const result = flatten([array]).filter(itemValue => (itemValue !== undefined && itemValue === ''))
  return result.length > 0
```

teneightfive: Yeah wasn't overly happy with it. Will look at that but might break the filter so its not all one hard to read line.

teneightfive: Also not sure if `arrayRequired` is the best name...what you think?

src/apps/omis/apps/edit/fields.js
```diff
@@ -23,6 +23,14 @@ function isDuration (value) {
   return valid
 }
 
+function arrayRequired (value) {
+  const invalidItems = flatten([value]).filter(itemValue => {
```
ztolley: 👍

src/apps/omis/apps/edit/fields.js
```diff
@@ -23,6 +23,14 @@ function isDuration (value) {
...
+    return itemValue === undefined || itemValue === ''
```
feedmypixel: do you need to also be checking for `null` here?

teneightfive: Not sure, what do you think? Will a field come back as `null`?

teneightfive: It's more likely to either not exist or come back as an empty string if it does exist in the form post.

PR #928, opened 2017-11-13
Add sector during order creation

teneightfive:
This adds the sector field to the creation journey rather than as part of the edit journey.

It can still be changed during the creation but is now required as part of the minimal initial creation journey.

### Step in creation journey
localhost_3000_omis_create_dcdabbc9-1781-e411-8955-e4115bead28a_sector_edit

### Creation summary step
localhost_3000_omis_create_dcdabbc9-1781-e411-8955-e4115bead28a_confirm 1

#### Todo

- [ ] update/add unit tests

### DIFF
--- a/src/apps/omis/apps/create/controllers/sector.js
+++ b/src/apps/omis/apps/create/controllers/sector.js
@@ -1,0 +1,21 @@
+const { get } = require('lodash')
+
+const metadataRepo = require('../../../../../lib/metadata')
+const { FormController } = require('../../../controllers')
+const { transformObjectToOption } = require('../../../../transformers')
+
+class SectorController extends FormController {
+  configure (req, res, next) {
+    req.form.options.fields.sector.options = metadataRepo.sectorOptions.map(transformObjectToOption)
+    super.configure(req, res, next)
+  }
+
+  saveValues (req, res, next) {
+    if (get(req.form, 'values.use_sector_from_company') === 'true') {
+      req.form.values.sector = get(res.locals, 'company.sector.id')
+    }
+    super.saveValues(req, res, next)
+  }
+}
+
+module.exports = SectorController
--- a/src/apps/omis/locales/en/default.json
+++ b/src/apps/omis/locales/en/default.json
@@ -22,6 +22,9 @@
     "description": {
       "label": "Description of the work or activity"
     },
+    "use_sector_from_company": {
+      "label": "Do you want to use the company's primary sector (shown above) for this order?"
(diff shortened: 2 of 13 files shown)


src/apps/omis/apps/create/controllers/sector.js
```diff
@@ -0,0 +1,21 @@
+const { get } = require('lodash')
+
+const metadataRepo = require('../../../../../lib/metadata')
+const { FormController } = require('../../../controllers')
+const { transformObjectToOption } = require('../../../../transformers')
+
+class SectorController extends FormController {
+  configure (req, res, next) {
+    req.form.options.fields.sector.options = metadataRepo.sectorOptions.map(transformObjectToOption)
```
feedmypixel: worth making this a `set`? is the nested `options` always going to be there?

src/apps/omis/apps/create/controllers/sector.js
```diff
@@ -0,0 +1,21 @@
...
+    super.configure(req, res, next)
+  }
+
+  saveValues (req, res, next) {
+    if (req.form.values.use_sector_from_company === 'true') {
```
feedmypixel: worth making this a `get` with a default to `false`?

src/apps/omis/locales/en/default.json
```diff
@@ -22,6 +22,9 @@
     "description": {
       "label": "Description of the work or activity"
     },
+    "use_sector_from_company": {
+      "label": "Do you want to use the company's primary sector (shown above) for this order?"
```
feedmypixel: should it be `companies` not `company's`?

src/apps/omis/apps/create/controllers/sector.js
teneightfive: In theory this is being set within the home office forms controller so we can rely on it being there. We have elsewhere so I'm reluctant to change it just for this controller.

Maybe if we want to be more robust in the future we can do it across all of them.

src/apps/omis/locales/en/default.json
teneightfive: No, it's the sector that belongs to the company, not a plural of company.

src/apps/omis/apps/create/controllers/sector.js
teneightfive: Same as above, we haven't actually done that elsewhere and can be confident what this method returns. But maybe with get there's no harm in doing a little overprotection.

PR #929, opened 2017-11-14
Handle unknown business type on company edit page

richtier:
# after the change
image

# before the change
image

### DIFF
--- a/test/unit/apps/companies/controllers/edit.test.js
+++ b/test/unit/apps/companies/controllers/edit.test.js
@@ -159,6 +159,12 @@ describe('Company export controller', () => {
       const label = this.controller.getBusinessTypeLabel(false, true, uuid)
       expect(label).to.equal('Foreign Company')
     })
+
+    it('should handle unrecognised uuids', () => {
+      const uuid = '98d14e94-milk-eggs-beef-e4115bead28a'
+      const label = this.controller.getBusinessTypeLabel(false, true, uuid)
+      expect(label).to.equal(undefined)
+    })
   })
 
   describe('renderForm', () => {
--- a/src/apps/companies/controllers/edit.js
+++ b/src/apps/companies/controllers/edit.js
@@ -18,7 +18,9 @@ function getBusinessTypeLabel (isLimitedCompany, isForeign, businessTypeUUID) {
     return 'UK limited company'
   }
   const businessTypeOption = getBusinessTypeOption(businessTypeUUID)
-  return (isForeign ? 'Foreign ' : 'UK ') + businessTypeOption.label
+  if (businessTypeOption) {
+    return (isForeign ? 'Foreign ' : 'UK ') + businessTypeOption.label
+  }
 }
 
 function renderForm (req, res) {


ztolley:
Should only show the business type and change option for adding a company, never when editing an existing one.

test/unit/apps/companies/controllers/edit.test.js
```diff
@@ -204,5 +211,29 @@ describe('Company export controller', () => {
...
+            const element = document.querySelector('.c-meta-list__item-label ')
+            expect(element.textContent).to.contain('UK Partnership')
+          })
+      })
+
+      it('should handle unknown business type label', () => {
+        const context = {
+          businessTypeLabel: undefined,
+        }
+        return render('../../src/apps/companies/views/edit.njk', context)
```
ztolley: Can we test what is passed to the template instead of rendering?

Also use '~/src' instead of '../../src'

test/unit/apps/companies/controllers/edit.test.js
```diff
@@ -204,5 +211,29 @@ describe('Company export controller', () => {
       this.controller.renderForm(reqMock, resMock, this.nextSpy)
       expect(this.renderSpy.args[0][1].businessTypeLabel).to.equal('UK limited company')
     })
+
+    describe('render', function () {
```
ztolley: use () => instead of function to be consistent

test/unit/apps/companies/controllers/edit.test.js
```diff
@@ -204,5 +211,29 @@ describe('Company export controller', () => {
...
+      it('should render the business type label', () => {
+        const context = {
+          businessTypeLabel: 'UK Partnership',
+        }
+        return render('../../src/apps/companies/views/edit.njk', context)
+          .then((document) => {
+            const element = document.querySelector('.c-meta-list__item-label ')
```
ztolley: extra space...

richtier: where? it's indented by 2 spaces, and there is only one space around the equals

test/unit/apps/companies/controllers/edit.test.js
richtier: we already [do that](https://github.com/uktrade/data-hub-frontend/blob/develop/test/unit/apps/companies/controllers/edit.test.js#L205)

but that kind of test would test this logic from the template:

```
businessTypeLabel | default('(type of organisation not known)')
```

so we need to render the template here to test it.

test/unit/apps/companies/controllers/edit.test.js
feedmypixel: I'm not convinced we should be testing this functionality in unit tests by rendering the nunjucks file? This feels like a UAT test if anything. Is this functionality not covered in a `unit` by testing the `getBusinessTypeLabel` method?

richtier: the template contains this logic:

```
businessTypeLabel | default('(type of organisation not known)')
```

so it should be tested somewhere.

I'll remove it to be added to tested by UAT sure.

feedmypixel: > so it should be tested somewhere.

I'm not convinced it should be. Any changes to the view layer are usually known about. Tests are to catch accidental changes, knock on effects, non pure code that cause another part of the application to blow up. So with the unit test on `getBusinessTypeLabel` output the view layer then does what it needs to do with that output.

Testing all views to make sure they are doing what was intended feels a bit overkill and I'm not convinced of the value. Yes we should test functionality via UAT, but testing visual output like this I'm not sold on.

Testing code that passes data to the view in a unit sense is much more valuable IMO

Always open for different appraoches

PR #930, opened 2017-11-14
Companies acceptance tests clean up

yeahfro:
Before starting the CDMS reference display on company details page I spotted some opportunities to clean up.

Included:
- removing redundant code
- consistency with other features
- reinstate companies search test


test/acceptance/features/companies/page-objects/Company.js
```diff
@@ -106,7 +108,8 @@ module.exports = {
               .waitForElementPresent('@continueButton')
               .click('@continueButton')
               .waitForElementPresent('@pageHeading')
-              .api.perform((done) => {
+              .api
+              .perform((done) => {
```
teneightfive: Does this need a new line? Seemed to make more sense before as each new line is a method call. `.api` looks a little lost on its own...

test/acceptance/features/companies/page-objects/Company.js
```diff
@@ -180,29 +214,61 @@ module.exports = {
                 })
               })
               .perform((done) => {
+                this.getRadioOption('headquarter_type', (result) => {
```
teneightfive: These changes feel like there's starting to get a bit repetitive. I don't really know nightwatch that well yet but wonder if there's something that can be done to improve this.

Probably more likely a follow-up change though so not blocking.

test/acceptance/features/companies/save.feature
```diff
@@ -1,28 +1,34 @@
-@companies-create
+@companies-save
```
teneightfive: What's the purpose of this rename? Feels like its still part of a create journey?

test/acceptance/features/companies/page-objects/Company.js
feedmypixel: @teneightfive FYI there are a number of state helpers coming that will improve all this repetition

test/acceptance/commands/getRadioOption.js
```diff
@@ -0,0 +1,17 @@
+/**
+ * Get a random option from a radio button list
+ * @param selector
+ * @param callback
+ */
+exports.command = function getRadioOption (name, callback) {
+  return this.elements('css selector', `input[name="${name}"] + label span`, (result) => {
+    const random = Math.floor(Math.random() * (result.value.length - 1))
```
feedmypixel: FYI I have wrapped a check around this for if (result.value.length) { and then log an error on the else for getListOption

PR #931, opened 2017-11-14
Do not render business type change link on edit

richtier:
image

---------------

image

### DIFF
--- a/src/apps/companies/views/edit.njk
+++ b/src/apps/companies/views/edit.njk
@@ -9,7 +9,11 @@
     <h2 class="heading-medium">Business type</h2>
     {{ MetaList({
         items: [
-          { label: businessTypeLabel | default('(type of organisation not known)'), value: '<a href="/companies/add-step-1">Change</a>', safe: true }
+          {
+            label: businessTypeLabel | default('(type of organisation not known)'),
+            value: '&nbsp;' if company.id else '<a href="/companies/add-step-1">Change</a>',
+            safe: true
+          }
         ]
       }) }}
   </div>


src/apps/companies/middleware/form.js
```diff
@@ -93,7 +93,13 @@ async function handleFormPost (req, res, next) {
   }
 }
 
+function setIsCompanyEdit (req, res, next) {
+  res.locals.isCompanyEdit = true
```
richtier: This helps answer the question of "is the user is edit mode or add mode".

I considered having similar logic to this in the controller (check is `res.locals.company` is set - if it is then the user is in edit mode). I avoided that when I realised that that this would make the controller aware of what another middleware was doing so leaking knowledge between middleware and controller layer.

This solves that problem.

src/apps/companies/controllers/edit.js
```diff
@@ -52,6 +52,7 @@ function renderForm (req, res) {
     .render('companies/views/edit', {
       isForeign,
       businessTypeLabel,
+      isCompanyEdit: res.locals.isCompanyEdit,
```
teneightfive: You shouldn't need to add this line as anything defined on `res.locals` in earlier middleware will be available in the view layer.

richtier: oh smashing :)

src/apps/companies/middleware/form.js
ztolley: What about looking for company id?

richtier: I don't follow your line of thought. care you explain more :)

ztolley: If company has an id you are editing, if it doesn't it's an add

PR #933, opened 2017-11-14
Fix website link in investment project management page

ztolley:
Previously the project management page incorrectly rendered links to an investor company website. 

Added some tests and more robust date handling while fixing the bug.

dhcrm-investment-project-assign-pm-website

link

### DIFF
--- a/src/apps/investment-projects/services/formatting.js
+++ b/src/apps/investment-projects/services/formatting.js
@@ -308,6 +308,7 @@ function transformBriefInvestmentSummary (data) {
   const investorCompany = data.investor_company
   const competitorCountries = data.competitor_countries || []
   const regionLocations = data.uk_region_locations || []
+  const date = moment(data.estimated_land_date)
 
   return {
     sector: get(data, 'sector.name', null),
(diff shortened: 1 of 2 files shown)


src/apps/investment-projects/services/formatting.js
```diff
@@ -308,18 +308,22 @@ function transformBriefInvestmentSummary (data) {
...
+  const date = moment(data.estimated_land_date)
 
   return {
     sector: get(data, 'sector.name', null),
     investor_company: {
       name: investorCompany.name,
       url: `/companies/${investorCompany.id}`,
     },
-    website: (investorCompany.website) ? `<a href="${investorCompany.website}">${investorCompany.website}</a>` : '',
+    website: investorCompany.website ? {
```
jim68000: We need to fix the initial capital thing too...

ztolley: That's in css , not in the code. It's applied to all tables. But yes that bugs me.

jim68000: So... we fix the CSS with a '.url' class?

ztolley: I think we remove the forced capitalisation for all values, but I'll do that in a new PR as it may invoke a lot of discussion.

PR #939, opened 2017-11-15
Add UK region to display of advisers in the UK

teneightfive:
This changes the format of the adviser display to include the UK
region to the list of advisers on the work order.

It also adds functionality to display if _you_ are one of the
advisers in the list that is currently present in the advisers in
the market.

## Before

image

## After

image

### DIFF
--- a/src/apps/omis/apps/view/controllers.js
+++ b/src/apps/omis/apps/view/controllers.js
@@ -6,10 +6,10 @@ const logger = require('../../../../../config/logger')
 async function renderWorkOrder (req, res) {
   const order = res.locals.order
   let values
+  const assignees = res.locals.assignees
 
   try {
     const subscribers = await Order.getSubscribers(req.session.token, order.id)
-    const assignees = await Order.getAssignees(req.session.token, order.id)
 
     values = merge({}, order, {
       subscribers,
--- a/src/apps/omis/apps/view/controllers.js
+++ b/src/apps/omis/apps/view/controllers.js
@@ -1,25 +1,16 @@
 const { get, merge, sumBy } = require('lodash')
 
-const { Order } = require('../../models')
-const logger = require('../../../../../config/logger')
-
-async function renderWorkOrder (req, res) {
+function renderWorkOrder (req, res) {
   const order = res.locals.order
-  let values
   const assignees = res.locals.assignees
+  const subscribers = res.locals.subscribers
 
-  try {
-    const subscribers = await Order.getSubscribers(req.session.token, order.id)
-
-    values = merge({}, order, {
-      subscribers,
-      assignees,
-      contact: res.locals.contact,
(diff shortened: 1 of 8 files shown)


src/apps/omis/apps/view/controllers.js
```diff
@@ -9,10 +9,22 @@ async function renderWorkOrder (req, res) {
   let values
 
   try {
-    const subscribers = await Order.getSubscribers(req.session.token, order.id)
+    const subscribersResponse = await Order.getSubscribers(req.session.token, order.id)
     const assignees = await Order.getAssignees(req.session.token, order.id)
     const contact = await getContact(req.session.token, order.contact.id)
 
+    const subscribers = map(subscribersResponse, (subscriber) => {
```
ztolley: Whats wrong with native map?

src/apps/omis/apps/view/views/work-order.njk
```diff
@@ -130,7 +130,7 @@
       url: 'edit/subscribers' if order.isEditable,
       label: 'Add or remove'
     }],
-    items: values.subscribers | map('name') if values.subscribers.length,
+    items: values.subscribers if values.subscribers.length,
```
ztolley: What happens if you just pass values.subscribers?

teneightfive: Yeah meant to actually just try that. Probably works just as well. Think it was there before because it was also doing `| map(values)`.

ztolley: Hence my question, I think you can just pass `value.subscribers`

src/apps/omis/apps/view/controllers.js
ztolley: Does it need a flatten added there now for when it's passed a null? Unit tests think so :)

src/apps/omis/apps/view/views/work-order.njk
teneightfive: Should be that now.

src/apps/omis/apps/view/controllers.js
teneightfive: This has been rewritten to use transformers :)

test/unit/apps/omis/apps/view/controllers.test.js
```diff
@@ -60,95 +51,41 @@ describe('OMIS View controllers', () => {
       }
     })
 
-    it('should not set a breadcrumb option', async () => {
-      await this.controllers.renderWorkOrder(this.reqMock, this.resMock)
+    it('should not set a breadcrumb option', () => {
+      this.controllers.renderWorkOrder(this.reqMock, this.resMock)
```
yeahfro: All of these `this.controllers.renderWorkOrder(this.reqMock, this.resMock)` could go into the `beforeEach`

PR #940, opened 2017-11-15
Fix events filtering on firefox

richtier:
Filtering of events fails on firefox, but not on chrome

## The cause

`history.pushState` has a lower size limited for state on firefox

browser| limit
------------ | -------------
Firefox | [640 kb](https://developer.mozilla.org/en-US/docs/Web/API/History_API#The_pushState()_method)
Chrome | Cannot find in docs, but it appears to be *much* higher than firefox's - I've heard as much as 10MB.

This affect's how the [`updateOutlet`](https://github.com/uktrade/data-hub-frontend/blob/develop/assets/javascripts/lib/xhr.js#L33) behaves when it encounters content with more than 320,000 characters (events filtering returns 340,000):

browser| behaviour
------------ | -------------
Firefox | Throws an exception when the response is too damn high
Chrome | Does not throw an exception because it can handle large state.

## The fix

- If the browser is unable to perform the `pushState` operation then perform a traditional page load.
- Reduce the size of the response by to less than 270,000 characters by removing whitespace. Note 230,000 of these characters are due to a list of event organisers, needed to populate this field:

image
 
A follow-up change could be to populate that field only once and not every time the filter changes.
 
### demo

firefox-fix-filter

# todo
 - [x] reduce response sizes
 - [x] handle responses that are too large
 - [x] add tests


assets/javascripts/lib/xhr.js
```diff
@@ -30,7 +30,8 @@ const XHR = {
     this.injectResponseInHtml(res.data)
 
     if (params) {
-      history.push(`?${queryString.stringify(params)}`, { data: res.data })
+      const data = res.data.replace(/(\r\n|\n|\r| {2}|)/gm, '')
```
richtier: oops this was a potential fix. will remove.

src/templates/_layouts/xhr.njk
```diff
@@ -1,3 +1,5 @@
+{% safespaceless %}
 <div class="main-content__inner l-container" id="xhr-outlet">
```
teneightfive: Want to add an indent to these lines to be consistent?

config/nunjucks/index.js
```diff
@@ -45,6 +48,27 @@ function ComponentExtension (env) {
   }
 }
 
+function SafeSpacelessExtension () {
+  this.tags = ['safespaceless']
+
+  this.parse = function (parser, nodes) {
+    var token = parser.nextToken()
```
teneightfive: Should we use `const` to be consistent?

PR #941, opened 2017-11-15
Add VAT Number field to company edit

ztolley:
Allow the user to specify a VAT Number associated with a UK company.

Also has minor changes to behaviour due to API requiring new text fields to specify no value as '' instead of null.

vat

### DIFF
--- a/src/apps/companies/services/formatting.js
+++ b/src/apps/companies/services/formatting.js
@@ -1,4 +1,5 @@
 const { title } = require('case')
+const { omitBy } = require('lodash')
 
 const { companyDetailsLabels, chDetailsLabels, hqLabels } = require('../labels')
 const { getFormattedAddress } = require('../../../lib/address')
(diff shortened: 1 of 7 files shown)


src/apps/companies/services/formatting.js
```diff
@@ -61,7 +61,8 @@ function getDisplayCompany (company) {
 
   if (company.uk_region && company.uk_region.name && company.uk_region.name !== 'Undefined') displayCompany.uk_region = company.uk_region.name
 
-  return displayCompany
+  // Strip out empty or null fields
+  return omitBy(displayCompany, val => (!val || val === ''))
```
yeahfro: `!val` will be fine here

src/apps/companies/services/formatting.js
```diff
@@ -62,7 +62,7 @@ function getDisplayCompany (company) {
   if (company.uk_region && company.uk_region.name && company.uk_region.name !== 'Undefined') displayCompany.uk_region = company.uk_region.name
 
   // Strip out empty or null fields
-  return omitBy(displayCompany, val => (!val || val === ''))
+  return omitBy(displayCompany, val => !val)
```
yeahfro: does `return omitBy(displayCompany, isNil)` work here?

ztolley: no

PR #944, opened 2017-11-15
Handle absent user when rendering investment project audit log

richtier:
## user known
image

## user unknown
image

### DIFF
--- a/src/apps/investment-projects/views/audit.njk
+++ b/src/apps/investment-projects/views/audit.njk
@@ -6,7 +6,7 @@
   <ul class="list-lined">
     {% for logEntry in auditLog %}
       <li>
-        User {{ logEntry.name }} changed {{ logEntry.changes }} {{ 'item' | pluralise(logEntry.changes) }} on {{ logEntry.timestamp }}
+        User {{ logEntry.name | default('Unknown adviser') }} changed {{ logEntry.changes }} {{ 'item' | pluralise(logEntry.changes) }} on {{ logEntry.timestamp }}
       </li>
     {% endfor %}
   </ul>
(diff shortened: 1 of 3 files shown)


src/apps/investment-projects/views/audit.njk
```diff
@@ -6,7 +6,7 @@
   <ul class="list-lined">
     {% for logEntry in auditLog %}
       <li>
-        User {{ logEntry.name }} changed {{ logEntry.changes }} {{ 'item' | pluralise(logEntry.changes) }} on {{ logEntry.timestamp }}
+        User {{ logEntry.name | default('Unknown Adviser') }} changed {{ logEntry.changes }} {{ 'item' | pluralise(logEntry.changes) }} on {{ logEntry.timestamp }}
```
richtier: I put the default in the template instead of the controller because "what to display when there is no data" seems to me to be a display logic thing.

yeahfro: Are we really really going for this rather than `An unknown adviser`?

test/unit/apps/investment-projects/controllers/audit.test.js
```diff
@@ -215,4 +215,51 @@ describe('Investment audit controller', () => {
       },
     }, this.next)
   })
+
+  describe('formatAuditLog', () => {
+    it('should return the user name', () => {
+      const actual = this.controller.formatAuditLog({
+        user: { name: 'James Example' },
+        timestamp: '2017-08-09T13:25:29.568665Z',
```
yeahfro: could we move some of this repetition to a variable or `beforeEach`?

src/apps/investment-projects/views/audit.njk
teneightfive: Adviser doesn't need a capital (channelling my inner Mills 😄 )

PR #945, opened 2017-11-16
Fix company edit not showing business type

richtier:
# cause of the bug

Misuse of the variable `isCompaniesHouse`. I thought it means "is the company in companies house". It actually means "have we retrieved the `companiesHouseRecord` from companies house? (which is only retrieved when adding a company, not editing a company)"

# the fix

- set `locals.companiesHouseCategory` in `getCompany` and `getCompaniesHouseRecord` middleware which are called on company edit and company add
- in the controller, first, try to use `companiesHouseCategory` and fall back to getting the business type label using the business type uuid.

# demo

## adding non-limited company

image

## adding companies from companies house (note not all companies in companies house are limited)

image

image

## editing a company from companies house

image

### DIFF
--- a/src/apps/companies/controllers/edit.js
+++ b/src/apps/companies/controllers/edit.js
@@ -13,13 +13,14 @@ function getBusinessTypeOption (businessTypeUUID) {
   )
 }
 
-function getBusinessTypeLabel (isLimitedCompany, isForeign, businessTypeUUID) {
-  if (isLimitedCompany) {
-    return 'UK limited company'
+function getBusinessTypeLabel (companiesHouseCategory, isForeign, businessTypeUUID) {
+  let prefix = isForeign ? 'Foreign' : 'UK'
+  if (companiesHouseCategory) {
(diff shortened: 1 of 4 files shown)


src/apps/companies/controllers/edit.js
```diff
@@ -13,13 +13,16 @@ function getBusinessTypeOption (businessTypeUUID) {
   )
 }
 
-function getBusinessTypeLabel (isLimitedCompany, isForeign, businessTypeUUID) {
-  if (isLimitedCompany) {
-    return 'UK limited company'
+function getBusinessTypeLabel (companiesHouseCategory, isForeign, businessTypeUUID) {
+  let label
+  if (companiesHouseCategory) {
```
teneightfive: Wonder if this can be restructured slightly with an idea of returning early rather than the if/else blocks.

Maybe set the foreign/uk as a variable at the top and then if companies house category exists, return the label, otherwise return what's in the else block by default.

PR #946, opened 2017-11-16
Company address transformation

ztolley:
In preparation for moving company to use the transformer pattern used everywhere else, to move off the key value table macro, this PR tidies up the formatting in the current transformer.

Changes all address generation to use the getFormattedAddress method (each transformer was doing it differently). Also extracts the code to generate the sic code list as that will be reused in future.

### DIFF
--- a/src/apps/companies/transformers.js
+++ b/src/apps/companies/transformers.js
@@ -9,6 +9,7 @@ const {
 } = require('lodash')
 
 const metadataRepository = require('../../lib/metadata')
+const { getFormattedAddress } = require('../../lib/address')
 
 function transformCompanyToListItem ({
   id,
@@ -92,6 +93,7 @@ function transformCompaniesHouseCompanyToListItem ({
   registered_address_town,
   registered_address_county,
   registered_address_postcode,
+  registered_address_country,
   company_number,
   company_category,
   company_status,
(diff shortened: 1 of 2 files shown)


src/apps/companies/transformers.js
```diff
@@ -101,24 +117,7 @@ function transformCompaniesHouseCompanyToListItem ({
...
-
-  const sicCodes = [
-    sic_code_1,
-    sic_code_2,
-    sic_code_3,
-    sic_code_4,
-  ]
-    .filter(item => item.length)
-    .join(', ')
+  const sicCodes = transformSicCodes({ sic_code_1, sic_code_2, sic_code_3, sic_code_4 })
```
teneightfive: Not sure we need to split this into a separate function?

test/unit/apps/companies/transformers.test.js
```diff
@@ -135,7 +135,7 @@ describe('Company transformers', function () {
     it('should format the address and return as meta data', () => {
       expect(this.transformed.meta[5]).to.deep.equal({
         label: 'Address',
-        value: 'Fred cottage smith lane, Winkfield, Windsor, Berkshire, SL9 9RE',
+        value: 'Fred Cottage Smith Lane, Winkfield, Windsor, Berkshire, SL9 9RE, United Kingdom',
```
teneightfive: Not sure about capitalising every word...maybe we just leave the address in the format it was entered. This seems to be what companies house do?

src/apps/companies/transformers.js
ztolley: we do as a follow up pr will use it

test/unit/apps/companies/transformers.test.js
ztolley: companies house make it all caps.  The aim here isn't to fix the formatter, but to consistently use it, then we can fix it later in one place.

ztolley: The address formatter itself could and *should* be improved. But this ticket just switches the code to use it to provide consistent addresses. Currently, addresses are generated in 3 different ways in the same controller and are not consistent between themselves and addresses elsewhere in the stem.

PR #947, opened 2017-11-16
Lead adviser (assignee) validation

teneightfive:
As setting the lead assignee isn't part of the standard edit steps
it needs to be handled slightly differently. This adds support for a
new custom variable for the view which determines whether the
error message is displayed.

## What it looks like 

localhost_3001_omis_b85a4802-17f0-4fec-82ee-5f49225dfe83_quote

### DIFF
--- a/src/apps/omis/apps/view/middleware.js
+++ b/src/apps/omis/apps/view/middleware.js
@@ -118,6 +118,7 @@ async function setQuotePreview (req, res, next) {
       }
     })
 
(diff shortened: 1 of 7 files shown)


teneightfive:
>Would this mean that potentially you'd go back to the page because there is an error but not see an error?

Not sure what you mean? You'd only get "error" messages when there are things you still need to complete. The other errors are the same and each has a step to link to so are handled differently. This one needs to be correct by going to the overview and selecting a lead adviser which is why it's coded slightly differently.

PR #949, opened 2017-11-16
Use the correct method to reset the value of conditional subfields

teneightfive:
When resetting the value for subfields this method was setting
the value to an empty string and setting checked to false.

This caused an issue for radio or checkbox elements as they were having
the value they needed removed. So when the form was submitted with
their value it was empty.

### DIFF
--- a/assets/javascripts/modules/conditional-subfields.js
+++ b/assets/javascripts/modules/conditional-subfields.js
@@ -139,8 +139,11 @@ const ConditionalSubfields = {
 
       Array.from(children).forEach((field) => {
         if (!field.getAttribute('data-persist-values')) {
-          field.value = ''
-          field.checked = false
+          if (['select', 'radio', 'checkbox'].indexOf(field.type) > -1) {
+            field.checked = false
+          } else {
+            field.value = ''
+          }
         }
 
         const event = this.wrapper.createEvent('HTMLEvents')


assets/javascripts/modules/conditional-subfields.js
```diff
@@ -139,8 +139,11 @@ const ConditionalSubfields = {
 
       Array.from(children).forEach((field) => {
         if (!field.getAttribute('data-persist-values')) {
-          field.value = ''
-          field.checked = false
+          if (['text', 'textarea'].indexOf(field.type) > -1) {
```
teneightfive: Not sure whether its better to check for text based fields and have fallback as `checked = false` or to check against `['select', 'radio', 'checkbox']` as have fallback as `value = ''`.

What are people's thoughts?

teneightfive:
>I still think it may be a better idea to set disabled=true on the field and the reverse to enable it again. Gets around the value issue and also works the same for all form controls

assets/javascripts/modules/conditional-subfields.js
```diff
@@ -139,8 +139,11 @@ const ConditionalSubfields = {
 
       Array.from(children).forEach((field) => {
```
feedmypixel: could use lodash `forEach` that iterates over all collection types

assets/javascripts/modules/conditional-subfields.js
feedmypixel: do we need to consider `tel`, `time`, `email`?

feedmypixel: > Not sure whether its better to check for text based fields and have fallback as checked = false or to check against ['select', 'radio', 'checkbox'] and have the fallback as value = ''.
I would whitelist elements that have the attribute `checked`

so I would be inclined to go with `if (includes(['select', 'radio', 'checkbox'], field.type)){`

assets/javascripts/modules/conditional-subfields.js
teneightfive: This is for the client so didn't want to bulk the dependencies more than we had to when there is native support.

assets/javascripts/modules/conditional-subfields.js
teneightfive: Yeh, think your comment about switching to the other way around is better.

assets/javascripts/modules/conditional-subfields.js
feedmypixel: ah ok - sorry in node mode :)

PR #956, opened 2017-11-16
Migrate company to table component

ztolley:
The company details and edit pages use the key value table macro to render their content, this macro is old and potentially unsafe. This change migrates the company details and edit pages to the newer key value table component.

As part of the change, the old company formatting code has been replaced with newer transformers, following patterns from the other sub apps. The unit tests have also been expanded to provide more coverage and updated to follow newer, better, patterns.

company details before
*Company details before*

company details after
Company details after

company edit before
Company edit before

company edit after
Company edit after

The styles and spacing for the tables have changed slightly with the migration, though this should make them more consistent with the rest of the tables on the site as all are being migrated off the macro version


ztolley:
Note the current failing test is due to bug in Leeloo being addressed

src/apps/companies/transformers.js
```diff
@@ -197,6 +204,72 @@ function transformCompanyResponseToForm (body) {
...
+  const viewRecord = {
+    trading_name,
+    vat_number,
+    reference_code,
+    description,
+    uk_region: get(uk_region, 'name'),
+    sector: getPrimarySectorName(get(sector, 'name')),
+    website: (website && website.length > 0) ? {
+      name: website,
+      url: `<a href="${website}">${website}</a>`,
```
teneightfive: Still don't like this....shouldn't be doing HTML in these places.

We should do something more like meta items where we send an object that expects `label` and `value` and an optional `type` which can handle how its displayed.

test/unit/data/companies/datahubOnlyCompany.json
```diff
@@ -0,0 +1,83 @@
+{
```
teneightfive: Can we use kebab case for file names please.

src/apps/companies/transformers.js
```diff
@@ -270,6 +273,43 @@ function transformCompanyResponseToViewRecord ({
...
+  sic_code_1,
+  sic_code_2,
+  sic_code_3,
+  sic_code_4,
+  incorporation_date,
+} = {}) {
+  const viewRecord = {
+    company_number,
+    company_status,
+    name: title(name),
```
teneightfive: Wonder whether we shouldn't destructure this? Just found it harder to understand what was going on as the method `title` was a bit confusing. Thinking `Case.title(name)` would be easier to read when looking at this transformer.

src/apps/companies/transformers.js
```diff
@@ -270,6 +273,43 @@ function transformCompanyResponseToViewRecord ({
...
+      address_town: registered_address_town,
+      address_county: registered_address_county,
+      address_postcode: registered_address_postcode,
+      address_country: registered_address_country,
+    }),
+    incorporation_date: formatLongDate(incorporation_date),
+    sic_code: transformSicCodes({ sic_code_1, sic_code_2, sic_code_3, sic_code_4 }),
+  }
+
+  return omitBy(getDataLabels(viewRecord, chDetailsLabels), val => !val)
```
teneightfive: Are you essentially filtering empty values here?

[OmitBy](https://lodash.com/docs/4.17.4#omitBy) uses [identity](https://lodash.com/docs/4.17.4#identity) as the default predicate which will already do this for you.

src/apps/companies/transformers.js
```diff
@@ -310,6 +311,15 @@ function transformCompaniesHouseResponseToViewRecord ({
   return omitBy(getDataLabels(viewRecord, chDetailsLabels), val => !val)
 }
 
+function transformCompanyResponseToOneListViewRecord ({ one_list_account_owner, classification }) {
+  const viewRecord = {
+    oneListAccountManager: get(one_list_account_owner, 'name', 'None'),
```
teneightfive: We should have a review of the content for fallbacks at some point. Always just saying `None` feels a bit lazy.

src/apps/companies/transformers.js
ztolley: Actually thats a bug, it shouldn't be putting html into the url AT ALL

src/apps/companies/controllers/details.js
```diff
@@ -1,78 +1,20 @@
...
-      chDetails: companyFormattingService.getDisplayCH(res.locals.company.companies_house_data),
-      chDetailsDisplayOrder: chDetailsDisplayOrder,
-      chDetailsLabels: chDetailsLabels,
-    })
-  }
-
   res
     .breadcrumb(company.name)
-    .render('companies/views/details')
+    .render('companies/views/details', {
```
teneightfive: This looks a lot neater 😃

src/apps/companies/views/details.njk
```diff
@@ -5,13 +5,13 @@
 
   {% if chDetails %}
     <div class="dark-panel">
-      {{ keyvaluetable(chDetails, labels=chDetailsLabels, keyorder=chDetailsDisplayOrder, readonly=true) }}
+      {% component 'key-value-table', variant='readonly', items=chDetails %}
```
teneightfive: This is obviously making use of the old style components which we also need to still deprecate.

I wonder if when we move its worth considering the use of the [AnswerSummary](https://github.com/uktrade/data-hub-frontend/blob/f9e4ef46b8864bacf9813a0c9ee6775162da78ab/src/templates/_macros/common/answers-summary.njk) macro that we've been using for OMIS. It's tested very well and also the way in which edit actions are associated with the tables. 

It also supports a unique way of doing fallbacks to give them a different style visually making it easier to scan not filled in data. And we can look to make better use of how we handle formatting using a `type` property.

src/apps/companies/services/formatting.js
```diff
@@ -1,75 +0,0 @@
...
-const { companyDetailsLabels, chDetailsLabels, hqLabels } = require('../labels')
-const { getFormattedAddress } = require('../../../lib/address')
-const { formatLongDate } = require('../../../../common/date')
-const { getPrimarySectorName } = require('../../../../common/transform-sectors')
-
-const companyDetailsDisplayOrder = Object.keys(companyDetailsLabels)
-const chDetailsDisplayOrder = Object.keys(chDetailsLabels)
-const companyTableKeys = ['name', 'address']
-
-function getDisplayCH (companyHouseData) {
```
teneightfive: 🙌

src/apps/companies/views/details.njk
ztolley: Well maybe we look at that and create a new macro after launch and bulk move off the key value table component. For now this gets us off the macro that has the xss issue and all tables use the same component.

test/unit/data/companies/datahubOnlyCompany.json
ztolley: I copied the case used for the other files already in the folder.

src/apps/companies/transformers.js
ztolley: removes null and empty and then joins them. How would you simplify it?

teneightfive: I just meant you don't need to pass the second argument to `omitBy` as the default predicate that lodash uses will do what you're doing (removing null and empty I believe).

src/apps/companies/views/details.njk
teneightfive: Yeh of course. Was just for a future suggestion. 

Although if we're thinking of making a change like this might be worth making soon so that people don't get too used to the current key/value table?

test/unit/data/companies/datahubOnlyCompany.json
teneightfive: Which is also probably wrong 🙄 

Let's try and always stick to kebab case for file names please. We can change existing ones in another commit if we want.

test/unit/apps/companies/transformers.test.js
```diff
@@ -200,7 +200,7 @@ describe('Company transformers', function () {
       it('should convert website to link', () => {
         expect(this.viewRecord['Website']).to.deep.equal({
           name: 'http://www.test.com',
-          url: '<a href="http://www.test.com">http://www.test.com</a>',
+          url: 'http://www.test.com">http://www.test.com',
```
teneightfive: This test can't be correct? Surely is should be looking for the value? Not the back half of an anchor tag too?

src/apps/companies/views/details.njk
ztolley: I want us to do the next version *right* and not rushed

src/apps/companies/transformers.js
ztolley: think I tried and it didn't but will try again

ztolley: Tried it, didn't work.

teneightfive: Ah, just looked myself and what you actually want is [pickBy](https://lodash.com/docs/4.17.4#pickBy) with no second argument. Not `omitBy` with manually checking for value.

src/apps/companies/controllers/details.js
```diff
@@ -0,0 +1,22 @@
...
+  transformCompanyResponseToOneListViewRecord,
+} = require('../transformers')
+
+function renderDetails (req, res) {
+  const company = res.locals.company
+
+  res
+    .breadcrumb(company.name)
+    .render('companies/views/details', {
+      tab: 'details',
```
teneightfive: Is this still needed? Feels like a legacy thing 🤔

PR #957, opened 2017-11-17
Fix postcode lookup error handling

richtier:
## the bug

unrecognised postcodes result in `undefined` being displayed on the address selector

image

this is caused by non-200 responses from the external postcode service being returned to the browser as a 200 response.

## the fix

Return the external postcode service's actual response, so the browser will get a 400, which will mean the "render the dropdown" js code is not called

### DIFF
--- a/src/apps/api/controllers.js
+++ b/src/apps/api/controllers.js
@@ -15,15 +15,16 @@ async function postcodeLookupHandler (req, res) {
   try {
     const postcode = req.params.postcode
     const addresses = await lookupAddress(postcode)
+    const unitedKingdomCountryId = metadata.countryOptions.find(
+      country => country.name.toLowerCase() === 'united kingdom'
+    ).id
     const augmentedAddresses = addresses.map(address => {
-      return Object.assign({}, address, {
-        country: metadata.countryOptions.find(country => country.name.toLowerCase() === 'united kingdom').id,
-      })
+      return Object.assign({}, address, { country: unitedKingdomCountryId })
     })
 
     res.json(augmentedAddresses)
   } catch (error) {
-    res.json({ error })
+    res.status(error.statusCode).json({ message: error.message })
   }
 }
 
--- a/test/unit/apps/api/controllers.test.js
+++ b/test/unit/apps/api/controllers.test.js
@@ -1,0 +1,55 @@
+describe('api controllers', () => {
+  beforeEach(() => {
+    this.req = { params: { postcode: 'dn21 6fg' } }
+    this.sandbox = sinon.sandbox.create()
+    this.resMock = {
+      send: this.sandbox.spy(),
+      json: this.sandbox.spy(),
+      status: this.sandbox.stub().returnsThis(),
+    }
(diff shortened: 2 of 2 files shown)


src/apps/api/controllers.js
```diff
@@ -11,19 +11,22 @@ async function adviserSearchHandler (req, res, next) {
...
   try {
     const postcode = req.params.postcode
     const addresses = await lookupAddress(postcode)
+    const unitedKingdomCountryId = metadata.countryOptions.find(
+      country => country.name.toLowerCase() === 'united kingdom'
+    ).id
     const augmentedAddresses = addresses.map(address => {
-      return Object.assign({}, address, {
-        country: metadata.countryOptions.find(country => country.name.toLowerCase() === 'united kingdom').id,
-      })
```
richtier: as far as I can tell there is no reason to find UK's id on ever address - so I moved it out of the loop.

src/apps/api/controllers.js
```diff
@@ -11,19 +11,22 @@ async function adviserSearchHandler (req, res, next) {
   }
 }
 
-async function postcodeLookupHandler (req, res) {
+async function postcodeLookupHandler (req, res, next) {
```
richtier: I added `next` to facilitate async testing.

teneightfive: Not sure this is necessary. You can test async tests using `async/await` too.

src/apps/api/controllers.js
```diff
@@ -11,19 +11,22 @@ async function adviserSearchHandler (req, res, next) {
...
-        country: metadata.countryOptions.find(country => country.name.toLowerCase() === 'united kingdom').id,
-      })
+      return Object.assign({}, address, { country: unitedKingdomCountryId })
     })
 
     res.json(augmentedAddresses)
+    next()
   } catch (error) {
-    res.json({ error })
+    res.status(error.statusCode).send(error.message)
```
teneightfive: Should it return the error as JSON as well?

src/apps/api/controllers.js
richtier: using `await` in the test would require the controller return the promise, no? that would be be a bit odd.

ztolley: Our old old tests did that but a cleaner pattern is to call the controller with await and then test the mocks. Works and easier to read the tests.

test/unit/apps/api/controllers.test.js
```diff
@@ -0,0 +1,44 @@
...
+  afterEach(() => {
+    this.sandbox.restore()
+  })
+
+  describe('postcodeLookupHandler', () => {
+    it('should return postcode response', async () => {
+      await this.controller.postcodeLookupHandler(this.req, this.resMock)
+      expect(this.resMock.json).to.be.calledWith([{ country: '1234' }])
+    })
+    it('should return correct status code and message on error', async () => {
```
ztolley: Could the test be split into 
context
  beforeEach
  it

I know it's repetitive but we've started adopting this pattern and it's a good pattern to follow, it also tends to make you test one thing at a time

PR #958, opened 2017-11-17
Move whitespace removal to source of inflated responses

richtier:
Removing whitepsace from entire XHR responses had unwanted side effects.

By narrowing down where we remove whitespace we solve for the largest offenders but reduce the scope for unwanted side-effects.

### DIFF
--- a/src/templates/_macros/form/select-box.njk
+++ b/src/templates/_macros/form/select-box.njk
@@ -29,9 +29,11 @@
     {% for option in options %}
       {% set label = option.name or option.label %}
       {% set value = option.id or option.value %}
-      <option value="{{ value }}" {% if value and fieldValue === value %}selected{% endif %}>
-        {{ label }}
-      </option>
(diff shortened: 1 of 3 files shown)


src/templates/_macros/form/select-box.njk
```diff
@@ -29,9 +29,11 @@
     {% for option in options %}
       {% set label = option.name or option.label %}
       {% set value = option.id or option.value %}
-      <option value="{{ value }}" {% if value and fieldValue === value %}selected{% endif %}>
-        {{ label }}
-      </option>
+      {% safespaceless %}
```
feedmypixel: maybe not as this PR but can we improve the name [here](https://github.com/uktrade/data-hub-frontend/blob/39906a555d8d03321212831917baafdabce7499e/config/nunjucks/index.js#L52) this is just the name of the extension, maybe`removeWhiteSpace` would mean more to a dev looking at the code

PR #959, opened 2017-11-17
Add acceptance tests around support form

yeahfro:
Changes cover:
- acceptance tests around support form
- making email address a mandatory field


src/apps/support/macros.js
```diff
@@ -37,6 +37,7 @@ const feedbackFormConfig = (browserInfo) => ({
     },
     {
       macroName: 'TextField',
+      optional: true,
```
feedmypixel: should this be `optional`? How else are we going to contact them?

test/acceptance/features/support/page-objects/Support.js
```diff
@@ -0,0 +1,66 @@
...
+          })
+          .perform(() => {
+            for (const key in supportRequest) {
+              if (supportRequest[key]) {
+                this.section.form
+                  .setValue(`@${key}`, supportRequest[key])
+              }
+            }
+            for (const key in supportRequestRadioOptions) {
+              this.api.useCss().click(supportRequestRadioOptions[key].labelSelector)
```
feedmypixel: id push the `click` onto another line

test/acceptance/features/support/step_definitions/support.js
```diff
@@ -0,0 +1,52 @@
...
+
+  Then(/^the support fields have error messages$/, async function () {
+    await Support.section.form
+      .assert.visible('@titleError')
+      .assert.visible('@chooseOneOfTheseError')
+  })
+
+  Then(/^a new support request is created$/, async function () {
+    await Support
+      .createSupportRequest({}, (supportRequest) => { set(this.state, 'supportRequest', supportRequest) })
```
feedmypixel: i'd drop this down to the next line, its a bit difficult to scan like this

test/acceptance/features/support/support.feature
```diff
@@ -0,0 +1,27 @@
+@support
+Feature: Support
+
+  @support--form
+  Scenario: Load support form
+
+    When I navigate to the dashboard
+    And I navigate to the support page
+    Then there are support fields
```
feedmypixel: future wise "form has the expected fields" and then a `dataTable`

test/acceptance/helpers/zendesk.js
```diff
@@ -0,0 +1,14 @@
+const http = require('http')
```
feedmypixel: Rather than spinning up a server can we not just use `nock` here?

```
nock(https://${process.env.ZEN_DOMAIN}.zendesk.com)
      .post(`/api/v2/tickets.json`)
      .reply(200, response)
```


As an aside it looks like we could do with splitting `config.zenUrl` into domain and path so its easy to use in tests.

And we can then add these in for CircleCi

feedmypixel: This does feel like the only way. I kind of want to suggest creating a docker for this to run on circleCi but for the moment this will do.

test/acceptance/helpers/zendesk.js
```diff
@@ -0,0 +1,15 @@
...
+
+module.exports = {
+  start (response) {
+    const port = process.env.ZEN_TEST_URL.split(':')[2]
+    this.server = http.createServer((req, res) => {
+      res.writeHead(200, { 'Content-Type': 'application/json' })
+      res.write(JSON.stringify(response))
+      res.end()
+    }).listen(port)
+  },
```
feedmypixel: can you add an auto kill timer on this so if `stop` is not called within a certain amount of time then the server is killed?

test/acceptance/helpers/zendesk.js
```diff
@@ -0,0 +1,15 @@
+const http = require('http')
+
+module.exports = {
+  start (response) {
+    const port = process.env.ZEN_TEST_URL.split(':')[2]
```
feedmypixel: [url.parse](https://nodejs.org/api/url.html#url_url_parse_urlstring_parsequerystring_slashesdenotehost) may be helpful here

config/index.js
```diff
@@ -35,13 +35,15 @@ const config = {
...
   logLevel: process.env.LOG_LEVEL || (isDev ? 'debug' : 'error'),
-  zenUrl: `https://${process.env.ZEN_DOMAIN}.zendesk.com/api/v2/tickets.json`,
-  zenToken: process.env.ZEN_TOKEN,
-  zenEmail: process.env.ZEN_EMAIL,
-  zenBrowser: process.env.ZEN_BROWSER,
-  zenImpact: process.env.ZEN_IMPACT,
-  zenService: process.env.ZEN_SERVICE,
-  zenServiceChannel: process.env.ZEN_SERVICE_CHANNEL || 'datahub',
+  zen: {
+    url: process.env.ZEN_TEST_URL || `https://${process.env.ZEN_DOMAIN}.zendesk.com/api/v2/tickets.json`,
```
feedmypixel: feels like the `ZEN_DOMAIN` is all wrong in the first place before you even got here 😞

src/apps/support/services.js
```diff
@@ -19,19 +19,19 @@ function createZenDeskMessage ({
       body: description,
     },
     custom_fields: [
-      { id: config.zenBrowser, value: browser },
-      { id: config.zenService, value: config.zenServiceChannel },
+      { id: config.zen.browser, value: browser },
```
feedmypixel: worth `get` ing these? quite likely `envs` are not there

test/acceptance/features/support/page-objects/Support.js
```diff
@@ -0,0 +1,68 @@
...
+      createSupportTicket (details = {}, callback) {
+        const supportRequest = assign({}, {
+          title: faker.lorem.words(),
+          description: faker.lorem.sentence(),
+          email: faker.internet.email(),
+          webBrowser: faker.random.words(),
+        }, details)
+        const supportRequestRadioOptions = {}
+
+        this.section.form
```
feedmypixel: you can return this here

test/acceptance/features/support/page-objects/Support.js
```diff
@@ -0,0 +1,68 @@
...
+            }
+          })
+          .perform(() => {
+            this.section.form
+              .click('@sendButton')
+
+            callback(assign({}, supportRequest, supportRequestRadioOptions))
+          })
+
+        return this
```
feedmypixel: you can return above

test/acceptance/features/support/support.feature
ztolley: Yeah that will make it super useful

test/acceptance/helpers/zendesk.js
ztolley: I like the idea of the nock way. Is there a reason why not?

feedmypixel:
I'll have a look at this Monday if @yeahfro isn't in

yeahfro:
I will deal with this as a priority

feedmypixel:
Tests wise I think drop the sending functionality for AT tests. Cover that with unit and nock. This spinning up a little server doesn’t feel quite right and I’m not sure if it sits in the Acceptance Tests domain/area? What do you think?

PR #960, opened 2017-11-17
Fix missing user on log line on company audit

richtier:
## missing user

image

## user present 

image


src/apps/audit/transformers.js
```diff
@@ -30,7 +31,7 @@ function transformAuditLogToListItem (labels = {}) {
       contentMetaModifier: 'stacked',
       meta: [{
         label: 'Adviser',
-        value: logEntry.user.name,
+        value: get(logEntry, 'user.name') || 'Unknown adviser',
```
richtier: I would prefer to put the default value in the template, but `MetaList` renders this and that does not support such behaviour.

test/unit/apps/audit/transformers.test.js
```diff
@@ -41,6 +41,26 @@ describe('Audit transformers', () => {
     })
   })
 
+  it('should handle absent user', () => {
```
teneightfive: Maybe look at using `context` to describe the situation and then use the `it` block for saying what the assertion should be. So something like `context('when the user is absent')` and `it('should use default text')`.

ztolley: I would echo Dom's words. Context has changed how I look at tests, try it

src/apps/audit/transformers.js
ztolley: `get(logEntry, 'user.name', 'Unknown adviser')`

richtier: that would not show "unknown adviser" for the following log line:

```
{ user: { name: '' } }
```

teneightfive: I've considered adapting `MetaList` to support a `fallback` or `default` property so that we don't need to do things like this. 

One for the future though. It's a pattern I added to `AnswersSummary` and found nicer.

PR #962, opened 2017-11-17
New company exports transformer

ztolley:
Moves the company export screen off the old key value table macro to the newer component version to fix XSS issue.

### DIFF
--- a/src/apps/companies/transformers/company-response-to-export-details-view-record.js
+++ b/src/apps/companies/transformers/company-response-to-export-details-view-record.js
@@ -1,0 +1,21 @@
+/* eslint-disable camelcase */
+const { flatten } = require('lodash')
+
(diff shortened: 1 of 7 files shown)


src/apps/companies/transformers/company-response-to-export-details-view-record.js
```diff
@@ -0,0 +1,21 @@
...
+
+const { getDataLabels } = require('../../../lib/controller-utils')
+const { exportDetailsLabels } = require('../labels')
+
+module.exports = function transformCompanyResponseToExportDetailsViewRecord ({
+  export_to_countries,
+  future_interest_countries,
+}) {
+  const viewRecord = {
+    exportToCountries: flatten([export_to_countries])
```
yeahfro: After discussion we agree that this is consistent with other implementations.

PR #963, opened 2017-11-17
Convert contact details to transformer/key value table component combination

ztolley:
Converts the contact details formatter to a transformer following the same pattern as other details screens.

Adopts the key value table component to replace the old macro version, to avoid attempted XSS attacks.

Some improvements over the previous unit tests.


ztolley:
Couple of failing acceptance tests and the old unit test code just isn't cutting it so fixing those before re-offering it

ztolley:
Giving @feedmypixel a turn at merging then I'll have to fix conflicts here

src/apps/contacts/transformers.js
```diff
@@ -73,6 +85,58 @@ function transformContactToListItem ({
...
 
+/**
+ * Translate a raw contact object into a formatted contact
+ * to display on the screen
+ * @param {object} contact
+ * @param {object} company
+ * @returns {object} displayContact A contact that can be put into a key value table
+ *
+ */
+function transformContactResponseToViewRecord ({
```
teneightfive: Just `transformContactResponseToView` we don't really say `ViewRecord` elsewhere.

src/apps/contacts/views/details.njk
```diff
@@ -3,14 +3,7 @@
...
   <h2 class="heading-medium">Contact details</h2>
-  {{ keyvaluetable(
-      contactDetails,
-      stripey=true,
-      labels=contactDetailsLabels,
-      keyorder=contactDetailsDisplayOrder,
-      id="contact-details",
-      shownull=true)
-  }}
+  {% component 'key-value-table', variant='striped', items=contactDetails, id="contact-details" %}
```
teneightfive: Mix of quote style.

src/apps/contacts/transformers.js
ztolley: actually, we do use viewRecord in Lots of places, though it's a valid point. Will change here and will update other recent changes later to be consistent

test/unit/data/contacts/contact.json
```diff
@@ -0,0 +1,45 @@
+{
```
teneightfive: Don't we already have a contact object at `/test/unit/data/simple-contact.js`?

ztolley: So we do, but that one is old. I'll switch to this one as it's more up to date and in the correct place.

teneightfive: Yeh maybe remove the out of date one.

It would actually be good to review the mock data we have and make them consistent. Also use the same format. We have some as JSON and some as JS objects.

ztolley: simple contact is a search result format, the new one is a /v3/api/contact format. Renamed the file to make this clear, and updated tests that depend on them.

PR #952, opened 2017-11-16
Acceptance tests collections and Investment Projects create

feedmypixel:
This work is around AT, it is all related and I have separated it out into sensible commits

### TL;DR
- Investment Project create AT
- Bit update to coverage on collections AT
- numerous small bug fixes
  - Removal of CTI from Investment Project Flow
  - Sorting of contacts in company
  - exposure of channel in interactions
  - fixing of event input error in create Investment projects
  - how we show dates in Investment Projects collections
  - remove appending of "United Kingdom" to non uk companies
- AT helpers around state and obtaining information from a form to submit it

PR #964, opened 2017-11-18
Feature/add sort and filters to contacts

feedmypixel:
⚡️  [Company contacts demo](https://datahub-beta2-pr-964.herokuapp.com/companies/0fb3379c-341c-4da4-b825-bf8d47b26baa/contacts?sortby=modified_on%3Adesc) ⚡️ 

⚡️   [Contacts demo](https://datahub-beta2-pr-964.herokuapp.com/contacts?sortby=modified_on%3Adesc) ⚡️ 

This work came about due to issues with selecting the first contact when creating one for UAT in https://github.com/uktrade/data-hub-frontend/pull/952

After digging in a little deeper it became clear that the contact work within a company `/company/<id>/contacts`, to show 2 different collections on the page needed some attention. With changes in the backend it also meant that a `active/archived` filter could easily be dropped in to the `/contacts` page

## This work:
- adds sorting and filtering to `/company/<id>/contacts`
- adds filtering of `active/archived` for `/contacts` and `/company/<id>/contacts`

## ToDO
- [x] sorting on `/company/<id>/contacts`
- [x] `active/archived` filtering on `/company/<id>/contacts`
- [x] `active/archived` filtering on `/contacts`
- [x] Unit tests 
- [x] AT tests 
- [x] Make sure AT tests from https://github.com/uktrade/data-hub-frontend/pull/952 work with this work 

## Screen shot

screen shot 2017-11-20 at 20 33 43

screen shot 2017-11-20 at 20 34 49

## Screen Captures

contacts-company

contacts


ztolley:
The approach of putting the filters on the top will not scale when it comes to adding filters to other sub-collections (which is a foregone conclusion) and takes up a lot of space.

Can we try putting the filters on the left under the sub-navigation? I know you could argue you might have to scroll to see them but I think it would be better, and users will find it and use it when they need too. 

I DO agree that the current design of 2 lists in the screen doesn't work for real world companies with many contacts and the filter approach is better.

A 3rd way may be to remove the 'filters' title and have the 2 options side by side, but it still wont scale. 

Or hide the filter section and reveal using JS, like Google search

feedmypixel:
> Can we try putting the filters on the left under the sub-navigation?

Yeah tried that it felt really disjointed from the collection list. 

I had a look at nesting it under the sort so it was in the collection header which worked better visually but I had overflow problems with the macros. Which I need to spend some time to understand properly. 

> A 3rd way may be to remove the 'filters' title and have the 2 options side by side, but it still wont scale.

Also tried this. The checkbox next to each other really didn’t look great and didn’t read well and loosing the title felt like too much was lost 

My overall conclusion was that this is functional and works and to iterate. It will be good to hear the rest of the chaps thoughts to see if anything comes out of that.

ztolley:
I get the disjointed. See what the others say. Just feel it could be better, but it's better than it was... iterate

src/apps/companies/controllers/contacts.js
```diff
@@ -1,26 +1,28 @@
...
-  const transformedContacts = contacts
-    .map(transformContactToListItem)
-    .map(contact => {
-      const meta = reject(contact.meta, ['label', 'Company'])
-      return Object.assign({}, contact, { meta })
-    })
-  const activeContacts = transformedContacts.filter(contact => !contact.isArchived)
-  const archivedContacts = transformedContacts.filter(contact => contact.isArchived)
+  const sortForm = merge({}, contactSortForm, {
+    hiddenFields: Object.assign({}, omit(req.query, 'sortby')),
```
yeahfro: `assign`

src/apps/companies/middleware/contact-collection.js
```diff
@@ -0,0 +1,61 @@
...
+      page: req.query.page,
+      isAggregation: false,
+    })
+      .then(
+        transformApiResponseToCollection(
+          { query: req.query },
+          transformContactToListItem,
+          (contact) => {
+            const meta = reject(contact.meta, ['label', 'Company'])
+            return Object.assign({}, contact, { meta })
```
yeahfro: `assign`

src/apps/contacts/macros.js
```diff
@@ -19,6 +19,16 @@ const contactFiltersFields = [
...
+    name: 'archived',
+    type: 'checkbox',
+    label: 'Status',
+    options: [
+      { value: 'false', label: 'Active' },
+      { value: 'true', label: 'Archived' },
+    ],
+  },
 ].map(filter => {
   return Object.assign(filter, {
```
yeahfro: `assign`

yeahfro: assign to `{}`

src/apps/companies/controllers/contacts.js
```diff
@@ -1,26 +1,28 @@
-const { reject } = require('lodash')
+const { merge, omit } = require('lodash')
 
-const { transformContactToListItem } = require('../../contacts/transformers')
+const { contactFiltersFields: filtersFields, contactSortForm } = require('../macros')
```
teneightfive: Do we need to rename this destructured variable? Feels a little unnecessary as its only being used on line 16 so could just as easily be the name from the object.

src/apps/companies/macros.js
```diff
@@ -45,7 +45,46 @@ const companySortForm = {
   ],
 }
 
+const contactSortForm = {
```
teneightfive: Can we not share the sort form from the collection? I imagine it's very similar if not the same?

src/apps/companies/middleware/contact-collection.js
```diff
@@ -0,0 +1,61 @@
...
+        )
+      )
+
+    next()
+  } catch (error) {
+    next(error)
+  }
+}
+
+function getCompanyContactRequestBody (req, res, next) {
```
teneightfive: Could some of this file make use of the middleware from the contacts collection?

src/apps/companies/router.js
```diff
@@ -86,7 +87,13 @@ router.use('/:companyId', setLocalNav(LOCAL_NAV))
 
 router.get('/:companyId', redirectToFirstNavItem)
 router.get('/:companyId/details', renderDetails)
-router.get('/:companyId/contacts', renderContacts)
+router.get('/:companyId/contacts',
+  setDefaultQuery(DEFAULT_COLLECTION_QUERY),
+  getCompanyContactRequestBody,
+  getCompanyContactCollection,
+  renderContacts,
```
teneightfive: Does this trailer comma not trip an error? Not sure we need it in a list of parameters.

src/apps/companies/router.js
```diff
@@ -86,7 +87,13 @@ router.use('/:companyId', setLocalNav(LOCAL_NAV))
 
 router.get('/:companyId', redirectToFirstNavItem)
 router.get('/:companyId/details', renderDetails)
-router.get('/:companyId/contacts', renderContacts)
+router.get('/:companyId/contacts',
+  setDefaultQuery(DEFAULT_COLLECTION_QUERY),
+  getCompanyContactRequestBody,
```
teneightfive: Should we use the more descriptive pattern of `set` rather than `get` for middleware that sets a local. In effect this method isn't returning anything so `get` feels inappropriate.

src/apps/contacts/macros.js
feedmypixel: I don't feel we need to assign to an empty object here to break reference - this is merging items on to an existing object

src/apps/companies/middleware/contact-collection.js
feedmypixel: So for `company/<id>/contacts` we drop out any company information as you know which company it is for. I think reuse here would create complicated and couple the methods between the two. Maybe there is an opportunity for refactor in the future

src/apps/companies/router.js
feedmypixel: > Does this trailer comma not trip an error? 
nope

>Not sure we need it in a list of parameters.
agreed

teneightfive: > nope

So I believe this is based on node version. I remember Tyom adding a trailing comma and it breaking in my version. I would say better to not add a trailing comma for function calls to be safe.

src/apps/companies/controllers/contacts.js
```diff
@@ -1,26 +1,37 @@
...
+  const visibleFiltersFields = companyContactFiltersFields.filter(field => {
+    if (field.name === 'company_uk_region') {
+      return isUKSelected
+    }
+    return true
+  })
 
   res
-    .breadcrumb(name, `/companies/${id}`)
+    .breadcrumb(companyName, `/companies/${companyId}`)
```
teneightfive: Not 100% sure this change was needed but if you feel it's easier to read then happy with it.

src/apps/companies/views/contacts.njk
```diff
@@ -7,20 +7,20 @@
...
-    items: activeContacts,
-    count: activeContacts.length,
-    countLabel: 'active contact',
-    summaryActionsHTML: addProjectButton
-  }) }}
+  {{ CollectionFilters({
+      query: QUERY,
+      filtersFields: filtersFields,
+      action: CURRENT_PATH
+    })
```
teneightfive: Could we keep the formatting same as before? The `}}` dropped on to the next line doesn't line up with the calls above.

src/apps/companies/views/contacts.njk
```diff
@@ -7,20 +7,20 @@
...
-    }) }}
-  {% endif %}
+  {{ Collection(contactResults | assign({
+      countLabel: 'contact',
+      sortForm: assign({}, sortForm, {action: CURRENT_PATH}),
+      selectedFilters: selectedFilters,
+      query: QUERY,
+      action: CURRENT_PATH,
+      summaryActionsHTML: addProjectButton
+    }))
```
teneightfive: Same as above with the bracket formatting.

PR #965, opened 2017-11-20
Handle no advisers when saving project members

richtier:
## the bug

The update project member middleware did not handle the case when all advisers are removed from the project.

no-adviser-bug

## the fix
no-adviser-fix

### DIFF
--- a/src/apps/investment-projects/middleware/forms/team-members.js
+++ b/src/apps/investment-projects/middleware/forms/team-members.js
@@ -1,4 +1,4 @@
-const { zipWith, get } = require('lodash')
+const { zipWith, get, isArray, isString } = require('lodash')
 const { getAdvisers } = require('../../../adviser/repos')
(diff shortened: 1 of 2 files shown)


src/apps/investment-projects/middleware/forms/team-members.js
```diff
@@ -7,17 +7,13 @@ const { transformObjectToOption } = require('../../../transformers')
...
-    }]
+  let teamMembers = []
+  if (isArray(body.adviser)) {
+    teamMembers = zipWith(body.adviser, body.role, (adviser, role) => ({ adviser, role }))
+  } else if (isString(body.adviser)) {
+    teamMembers = [{ adviser: body.adviser, role: body.role }]
   }
-
-  return []
+  return teamMembers.filter(member => member.adviser.length > 0)
```
yeahfro: `member.adviser.length` rather than `member.adviser.length > 0` would be fine

PR #967, opened 2017-11-20
make retrieving nested object property more robust - fixes sentry #2351

feedmypixel:
as the title says ☝️


teneightfive:
@feedmypixel I believe this has already just been fixed in #960. 

Might be able to close this one and reference in the sentry error.

feedmypixel:
fixed in https://github.com/uktrade/data-hub-frontend/pull/960

PR #969, opened 2017-11-20
Rename company transformers

ztolley:
After discuissions it was decided that transformer names could be simplified. This changes the name from entityResponseToViewRecord to entityToView


src/apps/companies/controllers/details.js
```diff
@@ -1,7 +1,7 @@
 const {
-  transformCompanyResponseToViewRecord,
-  transformCompaniesHouseResponseToViewRecord,
-  transformCompanyResponseToOneListViewRecord,
+  transformCompanyToView,
+  transformCompaniesHouseToView,
```
teneightfive: `transformCompaniesHouseToView`-> `transformCompaniesHouseCompanyToView`?

Never been sure what the best thing to names these is. Feels like the first is missing that it is still a _company_ but then when you add that it feels a bit repetitive. What do you think? 🤔

ztolley: I think it is potentially confusing and adding company doesn't help in any way, hence I picked this.

PR #973, opened 2017-11-20
403 support

teneightfive:
This change adds better support when the API returns a `403` status code for OMIS orders.

It adds work around the global error handler to support a `403` code and render a more useful message to users.

It also adds the work to error OMIS middleware correctly.

#### 404

image

#### 500

image

#### 404

image

#### 403

image

#### 500

image

### DIFF
--- a/src/middleware/errors.js
+++ b/src/middleware/errors.js
@@ -1,23 +1,31 @@
 const config = require('../../config')
 const logger = require('../../config/logger')
 
+function getStatusMessage (error) {
+  if (error.code === 'EBADCSRFTOKEN') {
+    return 'This form has been tampered with'
+  }
+
+  if (error.statusCode === 404) {
+    return 'Page not found'
+  }
+
+  if (error.statusCode === 403) {
+    return 'You don’t have permission to view this page'
+  }
+
+  return 'Page unavailable'
+}
+
 function notFound (req, res, next) {
   const error = new Error('Not Found')
   error.statusCode = 404
-  res.locals.BREADCRUMBS = null
 
   next(error)
 }
 
 function catchAll (error, req, res, next) {
-  const statusCode = error.statusCode = (error.statusCode || 500)
-  let statusMessage = statusCode === 404
-    ? 'Sorry we couldn\'t find that page!'
-    : 'Sorry something has gone wrong!'
-
-  if (error.code === 'EBADCSRFTOKEN') {
-    statusMessage = 'This form has been tampered with'
-  }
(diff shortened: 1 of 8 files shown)


src/middleware/errors.js
```diff
@@ -1,23 +1,31 @@
...
+  if (error.code === 'EBADCSRFTOKEN') {
+    return 'This form has been tampered with'
+  }
+
+  if (error.statusCode === 404) {
+    return 'Page not found'
+  }
+
+  if (error.statusCode === 403) {
+    return 'You don’t have permission to view this page'
```
feedmypixel: is `’` the correct character?

src/middleware/errors.js
```diff
@@ -1,23 +1,31 @@
 const config = require('../../config')
 const logger = require('../../config/logger')
 
+function getStatusMessage (error) {
+  if (error.code === 'EBADCSRFTOKEN') {
+    return 'This form has been tampered with'
```
feedmypixel: not sure how useful this message is to the user?

src/templates/errors.njk
```diff
@@ -1,16 +1,33 @@
...
 {% block body_main_content %}
-  {% if devErrorDetail %}
+  {% if statusCode == '404' %}
+    <p>If you entered a web address check it was correct.</p>
+
+    <p>If you followed a link here <a href="/support">let us know</a>.</p>
+  {% elseif statusCode == '403' and error.code === 'EBADCSRFTOKEN' %}
+    <p>Try to submit the form again in a few moments.</p>
+
+    <p>If you’re still having trouble <a href="/support">report a problem</a>.</p>
```
feedmypixel: same with the `’` looks funny should it not be `'`

src/middleware/errors.js
teneightfive: Yes, I think we should be using that character in sentences rather than `'`. If you look on GOV.UK this is what they use.

src/middleware/errors.js
teneightfive: This is just the heading to let them know something has happened to the form. The body copy in `errors.njk` gives some actions and more explanation.

teneightfive:
>Also are the screenshots correct?

No, I actually forgot to update them after I updated the text from the CRM ticket 😳They're more for reference that extra body text was added.

>A few small comments about the characters used.

Took me a little while to actually dig out a good article on the reason but found this one:
https://webdesignledger.com/common-typography-mistakes-apostrophes-versus-quotation-marks/

So we should try and use `’` where we remember for text.

PR #974, opened 2017-11-20
Handle *_id fields when converting field names to labels in audit log

richtier:
The `labels.js` files are not expecting a few fields that end with `_id`, meaning users are shown the field name instead of the label in audit pages. Note `labels.js` expects `"account_manager"` but not `"account_manager_id"`

This change handles that by removing `_id` from the field name during the label lookup, but it might be better to update `labels.js` so `account_manager` becomes `account_manager_id` - but I'm not sure because:
 - there may be backwards compatibility issues as field names changed over time
 - I don't know enough about the backend to know which fields should end in an `_id`.

thoughts @ztolley @teneightfive @feedmypixel ?

## after the change
image

## before the chage
image

### DIFF
--- a/src/apps/audit/transformers.js
+++ b/src/apps/audit/transformers.js
@@ -5,7 +5,7 @@ const { mediumDateTimeFormat } = require('../../../config')
 
 function transformChanges (changes, labels) {
   return Object.keys(changes)
-    .map(key => labels[key] || key)
+    .map(key => labels[key] || labels[key.replace(/_id$/, '')] || key)
     .join(', ')
 }
 
@@ -46,4 +46,5 @@ function transformAuditLogToListItem (labels = {}) {
 
 module.exports = {
   transformAuditLogToListItem,
+  transformChanges,
 }
--- a/src/apps/audit/transformers.js
+++ b/src/apps/audit/transformers.js
@@ -3,6 +3,18 @@ const { get } = require('lodash')
 const dateFns = require('date-fns')
 const { mediumDateTimeFormat } = require('../../../config')
 
+/**
+ * Transforms an audit log of changes into a common seperated list
+ * summerizing the fields that were modified in the change.
+ *
+ * Note that the field name that changed is stripped of '_id', if it ends in it.
+ * by convention the back end API includes the _id part of a fieldname in the
+ * original table even though all of references to the field do not include it
+ *
+ *
+ * @param {Array[Object]} changes
+ * @param {Object} labels
+ */
 function transformChanges (changes, labels) {
   return Object.keys(changes)
     .map(key => labels[key] || labels[key.replace(/_id$/, '')] || key)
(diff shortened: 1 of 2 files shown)


teneightfive:
I don't know this piece of code well enough but I'm not sure I'm keen on removing `_id` from the field name.

richtier:
@teneightfive I see no other option because:

1. we can't replace the existing `account_manager` key in `labels.js` with `account_manager_id` because some other code treats labels.js as a whitelist of resource fields (e.g., company record). The company record contains `account_manager`, not `account_manager_id`.

2. `lables.js` cannot have duplicate *values* because [`getDataLabels`](https://github.com/uktrade/data-hub-frontend/blob/0136b597e73c6a9bae198f2d64b889ced27a693b/src/lib/controller-utils.js#L116) inverts the object so `{'key': 'value'}` becomes `{'value': 'key'}`...so adding the following to labels.js break a lot of things:

```
  account_manager: 'Account manager',
  account_manager_id: 'Account manager',
```

feedmypixel:
This feels quite magic I would prefer to be more explicit and rename the work in `labels.js` otherwise this will just come back to bite us. It also makes it difficult for a developer to work out what is going on in the future, which again is a false economy.

> be better to update labels.js so account_manager becomes account_manager_id

I would vote for this

> there may be backwards compatibility issues as field names changed over time

Are there?

> I don't know enough about the backend to know which fields should end in an _id.

You can have a search on the [data-hub-leeoo repo](https://github.com/uktrade/data-hub-leeloo) or speak with one of our BE developers @reupen @elcct @marcofucci

richtier:
summary of the problem:

1. this code wants to use `labels.js` when transforming backend field names to human-friendly labels for the audit log: https://github.com/uktrade/data-hub-frontend/blob/develop/src/apps/audit/transformers.js#L7

2. this code want to use labels.js as a whitelist for resources returned by the api: https://github.com/uktrade/data-hub-frontend/blob/develop/src/apps/companies/transformers/companies-house-to-view.js#L45

3. this code expects values in `labels.js` must be unique: https://github.com/uktrade/data-hub-frontend/blob/0136b597e73c6a9bae198f2d64b889ced27a693b/src/lib/controller-utils.js#L116

combine these three uses of `labels.js` and we're a bit painted into a cornder.

Point 1 and 2 mean are mutually exclusive because the audit log has the field name `campaign_manager_id`, while company resource has `company_manager`, so we cannot do this:
```
{
account_manager_id: 'Account manager',  // changed account_manager to account_manager_id
...
}
```
Doing that would result in company resource no longer working

Point 3 means we cannot do this because values must be unique.

```
{
 account_manager: 'Account manager',
 account_manager_id: 'Account manager',
  ...
}
```

ztolley:
So I'm going to assume that django uses the convention of naming a field with _id when it is related to something else, in which case this solution works.

I also understand the concern that it may not be obvious to future developers the reason for stripping _id from a field.

Can we go with this solution and explain in the code Why the _id is stripped? Makes maintaining the labels a little easier as it may not be obvious to future developers that they sometimes need to add 2 items to the list instead of one. I know as a dev when I add a field I wouldn't realise I would have to do that.

ztolley:
The alternative would be to strip the _id from the field in the backend api call

ztolley:
Actually the more I think about it the more I think this is a backend issue. The back end is inconsistent in its field naming and that is causing this issue. The underlying reason is the way the serializers work and name fields compared to the raw field name.

ztolley:
Parking this... more important fish to fry

PR #975, opened 2017-11-20
Remove inserting user data into page

ztolley:
Removed a potentially dangerous part of the template that created a javscript copy of the users' information in the webpage. Have searched through all the client side javascript and can find no reason to have this. Removing

### DIFF
--- a/config/nunjucks/filters.js
+++ b/config/nunjucks/filters.js
@@ -72,7 +72,6 @@ const filters = {
   isPlainObject,
   isString,
   pluralise,
-  stringify: JSON.stringify,
   sentenceCase: Case.sentence,
 
   encodeQueryString (value) {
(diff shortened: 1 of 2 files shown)


ztolley:
As you can see from https://github.com/uktrade/data-hub-frontend/commit/28cf84908ba4c115ee2677dcad5f9530bfd87aa4 , there is no clue why it was added and I see nothing that refers to them,

ztolley:
I wonder if this was carried over from the layouts that Tyom brought over from NHS?

teneightfive:
>I wonder if this was carried over from the layouts that Tyom brought over from NHS?

Nope, I know those layouts too and we never had any kind of authentication so would have been no use for user. 

Also, the namespace is `DIT` 😉

My gut says it was something to do with the default value for advisers in forms like interactions or contacts?

PR #976, opened 2017-11-21
Filter out advisers with no name

richtier:
The database has some bad data (advisers with no name). We can guard against bad data by filtering out those advisers.

## after the fix

image

## before the fix

image

### DIFF
--- a/src/apps/adviser/repos.js
+++ b/src/apps/adviser/repos.js
@@ -1,11 +1,17 @@
+const { assign } = require('lodash')
 const config = require('../../../config')
 const logger = require('../../../config/logger')
 const authorisedRequest = require('../../lib/authorised-request')
 
 // TODO: Make sure this is removed and replaced with a better way to
 // filter/select advisers
+// TODO: advisers with no name are filtered on the front end. Make changes on
+// the back end to make this not necessary.
 function getAdvisers (token) {
   return authorisedRequest(token, `${config.apiRoot}/adviser/?limit=100000&offset=0`)
+    .then(data => assign({}, data, {
+      results: data.results.filter(adviser => Boolean(adviser.first_name || adviser.last_name)),
+    }))
 }
 
 function getAdviser (token, id) {
--- a/test/unit/apps/adviser/repos.test.js
+++ b/test/unit/apps/adviser/repos.test.js
@@ -1,0 +1,32 @@
+const nock = require('nock')
+const adviserData = require('~/test/unit/data/advisers/advisers.json')
+const badAdviserData = require('~/test/unit/data/advisers/advisers-with-bad-data.json')
+const config = require('~/config')
+const repos = require('~/src/apps/adviser/repos')
+
+describe('Adviser repository', () => {
+  describe('getAdvisers', () => {
+    context('when an adviser without a name is encountered', () => {
+      beforeEach(() => {
+        nock(config.apiRoot)
+          .get(`/adviser/?limit=100000&offset=0`)
+          .reply(200, badAdviserData)
+      })
+      it('will be filtered out', async () => {
+        const actual = await repos.getAdvisers(123)
+        expect(actual.results.length).to.equal(4)
+      })
+    })
+    context('when all advisers have names', () => {
+      beforeEach(() => {
+        nock(config.apiRoot)
+          .get(`/adviser/?limit=100000&offset=0`)
+          .reply(200, adviserData)
+      })
+      it('will not filter out any advisers', async () => {
+        const actual = await repos.getAdvisers(123)
+        expect(actual.results.length).to.equal(5)
+      })
+    })
+  })
+})
--- a/test/unit/data/advisers/advisers-with-bad-data.json
+++ b/test/unit/data/advisers/advisers-with-bad-data.json
@@ -1,0 +1,90 @@
+{
+  "count": 6,
+  "results": [
+    {
+      "id": "a0dae366-1134-e411-985c-e4115bead28b",
+      "name": " ",
+      "last_login": null,
+      "first_name": "",
+      "last_name": "",
+      "email": "james.thing@mockexample.com",
+      "contact_email": "",
+      "telephone_number": "",
+      "dit_team": {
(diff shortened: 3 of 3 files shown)


ztolley:
What happens if the record in the DB is related to an adviser with no name? Are there any?

src/apps/adviser/repos.js
```diff
@@ -6,6 +6,9 @@ const authorisedRequest = require('../../lib/authorised-request')
 // filter/select advisers
 function getAdvisers (token) {
   return authorisedRequest(token, `${config.apiRoot}/adviser/?limit=100000&offset=0`)
+    .then(data => Object.assign({}, data, {
+      results: data.results.filter(adviser => !!(adviser.first_name || adviser.last_name)),
```
feedmypixel: Personal preference is to use `Boolean` to cast to a `Boolean` rather than `!!`. But maybe thats just preference

src/apps/adviser/repos.js
```diff
@@ -6,6 +6,9 @@ const authorisedRequest = require('../../lib/authorised-request')
 // filter/select advisers
 function getAdvisers (token) {
   return authorisedRequest(token, `${config.apiRoot}/adviser/?limit=100000&offset=0`)
+    .then(data => Object.assign({}, data, {
```
feedmypixel: please use lodash `assign`

test/unit/apps/adviser/repos.test.js
```diff
@@ -0,0 +1,29 @@
+const { cloneDeep } = require('lodash')
+const adviserData = require('~/test/unit/data/advisers/advisers.json')
+
+describe('Adviser repository', () => {
```
feedmypixel: you could use `nock` here to provide a response to the call to `${config.apiRoot}/adviser/`

src/apps/adviser/repos.js
feedmypixel: might also be worth looking at `pick` from `lodash`

src/apps/adviser/repos.js
richtier: in other PRs I've been guided to avoid mutating objects. `lodash.assign` mutates the object.

src/apps/adviser/repos.js
richtier: I'm not sure how `pick` will help here? `pick` is for picking whitelisted keys from an object. Here we want to filter based on the object's values.

richtier:
@teneightfive 

somewhere somehow bad data creeps into the app, so it seems ok for the consumer of the data to guard against it. It's not up to the data provider to know how the data consumer will use it. Indeed, it might be here that "bad data" is subjective: the fact the backend allows an adviser to not have a name suggests the backend may have a good reason to allow advisers with no name? Therefore only the frontend considers this to be "bad data".

Tour point on hiding the error is very valid, but given the above it seems the appropriate place to handle the bad data is the frontend. We could do:

- filter out the data it considers bad.
- log to sentry when it finds bad data - so a dev or someone can react and fix it.

src/apps/adviser/repos.js
feedmypixel: If you use an empty object as the first argument then there is no mutation in source objects.
The codebase preference is `assign` from lodash. IMO having different ways of doing similar things that are named the same is not ideal.

> I was led to understand that I should avoid mutating objects in the codebase for (at the very least) code hygiene reasons.

Yes this is correct and will lead to less "magical" bugs. `assign` also has this capability.

test/unit/apps/adviser/repos.test.js
```diff
@@ -0,0 +1,24 @@
...
+        adviserDataWithMissingName.results[0].first_name = ''
+        adviserDataWithMissingName.results[0].last_name = ''
+        nock(config.apiRoot)
+          .get(`/adviser/?limit=100000&offset=0`)
+          .reply(200, adviserDataWithMissingName)
+      })
+      it('will be filtered out', async () => {
+        const actual = await repos.getAdvisers(123)
+        expect(actual.results.length).to.equal(4)
+      })
```
feedmypixel: worth explicitly checking the happy path also?

feedmypixel:
@richtier can you check the validity of data issue in [#data-hub-dev](https://ditdigitalteam.slack.com/messages/C6P36E2VC/convo/G82UVL87N-1511273348.000463/) please. 

As @teneightfive has mentioned
>I'm skeptical about this fix. I'm not sure it should be the responsibility of the frontend to be filtering bad data. The frontend should be able to trust that the list of advisers returned are usable.

This actually may not be something we need to fix.

src/apps/adviser/repos.js
richtier: a the `lodash.assign` documentation suggests only two arguments are allowed. cool will update.

test/unit/apps/adviser/repos.test.js
richtier: the remaining 4 items are the happy path really - but sure, will do

richtier:
@feedmypixel [see here](https://github.com/uktrade/data-hub-frontend/pull/976#issuecomment-346058207) for my argument that filtering on the frontend may be valid.

test/unit/apps/adviser/repos.test.js
```diff
@@ -0,0 +1,35 @@
...
+const adviserData = require('~/test/unit/data/advisers/advisers.json')
+const config = require('~/config')
+const repos = require('~/src/apps/adviser/repos')
+
+describe('Adviser repository', () => {
+  describe('getAdvisers', () => {
+    context('when an adviser without a name is encountered', () => {
+      beforeEach(() => {
+        const adviserDataWithMissingName = cloneDeep(adviserData)
+        adviserDataWithMissingName.results[0].first_name = ''
```
feedmypixel: Feels like it would be more explicit just adding in an advisor to the data fixtures `/test/unit/data/advisers/advisers.json` that doesn't have a `firstname/lastname`.
If doing this has minimal knock on effects with the rest of the tests I would do that.

richtier: I'll check the effect it has on the other tests

PR #977, opened 2017-11-21
extra logging for OAuth failures

jim68000:
There have been issues with stateID mismatch reporting around the [OAuth code](https://github.com/uktrade/data-hub-frontend/blob/develop/src/apps/oauth/controllers.js#L34-L36). This work is to log out to confirm what we suspect is happening, is happening.

### DIFF
--- a/src/apps/oauth/controllers.js
+++ b/src/apps/oauth/controllers.js
@@ -32,7 +32,7 @@ async function callbackOAuth (req, res, next) {
   }
 
   if (sessionOAuthState !== stateQueryParam) {
-    return next(Error('There has been an OAuth stateId mismatch sessionOAuthState = ' + sessionOAuthState + ' stateQueryParam = ' + stateQueryParam))
+    return next(Error('There has been an OAuth stateId mismatch sessionOAuthState = ' + sessionOAuthState + ' stateQueryParam = ' + stateQueryParam + '. Original URL was req.originalUrl'))
   }
 
   try {
--- a/src/apps/oauth/controllers.js
+++ b/src/apps/oauth/controllers.js
@@ -32,7 +32,7 @@ async function callbackOAuth (req, res, next) {
   }
 
   if (sessionOAuthState !== stateQueryParam) {
-    return next(Error('There has been an OAuth stateId mismatch'))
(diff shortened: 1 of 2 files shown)


test/unit/apps/oauth/controllers/oauth.test.js
```diff
@@ -64,7 +64,7 @@ describe('OAuth controller', () => {
         this.controller.callbackOAuth(this.reqMock, this.resMock, this.nextSpy)
         expect(this.nextSpy.calledOnce).to.be.true
         expect(this.nextSpy.args[0][0] instanceof Error).to.be.true
-        expect(this.nextSpy.args[0][0].message).to.equal('There has been an OAuth stateId mismatch')
+        expect(this.nextSpy.args[0][0].message).to.have.string('There has been an OAuth stateId mismatch')
```
yeahfro: Can we test that the fields `sessionOAuthState` and `stateQueryParam` are being appended?

src/apps/oauth/controllers.js
```diff
@@ -32,7 +32,7 @@ async function callbackOAuth (req, res, next) {
   }
 
   if (sessionOAuthState !== stateQueryParam) {
-    return next(Error('There has been an OAuth stateId mismatch'))
+    return next(Error('There has been an OAuth stateId mismatch sessionOAuthState = ' + sessionOAuthState + ' stateQueryParam = ' + stateQueryParam + '. Original URL was req.originalUrl'))
```
feedmypixel: if this is for  temporary logging only please put a `//TODO` in

feedmypixel: also can you not just `logger.error` these and then check the logs?

feedmypixel: so this
```
logger.error('There has been an OAuth stateId mismatch sessionOAuthState`)
logger.error(`sessionOAuthState: ${stateQueryParam}`)
logger.error(`stateQueryParam: ${stateQueryParam}`)
logger.error(`Original URL: ${req.originalUrl}`)
```

jim68000: do I still need to `return next`?

PR #978, opened 2017-11-21
Provide a default subject for interaction lists

ztolley:
When sorting interactions by subject some interactions appear to not have a subject so the list looks broken. This change introduces a message to make the list not look broken and inform the user there is no subject.

### DIFF
--- a/test/unit/apps/interactions/transformers.test.js
+++ b/test/unit/apps/interactions/transformers.test.js
@@ -147,6 +147,26 @@ describe('Interaction transformers', () => {
       })
     })
 
+    context('when the source is an interaction with an empty subject', () => {
+      beforeEach(() => {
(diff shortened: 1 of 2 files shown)


test/unit/apps/interactions/transformers.test.js
```diff
@@ -147,6 +147,26 @@ describe('Interaction transformers', () => {
       })
     })
 
+    context('when the source is an interaction with an empty subject', () => {
+      beforeEach(() => {
+        this.transformed = transformInteractionToListItem(assign({}, mockInteraction, { subject: '' }))
```
teneightfive: This could be called in the `it` block seeing as there's only one in this instance.

ztolley: This is the pattern I've used for my recent tests as the tests as the 'transformed' value is the subject of the test, plus it just makes it easy to add more tests to the context.

PR #982, opened 2017-11-21
XSS protection

teneightfive:
This work ensures that any uses of the `safe` nunjucks filter have had any values that could have come from user inputted data already escaped.

This has been addressed after results of PEN test quickly highlighted some areas where HTML/JavaScript could be injected into the page and run.

### DIFF
--- a/src/apps/companies/views/edit.njk
+++ b/src/apps/companies/views/edit.njk
@@ -11,8 +11,7 @@
         items: [
           {
             label: businessTypeLabel | default('(type of organisation not known)'),
-            value: '&nbsp;' if isEditMode else '<a href="/companies/add-step-1">Change</a>',
-            safe: true
+            value: ('&nbsp;' if isEditMode else '<a href="/companies/add-step-1">Change</a>') | safe
           }
         ]
       }) }}
--- a/src/apps/omis/apps/edit/views/payment.njk
+++ b/src/apps/omis/apps/edit/views/payment.njk
@@ -2,9 +2,9 @@
 
 {% block fields %}
   {% set billingContact = [
-    order.billing_contact_name,
-    order.billing_email,
-    order.billing_phone
+    order.billing_contact_name | escape,
+    order.billing_email | escape,
+    order.billing_phone | escape
   ] | removeNilAndEmpty %}
 
   {% if billingContact | length %}
@@ -22,30 +22,30 @@
   {% endif %}
 
   {% set billingAddress = [
-    order.billing_address_1,
-    order.billing_address_2,
-    order.billing_address_town,
-    order.billing_address_county,
-    order.billing_address_postcode,
-    order.billing_address_country.name
+    order.billing_address_1 | escape,
+    order.billing_address_2 | escape,
+    order.billing_address_town | escape,
+    order.billing_address_county | escape,
+    order.billing_address_postcode | escape,
+    order.billing_address_country.name | escape
   ] | removeNilAndEmpty %}
 
   {% set registeredAddress = [
-    company.registered_address_1,
-    company.registered_address_2,
-    company.registered_address_town,
-    company.registered_address_county,
-    company.registered_address_postcode,
-    company.registered_address_country.name
+    company.registered_address_1 | escape,
+    company.registered_address_2 | escape,
+    company.registered_address_town | escape,
+    company.registered_address_county | escape,
+    company.registered_address_postcode | escape,
+    company.registered_address_country.name | escape
   ] | removeNilAndEmpty %}
 
   {% set tradingAddress = [
-    company.trading_address_1,
-    company.trading_address_2,
-    company.trading_address_town,
-    company.trading_address_county,
-    company.trading_address_postcode,
-    company.trading_address_country.name
+    company.trading_address_1 | escape,
+    company.trading_address_2 | escape,
+    company.trading_address_town | escape,
+    company.trading_address_county | escape,
+    company.trading_address_postcode | escape,
+    company.trading_address_country.name | escape
   ] | removeNilAndEmpty %}
 
   <div class="c-form-group">
--- a/src/apps/omis/apps/view/views/_layouts/view.njk
+++ b/src/apps/omis/apps/view/views/_layouts/view.njk
@@ -62,7 +62,7 @@
 
     {{ MetaList({
       items: [
-        { label: 'Client company', value: companyText, safe: true },
+        { label: 'Client company', value: companyText | safe },
         { label: 'Market (country)', value: order.primary_market }
       ],
       itemModifier: 'stacked'
@@ -72,7 +72,7 @@
       items: [
         { label: 'Created on', value: order.created_on, type: 'datetime' },
         { label: 'Updated on', value: order.modified_on, type: 'datetime' },
-        { label: 'Status', value: statusText, safe: true }
+        { label: 'Status', value: statusText | safe }
       ],
       itemModifier: 'stacked'
     }) }}
--- a/src/apps/omis/apps/view/views/_layouts/view.njk
+++ b/src/apps/omis/apps/view/views/_layouts/view.njk
@@ -49,7 +49,7 @@
       (expire{{ 'd' if quote.expired else 's' }} {{ FromNow({ datetime: quote.expires_on }) }})
     {%- endif -%}
 
-    {% if order.status === 'complete' %}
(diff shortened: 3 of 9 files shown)


src/templates/_macros/entity/meta-item.njk
```diff
@@ -35,11 +34,7 @@
       {% set metaItemValue = props.value.name | default(props.value) %}
     {% endif %}
 
-    {% if props.safe %}
-      {% set safeValue = metaItemValue | highlight(props.highlightTerm, true) | safe %}
-    {% else %}
-      {% set safeValue = metaItemValue | highlight(props.highlightTerm, true) %}
-    {% endif %}
+    {% set value = metaItemValue | highlight(props.highlightTerm, true) %}
```
ztolley: Have you tested what a highlighted term renders as? How will this affect search results?

src/apps/companies/views/edit.njk
```diff
@@ -11,8 +11,7 @@
         items: [
           {
             label: businessTypeLabel | default('(type of organisation not known)'),
-            value: '&nbsp;' if isEditMode else '<a href="/companies/add-step-1">Change</a>',
-            safe: true
+            value: '&nbsp;' if isEditMode else '<a href="/companies/add-step-1">Change</a>' | safe
```
ztolley: AhI like what you've done here, taken away the need for safe mode in the meta-list item

src/apps/omis/apps/edit/views/payment.njk
```diff
@@ -2,9 +2,9 @@
 
 {% block fields %}
   {% set billingContact = [
-    order.billing_contact_name,
-    order.billing_email,
-    order.billing_phone
+    order.billing_contact_name | escape,
```
ztolley: Can we not format the adders in the controller/middleware so that we don't have to escape then use safe?

src/apps/omis/apps/view/views/work-order.njk
```diff
@@ -6,31 +6,33 @@
...
 {% set contactEmail %}
   {% set emails = [
-    values.contact_email,
-    values.contact.email,
-    values.contact.email_alternative
-  ] %}
+    values.contact_email | escape,
+    values.contact.email | escape,
+    values.contact.email_alternative | escape
+  ] | removeNilAndEmpty %}
```
ztolley: can this happen in middleware or controller and not have to escape?

src/apps/omis/apps/view/views/work-order.njk
```diff
@@ -56,13 +58,13 @@
     }],
     items: [{
       label: 'Name',
-      value: contactName
+      value: contactName | safe
     }, {
       label: 'Phone',
-      value: contactPhone
+      value: contactPhone | safe
```
ztolley: just confused why we've added safe but lets talk through it

src/apps/omis/apps/view/views/_layouts/view.njk
```diff
@@ -62,7 +62,7 @@
 
     {{ MetaList({
       items: [
-        { label: 'Client company', value: companyText, safe: true },
+        { label: 'Client company', value: companyText | safe },
         { label: 'Market (country)', value: order.primary_market }
```
jim68000: This still leaves a vector open surely? If someone has renamed a company to display bad stuff, then it will be passed straight into the page here?

src/apps/omis/apps/view/views/work-order.njk
```diff
@@ -56,13 +58,13 @@
...
     }, {
       label: 'Email',
-      value: contactEmail
+      value: contactEmail | safe
```
jim68000: Isn't this the opposite of what we want? 'safe' tells Nunjucks you've dealt with escaping so it can render direct to page - is this handled in the API now?

src/apps/companies/views/edit.njk
teneightfive: Yeh, I think I added it initially because I wasn't aware you could just do this. 

Seems nicer not to complicate the macro if we don't need to.

src/apps/omis/apps/edit/views/payment.njk
teneightfive: I don't like the idea of adding markup in middleware/controller. 

One other way I was thinking was to have an address formatter that allows you to selecting the join, for example comma or line break.

You would still have the same problem though because you would need to mark the variable where you output the result as safe so that it won't escape the line breaks (if you choose line breaks as your separator).

It feels this might be the easiest way to ensure we don't just allow markup into the address fields. Using escape specifically here will ensure any bad characters are escaped before we use the combined value with the safe filter.

src/apps/omis/apps/view/views/work-order.njk
teneightfive: See response on previous comment.

src/apps/omis/apps/view/views/work-order.njk
teneightfive: Not necessarily. Our goal is to ensure any data that has been input is escaped. Nunjucks does do that by default but there are some instances that we want to render markup in a variable.

In this particular case this variable is built on lines 7-19. That set statement includes some markup that we can treat as safe as we have generated it. There is however some parts of that statement that includes data entered by a user but our way or protecting in this case is to force escape when calling those variables. So even by this point where we're marking it as safe the bad characters from that data has been escaped so it doesn't allow it to be rendered here.

Does that make sense? I'm not 100% convinced on this method but trying to see if there's an alternative when we want to combine user data with allowed markup.

src/apps/omis/apps/view/views/_layouts/view.njk
teneightfive: Same response as [this conversation](https://github.com/uktrade/data-hub-frontend/pull/982/files#r152531958).

src/apps/omis/apps/view/views/_layouts/view.njk
```diff
@@ -49,7 +49,7 @@
       (expire{{ 'd' if quote.expired else 's' }} {{ FromNow({ datetime: quote.expires_on }) }})
     {%- endif -%}
 
-    {% if order.status === 'complete' %}
+    {% if order.status === 'complete' and order.completed_on %}
```
ztolley: this fixes xss?

src/apps/omis/apps/view/views/work-order.njk
```diff
@@ -56,13 +58,13 @@
     }],
     items: [{
       label: 'Name',
-      value: contactName
+      value: contactName | safe
```
ztolley: The contact name is still potentially unsafe isn't it? Why not pass name and url ?

src/apps/omis/apps/edit/views/payment.njk
ztolley: Can we add a todo or something to remind is to write a filter or something to do this better?

src/apps/omis/apps/view/views/_layouts/view.njk
teneightfive: Yes 🙄

src/apps/omis/apps/view/views/work-order.njk
teneightfive: We've added safe here as previously the macro was marking all values as `safe`. By default we shouldn't have that but in some cases, like `MetaList` we want to accept the value as safe.

This is one of those cases as what we're doing is specifically escaping the values that come back from the API that can contain user entered data (line 3). Then adding markup around it but saying when we pass the value to the macro treat that value as safe.

The value of contact name already gets escaped on line 4 when we call that value using normal nunjucks variable brackets so we can accept it being safe by this point.

src/apps/omis/apps/view/views/work-order.njk
teneightfive: Same reason as on `contactName`.

ztolley:
omis layouts view.njk 
What happens if someone inserts markup into the company name associated with the order?
Doesn't that get put into into the companyText in line 40 and then allowed through iun line 65

I'm guessing there that status text is a system value so users can't set that with markup


quote.njk
Where does the content come from for lines 78? Could this contain markup from a field value?

work-order.njk
Have you confirmed that if the contact name contains markup that is doesn't make it's way though?

ztolley:
Ok happy that we can pass values with safe that have been set using {{}} after testing it, and that addresses the other files I had worries about.

Think there is stuff we can do better in the future (like how we handle addresses) but good to go

PR #984, opened 2017-11-22
the breadcrumb is not called on this page as the user is not logged in

feedmypixel:
Noticed that this isn't being used


feedmypixel:
@teneightfive A.M.A.Z.I.N.G
tim-and-eric-mind-blown

PR #985, opened 2017-11-22
Fix error in highlight filter when term contains regex

ztolley:
If a search term is surrounded by * or contains ? or + it breaks the regex builder within the highlight filter. This change removes regex characters from the search term before using it.

Also adds a try/catch to make the function more robust so a page does not fail if highlight does.

This was the cause of the 500 message when searching for a term like *fred*

### DIFF
--- a/config/nunjucks/filters.js
+++ b/config/nunjucks/filters.js
@@ -89,12 +89,21 @@ const filters = {
   },
 
   highlight (string, searchTerm, shouldMatchFullWord = false) {
-    if (!isString(string) || !isString(searchTerm) || !searchTerm.trim()) { return string }
+    if (!isString(string) || !isString(searchTerm) || !searchTerm.trim()) {
+      return string
+    }
 
-    const regEx = new RegExp(`(${searchTerm})${shouldMatchFullWord ? '\\b' : ''}`, 'gi')
-    const result = string.replace(regEx, '<span class="u-highlight">$1</span>')
+    try {
+      // Remove regex characters from the search term
+      // as they wont be in the result and will cause an error in the regular expression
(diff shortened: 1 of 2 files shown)


test/unit/config/nunjucks/filters.test.js
```diff
@@ -46,6 +46,35 @@ describe('nunjucks filters', () => {
 
       expect(highlightedString.val).to.equal(mockString)
     })
+
+    describe('should not throw an error when a term that looks like regex', () => {
```
teneightfive: What about using `context`. A good base is to start `context` and `describe` with _when_ and `it` with _should_.

teneightfive: My main point was about this initial describe that includes both _should_ and _when_. The purpose of `context` is to help separation of the scenario from the expectation.

Maybe change this to `when search term includes special characters`?

config/nunjucks/filters.js
```diff
@@ -89,12 +89,22 @@ const filters = {
...
+    try {
+      // Remove regex characters from the search term
+      // as they wont be in the result and will cause an error in the regular expression
+      const cleanTerm = searchTerm.replace(/[.*+?^${}()|[\]\\]/g, '') // $& means the whole matched string
+
+      const regEx = new RegExp(`(${cleanTerm})${shouldMatchFullWord ? '\\b' : ''}`, 'gi')
+      const result = string.replace(regEx, '<span class="u-highlight">$1</span>')
+      return new nunjucks.runtime.SafeString(result)
+    } catch (error) {
+      console.log('error')
```
teneightfive: 🚨

PR #990, opened 2017-11-23
Allow advisers to be edited after a quote has been sent

teneightfive:
Previously we stopped anything being edited when an order is not in draft. There has been a need to change the list of people on an order past this point though.

This set of changes allows assignees (advisers in the market) and subscribers (advisers in the UK) to be edited after this point. We still don't allow them to be changed if an order is cancelled or completed though.

Due to hours and lead assignee being used as part of the quote we need to stop lead adviser being changed and not allow assignees to be removed when an order is not in draft state.

## Before
localhost_3000_omis_9e9d23d4-f1e3-4116-b0c3-cc4cf4538525_work-order 1

## After
localhost_3000_omis_9e9d23d4-f1e3-4116-b0c3-cc4cf4538525_work-order

### DIFF
--- a/src/apps/omis/apps/edit/controllers/assignees.js
+++ b/src/apps/omis/apps/edit/controllers/assignees.js
@@ -7,30 +7,32 @@ const { Order } = require('../../../models')
 
 class EditAssigneesController extends EditController {
   async configure (req, res, next) {
-    const orderId = get(res.locals, 'order.id')
-    const canEditOrder = get(res.locals, 'order.canEditOrder')
-    const canEditAdvisers = get(res.locals, 'order.canEditAdvisers')
-    const token = get(req.session, 'token')
-    const advisers = await getAdvisers(token)
-    const assignees = await Order.getAssignees(token, orderId)
-    const options = advisers.results.map(transformObjectToOption)
+    try {
+      const orderId = get(res.locals, 'order.id')
+      const canEditOrder = get(res.locals, 'order.canEditOrder')
+      const canEditAdvisers = get(res.locals, 'order.canEditAdvisers')
+      const token = get(req.session, 'token')
+      const advisers = await getAdvisers(token)
+      const assignees = await Order.getAssignees(token, orderId)
+      const options = advisers.results.map(transformObjectToOption)
 
-    req.form.options.fields.assignees.options = sortBy(options, 'label')
-    req.form.options.fields.assignees.canRemove = canEditOrder
-    req.form.options.disableFormAction = !canEditAdvisers
+      req.form.options.fields.assignees.options = sortBy(options, 'label')
+      req.form.options.fields.assignees.canRemove = canEditOrder
+      req.form.options.disableFormAction = !canEditAdvisers
 
-    res.locals.order.assignees = assignees
+      res.locals.order.assignees = assignees
 
-    super.configure(req, res, next)
+      super.configure(req, res, next)
+    } catch (error) {
+      next(error)
+    }
   }
 
   async successHandler (req, res, next) {
     const data = pick(req.sessionModel.toJSON(), Object.keys(req.form.options.fields))
     const assignees = data.assignees.map((id) => {
       return {
-        adviser: {
-          id,
-        },
+        adviser: { id },
       }
     })
 
--- a/src/apps/omis/apps/edit/controllers/assignees.js
+++ b/src/apps/omis/apps/edit/controllers/assignees.js
@@ -33,8 +33,16 @@ class EditAssigneesController extends EditController {
     })
 
     try {
-      await Order.forceSaveAssignees(req.session.token, res.locals.order.id, assignees)
-      const nextUrl = req.form.options.next || `/omis/${res.locals.order.id}`
+      const orderId = get(res.locals, 'order.id')
+      const canEditOrder = get(res.locals, 'order.canEditOrder')
+      const token = get(req.session, 'token')
+      const nextUrl = get(req, 'form.options.next') || `/omis/${orderId}`
+
+      if (canEditOrder) {
+        await Order.forceSaveAssignees(token, orderId, assignees)
+      } else {
+        await Order.saveAssignees(token, orderId, assignees)
+      }
 
       req.journeyModel.reset()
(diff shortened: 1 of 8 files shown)


src/apps/omis/apps/view/views/work-order.njk
```diff
@@ -72,7 +72,7 @@
     id: 'assignees',
     heading: 'Advisers in the market',
     actions: [{
-      url: 'edit/assignees' if order.canEditOrders,
+      url: 'edit/assignees' if order.canEditAdvisers,
```
ztolley: You edited it in one commit and then change it again in another commit?

teneightfive: Yeh, was trying to make the changes more logical. One of the commits changes the name of the variable to set it up for the second variable.

The second commit then changes it from just checking if the order is in draft to checking if advisers can be edited. 

The main reason for this change was to allow advisers to be edited after a quote has been sent. But all the other parts of the quote can't be edited after this point.

teneightfive:
>I dont see much testing going on

Yes there is a bit of a gap I'm planning on plugging with our next sprint. Both with unit tests around edit and high level acceptance tests.

src/apps/omis/apps/edit/controllers/assignees.js
```diff
@@ -7,30 +7,46 @@ const { Order } = require('../../../models')
...
+    try {
+      const orderId = get(res.locals, 'order.id')
+      const canEditOrder = get(res.locals, 'order.canEditOrder')
+      const canEditAdvisers = get(res.locals, 'order.canEditAdvisers')
+      const token = get(req.session, 'token')
+      const advisers = await getAdvisers(token)
+      const assignees = await Order.getAssignees(token, orderId)
+      const options = advisers.results.map(transformObjectToOption)
+
+      req.form.options.fields.assignees.options = sortBy(options, 'label')
```
feedmypixel: worth using an `assign` here?

src/apps/omis/apps/edit/controllers/assignees.js
```diff
@@ -7,30 +7,46 @@ const { Order } = require('../../../models')
...
+      req.form.options.fields.assignees.canRemove = canEditOrder
```
feedmypixel: could this be done using `set`?

src/apps/omis/apps/edit/controllers/assignees.js
```diff
@@ -7,30 +7,46 @@ const { Order } = require('../../../models')
...
       }
     })
 
     try {
-      await Order.forceSaveAssignees(req.session.token, res.locals.order.id, assignees)
-      const nextUrl = req.form.options.next || `/omis/${res.locals.order.id}`
+      const orderId = get(res.locals, 'order.id')
+      const canEditOrder = get(res.locals, 'order.canEditOrder')
+      const token = get(req.session, 'token')
+      const nextUrl = req.form.options.next || `/omis/${orderId}`
```
feedmypixel: need a `get`?

src/apps/omis/apps/edit/controllers/subscribers.js
```diff
@@ -7,17 +7,23 @@ const { Order } = require('../../../models')
...
+    try {
+      const orderId = get(res.locals, 'order.id')
+      const canEditAdvisers = get(res.locals, 'order.canEditAdvisers')
+      const token = get(req.session, 'token')
+      const advisers = await getAdvisers(token)
+      const subscribers = await Order.getSubscribers(token, orderId)
+      const options = advisers.results.map(transformObjectToOption)
 
-    req.form.options.fields.subscribers.options = sortBy(options, 'label')
+      req.form.options.disableFormAction = !canEditAdvisers
```
feedmypixel: `set` maybe?

src/apps/omis/middleware.js
```diff
@@ -18,7 +18,8 @@ async function setOrder (req, res, next, orderId) {
     const order = await Order.getById(req.session.token, orderId)
 
     res.locals.order = assign({}, order, {
-      isEditable: order.status === 'draft',
+      canEditOrder: order.status === 'draft',
+      canEditAdvisers: !['cancelled', 'complete'].includes(order.status),
```
feedmypixel: maybe place this in a variable to better explain what it is?

src/apps/omis/apps/edit/controllers/assignees.js
teneightfive: Potentially because there are two properties being set. Will use `assign`.

src/apps/omis/apps/edit/controllers/assignees.js
teneightfive: I think that's a bit unnecessary and over-guarding. If the `assignees` property doesn't exist this controller should probably blow up. Don't want it to silently fail.

src/apps/omis/apps/edit/controllers/subscribers.js
teneightfive: `req.form.options` is available to all HMPO form controllers so can be confident that it will exist.

src/apps/omis/middleware.js
teneightfive: Not sure if that's slightly overkill. The variable is whether advisers can be edited and is being set to whether the order is in cancelled or complete.

src/apps/omis/apps/edit/controllers/assignees.js
teneightfive: On second thoughts not sure if this is harder to read.

Either as is:

```js
...

req.form.options.fields.assignees.options = sortBy(options, 'label')
req.form.options.fields.assignees.canRemove = canEditOrder
req.form.options.disableFormAction = !canEditAdvisers

...
```

or with `assign`:

```js
...

req.form.options.fields.assignees = assign({}, req.form.options.fields.assignees, {
  options: sortBy(options, 'label'),
  canRemove: canEditOrder,
})
req.form.options.disableFormAction = !canEditAdvisers

...
```

feedmypixel: yeah I was a little concerned it might over complicate

PR #992, opened 2017-11-23
Add support for disallowing removing items from form wizard field

teneightfive:
This adds support to define whether a particular field allows removal
of items. In an instance we no longer want to be able to remove items
but still be able to add.


ztolley:
Need to start writing acceptance tests

PR #994, opened 2017-11-23
EU VAT number validation

teneightfive:
This change validates the EU VAT number to make sure it includes the country code.

We need to play this back using the correct country code on the customer invoice so need to ensure the correct format is entered.


src/apps/omis/apps/edit/fields.js
```diff
@@ -31,6 +32,12 @@ function arrayRequired (value) {
   return invalidItems.length === 0
 }
 
+function validEUVatNumber (value) {
```
ztolley: Wow that looks like a whole load of fun

PR #996, opened 2017-11-23
Change company list items to display trading name below registered name

ztolley:
Previously the transformer would pick the trading name as the link if it was present, or show the registered name, but never both. This changes the list item now so the main name appears at the link and if there is a registered name it is shown below. This behaviour is consistent with how the details view works.

While working on the ticket I rewrote the unit tests to use `context` and removed tests associated with companies house results as they no longer appear in company search results.

screencapture-localhost-3000-companies-1511461461727

### DIFF
--- a/package.json
+++ b/package.json
@@ -135,6 +135,7 @@
   },
   "devDependencies": {
     "chai-as-promised": "^7.1.1",
+    "chai-subset": "^1.6.0",
     "chromedriver": "^2.33.1",
     "codeclimate-test-reporter": "^0.5.0",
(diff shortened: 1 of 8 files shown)


feedmypixel:
Please also add trading name to the company collection checks in AT tests

ztolley:
Will update the create ltd company tests so they search on, and verify the trading name and see if the others can stay as they are so they check the name and heading but will take a little while.

For now just check the PR to make sure the transformer and unit tests are ok.

PR #997, opened 2017-11-23
fix for dates that were giving errors in circle

feedmypixel:
So we have been getting random errors around dates in [circle](https://12536-83823675-gh.circle-artifacts.com/4/root/data-hub-frontend/cucumber/report.html)
download 1

I think this is because of the work in this pr setting the wrong date
There is not 31 days in November. Even in 2042!

### Circle CI data hub frontend logs
screen shot 2017-11-23 at 20 53 32


ztolley:
> this is because of the work in this pr setting the wrong date
> There is not 31 days in November. Even in 2042!

:trollface:

PR #1006, opened 2017-11-24
Make project badge settable from env vars and add demo

jim68000:
So this is in response to a request from Live Services so that users can distinguish between 'demo', which will continue for a few weeks after we push the main system live, and Data Hub production

### DIFF
--- a/config/nunjucks/globals.js
+++ b/config/nunjucks/globals.js
@@ -1,10 +1,11 @@
 const nunjucks = require('nunjucks')
 const { assign, omit, isFunction, isArray, map } = require('lodash')
 const queryString = require('query-string')
+const config = require('../../config')
 
 module.exports = {
   serviceTitle: 'Data Hub',
-  projectPhase: 'beta',
+  projectPhase: config.projectPhase,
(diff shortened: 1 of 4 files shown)


feedmypixel:
whats that for? can you add a description? Doesn't really feel like `alpha, beta, demo` belong together? 
Also please add this env in the README

config/nunjucks/globals.js
```diff
@@ -4,7 +4,7 @@ const queryString = require('query-string')
 
 module.exports = {
   serviceTitle: 'Data Hub',
-  projectPhase: 'beta',
+  projectPhase: process.env.PROJECT_PHASE || 'beta',
```
teneightfive: Can this be set in the [config file](https://github.com/uktrade/data-hub-frontend/blob/04a355bac460f108c936c6b0fa2954005d7c1441/config/index.js) please. It keeps all `process.env` references in one place.

Might also need to add an item to the readme to explain what this is for.

Maybe even with a `// TODO` if its only temporary?

jim68000: Certainly

PR #1007, opened 2017-11-24
Improve company address selection

ztolley:
There a number of issues reported, the main ones being the address on the search list for companies did not match that on the company details.

While investigating this it turns out there's 2 issues. One is that the API is overriding the company address with the companies house address if it exists, the other was the address in the company header was also doing this.

This pull request does the following:
- [x] Previously the logic to select the company address was duplicated in 2 places, and they differed. A common company address selector/formatter has been created `getCompanyAddress`.
- [x] The parameter handler, which inserted an address into the company now uses the common address function and keeps the value separate from company
- [x] The tests for the parameter handler have been improved
- [x] The company list transformer now uses the common getCompanyAddress function

As a result, the companies house address is no longer used in the page heading for company details.


feedmypixel:
You will need to update Acceptance Tests around this

ztolley:
There are unit tests, I  doubt I'll have time for AT

ztolley:
That said I just spotted a couple of failing test due to label changes.

ztolley:
Tests updated

src/apps/companies/transformers/shared.js
```diff
@@ -16,6 +18,50 @@ function transformSicCodes ({
...
+      value: getFormattedAddress({
+        address_1: trading_address_1,
+        address_2: trading_address_2,
+        address_town: trading_address_town,
+        address_county: trading_address_county,
+        address_postcode: trading_address_postcode,
+        address_country: trading_address_country,
+      }),
+    }
+  } else {
```
feedmypixel: if we don't have `isTradingAddress` do we then always have `registered_address_1`, `registered_address_town` and `registered_address_postcode`?

test/acceptance/features/investment-projects/create.feature
```diff
@@ -36,7 +36,7 @@ Feature: Create a new Investment project
     And I search for the foreign source of equity Mars Exports Ltd
     Then I can view the Equity Source in the collection
       | text               | expected                               |
-      | Registered address | investmentProject.equitySource.address |
+      | Primary address | investmentProject.equitySource.address |
```
feedmypixel: alignment 👮

test/acceptance/features/investment-projects/create.feature
```diff
@@ -56,7 +56,7 @@ Feature: Create a new Investment project
     And I search for the foreign source of equity Lambda plc
     Then I can view the Equity Source in the collection
       | text               | expected                               |
-      | Registered address | investmentProject.equitySource.address |
+      | Primary address | investmentProject.equitySource.address |
```
feedmypixel: alignment 👮

feedmypixel:
> There are some tests already there that check for this and none of the changes here affect them.

my bad I thought this was the exposing the trading name work

src/apps/companies/transformers/shared.js
ztolley: Yes.

ztolley: Thre is always either be a trading address, that has town, postcode and address 1 as required, or a registered address (which has rules that differ for UK vs foreign)

The getFormattedAddress handle if fields are null or not.

PR #1009, opened 2017-11-24
Feature/add form link helper to investment projects

feedmypixel:
This work is the first part to provide links to help people know what they need to complete in an Investment Project.

- Add links dependant on the `incomplete_fields` array sent from the backend
- remove all `(optional)` flags and let backend validation provide what is required

### Gif, starting from creating an Investment Project
investment-links

screen shot 2017-11-24 at 17 23 46


MillsWest:
This works for me. Happy with the copy

src/apps/investment-projects/helpers/buildIncompleteFormList.js
```diff
@@ -0,0 +1,79 @@
...
+    formLinks.push(get(links, formName))
+  })
+
+  return formLinks
+}
+
+function buildIncompleteFormList (incompleteFields = [], formContents = formFields, links = linkDetails) {
+  let forms = []
+
+  forEach(incompleteFields, (incompleteField) => {
```
ztolley: I'm surprised there isn't a lodash thing to calculate where two collections intersect. but this works.

assets/stylesheets/components/_messages.scss
```diff
@@ -24,6 +24,10 @@
     margin-bottom: 0;
   }
 
+  .c-message__to-complete + .c-form-actions {
+    margin-top: $default-spacing-unit;
```
teneightfive: Not sure whether this could be better tackled with assemblies?

src/apps/investment-projects/helpers/buildIncompleteFormList.js
```diff
@@ -0,0 +1,79 @@
+const { get, indexOf, forEach, uniq } = require('lodash')
```
teneightfive: Can we use kebab case for file names to be consistent with rest of codebase.

src/apps/investment-projects/middleware/shared.js
```diff
@@ -1,6 +1,7 @@
 const { get, upperFirst } = require('lodash')
 
 const metadata = require('../../../lib/metadata')
+const buildIncompleteFormList = require('../helpers/buildIncompleteFormList')
```
teneightfive: It seems theres only one method within the `helpers` directory so maybe use the pattern we've used elsewhere and create a `helpers.js` file for now that exports the `buildIncompleteFormList` method.

If it grows we can split to separate files.

assets/stylesheets/components/_messages.scss
teneightfive: In fact it looks like this adds a new element which is tied to its content.

Maybe doesn't need the extra class and could be achieved with:

```
* + .c-form-actions {
  ...
}
```

but nested within the correct selector.

src/apps/investment-projects/views/_layout.njk
```diff
@@ -44,13 +44,18 @@
...
       }) %}
-        <p>
           {% if investmentStatus.currentStage.isComplete %}
-            You have added all required information to complete this stage.
-            {% else %}
-            Complete this stage to move to the {{ investmentStatus.nextStage.name }} stage.
+            <p>You have added all required information to complete this stage.</p>
+          {% else %}
+            <p>To move to the {{ investmentStatus.nextStage.name }} stage complete the following:</p>
+            <ul class="c-message__to-complete list-disc">
```
teneightfive: Think we can drop this content specific class name.

src/apps/investment-projects/helpers/buildIncompleteFormList.js
feedmypixel: yeah I will revisit - tired code and agreed could possibly be made more elegant

feedmypixel:
@teneightfive I have addressed your comments - am merging due to time pressures

PR #1010, opened 2017-11-25
Filter inactive events in interaction creation

ztolley:
When creating an interaction there is a drop-down to select the event associated. In production this list is 40,000 records long, whilst there are only 2,000 active events. Users are reporting that the full list is not shown in the browser.

Step 1 in making the users' life a bit easier is to filter the list of events in that dropdown so that it is only 2,000 active events instead of the full 40,000 are included.

I moved the code that got that events into the events repo and use a previously created filter to filter out the disabled entries.

Like a previous dropdown this also allows the dev to pass a current value in so if an existing record is associated with a disabled event, that is still on the options list.

### DIFF
--- a/test/unit/apps/events/repos.test.js
+++ b/test/unit/apps/events/repos.test.js
@@ -1,13 +1,21 @@
+const nock = require('nock')
+
 const config = require('~/config')
+const { search } = require('~/src/apps/search/services')
+
 const token = 'abcd'
 
 describe('Event repos', () => {
   beforeEach(() => {
     this.sandbox = sinon.sandbox.create()
     this.authorisedRequestStub = this.sandbox.stub().resolves()
+    this.searchSpy = this.sandbox.spy(search)
 
     this.repos = proxyquire('~/src/apps/events/repos', {
       '../../lib/authorised-request': this.authorisedRequestStub,
+      '../search/services': {
+        search: this.searchSpy,
+      },
     })
   })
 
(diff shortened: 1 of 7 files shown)


test/unit/apps/events/repos.test.js
```diff
@@ -64,4 +69,50 @@ describe('Event repos', () => {
         .to.be.calledWith(token, `${config.apiRoot}/v3/event/123`)
     })
   })
+
+  describe('#getActiveEvents', () => {
+    context('When there is a mix of active and inactive evets', () => {
+      context('and no current event', () => {
+        beforeEach(async () => {
+          this.searchStub.resolves(eventCollection)
```
feedmypixel: we really should be using `nock` for this sort of thing so the internals of the search method actually gets used, rather than not. Lots of [examples](https://github.com/uktrade/data-hub-frontend/blob/caae0d37522a1ae2cead64e88da90277bbb656c7/test/unit/apps/oauth/controllers/oauth.test.js#L89-L91) around.

test/unit/apps/events/repos.test.js
```diff
@@ -64,4 +69,50 @@ describe('Event repos', () => {
...
+            searchEntity: 'event',
+            requestBody: { sortby: 'name:asc' },
+            token: '1234',
+            limit: 100000,
+            isAggregation: false,
+          })
+        })
+
+        it('should return only the active events', () => {
+          expect(this.events.length).to.eq(2)
```
feedmypixel: can we `deep.equal` on the expected array?

test/unit/apps/events/repos.test.js
```diff
@@ -64,4 +69,50 @@ describe('Event repos', () => {
...
+        })
+      })
+
+      context('and a current inactive event', () => {
+        beforeEach(async () => {
+          this.searchStub.resolves(eventCollection)
+          this.events = await this.repos.getActiveEvents('1234', '4aa90860-ddb8-4cc0-8d5f-e0c26567abe4')
```
feedmypixel: Maybe assign this to a vaiable to explain what it is? at the moment its a random `guid`

test/unit/apps/events/repos.test.js
```diff
@@ -64,4 +69,50 @@ describe('Event repos', () => {
...
+            searchEntity: 'event',
+            requestBody: { sortby: 'name:asc' },
+            token: '1234',
+            limit: 100000,
+            isAggregation: false,
+          })
+        })
+
+        it('should return active events and the current inactive one', () => {
+          expect(this.events.length).to.eq(3)
```
feedmypixel: same can we `deep.equal` here?

ztolley: I'm testing that it filters items out, not what is in them.

test/unit/apps/events/repos.test.js
ztolley: The thing is I don't want to know how search works, just that I call it with the correct params, the search is tested elsewhere. If I use nock I have to know how search builds the query and I can't test what I call search with.

test/unit/apps/events/repos.test.js
feedmypixel: your just testing the length right? not that it filters out the correct things?

test/unit/apps/events/repos.test.js
feedmypixel: your just testing the length right? not that it filters out the correct things?

test/unit/apps/events/repos.test.js
feedmypixel: I'd be explicit, so that if anotehr change introduces something unexpected the tests catch it

test/unit/apps/events/repos.test.js
feedmypixel: I'd be explicit, so that if another change introduces something unexpected the tests catch it

test/unit/apps/events/repos.test.js
```diff
@@ -64,4 +72,78 @@ describe('Event repos', () => {
...
+            searchEntity: 'event',
+            requestBody: { sortby: 'name:asc' },
+            token: '1234',
+            limit: 100000,
+            isAggregation: false,
+          })
+        })
+
+        it('should return only the active events', () => {
+          expect(this.events).to.deep.eq([{
```
feedmypixel: probs could do something nicer with the expected test data rather than just repeating it - but will do for now

test/unit/apps/events/repos.test.js
```diff
@@ -64,4 +72,78 @@ describe('Event repos', () => {
...
+            searchEntity: 'event',
+            requestBody: { sortby: 'name:asc' },
+            token: '1234',
+            limit: 100000,
+            isAggregation: false,
+          })
+        })
+
+        it('should return active events and the current inactive one', () => {
+          expect(this.events).to.deep.eq([{
```
feedmypixel: probably could do something nicer with the expected test data, caching it in a local const, rather than just repeating it - but will do for now

PR #1011, opened 2017-11-25
add messages to be shown before moving to the next stage

feedmypixel:
This is the second part of the messages for moving between stages on Investment Projects:

It adds a message after you have completed the required steps on a stage and before you move to the next stage. Currently there is messages for 
- `prospect` > `assign PM`
- `active` > `verify win`

## Of note
I don't think these message work particularly well and have cut them down quite a bit from copy. We also could do with linking to somewhere for the `The Trade and Investment Analysis and Performance team` but I don't have an email of web url for them.

## Prospect > Assign PM
stage

screen shot 2017-11-25 at 14 22 33

## Active > Verify win
screen shot 2017-11-25 at 14 24 12

### DIFF
--- a/src/apps/investment-projects/helpers.js
+++ b/src/apps/investment-projects/helpers.js
@@ -36,7 +36,7 @@ const formFields = {
 const linkDetails = {
   requirements: {
     url: 'edit-requirements',
-    text: 'Requirements and location',
+    text: 'Requirements and location form',
   },
   value: {
     url: 'edit-value',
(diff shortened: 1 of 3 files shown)


feedmypixel:
@MillsWest FYI

MillsWest:
I agree the contact investment team isn't ideal. Unfortunately they 

1- don't have a group email
2 - have recently changed their name

So the user journey is now bobbins. I am tempted to take it out like we did for verification.
I think it needs a conversation with Tord

feedmypixel:
> . I am tempted to take it out like we did for verification.
@MillsWest I'm inclined to agree. Its going to cause more confusion.

Currently they can just move it to Assign PM and then Tords team can filter by Assign PM stage

PR #1013, opened 2017-11-25
Fix 500 error when editing an investment interaction

ztolley:
The interaction middleware expects the investment to have a company associated with it when building edit options, but investment interactions do not.

This small change gets the company from the contact associated with the interaction if there is no company directly associated.

I know there are no acceptance tests for this but there are no acceptance tests at all for interaction edit at this time and I would like to add some post-launch to fill the gap in a separate set of pull requests.

### DIFF
--- a/src/apps/interactions/middleware/details.js
+++ b/src/apps/interactions/middleware/details.js
@@ -4,9 +4,10 @@ const { sentence } = require('case')
 const { transformInteractionFormBodyToApiRequest } = require('../transformers')
 const { fetchInteraction, saveInteraction } = require('../repos')
 const metaDataRepository = require('../../../lib/metadata')
-const { getContactsForCompany } = require('../../contacts/repos')
+const { getContactsForCompany, getContact } = require('../../contacts/repos')
 const { getAdvisers } = require('../../adviser/repos')
 const { getActiveEvents } = require('../../events/repos')
+const { getDitCompany } = require('../../companies/repos')
 
 async function postDetails (req, res, next) {
   res.locals.requestBody = transformInteractionFormBodyToApiRequest(req.body)
@@ -37,8 +38,25 @@ async function postDetails (req, res, next) {
 
 async function getInteractionDetails (req, res, next, interactionId) {
   try {
-    res.locals.interaction = await fetchInteraction(req.session.token, interactionId)
-    res.locals.company = res.locals.interaction.company
+    const token = req.session.token
+    const interaction = res.locals.interaction = await fetchInteraction(token, interactionId)
+
+    // Get the company associated with the interaction. This can be in the interaction
+    // record, or in the case of editing investment interactions it is the company
+    // associated with the interaction contact.
+    if (interaction.company) {
+      res.locals.company = interaction.company
+      return next()
+    }
+
+    const contactId = get(interaction, 'contact.id')
+    if (!contactId) {
+      return next(new Error('An interaction must have a company or contact associated with it'))
+    }
+
+    const contact = await getContact(token, contactId)
+    res.locals.company = await getDitCompany(token, contact.company.id)
+
     next()
   } catch (err) {
     next(err)
--- a/test/unit/apps/interactions/middleware/details.test.js
+++ b/test/unit/apps/interactions/middleware/details.test.js
@@ -23,6 +23,9 @@ describe('Interaction details middleware', () => {
     this.transformInteractionFormBodyToApiRequestStub = this.sandbox.stub()
     this.transformInteractionResponseToViewRecordStub = this.sandbox.stub()
     this.getContactsForCompanyStub = this.sandbox.stub()
+    this.getContactStub = this.sandbox.stub()
+    this.getDitCompanyStub = this.sandbox.stub()
+
     this.middleware = proxyquire('~/src/apps/interactions/middleware/details', {
       '../repos': {
         saveInteraction: this.saveInteractionStub.resolves({ id: '1' }),
@@ -40,11 +43,16 @@ describe('Interaction details middleware', () => {
       },
       '../../contacts/repos': {
         getContactsForCompany: this.getContactsForCompanyStub.returns(contactsData),
+        getContact: this.getContactStub,
       },
       '../../events/repos': {
         getActiveEvents: this.sandbox.stub().resolves(eventsData.results),
       },
+      '../../companies/repos': {
+        getDitCompany: this.getDitCompanyStub,
+      },
     })
+
     this.req = {
       session: {
         token: 'abcd',
@@ -58,6 +66,7 @@ describe('Interaction details middleware', () => {
         kind: 'interaction',
       },
     }
+
     this.res = {
       breadcrumb: this.sandbox.stub().returnsThis(),
       render: this.sandbox.spy(),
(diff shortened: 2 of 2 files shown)


src/apps/interactions/middleware/details.js
```diff
@@ -37,8 +38,19 @@ async function postDetails (req, res, next) {
...
+    res.locals.interaction = await fetchInteraction(token, interactionId)
+
+    // Get the company associatew the interaction. This can be in the interaction
+    // record, or in the case of editing investment interactions it is the company
+    // associated with the interaction contact.
+    if (res.locals.interaction.company) {
+      res.locals.company = res.locals.interaction.company
+    } else {
+      // Get the contact record and the company associated with it
+      const contact = await getContact(token, res.locals.interaction.contact.id)
```
teneightfive: Maybe use `get()`? Wonder if its worth checking there is a contact ID before doing this block too?

ztolley: If an interaction doesn't have a company it always has a contact, no matter what.

ztolley: If it didn't have one or the other, then I Want the screen to fail

teneightfive: I'd probably still consider using `get()`

feedmypixel: @ztolley anything nested should be using `get`

src/apps/interactions/middleware/details.js
```diff
@@ -37,8 +38,19 @@ async function postDetails (req, res, next) {
 
 async function getInteractionDetails (req, res, next, interactionId) {
   try {
-    res.locals.interaction = await fetchInteraction(req.session.token, interactionId)
-    res.locals.company = res.locals.interaction.company
+    const token = req.session.token
+    res.locals.interaction = await fetchInteraction(token, interactionId)
+
+    // Get the company associatew the interaction. This can be in the interaction
```
feedmypixel: typos

src/apps/interactions/middleware/details.js
```diff
@@ -37,8 +38,19 @@ async function postDetails (req, res, next) {
...
+    // record, or in the case of editing investment interactions it is the company
+    // associated with the interaction contact.
+    if (res.locals.interaction.company) {
+      res.locals.company = res.locals.interaction.company
```
feedmypixel: this can not only be done using `get` you can also store in a local `const`

test/unit/apps/interactions/middleware/details.test.js
```diff
@@ -161,15 +171,52 @@ describe('Interaction details middleware', () => {
   })
 
   describe('#getInteractionDetails', () => {
-    context('when success', () => {
-      it('should set interaction data on locals', async () => {
+    context('when provided an interaction with a company associated', () => {
+      beforeEach(async () => {
+        this.company = this.sandbox.mock()
+        this.interaction = assign({}, interactionData, { company: this.company })
+        this.fetchInteractionStub.resolves(this.interaction)
```
feedmypixel: can you start using `nock` more please as mentioned in the PR yesterday if you stub out methods nothing in the method gets called. If you use `nock` then the `http` call is intercepted and the methods internals are called.

test/unit/apps/interactions/middleware/details.test.js
```diff
@@ -161,15 +171,52 @@ describe('Interaction details middleware', () => {
...
+      it('should set company to the one associated with the interaction', async () => {
+        expect(this.res.locals.company).to.deep.equal(this.company)
+      })
+    })
+
+    context('when provided an investment interaction with no company', () => {
+      beforeEach(async () => {
+        this.interaction = assign({}, interactionData, {
+          company: null,
+          contact: {
```
feedmypixel: its worth putting this things in local `const` then its much easier to read what they are rather than having to work out what the object contains. You can also then reuse.

src/apps/interactions/middleware/details.js
ztolley: I'll add it in but it will make no difference. If the get can't find the thing the page will still break.

src/apps/interactions/middleware/details.js
ztolley: I like to think of it as enthusiastic typing.

test/unit/apps/interactions/middleware/details.test.js
ztolley: I think this comes down to how we want to approach testing. 

Historically we've used stub unless we were dealing with a test for a component that directly makes network calls and the majority of tests we have do this and this comes down the fact they are 'unit test' in nature, concentrating on testing one single file. 

In this case if we change the tests to use Nock then the test will be testing this file and the interactions/repos.js file. If a test fails it could be because of an issue in the repos file.

All the other tests in this file use the stub method and if I use nock for the context I added it would either mean rewriting all the tests in this file to use nock or have 2 different methods. I would prefer to use stub for the new test cases to keep the file consistent.

I agree we all need to have a catchup about testing going forward so we can all talk about Stub vs Nock and how we want to evolve testing in general.

test/unit/apps/interactions/middleware/details.test.js
ztolley: Not sure what you mean here, this takes the data already provided and just modifies it a little and is then used for assertions. You can't make it a const and making the inner properties const won't make it any easier to read.

test/unit/apps/interactions/middleware/details.test.js
feedmypixel: Not sure if your understanding what I mean.
- If using http calls, `nock` mocks the node `http` functionality so the only think that is mocked is when the code requests using `http.get|http.post`, but the code runs through everything else until that call and everything after that call.
- When you mock a function to return something the internals of that function are not tested, period. The function is

In general mocks are a smell. Have a read of this https://medium.com/javascript-scene/mocking-is-a-code-smell-944a70c90a6a

test/unit/apps/interactions/middleware/details.test.js
feedmypixel: yeah soz I'm hallucinating from lack of sleep!

test/unit/apps/interactions/middleware/details.test.js
ztolley: I understand how they work. I agree that going forward nock is a good way that we all need to talk and adopt but in this particular test it uses stubs and I want to keep consistent with that instead of have to rewrite the rest of the test file,

src/apps/interactions/middleware/details.js
```diff
@@ -37,9 +38,24 @@ async function postDetails (req, res, next) {
...
+    // record, or in the case of editing investment interactions it is the company
+    // associated with the interaction contact.
+    if (interaction.company) {
+      res.locals.company = interaction.company
+      return next()
+    }
+
+    const contactId = get(interaction, 'contact.id')
+    if (!contactId) {
+      return next(new Error('An interaction must have a company or contact associated with it'))
```
feedmypixel: should we be throwing an error like this here? is this not in need of a polite message to the user?

ztolley: I thought I was polite :) suggestions?

PR #1014, opened 2017-11-25
Active adviser options

ztolley:
Add support to `getAdvisers` to return only active advisers, plus whatever adviser is in a drop-down if it is no longer active but required to make the select work.

Modified the return format for`getAdvisers` to just return the array of advisers and not a collection interface, as it's not required.

Updated all the sub-apps to use the new return structure and only show active advisers in forms and the full list for the filter form.

Todo: The investment project still shows all advisers as some forms have multiple adviser drop downs and the work required warrants a PR of its own.


src/apps/adviser/repos.js
```diff
@@ -1,17 +1,33 @@
...
+
+/**
+ *
+ * @param {string} token Session token for API calls
+ * @param {boolean} includeDisabled Should the response include advisers marked as disabled?
+ * @param {string} currentAdviser The ID of an adviser that should be included in the list irrespective of if
+ *                                the list excludes disabled advisers and the adviser is disabled.
+ * @returns {promise[Array]} Returns an array of adviser objects
+ */
+async function getAdvisers (token, includeDisabled = true, currentAdviser = null) {
```
teneightfive: Wonder whether `includeDisabled` should be `false` by default?

Think it might also be nicer to pass options as an object so when calling it its clearer to see what it being passed.

This:

```js
getAdvisers(token, { includeDisabled: true })
```

is easier to read than:

```js
getAdvisers(token, true)
```

src/apps/adviser/repos.js
```diff
@@ -1,17 +1,33 @@
...
+  const { results } = await authorisedRequest(token, `${config.apiRoot}/adviser/?limit=100000&offset=0`)
+
+  if (includeDisabled) {
+    return results.filter(adviserHasName)
```
teneightfive: I still really don't like that we're doing this on the frontend. Feels slow. 

Can't we adapt the API to have a flag to only return advisers with name?

src/apps/adviser/repos.js
```diff
@@ -1,17 +1,33 @@
...
+  }
+
+  return results
+    .filter(filterDisabledOption(currentAdviser))
+    .filter(adviserHasName)
```
teneightfive: I thought this should only be being done in the if block if `includeDisabled` is true?

src/apps/interactions/controllers/list.js
```diff
@@ -1,11 +1,9 @@
-const { getAdvisers } = require('../../adviser/repos')
 const { interactionFiltersFieldConfig } = require('../macros')
 const { buildSelectedFiltersSummary } = require('../../builders')
 
 async function renderInteractionList (req, res, next) {
   try {
```
teneightfive: Don't need this `try`/`catch` anymore or the `async` on the function if we're not doing any async calls.

src/apps/adviser/repos.js
ztolley: I kept it true to preserve the current behaviour. I did have a version with options as params but this is not consistent with other function declarations across the application. I know they can make it easier to read though, thankfully most IDE's pickup on the jsdoc to help with autocomplete and explain params.

src/apps/adviser/repos.js
ztolley: I agree, but we are stuck with this for now.

src/apps/adviser/repos.js
ztolley: No it's applied if includeDisabled is false, so this is correct.

src/apps/adviser/repos.js
jim68000: Yeah, clear bit of tech debt. I want to sort out the API endpoints after launch as we've lost consistency.

src/apps/interactions/controllers/list.js
ztolley: I'll remove the async but there could still be an error during the filter form construction or summary.

src/apps/adviser/repos.js
teneightfive: I'd disagree with it not being consistent. The [search](https://github.com/uktrade/data-hub-frontend/blob/develop/src/apps/search/services.js#L52) method uses this format. 

I think we should make options an object. What are others thoughts on this @yeahfro @feedmypixel?

teneightfive: >I kept it true to preserve the current behaviour.

We have the option to change this behaviour if we feel its no longer correct though. 

I suppose the tricky thing is migrated orders/investment projects that need the value in the dropdown even if it was disabled...

feedmypixel:
> Add support to getAdvisers to return only active advisers, plus whatever adviser is in a drop-down if it is no longer active but required to make the work. Modified the return format forgetAdvisers to just return the array of advisers and not a collection interface, as it's not required. Updated all the sub-apps to use the new return structure and only show active advisers in forms and the full list for the filter form. Todo: The investment project still shows all advisers as some forms have multiple adviser drop downs and the work required warrants a PR of its own.

Any chance of giving this some formatting its quite hard to understand work if the description is hard to understand.

src/apps/adviser/repos.js
```diff
@@ -1,17 +1,33 @@
...
-// filter/select advisers
 // TODO: advisers with no name are filtered on the front end. Make changes on
 // the back end to make this not necessary.
-function getAdvisers (token) {
-  return authorisedRequest(token, `${config.apiRoot}/adviser/?limit=100000&offset=0`)
-    .then(data => assign({}, data, {
-      results: data.results.filter(adviser => Boolean(adviser.first_name || adviser.last_name)),
-    }))
+
+function adviserHasName (adviser) {
```
feedmypixel: should this be `adviserHasFirstOrLastName`?

feedmypixel: is the `||` what you want to be checking here? or is this meant to be an `&&`?

src/apps/adviser/repos.js
feedmypixel: @ztolley we use options in most places TBH, much easier to read.

src/apps/adviser/repos.js
feedmypixel: I don't think `adviserHasName` is doing what you want?

src/apps/interactions/middleware/details.js
```diff
@@ -47,7 +47,8 @@ async function getInteractionDetails (req, res, next, interactionId) {
 
 async function getInteractionOptions (req, res, next) {
   try {
-    res.locals.advisers = await getAdvisers(req.session.token)
+    const interactionAdvisor = get(res.locals, 'interaction.dit_advisor.id')
+    res.locals.advisers = await getAdvisers(req.session.token, false, interactionAdvisor)
```
feedmypixel: yeah this would read better as options

src/apps/investment-projects/middleware/forms/client-relationship-management.js
```diff
@@ -9,8 +9,10 @@ async function populateForm (req, res, next) {
   try {
     const investmentData = res.locals.investmentData
 
+    // Todo - update to pass parameter to just get active advisers and
+    // pass in the existing adviser id to make sure it is not filtered out
     const advisersResponse = await getAdvisers(req.session.token)
-    const advisers = advisersResponse.results.map(transformObjectToOption)
+    const advisers = advisersResponse.map(transformObjectToOption)
```
feedmypixel: `count` is outside of the results do we need it here?

test/unit/apps/adviser/repos.test.js
```diff
@@ -12,11 +12,13 @@ describe('Adviser repository', () => {
           .get(`/adviser/?limit=100000&offset=0`)
           .reply(200, badAdviserData)
       })
+
       it('will be filtered out', async () => {
         const actual = await repos.getAdvisers(123)
-        expect(actual.results.length).to.equal(4)
+        expect(actual.length).to.equal(4)
```
feedmypixel: `deep.equal` please to explicitly check

test/unit/apps/adviser/repos.test.js
```diff
@@ -25,7 +27,79 @@ describe('Adviser repository', () => {
       })
       it('will not filter out any advisers', async () => {
         const actual = await repos.getAdvisers(123)
-        expect(actual.results.length).to.equal(5)
+        expect(actual.length).to.equal(5)
```
feedmypixel: `deep.equal` please to explicitly check

test/unit/apps/adviser/repos.test.js
```diff
@@ -25,7 +27,79 @@ describe('Adviser repository', () => {
...
+
+      context('and when the caller wishes to only see active users', () => {
+        context('and when the caller does not specify an adviser that must be returned', () => {
+          beforeEach(async () => {
+            this.advisersResult = await repos.getAdvisers('1234', false)
+          })
+
+          it('should return just the active adviser', () => {
+            expect(this.advisersResult).to.deep.equal([{
+              id: '3',
```
feedmypixel: be worth putting these in local `const` so its immediately obvious what it is rather than having to read through the object e.g `expectedActiveAdvisers`

test/unit/apps/adviser/repos.test.js
```diff
@@ -25,7 +27,79 @@ describe('Adviser repository', () => {
...
+          })
+        })
+
+        context('and when the caller specified an adviser that must be returned', () => {
+          beforeEach(async () => {
+            this.advisersResult = await repos.getAdvisers('1234', false, '2')
+          })
+
+          it('should return just the active adviser', () => {
+            expect(this.advisersResult).to.deep.equal([{
```
feedmypixel: be worth putting these in local `const` so its immediately obvious what it is rather than having to read through the object e.g `expectedActiveAdvisers`

src/apps/adviser/repos.js
ztolley: The other tricky thing @teneightfive is if you say only return disabled you always need to pass it the current value or you risk breaking dropdowns so this way is by far the safest. I agree that options is better, the first version I did used options so will go back to that.

ztolley:
@feedmypixel I put formatting in but github stripped it out... reformatting again

src/apps/adviser/repos.js
ztolley: || is correct, this actually code that already existing in there, just extracted to it's own function to make it easier to read.

src/apps/adviser/repos.js
ztolley: It checks to see if they have a first or last name, some records only have first, some only have last, but it is checking to see if the adviser has a name

src/apps/adviser/repos.js
ztolley: I think the function name is fine, it's checking if they have a name, the fact its having to check for first or last is not something the caller needs to know

src/apps/investment-projects/middleware/forms/client-relationship-management.js
ztolley: No, nowhere used the count, length or any of the collection information so it just returns the array of advisers

src/apps/adviser/repos.js
feedmypixel: So we want advisers with just a `first_name` or just a `last_name`? am I understanding what this is doing?

src/apps/interactions/middleware/details.js
```diff
@@ -47,7 +47,8 @@ async function getInteractionDetails (req, res, next, interactionId) {
 
 async function getInteractionOptions (req, res, next) {
   try {
-    res.locals.advisers = await getAdvisers(req.session.token)
+    const currentAdviser = get(res.locals, 'interaction.dit_advisor.id')
+    res.locals.advisers = await getAdvisers({ token: req.session.token, includeDisabled: false, currentAdviser })
```
teneightfive: Lots of `res.locals` repetition here. Could we use `assign()`?

src/apps/adviser/repos.js
ztolley: we want advisers that have either or both, but cannot have neither

src/apps/interactions/middleware/details.js
ztolley: yarp

src/apps/adviser/repos.js
feedmypixel: Interesting, so we will have a list of peoples firstnames only?

ztolley: yarp

feedmypixel: sounds wrong, right?

ztolley: no one said cdms the data was good

src/apps/adviser/repos.js
```diff
@@ -1,17 +1,33 @@
...
+
+/**
+ *
+ * @param {string} options token Session token for API calls
+ * @param {boolean} options.includeDisabled Should the response include advisers marked as disabled?
+ * @param {string} options.currentAdviser The ID of an adviser that should be included in the list irrespective of if
+ *                                the list excludes disabled advisers and the adviser is disabled.
+ * @returns {promise[Array]} Returns an array of adviser objects
+ */
+async function getAdvisers ({ token, includeDisabled = true, currentAdviser = null }) {
```
teneightfive: Just taking another look at this I think we only need the options as an object.

I think the first param can be token like other calls to _repos_ and the option where they're not as clear can be an object.

ztolley: I copied search, which was the example given, which does this

ztolley: done

PR #1015, opened 2017-11-26
Feature/redirect back to details form

feedmypixel:
When users click through to nested forms and post results they are not returned to the details page. This work enables the redirection of a form to a url specified in the `returnToUrl` query param.
Currently this is just for the `edit-project-management` form but I can see it being useful elsewhere.

redirect

### DIFF
--- a/src/apps/investment-projects/middleware/forms/project-management.js
+++ b/src/apps/investment-projects/middleware/forms/project-management.js
@@ -22,6 +22,9 @@ async function populateForm (req, res, next) {
       },
       buttonText: 'Save',
       returnLink: `/investment-projects/${investmentData.id}/team`,
+      hiddenFields: {
+        returnUrl: get(req.query, 'returnUrl'),
+      },
     })
 
     next()
--- a/src/apps/investment-projects/router.js
+++ b/src/apps/investment-projects/router.js
@@ -150,8 +150,7 @@ router
     getBriefInvestmentSummary,
     projectManagementFormMiddleware.populateForm,
     projectManagementFormMiddleware.handleFormPost,
-    team.editProjectManagement.postHandler,
-    team.details.getDetailsHandler
+    team.editProjectManagement.postHandler
   )
 
 router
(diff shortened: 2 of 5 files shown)


src/apps/investment-projects/middleware/shared.js
```diff
@@ -14,6 +14,14 @@ function getNextStage (currentStage, projectStages) {
   return projectStages[projectStageIndex + 1]
 }
 
+function redirectToReturnUrl (req, res, next) {
+  const returnToUrl = get(req.body, 'returnToUrl')
```
teneightfive: Could we maybe call this `returnUrl`? Would make it consistent with what I've added to OMIS and then we could consistently use this for all form saves.

Could extract this out to global middleware at that point too.

src/apps/investment-projects/views/team/edit-project-management.njk
```diff
@@ -9,7 +9,11 @@
 
   {% component 'key-value-table', items=briefInvestmentSummary, variant='striped' %}
 
-  {% call Form(form) %}
+  {% call Form(form | assign({
+    hiddenFields: {
+      returnToUrl: returnToUrl
```
teneightfive: Could this not be set in the middleware where the `form` variable is probably being constructed.

Or we add support for a `returnUrl` variable to the form object itself like we do with CSRF token?

src/apps/investment-projects/middleware/forms/project-management.js
```diff
@@ -10,6 +10,7 @@ async function populateForm (req, res, next) {
 
     const advisersResponse = await getAdvisers(req.session.token)
     const advisers = advisersResponse.results.map(transformObjectToOption)
+    const returnUrl = get(res.locals, 'returnUrl')
```
teneightfive: Wonder if this could just directly look at the query rather than the locals object? As its not being passed directly to the view anymore.

src/apps/investment-projects/router.js
```diff
@@ -150,8 +150,7 @@ router
     getBriefInvestmentSummary,
     projectManagementFormMiddleware.populateForm,
     projectManagementFormMiddleware.handleFormPost,
-    team.editProjectManagement.postHandler,
-    team.editProjectManagement.getHandler
```
teneightfive: Won't it still need this to render form errors?

feedmypixel: Good spot

PR #1022, opened 2017-11-26
bump dependencies in package.json

feedmypixel:
If people are not happy with this I understand as we are close to launch and IMO do not have enough test coverage to be sure everything is ok just from running tests.

Of not I have not updated:

```
"govuk-elements-sass": "^3.1.2",
"govuk_frontend_toolkit": "^7.2.0",
"hmpo-form-wizard": "^9.5.0",
```


package.json
```diff
@@ -60,7 +60,7 @@
     "jsdom": "^11.2.0",
     "lodash": "^4.16.4",
     "markdown-it": "^8.4.0",
-    "mocha": "^3.0.2",
```
ztolley: Have you tested this in regular and watch mode? The reason I ask is that they made some breaking changes to how mocha exits. I had a suite of tests that used knex and because knex wasn't closing the connection (unless told) then the test suite would hang at the end.

https://github.com/mochajs/mocha/blob/master/CHANGELOG.md

feedmypixel:
> there's not updated lock file to go with this so did you update Yarn?

Good spot, just adding

package.json
feedmypixel: Yup yarn `watch:test` and `yarn test` working as expected

feedmypixel:
Ok so the [job is failing](https://circleci.com/gh/uktrade/data-hub-frontend/13080) due to issues with the `wavy` module after a fresh `yarn install`. Going to park this and look at it another time

feedmypixel:
> Ok so the job is failing due to issues with the wavy module after a fresh yarn install. Going to park this and look at it another time

Actually my pure Genius has fixed it 👑  :trollface:

PR #1024, opened 2017-11-27
Reinstate sign out feature

yeahfro:
Change reinstates the sign out feature including unit tests.

screen shot 2017-11-27 at 10 55 58

TODO: Redirect to correct page

### DIFF
--- a/config/index.js
+++ b/config/index.js
@@ -63,6 +63,7 @@ const config = {
     redirectUri: process.env.OAUTH2_REDIRECT_URL,
     tokenFetchUrl: process.env.OAUTH2_TOKEN_FETCH_URL,
     token: process.env.OAUTH2_DEV_TOKEN,
+    logoutUrl: process.env.OAUTH2_LOGOUT_URL,
   },
 }
 
--- a/src/apps/oauth/controllers.js
+++ b/src/apps/oauth/controllers.js
@@ -65,8 +65,15 @@ function renderHelpPage (req, res, next) {
     })
 }
 
+function signOutOAuth (req, res) {
+  req.session = null
+  res.clearCookie('datahub.sid')
+  res.redirect(config.oauth.logoutUrl)
+}
+
 module.exports = {
   callbackOAuth,
   redirectOAuth,
   renderHelpPage,
+  signOutOAuth,
 }
--- a/src/apps/oauth/router.js
+++ b/src/apps/oauth/router.js
@@ -1,7 +1,8 @@
 const router = require('express').Router()
-const { callbackOAuth, redirectOAuth } = require('./controllers')
+const { callbackOAuth, redirectOAuth, signOutOAuth } = require('./controllers')
 
 router.get('/', redirectOAuth)
 router.get('/callback', callbackOAuth)
+router.get('/sign-out', signOutOAuth)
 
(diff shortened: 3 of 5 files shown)


config/index.js
```diff
@@ -63,6 +63,7 @@ const config = {
     redirectUri: process.env.OAUTH2_REDIRECT_URL,
     tokenFetchUrl: process.env.OAUTH2_TOKEN_FETCH_URL,
     token: process.env.OAUTH2_DEV_TOKEN,
+    logoutUrl: process.env.OAUTH2_LOGOUT_URL || 'https://sso.trade.gov.uk/logout',
```
feedmypixel: do we want this in a public repo? probs just use the `OAUTH2_LOGOUT_URL`

src/apps/oauth/controllers.js
```diff
@@ -65,8 +65,16 @@ function renderHelpPage (req, res, next) {
     })
 }
 
+function signOut (req, res) {
+  req.session.returnTo = null
+  req.session.token = null
+  req.session.user = null
+  res.redirect(config.oauth.logoutUrl)
+}
```
feedmypixel: I wonder if `req.session = null` is more final?

src/apps/oauth/router.js
```diff
@@ -1,7 +1,8 @@
 const router = require('express').Router()
-const { callbackOAuth, redirectOAuth } = require('./controllers')
+const { callbackOAuth, redirectOAuth, signOut } = require('./controllers')
 
 router.get('/', redirectOAuth)
 router.get('/callback', callbackOAuth)
+router.get('/sign-out', signOut)
```
feedmypixel: can we call this `signoutOauth`?

config/index.js
teneightfive: Yep, I'd agree with that. Let's not use this value as a fallback.

src/templates/_layouts/datahub-base.njk
```diff
@@ -33,6 +33,11 @@
           <a href="/profile" class="proposition-menu__link {{ 'is-active' if CURRENT_PATH == '/profile' }}">{{ user.name }}</a>
         </li>
       {% endif %}
+      <li class="proposition-menu__item">
+        {% if user %}
```
teneightfive: Should this wrap the whole `li`? Probably don't want an empty list item if user is not signed in.

feedmypixel:
@yeahfro FYI I have tested this against `uat` SSO and found that I also needed to clear the `datahub.sid` cookie on our side.

https://github.com/uktrade/data-hub-frontend/pull/1024/files#diff-f4ecd7fceda4a51f2f7d9de1e31c6c4aR70

src/templates/_layouts/datahub-base.njk
```diff
@@ -33,6 +33,11 @@
           <a href="/profile" class="proposition-menu__link {{ 'is-active' if CURRENT_PATH == '/profile' }}">{{ user.name }}</a>
         </li>
       {% endif %}
+      {% if user %}
+      <li class="proposition-menu__item">
```
teneightfive: Maybe indent one to make it easier to read like if block above?

PR #1025, opened 2017-11-27
Change messages around!

feedmypixel:
## add a message to the Verify win stage
screen shot 2017-11-27 at 11 56 40

## remove the message from the prospect stage
screen shot 2017-11-27 at 11 57 45

### DIFF
--- a/src/apps/investment-projects/middleware/shared.js
+++ b/src/apps/investment-projects/middleware/shared.js
@@ -1,4 +1,4 @@
-const { get, upperFirst, lowerCase } = require('lodash')
+const { get, upperFirst, camelCase } = require('lodash')
 
 const metadata = require('../../../lib/metadata')
 const { buildIncompleteFormList, toCompleteStageMessages } = require('../helpers')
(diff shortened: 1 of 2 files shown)


src/apps/investment-projects/middleware/shared.js
```diff
@@ -79,7 +79,7 @@ async function getInvestmentDetails (req, res, next) {
         name: stageName,
         isComplete: investmentData.team_complete && investmentData.requirements_complete && investmentData.value_complete,
         incompleteFields: buildIncompleteFormList(get(investmentData, 'incomplete_fields', [])),
-        messages: get(toCompleteStageMessages, lowerCase(stageName), []),
+        messages: get(toCompleteStageMessages, camelCase(stageName), []),
```
teneightfive: Do you want `camelCase`? Won't this strip spaces if it's multiple words?

ztolley: it will

feedmypixel: Yup this is so it matches the value in the `toCompleteStageMessages` object

teneightfive: Ah, gotcha! Yep makes sense.

PR #1027, opened 2017-11-27
Remove disabled advisers from event organiser field.

ztolley:
Made some improvement to the adviser repo and tests

Refactored the event sub app to get all advisers for filter forms and filter out disabled advisers in the organiser field in the edit/create form.

### DIFF
--- a/src/apps/adviser/repos.js
+++ b/src/apps/adviser/repos.js
@@ -1,4 +1,4 @@
-const { assign } = require('lodash')
+const { get } = require('lodash')
 const config = require('../../../config')
 const logger = require('../../../config/logger')
 const authorisedRequest = require('../../lib/authorised-request')
@@ -9,9 +9,14 @@ const authorisedRequest = require('../../lib/authorised-request')
 // the back end to make this not necessary.
 function getAdvisers (token) {
   return authorisedRequest(token, `${config.apiRoot}/adviser/?limit=100000&offset=0`)
-    .then(data => assign({}, data, {
-      results: data.results.filter(adviser => Boolean(adviser.first_name || adviser.last_name)),
-    }))
+    .then(response => {
+      const results = response.results.filter(adviser => get(adviser, 'name', '').trim().length)
+
+      return {
+        results,
+        count: results.length,
+      }
+    })
 }
 
 function getAdviser (token, id) {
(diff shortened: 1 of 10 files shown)


src/apps/adviser/repos.js
```diff
@@ -9,9 +9,14 @@ const authorisedRequest = require('../../lib/authorised-request')
...
 function getAdvisers (token) {
   return authorisedRequest(token, `${config.apiRoot}/adviser/?limit=100000&offset=0`)
-    .then(data => assign({}, data, {
-      results: data.results.filter(adviser => Boolean(adviser.first_name || adviser.last_name)),
-    }))
+    .then(response => {
+      const results = response.results.filter(adviser => get(adviser, 'name', '').trim().length)
+
+      return {
+        count: results.length,
```
teneightfive: 👍

src/apps/adviser/repos.js
```diff
@@ -9,9 +9,14 @@ const authorisedRequest = require('../../lib/authorised-request')
...
+        results,
```
teneightfive: Swap the order of these. We use the pattern of shorthand declared first for easier reading.

src/apps/events/controllers/edit.js
```diff
@@ -3,30 +3,42 @@ const { eventFormConfig } = require('../macros')
...
-  const lead_team = eventData.lead_team || get(req, 'session.user.dit_team.id')
-  const mergedData = pickBy(merge({}, eventData, { lead_team }, res.locals.requestBody))
-  const eventForm =
-    buildFormWithStateAndErrors(
-      eventFormConfig({ eventId, advisers }),
-      mergedData,
-      get(res.locals, 'form.errors.messages'),
-    )
+async function getActiveAdvisers (token, currentAdviser) {
+  const allAdvisers = await getAdvisers(token)
```
teneightfive: try/catch around this?

src/apps/adviser/repos.js
ztolley: Wish there was a lint rule for that

src/apps/events/controllers/edit.js
ztolley: It gets caught in the renderEditPage function, I tested it.

src/apps/adviser/repos.js
teneightfive: Yeh, I looked before but couldn't find one for eslint. Think there is one for JSHint though.

src/apps/events/controllers/edit.js
teneightfive: Ah I see because its not being called as part of middleware like we've used elsewhere.

ztolley: Trying not to overdo middleware, in this case it's just split out to a function to try and keep the code clean and follow the mantra 'Do one thing and do it well'

PR #1031, opened 2017-11-27
Add unit test around clearing of cookie for sign out

yeahfro:
As the title says


test/unit/apps/oauth/controllers/oauth.test.js
```diff
@@ -199,6 +200,12 @@ describe('OAuth controller', () => {
       expect(this.reqMock.session).to.be.null
     })
 
+    it('should clear the cookie', () => {
```
ztolley: can you reorganise it so all the common stuff including the call to the controller is in the beforeEach and consider using context :) A context in which you call it when logged in and a context if it is called when the user is already logged out?

yeahfro: I have moved the common code into the `beforeEach`. The rest of your comment is beyond the scope of this change but I will ask the question.

PR #1032, opened 2017-11-27
Introduce "Then there are form fields" step definition

yeahfro:
This change introduces a step definition that could be used for all assertions of form field visibility.

### DIFF
--- a/src/apps/components/form/macros.js
+++ b/src/apps/components/form/macros.js
@@ -24,7 +24,7 @@ const standardFormConfig = {
     {
       macroName: 'MultipleChoiceField',
       type: 'radio',
-      name: 'feedbackType',
+      name: 'feedback_type',
       modifier: 'inline',
       label: 'Inline inputs',
       options: [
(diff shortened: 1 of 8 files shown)


test/acceptance/features/save/step_definitions/save.js
```diff
@@ -0,0 +1,33 @@
...
+        return {
+          label: `#group-field-${snakeCase(name)} > legend > span`,
+          element: `#group-field-${snakeCase(name)}`,
+        }
+    }
+  }
+
+  Then(/^there are form fields$/, async function (dataTable) {
+    for (const row of dataTable.hashes()) {
+      const selectors = getElementSelector(row.type, row.name)
```
feedmypixel: I would put a `waitForElementPresent` in here also

test/acceptance/features/save/step_definitions/save.js
```diff
@@ -0,0 +1,33 @@
+const { client } = require('nightwatch-cucumber')
```
feedmypixel: Not convinced of the name of this. I would say this is more a form step_definition than a save one

PR #1034, opened 2017-11-28
Hide disabled advisers for interaction advisers drop-down

ztolley:
Excludes advisers marked as disabled in the adviser dropdown in the interaction edit form, though includes the current one from the interaction record if it is disabled.

### DIFF
--- a/src/apps/interactions/middleware/details.js
+++ b/src/apps/interactions/middleware/details.js
@@ -8,6 +8,12 @@ const { getContactsForCompany, getContact } = require('../../contacts/repos')
 const { getAdvisers } = require('../../adviser/repos')
 const { getActiveEvents } = require('../../events/repos')
 const { getDitCompany } = require('../../companies/repos')
+const { filterDisabledOption } = require('../../filters')
+
+async function getActiveAdvisers (token, currentAdviser) {
+  const allAdvisers = await getAdvisers(token)
+  return allAdvisers.results.filter(filterDisabledOption(currentAdviser))
+}
 
 async function postDetails (req, res, next) {
   res.locals.requestBody = transformInteractionFormBodyToApiRequest(req.body)
@@ -65,7 +71,7 @@ async function getInteractionDetails (req, res, next, interactionId) {
 
 async function getInteractionOptions (req, res, next) {
   try {
-    res.locals.advisers = await getAdvisers(req.session.token)
+    res.locals.advisers = await getActiveAdvisers(req.session.token, get(res.locals, 'interaction.dit_adviser.id'))
     res.locals.contacts = await getContactsForCompany(req.session.token, res.locals.company.id)
     res.locals.services = await metaDataRepository.getServices(req.session.token)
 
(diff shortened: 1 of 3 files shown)


src/apps/interactions/middleware/details.js
```diff
@@ -8,6 +8,12 @@ const { getContactsForCompany, getContact } = require('../../contacts/repos')
 const { getAdvisers } = require('../../adviser/repos')
 const { getActiveEvents } = require('../../events/repos')
 const { getDitCompany } = require('../../companies/repos')
+const { filterDisabledOption } = require('../../filters')
+
+async function getActiveAdvisers (token, currentAdviser) {
+  const allAdvisers = await getAdvisers(token)
+  return allAdvisers.results.filter(filterDisabledOption(currentAdviser))
```
teneightfive: Wonder if we should adapt the filter to accept an array like we talked about before?

ztolley: No because the people to force inclusion is different for each dropdown. On one page I will have 2 dropdowns on investment and the adviser that Must be included in them will be based on That field value, so will differ

ztolley: I'm working on the investment project forms now, they will demonstrate this. And yes, we will only make 1 call to get the advisers from the back end.

src/apps/interactions/middleware/details.js
```diff
@@ -65,7 +71,7 @@ async function getInteractionDetails (req, res, next, interactionId) {
 
 async function getInteractionOptions (req, res, next) {
   try {
-    res.locals.advisers = await getAdvisers(req.session.token)
+    res.locals.advisers = await getActiveAdvisers(req.session.token, get(res.locals, 'interaction.dit_adviser.id'))
```
feedmypixel: There is an element of reverse engineering needed here to work out that you are sending in the current adviser into a method called `filterDisabledOption` to not filter then out. Not for this PR but please revisit this as just glancing at the code we could tell what is going on with much better use of variables.


Something like:
```
getActiveAdvisers(token, { currentUser } )
```

and
```
return allAdvisers.results.filter(filterDisabledOption({ keepUser: currentUser }))
```

src/apps/interactions/middleware/details.js
teneightfive: Yeh but the filter may need to support multiples. Will need this for omis. Maybe either support and array and instances where you only need one kept you pass one item in the array.

Or accept a string or an array.

Probably prefer first option.

src/apps/interactions/middleware/details.js
ztolley: In the previous iteration that's what I had, then we discovered that there are instances that will require multiple separate filtered lists to be built and we didn't want to have to make multiple calls to the server. Interaction and event are very simple, so I just created this function to do the job needed there and does just one thing.

The next PR will be investment and it will introduce something like:

```javascript
function getActiveAdvisers({ advisers, data, property }) {
  const currentAdviser = get(data, `property.id`)
  return advisers.results.filter(filterDisabledOption(currentAdviser))
}

const advisers = await getAdvisers(token)
const clientRelationshipManagerAdvisers = getActiveAdvisers({ advisers, data: investment, property: 'client_relationsip_manager' })
const referralSourceAdvisers = getActiveAdvisers({ advisers, data: investment, property: 'referral_source_adviser' })
```

PR #1035, opened 2017-11-28
Add custom implementation for formatting addresses to title case

yeahfro:
Changes:

- Maintains punctuation after formatting
- Capitalises all words

This was discovered as there is a tiny bug when formatting addresses with `V` as address line 1. Using the current implementation of title case will ignore several words, because there is a regex `/\b(A|An|And|As|At|But|By|En|For|If|In|Of|On|Or|The|To|Vs?\.?|Via)\b/g`.

`startCase` is not viable because it removes punctuation.

## Bug

screen shot 2017-11-28 at 14 19 37


src/lib/address.js
```diff
@@ -1,32 +1,42 @@
...
+    { name: `${prefix}address_1`, formatter: titleCase },
+    { name: `${prefix}address_2`, formatter: titleCase },
+    { name: `${prefix}address_town`, formatter: titleCase },
+    { name: `${prefix}address_county`, formatter: titleCase },
+    { name: `${prefix}address_postcode` },
+  ]
 
-  if (object[`${key}address_country`] && object[`${key}address_country`].name && object[`${key}address_country`].name.length > 0) {
-    addressStr += title(object[`${key}address_country`].name)
+  const addressParts = compact(keys.map((key) => {
```
feedmypixel: feels like you could add `addressCountryKey` in the `keys` array as an object (using `get`), 
Then then `map` and format if there is a `formatter` property, or even just have a property that says `format: true` and use the `titleCase` on that?

```
[
  ...
   { name: get(entity, `${prefix}address_postcode`) },
   { name: get(entity, `${prefix}address_country.name`), format: true },
]
```

```map(pickBy(addressParts, (part) => part.name)), (addressPart) => {
   return addressPart.format ? titleCase(addressPart) : addressPart
))
```

PR #1040, opened 2017-11-29
Always show uk company in invest requirements

ztolley:
Currently, the investment details screen does not show the UK company if there are no other requirements details.

This change splits the decision to show the requirements table from the decision to show the add requirements message.

In addition, the logic for this decision, and the removal of empty values have moved into the controller so different combinations can be tested.

The result is if there is just a UK company in the requirements section, it is shown with the add button/message if there are more details the edit button is correctly shown.

Before
before

After
after

Also just to prove it works when there are requirements
company furher along

### DIFF
--- a/src/apps/investment-projects/views/details.njk
+++ b/src/apps/investment-projects/views/details.njk
@@ -20,14 +20,16 @@
   {{ renderButton('details', 'Edit summary', true) }}
 
   <h2 class="heading-medium">Requirements and location</h2>
-  {% if requirements | removeNilAndEmpty | length > 1 %}
-    {% component 'key-value-table', items=requirements, variant='striped' %}
+  {% component 'key-value-table', items=requirements, variant='striped' %}
+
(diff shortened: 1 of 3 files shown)


src/apps/investment-projects/views/details.njk
```diff
@@ -20,15 +20,17 @@
   {{ renderButton('details', 'Edit summary', true) }}
 
   <h2 class="heading-medium">Requirements and location</h2>
-  {% if requirements | removeNilAndEmpty | length > 1 %}
-    {% component 'key-value-table', items=requirements, variant='striped' %}
-    {{ renderButton('requirements', 'Edit requirements', true) }}
-  {% else %}
+  {% component 'key-value-table', items=requirements, variant='striped' %}
+
+  {% if showAddRequirements %}
```
teneightfive: `isRequirementsStarted`? Then switch this condition.

Reads better than `show` because something is always shown anyway.

PR #1043, opened 2017-11-30
Removed optional from description in support form

jim68000:
And changed wording around email

### DIFF
--- a/src/apps/support/macros.js
+++ b/src/apps/support/macros.js
@@ -29,7 +29,6 @@ const feedbackFormConfig = (browserInfo) => ({
     },
     {
       macroName: 'TextField',
-      optional: true,
       type: 'textarea',
       name: 'description',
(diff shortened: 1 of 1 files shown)


src/apps/support/macros.js
```diff
@@ -39,7 +38,7 @@ const feedbackFormConfig = (browserInfo) => ({
       macroName: 'TextField',
       name: 'email',
       label: 'Email',
-      hint: 'If you wish to be contacted, please provide your email address',
+      hint: 'You need to provide your email address',
```
teneightfive: Maybe just remove this hint altogether now? Unless we want to say why we're collecting it?

teneightfive: Maybe just remove this hint altogether now? Unless we want to say why we're collecting it?

PR #1046, opened 2017-11-30
Fix active adviser filter

ztolley:
Fixes the issue using the wrong field to filter out inactive advisers.

Created a common function to fetch advisers and filter out inactive entries.
Allow caller to pass in an adviser collection to save multiple API calls if filtering multiple adviser fields
Guards against non-existence of 'is_active' property.

Updates the event edit and interaction edit screens to use this new feature.

And YES! I tested this manually

### DIFF
--- a/src/apps/adviser/filters.js
+++ b/src/apps/adviser/filters.js
@@ -1,0 +1,10 @@
+const { get } = require('lodash')
+
+function filterActiveAdvisers ({ advisers = [], includeAdviser }) {
+  return advisers.filter((adviser) => {
+    return get(adviser, 'is_active', true) || adviser.id === includeAdviser
+  })
+}
+module.exports = {
+  filterActiveAdvisers,
+}
--- a/src/apps/events/controllers/edit.js
+++ b/src/apps/events/controllers/edit.js
@@ -1,20 +1,20 @@
 /* eslint camelcase: 0 */
+const { assign, merge, get, pickBy } = require('lodash')
+
 const { eventFormConfig } = require('../macros')
 const { transformEventResponseToFormBody } = require('../transformers')
 const { buildFormWithStateAndErrors } = require('../../builders')
-const { assign, merge, get, pickBy } = require('lodash')
 const { getAdvisers } = require('../../adviser/repos')
-const { filterDisabledOption } = require('../../filters')
-
-async function getActiveAdvisers (token, currentAdviser) {
-  const allAdvisers = await getAdvisers(token)
-  return allAdvisers.results.filter(filterDisabledOption(currentAdviser))
-}
+const { filterActiveAdvisers } = require('../../adviser/filters')
 
 async function renderEditPage (req, res, next) {
   try {
     const eventData = transformEventResponseToFormBody(res.locals.event)
-    const advisers = await getActiveAdvisers(req.session.token, get(eventData, 'organiser.id'))
+
+    const currentAdviser = get(eventData, 'organiser.id')
+    const advisers = await getAdvisers(req.session.token)
+    const activeAdvisers = filterActiveAdvisers({ advisers: advisers.results, includeAdviser: currentAdviser })
+
     const eventId = get(eventData, 'id', '')
     const eventName = get(eventData, 'name')
     const lead_team = eventData.lead_team || get(req, 'session.user.dit_team.id')
@@ -22,7 +22,7 @@ async function renderEditPage (req, res, next) {
 
     const eventForm =
       buildFormWithStateAndErrors(
-        eventFormConfig({ eventId, advisers }),
+        eventFormConfig({ eventId, advisers: activeAdvisers }),
         mergedData,
         get(res.locals, 'form.errors.messages'),
       )
--- a/src/apps/interactions/middleware/details.js
+++ b/src/apps/interactions/middleware/details.js
@@ -6,14 +6,9 @@ const { fetchInteraction, saveInteraction } = require('../repos')
 const metaDataRepository = require('../../../lib/metadata')
 const { getContactsForCompany, getContact } = require('../../contacts/repos')
 const { getAdvisers } = require('../../adviser/repos')
+const { filterActiveAdvisers } = require('../../adviser/filters')
 const { getActiveEvents } = require('../../events/repos')
 const { getDitCompany } = require('../../companies/repos')
-const { filterDisabledOption } = require('../../filters')
-
-async function getActiveAdvisers (token, currentAdviser) {
-  const allAdvisers = await getAdvisers(token)
-  return allAdvisers.results.filter(filterDisabledOption(currentAdviser))
-}
 
 async function postDetails (req, res, next) {
   res.locals.requestBody = transformInteractionFormBodyToApiRequest(req.body)
@@ -71,12 +66,19 @@ async function getInteractionDetails (req, res, next, interactionId) {
 
 async function getInteractionOptions (req, res, next) {
   try {
-    res.locals.advisers = await getActiveAdvisers(req.session.token, get(res.locals, 'interaction.dit_adviser.id'))
-    res.locals.contacts = await getContactsForCompany(req.session.token, res.locals.company.id)
-    res.locals.services = await metaDataRepository.getServices(req.session.token)
+    const token = req.session.token
+    const currentAdviser = get(res.locals, 'interaction.dit_adviser.id')
+    const company = get(res.locals, 'company.id')
+    const advisers = await getAdvisers(token)
+
+    res.locals = assign({}, res.locals, {
+      advisers: filterActiveAdvisers({ advisers: advisers.results, includeAdviser: currentAdviser }),
+      contacts: await getContactsForCompany(token, company),
+      services: await metaDataRepository.getServices(token),
+    })
 
     if (req.params.kind === 'service-delivery') {
-      res.locals.events = await getActiveEvents(req.session.token)
+      res.locals.events = await getActiveEvents(token)
     }
 
     next()
--- a/test/unit/apps/events/controllers/edit.test.js
+++ b/test/unit/apps/events/controllers/edit.test.js
@@ -2,14 +2,22 @@ const { assign, find } = require('lodash')
 const nock = require('nock')
 
 const config = require('~/config')
-const eventData = require('../../../data/events/event.json')
+const eventData = require('~/test/unit/data/events/event.json')
+const adviserFilters = require('~/src/apps/adviser/filters')
 
 describe('Event edit controller', () => {
   const currentUserTeam = 'team1'
 
   beforeEach(() => {
     this.sandbox = sinon.sandbox.create()
-    this.controller = require('~/src/apps/events/controllers/edit')
+
+    this.filterActiveAdvisersSpy = this.sandbox.spy(adviserFilters, 'filterActiveAdvisers')
+
+    this.controller = proxyquire('~/src/apps/events/controllers/edit', {
+      '../../adviser/filters': {
+        filterActiveAdvisers: this.filterActiveAdvisersSpy,
+      },
+    })
 
     this.req = {
       session: {
@@ -32,23 +40,20 @@ describe('Event edit controller', () => {
 
     this.next = this.sandbox.spy()
 
-    this.advisers = [{
-      id: '1',
-      name: 'Fred Flintstone',
-      disabled_on: '2017-01-01',
-    }, {
-      id: '2',
-      name: 'Wilma Flintstone',
-      disabled_on: '2017-01-01',
-    }, {
-      id: '3',
-      name: 'Barney Rubble',
-      disabled_on: null,
-    }]
+    this.activeInactiveAdviserData = {
+      count: 5,
+      results: [
+        { id: '1', name: 'Jeff Smith', is_active: true },
+        { id: '2', name: 'John Smith', is_active: true },
+        { id: '3', name: 'Zac Smith', is_active: true },
+        { id: '4', name: 'Fred Smith', is_active: false },
+        { id: '5', name: 'Jim Smith', is_active: false },
+      ],
+    }
 
     nock(config.apiRoot)
       .get(`/adviser/?limit=100000&offset=0`)
-      .reply(200, { results: this.advisers })
+      .reply(200, this.activeInactiveAdviserData)
   })
 
   afterEach(() => {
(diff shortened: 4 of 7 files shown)


test/unit/apps/adviser/services.test.js
```diff
@@ -0,0 +1,97 @@
...
+        })
+
+        it('should include active records and the current adviser', () => {
+          const expectedAdvisers = adviserData.results.slice(0, 4)
+          expect(this.activeAdvisers).to.deep.equal(expectedAdvisers)
+        })
+      })
+    })
+
+    context('when an adviser data has no is_active property', () => {
```
yeahfro: remove `an` - sorry, silly little change

test/unit/apps/interactions/middleware/details.test.js
```diff
@@ -241,31 +234,28 @@ describe('Interaction details middleware', () => {
...
-          id: '2',
-          name: 'Wilma Flintstone',
-          disabled_on: '2017-01-01',
-        }, {
-          id: '3',
-          name: 'Barney Rubble',
-          disabled_on: null,
-        }]
+      it('should get active advisers and the current adviser', () => {
+        expect(this.getActiveAdvisersSpy).to.be.calledWith({ token: 'abcd', currentAdviser: this.currentAdviser.id })
```
yeahfro: Normally I just make sure it was `calledOnce` as an extra check

test/unit/apps/interactions/middleware/details.test.js
```diff
@@ -280,29 +270,28 @@ describe('Interaction details middleware', () => {
...
```
yeahfro: Again I do `calledOnce`

src/apps/adviser/services.js
```diff
@@ -0,0 +1,25 @@
...
+
+/**
+ * @param {Object} options
+ * @param {string} [options.token] - The API token to use if no adviser collection is provided
+ * @param {Object} [options.advisers] - An adviser array collection, comprising of count and an array of adviser objects
+ * @param {string} [option.currentAdviser] - The id of an adviser that must be included in the result array, even if it is disabled
+ *
+ * @returns {array[object]} - Returns an array of adviser objects
+ */
+async function getActiveAdvisers ({ token, advisers, currentAdviser }) {
```
yeahfro: Does this really need to be documented?

test/unit/apps/adviser/services.test.js
```diff
@@ -0,0 +1,97 @@
...
+  })
+
+  describe('#getActiveAdvisers', () => {
+    context('when the advisers data is good', () => {
+      beforeEach(() => {
+        nock(config.apiRoot)
+          .get(`/adviser/?limit=100000&offset=0`)
+          .reply(200, adviserData)
+      })
+
```
feedmypixel: can you add
 ```     
afterEach(() => {
  nock.cleanAll()
})
``` 
please

File: src/apps/adviser/services.js (hunk 4 above)
Comment (feedmypixel) [reply src/apps/adviser/services.js]: I'd agree we should be light on comments and documentation. Please document via variable/function naming and argument objects etc.

Comment (feedmypixel) [inline src/apps/adviser/services.js]: should `advisers` have a default?

File: src/apps/adviser/services.js [hunk 6]
```diff
@@ -0,0 +1,25 @@
...
+  if (!token && !advisers) {
+    return
+  }
+
+  const allAdvisers = advisers || await getAdvisers(token)
```
feedmypixel: It feels like this check should be being done outside of this method. If you are calling this method then you shouldn't have any `advisers` is too much trying to be done in this method? or is this method being used differently in two places. I would align the fuctionality

test/unit/apps/events/controllers/edit.test.js
```diff
@@ -90,13 +85,23 @@ describe('Event edit controller', () => {
...
-        expect(organiserFieldOptions).to.deep.equal(expectedAdvisers)
+      it('should get all active advisers', () => {
+        expect(this.getActiveAdvisersSpy).to.be.calledWith({
+          token: 'abcd',
+          currentAdviser: undefined,
+        })
+      })
+
+      it('should render the form with the active advisers', () => {
+        const expectedOptions = adviserData.results
```
feedmypixel: might be clearer to just use a fixture with 3 advisers in the format:

```
{ label: 'frank parks', value: '87987986986' }
```

src/apps/adviser/services.js
ztolley: I put it in because my IDE picks it up and uses it to help autocomplete my code.

test/unit/apps/interactions/middleware/details.test.js
```diff
@@ -96,6 +85,10 @@ describe('Interaction details middleware', () => {
     }
 
     this.nextSpy = this.sandbox.spy()
+
+    nock(config.apiRoot)
+      .get(`/adviser/?limit=100000&offset=0`)
+      .reply(200, adviserData)
```
feedmypixel: also please add a 
```     
afterEach(() => {
  nock.cleanAll()
})
```

test/unit/apps/interactions/middleware/details.test.js
```diff
@@ -241,31 +234,28 @@ describe('Interaction details middleware', () => {
...
-          id: '2',
+          id: adviserData.results[4].id,
         },
       })
     })
 
     context('when interaction', () => {
       beforeEach(async () => {
+        this.currentAdviser = adviserData.results[3]
+        this.res.locals.interaction.dit_adviser = this.currentAdviser
```
feedmypixel: worth using a `set`?

test/unit/apps/interactions/middleware/details.test.js
```diff
@@ -241,31 +234,28 @@ describe('Interaction details middleware', () => {
...
-          id: '3',
-          name: 'Barney Rubble',
-          disabled_on: null,
-        }]
+      it('should get active advisers and the current adviser', () => {
+        expect(this.getActiveAdvisersSpy).to.be.calledWith({ token: 'abcd', currentAdviser: this.currentAdviser.id })
+      })
 
+      it('should set the active advisers on the response object', () => {
+        const expectedAdvisers = adviserData.results.slice(0, 4)
```
feedmypixel: id be more explicit with the expected. If someone adds something into the fixture this will break

test/unit/apps/interactions/middleware/details.test.js
```diff
@@ -280,29 +270,28 @@ describe('Interaction details middleware', () => {
...
```
feedmypixel: same I'd be more explicit

src/apps/adviser/services.js
ztolley: @feedmypixel Some ide's pickup jsdoc so it makes it easier know the name or spelling of parameter names

src/apps/adviser/services.js
ztolley: You Need this, the idea is that you can pass in a list of advisers already fetched from the API to avoid repeated calls to get advisers. I am using this in the next PR for investment and @teneightfive needs support for this in Omis.

test/unit/apps/adviser/services.test.js
ztolley: I'll add it back, I had it and the tests broke. Have you merged the fix?

test/unit/apps/events/controllers/edit.test.js
ztolley: I did but then I ended up with lots all over the place so it's better to use the json file to avoid repetition.

test/unit/apps/interactions/middleware/details.test.js
ztolley: see previous comment.

test/unit/apps/interactions/middleware/details.test.js
ztolley: ?

test/unit/apps/interactions/middleware/details.test.js
ztolley: it will make it more fragile to do that. If I specify it here and someone adds a property then it will break the test.

src/apps/adviser/services.js
ztolley: No. It's optional, it's only passed when you want to use data already fetched to avoid multiple calls.

feedmypixel: @ztolley I think lets have a chat about documentation and decided if as a team we wish to do it please

feedmypixel:
So having a chat it with @ztolley I feel this is better to rename the method `getActiveAdvisers` to `filterActiveAdvisers` and to just have this method filter the advisers. The getting of the advisers should be done outside of this method.

src/apps/adviser/services.js
ztolley: it does now

test/unit/apps/events/controllers/edit.test.js
ztolley: did it anyway

test/unit/apps/interactions/middleware/details.test.js
ztolley: done

ztolley:
As @feedmypixel described, and on reflection, the previous method was not 'doing one thing and doing it well' so reworked to simply make this a filter.

Tied up the tests with suggestions and re-tested

ztolley:
Rebase and squashed as the code change alot.

test/unit/apps/interactions/middleware/details.test.js
ztolley: Cannot do that at this time as the existing test suite has a bug when this is done. @feedmypixel either we wait for that or we merge this and as part of your fix for cleanAll you include these tests.

src/apps/adviser/filters.js
```diff
@@ -0,0 +1,8 @@
+const { get } = require('lodash')
+
+function filterActiveAdvisers ({ advisers = [], includeAdviser }) {
+  return advisers.filter(adviser => (get(adviser, 'is_active', true) || adviser.id === includeAdviser))
+}
```
feedmypixel: This is really hard to read and its doing stuff we don't do around the code base, please reformat:
Alo I don't really see a need for the `get`
```
function filterActiveAdvisers ({ advisers = [], includeAdviser }) {
  return advisers.filter((adviser) => {
    return adviser.is_active || adviser.id === includeAdviser)
}
```

src/apps/events/controllers/edit.js
```diff
@@ -1,28 +1,29 @@
...
-
-async function getActiveAdvisers (token, currentAdviser) {
-  const allAdvisers = await getAdvisers(token)
-  return allAdvisers.results.filter(filterDisabledOption(currentAdviser))
-}
+const { filterActiveAdvisers } = require('../../adviser/filters')
 
 async function renderEditPage (req, res, next) {
   try {
+    const token = req.session.token
```
feedmypixel: does this need to be in a `const`? its only used once

src/apps/events/controllers/edit.js
```diff
@@ -1,28 +1,29 @@
...
     const eventData = transformEventResponseToFormBody(res.locals.event)
-    const advisers = await getActiveAdvisers(req.session.token, get(eventData, 'organiser.id'))
+
+    const currentAdviser = get(eventData, 'organiser.id')
+    const { results: advisers } = await getAdvisers(token)
```
feedmypixel: not overly sure we need this? feels like `advisers` is now something I would expect `results` on but it doesn't have `.results` after this point. Also feels like we don't really do this elsewhere

src/apps/interactions/middleware/details.js
```diff
@@ -71,13 +66,17 @@ async function getInteractionDetails (req, res, next, interactionId) {
...
-    res.locals.contacts = await getContactsForCompany(req.session.token, res.locals.company.id)
-    res.locals.services = await metaDataRepository.getServices(req.session.token)
-
-    if (req.params.kind === 'service-delivery') {
-      res.locals.events = await getActiveEvents(req.session.token)
-    }
+    const token = req.session.token
+    const currentAdviser = get(res.locals, 'interaction.dit_adviser.id')
+    const company = get(res.locals, 'company.id')
+    const { results: advisers } = await getAdvisers(token)
```
feedmypixel: same comment as previously
> not overly sure we need this? feels like advisers is now something I would expect results on but it doesn't have .results after this point. Also feels like we don't really do this elsewhere

src/apps/interactions/middleware/details.js
```diff
@@ -71,13 +66,17 @@ async function getInteractionDetails (req, res, next, interactionId) {
...
+
+    res.locals = assign({}, res.locals, {
+      advisers: filterActiveAdvisers({ advisers, includeAdviser: currentAdviser }),
+      contacts: await getContactsForCompany(token, company),
+      services: await metaDataRepository.getServices(token),
+      events: (req.params.kind === 'service-delivery') ? await getActiveEvents(token) : null,
```
feedmypixel: feels like the previous way of `if (req.params.kind === 'service-delivery') {` and being above of this res.locals code was clearer.

Maybe

```
let activeEvents
if (req.params.kind === 'service-delivery') {
  activeEvents = await getActiveEvents(token)
}

...
services: await metaDataRepository.getServices(token),
events: activeEvents
...

```

src/apps/adviser/filters.js
ztolley: that wont work, you need the get to guard against the back end deployment being out of sync with the front end.

src/apps/events/controllers/edit.js
ztolley: its down to getAdvisers returning a collection with a count and results, so having to expand and rename it when you use it. I'd rather change getAdvisers to not return the collection object and just return the array of advisers, especially as nowhere uses the collection format and just gets the results

src/apps/interactions/middleware/details.js
ztolley: I did this because every time we add multiple properties to an object @teneightfive asks people to use assign.

src/apps/adviser/filters.js
ztolley: if I extract it out and say `isActive = adviser['is_active'] || true` then @teneightfive will tell me to use `get`

test/unit/apps/interactions/middleware/details.test.js
feedmypixel: ill pick this up once in - just finishing up the `nock` work

test/unit/apps/interactions/middleware/details.test.js
feedmypixel: > it will make it more fragile to do that. If I specify it here and someone adds a property then it will break the test.

lols this is what I'm saying also

would be better to have:
```
expectedAdvisors = [
  'advisor',
  'advisor1',
  'advisor2',
]
```
Rather than slicing the results and then
>  If someone adds something into the fixture this will break

As mentioned before much better for scanning if you are making variable names also.

src/apps/adviser/filters.js
feedmypixel: ok use the `get` if you need to guard. The bigger problem here is the formatting, please also address that.

src/apps/events/controllers/edit.js
feedmypixel: you could just do `advisers.results` like we do everywhere else?

src/apps/interactions/middleware/details.js
feedmypixel: you could just do advisers.results like we do everywhere else?

src/apps/interactions/middleware/details.js
feedmypixel: yes use assign, but move the conditional up to be more explicit

src/apps/adviser/filters.js
ztolley: I have

src/apps/events/controllers/edit.js
ztolley: I have

src/apps/interactions/middleware/details.js
ztolley: I have

PR #1056, opened 2017-12-01
Tidy up nock use

feedmypixel:
This Work includes:
- Tidy up use of nock to avoid leaky global http mocks,
- Just mock http on tests where it is required,
- Introduce this.nockScope,
- Introduce "expect(this.nockScope.isDone()).to.be.true" assertions,
- Refactor a number of tests
- Make `nock` a global in unit tests
- Introduce `global-helpers.js` and include a global `afterEach` that clears down nock usage via `nock.cleanAll()`

### DIFF
--- a/test/unit/apps/adviser/repos.test.js
+++ b/test/unit/apps/adviser/repos.test.js
@@ -1,4 +1,3 @@
-const nock = require('nock')
 const adviserData = require('~/test/unit/data/advisers/advisers.json')
 const badAdviserData = require('~/test/unit/data/advisers/advisers-with-bad-data.json')
 const config = require('~/config')
--- a/test/unit/apps/adviser/repos.test.js
+++ b/test/unit/apps/adviser/repos.test.js
@@ -7,10 +7,11 @@ const repos = require('~/src/apps/adviser/repos')
 describe('Adviser repository', () => {
   describe('getAdvisers', () => {
     context('when an adviser without a name is encountered', () => {
-      beforeEach(() => {
-        nock(config.apiRoot)
+      beforeEach(async () => {
+        this.nockScope = nock(config.apiRoot)
           .get(`/adviser/?limit=100000&offset=0`)
           .reply(200, badAdviserData)
+        this.advisers = await repos.getAdvisers(123)
       })
 
       it('will be filtered out', async () => {
@@ -24,22 +25,28 @@ describe('Adviser repository', () => {
           ],
         }
 
-        const actual = await repos.getAdvisers(123)
+        expect(this.advisers).to.deep.equal(expected)
+      })
 
-        expect(actual).to.deep.equal(expected)
+      it('nock mocked scope has been called', () => {
+        expect(this.nockScope.isDone()).to.be.true
       })
     })
 
     context('when all advisers have names', () => {
-      beforeEach(() => {
-        nock(config.apiRoot)
+      beforeEach(async () => {
+        this.nockScope = nock(config.apiRoot)
           .get(`/adviser/?limit=100000&offset=0`)
           .reply(200, adviserData)
+        this.advisers = await repos.getAdvisers(123)
       })
 
-      it('will not filter out any advisers', async () => {
-        const actual = await repos.getAdvisers(123)
-        expect(actual).to.deep.equal(adviserData)
+      it('will not filter out any advisers', () => {
+        expect(this.advisers).to.deep.equal(adviserData)
+      })
+
+      it('nock mocked scope has been called', () => {
+        expect(this.nockScope.isDone()).to.be.true
       })
     })
   })
--- a/test/unit/apps/contacts/repos.test.js
+++ b/test/unit/apps/contacts/repos.test.js
@@ -1,4 +1,3 @@
-const nock = require('nock')
 const config = require('~/config')
 
 const {
--- a/test/unit/apps/contacts/repos.test.js
+++ b/test/unit/apps/contacts/repos.test.js
@@ -10,14 +10,19 @@ const companyId = '23232323'
 
 describe('Investment repository', () => {
   describe('#getContactsForCompany', () => {
-    nock(config.apiRoot)
-      .get(`/v3/contact?company_id=${companyId}&limit=500`)
-      .reply(200, contactsApiResult)
+    beforeEach(async () => {
+      this.nockScope = nock(config.apiRoot)
+        .get(`/v3/contact?company_id=${companyId}&limit=500`)
+        .reply(200, contactsApiResult)
+      this.contacts = await getContactsForCompany('token', companyId)
+    })
 
     it('should return company contacts array', async () => {
-      const actual = await getContactsForCompany('token', companyId)
+      expect(this.contacts).to.deep.equal(contactsApiResult.results)
+    })
 
-      return expect(actual).to.deep.equal(contactsApiResult.results)
+    it('nock mocked scope has been called', () => {
+      expect(this.nockScope.isDone()).to.be.true
     })
   })
 })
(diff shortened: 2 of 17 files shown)


test/unit/apps/adviser/repos.test.js
```diff
@@ -5,10 +5,19 @@ const config = require('~/config')
 const repos = require('~/src/apps/adviser/repos')
 
 describe('Adviser repository', () => {
+  beforeEach(() => {
```
ztolley: Why create a sandbox that isn't used?

test/unit/apps/contacts/repos.test.js
```diff
@@ -9,15 +9,27 @@ const contactsApiResult = require('~/test/unit/data/contacts/contact-api-result.
 const companyId = '23232323'
 
 describe('Investment repository', () => {
+  beforeEach(() => {
+    this.sandbox = sinon.sandbox.create()
```
ztolley: Same question as above

test/unit/apps/investment-projects/repos.test.js
```diff
@@ -15,73 +15,104 @@ const investmentData = require('~/test/unit/data/investment/investment-data.json
...
+    beforeEach(() => {
+      this.nockScope = nock(config.apiRoot)
+        .get(`/v3/investment/${investmentData.id}`)
+        .reply(200, investmentData)
+    })
 
-    it('should return an investment object', () => {
-      const actual = getInvestment('token', investmentData.id)
+    it('should return an investment object', async () => {
+      const actual = await getInvestment('token', investmentData.id)
```
ztolley: Well this all explains the requests that leaked but how did they ever pass?

test/unit/apps/adviser/repos.test.js
feedmypixel: `this.nockScope` is being set and checked on the sandbox

test/unit/apps/contacts/repos.test.js
feedmypixel: this.nockScope is being set and checked on the sandbox

feedmypixel: good spot, yes not needed, has been removed

test/unit/apps/adviser/repos.test.js
feedmypixel: good spot, yes not needed, has been removed

test/unit/apps/adviser/repos.test.js
```diff
@@ -27,19 +31,21 @@ describe('Adviser repository', () => {
...
 
         expect(actual).to.deep.equal(expected)
+        expect(this.nockScope.isDone()).to.be.true
       })
     })
 
     context('when all advisers have names', () => {
       beforeEach(() => {
-        nock(config.apiRoot)
+        this.nockScope = nock(config.apiRoot)
```
ztolley: Should you clean all in the same context as you set the nock? I know this means two clean alls but then you can clean just that nock scope?

test/unit/apps/contacts/repos.test.js
```diff
@@ -9,15 +9,22 @@ const contactsApiResult = require('~/test/unit/data/contacts/contact-api-result.
 const companyId = '23232323'
 
 describe('Investment repository', () => {
+  afterEach(() => {
```
ztolley: Move this down below line 22?

test/unit/apps/adviser/repos.test.js
feedmypixel: so the `nock.cleanAll()` is actually a global wipeout of all. You can't do `this.nockScope.cleanAll()` unfortunately. So having this in the outer `afterAll` will cover it for this file.

ztolley:
Approved but then had one more thought: With the move to context testing we've been trying make a test just test one thing so maybe make the testing for the mock call to be it's own test, i.e. if a test has a title about testing that it rendered the page with the correct template, why do you test that nock was called?

feedmypixel:
> Approved but then had one more thought: With the move to context testing we've been trying make a test just test one thing so maybe make the testing for the mock call to be it's own test, i.e. if a test has a title about testing that it rendered the page with the correct template, why do you test that nock was called?

Maybe, its testing that nock has returned and nock is used in that call, testing it separately wont then test it on those calls. I did wonder if it was actually needed as the return of the `nock`ed data means it has been called. But this is almost a belt and braces that the `nock` mock has been used.

Will wait to see what thoughts @yeahfro and @teneightfive have on this

test/unit/apps/adviser/repos.test.js
```diff
@@ -1,14 +1,17 @@
-const nock = require('nock')
 const adviserData = require('~/test/unit/data/advisers/advisers.json')
 const badAdviserData = require('~/test/unit/data/advisers/advisers-with-bad-data.json')
 const config = require('~/config')
 const repos = require('~/src/apps/adviser/repos')
 
 describe('Adviser repository', () => {
+  afterEach(() => {
+    nock.cleanAll()
```
teneightfive: I'm wondering if we start having a global level `beforeEach` and `afterEach`?

Could always setup a sinon sandbox which is cleaned after each test and do the same with nock. 

That way we don't forget to setup and teardown when writing tests.

test/unit/apps/adviser/repos.test.js
```diff
@@ -27,19 +30,21 @@ describe('Adviser repository', () => {
         const actual = await repos.getAdvisers(123)
 
         expect(actual).to.deep.equal(expected)
+        expect(this.nockScope.isDone()).to.be.true
```
teneightfive: Am leaning towards @ztolley's suggestion of having a separate it block for this test.

I know we don't always do it but it does read much nicer when each it block is doing just one thing.

feedmypixel:
> Am leaning towards @ztolley's suggestion of having a separate it block for this test.

@ztolley @teneightfive have refactored with this in mind.

test/unit/apps/adviser/repos.test.js
ztolley: Still need to remove this line

test/unit/apps/events/controllers/list.test.js
```diff
@@ -106,6 +108,7 @@ describe('Event list controller', () => {
         )
       )
       expect(this.resMock.render).to.have.been.calledWith('events/views/list', sinon.match.hasOwn('selectedFilters'))
+      expect(this.nockScope.isDone()).to.be.true
```
ztolley: Move to own test like the others?

test/unit/apps/events/controllers/list.test.js
```diff
@@ -122,6 +125,7 @@ describe('Event list controller', () => {
...
```
ztolley: Move to own test like the others?

test/unit/apps/events/controllers/list.test.js
```diff
@@ -123,5 +125,10 @@ describe('Event list controller', () => {
       )
       expect(this.resMock.render).to.have.been.calledWith('events/views/list', sinon.match.hasOwn('selectedFilters'))
     })
+
+    it('should render collection without region if non-UK country is selected', async () => {
+      await this.controller.renderEventList(this.reqMock, this.resMock, this.nextSpy)
+      expect(this.nockScope.isDone()).to.be.true
```
ztolley: Move to own test?

test/unit/global-helpers.js
```diff
@@ -0,0 +1,3 @@
+afterEach(() => {
+  nock.cleanAll()
```
teneightfive: Maybe a follow up could be to restore the sandbox here and create one in a global `beforeEach`?

ztolley: Neat! 🌮

feedmypixel: > Maybe a follow up could be to restore the sandbox here and create one in a global beforeEach?

yup totally - will drop that in after this

PR #1057, opened 2017-12-01
One list account owner editing

yeahfro:
This change introduces to ability to edit the One List account owner for a company.

## 1. Edit button
screen shot 2017-12-04 at 14 47 06

## 2. Edit account management
screen shot 2017-12-04 at 14 47 29

## 3. New value is set
screen shot 2017-12-04 at 14 47 44

### DIFF
--- a/src/apps/companies/macros.js
+++ b/src/apps/companies/macros.js
@@ -1,4 +1,8 @@
+const { assign } = require('lodash')
+
 const { globalFields } = require('../macros')
+const formLabels = require('./labels')
+const { transformObjectToOption } = require('../transformers')
 
 const companyFiltersFields = [
   {
@@ -45,7 +49,34 @@ const companySortForm = {
   ],
 }
 
+const accountManagementFormConfig = function ({
+  returnLink,
+  advisers = [],
+}) {
+  return {
+    returnLink,
+    buttonText: 'Save',
+    returnText: 'Cancel',
+    children: [
+      {
+        macroName: 'MultipleChoiceField',
+        name: 'one_list_account_owner',
+        initialOption: '-- Select account manager --',
+        optional: true,
+        options () {
+          return advisers.map(transformObjectToOption)
+        },
+      },
+    ].map(field => {
+      return assign(field, {
+        label: formLabels.accountManagementDisplayLabels[field.name],
+      })
+    }),
+  }
+}
+
 module.exports = {
   companySortForm,
   companyFiltersFields,
+  accountManagementFormConfig,
 }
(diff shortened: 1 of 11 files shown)


src/apps/companies/middleware/account-management.js
```diff
@@ -12,6 +15,23 @@ async function populateAccountManagementForm (req, res, next) {
   }
 }
 
+async function postAccountManagementDetails (req, res, next) {
+  const data = assign({}, res.locals.company, {
```
ztolley: Do you need to do this? Can you just send the changed field and it updates just them?

src/apps/companies/middleware/account-management.js
```diff
@@ -12,6 +15,23 @@ async function populateAccountManagementForm (req, res, next) {
...
+  })
+
+  try {
+    const result = await saveCompany(req.session.token, data)
+
+    req.flash('success', 'Account management details updated')
+
+    res.redirect(`/companies/${result.id}`)
+  } catch (error) {
+    next(error)
```
ztolley: Will this trigger an error screen instead of a form with the field marked with an error?  I guess as it's a purely optional field it will be ok

src/apps/companies/router.js
```diff
@@ -88,6 +88,7 @@ router
 router
   .route('/:companyId/account-management/edit')
   .get(populateAccountManagementForm, renderAccountManagementEditPage)
+  .post(populateAccountManagementForm, postAccountManagementDetails, renderAccountManagementEditPage)
```
ztolley: By using next(error) it won't reach the render page, will it?

src/apps/companies/macros.js
```diff
@@ -45,7 +49,34 @@ const companySortForm = {
   ],
 }
 
+const accountManagementFormConfig = function ({
+  returnLink,
+  advisers = [],
+}) {
+  return {
+    returnLink,
+    buttonText: 'Save',
```
teneightfive: I think this is the default value so maybe don't need to declare it?

src/apps/companies/macros.js
```diff
@@ -45,7 +49,34 @@ const companySortForm = {
...
+    returnText: 'Cancel',
```
teneightfive: We should really tidy up our use of `returnText`. I've seen _back_, _cancel_, _return_ and _return without saving_.

src/apps/companies/middleware/account-management.js
teneightfive: Was also thinking that. Not sure if we need to send the whole company object.

src/apps/companies/router.js
teneightfive: Yes, maybe we pass through the post controller but set the error message on `res.locals`. Or even check the format our form errors need to appear and place the message in there.

src/apps/companies/middleware/account-management.js
teneightfive: A possible alternative we just discussed:

```js
} catch (error) {
  logger.error(error)
    errors: {
      summary: 'Please try submitting this form again.',
    }
  })
  next()
}
```

PR #1058, opened 2017-12-04
Print styles

teneightfive:
This starts to add support for base level print styles.

Previously it would render an un-styled page. This adds the ability to wrap particular component styles within a print specific block to change their appearance. 

May need to tweak particular components as we go but this provides a base level to start from.


ztolley:
Needs a description

assets/stylesheets/components/_global-nav.scss
```diff
@@ -6,9 +6,13 @@
   padding: $default-spacing-unit / 4 0;
   position: relative;
 
-  @include media(tablet) {
+  @include media-query(tablet) {
     padding: 0;
   }
+
+  @if $is-print {
```
tyom: Seems like a roundabout way of doing
```css
@media print {
  display: none;
}
```

teneightfive: Back from the dead!

ztolley: I'd say keep it simple and not worry about IE < 8 when only 6 people have used ie 7 and the fallback for them is still functional

assets/stylesheets/trumps/_utility.scss
```diff
@@ -35,15 +35,11 @@
...
-    float: left !important;
-  }
+  float: left !important;
 }
 
 .u-float-right {
-  @include media(tablet) {
-    float: right !important;
-  }
+  float: right !important;
```
ztolley: doesn't this screw-up phone?

teneightfive: Have a look at the [commit description](https://github.com/uktrade/data-hub-frontend/pull/1058/commits/c34194c5974dbc0a975164361c7afa4ac64b3b2c)

ztolley: I did, but it doesn't say why. What happens on mobile? (I know our mobile needs work but just asking what the affect is)

teneightfive: This change applies the style regardless of screen size. Previously it would have displayed block and not been floated either way.

May actually need to revisit. It should display block on mobile in the place its being used but float on print. Previously the print style was showing the mobile version...back to the drawing board 🙄

teneightfive: So having looked at it. The only place is on the payment receipt. I think rather than re-write the media query to support print override I'm going to remove this media query from here.

The print receipt may need some further work to support tables on mobile so can take a better look at that separately. But for now we need to be able to apply some styles to the print view on desktop for one user group.

PR #1063, opened 2017-12-06
Filter inactive advisers in investment project

ztolley:
Updated the forms in the investment project to filter out inactive advisers.

### DIFF
--- a/src/apps/investment-projects/middleware/forms/details.js
+++ b/src/apps/investment-projects/middleware/forms/details.js
@@ -4,6 +4,7 @@ const { transformToApi, transformFromApi } = require('../../services/formatting'
 const { isValidGuid } = require('../../../../lib/controller-utils')
 const metadata = require('../../../../lib/metadata')
 const { getAdvisers } = require('../../../adviser/repos')
+const { filterActiveAdvisers } = require('../../../adviser/filters')
 const {
   buildMetaDataObj,
   transformObjectToOption,
@@ -25,15 +26,14 @@ async function populateForm (req, res, next) {
   }
 
   try {
-    const advisersResponse = await getAdvisers(req.session.token)
+    const investmentData = transformFromApi(res.locals.investmentData)
     const {
       equityCompany,
       equityCompanyInvestment,
     } = await getEquityCompanyDetails(req.session.token, equityCompanyId)
 
-    const contacts = equityCompany.contacts.map(transformContactToOption)
-    const advisers = advisersResponse.results.map(transformObjectToOption)
-    const investmentData = transformFromApi(res.locals.investmentData)
+    const contacts = get(equityCompany, 'contacts', []).map(transformContactToOption)
+
     const investmentTypes = metadata.investmentTypeOptions.map(transformObjectToOption).filter((investmentType) => {
       return equityCompany.uk_based || investmentType.label.toLowerCase().includes('fdi')
     })
@@ -44,6 +44,18 @@ async function populateForm (req, res, next) {
       business_activities: [''],
     }, investmentData)
 
+    const advisersResponse = await getAdvisers(req.session.token)
+
+    const clientRelationshipManagers = filterActiveAdvisers({
+      advisers: advisersResponse.results,
+      includeAdviser: get(investmentData, 'client_relationship_manager'),
+    }).map(transformObjectToOption)
+
+    const referralSourceAdvisers = filterActiveAdvisers({
+      advisers: advisersResponse.results,
+      includeAdviser: get(investmentData, 'referral_source_adviser'),
+    }).map(transformObjectToOption)
+
     res.locals.clientCompanyId = clientCompanyId || equityCompanyId
     res.locals.equityCompany = equityCompany
     res.locals.equityCompanyInvestment = equityCompanyInvestment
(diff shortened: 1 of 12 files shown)


src/apps/investment-projects/middleware/forms/details.js
```diff
@@ -25,50 +26,62 @@ async function populateForm (req, res, next) {
...
       equityCompanyInvestment,
     } = await getEquityCompanyDetails(req.session.token, equityCompanyId)
 
-    const contacts = equityCompany.contacts.map(transformContactToOption)
-    const advisers = advisersResponse.results.map(transformObjectToOption)
-    const investmentData = transformFromApi(res.locals.investmentData)
-    const investmentTypes = metadata.investmentTypeOptions.map(transformObjectToOption).filter((investmentType) => {
+    const contacts = get(equityCompany, 'contacts', []).map(transformContactToOption)
+
+    const investmentTypes = flatten(metadata.investmentTypeOptions).map(transformObjectToOption).filter((investmentType) => {
```
teneightfive: What's the purpose of the `flatten` here?

ztolley: Just to make things a bit more robust to handle when the value isn't there, it made testing easier rather than having to create a fake fully populated metadata object

teneightfive: Does that actually work? Would it not be better to wrap it in another array to ensure the value is an array.

Not sure I like this approach though.

ztolley: Open to suggestions. The reason it went in was I didn't want to have to create a collection of fake metadata, but maybe I need to. It is Purely there for testing

teneightfive: >It is Purely there for testing

There's your answer.

No different to what we've done elsewhere. If a method expects things, provide them. Better than unnecessarily overcomplicating the code.

ztolley: Yup, painful, but yup.

src/apps/investment-projects/middleware/forms/details.js
```diff
@@ -44,14 +44,27 @@ async function populateForm (req, res, next) {
...
+
     res.locals.clientCompanyId = clientCompanyId || equityCompanyId
     res.locals.equityCompany = equityCompany
     res.locals.equityCompanyInvestment = equityCompanyInvestment
 
     res.locals.form = assign({}, res.locals.form, {
       state,
       options: {
-        advisers,
+        clientRelationshipManagers: clientRelationshipManagerOptions,
```
teneightfive: Could you rename `clientRelationshipManagerOptions` to just `clientRelationshipManager`? Then you can just use shorthand here?

Same for `referralSourceAdvisers `?

src/apps/investment-projects/middleware/forms/details.js
```diff
@@ -63,14 +76,14 @@ async function populateForm (req, res, next) {
         primarySectors: metadata.sectorOptions.map(transformObjectToOption),
         businessActivities: metadata.businessActivityOptions.map(transformObjectToOption),
         investmentSpecificProgramme: metadata.investmentSpecificProgrammeOptions
-          .map(transformObjectToOption)
-          .filter(filterDisabledOption(state.specific_programme)),
+          .filter(filterDisabledOption(state.specific_programme))
```
teneightfive: Good spot, better to filter before mapping everything 👍 

Although it doesn't look for the transformed format to filter on does it?

src/apps/investment-projects/middleware/forms/details.js
```diff
@@ -93,6 +106,7 @@ async function populateForm (req, res, next) {
 
     next()
   } catch (error) {
+    console.log(error)
```
teneightfive: 🚨

src/apps/investment-projects/middleware/forms/project-management.js
```diff
@@ -1,24 +1,37 @@
...
       labels: projectManagementLabels.edit,
       state: {
-        project_manager: get(investmentData, 'project_manager.id'),
-        project_assurance_adviser: get(investmentData, 'project_assurance_adviser.id'),
+        project_manager: projectManager,
+        project_assurance_adviser: projectAssuranceAdviser,
       },
       options: {
-        advisers,
+        projectManagers: projectManagerOptions,
```
teneightfive: Same as other file here. Can we use this property name as the variable name and use shorthand here.

src/apps/investment-projects/middleware/forms/details.js
ztolley: That would be confusing, it would read like the field value and not the option list. The only reason it's called clientRelationshipManager is because it's below options.

src/apps/investment-projects/middleware/forms/details.js
ztolley: It was a bug so nothing got filtered out. It does not expect the transformed format.

src/apps/investment-projects/middleware/forms/team-members.js
```diff
@@ -16,31 +17,32 @@ function transformDataToTeamMemberArray (body) {
   return teamMembers.filter(member => member.adviser.length)
 }
 
+function getTeamMember ({ teamMember, advisers }) {
```
teneightfive: This doesn't quite feel like the right name. 

This function seems to return an adviser with a list of options not just a team member...

ztolley: It returns a team member, which is a combination of adviser and role... you are saying an advisor played a role in something. The options are for the drop down to pick which advisor you wish to assign a role.

ztolley: getTeamMemberSection ?

PR #1064, opened 2017-12-06
Deprecated tidy up

teneightfive:
This is some work around tidying up and removing some of the legacy styles so that we have more control of them.

It reduces the use of govuk-elements by quite a bit as well as removing some unused styles from trade elements.


assets/stylesheets/components/_button.scss
```diff
@@ -1,4 +1,38 @@
 @import "~govuk_frontend_toolkit/stylesheets/design-patterns/buttons";
```
ztolley: If moving the styles into this project, why not the mixin?

assets/stylesheets/components/_button.scss
```diff
@@ -25,28 +24,19 @@
...
 // modifiers
 .button--secondary {
   @include button ($grey-2);
-
-  &:link,
-  &:visited {
-    background-color: $grey-2;
-    color: $black;
-  }
+  padding-bottom: .526315em;
```
ztolley: Where did this number come from?

src/apps/companies/views/edit.njk
```diff
@@ -18,10 +18,11 @@
   </div>
 
   {% if chDetails %}
-    <div class="section">
-      {% component 'key-value-table', variant='readonly', items=chDetails %}
-      <div class="table_footnote">Data from <strong>Companies House</strong></div>
-    </div>
+    {# TODO Do we even still need this? It's not currently being displayed (Bug?) #}
```
ztolley: I would say yes and we should fix that bug

assets/stylesheets/_deprecated/trade-elements.scss
```diff
@@ -1,23 +1,21 @@
 $gt-big-phone: ">414px";
 $asset-path: "../" !default;
 
+@import "../tools/mixins";
```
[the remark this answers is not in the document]
teneightfive: Not sure what you mean? They're currently in the deprecated folder.

Need to revisit some more of this later but means replacing some components we use in lots of places like the key/value table.

assets/stylesheets/components/_button.scss
teneightfive: It's now included only in the file it's needed.

Rather than including things just at a global level and assuming they're always there it's better to declare file dependencies within the file. That way it's easier to understand what else it requires if you move/remove it/use it for a different project.

assets/stylesheets/_deprecated/trade-elements.scss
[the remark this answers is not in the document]
ztolley: Just questioning why update something marked as deprecated, and wonder what is replacing it and should we update that instead... I know this probably doesn't make sense. I'm just wondering what replaces trade elements

assets/stylesheets/components/_button.scss
teneightfive: It's the number currently used in the frontend toolkit button mixin. The GOV.UK ones use the Transport font family which had a different line-height and it was compensated for by the padding. 

With our font-family this meant that button was not actually aligned in the middle of the button. This is a small fix to address that.

We may want to look at replacing that mixin when we remove the toolkit but we need to a do a few more things before we can do that.

Was trying to take small steps although this has already grown quite large :/

src/apps/companies/views/edit.njk
teneightfive: Separate PR?

assets/stylesheets/_deprecated/trade-elements.scss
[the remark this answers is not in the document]
ztolley: maybe we add them to the top level import file instead of trade elements: after all the idea is to get rid of the trade elements dependency.

And when do we talk Trade Elements 2.0 :) to be used across MI etc?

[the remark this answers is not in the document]
teneightfive: >Just questioning why update something marked as deprecated

This change is just to change the order of these imports.

>maybe we add them to the top level import file instead of trade elements

As mentioned in another comment, some of the files in trade elements currently rely on these so we import them here.

We want to move away from the pattern of define settings/mixins once at the top and more to nested dependencies within each component.

If we move them to the current top level import we break that pattern.

>And when do we talk Trade Elements 2.0 :) to be used across MI etc?

We can start moving there soon. That's what this work is. The patterns we have within this repo will be the base for new components.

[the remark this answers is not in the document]
ztolley: We also need to look at the issue with stack overflows then if we want to move away from top level imports as the current problem will only get worse.

PR #1066, opened 2017-12-07
Add job title label to contact card

yeahfro:
Change adds label for job title with other meta data when contacts are displayed in lists.

screen shot 2017-12-07 at 14 36 18

### DIFF
--- a/src/apps/contacts/transformers.js
+++ b/src/apps/contacts/transformers.js
@@ -1,9 +1,9 @@
 /* eslint-disable camelcase */
-const { get, pickBy, findIndex } = require('lodash')
+const { get, pickBy, compact, assign } = require('lodash')
 
 const { getFormattedAddress } = require('../../lib/address')
 const { getDataLabels } = require('../../lib/controller-utils')
-const { contactDetailsLabels } = require('./labels')
+const { contactDetailsLabels, contactMetaItemLabels } = require('./labels')
 
 function getContactAddress (address_same_as_company, contactAddressFields, company) {
   if (!address_same_as_company) {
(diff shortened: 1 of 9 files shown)


src/apps/transformers.js
```diff
@@ -97,6 +105,14 @@ function transformApiResponseToCollection (options = {}, ...itemTransformers) {
   }
 }
 
+const transformToMetaElement = (value, label, type, badgeModifier, additionalFields = {}) => {
```
ztolley: I see that transformToMetaElement gives us a tidier syntax because of how you've combined it with `compact`. What I don't get is what the additional fields param is there for.

Also, how about switching to use named params?

`({value, label, type, ... })`

src/apps/contacts/transformers.js
```diff
@@ -23,74 +24,30 @@ function transformContactToListItem ({
...
+  const meta = compact([
+    transformToMetaElement(job_title, 'Job title'),
+    transformToMetaElement(get(company, 'name'), 'Company'),
+    transformToMetaElement(get(company_sector, 'name'), 'Sector'),
+    transformToMetaElement(get(address_country, 'name'), 'Country'),
+    transformToMetaElement(company_uk_region, 'UK region'),
+    transformToMetaElement(modified_on, 'Updated on', 'datetime'),
+    transformToMetaElement(primary ? 'Primary' : null, 'Contact type', 'badge', 'secondary'),
+    transformToMetaElement(archived_on ? 'Archived' : null, 'Status', 'badge'),
+  ])
```
ztolley: how about we get the labels from a labels file?

src/apps/investment-projects/transformers.js
```diff
@@ -12,48 +13,13 @@ function transformInvestmentProjectToListItem ({
...
-      name: 'sector',
-      label: 'Sector',
-      value: sector,
-    })
-  }
+  const meta = compact([
+    transformToMetaElement(stage, 'Stage', 'badge'),
+    transformToMetaElement(investment_type, 'Investment type', 'badge', 'secondary'),
+    transformToMetaElement(investor_company, 'Investor'),
+    transformToMetaElement(estimated_land_date, 'Land date', 'dateMonthYear', null, { isInert: true }),
```
ztolley: Ah that's what the other fields param was for

test/unit/apps/investment-projects/transformers.test.js
```diff
@@ -19,17 +19,14 @@ describe('Investment project data transformers', () => {
       expect(firstItem.code).to.a('string')
       expect(firstItem.meta).to.an('array').that.have.length(5)
 
-      expect(firstItem.meta[0].name).to.equal('stage')
```
ztolley: ??

yeahfro: We are not using `name` for any of the other badges so this seemed like it was unnecessary

src/apps/contacts/transformers.js
feedmypixel: Feels like we have the established `labels.js` pattern. I am up for considering different approaches but what you have here is labels in a transformer file, feels like it would be good to keep the abstraction of labels in a different file.

src/apps/investment-projects/transformers.js
```diff
@@ -12,48 +13,13 @@ function transformInvestmentProjectToListItem ({
...
-  }
-  if (sector) {
-    meta.push({
-      name: 'sector',
-      label: 'Sector',
-      value: sector,
-    })
-  }
+  const meta = compact([
+    transformToMetaElement(stage, 'Stage', 'badge'),
```
feedmypixel: I'm not totally convinced that this abstraction aids communication to the developer. 
Showing that something won't appear if it doesn't exist is now wrapped up in `compact` rather than previously it being shown by: `if (estimated_land_date) {`
The `compact` is quite nice so maybe have a play, and the compact may read better if you also do the following`:

I wonder if `transformToMetaElement` is an actual transformation? Maybe:
- `buildMetaItem` (pretty much what `transformToMetaElement` does now)
- `buildMetaBadge` (curry `transformToMetaElement` with `type: 'badge'`)
- `buildMetaDate` (curry `transformToMetaElement` with `type: 'dateMonthYear'` as default)

src/apps/transformers.js
feedmypixel: I would also suggest named params, it can communicate when looking at the call to this function, what the params are for.

src/apps/investment-projects/transformers.js
feedmypixel: Also if you use named params for the call to a method then you can do:
`buildMetaItem({value: sector, label: 'Sector'})`

ztolley: It's more of a factory than a transformer

feedmypixel:
As a side note I like the code but its on the edge of KISS, a big part of writing code is readability. Not only for others but for your future self.

PR #1068, opened 2017-12-08
Filter disabled on date

ztolley:
Improves the disabled on filter so that rather than simply test for the presence of a value it filters items marked as disabled after a given date or the current date.

By using the filter in this manner the front end can build a list of items that were enabled at a given point in time, such as the creation date and time of a record being created.

### DIFF
--- a/src/apps/filters.js
+++ b/src/apps/filters.js
@@ -1,12 +1,15 @@
-/**
- *
- * @param {string} [currentValue] - The current value for a field so that the options filtered include the current
- *                                  even if invalid
- * @returns {function} - a filter function that will filter out inactive or non-current values
- */
-function filterDisabledOption (currentValue = null) {
+function filterDisabledOption ({ currentValue = null, createdOn } = {}) {
+  const createdOnTime = Date.parse(createdOn) || Date.now()
+
   return function (item) {
-    return !item.disabled_on || item.id === currentValue
+    if (!item.disabled_on) {
+      return true
+    }
+
+    const isDisabled = Date.parse(item.disabled_on) > createdOnTime
+    const isCurrentValue = item.id === currentValue
+
+    return (isDisabled || isCurrentValue)
   }
 }
 
--- a/src/apps/investment-projects/middleware/forms/details.js
+++ b/src/apps/investment-projects/middleware/forms/details.js
@@ -27,6 +27,8 @@ async function populateForm (req, res, next) {
 
   try {
     const investmentData = transformFromApi(res.locals.investmentData)
+    const createdOn = get(investmentData, 'created_on')
+
     const {
       equityCompany,
       equityCompanyInvestment,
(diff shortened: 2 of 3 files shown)


src/apps/filters.js
```diff
@@ -1,12 +1,14 @@
...
+function filterDisabledOption ({ currentValue = null, createdOn } = {}) {
+  const createdOnDateTime = createdOn ? moment(createdOn, moment.ISO_8601) : moment()
+
   return function (item) {
-    return !item.disabled_on || item.id === currentValue
+    if (!item.disabled_on) {
+      return true
+    }
+
+    return moment(item.disabled_on, moment.ISO_8601).isAfter(createdOnDateTime) || item.id === currentValue
```
teneightfive: Maybe wrap brackets around the second condition to make it easier to read?

feedmypixel: could this be put in a variable to help with communication? what is it that its doing? 
1) has been disabled after its creation
2) is the current value?

```
const disabledOrCurrentSelection = (moment(item.disabled_on, moment.ISO_8601).isAfter(createdOnDateTime) || item.id === currentValue)

return disabledOrCurrentSelection
```

Kind of feels like this is doing too much?

maybe:

```
const isDisabled = moment(item.disabled_on, moment.ISO_8601).isAfter(createdOnDateTime) 
const isCurrentValue = item.id === currentValue

if (isDisabled || isCurrentValue) {
   return true
}
```

src/apps/filters.js
```diff
@@ -1,12 +1,14 @@
...
- *
- * @param {string} [currentValue] - The current value for a field so that the options filtered include the current
- *                                  even if invalid
- * @returns {function} - a filter function that will filter out inactive or non-current values
- */
-function filterDisabledOption (currentValue = null) {
+const moment = require('moment')
+
+function filterDisabledOption ({ currentValue = null, createdOn } = {}) {
+  const createdOnDateTime = createdOn ? moment(createdOn, moment.ISO_8601) : moment()
```
feedmypixel: Are we preferring `moment` over `date-fns`? 
You could have could just use `createdOn = Date.now()` as a default argument 
Then `date-fns` and the default in `const createdOnDateTime = format(createdOn)`

src/apps/filters.js
```diff
@@ -1,12 +1,14 @@
...
+
   return function (item) {
-    return !item.disabled_on || item.id === currentValue
+    if (!item.disabled_on) {
+      return true
+    }
+
+    return moment(item.disabled_on, moment.ISO_8601).isAfter(createdOnDateTime) || (item.id === currentValue)
```
feedmypixel: could this be put in a variable to help with communication? what is it that its doing? 
1) has been disabled after its creation
2) is the current value?

```
const disabledOrCurrentSelection = (moment(item.disabled_on, moment.ISO_8601).isAfter(createdOnDateTime) || item.id === currentValue)

return disabledOrCurrentSelection
```

Kind of feels like this is doing too much?

maybe:

```
const isDisabled = moment(item.disabled_on, moment.ISO_8601).isAfter(createdOnDateTime) 
const isCurrentValue = item.id === currentValue

if (isDisabled || isCurrentValue) {
   return true
}
```

src/apps/filters.js
ztolley: I asked when looking at this. date-fns was picked for client side stuff because it's more modular and lightweight compared to moment due to size of locale files.  For server stuff moment feels simpler and easier to read.

Both solutions are equally valid, I went with this one.

src/apps/filters.js
ztolley: If that makes it easier to understand I'll do that.

src/apps/filters.js
ztolley: Also the whole 'isAfter' reads nice

feedmypixel: > date-fns was picked for client side stuff because it's more modular and lightweight compared to moment due to size of locale files. For server stuff moment feels simpler and easier to read.

Oh ok, don't think I was here for that. I have always used `date-fns` here. Would be nice not to have 2 things doing the same thing.

feedmypixel: > Also the whole 'isAfter' reads nice

Agreed. `date-fns` also has `isAfter` ;)

ztolley: Coverted to raw Date, but keeping moment in tests for readability

PR #1072, opened 2017-12-11
Smarter event filter for interaction edit

ztolley:
The current version of service delivery editing simply filters out any events that are disabled, but this can cause issues when editing a service delivery after it's associated event has been disabled.

The improved version uses a new event filter that allows the API server to filter events based on when they were disabled. The form options function now asks for events enabled at the time the service delivery was created (or today).

### DIFF
--- a/src/apps/events/repos.js
+++ b/src/apps/events/repos.js
@@ -29,18 +29,15 @@ function getAllEvents (token) {
   return authorisedRequest(token, `${config.apiRoot}/v3/event?limit=100000&offset=0`)
 }
 
-/**
- *
- * @param {string} token Session token to use for API requests
- *
- * @returns {promise[Array]} Returns an array of event objects that have not been disabled or are the current event
- */
-async function getActiveEvents (token) {
+async function getActiveEvents (token, createdOn) {
   const eventsResponse = await search({
     searchEntity: 'event',
     requestBody: {
       sortby: 'name:asc',
-      disabled_on_exists: false,
+      disabled_on: {
+        after: createdOn || (new Date()).toISOString(),
+        exists: false,
+      },
     },
     token: token,
     limit: 100000,
--- a/src/apps/interactions/middleware/details.js
+++ b/src/apps/interactions/middleware/details.js
@@ -66,6 +66,7 @@ async function getInteractionDetails (req, res, next, interactionId) {
 
 async function getInteractionOptions (req, res, next) {
   try {
+    const interaction = res.locals.interaction
     const token = req.session.token
     const currentAdviser = get(res.locals, 'interaction.dit_adviser.id')
     const company = get(res.locals, 'company.id')
@@ -78,7 +79,7 @@ async function getInteractionOptions (req, res, next) {
     })
 
     if (req.params.kind === 'service-delivery') {
-      res.locals.events = await getActiveEvents(token)
+      res.locals.events = await getActiveEvents(token, interaction.created_on)
     }
 
     next()
--- a/src/apps/interactions/middleware/details.js
+++ b/src/apps/interactions/middleware/details.js
@@ -79,7 +79,7 @@ async function getInteractionOptions (req, res, next) {
     })
 
     if (req.params.kind === 'service-delivery') {
-      res.locals.events = await getActiveEvents(token, interaction.created_on)
+      res.locals.events = await getActiveEvents(token, get(interaction, 'created_on'))
     }
 
     next()
--- a/test/unit/apps/events/repos.test.js
+++ b/test/unit/apps/events/repos.test.js
@@ -82,19 +82,55 @@ describe('Event repos', () => {
           .reply(200, this.eventCollection)
       })
 
-      context('and no current event', () => {
+      context('and when asked for all active events', () => {
         beforeEach(async () => {
-          this.events = await this.repos.getActiveEvents('1234')
+          const now = new Date()
+          this.clock = sinon.useFakeTimers(now.getTime())
+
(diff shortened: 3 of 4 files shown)


src/apps/events/repos.js
```diff
@@ -29,18 +29,18 @@ function getAllEvents (token) {
...
 }
 
-/**
- *
- * @param {string} token Session token to use for API requests
- *
- * @returns {promise[Array]} Returns an array of event objects that have not been disabled or are the current event
- */
-async function getActiveEvents (token) {
+async function getActiveEvents ({ token, createdOn }) {
```
teneightfive: Not sure about moving the token into a named param. I think its okay to keep the token as the first argument but potentially use object as arguments for other options.

ztolley: What is so special about token that merits it not being in the main set or parameters?

ztolley: @feedmypixel Can you give us a casting vote?

function `function fred ({ token, id, type, date })` or `function fred (token, { id, type, date}) `

My view is that there is nothing 'special' about token, it's just another parameter so why not be consistent and put them all in the {}

ztolley: its just one option so would be overkill, so (token, createdOn)

src/apps/interactions/middleware/details.js
```diff
@@ -66,6 +66,7 @@ async function getInteractionDetails (req, res, next, interactionId) {
 
 async function getInteractionOptions (req, res, next) {
   try {
+    const interaction = res.locals.interaction
```
feedmypixel: does this needs a `get()`

src/apps/interactions/middleware/details.js
```diff
@@ -78,7 +79,7 @@ async function getInteractionOptions (req, res, next) {
     })
 
     if (req.params.kind === 'service-delivery') {
-      res.locals.events = await getActiveEvents(token)
+      res.locals.events = await getActiveEvents(token, get(interaction, 'created_on'))
```
feedmypixel: this would read better as:
```
const createdOn = get(interaction, 'created_on')
res.locals.events = await getActiveEvents(token, { createdOn })
```

src/apps/events/repos.js
```diff
@@ -29,18 +29,18 @@ function getAllEvents (token) {
...
-/**
- *
- * @param {string} token Session token to use for API requests
- *
- * @returns {promise[Array]} Returns an array of event objects that have not been disabled or are the current event
- */
-async function getActiveEvents (token) {
+async function getActiveEvents (token, createdOn) {
+  const now = new Date()
+  const after = createdOn || now.toISOString()
```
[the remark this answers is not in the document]
feedmypixel: you can also just use `format()` form `date-fns`

src/apps/interactions/middleware/details.js
ztolley: it wasn't, then @teneightfive suggested make it get... I think we overuse get.

src/apps/events/repos.js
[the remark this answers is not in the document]
ztolley: There's no need to use `date-fns` as javascript dates work just fine for this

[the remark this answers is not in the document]
ztolley: As for the name after, we are filters things After a date, plus the name of the param in the api query is after.

src/apps/interactions/middleware/details.js
ztolley: Seemed overkill to use {} syntax for params when there are just two, and one is required.

src/apps/interactions/middleware/details.js
ztolley: Getting the adviser does as for some reason we had duff data that had no adviser associated with it

src/apps/interactions/middleware/details.js
ztolley: Now it's `getActiveEvents(token, interaction.created_on)

src/apps/interactions/middleware/details.js
teneightfive: I tend to use it for deep nested properties. In this case we might be okay without it.

ztolley: I removed it for the created on but need to leave it for the adviser to guard against bad data

feedmypixel: personally any level deeper than 1 on an object I use `get` for as you will get an error if the parent `locals` doesn't exist.

src/apps/events/repos.js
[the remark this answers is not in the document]
feedmypixel: > There's no need to use date-fns as javascript dates work just fine for this

I think this got confused in the PR the other day. We now have 3 different ways of doing dates across the codebase.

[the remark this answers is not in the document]
feedmypixel: > As for the name after, we are filters things After a date, plus the name of the param in the api query is after.

What I meant is `(new Date).toISOString()` doesn't mean `after` in my mind.

src/apps/interactions/middleware/details.js
ztolley: It's safe to assume res.locals.interaction exists, and if it didn't then there is a bigger problem in this middleware

src/apps/events/repos.js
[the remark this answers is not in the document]
ztolley: The idea is you are telling the query return results `after` the `created` or `current` date 

I moved this straight into the query anyway now which seems obvious to me

PR #1073, opened 2017-12-11
Add support for displaying and filtering UK region on orders

teneightfive:
This adds the display of the UK region associated with an order to
the order overview and support for filtering on the region to the
collection page.

### Order overview
localhost_3001_omis_81fe292f-8427-4cce-a056-0c6191a2057c_work-order

### Listing page

### DIFF
--- a/src/apps/omis/apps/create/views/summary.njk
+++ b/src/apps/omis/apps/create/views/summary.njk
@@ -43,7 +43,13 @@
 
     <h2 class="heading-medium">What happens next</h2>
 
-    <p>Continuing with the order will notify the post manager for {{ values.primary_market.name }}.</p>
+    <p>
+      Continuing with the order will notify the post manager for {{ values.primary_market.name }}
+      {%- if company.uk_region -%}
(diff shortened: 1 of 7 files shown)


src/apps/omis/apps/create/views/summary.njk
```diff
@@ -43,7 +43,13 @@
 
     <h2 class="heading-medium">What happens next</h2>
 
-    <p>Continuing with the order will notify the post manager for {{ values.primary_market.name }}.</p>
+    <p>
+      Continuing with the order will notify the post manager for {{ values.primary_market.name }}
+      {%- if company.uk_region -%}
+        &nbsp;and the region manager for {{ company.uk_region.name }}
```
feedmypixel: bit of classic html work there 👍  haven't seen a non breaking space for a while 😂 💥

teneightfive: Haha, I know!! Didn't like doing it but needed to remove the whitespace using nunjucks but in this instance enforce one when this condition matches.

PR #1074, opened 2017-12-12
Remove deprecated scripts

teneightfive:
This change removes unused scripts that were in the deprecated folder and rename files being
used from deprecated to something more appropriate.


teneightfive:
Going to leave the codeclimate issues for now as this change was just a move of the file rather than changing its functionality so this is already a known issue rather than introducing a new one.

PR #1077, opened 2017-12-12
Feature/filter company reference data

ztolley:
Refactor of company edit middleware and the introduction of a new options fetcher that will replace metadata.

The refactoring of the middleware:
- Simplifies the code and makes it easier to read and follow
- Introduces filtered options
- Fixes a bug saving UK addresses in some cases

The change also allows the middleware order to change so that form population and option fetching only happens if it needs to, instead of being called and ignored.

### DIFF
--- a/src/apps/companies/controllers/edit.js
+++ b/src/apps/companies/controllers/edit.js
@@ -1,4 +1,4 @@
-const { assign, find, get } = require('lodash')
+const { assign, find, get, isEmpty } = require('lodash')
 
 const { transformCompaniesHouseToView } = require('../transformers')
 
@@ -44,14 +44,17 @@ function renderForm (req, res) {
     })
   }
   const businessTypeLabel = getBusinessTypeLabel(
-    res.locals.companiesHouseCategory, isForeign, res.locals.businessType
+    res.locals.companiesHouseCategory, isForeign, get(res.locals, 'formData.business_type')
   )
+
+  const showTradingAddress = !isEmpty(get(res.locals, 'formData.trading_address_1'))
 
   res
     .breadcrumb(pageTitle)
     .render('companies/views/edit', {
       isForeign,
       businessTypeLabel,
+      showTradingAddress,
     })
 }
 
--- a/src/apps/companies/middleware/form.js
+++ b/src/apps/companies/middleware/form.js
@@ -1,75 +1,92 @@
 const { assign, find, get } = require('lodash')
 
-const metadataRepository = require('../../../lib/metadata')
+const { getOptions } = require('../../../lib/options')
 const { hqLabels } = require('../labels')
 const companyFormService = require('../services/form')
 const { transformCompanyToForm } = require('../transformers')
-const { transformObjectToOption } = require('../../transformers')
 
-function populateForm (req, res, next) {
-  const countryQueryParam = get(req.query, 'country')
-  const headquarterOptions = metadataRepository.headquarterOptions
-    .map(transformObjectToOption)
-    .map(option => {
-      return {
-        value: option.value,
-        label: hqLabels[option.label],
-      }
+function getHeadquarterOptions (token) {
+  return getOptions(token, 'headquarter-type')
+    .then((options) => {
+      const headquarterOptions = options.map(option => {
+        return {
+          value: option.value,
+          label: hqLabels[option.label],
+        }
+      })
+
+      headquarterOptions.unshift({
+        value: 'not_headquarters',
+        label: 'Not a headquarters',
+      })
+
+      return headquarterOptions
+    })
+}
+
+async function getCompanyFormOptions (token, formData, createdOn) {
+  return {
+    headquarters: await getHeadquarterOptions(token),
+    regions: await getOptions(token, 'uk-region', { createdOn, currentValue: formData.uk_region }),
+    sectors: await getOptions(token, 'sector', { createdOn, currentValue: formData.sector }),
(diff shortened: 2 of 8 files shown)


src/apps/companies/controllers/edit.js
```diff
@@ -44,14 +44,17 @@ function renderForm (req, res) {
     })
   }
   const businessTypeLabel = getBusinessTypeLabel(
-    res.locals.companiesHouseCategory, isForeign, res.locals.businessType
+    res.locals.companiesHouseCategory, isForeign, res.locals.formData.business_type
```
teneightfive: `get`?

src/apps/companies/middleware/form.js
```diff
@@ -1,75 +1,96 @@
...
-      return {
-        value: option.value,
-        label: hqLabels[option.label],
-      }
+
+function getHeadquarterOptions (token) {
+  return getOptions({ key: 'headquarter-type', token })
+    .then((options) => {
+      const headquarterOptions = options.map(option => {
+        return {
```
teneightfive: This looks like our transformers?

src/apps/companies/middleware/form.js
```diff
@@ -1,75 +1,96 @@
...
+    turnovers: await getOptions({ key: 'turnover', createdOn, currentValue: formData.turnover_range, token }),
+    countries: await getOptions({ key: 'country', createdOn, currentValue: formData.registered_address_country, token }),
+  }
+}
 
-  headquarterOptions.unshift({
-    value: 'not_headquarters',
-    label: 'Not a headquarters',
-  })
+// Generates form options and the form data for the user to edit based on
```
teneightfive: Do we need these comments? Might want to re-read and make it read better if we feel we do.

src/lib/options.js
```diff
@@ -0,0 +1,21 @@
+const config = require('../../config')
+const authorisedRequest = require('../lib/authorised-request')
+const { filterDisabledOption } = require('../apps/filters')
+const { transformObjectToOption } = require('../apps/transformers')
+
+async function getOptions ({ key, createdOn, currentValue, includeDisabled = false, token }) {
+  const url = `${config.apiRoot}/metadata/${key}/`
+  const allOptions = await authorisedRequest(token, url)
```
teneightfive: Try/catch around this?

src/lib/options.js
```diff
@@ -0,0 +1,21 @@
...
+
+  if (includeDisabled) {
+    return allOptions.map(transformObjectToOption)
```
teneightfive: Would be nice to only do this `map` once. We're doing it here and on line 16.

Maybe set the response to a variable and map the result then return just that var if disabled and then the same var filtered if not.

src/lib/options.js
```diff
@@ -0,0 +1,21 @@
...
+  }
+
+  return allOptions
+    .filter(filterDisabledOption({ currentValue, createdOn }))
+    .map(transformObjectToOption)
+}
+
+module.exports = {
+  getOptions,
```
teneightfive: Missing unit tests for this new module 😉

teneightfive:
I'm also not sure about `options`. Feels very vague and non-descriptive.

src/apps/companies/controllers/edit.js
ztolley: Do we want to use Get even when we know it's there?  I'll add it for safety but think there is such a thing as overuse

src/apps/companies/middleware/form.js
ztolley: I am most likely going to remove it, it was more to get ideas out of my head but now the code has been simplified enough so the code speaks for itself.. plus this is still WIP

src/lib/options.js
ztolley: The code still gets tests via testing the middleware as I have used Nock instead of Stubbing. SO the question is do we still have tests here which results in duplication?

ztolley:
Agreed that options can feel vague but couldn't think of a good name at the time. Starting to think that 'referenceData' is an alternative but open to suggestions.... Still working on this ticket.

src/apps/companies/middleware/form.js
ztolley: It's not. it's a little different in how it gets the labels.

ztolley:
Actually, I'm going back to like the name 'options' It's used to create a list of options that are used in the form macros for the 'options' property after all so it's consistent with that.

Dont get hung on on the < option > tag

src/lib/options.js
```diff
@@ -0,0 +1,23 @@
+const config = require('../../config')
+const authorisedRequest = require('../lib/authorised-request')
+const { filterDisabledOption } = require('../apps/filters')
+const { transformObjectToOption } = require('../apps/transformers')
+
+async function getOptions (token, key, { createdOn, currentValue, includeDisabled = false } = {}) {
```
teneightfive: Do you need `= {}`?

src/lib/options.js
```diff
@@ -0,0 +1,23 @@
...
+    const url = `${config.apiRoot}/metadata/${key}/`
+    let options = await authorisedRequest(token, url)
+
+    if (!includeDisabled) {
+      options = options.filter(filterDisabledOption({ currentValue, createdOn }))
+    }
+
+    return options.map(transformObjectToOption)
+  } catch (error) {
+    return []
```
teneightfive: Do we want to log this error?

ztolley: yeah probably good idea

src/lib/options.js
ztolley: yes

PR #1078, opened 2017-12-13
add tag manager tracking for no javascript page views

feedmypixel:
Add a variable to be read by Google Tag Manager so we can see how many `no JavaScript` view we are having


src/templates/_includes/google-tag-manager-no-script.njk
```diff
@@ -1,5 +1,8 @@
 {% if GOOGLE_TAG_MANAGER_KEY %}
 <noscript>
-  <iframe src="https://www.googletagmanager.com/ns.html?id={{ GOOGLE_TAG_MANAGER_KEY }}" height="0" width="0" style="display:none;visibility:hidden"></iframe>
+  <iframe src="https://www.googletagmanager.com/ns.html?id={{ GOOGLE_TAG_MANAGER_KEY }}&no_javascript=true"
+          height="0"
+          width="0"
+          style="display:none;visibility:hidden"></iframe>
```
teneightfive: The formatting of this seems strange? 

Maybe just keep it all on one line or drop the closing tag and line it to the same indentation to the opening tag?

PR #1079, opened 2017-12-13
Feature/permission based global nav, search aggregations and local nav

feedmypixel:
This work enables the Global Navigation, search aggregations and local nav to render items dependant on user permissions.

### TODO
- [x] Once the backend Interactions permissions from @reupen work is available re run AT
- [x] Once the backend Documents permissions from @elcct work is available re run AT
- [x] Work out what we are doing with permissions and Acceptance Tests as they are currently failing on this PR
> once the work from @reupen and @elcct is in and the [data-hub-leeloo/fixtures/metadata/teams.yaml](https://github.com/uktrade/data-hub-leeloo/blob/fa45085c422d1ca6598f21a93a6d70701a643263/fixtures/metadata/teams.yaml#L122-L268) `DIT_staff` role has had the new permissions added, the AT on this branch should pass

#### Global nav and search aggregations
lep-all

#### Contact page view
lep-contacts

#### Company page view
lep-company

#### Global nav and search aggregations
da-all

#### Contact page view
da-contacts

#### Company page view
da-company

#### Global nav and search aggregations
dit_user-all

#### Contact page view
dit_user-contacts

#### Company page view
dit_user-company

### Logged out user
screen shot 2017-12-13 at 12 57 27

### DIFF
--- a/src/apps/contacts/controllers/documents.js
+++ b/src/apps/contacts/controllers/documents.js
@@ -1,7 +1,11 @@
-const { get } = require('lodash')
+const { get, isUndefined } = require('lodash')
 
-async function renderDocuments (req, res, next) {
+function renderDocuments (req, res, next) {
   const archivedDocumentPath = get(res.locals, 'contact.archived_documents_url_path')
+
+  if (isUndefined(archivedDocumentPath)) {
+    return next({ statusCode: 403 })
+  }
 
   return res
     .breadcrumb('Documents')
--- a/src/apps/filters.js
+++ b/src/apps/filters.js
@@ -1,3 +1,5 @@
+const { get, intersection } = require('lodash')
+
 function filterDisabledOption ({ currentValue = null, createdOn } = {}) {
   const createdOnTime = Date.parse(createdOn) || Date.now()
 
@@ -13,6 +15,17 @@ function filterDisabledOption ({ currentValue = null, createdOn } = {}) {
   }
 }
 
+function filterNonPermittedItem (userPermissions) {
+  return function (item) {
+    const hasPermissions = get(item, 'permissions.length', 0)
+    if (!hasPermissions) { return true }
+
+    const isPermitted = intersection(item.permissions, userPermissions).length > 0
+    return isPermitted
+  }
+}
+
 module.exports = {
   filterDisabledOption,
+  filterNonPermittedItem,
 }
--- a/src/apps/middleware.js
+++ b/src/apps/middleware.js
@@ -1,5 +1,7 @@
-const { isEmpty } = require('lodash')
+const { get, isEmpty, assign } = require('lodash')
 const queryString = require('query-string')
+
+const { filterNonPermittedItem } = require('./filters')
 
 function setHomeBreadcrumb (name) {
   return function (req, res, next) {
@@ -15,13 +17,17 @@ function setHomeBreadcrumb (name) {
 
 function setLocalNav (items = []) {
   return function buildLocalNav (req, res, next) {
-    res.locals.localNavItems = items.map(item => {
-      const url = item.isExternal ? item.url : `${req.baseUrl}/${item.path}`
-      return Object.assign(item, {
-        url,
-        isActive: res.locals.CURRENT_PATH === url,
+    const userPermissions = get(res, 'locals.user.permissions')
+
+    res.locals.localNavItems = items
+      .filter(filterNonPermittedItem(userPermissions))
+      .map((item) => {
+        const url = item.isExternal ? item.url : `${req.baseUrl}/${item.path}`
+        return assign({}, item, {
+          url,
+          isActive: res.locals.CURRENT_PATH === url,
+        })
       })
-    })
     next()
   }
 }
--- a/src/apps/search/builders.js
+++ b/src/apps/search/builders.js
@@ -1,15 +1,18 @@
-const { isArray } = require('lodash')
+const { isArray, assign } = require('lodash')
 
-const { entities } = require('./services')
+const { ENTITIES } = require('./constants')
+const { filterNonPermittedItem } = require('../filters')
 
-function buildSearchAggregation (apiResponseEntities) {
-  if (!isArray(apiResponseEntities)) { return }
+function buildSearchAggregation ({ aggregations, userPermissions, entityDetails = ENTITIES }) {
+  if (!isArray(aggregations)) { return }
 
-  return entities.map((defaultEntity) => {
-    return Object.assign(
+  const userEntities = entityDetails.filter(filterNonPermittedItem(userPermissions))
+
+  return userEntities.map((defaultEntity) => {
+    return assign(
       {},
       defaultEntity,
-      apiResponseEntities.find((apiResponseEntity) => {
+      aggregations.find((apiResponseEntity) => {
         return apiResponseEntity.entity === defaultEntity.entity
       })
     )
--- a/src/middleware/locals.js
+++ b/src/middleware/locals.js
@@ -1,26 +1,22 @@
+const { get } = require('lodash')
+
+const { GLOBAL_NAV_ITEMS } = require('../apps/constants')
 const logger = require('../../config/logger')
 const config = require('../../config')
+const { filterNonPermittedItem } = require('../apps/filters')
 
 let webpackManifest = {}
+
 try {
   webpackManifest = require(`${config.buildDir}/manifest.json`)
 } catch (err) {
   logger.error('Manifest file is not found. Ensure assets are built.')
 }
 
-const globalNavItems = [
-  { path: '/companies', label: 'Companies' },
-  { path: '/contacts', label: 'Contacts' },
-  { path: '/events', label: 'Events' },
-  { path: '/interactions', label: 'Interactions and services' },
-  { path: '/investment-projects', label: 'Investment projects' },
-  { path: '/omis', label: 'Orders (OMIS)' },
-  { path: config.performanceDashboardsUrl, label: 'MI dashboards' },
-]
-
 module.exports = function locals (req, res, next) {
   const baseUrl = `${(req.encrypted ? 'https' : req.protocol)}://${req.get('host')}`
   const breadcrumbItems = res.breadcrumb()
+  const userPermissions = get(res, 'locals.user.permissions')
 
   res.locals = Object.assign({}, res.locals, {
     BASE_URL: baseUrl,
@@ -32,14 +28,16 @@ module.exports = function locals (req, res, next) {
     BREADCRUMBS: breadcrumbItems,
     IS_XHR: req.xhr,
     QUERY: req.query,
-    GLOBAL_NAV_ITEMS: globalNavItems.map(globalNavItem => {
-      const url = globalNavItem.path
-
-      return Object.assign(globalNavItem, {
-        url,
-        isActive: req.path.startsWith(url),
-      })
-    }),
+    GLOBAL_NAV_ITEMS: GLOBAL_NAV_ITEMS
+      .filter(filterNonPermittedItem(userPermissions))
+      .map(navItem => {
+        const { path: url, label } = navItem
+        return {
+          label,
+          url,
+          isActive: req.path.startsWith(url),
+        }
+      }),
 
     getMessages () {
       return req.flash()
--- a/test/unit/apps/filters.test.js
+++ b/test/unit/apps/filters.test.js
@@ -1,146 +1,293 @@
 const moment = require('moment')
 
-const { filterDisabledOption } = require('~/src/apps/filters')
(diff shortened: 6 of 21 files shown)


src/lib/permissions/global-nav-items.js
```diff
@@ -0,0 +1,40 @@
...
+  { permission: 'omis', path: '/omis', label: 'Orders (OMIS)' },
+]
+
+// Currently we have no way of showing/hiding the MI nav item via permissions in DataHub. For now the MI link will
+// always show in the global nav and SSO will take care of users being able to access the service.
+const persistentNavItems = [
+  { path: config.performanceDashboardsUrl, label: 'MI dashboards' },
+]
+
+const globalNavItems = ({
```
teneightfive: Wonder if this could be simplified so that the nav items object has the actual permission rather than a key which is mapped to another object.

This could then be a filter on that list and check the items `permission` property against the available permissions for that user.

If there is no `permission` property it will always return it. Means you could do away with separate objects for _available_ and _persistent_ items.

src/apps/filters.js
```diff
@@ -13,6 +15,17 @@ function filterDisabledOption ({ currentValue = null, createdOn } = {}) {
   }
 }
 
+const filterPermittedItems = (userPermissions) => {
```
ztolley: Any reason for the different syntax?

test/unit/apps/filters.test.js
```diff
@@ -1,146 +1,293 @@
...
     beforeEach(() => {
-      this.options = [{
-        id: '1234',
-        name: 'Freds',
-        disabled_on: null,
-      }]
+      this.filterPermittedItems = filterPermittedItems
+      this.items = [
+        {
+          permissions: [
```
ztolley: maybe move onto a single line if just 1 value

src/middleware/locals.js
```diff
@@ -1,26 +1,80 @@
+const { get, map } = require('lodash')
+
 const logger = require('../../config/logger')
 const config = require('../../config')
+const { filterPermittedItems } = require('../apps/filters')
 
+const navigationItems = [
```
ztolley: can't help but wonder if labels should be in own file, but then again we don't need to support different languages

src/middleware/locals.js
```diff
@@ -1,26 +1,80 @@
...
-  { path: '/investment-projects', label: 'Investment projects' },
-  { path: '/omis', label: 'Orders (OMIS)' },
-  { path: config.performanceDashboardsUrl, label: 'MI dashboards' },
-]
+const globalNavItems = ({
+  reqPath,
+  userPermissions,
+  navItems = navigationItems,
+}) => {
+  const userNavItems = navItems.filter(filterPermittedItems(userPermissions))
```
ztolley: Could use native map so:

```
const result = navItems
  .filter(filterPermittedItems(userPermissions))
  .map((itemItem) => {
    const { url, label } = navItem
    return {
      url,
      label,
      isActive: reqPath.startsWith(url),
    }
  })

return result  
```

This pattern lets you move the map part it's into own function if you wanted, or in functions that get long.

Also put the url property first to follow our convention

src/apps/middleware.js
```diff
@@ -15,13 +17,16 @@ function setHomeBreadcrumb (name) {
...
-    res.locals.localNavItems = items.map(item => {
-      const url = item.isExternal ? item.url : `${req.baseUrl}/${item.path}`
-      return Object.assign(item, {
-        url,
-        isActive: res.locals.CURRENT_PATH === url,
+    const userPermissions = get(res, 'locals.user.permissions')
+    res.locals.localNavItems = items
+      .filter(filterPermittedItems(userPermissions))
+      .map(item => {
+        const url = item.isExternal ? item.url : `${req.baseUrl}/${item.path}`
```
ztolley: Changing the filter map in the earlier checking would make it consistent with this style

src/apps/middleware.js
```diff
@@ -15,13 +17,16 @@ function setHomeBreadcrumb (name) {
...
+        return Object.assign(item, {
```
ztolley: lodash assign?

src/apps/filters.js
feedmypixel: Not sure what you mean?

src/apps/middleware.js
feedmypixel: Sorry I don't understand what this comment is about/means?

test/unit/apps/filters.test.js
feedmypixel: yeah I started with this but swapping between multi line and single line was harder to read, so hence doing all multiline even if there is 1 element in the array read better.

src/middleware/locals.js
feedmypixel: I see what your saying, I'm swapping between native and lodash here. Will choose one. 👍

src/middleware/locals.js
feedmypixel: yeah this file does feel on the large side to me, I started off with abstracting the globalNavItems work into its own file but that didn't feel quite right.

src/apps/filters.js
ztolley: I meant using = () => instead of function ()

src/apps/contacts/controllers/documents.js
```diff
@@ -1,8 +1,12 @@
-const { get } = require('lodash')
+const { get, isUndefined } = require('lodash')
 
 async function renderDocuments (req, res, next) {
   const archivedDocumentPath = get(res.locals, 'contact.archived_documents_url_path')
 
+  if (isUndefined(archivedDocumentPath)) {
```
teneightfive: I know we were just discussing this on slack but wonder if whether then can access this particular route can be its own permission?

Then we can have a piece of shared middleware that accepts a list of permissions like we've used to nav items as to determine whether to call next with a `403` or to call next with no arguments.

Would keep our way of access routes consistent and not rely on always having to do an API call and hope it returns a `403` on one of the endpoints.

src/middleware/locals.js
teneightfive: I was wondering whether things like a global nav could be extracted out also. 

Don't think it was quite right as a _library_ but could these be part of an app-wide set of constants? And maybe also extract the local nav items into an app-level set of constants instead of in each router?

src/middleware/locals.js
```diff
@@ -82,3 +129,8 @@ module.exports = function locals (req, res, next) {
   })
   next()
 }
+
+module.exports = {
+  globalNavItems,
```
teneightfive: Not sure we need to export the `globalNavItems` from this module do we?

If they are used elsewhere we could solve that by extracting the nav items into their own module (maybe `constants`) and have this module just export the same function it did previously.

feedmypixel: I'm exporting here to test but have been thinking the test may not actually be needed as the permissions filter has its own test

src/apps/contacts/controllers/documents.js
feedmypixel: I like this idea but I feel this could be a future iteration. I am under pressure to get this work in and once its in then we can understand what can be fine tuned.

src/middleware/locals.js
feedmypixel: I like this idea but I feel this could be a future iteration. I am under pressure to get this work in and once its in then we can understand what can be fine tuned.

src/middleware/locals.js
feedmypixel: Saying that out loud confirmed it. Have pushed a fixup that cleans this up

src/middleware/locals.js
feedmypixel: have moved relevant work in this PR into `constants.js` files. Will do a follow up PR to do this for the other apps

src/middleware/locals.js
feedmypixel: > If they are used elsewhere we could solve that by extracting the nav items into their own module (maybe constants) and have this module just export the same function it did previously.

done 👍

src/apps/search/builders.js
```diff
@@ -1,15 +1,18 @@
...
-function buildSearchAggregation (apiResponseEntities) {
-  if (!isArray(apiResponseEntities)) { return }
+function buildSearchAggregation ({ aggregations, userPermissions, entityDetails = ENTITIES }) {
+  if (!isArray(aggregations)) { return }
 
-  return entities.map((defaultEntity) => {
+  const userEntities = entityDetails.filter(filterPermittedItems(userPermissions))
+
+  return userEntities.map((defaultEntity) => {
     return Object.assign(
```
yeahfro: Oh I hate to do this but we could put a `lodash` `assign` here

src/middleware/locals.js
```diff
@@ -32,14 +28,17 @@ module.exports = function locals (req, res, next) {
...
-
-      return Object.assign(globalNavItem, {
-        url,
-        isActive: req.path.startsWith(url),
-      })
-    }),
+    GLOBAL_NAV_ITEMS: GLOBAL_NAV_ITEMS
+      .filter(filterPermittedItems(userPermissions))
+      .map(navItem => {
+        const url = navItem.path
```
yeahfro: Only a minor one but couldn't we just keep this consistent and add it to the object? i.e.

```javascript
return {
  label: navItem.label,
  url: navItem.path,
  isActive: req.path.startsWith(url),
}
```

test/unit/apps/filters.test.js
```diff
@@ -1,146 +1,293 @@
...
-  context('when an option is not disabled', () => {
+  describe('#filterPermittedItems', () => {
     beforeEach(() => {
-      this.options = [{
-        id: '1234',
-        name: 'Freds',
-        disabled_on: null,
-      }]
+      this.filterPermittedItems = filterPermittedItems
+      this.items = [
```
yeahfro: As discussed, it could be worth having a look at other options for this data i.e. data file

src/apps/search/builders.js
teneightfive: Do we maybe want to do a PR that changes our existing `Object.assign` to a lodash assign for consistency too?

src/apps/filters.js
```diff
@@ -13,6 +15,17 @@ function filterDisabledOption ({ currentValue = null, createdOn } = {}) {
   }
 }
 
+function filterPermittedItems (userPermissions) {
```
teneightfive: Should this name be singular like the other filter in the file? It's only being done on one item too so makes more sense.

src/middleware/locals.js
```diff
@@ -1,26 +1,22 @@
...
-  { path: '/contacts', label: 'Contacts' },
-  { path: '/events', label: 'Events' },
-  { path: '/interactions', label: 'Interactions and services' },
-  { path: '/investment-projects', label: 'Investment projects' },
-  { path: '/omis', label: 'Orders (OMIS)' },
-  { path: config.performanceDashboardsUrl, label: 'MI dashboards' },
-]
-
-module.exports = function locals (req, res, next) {
+module.exports = function (req, res, next) {
```
teneightfive: I'd keep the function name on this line, makes it easier to trace in error stack traces.

src/middleware/locals.js
teneightfive: I'd agree. If if we're not we should have shorthand properties first too.

src/apps/search/builders.js
feedmypixel: > Do we maybe want to do a PR that changes our existing Object.assign to a lodash assign for consistency too?

Sounds good. I'll do this one for the moment to make @yeahfro happy

src/apps/contacts/controllers/documents.js
feedmypixel: So had a good look at this today and have a nice router permission solution but it will take a little bit of discussion. So as mentioned on slack I will raise a PR once this one is in and then there is less pressure time wise.

test/unit/apps/filters.test.js
feedmypixel: Will pick this up in a follow up PR

src/middleware/locals.js
```diff
@@ -32,14 +28,16 @@ module.exports = function locals (req, res, next) {
...
-
-      return Object.assign(globalNavItem, {
-        url,
-        isActive: req.path.startsWith(url),
-      })
-    }),
+    GLOBAL_NAV_ITEMS: GLOBAL_NAV_ITEMS
+      .filter(filterNonPermittedItem(userPermissions))
+      .map(navItem => {
+        const { path: url, label } = navItem
```
yeahfro: Nice piece of code, I will take note ;)

feedmypixel: Ah thanks. Glad you like my approach. So much slicker than your suggestion :troll: :wink:

PR #1080, opened 2017-12-13
Feature/temporary oauth logging

jim68000:
Adds a few lines to the syslog to expose what's happening when we get the OAuth state mismatch

### DIFF
--- a/src/apps/oauth/controllers.js
+++ b/src/apps/oauth/controllers.js
@@ -1,9 +1,12 @@
 const request = require('request-promise')
 const queryString = require('query-string')
 const uuid = require('uuid')
+
 const { get, set } = require('lodash')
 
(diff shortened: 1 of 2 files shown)


src/apps/oauth/controllers.js
```diff
@@ -32,7 +34,12 @@ async function callbackOAuth (req, res, next) {
   }
 
   if (sessionOAuthState !== stateQueryParam) {
-    return next(Error('There has been an OAuth stateId mismatch'))
+    // @TODO remove this once we've diagnosed the cause of teh mismatches
```
feedmypixel: lols, classic `teh` - change if you wish to

jim68000: Oh FSF

PR #1082, opened 2017-12-13
improve the robustness of the search feature

feedmypixel:
I am seeing [failures around this work on CircleCi](https://14274-83823675-gh.circle-artifacts.com/3/root/data-hub-frontend/cucumber/report.html). Added a `waitForElementPresent` to make sure the element is on the page before it asserts which tab it is on.


feedmypixel:
Hmmm not sure what is going on with this I think its CircleCi playing up

PR #1083, opened 2017-12-13
Fix breadcrumb display problems

teneightfive:
Previously the position of the breadcrumbs variable meant that it
could be set from other requests and would show the incorrect set of
items for a page.

Nesting the methods within the return functions keeps these values
isolated to that request.

This change also addresses a duplicate call to middleware within the contacts
sub-app.

### DIFF
--- a/src/middleware/breadcrumbs.js
+++ b/src/middleware/breadcrumbs.js
@@ -8,7 +8,7 @@ const {
   each,
   extend,
   find,
-  findIndex,
+  includes,
 } = require('lodash')
 
 /**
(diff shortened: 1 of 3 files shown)


src/middleware/breadcrumbs.js
```diff
@@ -21,48 +21,48 @@ const {
...
 function init () {
-  let breadcrumbs = []
+  return function (req, res, next) {
+    let breadcrumbs = []
 
-  function exists (breadcrumb) {
-    return findIndex(breadcrumbs, breadcrumb) !== -1
-  }
+    function exists (breadcrumb) {
+      return findIndex(breadcrumbs, breadcrumb) !== -1
```
feedmypixel: would `includes` read better? which then leads on to is there a need for this `exists` method?

src/middleware/breadcrumbs.js
```diff
@@ -21,48 +21,48 @@ const {
...
 
-  function addBreadcrumbs (name, url) {
-    if (arguments.length === 1) {
-      if (isArray(name)) {
-        each(name, (breadcrumb) => {
-          if (!exists(breadcrumb)) {
-            breadcrumbs.push(breadcrumb)
+    function addBreadcrumbs (name, url) {
+      if (arguments.length === 1) {
+        if (isArray(name)) {
```
feedmypixel: could this be represented as:

```
addCrumb: (crumb) => {
  if (!includes(breadcrumbs, crumb)){
    breadcrumbs.push(crumb)
  }
}

...

if (isArray(name)) {
  each(name, addCrumb)
} else if (isObject(name)) {
  addCrumb(name)
} else {
  addCrumb({ name: name }
}

...
```

src/middleware/breadcrumbs.js
```diff
@@ -21,48 +21,48 @@ const {
...
+            breadcrumbs.push({ name: name })
           }
-        })
-      } else if (isObject(name)) {
-        if (!exists(name)) {
-          breadcrumbs.push(name)
         }
-      } else {
+      } else if (arguments.length === 2) {
         if (!exists(name)) {
```
feedmypixel: Could this also then become:

```
addCrumb({
  name: name,
  url: url,
})
```

PR #1084, opened 2017-12-13
Unit tests tidy up sandbox usage

feedmypixel:
As a follow up to https://github.com/uktrade/data-hub-frontend/pull/1056 this work tidies up the sandbox usage we have across unit tests.

- makes a `sinon.sandbox.create()` a global named `sandbox`
- resets this `sandbox` in the global `afterEach`
- makes all functions arrow functions as a tidy up

### DIFF
--- a/test/unit/common.js
+++ b/test/unit/common.js
@@ -17,7 +17,7 @@ global.rootPath = `${process.cwd()}`
 global.rootPath = `${process.cwd()}`
 global.globalReq = reqres.req()
 global.globalRes = reqres.res()
-global.sandbox = sinon.sandbox.create()
+global.sandbox = null
 
 chai.config.truncateThreshold = 0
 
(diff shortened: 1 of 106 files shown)


test/unit/common.js
```diff
@@ -17,6 +17,7 @@ global.rootPath = `${process.cwd()}`
 global.rootPath = `${process.cwd()}`
 global.globalReq = reqres.req()
 global.globalRes = reqres.res()
+global.sandbox = sinon.sandbox.create()
```
teneightfive: Is this the right place to do it? Doesn't this mean each test will use the same instance of sandbox?

The previous way would scope a sandbox to that test and clean it up after.

ztolley: Yup, put it in the global file like the onExit

feedmypixel: > Yup, put it in the global file like the afterEach

The context in individual `describe` and `before` is not available in the `global-helpers` `beforeEach` or `afterEach`. Therefore this has been done as a global as its the only way.

feedmypixel: > Is this the right place to do it? Doesn't this mean each test will use the same instance of sandbox?

True but the `sandbox.restore()` in the global `afterEach` will reset that

ztolley: If the restore is in the afterEach then I'd put the create in the beforEach

PR #1091, opened 2017-12-15
Bring release numbers into dev

jim68000:
As otherwise the two heads will be forever out of sync


jim68000:
And now I've done two more releases - we will be back

PR #1093, opened 2017-12-18
Filter inactive countries in contact form

ztolley:
Minimal work to filter inactive countries and added a todo as the tests are old and not as well written as others, and we need to migrate the contact edit form to form generator.

### DIFF
--- a/src/apps/contacts/controllers/edit.js
+++ b/src/apps/contacts/controllers/edit.js
@@ -1,9 +1,10 @@
+const { get } = require('lodash')
+
 const { containsFormData } = require('../../../lib/controller-utils')
 const contactFormService = require('../services/form')
 const { contactLabels } = require('../labels')
-const metadataRepository = require('../../../lib/metadata')
 const companyRepository = require('../../companies/repos')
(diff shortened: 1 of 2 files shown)


src/apps/contacts/controllers/edit.js
```diff
@@ -47,9 +49,8 @@ async function editDetails (req, res, next) {
       res.breadcrumb(`Add contact at ${res.locals.company.name}`)
     }
 
-    // Labels and options needed for the form and error display
     res.locals.contactLabels = contactLabels
-    res.locals.countryOptions = metadataRepository.countryOptions.map(transformObjectToOption)
+    res.locals.countryOptions = await getOptions(token, 'country', { createdOn: get(res.locals, 'contact.created_on') })
```
teneightfive: I noticed you're actually doing this transformation in the `getOptions` method. I'm wondering if we should be. Or whether that should be up to where that piece of code uses the options?

ztolley: The function is only used to get options, so rather than have to transform it after every call it's much neater and less repetitive to do it there. So it does it what it says, it gets options

PR #1094, opened 2017-12-18
Fix issue with parsing field values into date formats

teneightfive:
Previously all fields were checked whether they were a date and if so
they would be formatted in the default date format when displaying
within a form.

The logic also contained exempt fields that could have returned
true incorrectly.

This change adjust that logic to only parse specified date fields
rather than checking all.

### DIFF
--- a/src/apps/omis/controllers/edit.js
+++ b/src/apps/omis/controllers/edit.js
@@ -43,6 +43,8 @@ class EditController extends FormController {
     delete sessionValues.errorValues
     delete sessionValues.errors
 
+    const dateFields = ['delivery_date']
+
     const orderValues = mapValues(req.form.options.fields, (fieldOptions, key) => {
       const newValue = get(res.locals, `order.${key}`)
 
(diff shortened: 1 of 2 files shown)


src/apps/omis/controllers/edit.js
```diff
@@ -65,16 +65,18 @@ class EditController extends FormController {
       return newValue
     })
 
-    const exemptFields = ['vat_number', 'po_number', 'amount']
+    const dateFields = ['delivery_date']
     // combine order values and error values
     let combinedValues = Object.assign({}, orderValues, sessionValues, errorValues)
     // convert dates to default format
     combinedValues = mapValues(combinedValues, (value, key) => {
```
feedmypixel: can you assign this to a new variable? feels a bit odd its a `let`. Maybe `const combinedValuesWithDefaultDateFormat = `

feedmypixel: or a shorted snappier name!

src/apps/omis/controllers/edit.js
```diff
@@ -65,16 +65,18 @@ class EditController extends FormController {
...
-      if (typeof value === 'string' && !exemptFields.includes(key)) {
-        const parsedDate = dateFns.parse(value.toString())
-        if (dateFns.isValid(parsedDate)) {
-          return dateFns.format(parsedDate, longDateFormat)
-        }
+      const parsedDate = dateFns.parse((value || '').toString())
+
+      if (
+        dateFields.includes(key) &&
```
feedmypixel: this formatting feels a little unusual, does it not work all on one line?

teneightfive: Think it was from when I had three statements so the line got long. Will combine back onto one line.

src/apps/omis/controllers/edit.js
teneightfive: Maybe `combinedValues` and `formattedValues`?

src/apps/omis/controllers/edit.js
ztolley: does it need datefns :) and what is toString producing?

PR #1095, opened 2017-12-18
Reconciliation as app

teneightfive:
This work refactors the listing view used for the reconciliation flow into its own app.

The research behind this is that the current journeys are slightly confusing as they jump between existing edit screens and new listing. The hierarchy also reflected in the breadcrumb isn't what is
desired for these journeys either.

This work creates a more intuitive and self-contained path for users within the reconciliation flow.

### Reconciliation listing view
image

### Reconciling a payment view
image

### Reconciled payment confirmation view
localhost_3001_omis_f804c606-fcf7-4eb7-bdd4-14fe7ecb7406_payment-receipt

### Reconciliation listing view
image

### Reconciling a payment view
image

### Reconciled payment confirmation view
localhost_3001_omis_reconciliation_f804c606-fcf7-4eb7-bdd4-14fe7ecb7406_payment-receipt


src/apps/middleware.js
```diff
@@ -13,6 +13,11 @@ function setHomeBreadcrumb (name) {
   }
 }
 
+function removeBreadcrumb (req, res, next) {
```
ztolley: Feels like an odd place to put this. Reason?

teneightfive: It's a simple middleware to allow you to remove a breadcrumb item at that point.

I call it before calling middleware and controllers that set subsequent breadcrumb items.

I went around a few different solutions to this but there may still be a better one if you spot it :)

Basically the alternative was to move the reconciliation app out to the top level but that felt stranger.

ztolley: Feels odd to set a breadcrumb as part of the logic executed in a regular piece of middleware or controller, but have a middleware just to remove breadcrumb. I can see that it works, just.. felt odd.

As for other places, is there a place where you would say `if x = y then removeBreadcrumb` or conditionally set the breadcrumb in the first place? Are there two routes to the same controller, in which one route needs a breadcrumb and another doesn't?

PR #1096, opened 2017-12-18
DH-1328 active event options

ztolley:
Set active options for event edit form.

Also discovered that don't need to set the `currentValue` parameter when calling getActiveOptions, as the option will be available because it was available at the creation time.

Will go back and modify company and contact to match.

### DIFF
--- a/src/apps/events/controllers/edit.js
+++ b/src/apps/events/controllers/edit.js
@@ -11,7 +11,10 @@ const { transformObjectToOption } = require('../../transformers')
 
 async function getEditOptions (token, createdOn, currentAdviser) {
   const advisers = await getAdvisers(token)
-  const activeAdvisers = filterActiveAdvisers({ advisers: advisers.results, includeAdviser: currentAdviser })
+  const activeAdvisers = filterActiveAdvisers({
+    advisers: advisers.results,
+    includeAdviser: currentAdviser,
+  })
 
   return {
     advisers: activeAdvisers.map(transformObjectToOption),
--- a/src/apps/events/controllers/edit.js
+++ b/src/apps/events/controllers/edit.js
@@ -6,23 +6,40 @@ const { transformEventResponseToFormBody } = require('../transformers')
 const { buildFormWithStateAndErrors } = require('../../builders')
 const { getAdvisers } = require('../../adviser/repos')
 const { filterActiveAdvisers } = require('../../adviser/filters')
+const { getOptions } = require('../../../lib/options')
+const { transformObjectToOption } = require('../../transformers')
+
+async function getEditOptions (token, createdOn, currentAdviser) {
+  const advisers = await getAdvisers(token)
+  const activeAdvisers = filterActiveAdvisers({ advisers: advisers.results, includeAdviser: currentAdviser })
+
+  return {
+    advisers: activeAdvisers.map(transformObjectToOption),
+    eventTypes: await getOptions(token, 'event-type', { createdOn }),
+    locationTypes: await getOptions(token, 'location-type', { createdOn }),
+    countries: await getOptions(token, 'country', { createdOn }),
+    teams: await getOptions(token, 'team', { createdOn }),
+    services: await getOptions(token, 'service', { createdOn }),
+    programmes: await getOptions(token, 'programme', { createdOn }),
+    ukRegions: await getOptions(token, 'uk-region', { createdOn }),
+  }
+}
 
 async function renderEditPage (req, res, next) {
   try {
     const eventData = transformEventResponseToFormBody(res.locals.event)
-
-    const currentAdviser = get(eventData, 'organiser.id')
-    const advisers = await getAdvisers(req.session.token)
-    const activeAdvisers = filterActiveAdvisers({ advisers: advisers.results, includeAdviser: currentAdviser })
 
     const eventId = get(eventData, 'id', '')
     const eventName = get(eventData, 'name')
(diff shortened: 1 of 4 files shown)


src/apps/events/controllers/edit.js
```diff
@@ -6,23 +6,40 @@ const { transformEventResponseToFormBody } = require('../transformers')
...
-    const advisers = await getAdvisers(req.session.token)
-    const activeAdvisers = filterActiveAdvisers({ advisers: advisers.results, includeAdviser: currentAdviser })
-
     const eventId = get(eventData, 'id', '')
     const eventName = get(eventData, 'name')
     const lead_team = eventData.lead_team || get(req, 'session.user.dit_team.id')
     const mergedData = pickBy(merge({}, eventData, { lead_team }, res.locals.requestBody))
 
+    const currentAdviser = get(eventData, 'organiser.id')
+    const options = await getEditOptions(req.session.token, mergedData.created_on, currentAdviser)
```
feedmypixel: So I haven't had 👀  on this `getOptions` work but it looks like you have nested `try/catch`. You have one here in the `renderEditPage` method but you also have a `try/catch` in the [getOptions](https://github.com/uktrade/data-hub-frontend/blob/b6b0b87d72cd0e8e695fe06dba2b143144d5a9fe/src/lib/options.js#L8-L20) method. 

Is this intended? feels like you don't need both and I would air towards not having one in the `getOptions` method.

src/apps/events/controllers/edit.js
```diff
@@ -6,23 +6,40 @@ const { transformEventResponseToFormBody } = require('../transformers')
...
+
+  return {
+    advisers: activeAdvisers.map(transformObjectToOption),
+    eventTypes: await getOptions(token, 'event-type', { createdOn }),
+    locationTypes: await getOptions(token, 'location-type', { createdOn }),
+    countries: await getOptions(token, 'country', { createdOn }),
+    teams: await getOptions(token, 'team', { createdOn }),
+    services: await getOptions(token, 'service', { createdOn }),
+    programmes: await getOptions(token, 'programme', { createdOn }),
+    ukRegions: await getOptions(token, 'uk-region', { createdOn }),
```
feedmypixel: is `getOptions` the right name to be used here? is this not just `getMetaData` or `getData` (bit rubbish that one)

src/apps/events/controllers/edit.js
```diff
@@ -6,23 +6,40 @@ const { transformEventResponseToFormBody } = require('../transformers')
 const { buildFormWithStateAndErrors } = require('../../builders')
 const { getAdvisers } = require('../../adviser/repos')
 const { filterActiveAdvisers } = require('../../adviser/filters')
+const { getOptions } = require('../../../lib/options')
+const { transformObjectToOption } = require('../../transformers')
+
+async function getEditOptions (token, createdOn, currentAdviser) {
+  const advisers = await getAdvisers(token)
+  const activeAdvisers = filterActiveAdvisers({ advisers: advisers.results, includeAdviser: currentAdviser })
```
feedmypixel: would this read better on separate lines?

test/unit/apps/events/controllers/edit.test.js
```diff
@@ -47,17 +88,30 @@ describe('Event edit controller', () => {
...
+        .get('/metadata/country/')
+        .reply(200, metadataMock.countryOptions)
+        .get('/metadata/team/')
+        .reply(200, metadataMock.teamOptions)
+        .get('/metadata/service/')
+        .reply(200, metadataMock.serviceOptions)
+        .get('/metadata/programme/')
+        .reply(200, metadataMock.programmeOptions)
+        .get('/metadata/uk-region/')
+        .reply(200, metadataMock.regionOptions)
```
feedmypixel: this is pure showing off 😉

test/unit/apps/events/controllers/edit.test.js
```diff
@@ -205,10 +349,10 @@ describe('Event edit controller', () => {
     })
   })
 
-  function getOrganiserFieldOptions (res) {
+  function getFormFieldOptions (res, fieldName) {
```
feedmypixel: can this sit at the top rather than the bottom of the file?

src/apps/events/controllers/edit.js
ztolley: It's not meta data and it's getting option data. It's ONLY used to get options for forms. Dom had the same question a few PR's back but couldn't think of a better name.

test/unit/apps/events/controllers/edit.test.js
ztolley: 💯

src/apps/events/controllers/edit.js
ztolley: Hmm was advised to add the try catch BUT you make a good point, the try catch should only be in the middleware or the controller, not something else.

test/unit/apps/events/controllers/edit.test.js
ztolley: will do

src/apps/events/controllers/edit.js
ztolley: Actually, the try-catch was there so that getting options doesn't break the whole page, but then again maybe it should as it could cause invalid data to be saved.

feedmypixel: To me it feels like there should only be a `try/catch` at the top level. 

> getting options doesn't break the whole page
It has potential for not letting you progress so in my mind that breaks the page

PR #1098, opened 2017-12-19
Optimise acceptance tests

yeahfro:
Changes to acceptance tests, as follows:
- removing creation of company where it is not required, favouring reusing the company fixture
- consolidation of steps
- removing anything that is redundant

### DIFF
--- a/test/acceptance/features/audit/contact.feature
+++ b/test/acceptance/features/audit/contact.feature
@@ -48,11 +48,11 @@ Feature: View Audit history of a contact
     Then a primary contact is added
     When I submit the form
     Then I see the success message
-    When I search for the contact
+    When I search for the created contact
     And the Contacts search tab is clicked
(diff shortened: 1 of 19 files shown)


yeahfro:
@ztolley I think what you are referring to is leaky tests and this was what @feedmypixel and I were agonising over. I feel like the ideal would be to treat each test in isolation by creating new entities for every scenario. The issue we are having is that tests are becoming a bottle neck because of performance.

This started out as an experiment and the performance gains are very good.

When we are testing the creation of companies and some collection actions we need to create new companies. As another example, when we are testing contacts which hang off companies then it is okay to reuse a company fixture.

ztolley:
It's a shame you can't create a company for a feature file, but yeah the performance gain is substantial compared to the job of creating a company for each run... 

The alternative would be to create the company via an API call instead of manually through the screens.

PR #1102, opened 2017-12-21
Fix export revenue (and others) in investment project view

yeahfro:
A bug means that certain booleans are not being presented correctly in the `Value` key value table for investment projects.

Note that acceptance tests will be in the next PR.

# Before

bad

screen shot 2017-12-21 at 11 51 16

# After

good

### DIFF
--- a/src/apps/investment-projects/services/formatting.js
+++ b/src/apps/investment-projects/services/formatting.js
@@ -140,56 +140,75 @@ function transformInvestmentDataForView (data) {
   })
 }
 
-function transformInvestmentValueForView (data) {
-  if (!isPlainObject(data)) { return }
-
-  function formatBoolean (boolean, { suffix = '', pos = 'Yes', neg = 'No' }) {
+function transformInvestmentValueForView ({
+  client_cannot_provide_total_investment,
+  total_investment,
+  client_cannot_provide_foreign_investment,
+  foreign_equity_investment,
+  number_new_jobs,
+  number_safeguarded_jobs,
+  government_assistance,
+  r_and_d_budget,
+  average_salary,
+  new_tech_to_uk,
+  export_revenue,
+  sector,
+  investor_company,
+  business_activities,
+  non_fdi_r_and_d_budget,
+  id,
+  associated_non_fdi_r_and_d_project,
+}) {
+  function formatBoolean (boolean, { pos, neg }) {
     if (isNull(boolean)) { return null }
-    return (data.government_assistance ? pos : neg) + suffix
+    return boolean ? pos : neg
   }
 
-  const value = {
-    total_investment: data.client_cannot_provide_total_investment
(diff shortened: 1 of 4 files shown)


test/unit/mocha.opts
```diff
@@ -1,3 +1,4 @@
+test/unit/global-helpers.js
```
ztolley: I thought this was already getting picked up

src/apps/investment-projects/services/formatting.js
```diff
@@ -140,56 +140,77 @@ function transformInvestmentDataForView (data) {
   })
 }
 
-function transformInvestmentValueForView (data) {
-  if (!isPlainObject(data)) { return }
-
-  function formatBoolean (boolean, { suffix = '', pos = 'Yes', neg = 'No' }) {
+function transformInvestmentValueForView ({
```
ztolley: So you had to re-write the whole thing?

test/unit/apps/investment-projects/formatting.test.js
```diff
@@ -213,76 +213,181 @@ describe('Investment project formatting service', () => {
...
+            new_tech_to_uk: true,
+            export_revenue: true,
+            sector: {
+              name: 'Renewable Energy : Wind : Renewable energy: Wind: Onshore',
+            },
+            investor_company: {
+              name: 'Venus Ltd',
+            },
+            business_activities: [
+              {
```
ztolley: it's like tabs vs spaces argument. I'd put [{ on a single line

src/apps/investment-projects/services/formatting.js
yeahfro: Yeah it was not very readable so I have brought it up to scratch

test/unit/mocha.opts
yeahfro: Required for running tests in my IDE

src/apps/investment-projects/services/formatting.js
```diff
@@ -140,56 +140,77 @@ function transformInvestmentDataForView (data) {
...
+  new_tech_to_uk,
+  export_revenue,
+  sector,
+  investor_company,
+  business_activities,
+  non_fdi_r_and_d_budget,
+  id,
+  associated_non_fdi_r_and_d_project,
+}) {
+  function formatBoolean (boolean, { pos, neg }) {
```
ztolley: Could these be moved outside this function and for european pass the activities into it.

ztolley: Also, I suspect the formatBoolean could be useful in other places.

yeahfro: The `hasEuropeanOrGlobalHeadquartersBusinessActivities` function was a refactor just to move some code out of the object and make it self documenting. We are not doing anything similar anywhere else. What would be the benefit of moving it outside?

With `formatBoolean` you might be onto something but that would be a separate PR.

PR #1104, opened 2017-12-21
Updates the interaction/service form to include active options.

ztolley:
DH-1332

Changed the interaction/service middleware to get active options instead of all of them, and put all options in an options property.
Simplified the controller a little when calling the form builder.
Updated the interaction form builder to use options passed into it.

### DIFF
--- a/src/apps/interactions/macros.js
+++ b/src/apps/interactions/macros.js
@@ -1,9 +1,6 @@
 const { assign } = require('lodash')
 
 const formLabels = require('./labels')
-const metaData = require('../../lib/metadata')
-const { transformContactToOption, transformObjectToOption } = require('../transformers')
-
 const currentYear = (new Date()).getFullYear()
 
 const interactionSortForm = {
@@ -66,27 +63,23 @@ const interactionFields = {
       macroName: 'MultipleChoiceField',
       name: 'contact',
       initialOption: '-- Select contact --',
-      options () {
-        return contacts.map(transformContactToOption)
-      },
+      options: contacts,
     }
   },
-  provider: {
-    macroName: 'MultipleChoiceField',
-    name: 'dit_team',
-    initialOption: '-- Select provider --',
-    options () {
-      return metaData.teams.map(transformObjectToOption)
-    },
+  provider (teams) {
+    return {
+      macroName: 'MultipleChoiceField',
+      name: 'dit_team',
+      initialOption: '-- Select provider --',
+      options: teams,
+    }
   },
   service (services) {
     return {
       macroName: 'MultipleChoiceField',
       name: 'service',
       initialOption: '-- Select service --',
-      options () {
-        return services.map(transformObjectToOption)
-      },
+      options: services,
     }
   },
   subject: {
@@ -107,18 +100,16 @@ const interactionFields = {
       macroName: 'MultipleChoiceField',
       name: 'dit_adviser',
       initialOption: '-- Select adviser --',
-      options () {
-        return advisers.map(transformContactToOption)
-      },
+      options: advisers,
     }
   },
-  communicationChannel: {
-    macroName: 'MultipleChoiceField',
-    name: 'communication_channel',
-    initialOption: '-- Select communication channel --',
-    options () {
-      return metaData.communicationChannelOptions.map(transformObjectToOption)
-    },
+  communicationChannel (channels) {
+    return {
+      macroName: 'MultipleChoiceField',
+      name: 'communication_channel',
+      initialOption: '-- Select communication channel --',
+      options: channels,
+    }
   },
 }
 
@@ -168,6 +159,8 @@ const interactionFormConfig = function ({
   contacts = [],
   advisers = [],
   services = [],
+  teams = [],
+  channels = [],
   hiddenFields,
 }) {
   return {
@@ -177,13 +170,13 @@ const interactionFormConfig = function ({
     hiddenFields,
     children: [
       interactionFields.contact(contacts),
-      interactionFields.provider,
+      interactionFields.provider(teams),
       interactionFields.service(services),
       interactionFields.subject,
       interactionFields.notes,
       interactionFields.date,
       interactionFields.adviser(advisers),
-      interactionFields.communicationChannel,
+      interactionFields.communicationChannel(channels),
     ].map(field => {
       return assign(field, {
         label: formLabels.interaction[field.name],
@@ -197,6 +190,7 @@ const serviceDeliveryFormConfig = function ({
   contacts = [],
   advisers = [],
   services = [],
+  teams = [],
   events = [],
   hiddenFields,
 }) {
@@ -207,7 +201,7 @@ const serviceDeliveryFormConfig = function ({
     hiddenFields,
     children: [
       interactionFields.contact(contacts),
-      interactionFields.provider,
+      interactionFields.provider(teams),
       interactionFields.adviser(advisers),
       // TODO this will be going once interactions are within events
       {
@@ -231,9 +225,7 @@ const serviceDeliveryFormConfig = function ({
         macroName: 'MultipleChoiceField',
         name: 'event',
         initialOption: '-- Select event --',
-        options () {
-          return events.map(transformObjectToOption)
-        },
+        options: events,
         modifier: 'subfield',
         condition: {
           name: 'is_event',
--- a/src/apps/interactions/macros.js
+++ b/src/apps/interactions/macros.js
@@ -113,7 +113,7 @@ const interactionFields = {
   },
 }
 
-const interactionFiltersFieldConfig = function (advisers) {
+const interactionFiltersFieldConfig = function (advisers = [], channels = []) {
   return [
     {
       macroName: 'MultipleChoiceField',
@@ -126,7 +126,7 @@ const interactionFiltersFieldConfig = function (advisers) {
     },
     assign(
       {},
-      interactionFields.communicationChannel,
+      interactionFields.communicationChannel(channels),
       { initialOption: '-- All channels --' }
     ),
     assign(
--- a/src/apps/interactions/middleware/details.js
+++ b/src/apps/interactions/middleware/details.js
@@ -3,12 +3,13 @@ const { sentence } = require('case')
 
 const { transformInteractionFormBodyToApiRequest } = require('../transformers')
 const { fetchInteraction, saveInteraction } = require('../repos')
-const metaDataRepository = require('../../../lib/metadata')
 const { getContactsForCompany, getContact } = require('../../contacts/repos')
 const { getAdvisers } = require('../../adviser/repos')
 const { filterActiveAdvisers } = require('../../adviser/filters')
 const { getActiveEvents } = require('../../events/repos')
 const { getDitCompany } = require('../../companies/repos')
+const { transformObjectToOption, transformContactToOption } = require('../../transformers')
+const { getOptions } = require('../../../lib/options')
 
 async function postDetails (req, res, next) {
   res.locals.requestBody = transformInteractionFormBodyToApiRequest(req.body)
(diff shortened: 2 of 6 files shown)


src/apps/interactions/macros.js
```diff
@@ -66,27 +63,23 @@ const interactionFields = {
...
-    initialOption: '-- Select provider --',
-    options () {
-      return metaData.teams.map(transformObjectToOption)
-    },
+  provider (teams) {
+    return {
+      macroName: 'MultipleChoiceField',
+      name: 'dit_team',
+      initialOption: '-- Select provider --',
+      options: teams,
```
feedmypixel: How come are the `options` being overridden here rather than just changing them in [globalFields.teams](https://github.com/uktrade/data-hub-frontend/blob/930d45d7cba5e085000790242d79aa3c88343ace/src/apps/macros.js#L105-L113)

src/apps/interactions/macros.js
```diff
@@ -66,27 +63,23 @@ const interactionFields = {
...
   },
   service (services) {
     return {
       macroName: 'MultipleChoiceField',
       name: 'service',
       initialOption: '-- Select service --',
-      options () {
-        return services.map(transformObjectToOption)
-      },
+      options: services,
```
feedmypixel: How come are the options being overridden here rather than just changing them in [globalFields.serviceDeliveryServices](https://github.com/uktrade/data-hub-frontend/blob/930d45d7cba5e085000790242d79aa3c88343ace/src/apps/macros.js#L95-L103)

src/apps/interactions/macros.js
```diff
@@ -107,22 +100,20 @@ const interactionFields = {
       macroName: 'MultipleChoiceField',
       name: 'dit_adviser',
       initialOption: '-- Select adviser --',
-      options () {
-        return advisers.map(transformContactToOption)
-      },
+      options: advisers,
```
feedmypixel: can this go into [globalFields](https://github.com/uktrade/data-hub-frontend/blob/930d45d7cba5e085000790242d79aa3c88343ace/src/apps/macros.js)?

src/apps/interactions/macros.js
```diff
@@ -107,22 +100,20 @@ const interactionFields = {
...
     }
   },
-  communicationChannel: {
-    macroName: 'MultipleChoiceField',
-    name: 'communication_channel',
-    initialOption: '-- Select communication channel --',
-    options () {
-      return metaData.communicationChannelOptions.map(transformObjectToOption)
-    },
+  communicationChannel (channels) {
```
feedmypixel: can this go into [globalFields](https://github.com/uktrade/data-hub-frontend/blob/930d45d7cba5e085000790242d79aa3c88343ace/src/apps/macros.js)?

src/apps/interactions/middleware/details.js
```diff
@@ -66,20 +67,30 @@ async function getInteractionDetails (req, res, next, interactionId) {
 
 async function getInteractionOptions (req, res, next) {
   try {
-    const interaction = res.locals.interaction
     const token = req.session.token
-    const currentAdviser = get(res.locals, 'interaction.dit_adviser.id')
+    const createdOn = get(res.locals, 'interaction.created_on')
+
     const company = get(res.locals, 'company.id')
```
feedmypixel: this is `companyId` right?

test/unit/apps/interactions/macros.test.js
```diff
@@ -1,70 +1,113 @@
...
-          { id: 'omega', name: 'Omega' },
-        ],
-      },
-    })
-  })
+const macros = require('~/src/apps/interactions/macros')
 
+function getFormFieldOptions (form, fieldName) {
+  const formFields = form.children
+  const field = find(formFields, field => field.name === fieldName)
```
feedmypixel: can this go on separate lines?

test/unit/apps/interactions/macros.test.js
```diff
@@ -1,70 +1,113 @@
...
-    })
+      it('should return object with specified hidden fields', () => {
+        expect(this.form).to.have.property('hiddenFields', this.formParams.hiddenFields)
+      })
 
-    it('should return object with transformed dit_adviser options', () => {
-      const actual = this.macros.interactionFormConfig({ advisers: sampleOptions })
-      actual.children.find(x => x.name === 'dit_adviser').options()
+      it('should return object with contact options', () => {
+        const fieldOptions = getFormFieldOptions(this.form, 'contact')
```
feedmypixel: would `<type>FieldOptions` read better?

test/unit/apps/interactions/macros.test.js
```diff
@@ -1,70 +1,113 @@
...
 
   describe('#serviceDeliveryFormConfig', () => {
-    it('should return object with transformed event options', () => {
-      const actual = this.macros.serviceDeliveryFormConfig({ events: sampleOptions })
-      actual.children.find(x => x.name === 'event').options()
+    context('when called with service delivery options', () => {
+      beforeEach(() => {
+        this.formParams = {
+          returnLink: 'test',
+          contacts: sandbox.stub(),
```
feedmypixel: is this correct that these are stubs? should they not be values?

test/unit/apps/interactions/middleware/details.test.js
```diff
@@ -245,32 +266,91 @@ describe('Interaction details middleware', () => {
...
+        }, {
+          label: 'Jane Smith, Director',
+          value: '12651151-2149-465e-871b-ac45bc568a64',
+        }]
+
+        expect(this.res.locals.options.contacts).to.deep.equal(expectedContacts)
+      })
+
+      it('should set the active adviser options', () => {
+        const expectedAdvisers = [{
```
feedmypixel: not sure if these expected objects should sit outside of the `it` definitions to make it easier to scan them. I'd say have a play to see how it feels.

test/unit/apps/interactions/middleware/details.test.js
```diff
@@ -298,36 +384,91 @@ describe('Interaction details middleware', () => {
           dit_adviser: this.currentAdviser,
         }
 
-        await this.middleware.getInteractionOptions(this.req, this.res, this.nextSpy)
+        const { getInteractionOptions } = require('~/src/apps/interactions/middleware/details')
```
feedmypixel: can this go at top of the file?

test/unit/apps/interactions/middleware/details.test.js
```diff
@@ -298,36 +384,91 @@ describe('Interaction details middleware', () => {
...
-        expect(this.res.locals.services).to.deep.equal(servicesData)
-      })
-
-      it('should get the active events for the record', () => {
-        expect(this.getActiveEventsSpy).to.be.calledWith(this.req.session.token, this.res.locals.interaction.created_on)
-      })
-
-      it('should set events data on locals', () => {
-        expect(this.res.locals.events).to.deep.equal(eventsData.results)
+        const expectedContacts = [{
```
feedmypixel: same as a above comment, feels like these `it` would scan/read better sat inside of the `context/describe` or at the top of the file?

test/unit/apps/interactions/middleware/details.test.js
```diff
@@ -298,36 +384,91 @@ describe('Interaction details middleware', () => {
...
+          value: '3',
+        }, {
+          label: 'Fred Smith',
+          value: '4',
+        }]
+
+        expect(this.res.locals.options.advisers).to.deep.equal(expectedAdvisers)
+      })
+
+      it('should set service options', async () => {
```
feedmypixel: does this need to be `async`?

test/unit/apps/interactions/middleware/details.test.js
```diff
@@ -298,36 +384,91 @@ describe('Interaction details middleware', () => {
...
+          value: '1',
+        }, {
+          label: 'sv3',
+          value: '3',
+        }]
+
+        expect(this.res.locals.options.services).to.deep.equal(expectedServiceOptions)
+      })
+
+      it('should set team options', async () => {
```
feedmypixel: does this need to be async?

test/unit/apps/interactions/middleware/details.test.js
```diff
@@ -298,36 +384,91 @@ describe('Interaction details middleware', () => {
...
+          value: '1',
+        }, {
+          label: 'te3',
+          value: '3',
+        }]
+
+        expect(this.res.locals.options.teams).to.deep.equal(expectedTeamOptions)
+      })
+
+      it('should set channel options', async () => {
```
feedmypixel: does this need to be async?

test/unit/apps/interactions/middleware/details.test.js
```diff
@@ -298,36 +384,91 @@ describe('Interaction details middleware', () => {
...
+          value: '1',
+        }, {
+          label: 'c3',
+          value: '3',
+        }]
+
+        expect(this.res.locals.options.channels).to.deep.equal(expectedChannelOptions)
+      })
+
+      it('should set event options', async () => {
```
feedmypixel: does this need to be async?

test/unit/apps/interactions/middleware/details.test.js
```diff
@@ -298,36 +384,91 @@ describe('Interaction details middleware', () => {
...
+          value: '31a9f8bd-7796-4af4-8f8c-25450860e2d1',
+          label: 'A United Kingdom Get together',
+        }, {
+          value: '2f4cbd6e-9cc7-4b03-93fe-80a4d8cea4af',
+          label: 'Greatest feast ever',
+        }, {
+          value: '4aa90860-ddb8-4cc0-8d5f-e0c26567abe4',
+          label: 'name',
+        }]
+        expect(this.res.locals.options.events).to.deep.equal(expectedEvents)
```
feedmypixel: this is a MEGA test file! can anything be split out?

src/apps/interactions/macros.js
ztolley: As described in other PR, we need to pass the options into the form as the options depend on the created date for the record.

src/apps/interactions/macros.js
ztolley: Maybe in a separate PR

src/apps/interactions/macros.js
ztolley: This is the only form that uses it

test/unit/apps/interactions/macros.test.js
ztolley: it can not sure why

test/unit/apps/interactions/macros.test.js
ztolley: Stubs/Mocks can represent functions or values. Would you rather I use mock() or an empty object?

test/unit/apps/interactions/middleware/details.test.js
ztolley: They belong inside the `it` so it's easy to read when reading the test to understand what you expect the answer to be.

Normally I'd put the value straight in the expect but I think you asked it to be moved to a variable in a previous PR to make it easier to read.

test/unit/apps/interactions/middleware/details.test.js
ztolley: I disagree. Moving them somewhere else would make it hard to find what the expected value is and it is not part of setting up the context, rather it is a value to be tested and if anything I'd put it inside the equal parameter of the expect call.

test/unit/apps/interactions/middleware/details.test.js
ztolley: It is indeed, maybe as a separate PR we can split the details file and the tests with it

test/unit/apps/interactions/middleware/details.test.js
ztolley: Turns out there was a TON of tests there and already exists as well as the new ones that should not be async

src/apps/interactions/macros.js
feedmypixel: > This is the only form that uses it

at the moment right? Always good to abstract

test/unit/apps/interactions/macros.test.js
feedmypixel: > it can not sure why

So its the same as the rest of the codebase

test/unit/apps/interactions/macros.test.js
feedmypixel: > Would you rather I use mock() or an empty object?

Values to test against

test/unit/apps/interactions/middleware/details.test.js
feedmypixel: > TON of tests there and already exists as well as the new ones that should not be async

Yup seen this also, copy and paste is the culprit I should imagine

src/apps/interactions/macros.js
feedmypixel: > As described in other PR, we need to pass the options into the form as the options depend on the created date for the record.

As spoke about on slack this all feels like it is in multiple places when it could be in 1. I will check the other PR's code out and have a look ASAP

src/apps/interactions/macros.js
ztolley: Premature optimisation and we have many other fields not moved into the global fields.

PR #1113, opened 2017-12-27
Route based permissions

feedmypixel:
Follow up pr to https://github.com/uktrade/data-hub-frontend/pull/1079

This work:

- Will render a `403` at the routing level, rather than hitting the API for a `403`. Uses previous permissions in the frontend application
- Only shows `Interactions` on the dashboard if user has correct permissions
- Renders a `403` if user tries to go to search url without correct permissions, rather than empty results
- Moves the documents permissions to routes

#### Dashboard
lep-dashboard

#### Events search
screen shot 2017-12-27 at 14 39 38

#### Dashboard
da-dashboard

#### Events search
da-events-search

#### Dashboard
dit_user-dashboard

#### Events search
screen shot 2017-12-27 at 14 44 35

### DIFF
--- a/src/apps/companies/constants.js
+++ b/src/apps/companies/constants.js
@@ -1,3 +1,8 @@
+const DEFAULT_COLLECTION_QUERY = {
+  sortby: 'modified_on:desc',
+  archived: false,
+}
+
 const LOCAL_NAV = [
   {
     path: 'details',
@@ -48,11 +53,6 @@ const LOCAL_NAV = [
     ],
   },
 ]
-
-const DEFAULT_COLLECTION_QUERY = {
-  sortby: 'modified_on:desc',
-  archived: false,
-}
 
 module.exports = {
   LOCAL_NAV,
--- a/src/apps/middleware.js
+++ b/src/apps/middleware.js
@@ -1,7 +1,12 @@
-const { get, isEmpty, assign } = require('lodash')
+const { get, isEmpty, assign, intersection, isUndefined } = require('lodash')
 const queryString = require('query-string')
+const { parse } = require('url')
 
 const { filterNonPermittedItem } = require('./filters')
+
+function userHasPermission (routePermissions, userPermissions) {
+  return intersection(routePermissions, userPermissions).length > 0
+}
 
 function setHomeBreadcrumb (name) {
   return function (req, res, next) {
@@ -18,6 +23,27 @@ function setHomeBreadcrumb (name) {
 function removeBreadcrumb (req, res, next) {
   res.removeBreadcrumb()
   next()
+}
+
+function isPermittedRoute (pathname, routes, userPermissions) {
+  const routePermissions = get(routes.find((route) => {
+    return pathname.endsWith(route.path)
+  }), 'permissions')
+
+  return isUndefined(routePermissions) || userHasPermission(routePermissions, userPermissions)
+}
+
+function handleRoutePermissions (routes) {
+  return function handleRestrictedRoute (req, res, next) {
+    const userPermissions = get(res, 'locals.user.permissions')
+    const pathname = parse(req.originalUrl).pathname
+
+    if (!isPermittedRoute(pathname, routes, userPermissions)) {
+      return next({ statusCode: 403 })
+    }
+
+    return next()
+  }
 }
 
 function setLocalNav (items = []) {
@@ -56,4 +82,6 @@ module.exports = {
   setLocalNav,
   redirectToFirstNavItem,
   setDefaultQuery,
+  handleRoutePermissions,
+  isPermittedRoute,
 }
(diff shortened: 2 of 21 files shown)


test/unit/apps/middleware.test.js
```diff
@@ -212,4 +212,170 @@ describe('Apps middleware', () => {
...
+      this.middleware.handleRoutePermissions([
+        {
+          path: mockUrl,
+          permissions: [
+            'permission1',
+          ],
+        },
+      ])(this.reqMock, this.resMock, this.nextSpy)
+
+      expect(this.nextSpy).to.be.calledWith()
```
yeahfro: Tempting to consolidate the previous test with this..?

test/unit/apps/middleware.test.js
```diff
@@ -212,4 +212,170 @@ describe('Apps middleware', () => {
...
+      this.middleware.handleRoutePermissions([
+        {
+          path: mockUrl,
+          permissions: [
+            'permission2',
+          ],
+        },
+      ])(this.reqMock, this.resMock, this.nextSpy)
+
+      expect(this.nextSpy).to.be.calledWith({ statusCode: 403 })
```
yeahfro: Tempting to consolidate the previous test with this..?

test/unit/apps/middleware.test.js
```diff
@@ -212,4 +212,170 @@ describe('Apps middleware', () => {
...
+
+    it('should call next with error when route not permitted', () => {
+      this.middleware.handleRoutePermissions([
+        {
+          path: mockUrl,
+          permissions: [],
+        },
+      ])(this.reqMock, this.resMock, this.nextSpy)
+
+      expect(this.nextSpy).to.be.calledWith()
```
yeahfro: Tempting to consolidate the previous test with this..?

test/unit/apps/middleware.test.js
```diff
@@ -212,4 +212,170 @@ describe('Apps middleware', () => {
...
+
+      expect(this.nextSpy).to.be.calledOnce
+    })
+
+    it('should call next with error when route not permitted', () => {
+      this.middleware.handleRoutePermissions([
+        { path: mockUrl },
+      ])(this.reqMock, this.resMock, this.nextSpy)
+
+      expect(this.nextSpy).to.be.calledWith()
```
yeahfro: Tempting to consolidate the previous test with this..?

src/apps/companies/constants.js
```diff
@@ -1,3 +1,8 @@
+const DEFAULT_COLLECTION_QUERY = {
```
ztolley: What does this have to do with permissions?

src/apps/middleware.js
```diff
@@ -20,6 +25,34 @@ function removeBreadcrumb (req, res, next) {
...
+  }), 'permissions')
+
+  if (
+    isUndefined(routePermissions) ||
+    userHasPermission(routePermissions, userPermissions)
+  ) {
+    return true
+  }
+
+  return false
```
yeahfro: could we just replace this with `return isUndefined(routePermissions) || userHasPermission(routePermissions, userPermissions)` ?

src/apps/companies/constants.js
feedmypixel: Please have a look at the commit message. Technically it doesn't but the permissions work has involved moving items into `constants.js` files. So this is a tidy up as I am in this area.

src/apps/middleware.js
feedmypixel: that gave me the eslint error:
screen shot 2017-12-27 at 15 31 55

so have compromised and gone with 
`return isUndefined(routePermissions) || userHasPermission(routePermissions, userPermissions)`

test/unit/apps/middleware.test.js
feedmypixel: So after @ztolley and @teneightfive mentioned a test should be doing 1 thing I have started to follow that methodology. I quite like it if you can move things into `beforeEach`. Maybe something we should all discuss?

test/unit/apps/middleware.test.js
feedmypixel: see above ☝️

test/unit/apps/middleware.test.js
feedmypixel: see above, above☝️

test/unit/apps/middleware.test.js
feedmypixel: see above, above, above ☝️ 😁

test/unit/apps/middleware.test.js
ztolley: Yup. The pattern we have followed that naturally flows from the use of context is context describes a situation, beforeEach sets it up and makes the call to the function and it is followed by multiple 'it' to test different aspects of the result. 

A test should test 'one thing' which not only keeps the test focussed but also makes it clear exactly what failed when it does it makes it easier to diagnose the cause.

PR #1114, opened 2017-12-27
update to latest semantic versions

feedmypixel:
- Update dependencies to latest versions.
- Fixes lint errors caused by bumping eslint version

- [x] `/locahost` runs
- [x] `yarn lint`
- [x] `yarn test:unit`
- [x] `yarn tests:acceptance`
- [x] a quick browse on the [demo app](https://datahub-beta2-pr-1114.herokuapp.com/) and all looks ok


package.json
```diff
@@ -8,95 +8,95 @@
     "url": "https://github.com/uktrade/data-hub-frontend"
```
jim68000: It might be worth bumping node to whatever the latest PaaS supports up there 👆

feedmypixel: > It might be worth bumping node to whatever the latest PaaS supports up there

Lets do this a separate PR and release. I will follow up with that if ok?

PR #1125, opened 2018-01-03
DH-1390 Display the "Date and time created" for an investment project

yeahfro:
DH-1390

Adding the `created_on` field to the investment project local header meta data.

Change also includes:
- new unit tests as the middleware was not covered
- modified acceptance tests to test for date presented in local header was within the past minute

screen shot 2018-01-05 at 23 06 55

### DIFF
--- a/test/acceptance/features/details/step_definitions/details.js
+++ b/test/acceptance/features/details/step_definitions/details.js
@@ -1,24 +1,9 @@
 const { client } = require('nightwatch-cucumber')
 const { defineSupportCode } = require('cucumber')
 const { get, includes } = require('lodash')
-const { addMinutes, isAfter } = require('date-fns')
-const moment = require('moment')
 
-const { mediumDateTimeFormat } = require('../../../../../config')
 const { getDetailsTableRowValue } = require('../../../helpers/selectors')
-
-const formatters = {
-  isEuropeanOrGlobalHeadquartersFormatter (expected, actual) {
-    const formatted = /^(european|global) headquarters$/i.test(expected) ? 'Yes' : 'No'
-    return formatted === actual
-  },
-  isRecentDateFormatter (expected, actual) {
-    const actualDate = moment(actual, mediumDateTimeFormat).toDate()
-    const oneMinuteAgo = addMinutes(actualDate, -1)
-
-    return isAfter(actualDate, oneMinuteAgo)
-  },
-}
+const formatters = require('../../../helpers/formatters')
 
 function getExpectedValue (row, state) {
   if (includes(row.value, '.') && !includes(row.value, ' ')) {
--- a/test/acceptance/features/details/step_definitions/details.js
+++ b/test/acceptance/features/details/step_definitions/details.js
@@ -1,12 +1,22 @@
 const { client } = require('nightwatch-cucumber')
 const { defineSupportCode } = require('cucumber')
 const { get, includes } = require('lodash')
+const { addMinutes, isAfter } = require('date-fns')
+const moment = require('moment')
 
+const { mediumDateTimeFormat } = require('../../../../../config')
 const { getDetailsTableRowValue } = require('../../../helpers/selectors')
 
 const formatters = {
-  formatEuropeanOrGlobalHeadquarters (value) {
-    return /^(european|global) headquarters$/i.test(value) ? 'Yes' : 'No'
+  isEuropeanOrGlobalHeadquartersFormatter (expected, actual) {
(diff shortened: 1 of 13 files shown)


test/acceptance/features/details/step_definitions/details.js
```diff
@@ -1,12 +1,20 @@
 const { client } = require('nightwatch-cucumber')
 const { defineSupportCode } = require('cucumber')
 const { get, includes } = require('lodash')
+const moment = require('moment')
 
 const { getDetailsTableRowValue } = require('../../../helpers/selectors')
 
 const formatters = {
```
feedmypixel: feels like this could be split out (but I think I mentioned that before but not sure if it was on this code)

test/acceptance/features/details/step_definitions/details.js
```diff
@@ -115,11 +119,15 @@ defineSupportCode(({ Then }) => {
...
       const detailsTableRowValueXPathSelector = detailsTableSelector.selector + rowValueSelector.selector
       const expectedValue = getExpectedValue(row, this.state)
-      const formattedExpectedValue = formatExpectedValue(row, expectedValue)
       await Details
         .api.useXpath()
         .waitForElementPresent(detailsTableRowValueXPathSelector)
-        .assert.containsText(detailsTableRowValueXPathSelector, formattedExpectedValue)
+        .getText(detailsTableRowValueXPathSelector, (result) => {
+          if (row.format) {
+            return client.expect(formatters[row.format](expectedValue, result.value)).to.be.true
```
feedmypixel: you kind of loose what this is actually doing, what is the expectation here? Communicating through the function name would really help at a quick glance.

```
formatters[row.format](expectedValue, result.value)
```
or:

```
isEuropeanOrGlobalHeadquartersFormat(expectedValue, result.value)
```

feedmypixel: if i'm understanding it correctly?

test/acceptance/features/investment-projects/collection.feature
```diff
@@ -16,18 +16,19 @@ Feature: View a list of Investment Projects
...
-      | Primary sector                | investmentProject.primarySector               |
-      | Business activity             | investmentProject.businessActivity            |
-      | Client contacts               | investmentProject.clientContact               |
-      | Project description           | investmentProject.description                 |
-      | Anonymised description        | investmentProject.anonymousDescription        |
-      | Estimated land date           | investmentProject.estimatedLandDate           |
-      | New or existing investor      | investmentProject.investorType                |
-      | Level of involvement          | investmentProject.levelOfInvolvement          |
-      | Specific investment programme | investmentProject.specificInvestmentProgramme |
+      | key                           | value                                         | format             |
```
feedmypixel: I'd call this `formatter` to align to the hash of `formatters`

test/acceptance/features/details/step_definitions/details.js
```diff
@@ -1,12 +1,20 @@
...
 const { getDetailsTableRowValue } = require('../../../helpers/selectors')
 
 const formatters = {
-  formatEuropeanOrGlobalHeadquarters (value) {
-    return /^(european|global) headquarters$/i.test(value) ? 'Yes' : 'No'
+  isEuropeanOrGlobalHeadquartersFormat (expected, actual) {
+    const formatted = /^(european|global) headquarters$/i.test(expected) ? 'Yes' : 'No'
+    return formatted === actual
+  },
+  isRecentDateFormat (expected, actual) {
```
feedmypixel: I'd change this to `Formatter`

src/apps/investment-projects/services/formatting.js
```diff
@@ -137,6 +137,7 @@ function transformInvestmentDataForView (data) {
     business_activities: businessActivities.map(i => i.name).join(', '),
     client_contacts: data.client_contacts.map(i => i.name).join(', '),
     estimated_land_date: data.estimated_land_date ? moment(data.estimated_land_date).format('MMMM YYYY') : null,
+    created_date: moment(data.created_on).format('DD/MM/YY HH:mm'),
```
teneightfive: Can we use the standard date format we've used everywhere else for this?

teneightfive: Would probably say the [long date format](https://github.com/uktrade/data-hub-frontend/blob/799b41adc8d089db50bfdba5c3b2d2a61c44df97/config/index.js#L52) from the config file.

teneightfive:
Having a little think about the placement of the date and it doesn't really feel right within that project summary block.

What about placing it in the high level header:

image

We do something similar already within OMIS orders:

image

PR #1126, opened 2018-01-03
Remove text referring to feature that doesn't exist yet

teneightfive:
There is some text within the OMIS order creation journey that refers
to a feature that hasn't yet been completed.

For now we're going to remove that line and we can revert the commit
when the work has been done.


ztolley:
I'm guessing the company sort fail will be fixed by @feedmypixel's fix

PR #1127, opened 2018-01-03
Remove logic around which address is used for OMIS orders

teneightfive:
We are not actually using the company's trading address by default
so we can remove the logic that surrounds this within the work
order and billing details.

### DIFF
--- a/src/apps/omis/apps/view/views/work-order.njk
+++ b/src/apps/omis/apps/view/views/work-order.njk
@@ -217,25 +217,14 @@
     company.registered_address_country.name
   ] | removeNilAndEmpty %}
 
-  {% set tradingAddress = [
-    company.trading_address_1,
-    company.trading_address_2,
-    company.trading_address_town,
-    company.trading_address_county,
-    company.trading_address_postcode,
-    company.trading_address_country.name
-  ] | removeNilAndEmpty %}
-
-  {% set companyAddress = tradingAddress if tradingAddress|length else registeredAddress %}
-
   {% set addressValue %}
     {% if billingAddress | length %}
       {{ billingAddress | join(', ') }}
     {% else %}
-      <p>{{ companyAddress | join(', ') }}</p>
+      <p>{{ registeredAddress | join(', ') }}</p>
 
       {% call Message({ type: 'muted' }) %}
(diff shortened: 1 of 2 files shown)


src/apps/omis/apps/view/views/work-order.njk
```diff
@@ -217,25 +217,14 @@
...
-  ] | removeNilAndEmpty %}
-
-  {% set companyAddress = tradingAddress if tradingAddress|length else registeredAddress %}
-
   {% set addressValue %}
     {% if billingAddress | length %}
       {{ billingAddress | join(', ') }}
     {% else %}
-      <p>{{ companyAddress | join(', ') }}</p>
+      <p>{{ registeredAddress | join(', ') }}</p>
```
ztolley: Don't we have a formatter for this yet? Using this method will lead to the appearance of double commas, though feel free to tell me that is outside the scope of the fix.

src/apps/omis/apps/view/views/work-order.njk
```diff
@@ -217,25 +217,14 @@
...
 
       {% call Message({ type: 'muted' }) %}
-        The company's {{ 'trading' if tradingAddress|length else 'registered' }} address is currently being used for the invoice.
+        The company’s registered address is currently being used for the invoice.
```
ztolley: it that comma consistent with all the other commas? :)

src/apps/omis/apps/view/views/work-order.njk
teneightfive: It won't in this case. The `registeredAddress` uses the `removeNilAndEmpty` filter which removes empty values. The join method will then also not apply a comma to the last item so you won't get any strange commas.

src/apps/omis/apps/view/views/work-order.njk
teneightfive: We should be using the apostrophe in text where we can. When I spot a single quote now I try and replace them in context of a similar change.

There is a [good article](https://webdesignledger.com/common-typography-mistakes-apostrophes-versus-quotation-marks/) I sent to @feedmypixel when this cropped up in a previous PR. Worth a read to understand the difference.

ztolley: yeah apostrophe, sorry. late

PR #1129, opened 2018-01-03
Fix/contact list transformer

ztolley:
DH-1418

Contact search results or contact lists that contain contacts with only a first or last name caused the server to throw a 500 error.

There were 2 issues that cause this. The first was the contact list item transformer would only transform a contact with both first and last name, if either was missing it would return undefined.

The 2nd issue is that in `transformApiResponseToCollection`  if an item transformer returns undefined for an item it is passed to any other transformers in the chain and the function calling the transformer.

This fix corrects the contact list item transformer to look for either first or last name and this is followed up by a filter that removes undefined items between calls to transformers in the chain.

### DIFF
--- a/src/apps/contacts/transformers.js
+++ b/src/apps/contacts/transformers.js
@@ -27,7 +27,7 @@ function transformContactToListItem ({
   company_sector,
   primary,
 } = {}) {
-  if (!id || !first_name || !last_name) { return }
+  if (!id || (!first_name && !last_name)) { return }
 
   const metaItems = [
     { key: 'company', value: get(company, 'name') },
--- a/test/unit/apps/transformers.test.js
+++ b/test/unit/apps/transformers.test.js
@@ -185,34 +185,69 @@ describe('Global transformers', () => {
       expect(this.buildPaginationSpy).to.be.called
     })
 
-    it('should return a collection object with items transformed by given transformer with arguments', () => {
-      const itemTransformerInnerSpy = sandbox.spy()
-      const itemTransformerSpy = sandbox.stub().returns(itemTransformerInnerSpy)
-      const itemTransformerOptions = { query: { term: 'bobby' } }
+    context('when an item transformer is specified with arguments', () => {
+      beforeEach(() => {
+        this.itemTransformerInnerSpy = sandbox.spy()
+        this.itemTransformerSpy = sandbox.stub().returns(this.itemTransformerInnerSpy)
+        this.itemTransformerOptions = { query: { term: 'bobby' } }
 
-      const actual = this.transformers.transformApiResponseToCollection(
-        undefined,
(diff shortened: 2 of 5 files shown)


src/apps/contacts/transformers.js
```diff
@@ -27,7 +27,7 @@ function transformContactToListItem ({
   company_sector,
   primary,
 } = {}) {
-  if (!id || !first_name || !last_name) { return }
+  if (!id || (!first_name && !last_name)) { return }
```
yeahfro: cannot see the unit test covering all of this

test/unit/apps/transformers.test.js
```diff
@@ -185,34 +185,66 @@ describe('Global transformers', () => {
...
+
+        this.transformers.transformApiResponseToCollection(
+          undefined,
+          this.itemTransformerSpy(this.itemTransformerOptions)
+        )(this.mockResponse)
+      })
+
+      it('call the item transformer with the arguments', () => {
+        expect(this.itemTransformerInnerSpy).to.be.calledTwice
+        expect(this.itemTransformerSpy).to.be.calledWith(this.itemTransformerOptions)
```
yeahfro: one `expect` per test

test/unit/apps/transformers.test.js
```diff
@@ -185,34 +185,66 @@ describe('Global transformers', () => {
...
+        this.result = this.transformers.transformApiResponseToCollection(
+          undefined,
+          this.firstItemTransformerStub,
+          this.secondItemTransformerStub
+        )(this.mockResponse)
+      })
+
+      it('should call both item transformers for each item', () => {
+        expect(this.firstItemTransformerStub).to.be.calledTwice
+        expect(this.secondItemTransformerStub).to.be.calledTwice
```
yeahfro: one `expect` per test

test/unit/apps/transformers.test.js
ztolley: hmm

test/unit/apps/transformers.test.js
ztolley: actually this time it's valid. it's not one expect per test but a test should focus on one 'thing' and here it is ony testing that both transformers are called

src/apps/contacts/transformers.js
ztolley: https://github.com/uktrade/data-hub-frontend/blob/64c7c9ebf33a1eb7278f49a5805a4d04acbd5ac2/test/unit/apps/contacts/transformers.test.js#L21

yeahfro: My point here is that there are multiple combinations which would result in a `return`. The test will still pass if the `if` statement change to `if (!first_name) { return }`

PR #1131, opened 2018-01-04
Faceted search improvements

teneightfive:
This change affects the way collection filters are displayed and interacted with. This is the first piece of work on ongoing changes to the way filters behave.

The main purpose of this work is to remove the use of dropdowns which limit to only being able to filter on one value for that facet and to set some foundations for future changes.

### Before

localhost_3001_omis_sortby created_on 3adesc 1

### After

## Questions

- I'm not convinced on the naming of the new modifiers for `c-form-group`. `--filter` feels a bit tied to the content and I'm not sure `--option-select` describers what that modifier is as such. Any suggestions would be welcome but these were initial names when prototyping the changes

## Known problems to address in future PRs

- The xhr content loads the whole content area. This causes the filters to be reloaded which removes the highlight from the last selected filter and if scrolled down a container causes it to jump back to the top. This feels unnecessary and could just update the results. 
  - This has some sideaffects as the `sort_by` value is stored in the filter form and then no longer updates. It also means conditional filters like `uk_region` in companies doesn't get display on change of `country`
- It would be good to move the _remove all filters_ to the left hand column to keep interactions with results in one consistent place. The right hand column them becomes only about the results and associated information
- Collapsing less used filters - looking at the similar pattern that is used on GOV.UK we should look at collapsing groups of filters that aren't as often used
- Group of filters - for _to_ and _from_ dates it would be good to group them into one grey container to show their more direct link
- Long list of values - this is always going to be a slight concern. We have some testing to see how people get on with them but there are a number of options we can look at in the future. Both with reducing the number based on the current data set or by allowing it to be filtered with a typeahead style box


assets/stylesheets/components/form/_multiple-choice.scss
```diff
@@ -125,12 +130,12 @@
...
   .c-multiple-choice__input {
-    @include _scale-input(14px);
+    @include _scale-input(9px);
   }
 
   .c-multiple-choice__label {
-    padding-left: round($default-spacing-unit * 3 * (14 / 19));
-    line-height: 16px;
+    padding-left: round($default-spacing-unit * 3 * (10 / 19));
+    line-height: 18px;
```
ztolley: Always makes me nervous using hard-coded numbers, is there a variable we can use or a calculation?

assets/stylesheets/components/form/_multiple-choice.scss
```diff
@@ -139,6 +139,40 @@
...
+      &:before {
+        margin-left: 8px;
+        margin-top: 8px;
+      }
+    }
+  }
+
+  .c-multiple-choice__label {
+    display: block;
+    padding-left: round($default-spacing-unit * 3 * (10 / 19) + 8);
```
ztolley: huh???

src/apps/contacts/macros.js
```diff
@@ -49,15 +49,18 @@ const contactFiltersFields = [
   },
   assign({}, globalFields.sectors, {
     name: 'company_sector',
-    initialOption: 'All sectors',
+    type: 'checkbox',
```
ztolley: There's not much left of the global field value in some of these. Do we need to create globalFilterFields separate from form fields?

src/apps/investment-projects/transformers.js
```diff
@@ -1,6 +1,5 @@
 /* eslint camelcase: 0 */
-const { get, isArray, assign, compact, pickBy } = require('lodash')
```
ztolley: Might want to hold back on this and do it in  seperate PR as I've JUST re-factored the investment transformers, plus I'd say it has nothing to do with the filter change. Would be nice to keep the PR focused on one thing.

teneightfive: I'll rebase and see what has changed. It's sort of related in terms of changes to the collection filtering functionality. If you're being pedantic it could be split out as a separate change but I think the commit message is detailed enough and feels like it fits with the set of changes in this PR.

ztolley: I still think it's not part of filtering, and what about `transformInvestmentListItemToDisableMetaLinks` ?

It'll be as easy, or easier, to do it in a separate PR with the clash.

If you at the current state of investments you'll see the formatting file and transformer file has been re-organised and now there is a `transformers` folder which contains separate files.

jim68000:
I think this is a great solution where we need multiple selects from smallish lists -> countries, regions, sectors, services etc

But longer lists (eg teams and advisers) will become even less usable than they are now, as now not even the primitive typeahead provided by browsers will work. Can we make sure this change is only applied to shorter lists?

src/apps/contacts/macros.js
teneightfive: Still contains the options, type and label so think it's fine to continue to inherit.

ztolley: The options will get phased out so they are gathered using `getOptions` instead of the properties in metadata due to the fact they are only fetched on startup

ztolley: Plus we often get the labels through a map function after building the form properties

teneightfive:
@jim68000 I'd be concerned with mixing two ways to select filters for now. I think consistency would be better.

You're right with there currently being an issue with the long lists of items for interactions for example. I've also noticed it causes a large document size too (over 2mb). My question would be, do we really need a list of _all_ advisers? Would using a filter for _my interactions_ be enough to meet the user need?

I imagine people are mainly concerned with interactions associated to themselves or is there a need to find interactions for other advisers that yourself?

ztolley:
I like the idea of 'my advisers' but also that assumes managers won't want to filter by people who work for them.... 2mb is REALLY bad. Remember people will be using this over poor 3g and worse.. remember China?

ztolley:
We need to ask the users what they want instead of guess or assume. And yes, am aware they are asking for multi select. Normally I'd say release and iterate but can we really do that with these page size?

test/acceptance/commands/clickMultipleChoiceOption.js
```diff
@@ -0,0 +1,9 @@
+/**
+ * Utility to click a specific option by textContent in a named select element
```
feedmypixel: the text needs updating `in a named multiple choice element`

test/acceptance/commands/clickMultipleChoiceOption.js
```diff
@@ -0,0 +1,9 @@
...
+ * @param name
+ * @param textContent
```
feedmypixel: and args need updating

test/acceptance/features/companies/collection.feature
```diff
@@ -45,15 +45,15 @@ Feature: View collection of companies
     When I navigate to the Company collection page
     And I filter the companies list by company
     Then the companies should be filtered by company name
-    When the company filter is cleared
+    When I clear all filters
     And I filter the companies list by sector
     Then the companies should be filtered by company sector
-    When the sector filter is cleared
+    When I clear all filters
```
feedmypixel: Maybe as a different PR these 3:

```
When I clear all filters
Then there are no filters selected
And the result count should be reset
```

 could become:
```
And I reset the collection
```

test/unit/apps/investment-projects/transformers.test.js
```diff
@@ -1,120 +0,0 @@
-const {
-  transformInvestmentProjectToListItem,
```
feedmypixel: how come this whole file has gone? Are we loosing some coverage with this going?

test/unit/apps/investment-projects/transformers/collection.test.js
```diff
@@ -31,78 +30,6 @@ describe('Investment project transformers', () => {
     })
   })
 
-  describe('#transformInvestmentListItemToHaveMetaLinks', () => {
-    beforeEach(() => {
```
feedmypixel: feels like a lot is going here, are we loosing test coverage?

test/acceptance/features/companies/collection.feature
teneightfive: Yeah I think there's a lot of improvement that can be done to all these tests. I see lots of repetition around collections across different areas but didn't want to bloat this PR too much.

test/unit/apps/investment-projects/transformers.test.js
teneightfive: So I believe this test file was missed when @ztolley converted the transformers into separate files.

The coverage for this method is being covered in `test/unit/apps/investment-projects/transformers/collection.test.js`.

test/unit/apps/investment-projects/transformers/collection.test.js
teneightfive: This method is no longer required because we're not transforming the meta items into links. There was too much interaction happening within an entity so this functionality has been removed in favour of filtering with the facets in the left hand column.

PR #1134, opened 2018-01-05
Introduce permission based tests

feedmypixel:
This work:
- Adds permission testing capability to acceptance tests
- Adds `lep` and `da` permissions jobs to CircleCi
- refactors `.circleci/config.yml` to make full use of yaml anchors
- updates documentation on running acceptance tests with specific permissions
- refactors some tests that due to username changing with different permissions
- add `//TODO` around making Acceptance tests more robust after seeing oddities when running all tests at the same time

### Associated backend PR
The PR https://github.com/uktrade/data-hub-leeloo/pull/743 contains the relevant backend work required for this frontend work

### Of note
- I have setup the following tokens on the dev api
  - `lepStaffToken`
  - `daStaffToken`
  - `ditStaffToken`
- Please remove `QA_USER_EMAIL` environment variable this is no longer in use
- The WIP [commit](https://github.com/uktrade/data-hub-frontend/pull/1134/commits/2ee30d914d0cc8f918f434b7556dd4946398c4d3) needs to be removed once the backend has been merged

screen shot 2018-01-05 at 00 12 46


test/acceptance/features/dashboard/step_definitions/dashboard.js
```diff
@@ -10,21 +10,48 @@ defineSupportCode(({ When, Then }) => {
...
-  Then(/^there should be global nav links$/, async () => {
-    await Dashboard.section.globalNav
-      .assert.visible('@companies')
-      .assert.visible('@contacts')
-      .assert.visible('@events')
-      .assert.visible('@interactionsAndServices')
-      .assert.visible('@investmentProjects')
-      .assert.visible('@ordersOmis')
-      .assert.visible('@miDashboards')
+  Then(/^there should only be (DIT|LEP|DA) staff global nav links present$/, async (staffType) => {
```
yeahfro: We could use a data table for this so that we make sure nav links only show for whatever is in the table.

test/acceptance/features/dashboard/dashboard.feature
```diff
@@ -1,17 +1,28 @@
...
-    Then there should be global nav links
+    Then there should only be DIT staff global nav links present
 
   @dashboard--support-page
   Scenario: Navigate to the support page
 
     When I navigate to the dashboard
     Then I navigate to the support page
+
+  # LEP tests
```
ztolley: 👍

test/acceptance/features/dashboard/step_definitions/dashboard.js
feedmypixel: > We could use a data table for this so that we make sure nav links only show for whatever is in the table.

Yeah sounds good. This was a bit of a late night implementation so please do improve as you see fit 👍

.circleci/config.yml
```diff
@@ -1,43 +1,200 @@
...
+    command: ./setup-uat.sh
+    environment: &data_hub_backend_env
+      AWS_DEFAULT_REGION: eu-west-2
+      AWS_ACCESS_KEY_ID: foo
+      AWS_SECRET_ACCESS_KEY: bar
+      CDMS_AUTH_URL: http://example.com
+      DATABASE_URL: postgresql://postgres@postgres/datahub
+      DATAHUB_SECRET: secret
+      DEBUG: 'True'
+      DOCUMENTS_BUCKET: baz
```
ztolley: Interesting name

.circleci/config.yml
```diff
@@ -195,76 +286,144 @@ jobs:
...
+          <<: *data_hub_base_envs
+          OAUTH2_DEV_TOKEN: *oauth2_da_staff_token
       - *docker_redis
       - *docker_elasticsearch
       - *docker_postgres
-      - <<: *docker_data_hub_leeloo
+      - *docker_mock_sso
+      - <<: *docker_data_hub_backend
+        environment:
+          <<: *data_hub_backend_env
```
ztolley: If this is only used in one place then why extend/overwrite the token?

.circleci/config.yml
```diff
@@ -1,43 +1,200 @@
-## TODO discover why the wavy package is causing us problems and why we need to yarn add wavy with a fresh install
-
 version: 2
 aliases:
-  - &node_version         node:8.5.0
-  - &qa_user_email        circleci@datahub.com
-  - &oauth2_dev_token     topSecretSquirrel
+  - &node_version               node:8.5.0
```
ztolley: Is it possible to break the file into multiple for the different jobs, may make it easier to read and review, future PR though.

README.md
```diff
@@ -367,6 +374,23 @@ Data hub uses [Nightwatch.js](http://nightwatchjs.org), [nightwatch-cucumber](ht
 #### Running acceptance tests
 For information on [cucumber-js](https://github.com/cucumber/cucumber-js) tags please see the `nightwatch-cucumber` docs [executing-individual-feature-files-or-scenarios](http://mucsi96.github.io/nightwatch-cucumber/#executing-individual-feature-files-or-scenarios)
 
+#### Running tests with specific user permissions
+To run tests against a specific user permissions type: 
+- Change your `OAUTH2_DEV_TOKEN` environment variable to use one of:
```
ztolley: I wonder if the env file could contain all the tokens and the test runner pick the one to use, this would save having to change the token in your config when you want to switch to test different users.

feedmypixel: Interesting idea, although I'm not clear on how that would work with [config](https://github.com/uktrade/data-hub-frontend/blob/develop/config/index.js#L66)

.circleci/config.yml
feedmypixel: Great spot, this isn't needed anymore 👍

.circleci/config.yml
feedmypixel: why thank in you 🙇

.circleci/config.yml
feedmypixel: 🤔  I'm not sure that would help. It would be quite a cognitive load to jump between files. I also don't think circleci would work this way or if you can `import` in yaml?

jim68000: In theory you can, but I've never been able to get it to work. The standard defines it, but then leaves how to represent a filepath up to the implementor, iirc

PR #1135, opened 2018-01-05
updating NodeJs docs

feedmypixel:
This work adds information around the various places the node version needs updating, when it is updated.


feedmypixel:
fyi @jim68000 ☝️

PR #1137, opened 2018-01-05
Replace individual collection templates with shared one

teneightfive:
At the moment each collection view uses a template within the sub-app.

The majority of these templates is the same so this change removes the
need for duplicate template code in favour of one shared template.

Any differences are then set within the controller and passed to the
view.

## Possible todos (?)

- [ ] Find way to set summary actions without having to inherit from collection template and duplicate `Collection` macro call
- [x] Find way to set `highlightTerm` without having to inherit collection template and duplicate `Collection` macro call


test/unit/apps/interactions/middleware/collection.test.js
```diff
@@ -60,8 +60,8 @@ describe('interaction collection middleware', () => {
     })
 
     it('should create a collection of interactions', () => {
```
yeahfro: one `expect` per test?

teneightfive: A good point, but felt slightly out of scope for this change. Our test patterns are still quite different across different areas.

Feels like we need to revisit that conversation about agreed patterns and maybe revisit old tests as one change?

ztolley: It's not strictly one 'it' per test, but a test should test one 'thing' at a time.. sometimes that means multiple it's but not unrelated.

PR #1138, opened 2018-01-05
DH-1266 Investment type not displaying in evaluations screen

yeahfro:
DH-1266

This change shows the correct investment type on the evaluations screen. Including acceptance tests.

# Before
screen shot 2018-01-05 at 23 23 50

# After
screen shot 2018-01-05 at 23 22 27

### DIFF
--- a/src/apps/investment-projects/services/formatting.js
+++ b/src/apps/investment-projects/services/formatting.js
@@ -3,7 +3,6 @@
 const moment = require('moment')
 const format = require('date-fns/format')
 const { compact, mapValues, get, isPlainObject, isNull } = require('lodash')
-const metadataRepository = require('../../../lib/metadata')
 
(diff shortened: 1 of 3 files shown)


src/apps/investment-projects/services/formatting.js
```diff
@@ -265,39 +264,24 @@ function transformUKCompany (data) {
...
-  function approvedInvestmentText (investments, defaultText) {
-    return investments.reduce((displayText, investmentDetail) => {
-      return investmentDetail.isApproved ? investmentDetail.displayText : displayText
-    }, defaultText)
-  }
-
+function transformInvestmentFDIForView ({
+  investment_type,
+  fdi_type,
+  investor_company,
```
ztolley: an ugly mix of cases but yeah I can't see an alternative.

PR #1140, opened 2018-01-08
Fix/put investment team

ztolley:
DH-1417

Change the investment team members update code to use the new *PUT* method, recently added to the investment team members endpoint.

While working on this also had to update the error handling code and then finally updated the middleware tests to extend their reach to the repository.


ztolley:
Closing to replace with a different PR in the morning that refactors some of this code to fix another bug to do with error handling.

PR #1142, opened 2018-01-09
Fix/fix investment team edit

ztolley:
DH-1417

* Fix bug when LEP's saved investment team changes it would wipe out all team members
* Switch to use new PUT endpoint in the API to save team members
* Fix bug when a user had an error editing an investment team the form would not re-render properly and loose adviser selections
* Refactor investment team routing to be simpler and follow pattern of saving and passing errors to common render controller
* Refactor investment team middleware to fix bugs mentioned above and better structured
* Improved unit testing related to investment team middleware

### DIFF
--- a/src/apps/investment-projects/middleware/forms/team-members.js
+++ b/src/apps/investment-projects/middleware/forms/team-members.js
@@ -1,24 +1,38 @@
-const { zipWith, get, isArray, isString, assign } = require('lodash')
+const { zipWith, get, assign, forOwn, isEmpty, castArray } = require('lodash')
+
 const { getAdvisers } = require('../../../adviser/repos')
 const { filterActiveAdvisers } = require('../../../adviser/filters')
 const { updateInvestmentTeamMembers } = require('../../repos')
 const { teamMembersLabels } = require('../../labels')
 const { transformObjectToOption } = require('../../../transformers')
 
-// Transform the form posted to an array of objects to save
-// and filter out any blanks
-function transformDataToTeamMemberArray (body) {
-  let teamMembers = []
-  if (isArray(body.adviser)) {
-    teamMembers = zipWith(body.adviser, body.role, (adviser, role) => ({ adviser, role }))
-  } else if (isString(body.adviser)) {
-    teamMembers = [{ adviser: body.adviser, role: body.role }]
-  }
-  return teamMembers.filter(member => member.adviser.length)
+function transformFormToTeamMemberArray ({ adviser, role }) {
+  const advisersArray = castArray(adviser)
+  const rolesArray = castArray(role)
+
+  const teamMembers = zipWith(advisersArray, rolesArray, (adviser, role) => ({ adviser, role }))
+  return teamMembers.filter(member => !isEmpty(member.adviser))
 }
 
-function getTeamMemberItem ({ teamMember, advisers }) {
-  const adviser = get(teamMember, 'adviser.id')
+function transformTeamMemberArrayToFields (teamMemberArray, advisers) {
+  return teamMemberArray.map((teamMember) => getTeamMemberField({ teamMember, advisers }))
+}
+
+function transformInvestmentTeamMemberstoTeamMemberArray (investmentData = {}) {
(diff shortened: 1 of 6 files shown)


src/apps/investment-projects/middleware/forms/team-members.js
```diff
@@ -1,24 +1,38 @@
...
-function transformDataToTeamMemberArray (body) {
-  let teamMembers = []
-  if (isArray(body.adviser)) {
-    teamMembers = zipWith(body.adviser, body.role, (adviser, role) => ({ adviser, role }))
-  } else if (isString(body.adviser)) {
-    teamMembers = [{ adviser: body.adviser, role: body.role }]
-  }
-  return teamMembers.filter(member => member.adviser.length)
+function transformFormToTeamMemberArray ({ adviser, role }) {
+  const advisersArray = isArray(adviser) ? adviser : [adviser]
```
yeahfro: can we use a pattern similar to [transforming `related_programmes` and `teams` in `events`](https://github.com/uktrade/data-hub-frontend/blob/develop/src/apps/events/transformers.js#L185)?

src/apps/investment-projects/middleware/forms/team-members.js
```diff
@@ -32,59 +46,86 @@ function getTeamMemberItem ({ teamMember, advisers }) {
...
+function transformErrorResponseToFormErrors (error) {
+  const messages = {}
 
-    res.locals.form = assign({}, res.locals.form, {
-      labels: teamMembersLabels.edit,
-      fields: { teamMembers },
-      buttonText: 'Save',
-      returnLink: `/investment-projects/${investmentData.id}/team`,
+  if (isArray(error)) {
+    for (let pos = 0; pos < error.length; pos += 1) {
```
yeahfro: can we `map` this and use the second `index` parameter instead?

src/apps/investment-projects/middleware/forms/team-members.js
```diff
@@ -32,59 +46,86 @@ function getTeamMemberItem ({ teamMember, advisers }) {
...
+async function populateTeamEditForm (req, res, next) {
+  try {
+    const token = req.session.token
+    const investmentId = req.params.investmentId
+
+    const { results: advisers } = await getAdvisers(token)
+
+    const teamMembers = transformInvestmentTeamMemberstoTeamMemberArray(res.locals.investmentData)
+
+    const fields = transformTeamMemberArrayToFields(teamMembers, advisers)
```
yeahfro: my preference would be to concatenate this rather than push

src/apps/investment-projects/middleware/forms/team-members.js
ztolley: I'm not returning an array I'm adding properties to an object.

src/apps/investment-projects/middleware/forms/team-members.js
ztolley: Why? I am a fan of not mutating but when the thing is created within the function and only has one purpose then push is simpler no? PLus it allows use of a const instead of assigning the value back on itself

src/apps/investment-projects/middleware/forms/team-members.js
ztolley: its different to them but there is stuff there I can use to tidy it up a little

src/apps/investment-projects/middleware/forms/team-members.js
ztolley: I'll use foreach

src/apps/investment-projects/middleware/forms/team-members.js
```diff
@@ -1,4 +1,4 @@
-const { zipWith, get, isArray, assign, forOwn, isEmpty } = require('lodash')
+const { zipWith, get, assign, forOwn, isEmpty, castArray } = require('lodash')
```
yeahfro: Right on the edge of becoming multi line..?

ztolley: Yeah we should agree how many before we split

PR #1143, opened 2018-01-09
DH-1420 Fix permissions for LEPs to see interactions belonging to investment

yeahfro:
DH-1420

# Issue

LEPs cannot see interactions that belong to an investment project.

# Reason

After investigation I noticed that there is a global check of permissions. If a user does not have access to `interactions` at a global level then they cannot view `interactions`. This does not consider that the user may have a more granular permission allowing access to the `interactions` that belong to an entity.

# Fix

The change includes:
- Moving each global nav item to each app which will move the permissions into one place (for an app)
- Each app is responsible for checking permissions


src/apps/globalNavItems.js
```diff
@@ -0,0 +1,23 @@
...
+const APP_GLOBAL_NAV_ITEMS = compact(subApps.map(subAppDir => {
+  const constantsPath = path.join(__dirname, subAppDir, 'constants.js')
+  if (!fs.existsSync(constantsPath)) {
+    return null
+  }
+
+  return require(constantsPath).GLOBAL_NAV_ITEM
+}))
+
+const GLOBAL_NAV_ITEMS = concat(sortBy(APP_GLOBAL_NAV_ITEMS, (globalNavItem) => globalNavItem.order), {
```
ztolley: maybe do the sort and assign to a var before this bit to make the contact easier to read

PR #1147, opened 2018-01-10
DH-1384 Acceptance tests around LEP/DA collections

yeahfro:
DH-1384

This change adds acceptance tests for LEP/DA visibility of collections and asserting a 403 when they do not have permission to view a resource.

### DIFF
--- a/test/acceptance/features/location/page-objects/Location.js
+++ b/test/acceptance/features/location/page-objects/Location.js
@@ -70,4 +70,11 @@ module.exports = {
       ],
     },
   },
+  commands: [
+    {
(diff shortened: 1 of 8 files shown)


test/acceptance/features/location/page-objects/Location.js
```diff
@@ -70,4 +70,16 @@ module.exports = {
       ],
     },
   },
+  commands: [
+    {
+      getHeading (el, text) {
+        return getSelectorForElementWithText(
+          text,
```
ztolley: Maybe reduce it down so it's not spread on so many lines

PR #1149, opened 2018-01-10
Refactor invest transformers

ztolley:
Before starting on adding new fields to the investment project I wanted to combine investment formatters and transformers: they both did the same thing and the transformer functions were called transformxx

After that, I split the transformer file into separate files for the different parts of investment to make it easier to maintain.


src/apps/investment-projects/transformers/index.js
```diff
@@ -2,7 +2,10 @@
 
 const moment = require('moment')
 const format = require('date-fns/format')
-const { compact, mapValues, get, isPlainObject, isNull } = require('lodash')
+const queryString = require('query-string')
+const { assign, compact, get, isArray, isPlainObject, isNull, mapValues, pickBy } = require('lodash')
```
yeahfro: Feeling this would read better over multiple lines

src/apps/investment-projects/transformers/index.js
```diff
@@ -383,6 +386,73 @@ function transformTeamMembersForView ({ adviser, role }) {
...
+  }
+}
+
+function transformInvestmentListItemToHaveMetaLinks (query = {}) {
+  return function transformListItem (item) {
+    if (!isArray(item.meta)) { return item }
+
+    item.meta.forEach(metaItem => {
+      const name = metaItem.name
+      const itemQuery = Object.assign(
```
yeahfro: `assign`

PR #1150, opened 2018-01-10
DH-1381 Acceptance tests around contact permissions

yeahfro:
DH-1381

This change is for the following:
- adding tests around the contact local nav for each user type
- adding tests around contact documents for each user type
- adding tests around contact interactions for each user type - note this fixes a bug which is the interactions are not loading underneath entities


test/acceptance/features/omis/collection.feature
```diff
@@ -4,5 +4,5 @@ Feature: OMIS collection
   @omis-collection--lep @lep
   Scenario: Navigate to OMIS shows 403 for LEPs
 
-    Given I navigate to /omis
+    Given I navigate directly to /omis
```
ztolley: why?

test/acceptance/features/contacts/documents-details.feature
```diff
@@ -0,0 +1,28 @@
+@contact-documents-details  @details
```
ztolley: Why is it document details and not just documents?

PR #1152, opened 2018-01-10
DH-1383 Acceptance tests around search

yeahfro:
DH-1383

Adding acceptance tests around:
- search tabs as different user types
- asserting 403 when searching directly against entities as different user types

### DIFF
--- a/test/acceptance/features/search/search.feature
+++ b/test/acceptance/features/search/search.feature
@@ -42,3 +42,32 @@ Feature: Search
       | Orders              |
     And the Companies search tab has 1 results
     And I can view the company in the search results
+
+  @search--companies--lep @lep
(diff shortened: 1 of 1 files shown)


test/acceptance/features/search/search.feature
```diff
@@ -42,3 +42,62 @@ Feature: Search
       | Orders              |
     And the Companies search tab has 1 results
     And I can view the company in the search results
+
+  @search--companies--lep @lep
+  Scenario: Search companies as a LEP
```
ztolley: Out of curiosity, why do we space the tests with a line between scenario and the first action? I only ask because in cases like this, when we have lots of small tests they are harder to separate.

PR #1158, opened 2018-01-12
Fix investment transformer tests and improve investment details transformer

ztolley:
The project transformation tests for investment had many issues, this fix:
* Fixes description for many of the tests that were incorrect due to cut and paste errors.
* Reformat tests for the brief summary to use context
* Adds tests for investment project summary in details page

Also refactored the transformer for investment project details to clean it up in preparation for adding another field.

### DIFF
--- a/src/apps/investment-projects/transformers/project.js
+++ b/src/apps/investment-projects/transformers/project.js
@@ -1,5 +1,5 @@
 /* eslint-disable camelcase */
-const { get, isPlainObject, mapValues } = require('lodash')
+const { castArray, get, isEmpty, isPlainObject, mapValues } = require('lodash')
 const format = require('date-fns/format')
 const moment = require('moment')
 
@@ -101,25 +101,42 @@ function transformFromApi (body) {
   return Object.assign({}, body, formatted)
 }
 
-function transformInvestmentDataForView (data) {
-  if (!isPlainObject(data)) { return }
-
-  const businessActivities = data.business_activities.slice()
-  if (data.other_business_activity) {
-    businessActivities.push({ name: data.other_business_activity })
+function transformInvestmentDataForView ({
+  business_activities,
+  other_business_activity,
+  investment_type,
+  fdi_type,
+  sector,
+  client_contacts,
+  description,
+  anonymous_description,
+  investor_company,
+  investor_type,
+  level_of_involvement,
+  specific_programme,
+  estimated_land_date,
+} = {}) {
+  const businessActivities = castArray(business_activities).map(i => i.name)
+  if (!isEmpty(other_business_activity)) {
+    businessActivities.push(other_business_activity)
   }
 
-  return Object.assign({}, data, {
-    investor_company: {
-      name: data.investor_company.name,
-      url: `/companies/${data.investor_company.id}`,
-    },
-    investment_type: getInvestmentTypeDetails(data.investment_type, data.fdi_type),
-    sector: get(data, 'sector.name', null),
-    business_activities: businessActivities.map(i => i.name).join(', '),
-    client_contacts: data.client_contacts.map(i => i.name).join(', '),
-    estimated_land_date: data.estimated_land_date ? moment(data.estimated_land_date).format('MMMM YYYY') : null,
-  })
+  return {
+    investor_company: isPlainObject(investor_company) ? {
+      name: get(investor_company, 'name'),
+      url: `/companies/${investor_company.id}`,
+    } : null,
+    investment_type: getInvestmentTypeDetails(investment_type, fdi_type),
+    sector,
+    business_activities: businessActivities.join(', '),
+    client_contacts: castArray(client_contacts).map(i => i.name).join(', '),
+    description,
+    anonymous_description,
+    investor_type,
+    level_of_involvement,
+    specific_programme,
+    estimated_land_date: !isEmpty(estimated_land_date) ? moment(estimated_land_date, 'YYYY-MM-DD').format('MMMM YYYY') : null,
+  }
 }
 
 function transformBriefInvestmentSummary (data) {
@@ -128,7 +145,6 @@ function transformBriefInvestmentSummary (data) {
   const investorCompany = data.investor_company
   const competitorCountries = data.competitor_countries || []
   const regionLocations = data.uk_region_locations || []
-  const date = moment(data.estimated_land_date, 'YYYY-MM-DD')
 
   return {
     sector: get(data, 'sector.name', null),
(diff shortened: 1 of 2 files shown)


src/apps/investment-projects/transformers/project.js
```diff
@@ -101,25 +101,47 @@ function transformFromApi (body) {
...
+    investment_type: getInvestmentTypeDetails(investment_type, fdi_type),
+    sector,
+    business_activities: businessActivities.join(', '),
+    client_contacts: castArray(client_contacts).map(i => i.name).join(', '),
+    description,
+    anonymous_description,
+    investor_type,
+    level_of_involvement,
+    specific_programme,
+    estimated_land_date: estimatedDate.isValid() ? estimatedDate.format('MMMM YYYY') : null,
```
teneightfive: Do we need to make sure its a valid date? Can't we be sure the API will be returning a date format if it has a value?

ztolley: This is the existing code simply moved and not changed. I think it's there as much to check for null or empty strings but doesn't hurt to check the date too

ztolley: If you check FormatDate filter in filters.js you'll see that we always check the date is valid when formatting. I wanted to be consistent and think it's good practice to be safe.

ztolley: Will see if I can move it into the njk file though, after all you can pass theformat into the filter.

ztolley: I'd move it to the view layer but that doesn't currently support setting the date format

src/apps/investment-projects/transformers/project.js
```diff
@@ -101,25 +101,47 @@ function transformFromApi (body) {
...
-      name: data.investor_company.name,
-      url: `/companies/${data.investor_company.id}`,
-    },
-    investment_type: getInvestmentTypeDetails(data.investment_type, data.fdi_type),
-    sector: get(data, 'sector.name', null),
-    business_activities: businessActivities.map(i => i.name).join(', '),
-    client_contacts: data.client_contacts.map(i => i.name).join(', '),
-    estimated_land_date: data.estimated_land_date ? moment(data.estimated_land_date).format('MMMM YYYY') : null,
-  })
+  const estimatedDate = moment(estimated_land_date, 'YYYY-MM-DD')
```
teneightfive: `estimatedLandDate`? That's what it's actually mapping to.

src/apps/investment-projects/transformers/project.js
```diff
@@ -101,25 +101,47 @@ function transformFromApi (body) {
...
+  const investorCompanyLink = investor_company ? {
+    id: investor_company.id,
+    name: get(investor_company, 'name'),
+    url: `/companies/${investor_company.id}`,
+  } : null
+
+  return {
+    investor_company: investorCompanyLink,
+    investment_type: getInvestmentTypeDetails(investment_type, fdi_type),
+    sector,
```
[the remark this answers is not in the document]
ztolley: as said there, I  am very aware of the pattern, but in this case order is important

src/apps/investment-projects/transformers/project.js
```diff
@@ -101,25 +101,47 @@ function transformFromApi (body) {
...
+  const estimatedDate = moment(estimated_land_date, 'YYYY-MM-DD')
+
+  const investorCompanyLink = investor_company ? {
+    id: investor_company.id,
+    name: get(investor_company, 'name'),
+    url: `/companies/${investor_company.id}`,
+  } : null
+
+  return {
+    investor_company: investorCompanyLink,
```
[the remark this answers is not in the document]
ztolley: I know. I also tend to get asked by Ben to use this pattern to make the code easier to read.

src/apps/investment-projects/transformers/project.js
```diff
@@ -101,25 +101,47 @@ function transformFromApi (body) {
...
-    investment_type: getInvestmentTypeDetails(data.investment_type, data.fdi_type),
-    sector: get(data, 'sector.name', null),
-    business_activities: businessActivities.map(i => i.name).join(', '),
-    client_contacts: data.client_contacts.map(i => i.name).join(', '),
-    estimated_land_date: data.estimated_land_date ? moment(data.estimated_land_date).format('MMMM YYYY') : null,
-  })
+  const estimatedDate = moment(estimated_land_date, 'YYYY-MM-DD')
+
+  const investorCompanyLink = investor_company ? {
+    id: investor_company.id,
```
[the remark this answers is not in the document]
ztolley: who knows, it was there when I got here :) Will remove it and see what happens

src/apps/investment-projects/transformers/project.js
ztolley: It's a copy of the estimated_land_date as a moment object to checks it's validity

src/apps/investment-projects/transformers/project.js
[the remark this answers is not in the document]
teneightfive: It seems to have only been added with this change?

src/apps/investment-projects/transformers/project.js
```diff
@@ -101,25 +101,42 @@ function transformFromApi (body) {
...
+    investment_type: getInvestmentTypeDetails(investment_type, fdi_type),
+    sector,
+    business_activities: businessActivities.join(', '),
+    client_contacts: castArray(client_contacts).map(i => i.name).join(', '),
+    description,
+    anonymous_description,
+    investor_type,
+    level_of_involvement,
+    specific_programme,
+    estimated_land_date: isEmpty(estimated_land_date) ? null : moment(estimated_land_date, 'YYYY-MM-DD').format('MMMM YYYY'),
```
teneightfive: Do you need to define the date format when creating the moment object? It wasn't before.

src/apps/investment-projects/transformers/project.js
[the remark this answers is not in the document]
ztolley: Then I cut and paste it from somewhere else, I've removed it now.

src/apps/investment-projects/transformers/project.js
teneightfive: Can we also change the ternary operator so the default is `null`:

```js
!isEmpty(estimated_land_date) ? moment(estimated_land_date, 'YYYY-MM-DD').format('MMMM YYYY') : null,
```

ztolley: It was and if you dont you get a deprecation message telling you how it falls back to the default date format but should be specified.

teneightfive: I see. That's fine then.

PR #1160, opened 2018-01-12
DH-1357 Add actual land date to edit investment form and details view

ztolley:
Pretty much as it says in the title :) 

screencapture-localhost-3000-investment-projects-4fd1a994-74a0-4ebb-890b-73bd6c780de1-details-1515780349121
screencapture-localhost-3000-investment-projects-4fd1a994-74a0-4ebb-890b-73bd6c780de1-edit-details-1515780330223

### DIFF
--- a/src/apps/investment-projects/transformers/project.js
+++ b/src/apps/investment-projects/transformers/project.js
@@ -118,6 +118,7 @@ function transformInvestmentDataForView ({
   level_of_involvement,
   specific_programme,
   estimated_land_date,
+  actual_land_date,
 } = {}) {
   const businessActivities = castArray(business_activities).map(i => i.name)
   if (!isEmpty(other_business_activity)) {
@@ -139,6 +140,10 @@ function transformInvestmentDataForView ({
     level_of_involvement,
     specific_programme,
     estimated_land_date: !isEmpty(estimated_land_date) ? moment(estimated_land_date, 'YYYY-MM-DD').format('MMMM YYYY') : null,
+    actual_land_date: !isEmpty(actual_land_date) ? {
+      type: 'date',
+      name: actual_land_date,
+    } : null,
   }
 }
 
--- a/src/apps/investment-projects/transformers/project.js
+++ b/src/apps/investment-projects/transformers/project.js
@@ -1,9 +1,10 @@
 /* eslint-disable camelcase */
-const { castArray, get, isEmpty, isPlainObject, mapValues } = require('lodash')
+const { assign, castArray, get, isEmpty, isPlainObject, mapValues } = require('lodash')
 const format = require('date-fns/format')
 const moment = require('moment')
 
 const { formatCurrency, getInvestmentTypeDetails } = require('./shared')
+const { transformDateObjectToDateString } = require('../../transformers')
 
 function transformToApi (body) {
   if (!isPlainObject(body)) { return }
@@ -60,7 +61,9 @@ function transformToApi (body) {
     '01',
   ].join('-')
 
-  return Object.assign({}, body, formatted)
+  formatted['actual_land_date'] = transformDateObjectToDateString('actual_land_date')(body)
+
+  return assign({}, body, formatted)
 }
 
 function transformFromApi (body) {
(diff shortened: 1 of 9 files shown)


teneightfive:
Can we do something smart with the difference? Maybe saying whether it was x months early or late?

Maybe even combine them in the view?

src/apps/investment-projects/transformers/project.js
```diff
@@ -98,28 +101,56 @@ function transformFromApi (body) {
...
-    sector: get(data, 'sector.name', null),
-    business_activities: businessActivities.map(i => i.name).join(', '),
-    client_contacts: data.client_contacts.map(i => i.name).join(', '),
-    estimated_land_date: data.estimated_land_date ? moment(data.estimated_land_date).format('MMMM YYYY') : null,
-  })
+  const estimatedDate = moment(estimated_land_date, 'YYYY-MM-DD')
+  const actualDate = moment(actual_land_date, 'YYYY-MM-DD')
+
+  const investorCompanyLink = investor_company ? {
+    id: investor_company.id,
```
teneightfive: What's the need for the addition of the `id` property?

src/apps/investment-projects/transformers/project.js
```diff
@@ -98,28 +101,56 @@ function transformFromApi (body) {
...
+    name: get(investor_company, 'name'),
+    url: `/companies/${investor_company.id}`,
+  } : null
+
+  return {
+    investor_company: investorCompanyLink,
+    investment_type: getInvestmentTypeDetails(investment_type, fdi_type),
+    sector,
```
teneightfive: Can we list all shorthand at the top of the object definition to help readability. It's a pattern we've been using elsewhere.

src/apps/investment-projects/transformers/project.js
```diff
@@ -98,28 +101,56 @@ function transformFromApi (body) {
...
+    business_activities: businessActivities.join(', '),
+    client_contacts: castArray(client_contacts).map(i => i.name).join(', '),
+    description,
+    anonymous_description,
+    investor_type,
+    level_of_involvement,
+    specific_programme,
+    estimated_land_date: estimatedDate.isValid() ? estimatedDate.format('MMMM YYYY') : null,
```
teneightfive: We should probably be doing the date formatting using the nunjucks filters. Or if that's not always possible, using the date formats from the config file.

src/apps/investment-projects/transformers/project.js
```diff
@@ -98,28 +101,56 @@ function transformFromApi (body) {
...
+    actual_land_date: actualDate.isValid() ? {
+      type: 'date',
+      name: actual_land_date,
```
teneightfive: How does this get the value? Only seems to be passing a type and name. 

Is it also overkill to check that it's a valid date? We should be able to assume that if the API returns a value it is a date format?

ztolley: I'll try without the check but currently it's passes the type and name which is what the date formatter uses as value (I was confused with it using name ). Alternatively I could format it here so the result would be a valid string or null

src/apps/investment-projects/transformers/project.js
ztolley: Nope. Sorry. The return needs to be in order

ztolley: And Trust me I know the pattern ;)

src/apps/investment-projects/transformers/project.js
ztolley: Excellent question. It was there before?

ztolley:
I like the idea about differences, but for now.. we just need to get the field in

src/apps/investment-projects/transformers/project.js
ztolley: This passes the type and the value is sent to the name property. This seems to be a bug in the key value table macro as it differs from the metadata item usage that correctly uses value.  We need to do a follow up to change key value table to accept the date value in the 'value' property.

jim68000:
Yeah, there's no user need for measuring the differences.

PR #1174, opened 2018-01-19
DH-1447 : Fix issue handling empty expected land dates

ztolley:
Previous if no land date was provided an empty string was sent to the server which parsed it and turned it into a date in Jan 1970, if there is no expected land date the server expects a null value. Updated the form handler to accommodate this, and this makes it consistent with actual land date.

### DIFF
--- a/src/apps/investment-projects/transformers/project.js
+++ b/src/apps/investment-projects/transformers/project.js
@@ -55,11 +55,15 @@ function transformToApi (body) {
     return { id: value }
   })
 
-  formatted['estimated_land_date'] = [
-    body['estimated_land_date_year'],
-    body['estimated_land_date_month'],
-    '01',
-  ].join('-')
+  if (body['estimated_land_date_year'] || body['estimated_land_date_month']) {
+    formatted['estimated_land_date'] = [
+      body['estimated_land_date_year'],
+      body['estimated_land_date_month'],
+      '01',
+    ].join('-')
+  } else {
+    formatted['estimated_land_date'] = null
+  }
 
   formatted['actual_land_date'] = transformDateObjectToDateString('actual_land_date')(body)
 
--- a/test/unit/apps/investment-projects/transformers/project.test.js
+++ b/test/unit/apps/investment-projects/transformers/project.test.js
@@ -727,7 +727,7 @@ describe('Investment project transformers', () => {
         })
       })
 
-      it('should return a malformed date which will be picked up by the API validator', () => {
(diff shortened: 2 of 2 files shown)


src/apps/investment-projects/transformers/project.js
```diff
@@ -55,11 +55,15 @@ function transformToApi (body) {
     return { id: value }
   })
 
-  formatted['estimated_land_date'] = [
-    body['estimated_land_date_year'],
-    body['estimated_land_date_month'],
-    '01',
-  ].join('-')
+  if (!body['estimated_land_date_year'] && !body['estimated_land_date_month']) {
```
teneightfive: Can we use dot notation for these? They don't need to use bracket notation.

test/unit/apps/investment-projects/transformers/project.test.js
```diff
@@ -744,4 +744,41 @@ describe('Investment project transformers', () => {
...
+  })
+
+  context('when called with a partial estimated land date', () => {
+    beforeEach(() => {
+      this.result = transformToApi({
+        estimated_land_date_year: '2016',
+      })
+    })
+
+    it('should return a malformed date which will be picked up by the API validator', () => {
```
teneightfive: Don't think you need `which will be picked up by the API validator`. This test doesn't care about where the data goes.

teneightfive: Would be nice to be a bit more flexible with this too. Like if they enter December it uses December of this year...

src/apps/investment-projects/transformers/project.js
teneightfive: This it would also be better to switch this if block. So check for a value and do something with it, otherwise set it to null.

ztolley: It was that way before but changing the order made it easier to set the if.

test/unit/apps/investment-projects/transformers/project.test.js
ztolley: hmm

ztolley: Lets do that later and think about doing it on the client side for any dates, so when you enter a month it auto fills the year with the year that month next happens in, and see what users think when they try it. I've had issues before where we decided to auto file the date field and the user didn't realise, they had their head down and just kept typing (though you could make it so the whole value is selected by default when the field becomes active so it accommodates that)

PR #1175, opened 2018-01-21
DH-1432 Prevent moving to the next stage when actual land date is incomplete

yeahfro:
DH-1432

Current issues is that the investment project `Move to x stage` button is enabled when `actual_land_date` has not been set.

This change does the following:
- disables the button when the `actual_land_date` has not been set
- alters the feedback in the callout so that it is clear what needs to be done with links e.g. `Field x in y form`

screen shot 2018-01-21 at 22 09 46

### DIFF
--- a/src/apps/investment-projects/helpers.js
+++ b/src/apps/investment-projects/helpers.js
@@ -1,48 +1,32 @@
-const { get, indexOf, forEach, uniq } = require('lodash')
+const { assign, map, sortBy, compact } = require('lodash')
 
 const config = require('../../../config')
+const labels = require('./labels')
+const logger = require('../../../config/logger')
 
-const formFields = {
-  value: [
-    'total_investment',
-    'government_assistance',
-    'number_new_jobs',
-    'number_safeguarded_jobs',
-    'r_and_d_budget',
-    'non_fdi_r_and_d_budget',
-    'new_tech_to_uk',
-    'export_revenue',
-    'client_cannot_provide_total_investment',
-    'client_cannot_provide_foreign_investment',
-    'fdi_value',
-  ],
-  requirements: [
-    'strategic_drivers',
-    'client_considering_other_countries',
-    'uk_region_locations',
-    'site_decided',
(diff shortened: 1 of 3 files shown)


src/apps/investment-projects/views/_layout.njk
```diff
@@ -62,6 +62,7 @@
               <p>You have added all required information to complete this stage.</p>
             {% endif %}
           {% else %}
+            <p>The investment cannot be moved to the {{ investmentStatus.nextStage.name }} stage</p>
```
ztolley: Do we really need this, isn't it saying the same thing twice?

assets/stylesheets/components/_messages.scss
```diff
@@ -24,6 +24,10 @@
     margin-bottom: 0;
   }
 
+  p {
+    margin-bottom: $default-spacing-unit;
```
teneightfive: We should try and avoid adding bottom margins to everything like this. If the paragraph is the last element you'll get double spacing on the bottom of the message box.

Would be better to use sibling selectors to apply spacing. The [lobotomized owl](https://alistapart.com/article/axiomatic-css-and-lobotomized-owls) article is a good read when thinking about how to space elements.

I have a feeling the declaration on line 35 should be sorting out spacing so maybe look at tweaking that.

ztolley: Do we even need this though if my first comment it applicable?

src/apps/investment-projects/helpers.js
```diff
@@ -67,28 +77,19 @@ const toCompleteStageMessages = {
...
+      })
+    }
 
-  forEach(incompleteFields, (incompleteField) => {
-    forEach(formContents, (formFields, key) => {
-      if (indexOf(formFields, incompleteField) !== -1) {
-        forms.push(key)
-      }
-    })
+    logger.log('error', `Could not find form field link for ${incompleteField}`)
```
teneightfive: What about returning the error first?

PR #1180, opened 2018-01-23
Filter OMIS sector list

teneightfive:
During the create journey the list of sectors was not being filtered
for disabled values.

This change updates the controllers within the create journey to use
the new `getOptions` method to retrieve options from metadata
endpoints.

This method handle filtering out disabled items in the list.

### DIFF
--- a/test/unit/apps/omis/apps/create/controllers/market.test.js
+++ b/test/unit/apps/omis/apps/create/controllers/market.test.js
@@ -1,22 +1,13 @@
-const FormController = require('~/src/apps/omis/controllers/form')
+const { assign } = require('lodash')
 
-const Controller = proxyquire('~/src/apps/omis/apps/create/controllers/market', {
-  '../../../../../lib/metadata': {
-    omisMarketOptions: [{
-      id: '1',
-      name: 'One',
-      disabled_on: null,
-    }, {
-      id: '2',
-      name: 'Two',
-      disabled_on: '2010-01-01T00:00:00',
-    }, {
-      id: '3',
-      name: 'Three',
-      disabled_on: null,
-    }],
-  },
-})
+const { apiRoot } = require('~/config')
+const FormController = require('~/src/apps/omis/controllers/form')
+const Controller = require('~/src/apps/omis/apps/create/controllers/market')
+
+const marketOptionsMock = [
+  { id: '9999', name: 'United Kingdom' },
+  { id: '8888', name: 'Germany' },
+]
 
 describe('OMIS create market controller', () => {
   beforeEach(() => {
@@ -25,34 +16,57 @@ describe('OMIS create market controller', () => {
   })
 
   describe('configure()', () => {
-    beforeEach(() => {
-      this.reqMock = Object.assign({}, globalReq, {
-        form: {
-          options: {
-            fields: {
-              primary_market: {},
+    context('when getOptions returns 200', () => {
+      beforeEach(async () => {
+        nock(apiRoot)
+          .get('/metadata/omis-market/')
+          .reply(200, marketOptionsMock)
+
+        this.reqMock = assign({}, globalReq, {
+          form: {
+            options: {
+              fields: {
+                primary_market: {},
+              },
             },
           },
-        },
+        })
+
+        sandbox.spy(FormController.prototype, 'configure')
+        await this.controller.configure(this.reqMock, globalRes, this.nextSpy)
       })
 
-      sandbox.spy(FormController.prototype, 'configure')
+      it('should set list of markets dynamically', () => {
+        expect(this.reqMock.form.options.fields.primary_market.options).to.deep.equal([
+          { value: '8888', label: 'Germany' },
+          { value: '9999', label: 'United Kingdom' },
+        ])
+      })
+
+      it('should call parent configure method', () => {
+        expect(FormController.prototype.configure).to.be.calledOnce
+        expect(FormController.prototype.configure).to.be.calledWith(this.reqMock, globalRes, this.nextSpy)
+      })
     })
 
-    it('should set the list of markets', () => {
-      this.controller.configure(this.reqMock, globalRes, this.nextSpy)
+    context('when getOptions returns an error', () => {
+      beforeEach(async () => {
+        this.errorMessageMock = 'ERROR_REASON'
 
-      expect(this.reqMock.form.options.fields.primary_market.options).to.deep.equal([
-        {
-          value: '1',
-          label: 'One',
-        },
-        {
-          value: '3',
-          label: 'Three',
-        },
-      ])
-      expect(FormController.prototype.configure).to.be.calledWith(this.reqMock, globalRes, this.nextSpy)
+        nock(apiRoot)
+          .get('/metadata/omis-market/')
+          .replyWithError(this.errorMessageMock)
+
+        await this.controller.configure(globalReq, globalRes, this.nextSpy)
+      })
+
+      it('should call next with the error', () => {
+        const errorArgument = this.nextSpy.args[0][0]
+
+        expect(this.nextSpy).to.be.calledOnce
+        expect(errorArgument instanceof Error).to.be.true
+        expect(errorArgument.message).to.equal(`Error: ${this.errorMessageMock}`)
+      })
     })
   })
 })
--- a/test/unit/apps/omis/apps/create/controllers/sector.test.js
+++ b/test/unit/apps/omis/apps/create/controllers/sector.test.js
@@ -1,21 +1,13 @@
+const { assign } = require('lodash')
+
+const { apiRoot } = require('~/config')
 const FormController = require('~/src/apps/omis/controllers/form')
+const Controller = require('~/src/apps/omis/apps/create/controllers/sector')
 
-const sectorOptionsMock = [{
-  id: '1',
-  name: 'Sector one',
-}, {
-  id: '2',
-  name: 'Sector two',
-}, {
-  id: '3',
-  name: 'Sector three',
-}]
-
-const Controller = proxyquire('~/src/apps/omis/apps/create/controllers/sector', {
-  '../../../../../lib/metadata': {
-    sectorOptions: sectorOptionsMock,
-  },
-})
+const sectorOptionsMock = [
+  { id: '9999', name: 'ICT' },
+  { id: '8888', name: 'Farming' },
+]
 
 describe('OMIS create sector controller', () => {
   beforeEach(() => {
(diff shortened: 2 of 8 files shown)


test/unit/apps/omis/apps/create/controllers/market.test.js
```diff
@@ -1,20 +1,21 @@
...
-      name: 'Two',
-      disabled_on: '2010-01-01T00:00:00',
-    }, {
-      id: '3',
-      name: 'Three',
-      disabled_on: null,
-    }],
+  '../../../../../lib/options': {
+    getOptions () {
+      return marketOptionsMock
```
ztolley: How about following the pattern we've used for all the other options tests that uses nock instead of mocking the call to get options?

feedmypixel: + 1 for `nock` here

test/unit/apps/omis/apps/create/controllers/market.test.js
```diff
@@ -39,19 +40,10 @@ describe('OMIS create market controller', () => {
...
-        {
-          value: '1',
-          label: 'One',
-        },
-        {
-          value: '3',
-          label: 'Three',
-        },
-      ])
+      expect(this.reqMock.form.options.fields.primary_market.options).to.deep.equal(marketOptionsMock)
```
feedmypixel: does this nested reference need a `get`

test/unit/apps/omis/apps/create/controllers/sector.test.js
```diff
@@ -38,23 +40,10 @@ describe('OMIS create sector controller', () => {
...
-        {
-          value: '2',
-          label: 'Sector two',
-        },
-        {
-          value: '3',
-          label: 'Sector three',
-        },
-      ])
+      expect(this.reqMock.form.options.fields.sector.options).to.deep.equal(sectorOptionsMock)
```
feedmypixel: does this nested reference need a get? Safety first!

test/unit/apps/omis/apps/create/controllers/market.test.js
teneightfive: Don't feel like we need to in tests. It's expecting these objects so would rather it fails if they're not there.

test/unit/apps/omis/apps/create/controllers/market.test.js
```diff
@@ -25,34 +14,57 @@ describe('OMIS create market controller', () => {
...
-          options: {
-            fields: {
-              primary_market: {},
+    context('when getOptions returns 200', () => {
+      beforeEach(async () => {
+        nock(apiRoot)
+          .get('/metadata/omis-market/')
+          .reply(200, marketOptionsMock)
+
+        this.reqMock = Object.assign({}, globalReq, {
```
ztolley: lodash assign maybe?

test/unit/apps/omis/apps/create/controllers/market.test.js
```diff
@@ -25,34 +14,57 @@ describe('OMIS create market controller', () => {
...
+          .get('/metadata/omis-market/')
+          .replyWithError(this.errorMessageMock)
+
+        await this.controller.configure(globalReq, globalRes, this.nextSpy)
+      })
+
+      it('should call next with the error', () => {
+        const errorArgument = this.nextSpy.args[0][0]
+
+        expect(this.nextSpy).to.be.calledOnce
```
feedmypixel: Are we no longer doing 1 expectation per test?

test/unit/apps/omis/apps/create/controllers/market.test.js
ztolley: Not used globalReq on any of my tests, found it easier to just construct the one I needed in the test itself

test/unit/apps/omis/apps/create/controllers/market.test.js
```diff
@@ -25,34 +14,57 @@ describe('OMIS create market controller', () => {
...
-          value: '3',
-          label: 'Three',
-        },
-      ])
-      expect(FormController.prototype.configure).to.be.calledWith(this.reqMock, globalRes, this.nextSpy)
+    context('when getOptions returns an error', () => {
+      beforeEach(async () => {
+        this.errorMessageMock = 'ERROR_REASON'
+
+        nock(apiRoot)
```
feedmypixel: Can you also do a ` this.nockScope` and the relevant associated assertion. An example can be seen [here]( https://github.com/uktrade/data-hub-frontend/blob/b30a15149c23a577c126f14e5d48ba7a994eec46/test/unit/apps/contacts/repos.test.js#L13-L25)

test/unit/apps/omis/apps/create/controllers/sector.test.js
```diff
@@ -24,38 +14,57 @@ describe('OMIS create sector controller', () => {
...
   describe('configure()', () => {
-    beforeEach(() => {
-      this.reqMock = Object.assign({}, globalReq, {
-        form: {
-          options: {
-            fields: {
-              sector: {},
+    context('when getOptions returns 200', () => {
+      beforeEach(async () => {
+        nock(apiRoot)
```
feedmypixel: same for ` this.nockScope` please

test/unit/apps/omis/apps/create/controllers/sector.test.js
```diff
@@ -24,38 +14,57 @@ describe('OMIS create sector controller', () => {
...
+        expect(this.reqMock.form.options.fields.sector.options).to.deep.equal([
+          { value: '8888', label: 'Farming' },
+          { value: '9999', label: 'ICT' },
+        ])
       })
 
-      sandbox.spy(FormController.prototype, 'configure')
+      it('should call parent configure method', () => {
+        expect(FormController.prototype.configure).to.be.calledOnce
+        expect(FormController.prototype.configure).to.be.calledWith(this.reqMock, globalRes, this.nextSpy)
```
feedmypixel: The 1 assertion rule should apply here also?

test/unit/apps/omis/apps/create/controllers/sector.test.js
```diff
@@ -24,38 +14,57 @@ describe('OMIS create sector controller', () => {
...
-          value: '2',
-          label: 'Sector two',
-        },
-        {
-          value: '3',
-          label: 'Sector three',
-        },
-      ])
-      expect(FormController.prototype.configure).to.be.calledWith(this.reqMock, globalRes, this.nextSpy)
+        nock(apiRoot)
```
feedmypixel: can you add a ` this.nockScope` and assertion here?

test/unit/apps/omis/apps/create/controllers/sector.test.js
```diff
@@ -24,38 +14,57 @@ describe('OMIS create sector controller', () => {
...
+          .get('/metadata/sector/')
+          .replyWithError(this.errorMessageMock)
+
+        await this.controller.configure(globalReq, globalRes, this.nextSpy)
+      })
+
+      it('should call next with the error', () => {
+        const errorArgument = this.nextSpy.args[0][0]
+
+        expect(this.nextSpy).to.be.calledOnce
```
feedmypixel: 1 assertion rule should apply here also?

test/unit/apps/omis/apps/create/controllers/sector.test.js
teneightfive: My thoughts are that this is still one assertions. But just combines the count and expectation. Felt one for the count of calls and one for the value was unnecessary in this instance.

test/unit/apps/omis/apps/create/controllers/sector.test.js
teneightfive: See comment above re assertions

test/unit/apps/omis/apps/create/controllers/sector.test.js
feedmypixel: Agreed `this,nockScope` is to be removed in another PR

test/unit/apps/omis/apps/create/controllers/market.test.js
feedmypixel: I suppose in this example I feel it would read better if split out:

```
it('should call next ', () => {
  expect(this.nextSpy).to.be.calledOnce
})

it('should call next with an error', () => {
  expect(this.nextSpy.args[0][0] instanceof Error).to.be.true
})

it('should call next with expected error message', () => {
  expect(this.nextSpy.args[0][0]).to.equal(`Error: ${this.errorMessageMock}`)
})
```

test/unit/apps/omis/apps/create/controllers/sector.test.js
feedmypixel: and in this example maybe:

```
it('should call parent configure method', () => {
  expect(FormController.prototype.configure).to.be.calledOnce
})
it('should call parent configure method with expected arguments', () => {
  expect(FormController.prototype.configure).to.be.calledWith(this.reqMock, globalRes, this.nextSpy)
})     
```

I'm easy either way, but I feel this is easier to read and debug.

test/unit/apps/omis/apps/create/controllers/market.test.js
feedmypixel: I suppose in this example I feel it would read better if split out:

```
it('should call next ', () => {
  expect(this.nextSpy).to.be.calledOnce
})

it('should call next with an error', () => {
  expect(this.nextSpy.args[0][0] instanceof Error).to.be.true
})

it('should call next with expected error message', () => {
  expect(this.nextSpy.args[0][0]).to.equal(`Error: ${this.errorMessageMock}`)
})
```

test/unit/apps/omis/apps/create/controllers/sector.test.js
feedmypixel: and in this example maybe:

```
it('should call parent configure method', () => {
  expect(FormController.prototype.configure).to.be.calledOnce
})

it('should call parent configure method with expected arguments', () => {
  expect(FormController.prototype.configure).to.be.calledWith(this.reqMock, globalRes, this.nextSpy)
})     
```

I'm easy either way, but I feel this is easier to read and debug.

ztolley: As mentioned in chat, I'm with Dom on this one

test/unit/apps/omis/apps/create/controllers/market.test.js
teneightfive: I've used the `globalReq` elsewhere and was already being used in this test so probably going to keep it.

test/unit/apps/omis/apps/create/controllers/market.test.js
teneightfive: I'm not sure. I have been sticking to one thing rather than literally only one assertion. I think in this case they are testing the same thing. That next is called with the correct value. Not sure splitting it in this instance adds much value.

test/unit/apps/omis/apps/create/controllers/sector.test.js
teneightfive: Same as the other next comment. I feel there's a bit of repetition when you split these assertions and makes it harder to follow. I think seeing as this is testing the same context it's okay to group these two.

PR #1185, opened 2018-01-24
DH-1432 Investment project callout mapping bugs

yeahfro:
DH-1432

# Issue

Incomplete fields are not being correctly mapped:
1. `delivery_partners` form link is not being displayed
2. `undefined` is showed for address fields

screen shot 2018-01-24 at 10 48 45

# Fix
Fixes have been applied as follows:
- Add missing labels so that address fields can be correctly mapped to the callout
- Log error for any missing labels
- Add `delivery_partners` to missing lookup object

screen shot 2018-01-24 at 11 44 28


src/apps/investment-projects/helpers.js
```diff
@@ -79,14 +80,22 @@ const toCompleteStageMessages = {
 
 function buildIncompleteFormList (incompleteFields = []) {
   const formList = map(incompleteFields, (incompleteField) => {
-    const formFieldLink = formFieldLinks[incompleteField]
-    if (formFieldLink) {
-      return assign({}, formFieldLink, {
-        text: `${allLabels[incompleteField]} in ${formFieldLink.text}`,
-      })
+    const incompleteFieldLabel = allLabels[incompleteField]
```
feedmypixel: can you use a `get` here?

src/apps/investment-projects/helpers.js
```diff
@@ -79,14 +80,22 @@ const toCompleteStageMessages = {
...
+    const incompleteFieldFormLink = formFieldLinks[incompleteField]
+
+    if (!incompleteFieldLabel) {
```
feedmypixel: Do these data/fields change a lot? to the point we need logging around it? Why are we logging? It feels like we could remove the logging and prioritise tests around this area.

PR #1188, opened 2018-01-24
DH-1463 Add actual land date collection filters

yeahfro:
DH-1463

Change adds fields to filter investment project collection by `actual_land_date`.

screen shot 2018-01-24 at 15 29 35


ztolley:
Acceptance tests?

ztolley:
Noticed we have some tests but marked ignore. Lets get some tests in as a follow  up then

PR #1189, opened 2018-01-24
Release/4.4.1

feedmypixel:
# Bug Fixes
Switch to use contenthash for webpack css production assets
Ensure correct back link URL is present
Ensure null values aren't formatted as a date
Ensure payment value is an integer before sending to API
Handle missing labels by logging error
Add missing delivery_partners configuration
Add address labels to labels to be correctly mapped when mapping incomplete fields
Display field label only and not the form name

# Other
bump node version


ztolley:
Only comment is put the release notes in release notes, not here.
Also, put ticket numbers next to items to allow them to be linked back to Jira if people ask what they mean

feedmypixel:
> Only comment is put the release notes in release notes, not here.


Have added then to [both](https://github.com/uktrade/data-hub-frontend/releases/tag/untagged-5872ec383165b4edb7c7). Maybe overkill...

> Also, put ticket numbers next to items to allow them to be linked back to Jira if people ask what they mean

No Jira numbers, this is a patch release

PR #1191, opened 2018-01-24
Edit billing address after quote has been sent

teneightfive:
This change allows fields within the invoice details section to be updated after a quote has been accepted.

It adds a new journey which asks the user to confirm the VAT status if they change the billing country:

#### Billing address update without country change

`Edit invoice details` -> `Edit billing address` -> `Edit invoice details`

#### Billing address update with country change

`Edit invoice details` -> `Edit billing address` -> `Confirm VAT status` -> `Edit invoice details`

### Invoice details

localhost_3001_omis_63b83f12-7426-4d99-87a2-f12b9ddbe99d_edit_invoice-details 1

### Billing addres

localhost_3001_omis_63b83f12-7426-4d99-87a2-f12b9ddbe99d_edit_billing-address_returnurl _omis_63b83f12-7426-4d99-87a2-f12b9ddbe99d_edit_invoice-details

### VAT status confirmation

localhost_3001_omis_63b83f12-7426-4d99-87a2-f12b9ddbe99d_edit_vat-status

### Billing country updated

localhost_3001_omis_63b83f12-7426-4d99-87a2-f12b9ddbe99d_edit_invoice-details 2


src/apps/omis/apps/edit/controllers/billing-address.js
```diff
@@ -1,12 +1,33 @@
...
+      req.form.options.disableFormAction = !get(res.locals, 'order.canEditInvoiceDetails')
+    }
     req.form.options.fields.billing_address_country.options = metadataRepo.countryOptions.map(transformObjectToOption)
     super.configure(req, res, next)
   }
+
+  successHandler (req, res) {
+    const data = req.form.values
+    const newValue = get(data, 'billing_address_country')
+    const oldValue = get(res.locals, 'order.billing_address_country.id')
```
feedmypixel: can these have better names? `newCountryBillAddress`

feedmypixel: should you be comparing `ids` here?

teneightfive: Yeah good shout. I'll update these once the other PRs are in.

teneightfive: I think that's safer isn't it? The ID is the value stored in the form. Otherwise you'd need to find the name from the list of options before comparing. Seems less efficient and safe.

feedmypixel: > I think that's safer isn't it?

agreed 👍 

ah sorry, my point was should `billing_address_country` be `billing_address_country_id`?

teneightfive: Ah, no probably not. That is the object returned from the API. Unless we want to flatten all the objects that contain `id` and `name`?

Basically during this edit lifecycle I set the response for the order on the locals object so it is available in views and middleware/controllers.

src/apps/omis/apps/edit/controllers/billing-address.js
```diff
@@ -1,12 +1,35 @@
...
+    }
+
     super.configure(req, res, next)
   }
+
+  successHandler (req, res) {
+    const newBillingCountry = get(req, 'form.values.billing_address_country')
+    const oldBillingCountry = get(res.locals, 'order.billing_address_country.id')
+
+    if ((newBillingCountry && oldBillingCountry) && (newBillingCountry !== oldBillingCountry)) {
```
yeahfro: how about setting these conditions as vars, e.g. `hasBillingCountries` and `hasBillingCountryChanged`

then `if (hasBillingCountries && hasBillingCountryChanged)`

src/apps/omis/controllers/edit.js
```diff
@@ -33,7 +33,9 @@ class EditController extends FormController {
     req.sessionModel.reset()
     req.sessionModel.destroy()
 
-    req.flash('success', req.form.options.successMessage)
+    if (req.form.options.successMessage) {
+      req.flash('success', req.form.options.successMessage)
```
yeahfro: No unit test around this?

teneightfive: Good spot

PR #1194, opened 2018-01-25
Update cucumber dependency

ztolley:
Updates the cucumber and Nightwatch cucumber dependencies to the latest version. These two dependencies seem to be tied to each other.


feedmypixel:
might also be worth updating `chromedriver` also as its part of this "team"

PR #1196, opened 2018-01-25
Fix fragility when testing collection header

ztolley:
Previously when testing the collection header the tests assumed that there would be 0 or > 1 results. If the result actually had 1 result the heading wouldn't match the test text. This is because the heading uses the pluralise function to decide if it should say 'company' or 'companies'.

This fix addresses these issues:
* Makes locating the collection intro non-dependant on its plural nature
* Renames the step for checking the collection heading to accurately reflect it's purpose
* Separates locating a collection title from asserting the title contains the correct text

### DIFF
--- a/test/acceptance/features/collection/step_definitions/collection.js
+++ b/test/acceptance/features/collection/step_definitions/collection.js
@@ -3,6 +3,7 @@ const { client } = require('nightwatch-cucumber')
 const { defineSupportCode } = require('cucumber')
 
 const { getButtonWithText } = require('../../../helpers/selectors')
+const { pluralise } = require('../../../../../config/nunjucks/filters')
 
 defineSupportCode(({ When, Then }) => {
   const Collection = client.page.Collection()
@@ -45,21 +46,18 @@ defineSupportCode(({ When, Then }) => {
       })
   })
 
-  Then(/^the results count header for (.+) is present$/, async function (collectionType) {
-    const resultsCountHeaderSelector = Collection
-      .getSelectorForResultsCountHeader(collectionType)
+  Then(/^the results summary for a (.+) collection is present$/, async function (collectionType) {
+    await Collection.captureResultCount((count) => {
+      set(this.state, 'collection.resultCount', count)
+    })
+
+    const resultCount = get(this.state, 'collection.resultCount')
+    const collectionTypeTitle = pluralise(collectionType, resultCount)
 
     await Collection
-      .captureResultCount((count) => {
-        set(this.state, 'collection.resultCount', count)
-      })
-
-    await Collection
-      .api.useXpath()
-      .assert.visible(resultsCountHeaderSelector.selector)
-      .assert.containsText(resultsCountHeaderSelector.selector, collectionType)
-      .assert.containsText(resultsCountHeaderSelector.selector, get(this.state, 'collection.resultCount'))
-      .useCss()
+      .section.collectionHeader
+      .assert.visible('@intro')
+      .assert.containsText('@intro', `${resultCount} ${collectionTypeTitle}`)
   })
 
   Then(/^there is an (.+) button in the collection header$/, async function (buttonText) {
--- a/test/acceptance/features/companies/collection.feature
+++ b/test/acceptance/features/companies/collection.feature
@@ -12,7 +12,7 @@ Feature: View collection of companies
     Then I see the success message
     When I navigate to the Company collection page
     Then I confirm I am on the Companies page
-    And the results count header for companies is present
+    And the results summary for a company collection is present
     And there is an Add company button in the collection header
     And I can view the Company in the collection
       | text               | expected                  |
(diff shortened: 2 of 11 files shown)


test/acceptance/features/collection/step_definitions/collection.js
```diff
@@ -46,19 +47,22 @@ defineSupportCode(({ When, Then }) => {
   })
 
   Then(/^the results count header for (.+) is present$/, async function (collectionType) {
```
teneightfive: Want to change this to something more appropriate to what it's doing:

```
the results summary for a (.+) collection is present
```

test/acceptance/features/companies/collection.feature
```diff
@@ -12,7 +12,7 @@ Feature: View collection of companies
     Then I see the success message
     When I navigate to the Company collection page
     Then I confirm I am on the Companies page
-    And the results count header for companies is present
+    And the results count header for company is present
```
teneightfive: I know it's not a pattern we've used but I think these statement support quotes which I've found makes it easier to see what the dynamic value is:

```
And the results count header for "company" is present
```

ztolley: I too have generally used quotes for my step definitions, but none of the others do so this is to keep it consistent

feedmypixel: > I know it's not a pattern we've used but I think these statement support quotes which I've found makes it easier to see what the dynamic value is:

I moved away from quotes a while back because my IDE highlights it

screen shot 2018-01-25 at 16 51 36


However if your IDE doesn't then that is not great

test/acceptance/features/collection/step_definitions/collection.js
feedmypixel: > Want to change this to something more appropriate to what it's doing:

I'd agree with this, it reads better than what is currently there

test/acceptance/features/companies/collection.feature
ztolley: Mine does not but I don't think we should pick the syntax based on the IDE

test/acceptance/features/collection/step_definitions/collection.js
```diff
@@ -46,19 +47,22 @@ defineSupportCode(({ When, Then }) => {
...
-
     await Collection
       .captureResultCount((count) => {
         set(this.state, 'collection.resultCount', count)
       })
 
+    const resultCount = get(this.state, 'collection.resultCount')
+    const collectionTypeTitle = pluralise(collectionType, resultCount)
+
+    const resultsCountHeaderSelector = Collection
```
feedmypixel: AS a tidy up potentially in another PR I feel this could be using a css selector rather than searching by text

```
<div class="c-collection__header-intro">
  <span class="c-collection__result-count">5</span> 
  companies
</div>
```

You could then use 
```
document.querySelector('.c-collection__header-intro').textcontent
```

and compare the result 
```
5 companies
```

ztolley: I can change getSelectorForResultsCount header to simply return `div.c-collection__header-intro`, that fixes the issue having the text in the selector AND the assertion

test/acceptance/features/collection/step_definitions/collection.js
```diff
@@ -3,6 +3,7 @@ const { client } = require('nightwatch-cucumber')
 const { defineSupportCode } = require('cucumber')
 
 const { getButtonWithText } = require('../../../helpers/selectors')
+const { pluralise } = require('../../../../../config/nunjucks/filters')
```
yeahfro: we could use wavy here

test/acceptance/features/collection/step_definitions/collection.js
```diff
@@ -45,21 +46,19 @@ defineSupportCode(({ When, Then }) => {
...
     await Collection
-      .api.useXpath()
-      .assert.visible(resultsCountHeaderSelector.selector)
-      .assert.containsText(resultsCountHeaderSelector.selector, collectionType)
-      .assert.containsText(resultsCountHeaderSelector.selector, get(this.state, 'collection.resultCount'))
-      .useCss()
+      .section.collectionHeader
+      .assert.visible('@intro')
+      .assert.containsText('@intro', collectionTypeTitle)
+      .assert.containsText('@intro', resultCount)
```
yeahfro: are these two lines better than `.assert.containsText('@intro', resultCount + ' ' + collectTypeTitle)`? Appreciate what it was before

test/acceptance/features/collection/step_definitions/collection.js
ztolley: Actually would like to get rid of wavy tbh (after every yarn install we have to re-run yarn add wavy) and replace it with `process.cwd()`. For now, I'll leave this as it to be consistent with the rest of the acceptance tests

test/acceptance/features/collection/step_definitions/collection.js
ztolley: I'll change it

PR #1198, opened 2018-01-25
DH-1473 add application version to meta tag

feedmypixel:
Add the application version to a meta tag so its easy to work out which version of the frontend you are currently looking at

### DIFF
--- a/src/templates/_layouts/dit-base.njk
+++ b/src/templates/_layouts/dit-base.njk
@@ -5,6 +5,7 @@
     {% block head %}
       <meta http-equiv="content-type" content="text/html; charset=UTF-8">
       <meta name="viewport" content="width=device-width, initial-scale=1">
+      <meta name="version" content="{{ APP_VERSION }}">
       <meta http-equiv="x-ua-compatible" content="ie=edge">
       <title>{% block head_title_content %}{{ pageTitle | default(titleDefault) }}{% endblock %}</title>
       {% if description -%}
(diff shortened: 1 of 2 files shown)


src/templates/_layouts/dit-base.njk
```diff
@@ -5,6 +5,7 @@
     {% block head %}
       <meta http-equiv="content-type" content="text/html; charset=UTF-8">
       <meta name="viewport" content="width=device-width, initial-scale=1">
+      <meta name="app-version" content="{{ APP_VERSION }}">
```
teneightfive: Might be nice to prefix these. GOV.UK does something similar and helps distinguish which ones are custom the app.

Maybe:

```html
<meta name="datahub:version" content="{{ APP_VERSION }}">
```

ztolley: Actually I like version as it is. A version is generic, it's not unique to datahub.

ztolley: If it was attributes about datahub, like the current entity name or something then I would prefix them

teneightfive: But it's the version of the datahub rendering application, no?

If you don't like the idea of prefixing things specific to the frontend app I think just `version` is enough.

ztolley: so just go with version and we can all agree with that

feedmypixel:
@teneightfive @ztolley have changed to `<meta name="version" content="{{ APP_VERSION }}">`

PR #1200, opened 2018-01-25
Move handling of back and next values to edit handler

teneightfive:
Previously these values were being set within the edit configure
controller. The setting of these values feels better placed as part
of the step definition which is set within the edit handler.

### DIFF
--- a/src/apps/omis/apps/edit/controllers/edit-handler.js
+++ b/src/apps/omis/apps/edit/controllers/edit-handler.js
@@ -12,6 +12,7 @@ const i18n = i18nFuture({
 function editHandler (req, res, next) {
   const step = steps[`/${req.params.step}`]
   const order = res.locals.order
+  const nextUrl = req.query.returnUrl || `/omis/${order.id}/work-order`
 
   if (!step || !order) { return next() }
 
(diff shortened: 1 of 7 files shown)


src/apps/omis/apps/edit/controllers/edit-handler.js
```diff
@@ -12,6 +12,7 @@ const i18n = i18nFuture({
 function editHandler (req, res, next) {
   const step = steps[`/${req.params.step}`]
   const order = res.locals.order
+  const nextUrl = req.query.returnUrl || `/omis/${order.id}/work-order`
```
feedmypixel: Is `nextUrl` the right name for this Variable? Feels a little like you are combing two things here when there isn't a need to? Unless I am misunderstanding something?

feedmypixel:
We should up the `CodeClimate` [allowed number of lines](https://codeclimate.com/github/uktrade/data-hub-frontend/pull/1200) to something more than 25

ztolley:
You already know you have my vote to up the 25 line setting.

src/apps/omis/apps/edit/controllers/edit-handler.js
teneightfive: Yeah I think it began as `nextUrl` because it was being used for the next property which is where the form redirects to in the `successHandler` method. But in this case it's also used as the return URL for the back button (in our case the default text is `Return without saving`).

This value is either the value set specifically as a URL parameter, an example is to close the loops from the quote page, but the fallback is to always return the original order overview.

Does that make more sense?

Happy for a better suggestion for this variable. The Form Controller from the form wizard will look for the previous step by default but in these instances we don't normally have a previous step.

PR #1201, opened 2018-01-25
DH-1447 Fix editing project with empty estimated date

ztolley:
The investment form populate function did not cater for an estimated land date being null, this ended up with a date being generated in 1970. The change tests for the presence of the date before parsing it.

### DIFF
--- a/src/apps/investment-projects/transformers/project.js
+++ b/src/apps/investment-projects/transformers/project.js
@@ -99,9 +99,10 @@ function transformFromApi (body) {
     return get(body, `${key}.id`)
   })
 
-  const date = new Date(body['estimated_land_date'])
-  if (date) {
-    formatted['estimated_land_date_year'] = date.getFullYear()
+  const estimatedLandDate = body.estimated_land_date
+  if (!isEmpty(estimatedLandDate)) {
+    const date = new Date(estimatedLandDate)
+    formatted['estimated_land_date_year'] = date.getFullYear().toString()
     formatted['estimated_land_date_month'] = format(date, 'MM')
   }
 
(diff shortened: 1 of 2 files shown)


src/apps/investment-projects/transformers/project.js
```diff
@@ -99,9 +99,10 @@ function transformFromApi (body) {
     return get(body, `${key}.id`)
   })
 
-  const date = new Date(body['estimated_land_date'])
-  if (date) {
-    formatted['estimated_land_date_year'] = date.getFullYear()
+  const estimatedLandDate = get(body, 'estimated_land_date')
+  if (!isEmpty(estimatedLandDate)) {
```
feedmypixel: I wonder if something like [isValid](https://date-fns.org/v1.29.0/docs/isValid) would work well here?

ztolley: That wont work unless you try to create a date object, and if you do that you get a date set in 1970, which is the original problem.

ztolley: actually I have an idea

ztolley: I tried this but doesn't work, that function tests a date object and if I create a date object with the field value and it's empty or null that is what causes the issue in the first place. The requirement is to test if the field has a value or not, if it does then parse it, otherwise don't.

teneightfive: You don't need `get` here. It's useful for deep nested properties. This is on an object we know will exist. 

Would this be enough:

```js
if (body.estimated_land_date) {
  const date = new Date(body.estimated_land_date)
  formatted['estimated_land_date_year'] = date.getFullYear().toString()
  formatted['estimated_land_date_month'] = format(date, 'MM')
}
```

ztolley: You still to test for an empty string as this can cause the bug too

PR #1205, opened 2018-01-26
Tidy use of nock in unit tests

teneightfive:
This changes reduces some of the use cases of nock in unit tests:

- `nock` is available globally so does not need to be included in test files
- testing of nock scope being done doesn't feel necessary as we are already testing for the result being mocked. If they're not called with the mocked response our tests will already fail.


feedmypixel:
FYI the error in AT is addressed by the work in https://github.com/uktrade/data-hub-frontend/pull/1207

PR #1209, opened 2018-01-28
Move saving of OMIS data to saveValues method

teneightfive:
This change splits out what the current successHandler method is
responsible for. It was previously handling saving of the data to the
API and redirecting on success.

Looking again at the lifecycle it makes more sense that the saving of
data is taken care of in the saveValues method then the next method
is called to move to the successHandler.

This means the successHandler is then only responsible for clearing of
the session data and redirecting to the next URL. It means repetition
in those code when a custom method for saving is required can be
removed.

### DIFF
--- a/src/apps/omis/controllers/edit.js
+++ b/src/apps/omis/controllers/edit.js
@@ -17,23 +17,24 @@ const FormController = require('./form')
 const { Order } = require('../models')
 
 class EditController extends FormController {
-  async successHandler (req, res, next) {
-    const data = pick(req.sessionModel.toJSON(), Object.keys(req.form.options.fields))
-
+  async saveValues (req, res, next) {
     try {
-      const order = await Order.update(req.session.token, res.locals.order.id, data)
-      const nextUrl = req.form.options.next || `/omis/${order.id}`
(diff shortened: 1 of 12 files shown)


ztolley:
Wouldn't mind catching up and you walking me through the flow of the middleware here. At first glance, it seems you are suggesting splitting 1 middleware into 2? If anything, in other areas it has made sense to have the first middleware a post hits be responsible for saving and on success redirect, not split it out into 2 but maybe this is different (hence wanting to understand the reasoning)

src/apps/omis/controllers/edit.js
```diff
@@ -17,25 +17,26 @@ const FormController = require('./form')
...
-      res.redirect(nextUrl)
+      next()
     } catch (error) {
       next(error)
     }
   }
 
+  successHandler (req, res) {
+    req.journeyModel.reset()
+    req.journeyModel.destroy()
```
feedmypixel: I presume these can't be chained?

teneightfive: Sadly not :(

PR #1215, opened 2018-01-29
Add HTTP auth support

ztolley:
Added support for HTTP Basic Auth via environment config and updated the Heroku config to fix a bug not passing the OAUTH2_BYPASS_SSO setting and require users to set the basic auth user name and password to avoid accidentally allowing access to sensitive information without a password.


package.json
```diff
@@ -67,7 +68,7 @@
     "moment-duration-format": "^2.1.0",
     "nock": "^9.1.5",
     "node-sass": "^4.7.2",
-    "nodemon": "^1.14.3",
+    "nodemon": "^1.14.11",
```
ztolley: This crept in when I wasn't looking, but lets keep it

README.md
```diff
@@ -257,6 +259,9 @@ You could point data-hub-frontend to the UAT SSO environment. If you wish to ple
 [#technology-sso](https://ditdigitalteam.slack.com/messages/C5FLP2DSM/details/) team to set up the correct access and 
 SSO details required.
 
+### HTTP Basic Auth
+When running with `OAUTH2_BYPASS_SSO` the frontend can be user/password protected using by setting `BASICAUTH_USER` and `BASICAUTH_PASSWORD`. This is usefull for deploying the server for users and other team members to test
```
feedmypixel: typo `usefull` should be `useful`

README.md
```diff
@@ -87,6 +87,8 @@ This file expects the following environment variables:
 | API_ROOT | The url for a back end server instance for the service |
 | ARCHIVED_DOCUMENTS_BASE_URL | The url for the archived documents store |
 | ASSETS_HOST | Optional host for assets CDN, defaults to app’s host |
+| BASICAUTH_USER | Optionl. If running server with OAUTH2_BYPASS_SSO=true then this can be used in conjunction with a password to setup basic HTTP auth to protect the server |
```
feedmypixel: can this be `BASIC_AUTH_USER`, which is a little easier to read

README.md
```diff
@@ -87,6 +87,8 @@ This file expects the following environment variables:
...
+| BASICAUTH_PASSWORD |  Used in conjuction with BASICAUTH_USER to setup security for the server |
```
feedmypixel: can this be `BASIC_AUTH_PASSWORD`, which is a little easier to read

README.md
feedmypixel: is `This is usefull for deploying the server for users and other team members to test` needed? Auth is pretty much what you see is what you get.

config/index.js
```diff
@@ -66,6 +66,10 @@ const config = {
     devToken: process.env.OAUTH2_DEV_TOKEN,
     bypassSSO: process.env.OAUTH2_BYPASS_SSO || false,
   },
+  basicauth: {
```
feedmypixel: can this become `basicAuth` a little easier to read

package.json
feedmypixel: technically I'd do this a separate PR, but is ok for now

src/middleware/httpbasicauth.js
```diff
@@ -0,0 +1,26 @@
+const { get } = require('lodash')
```
feedmypixel: can this file just be called `basic-auth.js`?

src/middleware/httpbasicauth.js
```diff
@@ -0,0 +1,26 @@
...
+const auth = require('basic-auth')
```
feedmypixel: maybe `basicAuth`?

src/middleware/httpbasicauth.js
```diff
@@ -0,0 +1,26 @@
...
+      res.statusCode = 401
+      res.setHeader('WWW-Authenticate', 'Basic realm="datahub"')
+      return res.end('Access denied')
+    }
+  }
+
+  next()
+}
+
+module.exports = httpBasicAuth
```
feedmypixel: we use a pattern on other middleware using a closure:

```
module.exports = () => {
  return function basicAuth (req, res, next) {
    ...
    next()
  }
}
```

Can we follow that pattern here?

src/middleware/httpbasicauth.js
```diff
@@ -0,0 +1,26 @@
...
+
+const config = require('../../config')
+
+const httpBasicAuth = (req, res, next) => {
+  const bypassSSO = get(config, 'oauth.bypassSSO')
+  const httpBasicUser = get(config, 'basicauth.user')
+  const httpBasicPassword = get(config, 'basicauth.password')
+
+  if (bypassSSO && httpBasicUser && httpBasicPassword) {
+    const credentials = auth(req)
```
feedmypixel: I'd be tempted to name this variable `user`

src/middleware/httpbasicauth.js
```diff
@@ -0,0 +1,26 @@
...
+
+    if (!credentials ||
+        httpBasicUser !== credentials.name ||
+        httpBasicPassword !== credentials.pass) {
+      res.statusCode = 401
+      res.setHeader('WWW-Authenticate', 'Basic realm="datahub"')
+      return res.end('Access denied')
```
feedmypixel: Can this work go into `src/templates/errors.njk` and you just set the `statusCode` to `401` here?

feedmypixel: and the `res.setHeader('WWW-Authenticate', 'Basic realm="datahub"')` could be set in `src/middleware/errors.js`

src/server.js
```diff
@@ -100,6 +101,7 @@ app.use(breadcrumbs.setHome())
 app.use(flash())
 
 app.use(ssoBypass())
+app.use(httpBasicAuth)
```
feedmypixel: We only want this in a non-prod environment? can we wrap the middleware here in something that won't run it in Prod?

src/middleware/httpbasicauth.js
ztolley: Then this clashes with the basi-auth module that it imports... you have `basic-auth` importing `basic-auth`

src/middleware/httpbasicauth.js
ztolley: I'd worry about this clashing if oAuth sends a 401 or something. I'd feel more comfortable if this is self contained as it's only for devs to show their stuff

src/middleware/basic-auth.js
```diff
@@ -0,0 +1,26 @@
...
+
+function basicAuthMiddleware (req, res, next) {
+  const bypassSSO = get(config, 'oauth.bypassSSO')
+  const authUser = get(config, 'basicAuth.user')
+  const authPassword = get(config, 'basicAuth.password')
+
+  if (bypassSSO && authUser && authPassword) {
+    const user = basicAuth(req)
+
+    if (!user ||
```
teneightfive: Not sure about this formatting. I'd say either:

```js
if (
  !user ||
  authUser !== user.name ||
  authPassword !== user.pass
) {
  ...
}
```

or oneline

```js
if (!user || authUser !== user.name || authPassword !== user.pass) {
  ...
}
```

src/server.js
```diff
@@ -100,6 +101,11 @@ app.use(breadcrumbs.setHome())
 app.use(flash())
 
 app.use(ssoBypass())
+
+if (config.isDev) {
```
teneightfive: Won't the heroku environment be marked as production still?

ztolley: yeah lemme check

ztolley: Pulled it, heroku does indeed run in production mode. Having it only apply if bypass sso is also set I think is guard enough

src/middleware/httpbasicauth.js
feedmypixel: more about having errors in 1 place TBH

ztolley: Ok done

src/middleware/httpbasicauth.js
ztolley: I 'made it work'

src/middleware/basic-auth.js
```diff
@@ -0,0 +1,25 @@
...
+function basicAuthMiddleware (req, res, next) {
+  const bypassSSO = get(config, 'oauth.bypassSSO')
+  const authUser = get(config, 'basicAuth.user')
+  const authPassword = get(config, 'basicAuth.password')
+
+  if (bypassSSO && authUser && authPassword) {
+    const user = basicAuth(req)
+
+    if (!user || authUser !== user.name || authPassword !== user.pass) {
+      res.setHeader('WWW-Authenticate', 'Basic realm="datahub"')
```
feedmypixel: I'd be tempted to move this into the `catchAll` method in `src/middleware/errors.js`

feedmypixel: Actually just seen this

>I'd worry about this clashing if oAuth sends a 401 or something. I'd feel more comfortable if this is self contained as it's only for devs to show their stuff

Thats a fair point as `Basic` is being set here

PR #1216, opened 2018-01-29
Ensure values are escaped in entity and meta item

teneightfive:
Due to the use of the `highlight` filter the output for the heading
within an entity and meta item will not be escaped.

Seeing as this value can come from user input we should ensure that
the initial value at least is always escaped before the highlight
filter is applied.


yeahfro:
Merging this to get it deployed

PR #1221, opened 2018-01-30
Remove advisers as facet from interactions

teneightfive:
This facet within the live environment currently returns a large
amount of options (9,000+) causing a lot of performance issues within
the browser.

After understanding the need for this facet and that providing all
possible options not being seen as a viable solution it is being
removed in favour of _My interactions_ temporarily until search
facets are more dynamic in the options they show.

## Before

image

## After

image


feedmypixel:
Nice approach 👍

PR #1223, opened 2018-01-31
DH-1491 Do not sort all options

yeahfro:
DH-1491

# Issue

Fields in the company edit and investment project value edit forms contain fields with multiple options. These options are being sorted by the new option library.

# Fix

Not all options should be sorted so a parameter is being passed into the function to prevent sorting. Options will be sorted by default.

# Before

screen shot 2018-01-31 at 14 46 27

screen shot 2018-01-31 at 14 47 06

# After

screen shot 2018-01-31 at 14 45 42

screen shot 2018-01-31 at 14 45 18

### DIFF
--- a/src/lib/options.js
+++ b/src/lib/options.js
@@ -5,7 +5,7 @@ const authorisedRequest = require('../lib/authorised-request')
 const { filterDisabledOption } = require('../apps/filters')
 const { transformObjectToOption } = require('../apps/transformers')
 
-async function getOptions (token, key, { createdOn, currentValue, includeDisabled = false } = {}) {
+async function getOptions (token, key, { createdOn, currentValue, includeDisabled = false, sorted = true } = {}) {
   const url = `${config.apiRoot}/metadata/${key}/`
   let options = await authorisedRequest(token, url)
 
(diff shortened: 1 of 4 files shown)


src/lib/options.js
```diff
@@ -5,15 +5,17 @@ const authorisedRequest = require('../lib/authorised-request')
 const { filterDisabledOption } = require('../apps/filters')
 const { transformObjectToOption } = require('../apps/transformers')
 
-async function getOptions (token, key, { createdOn, currentValue, includeDisabled = false } = {}) {
+async function getOptions (token, key, { createdOn, currentValue, includeDisabled = false, canSort = true } = {}) {
```
ztolley: canSort seems an odd name. Why not just sort?

teneightfive: I like the idea of is being more obvious it's a boolean. `sort` is a little ambiguous, it could be sort or don't sort (boolean) or the value to sort by.

What about `shouldSort` or `isSorted`?

ztolley: I'm thinking of it as in instruction, not saying it is something that has already been sorted.

You 'can' sort anything, so I don't like `canSort` is right and it's not asking a question 'is this already sorted'. 

So I like `sort` but would compromise on `shouldSort`

src/lib/options.js
```diff
@@ -5,15 +5,17 @@ const authorisedRequest = require('../lib/authorised-request')
 const { filterDisabledOption } = require('../apps/filters')
 const { transformObjectToOption } = require('../apps/transformers')
 
-async function getOptions (token, key, { createdOn, currentValue, includeDisabled = false } = {}) {
+async function getOptions (token, key, { createdOn, currentValue, includeDisabled = false, shouldSort = true } = {}) {
```
teneightfive: Last minute thought...what about `sorted`?

Seeing it again, `shouldSort` just looks odd :/

PR #1225, opened 2018-02-01
DH-1467 Fix oauth stateid mismatch part 2

feedmypixel:
## Suspected problem
A user comes into datahub and goes through SSO
- they are on instance `a`
- in the `redirectOAuth` method `stateId` is saved to the session
- then SSO redirects back to the application but this time to instance `b`
- the session on instance `b` is not in sync with redis so the `stateId` is not present in its session
- the error [in sentry](https://sentry.ci.uktrade.io/dit/front-end/issues/3045/?referrer=slack) then occurs

After reverse engineering the journey it appears that everything is ok apart from we don't sync instances. So part 2 of this investigation adds `reloadSession` which will reload a session from `redis` using the ID in our session cookie. 

Due to complications in testing this PR also has work that splits out the rather fat controller into teh relvant files, it also updates the tests to reflect the work and the splitting out.

### Of note
all the logging will be going ocne this fix is confirmed

### DIFF
--- a/src/apps/oauth/controllers.js
+++ b/src/apps/oauth/controllers.js
@@ -1,19 +1,31 @@
 const queryString = require('query-string')
 const uuid = require('uuid')
 
-const { get, set } = require('lodash')
+const { get, set, isUndefined } = require('lodash')
 
 const config = require('./../../../config')
 const logger = require('./../../../config/logger') // @TODO remove this once we've diagnosed the cause of the mismatches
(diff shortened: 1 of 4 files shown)


feedmypixel:
splitting this work into 2 PR's as the problem cannot be reproduced in staging which makes me think it is something wrong with the prod setup rather than the code in our frontend

PR #1228, opened 2018-02-02
DH-1467 Refactor/tidy up oauth controller

feedmypixel:
Update tests for Oauth controller

Originally it sat with #1229, but I have split it out as I'm not sure yet if the work in #1229 is actually needed.

### Of note
all the logging will be going once this fix is confirmed

### DIFF
--- a/src/apps/oauth/controllers.js
+++ b/src/apps/oauth/controllers.js
@@ -7,12 +7,12 @@ const { get, set } = require('lodash')
 const config = require('./../../../config')
 const logger = require('./../../../config/logger') // @TODO remove this once we've diagnosed the cause of the mismatches
 
-function getAccessToken (oauthCode) {
+function getAccessToken (code) {
   const options = {
     method: 'POST',
     url: config.oauth.tokenFetchUrl,
     formData: {
-      code: oauthCode,
+      code,
       grant_type: 'authorization_code',
       client_id: config.oauth.clientId,
       client_secret: config.oauth.clientSecret,
--- a/test/unit/apps/oauth/controllers.test.js
+++ b/test/unit/apps/oauth/controllers.test.js
@@ -9,6 +9,14 @@ describe('OAuth controller', () => {
       './../../../config': this.mockConfig,
       'uuid': this.mockUuid,
     })
+    this.mockOauthAccessToken = 'mockAccessToken'
+    this.mockAuthCode = 'mock-auth-code'
+    this.mockFetchUrl = {
+      host: 'http://mock-oauth-host',
+      path: '/example-fetch-token-url',
+    }
+    this.mockStateId = 'mock-state-id'
+
     this.resMock = assign({}, globalRes, {
       redirect: sandbox.spy(),
       render: sandbox.spy(),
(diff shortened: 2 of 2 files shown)


src/apps/oauth/controllers.js
```diff
@@ -1,28 +1,11 @@
...
-      client_id: config.oauth.clientId,
-      client_secret: config.oauth.clientSecret,
-      redirect_uri: config.oauth.redirectUri,
-    },
-    json: true,
-  }
-
-  return request(options)
-}
+const { getAccessToken } = require('./services')
```
teneightfive: I thought we were trying to remove the idea of `services`? They don't feel well enough defined to me.

src/apps/oauth/services.js
```diff
@@ -0,0 +1,24 @@
+const request = require('request-promise')
+
+const config = require('./../../../config')
+
+function getAccessToken (oauthCode) {
+  const options = {
+    method: 'POST',
+    url: config.oauth.tokenFetchUrl,
```
teneightfive: `get()` on these properties?

src/apps/oauth/services.js
```diff
@@ -0,0 +1,24 @@
+const request = require('request-promise')
+
+const config = require('./../../../config')
+
+function getAccessToken (oauthCode) {
```
teneightfive: What about just `code`? Then you can use shorthand in the object definition.

src/apps/oauth/controllers.js
feedmypixel: I'd disagree, `services` is a pretty well-defined pattern, outside and Inside of our `app`.
screen shot 2018-02-02 at 10 29 51

src/apps/oauth/services.js
feedmypixel: These are all required for Oauth to work so I don't think `get()` is needed here

test/unit/apps/oauth/controllers.test.js
```diff
@@ -18,12 +26,20 @@ describe('OAuth controller', () => {
       session: {},
     })
     this.nextSpy = sandbox.spy()
+
+    set(this.mockConfig, 'oauth', {
```
teneightfive: Wonder whether set is overkill here? You've defined the `mockConfig` property above so can be sure it exists. You're also not setting a deep property so don't think you'll need this helper.

test/unit/apps/oauth/controllers.test.js
```diff
@@ -18,12 +26,20 @@ describe('OAuth controller', () => {
...
+      tokenFetchUrl: this.mockFetchUrl.host + this.mockFetchUrl.path,
+      clientSecret: 'mock-client-secret',
+      redirectUri: '/mock/callback/url',
+      clientId: 'mockClientId',
+    })
   })
 
   describe('#redirectOAuth', () => {
+    this.mockUUIDvalue = 'mock-uuid-1234'
```
teneightfive: Not sure if this has access to the `this` that you want to set this property on. Might be worth double checking this works as expected.

test/unit/apps/oauth/controllers.test.js
```diff
@@ -116,30 +122,21 @@ describe('OAuth controller', () => {
...
-            path: '/example-fetch-token-url',
-          }
-
-          set(this.mockConfig, 'oauth', {
-            tokenFetchUrl: this.mockFetchUrl.host + this.mockFetchUrl.path,
-            clientSecret: 'mock-client-secret',
-            redirectUri: '/mock/callback/url',
-            clientId: 'mockClientId',
-          })
           set(this.reqMock, 'query', {
```
teneightfive: Again, not sure you need `set` here as you know `reqMock` property exists.

test/unit/apps/oauth/controllers.test.js
```diff
@@ -161,91 +158,97 @@ describe('OAuth controller', () => {
...
-      }
-
       beforeEach(() => {
-        set(this.mockConfig, 'oauth', {
-          tokenFetchUrl: mockFetchUrl.host + mockFetchUrl.path,
-          clientSecret: 'mock-client-secret',
-          redirectUri: '/mock/callback/url',
-          clientId: 'mockClientId',
-        })
         set(this.reqMock, 'query', {
```
teneightfive: Same as above re need for `set`.

test/unit/apps/oauth/controllers.test.js
feedmypixel: this is being tested on [line 69](https://github.com/uktrade/data-hub-frontend/pull/1228/files/f7702736f527422e2e5b72fd728927a38d728795#diff-224a60a8afda1ee300bdcdf49c88b4cdR69)

PR #1230, opened 2018-02-02
DEMO: list cards

feedmypixel:
Quick 👉 [demo](https://datahub-beta2-pr-1230.herokuapp.com/companies?archived=false&page=1000&sortby=modified_on%3Adesc) 👈 with some thoughts around how to display data in list cards.

FYI @yeahfro 

### OF Note
- Obvs tests wont pass this is a smoke and mirrors example
- OBVS all work is of DEMO quality, not for PROD!
- I like the alignment and knowing where everything is but I'm not totally sold on `n/a` I personally think it works better though than having things in different places
- ~~We also may want to think about having things like `Updated` always in the same place across collections. Or maybe it doesn't need to be part of the metaElements, feels like it could just be in 1 generic place, say top right of each card. See screenshot or example.~~
- We potentially need to do something about long titles (if its a real world case)
- Have added turnover to fill out the space by moving `updated` up

### Screen shot
screencapture-localhost-3000-companies-1517617469983

### DIFF
--- a/assets/stylesheets/components/_entity-list.scss
+++ b/assets/stylesheets/components/_entity-list.scss
@@ -3,7 +3,7 @@
 @import "../tools/functions/typography";
 
 .c-entity-list__item {
-  padding: $default-spacing-unit 0;
+  padding: $default-spacing-unit/2 0 $default-spacing-unit;
 
   & + & {
     border-top: 1px solid $grey-2;
--- a/assets/stylesheets/components/_entity.scss
+++ b/assets/stylesheets/components/_entity.scss
@@ -17,6 +17,10 @@
   }
 }
 
+.c-entity__updated {
+  margin-bottom: $default-spacing-unit;
+}
+
 @include media(tablet) {
   .c-entity__header,
   .c-entity__content {
--- a/assets/stylesheets/components/_meta-list.scss
+++ b/assets/stylesheets/components/_meta-list.scss
@@ -41,12 +41,13 @@
 
 .c-meta-list--inline {
   .c-meta-list__item-label {
-    display: inline-block;
+    display: block;
   }
 
   .c-meta-list__item {
     margin-top: $baseline-grid-unit;
-    margin-right: $default-spacing-unit;
+    padding-right: $default-spacing-unit;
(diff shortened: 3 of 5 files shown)


teneightfive:
Like the look of this @feedmypixel!

>but I'm not totally sold on n/a
I agree on this, feels really bad to put generic holding text like this and almost suggests that's the value somebody has entered. I found with OMIS the conditional data can be stacked in a way that even when it's there or isn't, the other items always remain in the same place. Something we could review for the entities.

>Or maybe it doesn't need to be part of the metaElements, feels like it could just be in 1 generic place, say top right of each card

I like the idea of a generic place too, almost like what we did with reference number on investment projects. 

Could take some inspiration from GitHub pull requests too and have that as one line underneath each title that could also say who it was last updated by (I'm pretty sure we have that information):

image

>We potentially need to do something about long titles

I think this new layout might not make long titles look as bad. As long as we create some space between titles and the tags on the right hand side.

feedmypixel:
example

An example (in photoshop) of what I was thinking Re: updated date (please excuse the slightly ropy alignment)

feedmypixel:
@teneightfive yes to all that 👍 💯

feedmypixel:
We also need to sort out the address formatting here.

teneightfive:
>We also need to sort out the address formatting here.

Very tricky territory...I think that's why companies house, for example, don't touch it and just display it as it was entered.

feedmypixel:
> Very tricky territory...I think that's why companies house, for example, don't touch it and just display it as it was entered.

Something as simple as putting breaks in after commas may really help with readability

feedmypixel:
@teneightfive @yeahfro I had a little play, have updated the `updated` text, see demo and screenshot

teneightfive:
>Something as simple as putting breaks in after commas may really help with readability

But then I think in this display it would make each _card_ quite long, reducing scanability. I'd say in a different view that will help but maybe not in these summary screens.

feedmypixel:
This Demo is no longer needed @ztolley and @yeahfro have this work in hand

PR #1231, opened 2018-02-02
Common steps for acceptance tests

teneightfive:
This work consolidates some of the step definition use into shared steps where some have very similar patterns.

Summary of changes:

- move step definitions in feature folders that contain no feature tests into a shared step definitions folder
- create a common step definition file for shared steps like navigating to a particular page
- remove use of navigating to a particular URL from feature files and use idea of page objects to define URLs
- use page object navigation step in place of collection specific step definition

For a more detailed description see individual commit messages.


test/acceptance/features/step_definitions/common.js
```diff
@@ -0,0 +1,9 @@
+const { get } = require('lodash')
+const { client } = require('nightwatch-cucumber')
+const { When } = require('cucumber')
+
+When(/^I navigate to the "(.+)" page$/, async function (pageName) {
+  const Page = get(client.page, pageName)()
```
feedmypixel: So your always relying on getting a function here, which means it will blow up when the `get` returns `undefined`. Maybe separate out the page object function call into an `isFunction` check?

test/acceptance/features/companies/collection.feature
```diff
@@ -10,7 +10,7 @@ Feature: View collection of companies
 
     When a "Foreign company" is created
     Then I see the success message
-    When I navigate to the Company collection page
+    When I navigate to the "companies.List" page
```
feedmypixel: I'm not convinced on the quotes. I thought this had been agreed to not use quotes? It feels like a lot of extra quotes when they're not really needed. Can I suggest looking for a plugin so your IDE does the same as everyone else's. And then see how you feel after that? I feel if we are having quotes or not then it needs to be a group decision.

test/acceptance/features/companies/collection.feature
```diff
@@ -26,14 +26,14 @@ Feature: View collection of companies
   @companies-collection--view--lep @lep
   Scenario: View companies collection as LEP
 
-    When I navigate to the Company collection page
+    When I navigate to the "companies.List" page
```
feedmypixel: I'm also not convinced of the readability of this. Written on paper it doesn't make sense and there is an element of needing to reverse engineer what it is going. I personally would vote for taking the complication into the page object so you just show `When I navigate to the Company collection page` and then in the background use a map to access the page Object. You could even just map the words `Companies List` to `companies.List`

test/acceptance/features/companies/collection.feature
teneightfive: Yeah I think you're right. I added them to see how people felt actually seeing them in the code as before it was just speculation.

Maybe quotes should be reserved for when pure strings are passed through to the step definition.

test/acceptance/features/companies/collection.feature
teneightfive: Yeah I'm not 100% sure on the best approach at the moment. I felt some of the existing step definitions were needing to do quite a lot to try and translate the language in feature files into useable things. Sometimes using helpers like case changing or pluralising which felt it could be flakey. 

This method is more specific but still readable. Maybe we rename the `List` page to `Collection` to keep our consistent language?

test/acceptance/features/step_definitions/common.js
teneightfive: I'm in two minds. I think it's good to maybe protect against that but we should still throw an error?

If you call this step definition with a bad value we want it to fail I suppose, not to silently carry on?

feedmypixel: Not sure if there is a need to throw an error. Won't it fail because it doesn't navigate to the expected page? Rather than failing because it blows up?

PR #1234, opened 2018-02-05
Acceptance tests/fixture step definition

teneightfive:
The idea behind this change is to move to more of a page object structure with acceptance tests.

Currently navigating to fixtures involves visiting the homepage and searching for a name then clicking the first result. This creates a number of extra requests and adds time to a growing set of acceptance tests.

Search can still be tested separately but we should be able to navigate straight to a fixture if we know it's ID.

I'm not convinced on the new name for the page objects but the current page objects for each entity contain a lot of code for other pages. We would need to split out functionality used for nested pages like _contacts_ within _companies_ from the `Company.js` page object first. The main `Company.js` page object can then contain the code from the `Fixture.js` page object which allows you to navigate directly to a fixture using the page objects `.url()` method.

These changes are the start of moving to more flexible, shared step definitions and having some of the heavy lifting of interacting with the page done in page objects.

### DIFF
--- a/test/acceptance/features/audit/contact.feature
+++ b/test/acceptance/features/audit/contact.feature
@@ -7,8 +7,8 @@ Feature: View Audit history of a contact
   @audit-contact--fields
   Scenario: View name of the person who made contact record changes
 
-    Given I navigate to company fixture Lambda plc
-    When I click the Contacts local nav link
+    When I navigate to the `companies.Fixture` page using `company` `Lambda plc` fixture
+    And I click the Contacts local nav link
     And I click the "Add contact" link
     And a primary contact is added
     When I submit the form
@@ -26,8 +26,8 @@ Feature: View Audit history of a contact
   @audit-contact--count
   Scenario: View the number of changes occurred on a contact record
 
-    Given I navigate to company fixture Lambda plc
-    When I click the Contacts local nav link
+    When I navigate to the `companies.Fixture` page using `company` `Lambda plc` fixture
+    And I click the Contacts local nav link
     And I click the "Add contact" link
     Then a primary contact is added
     When I submit the form
@@ -42,8 +42,8 @@ Feature: View Audit history of a contact
   @audit-contact--archived
   Scenario: View audit log for Archived contact
 
-    Given I navigate to company fixture Lambda plc
-    When I click the Contacts local nav link
+    When I navigate to the `companies.Fixture` page using `company` `Lambda plc` fixture
+    And I click the Contacts local nav link
     And I click the "Add contact" link
     Then a primary contact is added
     When I submit the form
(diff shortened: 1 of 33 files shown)


test/acceptance/features/audit/contact.feature
```diff
@@ -7,8 +7,8 @@ Feature: View Audit history of a contact
   @audit-contact--fields
   Scenario: View name of the person who made contact record changes
 
-    Given I navigate to company fixture Lambda plc
-    When I click the Contacts local nav link
+    When I navigate to the `companies.Fixture` page using `company` `Lambda plc` fixture
```
ztolley: The syntax feels very technical and not very English, if that makes sense.

teneightfive: Yeah have been thinking about how we can improve it.

Was thinking one option is to lowercase the page objects so they look less syntaxy. And then if we improve the existing page objects this could become more like:

```
When I navigate to the `companies.company` page as a `uk limited company`
```

ztolley: That still feels more like a programming language than a sentence.

`When I navigate to company details page for 'Lamda plc'`

teneightfive: Yes but the problem is translating that into a page object. One option is to have a map for reads better to page object:

```
const pages = {
  'company details': 'companies.Company',
}
```

But at the time is was deemed we didn't want the extra maintenance overhead. In reality it has worked out that only developers are the ones writing and reading these tests.

test/acceptance/pages/companies/Fixture.js
```diff
@@ -0,0 +1,13 @@
+const { find } = require('lodash')
+
+const { company } = require('../../features/setup/fixtures')
+
+module.exports = {
+  url: function companyFixtureUrl (companyName) {
+    const fixture = find(company, { name: companyName })
+    const companyId = fixture ? fixture.pk : company.ukLtd.pk
```
feedmypixel: I think `id` would work better here as `pk` (primary key) is a Django convention. It would also work better with the const name of `companyId`.

feedmypixel: it might also be worth storing the default in a const just to explain what it is more. Something like:
```
const defaultCompany = company.ukLtd

const companyId = fixture ? fixture.id : defaultCompany.id
```

feedmypixel: Maybe also look into using `get`, for the use of a fallback.

```
const companyId = get(fixture, 'id', defaultCompany.id)
```

teneightfive: Changing to `id` would mean updating all the fixtures. Currently our fixtures store the UUID as the `pk` property which you're right, isn't necessarily right. 

This change is just moving the logic of getting that value to a different step definition. Would say we should do that separately if we don't like that property name?

feedmypixel: > Would say we should do that separately if we don't like that property name?

Sounds good.

test/acceptance/features/audit/contact.feature
feedmypixel: Re: the conversation about a map means more reverse engineering. I'd say the plan English a map would provide in the `feature` file _should_ negate the need for any reverse engineering.

feedmypixel:
`pk` -> `id` I'm fine with this being in a follow-up PR

human readable `map` I'd vote for this being done in a follow-up PR I have [added a comment around this](https://github.com/uktrade/data-hub-frontend/pull/1234#discussion_r171519528)

PR #1235, opened 2018-02-05
Allow contact details to be edited after a quote has been sent

teneightfive:
This change allows the contact details to be changed during later
stages of the OMIS order.

It uses the same method as used for subscribers, assignees and
invoice details.

## What it looks like

localhost_3001_omis_b171225e-bade-4c11-8996-de9d30c8f9e0_edit_client-details 1

### DIFF
--- a/src/apps/omis/middleware.js
+++ b/src/apps/omis/middleware.js
@@ -16,11 +16,13 @@ async function setCompany (req, res, next, companyId) {
 async function setOrder (req, res, next, orderId) {
   try {
     const order = await Order.getById(req.session.token, orderId)
+    const activeOrder = !(['cancelled', 'complete'].includes(order.status))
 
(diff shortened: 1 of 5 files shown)


src/apps/omis/middleware.js
```diff
@@ -16,11 +16,13 @@ async function setCompany (req, res, next, companyId) {
 async function setOrder (req, res, next, orderId) {
   try {
     const order = await Order.getById(req.session.token, orderId)
+    const activeOrder = !['cancelled', 'complete'].includes(order.status)
```
yeahfro: This would probably be clearer as `!(['cancelled', 'complete'].includes(order.status))` as it's not obvious what the `!` is applied to

teneightfive:
Merging with admin privileges as @yeahfro's comments have been addressed.

PR #1236, opened 2018-02-06
DH-1470 Add TAP status field to service delivery form

yeahfro:
DH-1470

This change:
- introduces the `Status` field to the service delivery form
- refactors some code for tidy up

screen shot 2018-02-07 at 17 05 08


jim68000:
That wording is a bit confusing. Could we add a TAP to it so it's TAP Status? Otherwise people will assume this is something they can add to any service.

jim68000:
Had offline chat with Lee. Making display dependent on which service is selected is doable but would be painful as this is a user-editable list. As we are under the pump (as they say in Australia) to get this in, can we go with some explanatory text 'Trade Access Program (TAP) status - only use if a TAP service is selected above' or somesuch?

src/apps/interactions/transformers.js
```diff
@@ -26,19 +27,20 @@ function transformInteractionResponseToForm ({
     id: id,
     contact: get(contact, 'id'),
     dit_team: get(dit_team, 'id'),
+    dit_adviser: get(dit_adviser, 'id'),
+    is_event: isNil(event) ? 'false' : 'true',
+    event: get(event, 'id'),
     service: get(service, 'id'),
+    service_delivery_status: get(service_delivery_status, 'id', null),
     subject: subject,
```
teneightfive: Not related but these could be shorthand, no?

PR #1238, opened 2018-02-07
Allow an order to be started without a company

teneightfive:
This change adds the ability to start an order without the company preselected.

Currently the user has to find the company, go to the orders _tab_ and begin the creation journey from there. This wasn't the logical starting place for many users so these changes adds a step for selecting the company.

If the journey is started from the original starting point they will bypass this step.

🔥 [**DEMO**](https://datahub-beta2-pr-1238.herokuapp.com/omis) 🔥 

### Collection

### Company selection
localhost_3001_omis_create_company

### Company selection results
localhost_3001_omis_create_company_term samsung

### Edit company (either back or from summary edit)
localhost_3001_omis_create_company_edit

### Creation summary
localhost_3001_omis_create_confirm


test/unit/apps/omis/controllers/create.test.js
```diff
@@ -1,3 +1,136 @@
+const { assign } = require('lodash')
+
+const FormController = require('~/src/apps/omis/controllers/form')
+const companyMock = require('~/test/unit/data/company.json')
+
 describe('OMIS CreateController', () => {
+  beforeEach(() => {
+    this.nextSpy = sandbox.spy()
+    this.errorSpy = sandbox.spy()
+    this.getDitCompanyStub = sandbox.stub().resolves(companyMock)
```
ztolley: Consider using Nock

PR #1239, opened 2018-02-08
Fix/oauth state mismatch

feedmypixel:
This PR will supersede https://github.com/uktrade/data-hub-frontend/pull/1229
The approach here is one of saving the session to the session store before moving on through the SSO journey.

Thanks to @web-bert for his thoughts on this.

### DIFF
--- a/src/apps/oauth/controllers.js
+++ b/src/apps/oauth/controllers.js
@@ -4,6 +4,7 @@ const uuid = require('uuid')
 
 const { get, set } = require('lodash')
 
+const { saveSession } = require('./../../lib/session-helper')
 const config = require('./../../../config')
 const logger = require('./../../../config/logger') // @TODO remove this once we've diagnosed the cause of the mismatches
 
@@ -53,7 +54,7 @@ async function callbackOAuth (req, res, next) {
   }
 }
 
-function redirectOAuth (req, res) {
+async function redirectOAuth (req, res, next) {
   const stateId = uuid()
   const urlParams = {
     response_type: 'code',
@@ -74,8 +75,14 @@ function redirectOAuth (req, res) {
     set(urlParams, 'code', oAuthDevToken)
   }
 
-  set(req.session, 'oauth.state', stateId) // used to check the callback received contains matching state param
-  return res.redirect(`${config.oauth.url}?${queryString.stringify(urlParams)}`)
+  try {
+    // used to check the callback received contains matching state param
+    set(req.session, 'oauth.state', stateId)
+    await saveSession(req.session)
+    res.redirect(`${config.oauth.url}?${queryString.stringify(urlParams)}`)
+  } catch (error) {
+    next(error)
+  }
 }
 
 function renderHelpPage (req, res, next) {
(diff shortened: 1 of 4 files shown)


src/apps/oauth/controllers.js
```diff
@@ -74,8 +75,13 @@ function redirectOAuth (req, res) {
     set(urlParams, 'code', oAuthDevToken)
   }
 
-  set(req.session, 'oauth.state', stateId) // used to check the callback received contains matching state param
-  return res.redirect(`${config.oauth.url}?${queryString.stringify(urlParams)}`)
+  try {
+    set(req.session, 'oauth.state', stateId) // used to check the callback received contains matching state param
+    await saveSession(req)
+    return res.redirect(`${config.oauth.url}?${queryString.stringify(urlParams)}`)
```
teneightfive: Don't think we need to return this line.

src/apps/oauth/controllers.js
```diff
@@ -74,8 +75,13 @@ function redirectOAuth (req, res) {
...
+  } catch (error) {
+    return next(error)
```
teneightfive: Don't think we need to return here either.

src/lib/session-helper.js
```diff
@@ -0,0 +1,20 @@
+/**
+ * promise wrapper around express session save method
+ * @param req
+ * @returns {Promise}
+ */
+function saveSession ({ session }) {
+  return new Promise((resolve, reject) => {
+    session.save((error) => {
+      if (error) {
+        reject(error)
```
teneightfive: Does this need a `return` to stop it moving on and calling `resolve`?

src/lib/session-helper.js
```diff
@@ -0,0 +1,20 @@
+/**
+ * promise wrapper around express session save method
+ * @param req
+ * @returns {Promise}
+ */
+function saveSession ({ session }) {
```
teneightfive: Maybe have this function just expect the session rather than an object which expects a session property.

We never seem to be using any other properties so feels overkill.

src/apps/oauth/controllers.js
```diff
@@ -4,6 +4,7 @@ const uuid = require('uuid')
 
 const { get, set } = require('lodash')
 
+const { saveSession } = require('./../../lib/session-helper')
```
teneightfive: Maybe just call it `session` instead of `session-helper`?

`lib` + `helper` feels confusing and we're not quite sure what it is.

feedmypixel: yeah started with `session` but that felt like it belonged to `express-session` so changed it to `session-helper`

src/apps/oauth/controllers.js
```diff
@@ -74,8 +75,13 @@ function redirectOAuth (req, res) {
     set(urlParams, 'code', oAuthDevToken)
   }
 
-  set(req.session, 'oauth.state', stateId) // used to check the callback received contains matching state param
-  return res.redirect(`${config.oauth.url}?${queryString.stringify(urlParams)}`)
+  try {
+    set(req.session, 'oauth.state', stateId) // used to check the callback received contains matching state param
```
teneightfive: Can we move this comment to the line above? I thought our eslint prevented inline comments?

PR #1240, opened 2018-02-08
add comment about dev oauth token

feedmypixel:
Add a comment to explain what this work does so it makes sense for others looking over this code to implement SSO

FYI @web-bert


src/apps/oauth/controllers.js
```diff
@@ -70,6 +70,7 @@ function redirectOAuth (req, res) {
   // SSO journey
   const oAuthDevToken = get(config, 'oauth.devToken')
 
+  // When the OAUTH_DEV_TOKEN env is set then pass this token through the SSO process
```
web-bert: Add "When using Mock SSO" ?

PR #1242, opened 2018-02-08
Remove TAP status field

yeahfro:
Change removes TAP status field for this release. This will be added back in once the BE has been deployed.


feedmypixel:
@reupen is going to do a BE release so no need for this

yeahfro:
Closing, not required

PR #1245, opened 2018-02-08
DH-1485 Add Status to details screen

yeahfro:
DH-1485

This change includes:
- adding field to details screen
- refactoring transformer
- refactoring associated tests
- update wording on form field + remove hint

Note that this work is essentially toggled off because the `Status` field is not enabled/visible.

### Updates
- TAP related fields will only be showing when a TAP service is selected

# Details view

screen shot 2018-02-13 at 09 42 38


src/apps/interactions/transformers.js
```diff
@@ -109,63 +131,46 @@ function transformInteractionResponseToViewRecord ({
...
-      url: config.archivedDocumentsBaseUrl + archived_documents_url_path,
-      name: 'View files and documents',
-      hint: '(will open another website)',
-      hintId: 'external-link-label',
-    }
+  if (kind === 'service_delivery') {
+    return mapKeys({
+      company: transformEntityLink(company, 'companies'),
+      contact: transformEntityLink(contact, 'contacts'),
+      dit_team: dit_team,
```
teneightfive: Can these properties be shorthand?

yeahfro: It did cross my mind but then you will have shorthand properties mixed in (rather than at the top). Order is important.

src/apps/interactions/labels.js
```diff
@@ -18,11 +20,13 @@ const serviceDelivery = {
   date: 'Date of service delivery',
   dit_adviser: 'DIT adviser',
   service: 'Service',
-  service_delivery_status: 'Trade Access Program (TAP) status',
+  service_delivery_status: 'Status',
```
teneightfive: What's the reason for removing the TAP part of this label? It's being displayed within the view so seems like it should be consistent?

src/apps/interactions/transformers.js
ztolley: I'd shorthand them but keep the order as has been done elsewhere

src/apps/interactions/transformers.js
```diff
@@ -109,63 +131,46 @@ function transformInteractionResponseToViewRecord ({
...
-    'Service': service,
-    'Subject': subject,
-    'Notes': notes,
-    'Interaction date': {
+  return mapKeys({
+    company: transformEntityLink(company, 'companies'),
+    contact: transformEntityLink(contact, 'contacts'),
+    dit_team: dit_team,
+    service: service,
+    subject: subject,
```
ztolley: Same comment as earlier about shorthand

src/apps/interactions/hints.js
```diff
@@ -1,7 +0,0 @@
-const serviceDelivery = {
-  service_delivery_status: 'Only applies for TAP services',
```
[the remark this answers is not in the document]
yeahfro: We will only be displaying the fields for TAP services.

src/apps/interactions/transformers.js
```diff
@@ -109,63 +131,42 @@ function transformInteractionResponseToViewRecord ({
...
-    'Service': service,
-    'Subject': subject,
-    'Notes': notes,
-    'Interaction date': {
+  const transformed = kind === 'service_delivery' ? {
+    company: transformEntityLink(company, 'companies'),
+    contact: transformEntityLink(contact, 'contacts'),
+    dit_team,
+    service,
+    service_delivery_status: service_delivery_status,
```
teneightfive: Shorthand like the others?

src/apps/interactions/transformers.js
```diff
@@ -109,63 +131,42 @@ function transformInteractionResponseToViewRecord ({
...
-
-  const viewRecord = {
-    'Company': companyLink,
-    'Contact': contactLink,
-    'Service provider': dit_team,
-    'Service': service,
-    'Subject': subject,
-    'Notes': notes,
-    'Interaction date': {
+  const transformed = kind === 'service_delivery' ? {
```
teneightfive: Could this be restructured so instead of being a ternary we have a set of default and override those using `assign` for service deliveries.

Found these two large objects nested within a ternary quite hard to read.

teneightfive: Also feels like there's a lot of duplication between the two objects so an assign might reduce the lines of code too.

PR #1251, opened 2018-02-12
Fix rounding errors during OMIS reconciliation

teneightfive:
It appears the previous fix hadn't caught all possibilities.

It was using `parseInt` which would round down rather than to the
nearest whole number.

This caused the API to return a possible value mismatch on the
total cost of the order.

Using `Math.round` ensures it will round to the closest integer and
avoid the API returning a mismatch in values.


src/apps/omis/apps/edit/controllers/payment-reconciliation.js
```diff
@@ -75,7 +75,7 @@ class EditPaymentReconciliationController extends EditController {
     const data = req.form.values
 
     // convert pounds to pence
```
yeahfro: To me this comment is a smell and the line of code below could be self documenting

PR #1253, opened 2018-02-12
Adviser filter

ztolley:
Demonstrate autocomplete/multi-select filter for instances that have long filter lists and demonstrated using adviser in the interactions page.


assets/stylesheets/components/form/_form-group.scss
```diff
@@ -152,7 +152,7 @@ fieldset.c-form-group.has-error {
   .c-form-group__inner {
     background: $white;
     clear: left;
-    margin: 0 -4px -4px;
+    margin: 0;
```
teneightfive: This is actually there to align the form elements within a multi-select with the other form elements. Rather than aligning the white space with the other form elements. It's also based on the pattern used on GOV.UK:

image

PR #1255, opened 2018-02-13
DH-1486 Display Grant offered field on service delivery details

yeahfro:
DH-1486

This makes changes to the following:
- consolidates `formatCurrency` as it was being defined in multiple locations
- introduces `Grant offered` to the service delivery details screen

screen shot 2018-02-13 at 12 57 25

### DIFF
--- a/src/apps/investment-projects/transformers/project.js
+++ b/src/apps/investment-projects/transformers/project.js
@@ -3,7 +3,7 @@ const { assign, castArray, get, isEmpty, isPlainObject, mapValues } = require('l
 const format = require('date-fns/format')
 const moment = require('moment')
 
-const { formatCurrency, getInvestmentTypeDetails } = require('./shared')
+const { getInvestmentTypeDetails } = require('./shared')
 const { transformDateObjectToDateString } = require('../../transformers')
 
 function transformToApi (body) {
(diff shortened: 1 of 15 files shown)


src/apps/investment-projects/transformers/project.js
```diff
@@ -173,7 +173,10 @@ function transformBriefInvestmentSummary (data) {
     uk_region_locations: regionLocations.map(region => region.name).join(', '),
     competitor_countries: competitorCountries.map(country => country.name).join(', '),
     estimated_land_date: !isEmpty(data.estimated_land_date) ? moment(data.estimated_land_date, 'YYYY-MM-DD').format('MMMM YYYY') : null,
-    total_investment: formatCurrency(data.total_investment),
+    total_investment: data.total_investment ? {
+      type: 'currency',
+      name: data.total_investment,
```
ztolley: Having to use `name` for a value bugs the living daylights out of me still. I have a separate piece of work to correct this that I think I need to play to help with my OCD

src/templates/_components/key-value-table.njk
```diff
@@ -43,6 +43,8 @@
             {% elif data | isPlainObject %}
               {% if data.type == "date" %}
                 {{ data.name | formatDate }}
+              {% elif data.type == "currency" %}
+                {{ data.name | formatCurrency }}
```
ztolley: 🥇

src/apps/investment-projects/transformers/project.js
teneightfive: @ztolley I think we'd be better swapping out those old components for the answer summary style too with that change. And moving away from the striped tables with green/grey buttons.

Then look at extending the answers summary component to support types. Or maybe even just a `filters` property
which applieswhich

PR #1256, opened 2018-02-13
Updated styles and markup for collection lists

ztolley:
* Updates the collection list layout to put data stacked on the left and badge in the top right.
* Update card contents in response to user research

---------------------------------
### Companies - Before
companies-old

### Companies - After
companies-new

-----------------------------------------
### Contacts - Before
contacts-old

### Contacts - After
contacts-new

-----------------------------------------
### Events - Before
events-old

### Events - After
events-new

------------------------------------------
### Interactions - Before
interaction-old

interactions-new
Interactiopns - After

----------------------------------------
### Investments - Before
investment-old

### Investments - After
invest new

------------------------------------------
### Add Company - Before
add-company-old

### Add Company - After
add-company-new

------------------------------------------

### Mobile 

invest mobile

------------------------------------------
- [x] Review with team to check requested changes have been made
- [x] Test Tablet sizing
- [x] Test phone sizing
- [x] Test IE
- [x] Test Chrome
- [x] Test Firefox
- [x] Test Edge
- [x] Test with company hierarchy data
- [x] Test android
- [x] Test ios
- [x] Update unit tests
- [x] Update acceptance tests
- [x] Test variation used in add company flow
- [x] Create screenshots to show differences

Follow up: 
* Centralise the acceptance test code around setting the sort order and testing results sorted by updated date.
* Discovered that tests for sorts that don't have a reverse available cannot work with the current selectors as the selector for a second item when combined with the xpath created for a meta item causes the first item to be returned.  The selector for meta items need to have a parameter added to specify which result index to use


feedmypixel:
@ztolley the checklist you have here. It feels like it would be good to get it into our [PULL_REQUEST_TEMPLATE.md](https://github.com/uktrade/data-hub-frontend/blob/develop/.github/PULL_REQUEST_TEMPLATE.md)

assets/stylesheets/settings/_colours.scss
```diff
@@ -2,7 +2,7 @@ $govuk-blue: #005ea5;
 $dit-colour: #cf102d;
 
 // Greyscale
-$black: #0b0c0c;
+$black: #000;
```
teneightfive: I'd leave this as it was. I believe the existing colour isn't black black for a reason. I think it has something to do with `#000` on `#fff` being harder to read.

ztolley: Currently, the site is inconsistent, some places use #000 and others use $black. I found #000 to be easier to read and this makes the colour consistent at least.

ztolley: Will revert it though. Just realised that it's still inconsistent as some places use $text-colour which is also slightly off black

teneightfive: I think we should consistently use this value if we need _black_ and use the `$black` variable as the value for `$text-colour`. We should refrain from setting colours like `#000` within sass files.

Feels like that could be a separate change to standardise that if we feel it's an issue.

ztolley: agreed

assets/stylesheets/components/_entity.scss
```diff
@@ -1,5 +1,9 @@
 @import "../tools";
 
+.c-entity__header {
+  margin: 0 0 $default-spacing-unit;
```
teneightfive: Feels like these kind of spacings should in assemblies. I'm also get conscious about just applying `margin-top` and `margin-bottom`.

assets/stylesheets/components/_entity.scss
```diff
@@ -17,13 +21,34 @@
...
 
+.c-entity__subtitle {
+  @include core-font(14);
+  clear: both;
+  color: $grey-1;
+}
+
+.c-entity__badges {
+  margin: 0;
+  @include core-font(14);
```
teneightfive: `@include` statements should come first to make it easier to scan declarations.

src/templates/_macros/entity/entity.njk
```diff
@@ -19,7 +19,7 @@
     {% set urlPrefix = props.urlPrefix or props.type | pluralise + '/' %}
     {% set url = props.url | default('/' + urlPrefix + props.id) %}
     {% set metaBadges = props.meta | filter(['type', 'badge']) %}
-    {% set contentMetaModifier = props.contentMetaModifier| default(['inline', 'split']) %}
+    {% set contentMetaModifier = props.contentMetaModifier| default(['stacked']) %}
```
teneightfive: It looks like this modifier has been removed from the sass file? Do we still need to set this modifier?

Maybe keep the modifier as have the default still table display?

assets/stylesheets/components/_entity.scss
```diff
@@ -17,13 +21,34 @@
...
 @include media(tablet) {
   .c-entity__header,
   .c-entity__content {
     @include clearfix;
   }
 
-  .c-entity__title {
+  .c-entity__title,
+  .c-entity__subtitle {
     float: left;
```
teneightfive: Maybe only apply these floats on tablet and above?

assets/stylesheets/components/_entity.scss
```diff
@@ -17,13 +21,34 @@
...
+.c-entity__subtitle {
+  @include core-font(14);
+  clear: both;
+  color: $grey-1;
+}
+
+.c-entity__badges {
+  margin: 0;
+  @include core-font(14);
+  text-align: right;
```
teneightfive: Maybe look at the small screen display again and stack items like the badges and header rather than having these align right at all resolutions.

src/templates/_macros/entity/entity.njk
```diff
@@ -28,29 +28,30 @@
...
+        <h3 class="c-entity__title">
+          {% if props.id and not props.isBlockLink %}
+            <a href="{{ url }}">{{ highlightedName }}</a>
+          {% else %}
+            {{ highlightedName }}
+          {% endif %}
+        </h3>
+
+        {% if props.modified_on %}
+          <div class="c-entity__subtitle">Updated on {{ props.modified_on | formatDate('D MMM YYYY, h:mma') }}</div>
```
teneightfive: `props.modified_on` isn't documented in this macro. I'd also be tempted to stick with the same camel case naming convention used within all the other macros rather than snake case.

teneightfive: I think there's also a `formatDatetime` filter which will use the default format set within the application config and help keep our date/datetime formats consistent.

src/apps/companies/transformers/company-to-list-item.js
```diff
@@ -89,6 +83,7 @@ module.exports = function transformCompanyToListItem ({
     name,
     url,
     meta,
+    modified_on,
```
teneightfive: Some comments around this property name in the other commit. Will need to change this if you update the macro.

src/apps/events/transformers.js
```diff
@@ -12,47 +12,29 @@ function transformEventToListItem ({
...
         value: get(event_type, 'name'),
       },
       {
-        label: 'Updated',
-        type: 'datetime',
-        value: modified_on,
-      },
-      {
-        label: 'Begins',
+        label: 'Event date',
```
teneightfive: What about being more dynamic and appending the end date if it exists?

So for a single date it would read:

```
Event date: 12 March 2018
``` 

and an event over more days would read:

```
Event date: 12 March 2018 to 14 March 2018
```

src/apps/events/transformers.js
```diff
@@ -12,47 +12,29 @@ function transformEventToListItem ({
...
 
   const item = {
     id,
-    type: 'event',
     name,
+    modified_on,
+    type: 'event',
     meta: [
       {
         label: 'Type',
```
teneightfive: What about making the _type_ a badge? Noticed its a fixed list and seems to be something worth scanning similar to type of interaction?

src/apps/investment-projects/transformers/collection.js
```diff
@@ -30,7 +31,6 @@ function transformInvestmentProjectToListItem ({
     id,
     name,
     type: 'investment-project',
-    code: project_code,
```
teneightfive: Not completely convinced by moving this into the standard list of meta. Do people ever search on it? What's the point of the _code_? If it's a way to remember a project shouldn't we make it more prominent and allow it to be searched/filtered on?

src/apps/investment-projects/transformers/collection.js
```diff
@@ -6,19 +6,20 @@ const labels = require('../labels')
...
   investor_company,
   estimated_land_date,
   sector,
+  project_code,
 }) {
   const metaItems = [
     { key: 'stage', value: stage, type: 'badge' },
     { key: 'investment_type', value: investment_type, type: 'badge', badgeModifier: 'secondary' },
     { key: 'investor_company', value: investor_company },
     { key: 'estimated_land_date', value: estimated_land_date, type: 'dateMonthYear', isInert: true },
```
teneightfive: Might be worth looking at the order of these meta items.

I would put ones that aren't always available beneath ones that are so that it's easier to scan.

Notice how estimated land pops in and out of the middle of these meta items:

image

src/templates/_macros/entity/entity.njk
ztolley: The default is now stacked so I'll remove this

assets/stylesheets/components/_entity.scss
ztolley: yup I like the stacked mobile suggestion

src/apps/investment-projects/transformers/collection.js
ztolley: I'll move optional to the bottom to make scanning easier.

src/apps/events/transformers.js
ztolley: The change requested by the users is to remove the end date. They are only ever interested in the start date in a result

src/apps/investment-projects/transformers/collection.js
ztolley: It is no more special than any other property, yes people search on it and there's a separate ticket to handle highlighting terms in meta data.

src/templates/_macros/entity/entity.njk
ztolley: Actually, that looks like a fixup got lost. I had that in an earlier version but switched to calling the property `subTitle` and outputting that from the transformer to make the code less specific and the subtitle more re-usable.  Also, remember I switched to formatDateTime.

src/apps/companies/transformers/company-to-list-item.js
ztolley: As mentioned in other comment, that should be subTitle, which is then obvious

src/apps/events/transformers.js
ztolley: Type is not displayed, it programmatically indicated the entity type (though I can't remember why we even have it and another PR could remove it from all the transformers)

src/apps/companies/transformers/company-to-list-item.js
teneightfive: The only thing with changing to `subTitle` in the transformer is that you'll need to add the `Update on` text in the transformer, no?

ztolley: I created a macro for subtitle that works like meta fields and supports types, such as date

src/apps/events/transformers.js
teneightfive: Hmm, that may have been when it was 2 separate items though. If you ask a question like "Is there anything you don't need?" users may feel they have to say something so they'd learn towards the end date.

Feels like it wouldn't hurt to combine it as one meta item. Maybe even say how many days a multi day event is: `Event date: 12 to 14 March 2018 (3 days)`. Still provides the main thing people are looking for but gives that extra bit of potentially useful information when the event is spread over 3 days.

Think of if you're looking for a concert or a conference. It would be useful to know if that event was more than just one day. 

I'd just be careful completely removing this piece of information based on a small subset of users asked in a research session.

ztolley: we decided as a team based on the user feedback to just put start date, so we are just putting start date.

src/apps/events/transformers.js
teneightfive: I'm referring to the event type that is a property on the event object. We set `type` on the entity to determine it's display.

I was suggesting displaying the event type in the badge format, like is done with other entities rather than a plain meta item.

This:
image

rather than this:
image

Feels like it's then more consistent with other entity displays:
image

image

src/apps/investment-projects/transformers/collection.js
teneightfive: I think I was thinking more than it's like a badge in that it doesn't need the leading label. The items we display beneath could be a little less self-explanatory hence why they were displayed using the label. 

I feel if an item has a code, the previous display position is easier to scan. Think we could leave this one as it was.

teneightfive: Maybe also check there's a ticket to add the project code as a search facet too.

Might be worth having that _one search box_ which can look for project name, company name, project code, etc. rather than offering lots of facets on the left hand side.

assets/stylesheets/components/_entity.scss
ztolley: The setting the top and bottom margin is an existing rule just moved around slightly.

src/apps/events/transformers.js
teneightfive: Just feels like removing it could hide important information for some users. I think we need to be careful of red herrings like this in research sessions.

If we can provide better information but make it more succinct, why wouldn't we?

src/apps/investment-projects/transformers/collection.js
ztolley: There's already a prototype with a single search box that would search the code and the code is going back in as a subtitle now

assets/stylesheets/components/_entity.scss
ztolley: My worries about assemblies is the style rules for a component are spread out. I didn't even know about assemblies until you mentioned it

src/apps/events/transformers.js
ztolley: People are more interested in the country and region than the type

src/apps/investment-projects/transformers/collection.js
ztolley: Switched sector and the expected land date, plus put project code in sub-title, so scanning is easier

teneightfive: Not sure on project code in subtitle having just looked. 

I think it would be good to be consistent with _Updated on_ location. I think consistency should be a big consideration so when you look in a spot across any entity you expect certain things to be in the same place.

If on companies, contacts and events we used the subtitle for _Updated on_ we should do the same with interactions, investment projects and OMIS orders. At the moment updated on for OMIS orders is in the meta item list and we don't display it for interactions or investment procts.

assets/stylesheets/components/_entity.scss
teneightfive: Am wondering whether we don't float them right at all now looking at the mobile layout. We've noticed issues with some of the titles when they're longer.

Could just go with something like this:

image

src/apps/investment-projects/transformers/collection.js
ztolley: I know what you mean. Updated on doesn't make sense in some record types though.
Your concern is the reason why I put it in meta before but after a discussion with Alberto today I put it in the sub heading.

I'll review it with him and Paul tomorrow and let them decide then will go with that as it's already dragging.

assets/stylesheets/components/_entity.scss
ztolley: I'm keeping them right aligned, they help people scan the page.

src/apps/events/transformers.js
alberto-moro-dit: Hello. I see some discussion around content and design. If you have any concern, please come talk to researcher/designer as we are the closest to the users. Sometimes some decisions don't make sense to me at first, but happen to make a lot of sense for the users.

assets/stylesheets/components/_entity.scss
```diff
@@ -11,49 +10,49 @@
...
     float: left;
+    max-width: 80%;
+  }
+
+  .c-entity__subtitle {
+    clear: both;
   }
 
   .c-entity__badges {
+    text-align: right;
```
teneightfive: Not sure we need the `text-align` property. Removing it doesn't seem to change anything.

assets/stylesheets/components/_meta-list.scss
```diff
@@ -1,16 +1,9 @@
...
-}
-
-.c-meta-list__item {
-  display: inline-block;
-
-  & + & {
-    margin-left: $default-spacing-unit / 2;
-  }
+.c-meta-list__item + .c-meta-list__item {
+  margin-left: 0;
```
teneightfive: Probably don't need this property if the default doesn't apply a left margin.

assets/stylesheets/components/_meta-list.scss
```diff
@@ -45,8 +37,12 @@
   }
 
   .c-meta-list__item {
+    display: inline-block;
     margin-top: $baseline-grid-unit;
-    margin-right: $default-spacing-unit;
+  }
+
+  .c-meta-list__item + .c-meta-list__item {
+    margin-left: $default-spacing-unit / 2;
```
teneightfive: Can't see where this is applied, do you have a link to it on the demo app?

teneightfive: Seems to apply a left margin for meta items nested within a `.c-meta-list__item-value` class which doesn't seem like it would occur?

assets/stylesheets/layout/_assemblies.scss
```diff
@@ -53,3 +53,25 @@
 .c-results__sort + .c-entity-list {
   border-top: 1px solid $grey-2;
 }
+
+.c-entity__title {
+  margin-top: $default-spacing-unit / 4;
```
teneightfive: I would drop this and the margin on line 63 for the badges. Don't think we need that extra 4px. And if we do, apply it to the padding of the list item so it'll always be even, not margin top in 2 places.

assets/stylesheets/layout/_assemblies.scss
```diff
@@ -53,3 +53,25 @@
...
+  margin-bottom: 0;
```
teneightfive: Probably move this to the component as it's spacing internally within that component, not how it should be assembled in the wider layout.

assets/stylesheets/layout/_assemblies.scss
```diff
@@ -53,3 +53,25 @@
...
+}
+
+.c-entity__badges {
+  margin-top: $default-spacing-unit / 4;
```
teneightfive: Can drop this spacing I think.

assets/stylesheets/layout/_assemblies.scss
```diff
@@ -53,3 +53,25 @@
...
+}
+
+.c-entity__content {
+  margin-top: $default-spacing-unit;
```
teneightfive: Can move this to the component style.

assets/stylesheets/layout/_assemblies.scss
```diff
@@ -53,3 +53,25 @@
...
+}
+
+.c-entity--block-link {
```
teneightfive: Can also move this to inside the component.

assets/stylesheets/layout/_assemblies.scss
ztolley: Its needed to match spacing with the title

src/apps/components/controllers.js
```diff
@@ -137,6 +137,24 @@ function renderHiddenText (req, res) {
     .render('components/views/hidden-text')
 }
 
+function renderMetaList (req, res) {
```
teneightfive: Want to add the other variations too?

src/apps/components/views/metalist.njk
```diff
@@ -0,0 +1,10 @@
+{% extends "./_layout.njk" %}
+
+{% block body_main_content %}
+  {% call Example('Stacked meta list, used in header)') %}
```
teneightfive:
```
Meta list with `stacked` modifier (used in header)
```

src/apps/components/views/entity-list.njk
```diff
@@ -12,4 +12,17 @@
       {% endfor %}
     </ul>
   {% endcall %}
+
+  <hr/>
+
+  {% call Example('Entity list of block links') %}
```
teneightfive: Maybe:
```
`entity-list` of `Entity` items with `block-links` modifer
```

Keeps it consistent with other titles that show what class/modifier names are used.

assets/stylesheets/layout/_assemblies.scss
ztolley: Its required to keep it in line with the title

src/apps/events/transformers.js
teneightfive: @alberto-moro-dit just to add another angle to this thought, I just realised what the original code did.

It would show the start date as the end date when there was no end date. So in cases of a 1 day event users would see the same date twice. Which, if asked what wasn't needed it would be understandable that they suggest losing the end date.

So they may have seen something like this which isn't particularly helpful:
```
Begins: 20 March 2018
Ends: 20 March 2018
```

Might be worth seeing if the sample data used had events with multiple dates.

We could show them something like this which may still be of benefit:
```
Event date: 20 March 2018 (1 day)
```
or
```
Event date: 20 to 22 March 2018 (3 days)
```

It may still be less of a concern on the listing pages but something we could still consider when doing to the event detail pages?

src/templates/_macros/entity/entity-sub-title.njk
```diff
@@ -0,0 +1,20 @@
+{% macro SubTitle (subTitle) %}
```
teneightfive: Document the props?

Also maybe call the parameter `props` like other macros?

src/templates/_macros/entity/entity.njk
```diff
@@ -1,9 +1,12 @@
 {% from './meta-list.njk' import MetaList %}
+{% from './entity-sub-title.njk' import SubTitle %}
 
 {##
  # Render entity component
  # @param {object} props
  # @param {string} props.name - entity name
+ # @params{string|object} props.subTitle - Small sub-title below the main link/title such as the updated date
```
teneightfive: Drop `such as the updated date`. Probably don't need to say `Small` either.

src/templates/_macros/entity/entity.njk
```diff
@@ -1,9 +1,12 @@
...
+ #                                         Can either be plain text or an object with a label, value and type
```
teneightfive: Do you even need to say this? Isn't that described by the `{string|object}` part? 

And if you document the props in the other macro you don't need to say what it contains?

src/templates/_macros/entity/entity-sub-title.njk
```diff
@@ -0,0 +1,20 @@
...
+  <div class="c-entity__subtitle">
```
teneightfive: Is a `div` the right element for this? Maybe `h4`?

assets/stylesheets/components/_entity.scss
```diff
@@ -11,49 +11,51 @@
...
 
-  .c-entity__title {
+  .c-entity__title,
+  .c-entity__subtitle {
     float: left;
+    max-width: 80%;
+  }
+
+  .c-entity__subtitle {
+    clear: both;
```
teneightfive: Do you need to clear both? `clear: left` is probably enough.

assets/stylesheets/components/_entity.scss
```diff
@@ -11,49 +11,51 @@
...
-
 // modifiers
 
 .c-entity--block-link {
-  margin-top: -$default-spacing-unit;
-  margin-bottom: -$default-spacing-unit;
-  padding: $default-spacing-unit;
   border: 2px solid $grey-2;
   display: block;
+  padding: $default-spacing-unit;
```
teneightfive: Any reason to move these? Seems to just add noise. Maybe revert to the original position.

src/templates/_macros/entity/entity-sub-title.njk
ztolley: Not used externally, purely used to keep the code clean so didn't document the props

assets/stylesheets/components/_meta-list.scss
```diff
@@ -38,15 +29,33 @@
 }
 
 // Modifiers
+.c-meta-list--stacked {
```
teneightfive: Any reason to move this? Feels like it's creating extra noise in this change.

assets/stylesheets/components/_meta-list.scss
```diff
@@ -38,15 +29,33 @@
...
   }
 
   .c-meta-list__item {
+    display: inline-block;
     margin-top: $baseline-grid-unit;
-    margin-right: $default-spacing-unit;
+  }
+
+  .c-meta-list__item + .c-meta-list__item {
+    margin-top: $baseline-grid-unit;
```
teneightfive: Looking at the example of this not sure we do need the `.c-meta-list--inline` modifier anymore.

If we do I'd drop this extra margin top and remove it from line 53 too.

src/apps/components/views/metalist.njk
```diff
@@ -0,0 +1,26 @@
...
+      items: meta,
+      itemModifier: 'stacked'
+    }) }}
+  {% endcall %}
+
+  {% call Example('Meta list with `inline` modifier') %}
+    {{ MetaList({
+      items: meta,
+      modifier: 'inline',
+      itemModifier: 'inline'
```
teneightfive: Doesn't actually look like this is applied anymore. Look at the components example, it seems to always apply the `stacked` modifier to all `meta-list__item` classes.

PR #1260, opened 2018-02-14
Split responsibility of saving and handling success for create

teneightfive:
The edit controllers split the logic between saving the values
to the API and what to do on success of the form controller.

This change introduces that separation to the create journey too.

### DIFF
--- a/test/unit/apps/omis/apps/create/controllers/confirm.test.js
+++ b/test/unit/apps/omis/apps/create/controllers/confirm.test.js
@@ -103,21 +103,78 @@ describe('OMIS create confirm controller', () => {
     })
   })
 
-  describe('successHandler()', () => {
+  describe('saveValues()', () => {
     beforeEach(() => {
-      this.resetSpy = sandbox.spy()
-      this.destroySpy = sandbox.spy()
+      this.sessionSetSpy = sandbox.spy()
       this.reqMock = Object.assign({}, globalReq, {
         session: {
           token: 'token-12345',
         },
         sessionModel: {
+          set: this.sessionSetSpy,
           toJSON: sandbox.stub().returns({
             'csrf-secret': 'secret-key',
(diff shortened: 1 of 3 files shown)


test/unit/apps/omis/controllers/edit.test.js
```diff
@@ -102,15 +99,12 @@ describe('OMIS EditController', () => {
...
-        locals: {
-          order: { id: updateMockData.id },
-        },
       }
 
       sandbox.stub(FormController.prototype, 'getNextStep').returns(this.nextMock)
     })
 
-    context('when a success message exists', () => {
+    context('when a success message doesn\'t exists', () => {
```
yeahfro: too many `s`

test/unit/apps/omis/controllers/edit.test.js
```diff
@@ -121,8 +115,7 @@ describe('OMIS EditController', () => {
       })
 
       it('should set a flash message', () => {
```
yeahfro: missing `not`

test/unit/apps/omis/apps/create/controllers/confirm.test.js
```diff
@@ -103,80 +103,121 @@ describe('OMIS create confirm controller', () => {
...
+      this.flashSpy = sandbox.spy()
+      this.redirectSpy = sandbox.spy()
+
+      this.reqMock = {
+        form: {
+          options: {},
+        },
+        sessionModel: {
+          get: this.getStub,
           reset: this.resetSpy,
```
yeahfro: If this was me I would have assigned a different `resetSpy` for each model

teneightfive: I was considering that but checking the call count to be twice felt enough to make sure it was called on both methods.

PR #1273, opened 2018-02-19
Extract OMIS validators to separate library

teneightfive:
Extracting the validators will make it easier to maintain and test and
use across other parts of the sub-app if needed.


src/apps/omis/validators.js
```diff
@@ -0,0 +1,39 @@
+const { isArray, flatten } = require('lodash')
+const { Controller } = require('hmpo-form-wizard')
+
+const validators = {
+  isDuration (value) {
+    const regex = /^\d*$/
+
+    if (!isArray(value)) {
```
ztolley: Repeats the code doing the assertion. Split the assertion piece out and then call that from the array foreach and the none array call.

src/apps/omis/validators.js
```diff
@@ -0,0 +1,39 @@
...
+      if (!itemValue.match(regex)) {
+        valid = false
+      }
+    })
+
+    return valid
+  },
+  arrayRequired (value) {
+    const invalidItems = flatten([value]).filter(itemValue => {
+      return itemValue === undefined || itemValue === ''
```
ztolley: should also check for null which is different from undefined? What about using isEmpty or something?

src/apps/omis/validators.js
```diff
@@ -0,0 +1,39 @@
...
+    })
+
+    return invalidItems.length === 0
+  },
+  validEUVatNumber (value) {
+    const regex = /^(ATU[0-9]{8}|BE0[0-9]{9}|BG[0-9]{9,10}|HR[0-9]{11}|CY[0-9]{8}L|CZ[0-9]{8,10}|DE[0-9]{9}|DK[0-9]{8}|EE[0-9]{9}|(EL|GR)[0-9]{9}|ES[0-9A-Z][0-9]{7}[0-9A-Z]|FI[0-9]{8}|FR[0-9A-Z]{2}[0-9]{9}|GB([0-9]{9}([0-9]{3})?|[A-Z]{2}[0-9]{3})|HU[0-9]{8}|IE[0-9]S[0-9]{5}L|IT[0-9]{11}|LT([0-9]{9}|[0-9]{12})|LU[0-9]{8}|LV[0-9]{11}|MT[0-9]{8}|NL[0-9]{9}B[0-9]{2}|PL[0-9]{10}|PT[0-9]{9}|RO[0-9]{2,10}|SE[0-9]{12}|SI[0-9]{8}|SK[0-9]{10})$/
```
ztolley: I'm sure it works, and I'm sure it's the only way but long regex is evil

teneightfive: Yes, I know. If you have a nicer way to split it out I'd love that?

test/unit/apps/omis/validators.test.js
```diff
@@ -0,0 +1,158 @@
...
+    return `${testName(input[0])} with args: ${input.slice(1)}`
+  } else {
+    return `${typeof input} ${input}`
+  }
+}
+
+describe('OMIS form validators', () => {
+  describe('duration()', () => {
+    describe('invalid values', () => {
+      let inputs = [
```
ztolley: const?

Also as mentioned in chat, I find this complicated and unclear. it's doing lots of tests within the `each` block.

I would have added some tests that test the simplest possible thing, which helps when debugging potential errors.

```
    describe('empty duration', () => {
      beforeEach(() => {
        this.result = validators.duration('')
      })

      it('should say invalid', () => {
        expect(this.result).to.eq(false)
      })
    })

    describe('valid duration', () => {
      beforeEach(() => {
        this.result = validators.duration('1')
      })

      it('should say invalid', () => {
        expect(this.result).to.eq(true)
      })
    })

    describe('Multiple valid durations', () => {
      beforeEach(() => {
        this.result = validators.duration(['1', '2'])
      })

      it('should say invalid', () => {
        expect(this.result).to.eq(true)
      })
    })
```

src/apps/omis/validators.js
```diff
@@ -0,0 +1,35 @@
...
+
+const validators = {
+  duration (value) {
+    const regex = /^\d*$/
+
+    if (!isArray(value)) {
+      return Controller.validators.regex(value, regex)
+    }
+
+    let valid = true
```
ztolley:
```
const validValues = value.filter((itemValue) => {
  return Controller.validators.regex(itemValue, regex)
})

return validValues.length > 0
```

test/unit/apps/omis/validators.test.js
teneightfive: I'm not sure. This would mean the test file would be very, very long as you're repeating each it block and each method call.

src/apps/omis/validators.js
teneightfive: I like the thinking but this isn't quite right. This fails when an array contains any valid item. It will return `true` where it should return `false`.

This may work though:

```js
const invalidItems = value.filter((itemValue) => {
  return !Controller.validators.regex(itemValue, regex)
})

return invalidItems.length === 0
```

PR #1281, opened 2018-02-20
Options API endpoints

ztolley:
In preperation for adding the advisers filter component to interactions the node layer needs to provide an AJAX endpoint to lookup adviser options.

As the component that will be used for advisers has the ability to allow lookup and selection for any options, the API endpoint has been created to allow lookup of options for any metadata, though the advisers version differs in that it uses a more complex search and adds a subLabel property to the result.

A typical call would loook like:

`GET /api/options/advisers?term=alb`

Which in turn would return
```
[{
    "value":"5707c18a-454a-4bd2-b14b-218761a684bc",
    "label":"Zac Tolley"
  },
  {
    "value":"56163cae-342d-e411-985c-e4115bead28a",
    "label":"Zahrah Alawami",
    "subLabel":"British Trade Office Al Khobar Saudi Arabia"
  }
]
```

or

`GET /api/options/investment-type`

Which returns:

```
[
  {
    "value":"031269ab-b7ec-40e9-8a4e-7371404f0622",
    "label":"Commitment to invest"
  },
  {
    "value":"3e143372-496c-4d1e-8278-6fdd3da9b48b",
    "label":"FDI"
  },
  {
    "value":"9c364e64-2b28-401b-b2df-50e08b0bca44",
    "label":"Non-FDI"
  }
]
```


src/lib/options.js
```diff
@@ -5,14 +5,21 @@ const authorisedRequest = require('../lib/authorised-request')
...
+async function getOptions (token, key, { createdOn, currentValue, includeDisabled = false, sorted = true, term } = {}) {
   const url = `${config.apiRoot}/metadata/${key}/`
   let options = await authorisedRequest(token, url)
 
   if (!includeDisabled) {
     options = options.filter(filterDisabledOption({ currentValue, createdOn }))
   }
 
+  if (term) {
+    const lterm = term.toLowerCase()
```
teneightfive: `lowercaseTerm`? Better to be more verbose sometimes.

teneightfive: Or just do the `toLowerCase()` within the return where you're doing it on `name` anyway.

teneightfive: Is there any way we can do it on the API too? Feels heavy to do the search on the client.

src/apps/api/controllers/options.js
```diff
@@ -0,0 +1,21 @@
...
+  const key = req.params.entity
+
+  try {
+    const options = await getOptions(token, key, {
+      includeDisabled: true,
+      term: get(req.query, 'term'),
+    })
+
+    res.json(options)
+  } catch (exception) {
```
teneightfive: `catch (error)`? That's the language we've used everywhere else so better to be consistent.

src/apps/api/controllers/options.js
```diff
@@ -0,0 +1,21 @@
...
+const { getOptions } = require('../../../lib/options')
+
+async function getOptionsHandler (req, res, next) {
+  const token = req.session.token
+  const key = req.params.entity
+
+  try {
+    const options = await getOptions(token, key, {
+      includeDisabled: true,
+      term: get(req.query, 'term'),
```
teneightfive: Probably don't need `get` in this instance. Feels over protective because we know express will always provide the `query` property on `req` within a controller.

src/lib/options.js
ztolley: The only allows free text search and not begins with, so this just reduces down that result. I think it's ok here as it's relatively simple

src/apps/adviser/repos.js
```diff
@@ -23,43 +18,43 @@ function getAdviser (token, id) {
...
+  return 0
+}
+
+async function adviserSearch (token, term) {
+  if (isEmpty(term)) {
+    return
+  }
 
-      const data = await authorisedRequest(token, { url })
+  const [firstName, lastName] = term.trim().toLowerCase().split(' ')
```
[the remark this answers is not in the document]
ztolley: How so?

src/apps/api/controllers/advisers.js
```diff
@@ -0,0 +1,20 @@
+const { get } = require('lodash')
+
+const { adviserSearch } = require('../../adviser/repos')
+const { transformAdviserToOption } = require('../../adviser/transformers')
+
+async function getAdviserOptionsHandler (req, res, next) {
+  try {
+    const token = req.session.token
+    const term = get(req.query, 'term')
```
teneightfive: Can drop this `get`. `query` will always exist in `req` with express.

PR #1286, opened 2018-02-23
[DH-1492] Fix handling of blank fields in the investment value form

reupen:
This:
* refactors the investment value form code slightly so that it's more similar to the recently refactored requirements form
* corrects handling of the numeric fields and average salary when they are left blank
* improves test coverage of the form

Saving a blank value form is meant to work in the prospect stage, because investment projects use a save and carry on later model.

The errors were mostly happening due to empty strings and empty objects being sent for numerical fields and foreign keys, which the API doesn't allow. Incorrect handling of blank numerical fields seems to be a recurring problem, so it's begging for a more general solution. This fixes this specific form, though.

## Before

value-form-before

## After

value-form-after

image

### DIFF
--- a/src/apps/investment-projects/transformers/value.js
+++ b/src/apps/investment-projects/transformers/value.js
@@ -76,16 +76,22 @@ function transformInvestmentValueForView ({
 }
 
 function transformInvestmentValueFormBodyToApiRequest (props) {
-  const newProps = {
-    average_salary: {
-      id: props.average_salary,
-    },
-    total_investment: props.client_cannot_provide_total_investment === 'true' ? null : props.total_investment,
-    foreign_equity_investment: props.client_cannot_provide_foreign_investment === 'true' ? null : props.foreign_equity_investment,
-  }
+  const totalInvestment = props.client_cannot_provide_total_investment === 'true'
+    ? null
+    : props.total_investment || null
+
+  const foreignEquityInvestment = props.client_cannot_provide_foreign_investment === 'true'
+    ? null
+    : props.foreign_equity_investment || null
+
   return {
     ...props,
-    ...newProps,
+    ...{
+      total_investment: totalInvestment,
+      foreign_equity_investment: foreignEquityInvestment,
+      number_new_jobs: props.number_new_jobs || null,
+      number_safeguarded_jobs: props.number_safeguarded_jobs || null,
+    },
   }
 }
 
--- a/src/apps/investment-projects/transformers/value.js
+++ b/src/apps/investment-projects/transformers/value.js
@@ -75,6 +75,20 @@ function transformInvestmentValueForView ({
   }
 }
 
+function transformInvestmentValueFormBodyToApiRequest (props) {
+  const newProps = {
+    average_salary: {
+      id: props.average_salary,
+    },
+    total_investment: props.client_cannot_provide_total_investment === 'true' ? null : props.total_investment,
+    foreign_equity_investment: props.client_cannot_provide_foreign_investment === 'true' ? null : props.foreign_equity_investment,
+  }
+  return {
+    ...props,
+    ...newProps,
+  }
+}
+
 function transformAssociatedProject ({
   id,
   associated_non_fdi_r_and_d_project,
@@ -100,4 +114,7 @@ function transformAssociatedProject ({
   }
 }
 
-module.exports = { transformInvestmentValueForView }
+module.exports = {
+  transformInvestmentValueForView,
+  transformInvestmentValueFormBodyToApiRequest,
+}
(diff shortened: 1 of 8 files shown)


test/unit/apps/investment-projects/transformers/value.test.js
```diff
@@ -263,4 +266,123 @@ describe('Investment project transformers', () => {
...
+          fdi_value: '38e36c77-61ad-4186-a7a8-ac6a1a1104c6',
+          government_assistance: 'false',
+          r_and_d_budget: 'false',
+          non_fdi_r_and_d_budget: 'false',
+          new_tech_to_uk: 'false',
+          export_revenue: 'false',
+        }
+      })
+      it('sets the total investment value to null', () => {
+        const transformedData = transformInvestmentValueFormBodyToApiRequest(this.formData)
```
ztolley: Maybe move this into the beforeEach and let the `it` concentrate on testing the result

test/unit/apps/investment-projects/transformers/value.test.js
```diff
@@ -263,4 +266,123 @@ describe('Investment project transformers', () => {
...
+          number_safeguarded_jobs: '100',
+          fdi_value: '38e36c77-61ad-4186-a7a8-ac6a1a1104c6',
+          government_assistance: 'false',
+          r_and_d_budget: 'false',
+          non_fdi_r_and_d_budget: 'false',
+          new_tech_to_uk: 'false',
+          export_revenue: 'false',
+        }
+      })
+      it('transforms and returns the entire object', () => {
```
ztolley: Generally, we leave a blank line between tests etc

src/apps/investment-projects/transformers/value.js
```diff
@@ -76,12 +76,25 @@ function transformInvestmentValueForView ({
...
-      id: props.average_salary,
-    },
-    total_investment: props.client_cannot_provide_total_investment === 'true' ? null : props.total_investment,
-    foreign_equity_investment: props.client_cannot_provide_foreign_investment === 'true' ? null : props.foreign_equity_investment,
+    total_investment: totalInvestment,
+    foreign_equity_investment: foreignEquityInvestment,
+    number_new_jobs: props.number_new_jobs || null,
+    number_safeguarded_jobs: props.number_safeguarded_jobs || null,
   }
   return {
```
ztolley: I like the spread operator but maybe see what the others think before we start to adopt it or see if we want to keep using `assign`

src/apps/investment-projects/transformers/value.js
```diff
@@ -75,6 +75,33 @@ function transformInvestmentValueForView ({
...
+  const newProps = {
+    total_investment: totalInvestment,
+    foreign_equity_investment: foreignEquityInvestment,
+    number_new_jobs: props.number_new_jobs || null,
+    number_safeguarded_jobs: props.number_safeguarded_jobs || null,
+  }
+  return {
+    ...props,
+    ...newProps,
+  }
```
feedmypixel: This could be condensed quite a bit. And using `assign` means you don't have to store the newProbs in a const

A suggestion would be:

```
const totalInvestment = props.client_cannot_provide_total_investment === 'true' 
  ? null 
  : get(props, 'total_investment', null)

const foreignEquityInvestment = props.client_cannot_provide_foreign_investment === 'true' 
  ? null 
  : get(props, 'foreign_equity_investment', null)

return assign({}, props, {
  total_investment: totalInvestment,
  foreign_equity_investment: foreignEquityInvestment,
  number_new_jobs: props.number_new_jobs || null,
  number_safeguarded_jobs: props.number_safeguarded_jobs || null,
})
```

src/apps/investment-projects/transformers/value.js
```diff
@@ -75,6 +75,26 @@ function transformInvestmentValueForView ({
...
+  const totalInvestment = props.client_cannot_provide_total_investment === 'true'
+    ? null
+    : props.total_investment || null
+
+  const foreignEquityInvestment = props.client_cannot_provide_foreign_investment === 'true'
+    ? null
+    : props.foreign_equity_investment || null
+
+  return {
+    ...props,
```
feedmypixel: 👍

src/apps/investment-projects/transformers/value.js
```diff
@@ -75,6 +75,26 @@ function transformInvestmentValueForView ({
...
+    ...{
```
feedmypixel: 👍

src/apps/investment-projects/transformers/value.js
ztolley: You don't have to store them in a constant for the spread operator either, both assign and spread benefit from it being in it's on var to be readable though

src/apps/investment-projects/transformers/value.js
ztolley: Do you even need the spread on the second part?

reupen: > Do you even need the spread on the second part?

Yes, you need ... to unpack the object.

PR #1287, opened 2018-02-23
Improve meta lists

ztolley:
The behaviour for meta lists was somewhat unpredictable and the use of multiple modifiers in search results was simply wrong and made things more complex and harder to debug.

As part of improving the layout for search results, the metadata lists first need to be cleaned up

This PR :
* Adds examples of meta lists and entity lists to the component library
* Cleans up the sass for metadata to be simpler and more predictable
* Removes the use of defaulting to multiple modifiers for meta lists
* Makes it explicit what modifiers to use in collection lists and investment details

[View meta list component](https://datahub-beta2-pr-1287.herokuapp.com/components/meta-list)

### DIFF
--- a/assets/stylesheets/components/_meta-list.scss
+++ b/assets/stylesheets/components/_meta-list.scss
@@ -84,18 +84,16 @@
   }
 
   @include media(tablet) {
-    .c-meta-list__item {
-      max-width: 50%;
-    }
-
     .c-meta-list__item:nth-child(odd) {
       float: left;
       clear: both;
+      max-width: 50%;
     }
 
     .c-meta-list__item:nth-child(even) {
       float: right;
       text-align: right;
+      max-width: 50%;
     }
   }
 }
--- a/assets/stylesheets/components/_meta-list.scss
+++ b/assets/stylesheets/components/_meta-list.scss
@@ -1,15 +1,9 @@
 @import "../settings";
 @import "../tools";
 
-.c-meta-list {
-  display: table;
-}
-
 .c-meta-list__item {
-  display: inline-block;
-
   & + & {
-    margin-left: $default-spacing-unit / 2;
+    margin-top: $default-spacing-unit / 4;
   }
 }
 
@@ -20,7 +14,6 @@
 
 .c-meta-list__item-value {
   color: $text-colour;
-  display: inline-block;
   word-break: break-all;
 
   &:link,
@@ -40,13 +33,31 @@
 // Modifiers
 
 .c-meta-list--inline {
+  display: inline-block;
+  margin-top: -($default-spacing-unit / 2);
+
   .c-meta-list__item-label {
     display: inline-block;
   }
 
   .c-meta-list__item {
-    margin-top: $baseline-grid-unit;
-    margin-right: $default-spacing-unit;
+    display: inline-block;
+    margin-right: $default-spacing-unit * 2;
+    margin-top: $default-spacing-unit / 2;
+
+    &:last-child {
+      margin-right: 0;
+    }
+  }
+}
+
+.c-meta-list--condensed {
+  &.c-meta-list--inline .c-meta-list__item {
+    margin-right: $default-spacing-unit / 4;
+
+    &:last-child {
+      margin-right: 0;
+    }
   }
 }
 
@@ -54,18 +65,7 @@
   display: block;
 
   & + & {
-    margin-left: 0;
     margin-top: $default-spacing-unit / 2;
-  }
-
-  @include media(tablet) {
-    display: inline-block;
-    vertical-align: top;
-
-    & + & {
-      margin-left: $default-spacing-unit * 2;
-      margin-top: 0;
-    }
   }
 
   .c-meta-list__item-label {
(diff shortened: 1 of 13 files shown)


src/apps/components/controllers.js
```diff
@@ -137,6 +137,36 @@ function renderHiddenText (req, res) {
...
+        { label: 'Valuation', value: 'Not yet valued' },
+        { label: 'Created on', value: '5 Jan 2016, 3:00pm' },
+      ],
+      metaStacked: [
+        {
+          label: 'Status',
+          value: 'Ongoing',
+          url: '/investment-projects/5d341b34-1fc8-4638-b4b1-a0922ebf401e/status',
+          urlLabel: 'change',
+          modifier: 'stacked',
```
teneightfive: Do you need to pass a modifier to each item? Feels like that's what the `itemModifier` prop on `metaList` should do.

teneightfive: And if you really do, you could map the other object so you don't have as much duplication:

```
metaStacked: meta.map(item => {
  return assign({}, item, {
    modifier: 'stacked',
  })
})
```

ztolley: Yes normally you would use item modifier, but I found a bug where when you display data it modifies the original item, even though it uses assign. As there are no instances where the same data is used on the screen with different modifiers, except for the component library, this was the only fix available.

ztolley: AssignCopy fixed it.

assets/stylesheets/components/_meta-list.scss
```diff
@@ -1,16 +1,8 @@
...
-  display: table;
-}
-
 .c-meta-list__item {
-  display: inline-block;
-
-  & + & {
-    margin-left: $default-spacing-unit / 2;
-  }
+  margin-bottom: $default-spacing-unit / 2;
```
teneightfive: I would change this to a sibling selector. Otherwise you'll get bottom margin on every item and if its container has padding you'll get unwanted extra spacing.

This would be cleaner:

```scss
.c-meta-list__item {
  & + & {
  }
}
```

assets/stylesheets/components/_meta-list.scss
```diff
@@ -40,33 +31,46 @@
...
+
   .c-meta-list__item-label {
     display: inline-block;
   }
 
   .c-meta-list__item {
-    margin-top: $baseline-grid-unit;
-    margin-right: $default-spacing-unit;
+    display: inline-block;
+    margin-right: $default-spacing-unit * 2;
```
teneightfive: I would also use a sibling selector to apply a left margin rather than blanket apply a right margin.

assets/stylesheets/components/_meta-list.scss
```diff
@@ -40,33 +31,46 @@
...
-    margin-top: $default-spacing-unit / 2;
+  .c-meta-list__item-label {
+    display: inline-block;
   }
 
-  @include media(tablet) {
+  .c-meta-list__item {
     display: inline-block;
-    vertical-align: top;
+    margin-right: $default-spacing-unit * 2;
```
teneightfive: Same here with the blanket margin. Use a sibling selector to only apply left margin when needed.

assets/stylesheets/components/_meta-list.scss
```diff
@@ -40,33 +31,46 @@
...
+  }
+}
 
-    & + & {
-      margin-left: $default-spacing-unit * 2;
-      margin-top: 0;
-    }
+.c-meta-list--inline-condensed {
```
teneightfive: Seems this duplicates `.c-meta-list--inline`. 

What about this to have less duplication:

```scss
.c-meta-list--inline,
.c-meta-list--inline-condensed {
  display: inline-block;

  .c-meta-list__item-label {
    display: inline-block;
  }

  .c-meta-list__item {
    display: inline-block;
    margin-right: $default-spacing-unit * 2;
  }
}

.c-meta-list--inline-condensed .c-meta-list__item {
  margin-right: $default-spacing-unit / 4;
}
```

teneightfive: Or maybe even just a `.c-meta-list--condensed` modifier that can reduce the spacing based on its sibling modifiers.

Then it can be applied to all kinds to reduce spacing.

assets/stylesheets/components/_meta-list.scss
```diff
@@ -92,19 +95,7 @@
     .c-meta-list__item:nth-child(even) {
       float: right;
       margin-right: 0;
+      text-align: right;
```
teneightfive: I notice this side doesn't have a `max-width` set which means it can sometimes looks strange when the right hand side stretches wider than halfway when the left hand side is shorter.

We should probably apply `max-width: 50%` to both sides.

src/templates/_macros/entity/meta-item.njk
```diff
@@ -15,6 +15,7 @@
  # @param {string} [props.isLabelHidden=false] - whether the label should be visually hidden
  # @param {string} [props.badgeModifier] - modifier for badge
  # @param {string} [props.highlightTerm] - text to use to apply highlight filter
+ # @param {string|array} [props.modifier] - Modifier for the outer class, e.g. stacked
```
teneightfive: `item modifier (e.g. 'stacked')` to keep inline with other prop description?

assets/stylesheets/components/_meta-list.scss
ztolley: The top-level condensed one would lead to more complexity I fear. I like the first suggestion though

assets/stylesheets/components/_meta-list.scss
```diff
@@ -40,33 +33,36 @@
...
-    margin-top: $baseline-grid-unit;
-    margin-right: $default-spacing-unit;
-  }
-}
-
-.c-meta-list__item--stacked {
-  display: block;
+    display: inline-block;
+    margin-right: $default-spacing-unit * 2;
+    margin-top: 0;
```
teneightfive: Think this should actually be: `margin-top: $default-spacing-unit / 2;`

PR #1288, opened 2018-02-23
Feature/stacked collections

ztolley:
DH-1546 - Migrate collection display from split to stacked style

* Drops the use of the split modifier for collection metadata
* Introduces updated entity macro which removes code element and introduces subtitle
* Updates the investment collection to put the project code in the subtitle
* Adjusts font sizes in collections to make slightly clearer
* Correct an issue with underlines when using block link modifier

companies
addcompany
projects
interactions
events
contacts

### DIFF
--- a/assets/stylesheets/components/_entity.scss
+++ b/assets/stylesheets/components/_entity.scss
@@ -55,11 +55,16 @@
   padding: $default-spacing-unit;
   border: 2px solid $grey-2;
   display: block;
+  text-decoration: none;
 
   &:focus,
   &:hover {
     background-color: $grey-4;
     border-color: $link-hover-colour;
     color: $link-hover-colour;
+  }
+
+  .c-entity__title {
+    text-decoration: underline;
   }
 }
--- a/assets/stylesheets/components/_entity.scss
+++ b/assets/stylesheets/components/_entity.scss
@@ -11,39 +11,40 @@
       text-decoration: underline;
     }
   }
+}
 
-  @include media(tablet) {
-    max-width: 80%;
+.c-entity__subtitle {
+  @include core-font(14);
+  color: $grey-1;
+  margin: 0;
+}
+
+.c-entity__badges {
+  @include core-font(14);
+  margin: 20 - 14px 0;
+}
+
+.c-entity__content {
+  @include core-font(16);
+
+  * + & {
+    margin-top: $default-spacing-unit / 2;
   }
 }
 
 @include media(tablet) {
-  .c-entity__header,
-  .c-entity__content {
+  .c-entity__header {
     @include clearfix;
   }
 
-  .c-entity__title {
-    float: left;
+  .c-entity__title,
+  .c-entity__subtitle {
+    max-width: 80%;
   }
 
   .c-entity__badges {
     float: right;
   }
-}
-
-.c-entity__code,
-.c-entity__content,
-.c-entity__badges {
-  @include core-font(14);
-}
-
-.c-entity__badges {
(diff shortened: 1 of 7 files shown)


assets/stylesheets/components/_entity.scss
```diff
@@ -17,35 +17,45 @@
...
+    float: left;
+    max-width: 80%;
+  }
+
   .c-entity__title {
     float: left;
   }
 
+  .c-entity__subtitle {
+    clear: both;
```
teneightfive: Could you just use `clear: left`?

assets/stylesheets/components/_entity.scss
```diff
@@ -64,11 +64,16 @@
...
+  text-decoration: none;
 
   &:focus,
   &:hover {
     background-color: $grey-4;
     border-color: $link-hover-colour;
     color: $link-hover-colour;
   }
+
+  h3 {
```
teneightfive: Should we be using classname selectors here instead of element selectors?

assets/stylesheets/components/_entity.scss
```diff
@@ -17,35 +17,45 @@
...
-  .c-entity__header,
-  .c-entity__content {
+  .c-entity__header {
     @include clearfix;
   }
 
+  .c-entity__title,
+  .c-entity__subtitle {
+    float: left;
+    max-width: 80%;
```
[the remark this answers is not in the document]
ztolley: As we discussed before, lets look at left aligned badges soon but not now. And this is still better than what we have now.

assets/stylesheets/components/_entity.scss
ztolley: No, needs both otherwise it tries to appear next to the title

assets/stylesheets/components/_entity.scss
```diff
@@ -17,35 +17,45 @@
...
+  color: $grey-1;
+}
+
+.c-entity__badges {
+  @include core-font(14);
+}
+
+.c-entity__content {
+  @include core-font(16);
+  margin-top: $default-spacing-unit / 2;
```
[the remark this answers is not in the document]
ztolley: An entity only ever has one content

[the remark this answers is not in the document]
teneightfive: That wasn't the point I was making. It was using the lobotomized owl thought of only applying top margins to an element when there is an element before it.

That way if it's on its own you don't get double spacing if its container has padding.

assets/stylesheets/components/_entity.scss
teneightfive: I'm not sure it does. `.c-entity__title` is floated left so as long as you clear the left float this element will stack beneath it, even if you remove the badges which are floated right.

image


screen shot 2018-02-26 at 11 38 15

assets/stylesheets/components/_entity.scss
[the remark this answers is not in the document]
ztolley: ok well I'll add it but to me it looks more complicated for a rule that will never be used as you always have one title and one content

assets/stylesheets/components/_entity.scss
ztolley: I tried and in some situations, it appeared next to the title, but I came up with an even better solution.. don't float it.

ztolley: I could even remove the `@include clearfix` from the header as there's always a title in there, but it's staying there just as good practice to always clear something with floats in it.

PR #1295, opened 2018-02-27
DH-1564 Make OAuth more robust

feedmypixel:
This work tackles a number of issues with Oauth. Specifically the `OAuth stateId mismatch sessionOAuthState`

It also guards against:
- Visits to `/oauth/callback/` without a `session.token`
- Visits to `/oauth/callback/` without a `session.oauth.state`
- Visits to `/oauth/callback/` without a `state` query param

This work now means that if there is a `state` already in `session.oauth.state` then it will use that, rather than overriding it.

It removes logging as this is making a mess of the file and isn't really informing us of anything new.

I have also added more tests to all this work.

FYI @web-bert

### DIFF
--- a/src/apps/oauth/controllers.js
+++ b/src/apps/oauth/controllers.js
@@ -44,6 +44,11 @@ async function callbackOAuth (req, res, next) {
   if (get(req.session, 'token')) {
     return res.redirect('/')
   }
+
+  // No state query param
+  if (isUndefined(stateQueryParam)) {
+    return next({ statusCode: 403 })
+  }
   if (errorQueryParam) {
     return renderHelpPage(req, res, next)
   }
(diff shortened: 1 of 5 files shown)


src/apps/oauth/controllers.js
```diff
@@ -40,6 +40,10 @@ async function callbackOAuth (req, res, next) {
   const stateQueryParam = get(req.query, 'state')
   const sessionOAuthState = get(req.session, 'oauth.state')
 
+  // Already been through OAuth
+  if (get(req, 'session.token')) {
```
teneightfive: `res.session.token`? Don't we know `req.session` always exists?

teneightfive: Could also be a piece of middleware?

feedmypixel: > Could also be a piece of middleware?

Could be, however it felt a bit weird splitting it out TBH

teneightfive: Fair enough. Just looked like a real simple piece of middleware to decide whether to redirect or continue rather than adding another if block to the controller doing the rendering.

PR #1296, opened 2018-02-27
DH1-1546 - update fields shown on result cards

ztolley:
* Add phone and email to contact list
* Remove end date from events list
* Remove communication channel from interaction list
* Reorder investment list for easier scanning

### DIFF
--- a/src/apps/contacts/labels.js
+++ b/src/apps/contacts/labels.js
@@ -44,9 +44,10 @@ const contactMetaItemLabels = {
   company_sector: 'Sector',
   address_country: 'Country',
   company_uk_region: 'UK region',
-  modified_on: 'Updated on',
   contact_type: 'Contact type',
   archived_on: 'Status',
+  telephone: 'Phone number',
+  email: 'Email',
 }
 
 module.exports = {
--- a/test/unit/apps/contacts/transformers.test.js
+++ b/test/unit/apps/contacts/transformers.test.js
@@ -88,6 +88,8 @@ describe('Contact transformers', () => {
             { label: 'Job title', value: 'Director' },
             { label: 'Sector', value: 'Aerospace' },
             { label: 'Country', value: 'United Kingdom' },
+            { label: 'Phone number', value: '(+44) 07814 333 777' },
+            { label: 'Email', value: 'fred@test.com' },
             { label: 'Contact type', type: 'badge', value: 'Primary', badgeModifier: 'secondary' },
           ],
         })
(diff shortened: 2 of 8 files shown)


jim68000:
Why don't we have the ticket numbers in the title again? It makes scanning the commits in a release to summarise to the PM a right nightmare

teneightfive:
>Why don't we have the ticket numbers in the title again? It makes scanning the commits in a release to summarise to the PM a right nightmare

@jim68000 The ticket number is relatively meaningless in a scan of the commits though, right? You have to map it back to a ticket that has a description.

I think it's good to have it for reference but the ticket number shouldn't replace a well written PR title or commit message.

src/apps/events/transformers.js
```diff
@@ -44,15 +35,10 @@ function transformEventToListItem ({
         value: get(event_type, 'name'),
       },
       {
-        label: 'Begins',
+        label: 'Event date',
```
teneightfive: Still feel this could be improved to include an end date if it's a multi day event.

This is misleading for multi-day events now as it would only show the start date of the event so there is no way to tell if it goes on for longer. Label would then be more accurate as `First day of event` or `Event start date`. But they don't read well.

Do we actually feel that these examples would be bad:

```
Event date: 12 March 2018
```

```
Event date: 12 to 14 March 2018 (3 days)
```

Also, as previously mentioned, if you look at the previous implementation it could explain why when asked "Are there any bits of information you don't need?" users would answer with _End date_ (it often contained duplicated data). 

I think this is worth considering before removing end date and only displaying start date.

jim68000:
@teneightfive Yeah, I get the general idea that your commits should be clean and easy to read, and adding 7 arbitrary characters will diminish that. However a typical FE release will have say 12 commits of which maybe two are features the upstream team care about, and the other 10 are equally important but don't directly relate to stories in the sprint (eg, OMIS, AT clean up, resilience things). 

Our PM wants a fine grained sign off of things that are stories, so I'd like to be able to glance at the list of commit titles and say 'oh yeah, this and this you need to look at'.

src/apps/events/transformers.js
ztolley: I know you think that but this is what we agreed

src/apps/contacts/labels.js
```diff
@@ -44,9 +44,10 @@ const contactMetaItemLabels = {
   company_sector: 'Sector',
   address_country: 'Country',
   company_uk_region: 'UK region',
-  modified_on: 'Updated on',
   contact_type: 'Contact type',
   archived_on: 'Status',
+  telephone: 'Telephone',
```
teneightfive: `Phone` and `Email`? `Telephone` feels a little old school and doesn't feel like we need the `address` part of the email label.

teneightfive: If we want consistency, in the actual contact listing we say `Phone number` and `Email`.

test/unit/apps/contacts/transformers.test.js
```diff
@@ -88,6 +88,8 @@ describe('Contact transformers', () => {
             { label: 'Job title', value: 'Director' },
             { label: 'Sector', value: 'Aerospace' },
             { label: 'Country', value: 'United Kingdom' },
+            { label: 'Telephone', value: '(+44) 07814 333 777' },
```
teneightfive: Do these need updating too?

ztolley: :trollface:

PR #1298, opened 2018-02-27
Always show region option

ztolley:
Users are unaware of the region filter and do not realise that they need to select the UK in order to see the region filter.

This change causes the region filter to always be shown.


src/apps/companies/controllers/list.js
```diff
@@ -1,32 +1,24 @@
 const { get, omit, merge } = require('lodash')
-const { companyFiltersFields, companySortForm } = require('../macros')
+const { companyFiltersFields: filtersFields, companySortForm } = require('../macros')
```
teneightfive: Do you really need to do this?

Might be better to just assign `filterFields` to `companyFiltersFields` in the locals object.

ztolley: Makes the call to res cleaner and copies how contacts does it

PR #1299, opened 2018-02-27
Update XHR to only replace results

ztolley:
The XHR module currently updates the entire content area, this can be slow for pages with lots of filters and will not work with the multi-select autocomplete component.

This change modifies the page and templates so that only the results section is replaced.

To make this work a small piece of JS was added on the client side to watch for sort form changes and copy them to the filter form, so when the filter form parsed the current sort order is preserved.


assets/javascripts/modules/filter-sort.js
```diff
@@ -0,0 +1,30 @@
+/*
+ * Listens for change of sort order in the sort form and updates the filters form
+ * to know about the sort order change so it is applied when the filter changes
+*/
+const sortFieldSelector = '.js-filter-sort select[name="sortby"]'
+const filterFormSortField = '.c-collection-filters input[name="sortby"]'
```
teneightfive: Maybe use a `.js-` prefix class rather to target these elements.

src/apps/companies/macros.js
```diff
@@ -33,7 +33,7 @@ const companyFiltersFields = [
 
 const companySortForm = {
   method: 'get',
-  class: 'c-collection__sort-form js-AutoSubmit',
+  class: 'c-collection__sort-form js-AutoSubmit js-filter-sort',
```
teneightfive: `js-FilterSort` to keep the same pattern as elsewhere?

assets/javascripts/modules/filter-sort.js
feedmypixel: yeah agreed `.js-` communicates in the dom that there is `JavaScript` functionality attached to these elements

feedmypixel: Also, why are these variables? They are only used once. Could you just inline them?

ztolley: It was to keep the code neat and easy to see where they are looking

ztolley: The multiple choice control does not pass classes through to it and I was unable to figure out why, so I had to attach the selector to the form. 

I can add one to the collection filters too if you really think this will benefit things

ztolley:
Squashed the commits for macros, changed the case for the class to tag the component and inlined selectors.

As explained, the reason for using the selectors here is due to a bug in the multiple choice macro that has been around for quite a while and not sure how to fix it, plus I feel when we fix it, it should be in its own PR as it will affect almost all of the site.

feedmypixel:
> As explained, the reason for using the selectors here is due to a bug in the multiple choice macro that has been around for quite a while and not sure how to fix it, plus I feel when we fix it, it should be in its own PR as it will affect almost all of the site.

I'm really not a fan of adding code to fix something rather than fixing the underlying problem. In the majority of cases the code to patch a fix at a higher level ends up becoming permanent and the underlying problem doesn't get fixed.

What is the underlying problem? Has it been correctly identified and how difficult is it to fix?

ztolley:
Fixing a bug found in the company contact screen.

ztolley:
The underlying problem is twofold:
1. There is a bug in the multiple choice macro that means you cannot set the class on the select. That bug has been around since the multiple choice macro was written, and both Tyom and I have both tried to figure it out, as has Dom I think. The last fix for the bug had a side effect that caused many problems so it needs to be fixed in isolation.

2. The hidden fields for filter forms generated by simply passing the form macro a list of hidden fields, this does not support the ability to set a class on the hidden field.

Yes, it would be ideal to fix those issues, I just feel that for a small change this is the pragmatic approach.

teneightfive:
>There is a bug in the multiple choice macro that means you cannot set the class on the select. That bug has been around since the multiple choice macro was written, and both Tyom and I have both tried to figure it out, as has Dom I think. The last fix for the bug had a side effect that caused many problems so it needs to be fixed in isolation.

@ztolley see #1300 for a fix for the bug that you mentioned. Sometimes worth trying to understand the root cause rather than covering it up. In the end it was down to the class not being set with the correct value.

>The hidden fields for filter forms generated by simply passing the form macro a list of hidden fields, this does not support the ability to set a class on the hidden field.

The suggestion I gave in slack wouldn't require this functionality. If we use the idea of a data attribute on the field which is changing with a selector to the value that needs updating we may not need to set anything on the hidden field.

It might require adding support for data attributes on the select box but we have something similar in another macro. But could be map of name/value to add data attributes to a field.

feedmypixel:
> Yes, it would be ideal to fix those issues, 

I don't think _ideal_ is the correct word here. I think _necessary_ is more appropriate. It feels like the list you have made is the list of things that need to be done for this work to be merged.

> I just feel that for a small change this is the pragmatic approach.

Adding something at a higher level and not applying a simple fix to the underlying problem. IMO is not _pragmatic_

PR #1301, opened 2018-03-01
Add support for input data attributes

ztolley:
Adds the ability to specify data attributes for a text field, text area or select control

This enhances the capabilities of client-side javascript.

### DIFF
--- a/src/templates/_macros/form/input.njk
+++ b/src/templates/_macros/form/input.njk
@@ -9,6 +9,7 @@
  # @param {string} [props.placeholder] - Field placeholder
  # @param {string, array} [props.modifier] - form-control modifier
  # @param {string} [props.error] - Mark form-control with error
+ # @param {object} [props.data] - data attributes to add to control
  #}
 {% macro Input(props) %}
   {% set modifier = props.modifier | concat('') | reverse | join(' c-form-control--') if props.modifier %}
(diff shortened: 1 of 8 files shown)


src/templates/_macros/form/input.njk
```diff
@@ -22,5 +23,8 @@
     class="c-form-control {{ modifier }} {{ 'has-error' if props.error }} {{ props.class }}"
     {% if props.autofocus %}autofocus{% endif %}
     {% if props.hint %}aria-describedby="hint-{{ props.fieldId }}"{% endif %}
+    {% for key, value in props.data %}
+      data-{{ key }}="{{ value | escape }}"
```
teneightfive: Do you need to escape this? It's a user entered value and our config of nunjucks will only escape values by default.

src/templates/_macros/form/select-box.njk
```diff
@@ -22,6 +23,9 @@
     class="c-form-control {{ modifier }} {{ 'has-error' if props.error }} {{ props.class }}"
     {% if props.persistsConditionalValue %}data-persist-values="true"{% endif %}
     {% if props.hint %}aria-describedby="hint-{{ props.fieldId }}"{% endif %}
+    {% for key, value in props.data %}
+      data-{{ key }}="{{ value | escape }}"
```
teneightfive: Don't think you need to escape this value. Nunjucks will be doing it be default.

src/templates/_macros/form/text-area.njk
```diff
@@ -20,5 +21,8 @@
     {% if props.hint %}aria-describedby="hint-{{ props.fieldId }}"{% endif %}
     {% if props.autofocus %}autofocus{% endif %}
     rows="8"
+    {% for key, value in props.data %}
+      data-{{ key }}="{{ value | escape }}"
```
teneightfive: Don't think you need to escape this value. Nunjucks will be doing it be default.

PR #1302, opened 2018-03-01
DH-1563 solved search term highliting and escaping string input

yozzo:
Solved search term highliting and escaping string input of potentially hazardous characters

### DIFF
--- a/config/nunjucks/filters.js
+++ b/config/nunjucks/filters.js
@@ -91,21 +91,23 @@ const filters = {
     return value.split(separator)
   },
 
-  highlight (string, searchTerm, shouldMatchFullWord = false) {
-    if (!isString(string) || !isString(searchTerm) || !searchTerm.trim()) {
-      return string
+  highlight (searchResultText, searchTerm, matchFullWord = false) {
+    if (!isString(searchResultText) || !isString(searchTerm) || !searchTerm.trim()) {
+      return searchResultText
     }
 
     try {
       // Remove regex characters from the search term
       // as they wont be in the result and will cause an error in the regular expression
-      const cleanTerm = searchTerm.replace(/[.*+?^${}()|[\]\\]/g, '') // $& means the whole matched string
+      const cleanedSearchTerm = searchTerm.replace(/[.*+?^${}()<&|[\]\\]/g, '')
+      const searchPattern = matchFullWord
+        ? new RegExp(`\\b(${cleanedSearchTerm})\\b`, 'gi')
+        : new RegExp(`(${cleanedSearchTerm})`, 'gi')
 
-      const regEx = new RegExp(`(${cleanTerm})${shouldMatchFullWord ? '\\b' : ''}`, 'gi')
-      const result = string.replace(regEx, '<span class="u-highlight">$1</span>')
+      const result = searchResultText.replace(searchPattern, '<span class="u-highlight">$1</span>')
       return new nunjucks.runtime.SafeString(result)
     } catch (error) {
-      return string
+      return searchResultText
     }
   },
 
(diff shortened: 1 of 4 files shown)


config/nunjucks/filters.js
```diff
@@ -102,7 +103,8 @@ const filters = {
       const cleanTerm = searchTerm.replace(/[.*+?^${}()|[\]\\]/g, '') // $& means the whole matched string
 
       const regEx = new RegExp(`(${cleanTerm})${shouldMatchFullWord ? '\\b' : ''}`, 'gi')
-      const result = string.replace(regEx, '<span class="u-highlight">$1</span>')
+      const cleanString = escape(string)
```
feedmypixel: `escapedString` could be a more appropriate name?

config/nunjucks/filters.js
```diff
@@ -102,7 +103,8 @@ const filters = {
       const cleanTerm = searchTerm.replace(/[.*+?^${}()|[\]\\]/g, '') // $& means the whole matched string
 
       const regEx = new RegExp(`(${cleanTerm})${shouldMatchFullWord ? '\\b' : ''}`, 'gi')
-      const result = string.replace(regEx, '<span class="u-highlight">$1</span>')
```
feedmypixel: not part of this work but `string`, IMO is a terrible variable name!
I would go with something which at a look describes what it is, rather than its type. What is `string`? what does it contain?

`textValue`? or `searchResultText`? instead of `string`

config/nunjucks/filters.js
```diff
@@ -102,7 +103,8 @@ const filters = {
       const cleanTerm = searchTerm.replace(/[.*+?^${}()|[\]\\]/g, '') // $& means the whole matched string
```
feedmypixel: I wonder what is the comment here referring to here? should it be removed?

feedmypixel: Can I suggest `cleanedSearchTerm` rather than `cleanTerm`

config/nunjucks/filters.js
```diff
@@ -102,7 +103,8 @@ const filters = {
...
 
       const regEx = new RegExp(`(${cleanTerm})${shouldMatchFullWord ? '\\b' : ''}`, 'gi')
```
feedmypixel: Don't you also need a word boundary flag `\b` at the start?

feedmypixel: Could this could be `matchFullWord` rather than `shouldMatchFullWord`.

It also feels like this is quite complicated, can I suggest:

```
const searchPattern = matchFullWord 
  ? new RegExp(`\\b(${cleanedTerm})\\b`, 'gi')
  : new RegExp(`(${cleanedTerm})`, 'gi')

const escapedString = escape(string)
const resultWithHighlight = escapedString.replace(searchPattern, '<span class="u-highlight">$1</span>')
```

config/nunjucks/filters.js
yozzo: amended

config/nunjucks/filters.js
yozzo: amended

config/nunjucks/filters.js
yozzo: removed

config/nunjucks/filters.js
yozzo: amended

test/unit/config/nunjucks/filters.test.js
```diff
@@ -93,6 +93,26 @@ describe('nunjucks filters', () => {
         expect(this.highlightedString.val).to.equal('Here is an <span class="u-highlight">exam</span>ple phrase')
       })
     })
+
```
feedmypixel: Feels like it would be worth adding in a test around the word boundary work?

```
it('should highlight only a word')
it('should highlight matched text')
```

yozzo: added

PR #1303, opened 2018-03-01
Update XHR to only update results section

ztolley:
This change modifies the behaviour of the XHR module used to update collections when a filter is applied. The change causes only the results section to be refreshed instead of the entire content area. This change results in faster screen updates and less network traffic on collections with long filter lists.

The previous behaviour was required because the region filter was optional, but this has now been changed.

This change introduces a new module to synchronise the value between two fields, this allows the sort value to be kept in sync between the sort form and the filter form.

A second module was introduced to clear field values in a form, this is required as the filters now need to manually reset as the filter section does not get redrawn

### DIFF
--- a/assets/javascripts/modules/mirror-value.js
+++ b/assets/javascripts/modules/mirror-value.js
@@ -1,0 +1,21 @@
+/*
+ * Mirror Value
+ *
+ * Listens for a change on one form element and mirrors the value to another
+ */
+const MirrorValue = {
+  init (wrapper = document) {
+    const sourceField = wrapper.querySelector('.js-MirrorValue')
+
+    if (sourceField) {
+      const targetSelector = sourceField.getAttribute('data-target-selector')
+      const targetField = wrapper.querySelector(targetSelector)
+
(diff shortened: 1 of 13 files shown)


assets/javascripts/modules/mirror-value.js
```diff
@@ -0,0 +1,21 @@
+/*
+ * Mirror Value
+ *
+ * Listens for a change on one form element and mirrors the value to another
+ */
+const MirrorValue = {
+  init (wrapper = document) {
+    const sourceField = wrapper.querySelector('.js-MirrorValue')
```
yozzo: What is the house rule for selector classes?  I remember seeing  something mentioned here:
[https://github.com/alphagov/styleguides/blob/master/js.md#html-class-hooks](url) but doesn't specify if camelCase or lisp-cased (hyphenated) should be used. Examples use hyphen though. wouldn't ```.js-mirror-value``` be more HTML'y?

ztolley: We seem to use kebab case for style classes and this style for classes tied to a JS module.

PR #1305, opened 2018-03-01
MI 1391 Add policy feedback as a new interaction

web-bert:
Add new policy-feedback-form macro
Add Policy feedback to the existing radio options
Update transformer to handle new 'kind'
Add unit tests
Add acceptance tests
Put kind names in constants.js


web-bert:
policy-feedback

PR #1311, opened 2018-03-02
DH-1549 Autocomplete - Multi-select component

ztolley:
Adds a vue js based multi-select, autocomplete, component. The main highlights are:

- Vue js dependencies and the webpack build process
- Introduces  smart multiple-choice component
- Adds an example of the component in the gallery
- Adds separate unit tests for vue components
- Integrates new tests into circle

The introduces the control ready for follow up requests to add it to the interactions collection screen.

To-Do/Things learned: It was not possible to test the vue components with the current mocha testing setup as code needs to be transpiled before testing. For this reason, the current tests have been split into their own folder and use mocha-webpack to transpile code before testing. Jest JS appears to be better suited to client-side code testing as is allows testing markup against snapshots and has built in support for transpiling code, based on file extension. This would mean we can initially move client-side tests to Jest and then judge if it's worth moving server side too, to keep one testing codebase.

Try it out here
https://datahub-beta2-pr-1311.herokuapp.com/components/smart-multiple-choice?adviser=5707c18a-454a-4bd2-b14b-218761a684bc


assets/javascripts/vue/smart-multiple-choice.vue
```diff
@@ -0,0 +1,116 @@
+<template>
+  <div class="c-form-group c-form-group--light c-form-group--smaller c-form-fieldset--light c-form-fieldset--smaller c-form-group--filter">
+    <label class="c-form-group__label" for="ajax">
```
yozzo: I think the id attribute should be dynamic, as by the look of the design we might have multiple filters per page

assets/javascripts/vue/smart-multiple-choice.vue
```diff
@@ -0,0 +1,116 @@
...
+      :options-limit="500"
+      :show-no-results="false"
+      :showLabels="false"
+      :searchable="true"
+      @search-change="asyncFind">
+      <template slot="clear" slot-scope="props">
+        <div class="multiselect__clear" v-if="selectedOptions.length" @mousedown.prevent.stop="clearAll(props.search)"></div>
+      </template>
+
+      <span slot="noResult">Oops! No elements found. Consider changing the search query.</span>
```
yozzo: error verbiage is usually prone to change, should we have all error messages in a global file which we import where needed?

src/apps/components/controllers.js
```diff
@@ -1,9 +1,13 @@
+const { castArray } = require('lodash')
+
```
yozzo: please remove empty row

src/apps/components/views/smart-multiple-choice.njk
```diff
@@ -0,0 +1,26 @@
+{% extends "./_layout.njk" %}
+
+{% block body_main_content %}
+  <form class="js-vue-wrapper">
+    <smart-multiple-choice
+      entity="adviser"
```
yozzo: entity type shouldn't be dynamic?

test/vue/smart-multiple-choice.test.js
```diff
@@ -0,0 +1,263 @@
...
+    })
+
+    describe('#asyncFind', () => {
+      let asyncFind
+
+      beforeEach(() => {
+        asyncFind = SmartMultipleChoice.methods.asyncFind.bind(instance)
+      })
+
+      context('when the user enters more then 3 characters', () => {
```
yozzo: *cough

assets/javascripts/vue/smart-multiple-choice.vue
```diff
@@ -0,0 +1,116 @@
...
+  </div>
+</template>
+
+<script>
+  const axios = require('axios')
+  const Multiselect = require('vue-multiselect').default
+  const getFormData = require('get-form-data').default
+  const pickBy = require('lodash/pickBy')
+  const debounce = require('lodash/debounce')
+
```
yozzo: please remove empty row

assets/javascripts/vue/smart-multiple-choice.vue
```diff
@@ -0,0 +1,116 @@
...
+        if (this.isSubmitting) { return }
+        this.isSubmitting = true
+
+        const form = this.formSelector ? document.querySelector(this.formSelector) : this.$el.closest('form')
+        if (!form) { return }
+
+        const query = pickBy(getFormData(form))
+        query[this.name] = newOptions.map(option => option.value)
+
+        XHR.request(form.action, query)
```
yozzo: in this case it's not an issue, but if the filter is to become more complex, it would be a good idea to have the requests in a different `container` or `smart` component and only keep in the filter the `presentation` or `dumb` component logic, thus connecting  them through a publisher-subscriber pattern.  see vuex diagram here: https://vuex.vuejs.org/en/images/vuex.png

assets/javascripts/vue/smart-multiple-choice.vue
teneightfive: We've been using a pattern of a new line to signify the split between external and internal imports. It keeps it cleaner and makes it easier to scan.

So we load any libraries from external dependencies first and internal modules second using a new line inbetween.

teneightfive:
May want to look at the styling to bring it more inline with the app.

We don't really use rounded corners very much so I'd drop them from the border of each item:

image

The cross to remove an item is also hidden and the hover state doesn't feel quite right:

image

We use the yellow colour to signify focus whereas here it's used to signify hover state.

Not sure what the arrow on the right is doing either. Feels odd that it just pops up and down yet there is no list of options displayed when you click in it:

image

Here's a possible cleaner display of the selected items:

image

ztolley:
Thanks for the suggestions @teneightfive , I was thinking of going with those styles but with the highlight background, as it's subtle

assets/javascripts/vue/smart-multiple-choice.vue
ztolley: In this case, the `vue-multiselect` is the dumb component and the business logic is in this component. This isn't complex enough to need a store or external state/action management.

src/apps/components/controllers.js
ztolley: reason stated by Dom above, we seperate imports from vendors and local.

src/apps/components/views/smart-multiple-choice.njk
ztolley: Yes, but this is just an example of it's usage

test/vue/smart-multiple-choice.test.js
ztolley: ?

yozzo: `than`

ztolley: ah

ztolley:
@teneightfive The arrow on the right bothers me too, I've just found a way to override it's default behaviour so I'll only show it when there are options, so you don't get that odd behaviour when then are none

ztolley:
I've updated the styles now to use the highlight background, rid of the yellow delete selector background and got rid of the rounded edges.

The arrow toggle now only shows then there are options, so it doesn't have the strange behaviour when the list is closed.

Other minor changes made too following feedback.

highlight

assets/javascripts/app-vue.js
```diff
@@ -0,0 +1,15 @@
+/* eslint no-new: 0 */
+const Vue = require('vue')
+
+const SmartMultipleChoice = require('./vue/smart-multiple-choice.vue').default
```
teneightfive: Not convinced `SmartMultipleChoice` is a good name for this.

Maybe go with `typeahead` or `autocomplete` which feel more like common names for this type of component.

assets/stylesheets/application.scss
```diff
@@ -2,6 +2,8 @@ $images-dir: "/images/";
 
 @import "_deprecated/trade-elements";
 
+@import "vue-multiselect.min.css";
```
teneightfive: Wondering whether it's worth importing this at all.

Wonder if we can just take the base styles we use for this and set them in `form/vue-multiselect` rather than importing them all and overriding lots of them.

ztolley: I need to import the styles from the component otherwise if the component author changes their markup I risk the control breaking. The majority of the css brought in is required, just a small portion you pointed out regarding rtl.

assets/javascripts/app-vue.js
ztolley: Adding 2 properties to the control for `autoSubmit` and 'allowMultiple' to make it more generic and rename to `typeahead`.

Moving example into the forms gallery page.

ztolley:
@teneightfive Following comments I've corrected the border colour and text colour.

With regard to showing a message when there are no options found and the fact it has stray markup in the DOM marked as `dispay:none` I've suggested to the developer use `v-if` instead.

Also, I did try showing the message about no suggestions but that does not take into consideration the minimum term length which ends up being more confusing than not showing a message.

I will try to submit a PR to that project that lets you specify a minimum term length and take that into consideration when showing the message that no suggestions have been found, the alternative will be to fork the project or just copy their component into our project, but I would like to avoid that if possible.

assets/stylesheets/components/form/_typeahead.scss
```diff
@@ -0,0 +1,72 @@
...
+}
+
+.multiselect__content {
+  width: 100%;
+}
+
+.multiselect__content-wrapper {
+  border: 1px solid $grey-1;
+  border-top: 0;
+  border-bottom-left-radius: 0;
```
teneightfive: This could just be: `border-radius: 0;`

assets/stylesheets/components/form/_typeahead.scss
```diff
@@ -0,0 +1,72 @@
...
+  margin-bottom: 0;
+}
+
+.multiselect__input:focus,
+.multiselect__single:focus {
+  box-shadow: none;
+}
+
+.multiselect--active {
+  box-shadow: 0 0 0 3px $yellow;
```
teneightfive: Maybe `$focus-colour` instead?

src/apps/components/middleware.js
```diff
@@ -0,0 +1,29 @@
+const { castArray } = require('lodash')
+
+const { getAdviser } = require('../adviser/repos')
+
+async function adviserLookup (req, res, next) {
+  try {
+    const token = req.session.token
+    res.locals.advisers = []
+    const adviserParam = req.query.adviser
```
teneightfive: Move up a line to group the `const` together.

src/apps/components/views/form.njk
```diff
@@ -317,4 +317,28 @@
...
+        <h2>Advisers selected</h2>
+        <ul>
+          {% for adviser in advisers %}
+            <li>{{ adviser.label }}</li>
+          {% endfor %}
+        </ul>
+      </div>
+    {% endblock %}
+
+      {% endcall %}
```
teneightfive: Check the indentation

assets/stylesheets/components/form/_typeahead.scss
ztolley: No, because the thing it overrides specifies the corners, just using border-radius won't override it

PR #1325, opened 2018-03-07
Create journey acceptance tests

teneightfive:
This change adds acceptance tests around the OMIS order creation journey.

It starts with adding the journey from a particular company which skips the company search.

It also adds some more shared step definitions which can be used across the rest of the tests
once this change has been merged.

See the commit logs for more detailed descriptions of each change.

### DIFF
--- a/test/acceptance/features/omis/create.feature
+++ b/test/acceptance/features/omis/create.feature
@@ -1,0 +1,45 @@
+@omis @omis--create
+Feature: Create new order using company ID
+
+  Background: Create initial draft
+    When I navigate to the `omis.create.start` page using `company` `Venus Ltd` fixture
+    And I select a value for `contactField` on the `omis.create.contact` page
+    And I submit the form
+    And I select a value for `marketField` on the `omis.create.market` page
+    And I submit the form
+    And I choose the company’s primary sector
+    And I submit the form
+    Then I should see the correct text on the `omis.create.summary` page
+      | elementPath     | expectedText                     |
+      | contact.contact | omis.create.contact.contactField |
+      | market.market   | omis.create.market.marketField   |
+      | sector.sector   | omis.create.sector.sectorField   |
+
+  Scenario: Edit steps from summary
+    When I click `editContactLink` on the `omis.create.summary` page
+    Then I am on the `omis.create.contact` page
+    When I select a value for `contactField` on the `omis.create.contact` page
+    And I submit the form
+    Then I am on the `omis.create.summary` page
+    And I should see the correct text on the `omis.create.summary` page
+      | elementPath     | expectedText                     |
+      | contact.contact | omis.create.contact.contactField |
+
+    When I click `editMarketLink` on the `omis.create.summary` page
+    Then I am on the `omis.create.market` page
+    When I select a value for `marketField` on the `omis.create.market` page
+    And I submit the form
+    Then I am on the `omis.create.summary` page
+    And I should see the correct text on the `omis.create.summary` page
+      | elementPath     | expectedText                     |
+      | market.market   | omis.create.market.marketField   |
+
+    When I click `editSectorLink` on the `omis.create.summary` page
+    Then I am on the `omis.create.sector` page
+    When I click `useCustomSectorOption` on the `omis.create.sector` page
+    And I select a value for `sectorField` on the `omis.create.sector` page
+    And I submit the form
+    Then I am on the `omis.create.summary` page
+    And I should see the correct text on the `omis.create.summary` page
+      | elementPath     | expectedText                     |
+      | sector.sector   | omis.create.sector.sectorField   |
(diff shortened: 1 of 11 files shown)


test/acceptance/pages/omis/create/start.js
```diff
@@ -0,0 +1,13 @@
+const { find } = require('lodash')
+
+const { company: companyFixtures } = require('../../../fixtures')
+
+module.exports = {
+  url: function companyFixtureUrl (companyName) {
+    const fixture = find(companyFixtures, { name: companyName })
```
ztolley: I still prefer native find, but I can see in this instance it would end up messy

test/acceptance/features/step_definitions/common.js
```diff
@@ -57,3 +57,17 @@ When(/^I select a value for `(.+)` on the `(.+)` page$/, async function (element
     throw new Error(`The page object '${pageName}' does not exist`)
   }
 })
+
+Then(/^I am on the `(.+)` page$/, async function (pageName) {
+  try {
+    const page = get(client.page, pageName)()
```
ztolley: would appreciate you explaining what this does, not because its wrong but the opposite, I like it and just want to understand how it works

test/acceptance/features/step_definitions/common.js
```diff
@@ -71,3 +71,23 @@ Then(/^I am on the `(.+)` page$/, async function (pageName) {
     throw new Error(error)
   }
 })
+
+Then(/^I should see the correct text on the `(.+)` page$/, async function (pageName, data) {
```
ztolley: Something doesn't feel right. The step reads like it is just checking for text on a page but it looks like it's doing more than that.  Can the step text reflect that it's looking for text in a section or something more obvious it's not just checking for text on any old page?

test/acceptance/features/omis/create.feature
```diff
@@ -0,0 +1,45 @@
...
+    When I navigate to the `omis.create.start` page using `company` `Venus Ltd` fixture
+    And I select a value for `@contactField` on the `omis.create.contact` page
+    And I submit the form
+    And I select a value for `@marketField` on the `omis.create.market` page
+    And I submit the form
+    And I choose the company’s primary sector
+    And I submit the form
+    Then I should see the correct text on the `omis.create.summary` page
+      | elementPath      | expectedText                      |
+      | contact.@contact | omis.create.contact.@contactField |
```
ztolley: Still torn on the @ syntax vs using the label names. contact.@contact reads... weird, same for the expected text.

ztolley: can't we strip off the @ here and then add it if we find a page object that matches?

test/acceptance/features/omis/step_definitions/create.js
```diff
@@ -0,0 +1,13 @@
+const { set } = require('lodash')
```
ztolley: Why didn't this get added as a separate commit like the other step definitions?

test/acceptance/pages/omis/create/start.js
teneightfive: Yeah I'm thinking of removing the idea of finding them on name anyway and we reference them using the property path like we do page objects. But make the properties readable and clear. Think it'll actually be better to read and understand than `Lambda plc` too.

test/acceptance/features/step_definitions/common.js
teneightfive: Sure, let's catch up next week.

It's basically using lodash's get to deep find a page on the `client.page` object and call that function.

Then the other part of this method either checks the current url against the url returned from the inbuilt url method that nightwatch gives you or if we want to use a regex to match, it will use that. Works really nicely actually.

test/acceptance/features/step_definitions/common.js
teneightfive: Yeah wasn't 100% convinced on the name either. It started as just checking sections then I ended up refactoring it so that it was more powerful.

I see it having the ability to replace a lot of the step definitions we have.

It essentially allows you to check any element on any page against either a value from state or a string. Keeps the defining of elements to the page object too which felt nice.

test/acceptance/features/omis/create.feature
teneightfive: Hmm, possibly. We may be able to build the `@` in the step definition instead.

So these could read: `contact.contact` (might be helpful to rename that too) and `omis.create.contact.contactField`.

test/acceptance/features/omis/step_definitions/create.js
teneightfive: I was considering it but this felt like because it was tied to only this feature file it made sense to bundle these.

The other step definitions are more global and felt like they would benefit from a separate commit and description.

This step felt more straightforward so didn't warrant that.

PR #1328, opened 2018-03-07
Improve consistency of common step definitions

teneightfive:
This change includes two updates to the existing step definitions
around navigating to pages.

It instantiates the page object for the base navigation step definition
so it is consistent with the other uses of this method of finding a
page object.

It also ensures the fixture navigation step definition is wrapped in
a try catch block to render a more friendly error with the page object
isn't found.


test/acceptance/features/step_definitions/common.js
```diff
@@ -6,23 +6,27 @@ const fixtures = require('../../fixtures')
 
 When(/^I (?:navigate|go|open|visit).*? `(.+)` page$/, async function (pageName) {
   try {
-    const page = get(client.page, pageName)
-    await page().navigate()
+    const Page = get(client.page, pageName)()
```
ztolley: I'm still of the school in which you only use Capital to denote that this is something that needs to be instantiated with `new`

teneightfive:
>Only approving this is you help Alex to rebase Bens branch

🙄 you should be approving because it's helping your codebase and to make tests more readable and consistent...

PR #1332, opened 2018-03-09
added functionality to prototype the company contacts data to the inv…

yozzo:
…estment page

### DIFF
--- a/assets/javascripts/modules/add-items.js
+++ b/assets/javascripts/modules/add-items.js
@@ -65,8 +65,11 @@ const AddItems = {
   cacheEls () {
     const lastItem = this.wrapper.querySelector(this.settings.itemSelector)
 
-    this.itemContainer = lastItem.parentNode
-    this.template = lastItem.cloneNode(true)
+    if (lastItem) {
+      this.itemContainer = lastItem.parentNode
+      this.template = lastItem.cloneNode(true)
+    }
+
     this.addButton = this.wrapper.querySelector(this.settings.addButtonSelector)
   },
 
--- a/src/apps/investment-projects/middleware/forms/details.js
+++ b/src/apps/investment-projects/middleware/forms/details.js
@@ -26,8 +26,13 @@ async function populateForm (req, res, next) {
 
   try {
     const investmentData = transformFromApi(res.locals.investmentData)
+    const equityCompanyData = transformFromApi(res.locals.equityCompany)
     const createdOn = get(investmentData, 'created_on')
     const token = req.session.token
+
+    if (investmentData && (investmentData.client_contacts !== undefined && investmentData.client_contacts.length === 0)) {
+      investmentData.client_contacts = ['']
+    }
 
     const {
       equityCompany,
@@ -46,6 +51,13 @@ async function populateForm (req, res, next) {
     const state = assign({}, {
       client_contacts: [''],
       business_activities: [''],
+      company_contacts: () => {
+        if (equityCompanyData && equityCompanyData.contacts) {
+          return equityCompanyData.contacts
+        } else {
+          return []
+        }
+      },
     }, investmentData)
 
     const advisersResponse = await getAdvisers(token)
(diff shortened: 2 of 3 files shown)


src/apps/investment-projects/views/_details-form.njk
```diff
@@ -126,27 +126,42 @@
...
-          value: client_contact
-        }) }}
-      {% endfor %}
+        <p class="c-form-fieldset__add-another">
+          <button class="button button--secondary js-AddItems__add" type="submit" name="add-item" value="client_contacts" data-persist-values="true">Add another contact</button>
+        </p>
+      </div>
+
+    {% else %}
+      <h2>Pisica</h2>
```
teneightfive: What's the purpose of these headings?

src/apps/investment-projects/views/_details-form.njk
```diff
@@ -126,27 +126,42 @@
...
+      <h3>{{ form.state.company_contacts.length > 0  }}</h3>
+      <h3>{{ not form.errors.messages.client_contacts  }}</h3>
+      <h3>{{ form.state.company_contacts.length > 0 and not form.errors.messages.client_contacts  }}</h3>
+      <h3>{{ form.hiddenFields.investor_company }}</h3>
+
+      <p>This company has no contacts, please create a business contact</p>
```
teneightfive: Maybe use the [Message component](https://datahub-beta2.herokuapp.com/components/messages) for this as it's a pattern we've [elsewhere](https://github.com/uktrade/data-hub-frontend/blob/5d97cd03fafc842492c6adc238da9364b777066d/src/apps/omis/apps/create/views/contact.njk#L7-L9) to display messages like this.

src/apps/investment-projects/views/_details-form.njk
```diff
@@ -126,27 +126,42 @@
...
+      <a class="button button--secondary" href="/contacts/create?company={{ form.hiddenFields.investor_company }}">Add Contact</a>
+      <a class="button button--secondary" href="{{ form.options.returnLink }}">Return Link</a>
```
teneightfive: Do we need the return link?

src/apps/investment-projects/views/_details-form.njk
```diff
@@ -126,27 +126,42 @@
...
+    {% endif %}
+
+    {{ form | log | safe }}
```
teneightfive: Maybe remove this dump of the form object.

assets/javascripts/modules/add-items.js
```diff
@@ -65,8 +65,11 @@ const AddItems = {
   cacheEls () {
     const lastItem = this.wrapper.querySelector(this.settings.itemSelector)
 
-    this.itemContainer = lastItem.parentNode
-    this.template = lastItem.cloneNode(true)
+    if (lastItem) {
```
teneightfive: What's the purpose of this `if` block? In theory we should always be rendering at least one item.

And if we don't maybe we need to handle the template not being there in the `render` method too.

src/apps/investment-projects/middleware/forms/details.js
```diff
@@ -46,6 +51,13 @@ async function populateForm (req, res, next) {
     const state = assign({}, {
       client_contacts: [''],
       business_activities: [''],
+      company_contacts: () => {
+        if (equityCompanyData && equityCompanyData.contacts) {
+          return equityCompanyData.contacts
```
teneightfive: Lodash's [flatten](https://lodash.com/docs/4.17.4#flatten) method might allow this to not require the if/else block.

```js
...
company_contacts: flatten([equityCompanyData.contacts]),
...
```

Also is this being used anywhere in the form? Adding it to the `state` object doesn't feel quite right as it's not holding form state.

If it's to provide the logic in the view `equityCompany.contacts` will already be available in the view as it's set on the `res.locals` object in previous middleware.

src/apps/investment-projects/views/_details-form.njk
yozzo: work in progress, sir. had to pick up something else, this isn't yet ready for reviewing, therefore didn't attach any reviewers to it. but thank you.

teneightfive: Ah my bad. Sorry, was just going through all the PRs at the time.

PR #1333, opened 2018-03-09
Bugfix/dh 1563 search term highlighting

yozzo:
https://uktrade.atlassian.net/browse/DH-1563

Because the regex isn't escaping all characters, we added an `escape` filter on top of of the search results string to prevent the app from crashing.

### DIFF
--- a/config/nunjucks/filters.js
+++ b/config/nunjucks/filters.js
@@ -8,6 +8,7 @@ const queryString = require('query-string')
 const {
   assign,
   concat,
+  escape,
   isArray,
   isFunction,
   isPlainObject,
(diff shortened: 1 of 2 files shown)


config/nunjucks/filters.js
```diff
@@ -104,7 +105,7 @@ const filters = {
         ? new RegExp(`\\b(${cleanedSearchTerm})\\b`, 'gi')
         : new RegExp(`(${cleanedSearchTerm})`, 'gi')
 
-      const result = searchResultText.replace(searchPattern, '<span class="u-highlight">$1</span>')
+      const result = escape(searchResultText).replace(searchPattern, '<span class="u-highlight">$1</span>')
```
ztolley: any way to test it?

yozzo: added

PR #1334, opened 2018-03-12
Fix component modifiers

teneightfive:
Currently some modifiers can be applied with undefined values. It can lead to examples like this:

```scss
c-component-name c-component-name--undefined
```

This set of changes uses a filter to apply those modifiers rather than always manually building them out.

See the commit descriptions for a better breakdown of each change.

### DIFF
--- a/src/templates/_macros/form/entity-search-form.njk
+++ b/src/templates/_macros/form/entity-search-form.njk
@@ -23,10 +23,9 @@
   {% set inputPlaceholder = props.inputPlaceholder | default(props.inputLabel) -%}
   {% set isLabelHidden = props.isLabelHidden | default(true) %}
   {% set entityType = props.entityType | default('company') %}
-  {% set modifier = props.modifier | concat('') | reverse | join(' c-entity-search--') if props.modifier %}
 
   {% call Form(props | assign({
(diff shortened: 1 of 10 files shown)


src/templates/_macros/form/entity-search-form.njk
```diff
@@ -23,10 +23,9 @@
   {% set inputPlaceholder = props.inputPlaceholder | default(props.inputLabel) -%}
   {% set isLabelHidden = props.isLabelHidden | default(true) %}
   {% set entityType = props.entityType | default('company') %}
-  {% set modifier = props.modifier | concat('') | reverse | join(' c-entity-search--') if props.modifier %}
 
   {% call Form(props | assign({
-    class: 'c-entity-search ' + modifier,
+    class: 'c-entity-search' | applyClassModifiers(props.modifier),
```
ztolley: Does this support the ability to apply multiple modifiers?

teneightfive: Yep, supports the same behaviour as before where the modifier property can be a `string` or `array`.

Changing them to use the filter actually handles that better. It always creates an array and then filters out empty values so you don't have the example in the PR description or what we spotted of `c-component-name--`.

PR #1335, opened 2018-03-14
Removed duplicate params in XHR

ztolley:
With the addition of the adviser lookup control, it will be possible to submit an identical key-value pair if the user picks their own name from the adviser typeahead control.

This simple change ensures that the key-value pairs are unique when generating the URL to search and when updating the user URL.

Note: that we need to find a better way to test this component to allow unit tests for things like this to be added but I want to do that separate so we can also consider how it would work with Jest.

### DIFF
--- a/assets/javascripts/lib/xhr.js
+++ b/assets/javascripts/lib/xhr.js
@@ -1,6 +1,8 @@
 const axios = require('axios')
 const createHistory = require('history').createBrowserHistory
 const queryString = require('query-string')
+const uniqWith = require('lodash/uniqWith')
+const isEqual = require('lodash/isEqual')
 
 const history = createHistory()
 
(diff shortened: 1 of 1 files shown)


assets/javascripts/lib/xhr.js
```diff
@@ -46,20 +48,22 @@ const XHR = {
...
+    const uniqueParams = uniqWith(params, isEqual)
+
     if (showLoader) {
       outlet.classList.add('u-loading')
     }
 
     if (params) {
-      const url = `?${queryString.stringify(params)}`
-      history.push(url)
+      const historyUrl = `?${queryString.stringify(uniqueParams)}`
```
teneightfive: Might need to update line 35 too.

ztolley: `updateOutlet` uses the parms sent to it in line 66, which is the unique ones. Nothing else uses it.

ztolley:
Just realised a bug so I'm pulling this to re-do it

PR #1336, opened 2018-03-14
DH-1439 User can unintentionally create duplicate records

yozzo:
Some people still use double click to submit forms or even access links, in order to prevent that behaviour, which also creates unwanted multiple entries in our database, we can simply disable the submit button after the user interacts with it.

this also fixes the sorting issue on the interactions page

### DIFF
--- a/src/apps/contacts/views/interactions.njk
+++ b/src/apps/contacts/views/interactions.njk
@@ -9,12 +9,18 @@
   <p>
     An interaction could be a meeting, call, email or another activity where you have been in touch with a contact.
   </p>
-  {{
-    Collection(results | assignCopy({
-      countLabel: 'interaction',
-      summaryActionsHTML: addButton,
-      sortForm: assign({}, sortForm, {action: CURRENT_PATH}),
-      query: QUERY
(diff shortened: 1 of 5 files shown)


src/apps/contacts/views/interactions.njk
```diff
@@ -9,12 +9,18 @@
...
-      countLabel: 'interaction',
-      summaryActionsHTML: addButton,
-      sortForm: assign({}, sortForm, {action: CURRENT_PATH}),
-      query: QUERY
-    }))
-  }}
+
+  {% block xhr_content %}
+    <article id="xhr-outlet">
+        {{
```
ztolley: Can you reformat with 2 spaces instead of 4, and pull those braces back so `Collection` is indented inside them. ? .editorconfig is your friend

yozzo: amended

ztolley: Looks the same

yozzo: forgot to push

PR #1341, opened 2018-03-16
Content updates

teneightfive:
This change includes a number of small commits that address content changes which
came out of a full journey content review.

See each commit for a more detailed description of what has changed and why.


ztolley:
Looks fine except the unit test

PR #1348, opened 2018-03-20
DH-1627 Clear filters functionality doesn't apply for typeahead filter

yozzo:
Because typeahed filters have a different behaviour than normal filters we had to find a way of clearing them before the submit to the server.
We were able to clear the filter using normal vue behaviour but this was creating a race condition with the global filters submit,
therefore decided for this approach, by removing the filter field from the clear-filters submit.

### DIFF
--- a/assets/javascripts/modules/clear-inputs.js
+++ b/assets/javascripts/modules/clear-inputs.js
@@ -41,6 +41,10 @@ const ClearInputs = {
     if (targetForm) {
       const formControls = getFormControls(targetForm)
       clearValues(formControls)
+
+      Array.from(targetForm.querySelectorAll('.js-ClearInputs--removable-field'))
+        .forEach(input => input.parentNode.removeChild(input))
+
       targetForm.submit()
     }
   },
--- a/assets/javascripts/modules/print-dialog.js
+++ b/assets/javascripts/modules/print-dialog.js
@@ -8,7 +8,7 @@ const PrintDialog = {
   contentLoaded () {
     const printLinks = document.querySelectorAll(this.selector)
 
-    printLinks.forEach((link) => {
+    Array.from(printLinks).forEach((link) => {
       link.addEventListener('click', this.handleClick.bind(this))
     })
   },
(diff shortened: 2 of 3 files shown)


assets/javascripts/modules/clear-inputs.js
```diff
@@ -41,6 +41,7 @@ const ClearInputs = {
     if (targetForm) {
       const formControls = getFormControls(targetForm)
       clearValues(formControls)
+      targetForm.querySelectorAll('.js-ClearInputs--removable-field').forEach(input => input.remove())
```
ztolley: According to https://developer.mozilla.org/en-US/docs/Web/API/ChildNode/remove, the remove function doesn't work on iOS and Explorer.

You need to get the node's parent and then call `removeChild`

ztolley: Oh and check your spellings on the comments :)

assets/javascripts/modules/clear-inputs.js
```diff
@@ -41,6 +41,10 @@ const ClearInputs = {
     if (targetForm) {
       const formControls = getFormControls(targetForm)
       clearValues(formControls)
+      targetForm
+        .querySelectorAll('.js-ClearInputs--removable-field')
+        .forEach(input => input.parentNode.removeChild(input))
```
ztolley: Can't use `foreach` over a collection of elements from query selector.

You have to create an array from the result of the query selector all. I guess Chrome lets you get away with it but other browsers don't.

capture

try

```
const inputs= Array.from(targetForm.querySelectorAll('.js-ClearInputs--removable-field'))
inputs.forEach(input => input.parentNode.removeChild(input))
```

And either install IE locally using one of the free VM's from Microsoft or we have a browser stack account

ztolley: Still getting ForEach error when using ie and loading the page

assets/javascripts/modules/print-dialog.js
```diff
@@ -8,7 +8,7 @@ const PrintDialog = {
   contentLoaded () {
     const printLinks = document.querySelectorAll(this.selector)
 
-    printLinks.forEach((link) => {
+    Array.from(printLinks).forEach((link) => {
```
ztolley: I know it's one line but it's not related to typeahead. if it could be a separate PR I will review and approve it straight away, after checking in IE

PR #1350, opened 2018-03-21
DH-1439 changed the global submit type button selector to a more spec…

yozzo:
changed the global submit type button selector to a more specific classname selector.

@teneightfive  pointed out that there are other functionalities in the app that break when used in conjunction with the module that prevents multiple submits. So we decided to change the global selector to `.js-prevent-multiple-submits` instead, and add it to where it's required in the future.

### DIFF
--- a/assets/javascripts/modules/prevent-multiple-submits.js
+++ b/assets/javascripts/modules/prevent-multiple-submits.js
@@ -1,6 +1,6 @@
 const CONSTANTS = {
   selectors: {
-    button: 'button[type="submit"]',
+    button: 'js-prevent-multiple-submits',
     form: 'form',
   },
   types: {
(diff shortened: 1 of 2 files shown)


assets/javascripts/modules/prevent-multiple-submits.js
```diff
@@ -1,6 +1,6 @@
 const CONSTANTS = {
   selectors: {
-    button: 'button[type="submit"]',
+    button: 'js-prevent-multiple-submits',
```
teneightfive: `.js-PreventMultipleSubmits`? Follows the pattern we've used elsewhere of pascal casing the js prefixed selector.

assets/javascripts/modules/prevent-multiple-submits.js
```diff
@@ -19,16 +19,16 @@ const PreventMultipleSubmits = {
   counter: 0,
 
   init () {
-    if (document.querySelectorAll(CONSTANTS.selectors.button).length) {
+    if (document.querySelectorAll(`.${CONSTANTS.selectors.button}`).length) {
```
teneightfive: Maybe actually just include the dot in the selector as it's part of it. Otherwise if you change to an attribute selector you'd also have to change this line which voids the idea of having a constant that is used by this function in its form.

yozzo: Initially it was like that, but I'm also checking just for the class name in the event callback

PR #1352, opened 2018-03-21
[DH-1613] Added Triage field for SPI-1

foxleigh81:
Added field to satisfy the 'SPI-1' requirement of the following presentation:
https://docs.google.com/presentation/d/1XjsWYCD42oLun_WJNUkl583LO9ekhIrJkEd2jGPFKgM/edit#slide=id.p10

Screenshot of SPI1

- Added 'Date project arrived in triage' field
- Updated acceptance tests to accommodate new field.


src/apps/investment-projects/middleware/collection.js
```diff
@@ -32,6 +32,8 @@ function getRequestBody (req, res, next) {
     'sector',
     'investment_type',
     'investor_company',
+    'project_arrived_in_triage_on_before',
```
teneightfive: You might need to add the macros to show the filters on the listing page, or is that coming in a separate change?

src/apps/investment-projects/transformers/project.js
```diff
@@ -64,7 +66,6 @@ function transformToApi (body) {
   } else {
     formatted.estimated_land_date = null
   }
-
```
teneightfive: Maybe leave this spacing in.

src/apps/investment-projects/views/_details-form.njk
```diff
@@ -271,6 +271,14 @@
     ]
   }) }}
 
+  {{ DateFieldset({
+    name: 'project_arrived_in_triage_on',
+    label: 'Date project arrived in triage (IST use)',
+    hint: '',
```
teneightfive: Can just remove this property rather than setting to an empty string.

src/apps/investment-projects/middleware/collection.js
ztolley: I don't see this field added as a new filter so remove from here. This section is only used to process filters.

PR #1353, opened 2018-03-21
DH-1588 Sort investment projects results by their created date

yozzo:
Added the option to sort investment projects results by their `created date`

### DIFF
--- a/src/apps/investment-projects/macros.js
+++ b/src/apps/investment-projects/macros.js
@@ -100,6 +100,7 @@ const investmentSortForm = {
         'target-selector': '.c-collection-filters input[name="sortby"]',
       },
       options: [
+        { value: 'created_on:desc', label: 'Created date' },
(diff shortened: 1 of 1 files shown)


ztolley:
What is a fature?

src/apps/investment-projects/macros.js
```diff
@@ -100,6 +100,7 @@ const investmentSortForm = {
         'target-selector': '.c-collection-filters input[name="sortby"]',
       },
       options: [
+        { value: 'created_on:desc', label: 'Created date' },
```
teneightfive: What about `Most recently created`? Feels like it reads better and follows the pattern we'd used elsewhere.

yozzo: just following the acceptance criteria:  https://uktrade.atlassian.net/browse/DH-1588

PR #1356, opened 2018-03-22
Rename collection macros

ztolley:
The naming of collection related nunjucks macros meant
that we had 4 different files named collection.njk and it was
not clear what some of the collection files did from their
name.

This refactor renames the different macros and pages related
to collections to make it clear what their purpose is, and also
splits out the collection list portion to it's own macro.

The naming now follows the structure described in the wiki
https://github.com/uktrade/data-hub-frontend/wiki/Collections


src/templates/_macros/collection/collection-header.njk
```diff
@@ -29,22 +28,23 @@
...
+            href="?{{ buildQuery({}, { custom: true, sortby: props.sortForm.children[0].value }) }}"
+            class="c-collection__filter-remove-all js-xhr js-ClearInputs"
+            data-target-selector=".c-collection-filters"
+            aria-label="Reset filter results"
+          >
+            Remove all filters
+          </a>
+        {% endif %}
+
+        {% if props.addButton %}
```
teneightfive: `addButton` feels like it ties it directly to its content which we should avoid.

Maybe make this prop a list of actions and have the macro loop over those an create a button for each. Might be a case where we want a couple of buttons and feels like it keeps it more flexible.

src/apps/companies/views/orders.njk
```diff
@@ -1,18 +1,14 @@
 {% extends "./_layout-view.njk" %}
 
-{% set addButton %}
```
teneightfive: Could we remove this view altogether if we add this prop to the global layout?

src/templates/_macros/collection/collection-content.njk
```diff
@@ -12,31 +12,19 @@
...
- # @param {string}  [props.summaryActionsHTML] - HTML content for actions block
+ # @param {string}   [props.addButton] - An object with a label and url to generate a link to add a new item
  #
  # @param {function} props.caller - Inner contents
 #}
-{% macro Collection(props) %}
-  {% set hasBlockLinks = 'block-links' in [props.listModifier] | flatten %}
-  {% set listModifier = props.listModifier | concat('') | reverse | join(' c-entity-list--') if props.listModifier %}
-
+{% macro CollectionContent(props) %}
```
teneightfive: Still not convinced on the rename of this. `Content` just feels a bit generic.

`Collection` actually feels more appropriate as that's also what we call the sass component...

src/templates/_macros/collection/collection-header.njk
```diff
@@ -29,22 +28,23 @@
           {% endif %}
         {% endif %}
       </div>
-      {% if props.selectedFilters | length or props.actionsHTML %}
```
teneightfive: Why remove this if block? It prevents the div being shown with no content.

src/templates/_macros/collection/collection-list.njk
```diff
@@ -0,0 +1,17 @@
+{% from '../entity.njk' import Entity %}
+
+{% macro CollectionList(props) %}
+  {% set hasBlockLinks = 'block-links' in [props.listModifier] | flatten %}
+  {% set listModifier = props.listModifier | concat('') | reverse | join(' c-entity-list--') if props.listModifier %}
+
+  <ol class="c-entity-list {{ listModifier }}">
```
teneightfive: Maybe some protection around this element so it doesn't render if there are not items. Otherwise you get an empty `ol`.

src/templates/_macros/collection/collection-header.njk
ztolley: Good suggestion

src/templates/_macros/collection/collection-content.njk
ztolley: Just trying out an alternative name based on this being an entity list and after the move of collection to layout

ztolley: But need to get away from having a layout named collection And a macro named collection, it's confusing

src/apps/companies/views/orders.njk
ztolley: it's a view, within a company, that only has the list of entities so the view has to be here

src/templates/_macros/entity/entity-list.njk
```diff
@@ -0,0 +1,17 @@
+{% from './entity.njk' import Entity %}
+
+{% macro EntityList(props) %}
+  {% set hasBlockLinks = 'block-links' in [props.listModifier] | flatten %}
+  {% set listModifier = props.listModifier | concat('') | reverse | join(' c-entity-list--') if props.listModifier %}
+
+  <ol class="c-entity-list {{ listModifier }}">
+    {% for item in props.items %}
+      <li class="c-entity-list__item">
+        {{ Entity(item | assign({
```
teneightfive: Would be better to use `assignCopy` here rather than `assign` as assign mutates the object.

src/templates/_macros/entity/entity-list.njk
```diff
@@ -0,0 +1,17 @@
+{% from './entity.njk' import Entity %}
+
+{% macro EntityList(props) %}
+  {% set hasBlockLinks = 'block-links' in [props.listModifier] | flatten %}
+  {% set listModifier = props.listModifier | concat('') | reverse | join(' c-entity-list--') if props.listModifier %}
+
+  <ol class="c-entity-list {{ listModifier }}">
```
teneightfive: Would also add some protection so you don't end up with an empty `ol` when there are no items.

PR #1361, opened 2018-03-23
DH-1632 added polyfill to support ES6 Object.assign() in IE

yozzo:
Added polyfill to support ES6 Object.assign() in IE

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,6 +15,7 @@
     "babel-eslint": "^8.1.2",
     "babel-loader": "^7.1.2",
     "babel-plugin-transform-class-properties": "^6.24.1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -36,6 +36,7 @@
     "del-cli": "^1.1.0",
     "dotenv": "^5.0.0",
     "element-closest": "^2.0.2",
+    "es6-object-assign": "^1.1.0",
```
ztolley: What about using lodash assign? Be careful to just import the assign part otherwise the app.js size will balloon

ztolley:
This isn't the root cause. The cause was due to the query-string library not working with ie any more.

PR #1366, opened 2018-03-27
Release version 4.16.0

ztolley:


jim68000:
Shouldn't this be called _release_/v4.16.0?

PR #1368, opened 2018-03-27
DH-1621 Added Primary Business Sector filter

yozzo:
Because we were displaying all business sectors in the filter, in order to improve user experience and page performance
we added a level filter for the business sector, and only display and filter according to the main categories.


src/lib/metadata.js
```diff
@@ -6,7 +6,19 @@ let redisClient
 
 function getMetadata (path, key) {
   const ttl = config.redis.metadataTtl
-  const url = `${config.apiRoot}/metadata/${path}/`
+  let url
```
ztolley: metadata needs to be marked as deprecated. Instead call 'getOptions' from the middleware and then pass that into the macro building the form definition

src/apps/investment-projects/macros.js
```diff
@@ -30,6 +30,7 @@ const investmentFiltersFields = function ({ currentAdviserId }) {
     },
     Object.assign({}, globalFields.sectors, {
       type: 'checkbox',
+      name: 'sector_descends',
       modifier: 'option-select',
```
ztolley: Add `options: sectorOptions` where sector options is the result of a call to getOptions(). Though in this case the assumption that getOptions make about the url won't work so you will need to add a parameter to that function to allow the caller to override the url to use.

PR #1375, opened 2018-04-03
DH-1640 companies hierarchies filters

foxleigh81:
**Defect**: the selection of more than one filter option returns result = all the companies record.

select Global HQ = global HQ record
select Global HQ + European HQ = All companies record

**Expected result**
Select two of the 3 options = xxx Companies + xxx companies record.

- [x] Filter bug fixed by refactoring `getRequestBody` (see below)
- [x] Refactored macro to follow the same pattern as the other filter macros
- [x] Refactored macro to use database values instead of hardcoded values

```javascript
function getRequestBody (req, res, next) {
  const selectedFiltersQuery = removeArray(pick(req.query, [
    'name',
    'sector',
    'country',
    'uk_region',
    'headquarter_type',
  ]), 'headquarter_type')
```

to

```javascript
function getRequestBody (req, res, next) {
  const selectedFiltersQuery = pick(req.query, [
    'name',
    'sector',
    'country',
    'uk_region',
    'headquarter_type',
  ])
```

### DIFF
--- a/src/apps/macros.js
+++ b/src/apps/macros.js
@@ -1,5 +1,5 @@
 const metadata = require('../lib/metadata')
-const { transformObjectToOption, transformStringToOption } = require('./transformers')
+const { transformObjectToOption, transformStringToOption, transformHQCodeToLabelledOption } = require('./transformers')
 
 const foreignOtherCompanyOptions = [
   'Charity',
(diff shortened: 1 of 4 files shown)


src/apps/macros.js
```diff
@@ -32,6 +32,22 @@ const globalFields = {
...
 
+  headquarter_type: {
+    macroName: 'MultipleChoiceField',
+    name: 'headquarter_type',
+    type: 'checkbox',
+    label: 'Type',
+    options () {
+      return metadata.headquarterOptions.map(transformHQCodeToLabelledOption)
+    },
+    // options: [
```
yozzo: do we still need the commented values?

PR #1382, opened 2018-04-05
DH-1646: HQ filters in incorrect order on company collections page

foxleigh81:
**Expected**: Global is top as in

> Global HQ
> European HQ
> UK HQ

**Actual**: As they're coming straight from the db they render in alphabetical order, so European is top:

> European HQ
> Global HQ
> UK HQ

[x] Added Lodash to macros.js 
[x] Used the Lodash `sortby` function to sort headquarter types by the upcoming `order` field.

**Note:** The `sortby` function is used in a way that will allow it to fallback to name or ID to sort. This means that this fix can go live even if https://github.com/uktrade/data-hub-leeloo/pull/898 has not been merged into master yet. When the `order` field exists in the data, then it will automatically start to sort by that field.

### DIFF
--- a/src/apps/macros.js
+++ b/src/apps/macros.js
@@ -1,3 +1,4 @@
+const { sortBy } = require('lodash')
 const metadata = require('../lib/metadata')
 const { transformObjectToOption, transformStringToOption, transformHQCodeToLabelledOption } = require('./transformers')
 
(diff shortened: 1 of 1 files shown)


reupen:
`/metadata/headquarter-type/` will return the items already sorted by order. So, you don't need to sort the returned items. `order` isn't included in the response for the metadata models that have it at the moment, so we'd have to add it if you want to use it.

PR #1391, opened 2018-04-11
DH-1140 Retrieve and display details of a selected event in the service delivery form

yozzo:
Because of the vast amount of data available to the users at one point,
we need to offer them the option to drill down and see more details about a category (i.e. events)
without changing the page they are on.
In order to acheieve this we had to crate the following functionalities:

-  api call for events details
- refactored the advisers option handler to be generic and accept multiple types of entities
- created a generic form item in the form builder, to use as a container for the details
- created a client side module that makes async calls for the entity details


src/templates/_macros/form.njk
```diff
@@ -28,4 +30,4 @@
 {% set SelectBox = SelectBox %}
 {% set TextArea = TextArea %}
 {% set TextField = TextField %}
-{% set Typeahead = Typeahead %}
+{% set Typeahead = Typeahead %}
```
ztolley: Do you have your editor setup to recognise .editorconfig files?

assets/javascripts/modules/fetch-from-item-details.js
```diff
@@ -0,0 +1,58 @@
+const axios = require('axios')
```
ztolley: So far the general rule of thumb has been to do stuff like this on the server side. If we are going to change that approach we should all talk about it. @jim68000 @yozzo thoughts?

My thoughts are to use client-side .js only when it adds a real benefit, like the XHR stuff.

If we do this on the client side, have we considered using Vue as we did with the typeahead, it gives a nice common way of structuring code and is already loaded.

yozzo:
feature reprioritised

PR #1397, opened 2018-04-20
Bugfix/test stability

ztolley:
Previously the tests would create a company with a random headquarters type but this made
tests fail if a company was created as an HQ as the title for the company page changed.

This change allows the getOptions function in tests accept an option to select and tells the test to pick the 'not a headquarters' option

### DIFF
--- a/test/acceptance/global.nightwatch.js
+++ b/test/acceptance/global.nightwatch.js
@@ -1,6 +1,6 @@
 module.exports = {
-  waitForConditionPollInterval: 2000,
-  waitForConditionTimeout: 12000,
-  pauseDuration: 7500,
-  retryAssertionTimeout: 5000,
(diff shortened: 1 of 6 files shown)


test/acceptance/global.nightwatch.js
```diff
@@ -1,6 +1,6 @@
 module.exports = {
-  waitForConditionPollInterval: 2000,
-  waitForConditionTimeout: 12000,
-  pauseDuration: 7500,
-  retryAssertionTimeout: 5000,
+  waitForConditionPollInterval: 1000,
```
jim68000: _Reducing_ the timeouts?

ztolley: We increased them in a previous PR because I suspected that was causing the issues, turns out I was wrong so I'm returning them to their original values.

PR #1398, opened 2018-04-23
DH-1663 - Added the Event interaction type switch to the add interaction page

yozzo:


yozzo:
part of PR #1409, closing it

PR #1399, opened 2018-04-23
Feature/dh 1665 remove events from standard interaction

yozzo:


yozzo:
part of PR #1409, closing it

PR #1400, opened 2018-04-23
DH-1439 Fix for multiple submits on IE11

yozzo:
Native IE11 seems to post a form twice, after a single click, as I didn't take in account the native HTML
behaviour of the submit button, so I moved the preventDefault to be at the top of the event callback


ztolley:
Rebase to get a bunch of fixes for acceptance tests

PR #1407, opened 2018-05-02
Update Yarn and Node to current versions

ztolley:


reupen:
Need to update the build pack version in manifest.yml as well (the latest is v1.6.23: https://github.com/cloudfoundry/nodejs-buildpack/releases).

PR #1409, opened 2018-05-03
Feature/dh 1664 create event invitation form view

yozzo:


yozzo:
feature reprioritised

PR #1410, opened 2018-05-04
Policy Feedback

ztolley:
Add support for new policy feedback interaction type.

- [x] Add support for filtering metadata by context
- [x] Refactor interaction creation and edit flow to simplify and improve efficiency
- [x] Only select options related to form being edited
- [x] Improve test coverage and approach
- [x] Add support for policy feedback type
- [x] Add tests for policy feedback
- [x] Add acceptance test for adding policy feedback

Blocked until the policy feedback service has been added to metadata in Leeloo.


ztolley:
Splitting out the service filter work as this needs to go into the front end before the backend can add the policy feedback service otherwise this will result in potentially bad data.

PR #1416, opened 2018-05-09
Remove enter three characters hint.

stuaxo:
See the trello ticket

https://trello.com/c/afOyJHL1/30-remove-at-least-three-characters-hint-texts-on-search-fields


ztolley:
One other thing to consider. The typeahead control requires at least 3 characters before it will lookup an adviser, should this also change?

stuaxo:
OK, just rebased - any chance you can sanity check that I haven't got it into a pickle.

If's good, I'll merge it.

stuaxo:
@ztolley any idea why it's complaining about cucumber ?

PR #1418, opened 2018-05-10
Update minor dependecies

ztolley:
Just trying to get the dependencies up to date for datahub to address any potential bugs or security threats.

Leaving the major update of webpack to a separate PR as it's a breaking upgrade.

Also leaving the govuk stuff, for now, the idea is this should be broken into a separate project.


ztolley:
Circle is failing as the docker image this relies on is out of date. Updating that

ztolley:
Did some testing on IE and realised query-string will not work on IE as it needs to have object.assign, which IE does not support.

PR #1420, opened 2018-05-10
Update babel

ztolley:
- [x] Simplify linting to remove the dependency on babel
- [x] Update Babel config to use babel-preset-env 

Babel recommends that users move to a new way of specifying modules and plugins for transpiling code.

By moving to babel-preset-env you can specify the target browsers and it handles analysing the code
and fetching required modules. This also allows new javascript syntax to be adopted without changing the babel config.


.eslintrc
```diff
@@ -1,6 +1,5 @@
 {
   "extends": "standard",
-  "parser": "babel-eslint",
```
reupen: Should be able to remove `babel-eslint` from the dependencies then?

Also might want to explicitly set parserOptions.ecmaVersion.

ztolley: Not seen the ecma version thing, will take a look. Will remove the babel eslint

ztolley: It gets the ecmaversion from standard js

reupen: Ah, that explains that.

PR #1421, opened 2018-05-14
Policy Feedback

ztolley:
Adds support for listing, viewing, editing and creating policy feedback interactions.  Supports permissions to only allow permitted users to view, edit and create.


src/apps/interactions/macros/kind-form.js
```diff
@@ -1,32 +1,37 @@
...
+      label: 'A service that you have provided',
+      hint: 'For example, account management, a significant assist or an event',
+    },
+  ]
+
+  if (permissions.includes('interaction.add_policy_feedback_interaction')) {
+    options.push({
+      value: 'policy_feedback',
+      label: 'Capture policy feedback',
+      hint: 'For example, and issue or comment on government policy from a company',
```
reupen: Typo here – should be 'For example, **an** issue' rather than 'For example, **and** issue'.

reupen:
Does this cover showing/hiding the policy feedback filter option on the interaction collection page?

ztolley:
Doesn't currently cover the filter. I'll take a look and if it's easy will add.

reupen:
@ztolley I didn't think the filter option was meant to appear for those without permission (seems to be at the moment).

ztolley:
Updated the filters to only show if the user has permission

PR #1424, opened 2018-05-16
Merge back from 4.22

jim68000:


ztolley:
Bit confused why it says out of date and updating the branch feels risky.

when I merge master back into develop i just do it from the command line rather than pr, though more risky

jim68000:
What do you do after you merge though? Force push with lease?

PR #1432, opened 2018-05-21
Only show subsidiaries tab for global hq company

ztolley:


src/apps/companies/middleware/local-navigation.js
```diff
@@ -13,7 +13,7 @@ function setCompaniesLocalNav (req, res, next) {
 
   const navItems = LOCAL_NAV.filter((navItem) => {
     if (
-      (!headquarterType && navItem.path === 'subsidiaries') ||
+      (headquarterType !== 'ghq' && navItem.path === 'subsidiaries') ||
```
reupen: Can you make it return an expression directly instead of doing `if (...) return false; return true;`?

ztolley: OK, just isn't as nice to read

src/apps/companies/middleware/local-navigation.js
```diff
@@ -12,14 +12,10 @@ function setCompaniesLocalNav (req, res, next) {
   const companyNumber = get(company, 'company_number')
 
   const navItems = LOCAL_NAV.filter((navItem) => {
-    if (
-      (!headquarterType && navItem.path === 'subsidiaries') ||
+    return !(
```
reupen: You should be able to simplify this a bit more:

```javascript
const navItems = LOCAL_NAV.filter((navItem) => {
    return (navItem.path !== 'subsidiaries' || headquarterType === 'ghq') &&
        (navItem.path !== 'timeline' || !companyNumber)
  })
```

reupen: You should be able to simplify this a bit more:

```javascript
const navItems = LOCAL_NAV.filter((navItem) => {
    return (navItem.path !== 'subsidiaries' || headquarterType === 'ghq') &&
        (navItem.path !== 'timeline' || companyNumber)
  })
```

PR #1444, opened 2018-05-29
Instantly see new subsidiary links

ztolley:
Changes how subsidiaries are fetched so it uses the v3 API instead of the search API, that way the data is instantly available.

This fixes the issue of adding a subsidiary and not seeing it in the list when you press save.
subsidiaries

### DIFF
--- a/src/apps/companies/repos.js
+++ b/src/apps/companies/repos.js
@@ -1,7 +1,6 @@
 /* eslint camelcase: 0, prefer-promise-reject-errors: 0 */
 const config = require('../../../config')
 const authorisedRequest = require('../../lib/authorised-request')
-const { search } = require('../search/services')
 
 // Get a company and then pad out the interactions with related data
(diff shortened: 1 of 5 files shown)


src/apps/companies/repos.js
```diff
@@ -91,6 +91,19 @@ function getCompanyTimeline (token, companyId, page = 1) {
...
+function getCompanySubsidiaries (token, companyId, page = 1) {
+  const limit = 10
+  const offset = limit * (page - 1)
+  return authorisedRequest(token, {
+    url: `${config.apiRoot}/v3/company`,
+    qs: {
+      limit,
+      offset,
+      global_headquarters_id: companyId,
+    },
```
reupen: They're unsorted by default – pass `sortby: 'name'` to sort them by name for now?

ztolley: ok

PR #1445, opened 2018-05-29
Update copy in One List section and link for support

marcofucci:
This changes the copy in the One List section so that users can email the one list email address directly instead of going through the support team if they need changes.

*Before*

screen shot 2018-05-29 at 15 45 33

*After*

screen shot 2018-05-29 at 15 50 18

### DIFF
--- a/src/apps/companies/controllers/details.js
+++ b/src/apps/companies/controllers/details.js
@@ -1,3 +1,4 @@
+const config = require('../../../../config')
 const {
   transformCompanyToView,
   transformCompaniesHouseToView,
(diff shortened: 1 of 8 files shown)


src/apps/companies/controllers/details.js
```diff
@@ -13,6 +14,7 @@ function renderDetails (req, res) {
       companyDetails: transformCompanyToView(company),
       accountManagementDetails: transformCompanyToOneListView(company),
       chDetails: company.companies_house_data ? transformCompaniesHouseToView(company.companies_house_data) : null,
+      oneListEmail: config.oneList.email,
```
ztolley: Tests?

PR #1450, opened 2018-05-30
Remove Greenkeeper

ztolley:
Greenkeeper is designed to work with libraries not
applications and does not update or create pr's for
minor updates, so removing it.


ztolley:
Circle has been behaving strangely for sure, which may or may not be caused by this change, hence splitting this out into its own pull request.

Other builds also failed with the same issue and Circle makes it really hard to re-run the whole lot again.

The plan is merge the other changes in and then re-base this to trigger a re-run.

PR #1452, opened 2018-05-30
Reorganise and update dependencies

ztolley:
Organize dependencies to reduce security risks so only the required dependencies are loaded for live services.

Update minor dependency updates.

Remove uglify and shortid as not needed/


reupen:
It still doesn't deploy to PaaS – it's complaining about browser-sync-webpack-plugin at the moment.

ztolley:
DOH! - missing a commit

reupen:
It gets further now, but the app doesn't start due to not being able to load `basic-auth`.

ztolley:
ok I managed to start it locally after a minor change. This look better now @reupen ?

PR #1455, opened 2018-06-01
Lock down form fields for One List companies

marcofucci:
This locks down the fields `sector` and `headquarter_type` for companies on the One List.
Changing those fields in that case should only be allowed by specific people and only via the admin area.

screen shot 2018-06-01 at 10 15 46

### DIFF
--- a/src/apps/companies/controllers/edit.js
+++ b/src/apps/companies/controllers/edit.js
@@ -1,6 +1,11 @@
 const { assign, find, get, isEmpty } = require('lodash')
 
-const { transformCompaniesHouseToView } = require('../transformers')
+const config = require('../../../../config')
+const {
+  transformCompanyToView,
+  transformCompaniesHouseToView,
+} = require('../transformers')
+
 const { buildUkOtherCompanyOptions, buildForeignOtherCompanyOptions } = require('../options')
 const UK_BRANCH_OF_FOREIGN_COMPANY_ID = 'b0730fc6-fcce-4071-bdab-ba8de4f4fc98'
 
@@ -31,6 +36,10 @@ function isForeignCompany (req, res) {
   return req.query.country === 'non-uk'
 }
 
+function isOnOneList (req, res) {
+  return !isEmpty(get(res.locals, 'company.classification'))
+}
+
 function getHeading (company, isForeign) {
   const action = company ? 'Edit' : 'Add'
   const type = isForeign ? 'foreign' : 'UK'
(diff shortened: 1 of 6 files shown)


src/apps/companies/controllers/edit.js
```diff
@@ -31,6 +36,14 @@ function isForeignCompany (req, res) {
   return req.query.country === 'non-uk'
 }
 
+function onOneList (req, res) {
+  if (!res.locals.company) {
+    return false
+  }
+
+  return !isEmpty(res.locals.company.classification)
+}
```
ztolley: or

return !isEmpty(get(res.locals, 'company.classification'))

src/apps/companies/controllers/edit.js
```diff
@@ -62,10 +75,13 @@ async function renderForm (req, res) {
     .breadcrumb(res.locals.company ? 'Edit' : 'Add')
     .render('companies/views/edit', {
       isForeign,
+      onOneList: onOneList(req, res),
+      companyDetails: res.locals.company ? transformCompanyToView(res.locals.company) : {},
       businessTypeLabel,
       showTradingAddress,
       showCompanyNumber: businessType === UK_BRANCH_OF_FOREIGN_COMPANY_ID,
       heading: getHeading(res.locals.company, isForeign),
+      oneListEmail: config.oneList.email,
```
ztolley: what if that value isn't there?  `get(config, 'onelist.email')` maybe?

test/acceptance/features/step_definitions/form.js
```diff
@@ -91,3 +91,10 @@ Then(/^there are form fields$/, async function (dataTable) {
       .assert.visible(selectors.element)
   }
 })
+
+Then(/^I can(not)? see the field "(.+)"$/, async function (negate, fieldName) {
```
ztolley: 👍

test/unit/apps/companies/controllers/edit.test.js
```diff
@@ -98,31 +98,44 @@ describe('Company edit controller', () => {
       })
 
       it('should render the edit page', () => {
-        expect(this.resMock.render.getCall(0).args[0]).to.equal('companies/views/edit')
+        expect(this.resMock.render.firstCall.args[0]).to.equal('companies/views/edit')
```
ztolley: I wonder if we are getting the firstCalls.args[0] over and over, should it set that on 'this' in the beforeEach?

src/apps/companies/controllers/edit.js
marcofucci: If the value isn't there we want it to fail badly so that we know instead of suppress it silently.

marcofucci:
@ztolley changes made, have another look

src/apps/companies/controllers/edit.js
ztolley: In that case wrap this all in a try catch and in the catch call `next(error)`

marcofucci:
That should be an uncaught (unexpected) error so try/catch is technically incorrect and it should be left uncaught and dealt with by the framework (or custom global error handler). 
Is this how you are usually dealing with this? 🤔

PR #1457, opened 2018-06-05
Filter subsidiary suggestions to exclude Global HQ

ztolley:
When a user adds a subsidiary the list of companies returned by search should exclude companies that are flagged as a global HQ.

### DIFF
--- a/src/apps/companies/middleware/collection.js
+++ b/src/apps/companies/middleware/collection.js
@@ -1,5 +1,6 @@
 const { pick, pickBy, assign } = require('lodash')
 
+const { getOptions } = require('../../../lib/options')
 const { search, searchLimitedCompanies, searchCompanies } = require('../../search/services')
 const { transformApiResponseToSearchCollection } = require('../../search/transformers')
 const {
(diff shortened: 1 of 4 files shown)


src/apps/companies/middleware/collection.js
```diff
@@ -7,6 +7,13 @@ const {
   transformCompaniesHouseToListItem,
 } = require('../transformers')
 
+const globalHQId = '43281c5e-92a4-4794-867b-b4d5f801e6f3'
+const nonGLobalHQId = [
```
reupen: First l should be in lower case here (also would add an s on the end).

src/apps/companies/middleware/collection.js
```diff
@@ -7,6 +7,13 @@ const {
...
+  null,
+  '3e6debb4-1596-40c5-aa25-f00da0e05af9',
```
reupen: Give these two names as well so that we know what they are?

src/apps/companies/middleware/collection.js
ztolley: Changed the behaviour to fetch the ID's from the server, it was bugging me. See if this is any better

PR #1459, opened 2018-06-08
DHI-88 Display for SPI type interactions `Created on` value in the details view

yozzo:
DHI-88 Display for SPI type interactions `Created on` value in the details view

In order to prevent users from fudging the interaction creation dates, for IST specific interactions - SPIs - steakholders requested to display the system generated date amongst the manually created one

### DIFF
--- a/test/unit/apps/interactions/transformers.test.js
+++ b/test/unit/apps/interactions/transformers.test.js
@@ -814,6 +814,123 @@ describe('Interaction transformers', () => {
       })
     })
 
+    context('when provided with an IST only specific service ', () => {
+      beforeEach(() => {
+        const istService = assign({}, mockInteraction, {
+          service: {
+            name: 'Investment - IST Aftercare (IST use only)',
+            id: 'bd3d1be3-f9f7-e511-888e-e4115bead28a',
+          },
+        })
+
+        this.transformed = transformInteractionResponseToViewRecord(istService)
+      })
+
+      it('should transform to display format', () => {
+        expect(this.transformed).to.deep.equal({
+          'Company': {
+            url: '/companies/dcdabbc9-1781-e411-8955-e4115bead28a',
+            name: 'Samsung',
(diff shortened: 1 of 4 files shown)


test/unit/apps/interactions/transformers.test.js
```diff
@@ -814,6 +814,123 @@ describe('Interaction transformers', () => {
...
+            id: 'bd3d1be3-f9f7-e511-888e-e4115bead28a',
+          },
+        })
+
+        this.transformed = transformInteractionResponseToViewRecord(istService)
+      })
+
+      it('should transform to display format', () => {
+        expect(this.transformed).to.deep.equal({
+          'Company': {
```
yeahfro: could we do something to soak this up?

test/unit/apps/interactions/transformers.test.js
```diff
@@ -814,6 +814,123 @@ describe('Interaction transformers', () => {
...
+    })
+
+    context('when provided with a non IST specific service ', () => {
+      beforeEach(() => {
+        this.transformed = transformInteractionResponseToViewRecord(mockInteraction)
+      })
+
+      it('should transform to display format', () => {
+        expect(this.transformed).to.deep.equal({
+          'Company': {
```
yeahfro: could we do something to soak this up?

ztolley: YOU ARE BACK!

yozzo: Don't really understand what you mean by soaking in this context, but if you mean something like caching the object in a variable somewhere higher up and then just comparing the key differences, thus using a different pattern to the actual one, so adding extra logic in this case would make it harder to read overall. Welcome back.

yozzo: @yeahfro :nudge :nudge

PR #1460, opened 2018-06-12
Allow multiple policy areas in policy feedback

ztolley:
Previous the backend API had a field named 'policy_area' when creating a policy feedback. The model has changed to allow multiple areas to be specified, and the field associated with that given a new name.

This change:

[x] Refactors interactions to make it easier to add the change
[x] Uses 'policy_areas' instead of 'policy_area' and makes it support an array of values
[x] Fixes some quirks with form display and error display in a fieldset.


src/apps/interactions/transformers/interaction-form-body-to-api.js
```diff
@@ -1,13 +1,18 @@
-const { omit } = require('lodash')
+const { castArray, compact, omit } = require('lodash')
 
 const { transformDateObjectToDateString } = require('../../transformers')
 
+const castToArrayAndRemoveEmpty = (value) => compact(castArray(value))
```
yozzo: is there a reason why you're using a function declaration instead of a function expression here?

reupen:
Something odd happening with the styling?

image

The border is too wide and the font of the error message doesn't look right (although it's squished back down here).

Other add-another fields look fine:

image

src/apps/interactions/transformers/interaction-form-body-to-api.js
```diff
@@ -0,0 +1,19 @@
+const { castArray, compact, omit } = require('lodash')
+
+const { transformDateObjectToDateString } = require('../../transformers')
+
+const castToArrayAndRemoveEmpty = (value) => compact(castArray(value))
```
reupen: This is also defined in `src/apps/events/transformers.js` and `src/apps/investment-projects/middleware/forms/requirements.js` also does `compact(castArray(...))`. Can the function be defined somewhere common and then used consistently?

ztolley:
That's strange because they all use the same components, no styling was done at all for this. I'll take a look in case there is something I'm missing but not sure what.

src/apps/interactions/transformers/interaction-form-body-to-api.js
ztolley: I only didn't do that because it's a single line and felt like overkill, but can do

reupen:
It seems like the wrong border size might be because there's no left nav on the page.

As for the error message font and placement, the client contacts field mentioned doesn't seem to be using the same macro.

test/acceptance/pages/interactions/interaction.js
```diff
@@ -44,7 +44,7 @@ module.exports = {
     eventNo: 'label[for=field-is_event-2]',
     event: '#field-event',
     policyIssueType: '#field-policy_issue_type',
-    policyArea: '#field-policy_area',
+    policyArea: '#field-policy_areas',
```
yeahfro: Hello @ztolley 😃 

Strictly speaking this should belong to the commit before because the build is broken until this commit is applied.

ztolley: Github has done its usual thing of getting the commit order wrong after a rebase, but you are correct. I'll merge those two. Welcome back, we missed you.

ztolley: But, generally, I'm of the mind that it's ok to have a commit that fails tests, the idea is you have a branch that must pass them. In some TDD shops the first thing they do is create and commit tests that will fail and then say the feature it done when all the tests pass.

src/apps/interactions/transformers/interaction-form-body-to-api.js
ztolley: Its just a copy of something used elsewhere and the syntax seemed neat and compact. I may even do away with this and just do it inline

src/apps/interactions/transformers/interaction-form-body-to-api.js
ztolley: I may just put this inline tbh

ztolley:
capture

[x] Fixed the style issues in the form
[x] Merged the commit with the fixes for acceptance tests
[x] Refactored casting to compact arrays to a shared function

reupen:
Looks better, except that one of the acceptance tests failed (around contact error messages).

ztolley:
Yup looking at that now.

PR #1462, opened 2018-06-15
Update node to 8.11.3 and Yarn to 1.7.0

ztolley:
Switches to the latest version of node and yarn and updates the buildpack used during deployment to use those versions.


jim68000:
👍

PR #1464, opened 2018-06-15
Feature/dh 1664 create event invitation form view

yozzo:
Because of the vast amount of data available to the users at one point,
we need to offer them the option to drill down and see more details about a category (i.e. events)
without changing the page they are on.
In order to acheieve this we had to crate the following functionalities:

new api call for events details
refactored the advisers option handler to be generic and accept multiple types of entities
created a generic form item in the form builder, to use as a container for the details
created a client side module to make an asyncronous call for the entity details


yozzo:
removing this, as functionality doesn't really match events anymore

PR #1466, opened 2018-06-18
Reinstate sort contact list test

yeahfro:
Reinstating the acceptance test for sorting the contact list. It is understood that this was disabled because it was breaking and could not be fixed. It appears to be working now.

If this breaks again then I am happy to take responsibility for fixing it.

### DIFF
--- a/test/acceptance/features/contacts/collection.feature
+++ b/test/acceptance/features/contacts/collection.feature
@@ -83,31 +83,29 @@ Feature: View collection of contacts
     And I click the local header link
     Then the company details UK region is displayed
 
+   @contacts-collection--sort
+   Scenario: Sort contact list
 
-# TODO: This test is causing PR's to fail, it should be looked at when we have time to fix it.
-#   @contacts-collection--sort
-#   Scenario: Sort contact list
-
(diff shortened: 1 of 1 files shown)


test/acceptance/features/contacts/collection.feature
```diff
@@ -83,31 +83,29 @@ Feature: View collection of contacts
...
+     Then I wait and then refresh the page
+     When I navigate to the `contacts.list` page
+     When the contacts are sorted by Newest
+     When the contacts are sorted by Oldest
+     Then the contacts should have been correctly sorted by creation date
+     And the results are sorted by Recently updated
+     Then the results should be sorted by Recently updated
+     And the results are sorted by Least recently updated
+     Then the results should be sorted by Least recently updated
+  #    When the contacts are sorted by Last name: A-Z
```
yozzo: is this potential bug still being investigated?

ztolley: I think the issue was the test was fragile, enable it again if you wish but if the tests start to fail we will have to turn of off again.

PR #1467, opened 2018-06-18
Add event attendee

ztolley:
When viewing the event attendee list the user can search for a contact to add, once selected a service delivery is automatically created and the user taken straight back to the attendee list screen with the new attendee visible.

While on this I made some subtle changes to acceptance tests to make them more generic, consistent with the use of " and flexible to test this ticket.

Refactored the attendee code so it's now a sub-app of events, makes managing and navigating the code easier.

add_attendee


src/apps/events/attendees/controllers/create.js
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    const contact = await getContact(token, contactId)
+
+    const serviceDelivery = {
+      contact: contact.id,
+      company: contact.company.id,
+      date: event.start_date,
+      dit_adviser: adviser.id,
+      dit_team: event.lead_team.id,
```
reupen: Use lodash `get()`?

src/apps/events/attendees/controllers/find.js
```diff
@@ -0,0 +1,74 @@
...
+    const query = req.query
+    const searchTerm = res.locals.searchTerm = get(req, 'body.term', '').trim()
+
+    const contacts = await search({
+      searchTerm,
+      searchEntity: 'contact',
+      token: req.session.token,
+      page: query.page,
+    })
+      .then(transformApiResponseToSearchCollection(
```
reupen: Can we use two `await`s instead of `.then()`?

ztolley: We use this syntax everywhere, so was done here to be consistent.

reupen: OK, but it's quite hard to read (and this one seems worse than the others). Maybe you can move the arrow function out (assign it to a variable, use a function declaration or whatever...)

ztolley: yip I'll move it out to a function

ztolley: ah actually it has to be inline so it can pull in the event id so I can't do that.

test/acceptance/features/events/attendees.feature
```diff
@@ -0,0 +1,24 @@
...
+  When I click the "Attendees" link
+  And I click the "Add attendee" link
+  Then I see a search box
+  When I search for "Dean Cox"
+  Then I see "Dean Cox" in the search results
+  When I select "Dean Cox" in the search results
+  Then the results summary for an attendee collection is present
+  And I see "Dean Cox" in the collection
+  When I click the "View or edit service delivery" link
+  Then I can see the service delivery details
```
yeahfro: Maybe align `I can see` with `I see` - as above

test/acceptance/features/companies/subsidiaries.feature
```diff
@@ -4,7 +4,7 @@ Feature: Company details
 @companies-heirarchies-set-hq
   Scenario: Set a company to a Global HQ
     When I navigate to the `companies.fixture` page using `company` `Lambda plc` fixture
-    Then I click the 'Edit company details' link
+    Then I click the "Edit company details" link
```
yeahfro: 👍

PR #1470, opened 2018-06-21
Add sector to filter options and list display for omis orders

ztolley:


reupen:
The filter doesn't seem to be having any effect. Is it using `sector_descends`?

src/apps/omis/apps/list/controllers.js
```diff
@@ -1,31 +1,81 @@
...
+      selectedFilters,
+      filtersFields,
+      countLabel: 'order',
+      actionButtons: [{
+        label: 'Add order',
+        url: '/omis/create',
+      }],
+    })
+  } catch (error) {
+    console.log(error)
```
reupen: Did you mean to put a `console.log()` here?

reupen:
Now it's having an effect, but if you select and change the sectors too quickly, you end up with the wrong results.

ztolley:
The issue with clicking filters too fast it a general problem on all filter pages, caused by how the XHR module works. Maybe time to revisit it as part of moving the code to a shared framework

PR #1472, opened 2018-06-22
Bug/contact active filter

yeahfro:
https://trello.com/c/POfOIgX8/136-fix-active-inactive-filter-on-contacts-page

# Problem
Active/inactive filters were not working correctly because the query string parameter was being passed as an array element. Arrays are set to `null`.

# Solution
Specify the `arrayFormat` to be `repeat` so that if just one filter is selected then it is not understood as an array in the middleware.

In addition I have:
* Added unit tests around `getRequestBody` middleware
* Refactored code
* Updated acceptance tests

### DIFF
--- a/src/apps/contacts/middleware/collection.js
+++ b/src/apps/contacts/middleware/collection.js
@@ -1,4 +1,4 @@
-const { pick, mapValues, isArray, pickBy, get } = require('lodash')
+const { pick, pickBy } = require('lodash')
 
 const { search } = require('../../search/services')
 const { transformApiResponseToSearchCollection } = require('../../search/transformers')
@@ -34,10 +34,6 @@ function getRequestBody (req, res, next) {
     'address_country',
     'company_uk_region',
   ]), 'archived')
-
-  mapValues(get(selectedFiltersQuery, 'archived'), (value) => {
-    return isArray(value) ? null : value
-  })
 
   const selectedSortBy = req.query.sortby
     ? { sortby: req.query.sortby }
--- a/src/apps/contacts/middleware/collection.js
+++ b/src/apps/contacts/middleware/collection.js
@@ -39,7 +39,11 @@ function getRequestBody (req, res, next) {
     ? { sortby: req.query.sortby }
     : null
 
-  req.body = Object.assign({}, req.body, selectedSortBy, pickBy(selectedFiltersQuery))
+  req.body = {
+    ...req.body,
+    ...selectedSortBy,
+    ...pickBy(selectedFiltersQuery),
+  }
 
   next()
 }
--- a/test/unit/apps/contacts/middleware/collection.test.js
+++ b/test/unit/apps/contacts/middleware/collection.test.js
@@ -1,0 +1,107 @@
+describe('Contact collection middleware', () => {
+  beforeEach(() => {
+    this.middleware = require('~/src/apps/contacts/middleware/collection')
+    this.req = {
+      body: '',
+      query: {
+        name: 'contact name',
+        company_name: 'company name',
+        company_sector_descends: [
+          '9b38cecc-5f95-e211-a939-e4115bead28a',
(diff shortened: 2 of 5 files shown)


test/unit/apps/contacts/middleware/collection.test.js
```diff
@@ -0,0 +1,88 @@
...
+      },
+    }
+    this.res = {}
+    this.nextSpy = sinon.spy()
+  })
+
+  describe('#getRequestBody', () => {
+    context('when supplied with default parameters', () => {
+      it('should set the request body', () => {
+        const req = assign({}, this.req, {
```
ztolley: Starting to adopt ... cos it's new and cool.

test/unit/apps/contacts/middleware/collection.test.js
```diff
@@ -0,0 +1,88 @@
...
+          '934cd12a-6095-e211-a939-e4115bead28a',
+        ],
+      },
+    }
+    this.res = {}
+    this.nextSpy = sinon.spy()
+  })
+
+  describe('#getRequestBody', () => {
+    context('when supplied with default parameters', () => {
```
ztolley: Even it's just one test, put the setup in a `beforeEach` so the 'it' just focuses on the assertion

test/unit/apps/contacts/middleware/collection.test.js
```diff
@@ -0,0 +1,88 @@
...
+            '934cd12a-6095-e211-a939-e4115bead28a',
+          ],
+          name: 'contact name',
+        })
+      })
+    })
+
+    context('when archived is set twice as true and false', () => {
+      it('should not set archived on the request body', () => {
+        const req = assign({}, this.req, {
```
ztolley: why mix assign and ...

also you could just assign the result of the creation of the new value back to this.req

src/apps/contacts/middleware/collection.js
```diff
@@ -35,10 +35,6 @@ function getRequestBody (req, res, next) {
     'company_uk_region',
   ]), 'archived')
 
-  mapValues(get(selectedFiltersQuery, 'archived'), (value) => {
```
ztolley: 👍

src/apps/contacts/middleware/collection.js
```diff
@@ -39,7 +39,11 @@ function getRequestBody (req, res, next) {
     ? { sortby: req.query.sortby }
     : null
 
-  req.body = Object.assign({}, req.body, selectedSortBy, pickBy(selectedFiltersQuery))
+  req.body = {
```
ztolley: 👍

test/unit/apps/contacts/middleware/collection.test.js
jim68000: Did you read the David Flanagan thing about how ... is really a loop and should be treated with care though?

ztolley: I did not, plus I see how you have used ... elsewhere.  Is it that evil that we shouldn't use it?

ztolley: I wonder if that is only with script passed through babel which is then converted to es5 with lots of loops. In our case it's native

Anyway, after looking at the rest of the code it uses ... all over so maybe do it here to be consistent.

yozzo: I wouldn't penalise it, as it takes away from readability, and when it comes to objects it doesn't support deep nesting, so it kinda does half the job, but if we're to take the shorthand `...` route we should make it official and add it to the coding standards.

NB: I have the same opinion about ternary operators , and if we're to take any greenhorns on board  (e.g: designers helping rollout the fresh new webkit)  the  `foo ? bar : (bas) => {... baz}` syntax style will feel very daunting.

PR #1478, opened 2018-06-27
Add service filter to interaction collection

ztolley:

### DIFF
--- a/src/apps/events/attendees/controllers/find.js
+++ b/src/apps/events/attendees/controllers/find.js
@@ -1,4 +1,4 @@
-const { get } = require('lodash')
+const { get, isEmpty } = require('lodash')
 
 const { search } = require('../../../search/services')
 const { transformApiResponseToSearchCollection } = require('../../../search/transformers')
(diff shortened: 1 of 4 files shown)


src/lib/options.js
```diff
@@ -43,7 +47,7 @@ async function getOptions (token, key, { createdOn, currentValue, includeDisable
   return sorted ? sortBy(mappedOptions, 'label') : mappedOptions
 }
 
-async function getOptionsForId (token, key, id) {
+async function getOptionsForEntityId (token, key, id) {
```
yeahfro: Is there any way this `getOptions` can be improved to avoid needing `getOptionsForMetadataId ` that is doing something very similar?

ztolley: They work different, one filters on the server the other on the node side. There's no support for filtering metadata items so this was the simplest solution

ztolley: Talking to @reupen to see if metadata can be made to act like v3 so I'd then go back to only needing the one common function

ztolley: going to switch it back to a regular radio list so this change will be reversed

PR #1481, opened 2018-06-27
Add service as filter in Interactions

ztolley:
Adds 'Service' as a filter option in the interactions list. Also adds the service in the result card

### DIFF
--- a/src/apps/interactions/controllers/list.js
+++ b/src/apps/interactions/controllers/list.js
@@ -1,5 +1,3 @@
-const { get } = require('lodash')
-
 const { collectionFilterFields } = require('../macros')
 const { buildSelectedFiltersSummary, buildFieldsWithSelectedEntities } = require('../../builders')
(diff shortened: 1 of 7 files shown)


src/apps/interactions/controllers/list.js
```diff
@@ -8,26 +6,34 @@ const FILTER_CONSTANTS = require('../../../lib/filter-constants')
 const QUERY_STRING = FILTER_CONSTANTS.INTERACTIONS.SECTOR.PRIMARY.QUERY_STRING
 const SECTOR = FILTER_CONSTANTS.INTERACTIONS.SECTOR.NAME
 
+async function getInteractionOptions (token) {
```
yeahfro: These changes would have been better if they were split out

1. adding the filter
2. refactor

yeahfro: Other thoughts around this function are whether we should be going it in middleware like other applications

PR #1483, opened 2018-06-28
Refactor get collection middleware for contacts

yeahfro:
Description of change:
- start at `/src/apps/contacts/middleware/collection.js`
- concept of modules
-- search
-- permissions
-- API
- moving away from horizontal splits such as `middleware`, `filters.js`, `builders.js`


src/modules/api/transformers/api-response-to-collection.js
```diff
@@ -0,0 +1,44 @@
...
+const { buildPagination } = require('../../../lib/pagination')
+
+/**
+ * @param {object} [options] {object}
+ * @param {string} [options.searchTerm] - search term used for highlighting words in collection macro
+ * @param {object} [options.query] - URL query object used in pagination
+ * @param {...function} [itemTransformers] - an array of transformer functions to apply for each item in the list
+ * @returns {object, undefined}
+ */
+function transformApiResponseToCollection (options = {}, ...itemTransformers) {
```
yozzo: Now that we're at the point of refactoring stuff, I find the three callbacks here a bit hard to follow and I think having each function declared individually outside the scope of `transformApiResponseToCollection`  and  only executing them inside would tick my box of clean refactoring. Also passing `items` as an argument to the arrow function, rather than accessing it from the parent function would make it more clear, I'm not a big fan of scope leaks

src/modules/search/constants.js
```diff
@@ -0,0 +1,68 @@
+// TODO this does not feel like the right place for this as the search module should not know about apps
```
yozzo: it does what it says on the label, `constants`, I'm happy with that.

src/modules/search/services.js
```diff
@@ -0,0 +1,35 @@
+const { assign } = require('lodash')
+
+const authorisedRequest = require('../../lib/authorised-request')
+const config = require('../../../config')
+
+function search ({ token, searchTerm = '', searchEntity, requestBody, isAggregation = true, limit = 10, page = 1 }) {
```
yozzo: aligning arguments in a column would make it more readable.

ztolley: I started using prettier elsewhere, and it solves discussions like this

src/modules/api/transformers/api-response-to-collection.js
yeahfro: Most of this I would agree with but it's outside the scope of this change

src/modules/search/constants.js
yeahfro: I am not comfortable with this common library having knowledge of the dependent applications

src/apps/contacts/middleware/collection.js
```diff
@@ -1,27 +1,7 @@
...
-    next()
-  } catch (error) {
-    next(error)
-  }
-}
+const { getCollection } = require('../../../middleware/collection')
 
 module.exports = {
-  getContactsCollection,
+  getContactsCollection: getCollection('contact', transformContactToListItem),
```
yeahfro: Is everyone happy with this usage? Should we just sack of `collection.js` altogether and do this in `router.js`?

yozzo: I'd go for the `router` option

src/modules/search/services.js
ztolley: Going forward I'd advise people to change the lint config and adopt prettier & Airbnb. A config for this can be found in https://github.com/ztolley/blubber/blob/master/package.json

It will be a mahoosive change though as the one thing prettier won't support is `function doStuff (param)` where there is a space between the function name and the bracket, and the author refuses to allow people to do that.

src/modules/api/transformers/api-response-to-collection.js
ztolley: I tend to refactor as I touch code, though I know this can cause inconsistent code :/

PR #1485, opened 2018-07-03
Improve authorised request

ztolley:
Previously the authorised request function would send things such as `qs=undefined` or `term=''`. This strips out empty params, body and headers

Also removes the error transformer for V2 calls as we no longer make those calls.

### DIFF
--- a/src/lib/authorised-request.js
+++ b/src/lib/authorised-request.js
@@ -1,7 +1,12 @@
+const { isNil, pickBy, set } = require('lodash')
 const request = require('request-promise')
-const config = require('../../config')
 
+const config = require('../../config')
 const logger = require('../../config/logger')
+
+function hasValue (value) {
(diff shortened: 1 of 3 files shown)


src/lib/authorised-request.js
```diff
@@ -30,28 +35,20 @@ function jsonReviver (key, value) {
...
-    requestOptions.headers = {}
-  } else {
-    requestOptions.url = opts.url
-    requestOptions.headers = opts.headers || {}
-    requestOptions.method = opts.method || 'GET'
-    requestOptions.qs = opts.qs
-
-    if (opts.body) {
-      requestOptions.body = opts.body
+  let requestOptions = (isString(opts))
```
yeahfro: Just an observation: I find this hard to read but I also find the previous version hard to read too.

PR #1487, opened 2018-07-05
Stop user adding attendees to incomplete event

ztolley:
Some events do not have a lead team or service which are required when creating a service delivery.

If the event is incomplete, block the user from adding an attendee.

### DIFF
--- a/src/apps/events/attendees/controllers/list.js
+++ b/src/apps/events/attendees/controllers/list.js
@@ -18,33 +18,40 @@ async function renderAttendees (req, res, next) {
     const page = query.page || 1
     const token = req.session.token
     const sortby = req.query.sortby || defaultAttendeeSort
+    const incompleteEvent = (!event.service || !event.lead_team)
 
-    const sortForm = merge({}, attendeeSortForm, {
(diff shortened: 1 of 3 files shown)


src/apps/events/attendees/controllers/list.js
```diff
@@ -18,33 +18,40 @@ async function renderAttendees (req, res, next) {
     const page = query.page || 1
     const token = req.session.token
     const sortby = req.query.sortby || defaultAttendeeSort
+    const incompleteEvent = (!event.service || !event.lead_team)
```
yeahfro: I am feeling a bit funny about this `incompleteEvent` name. I would find it more readable if it was `const completeEvent = (event.service && event.lead_team)`

Understand this changes a few things but it will be much easier to read `if (completeEvent)`

ztolley: I like to do positive comparisons and in the template I say show something if the event is incomplete

ztolley: Also, it feels odd to say that something is complete if it has just 2 properties, but you know it's incomplete is 2 particular properties are not there.

PR #1495, opened 2018-07-11
Event attendees - double-clicking adds the same attendee twice

yozzo:
Some users still double click on links and this creates an unwanted behaviour, i.e. multiple entries.
I've done some discovery on what's the best way to approach this, and because in the node layer it creates multiple "threads" you can't really control the state of multiple requests unless you create an observer pattern in the authorised-request file.

Knowing that the authorised-request file is being used with almost all request, this pattern is not fit for purpose so I decided to solve it very simply in the client by preventing any double or multi click actions on a link with a predefined class, thus preserving the native browser behaviour.

### DIFF
--- a/assets/javascripts/app.js
+++ b/assets/javascripts/app.js
@@ -16,6 +16,7 @@ const PrintDialog = require('./modules/print-dialog')
 const MirrorValue = require('./modules/mirror-value.js')
 const ClearInputs = require('./modules/clear-inputs.js')
 const PreventMultipleSubmits = require('./modules/prevent-multiple-submits.js')
+const PreventLinkDoubleClick = require('./modules/prevent-link-double-click.js')
 
 const CompanyAdd = require('./_deprecated/company-add')
 const CompanyEdit = require('./_deprecated/company-edit')
@@ -33,6 +34,7 @@ PrintDialog.init()
 MirrorValue.init()
 ClearInputs.init()
 PreventMultipleSubmits.init()
+PreventLinkDoubleClick.init()
 
 // Deprecated
 CompanyAdd.init()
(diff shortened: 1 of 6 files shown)


ztolley:
@marcofucci and @reupen Both raised concerns that we use GET instead of POST

yozzo:
@ztolley that would require a completely new view for 'search attendees results' and the use of a UI pattern like radio-buttons. No matter how you put it, you can't post a form through the click of an anchor tag. I haven't looked into how the list is generated but I'm not sure it's a quick change which means it would need to be prioritised otherwise could affect our (doc upload) current sprint scope.

reupen:
> that would require a completely new view for 'search attendees results' and the use of a UI pattern like radio-buttons.

Not sure about that. Surely buttons can be styled as you want? Then it would just be a list of buttons styled as it currently looks.

It's also missing CSRF protection at the moment.

yozzo:
@reupen styling is not the problem here, but it's more about web standards - W3C et all, otherwise anchors would've worked perfectly fine. In the simplest shape with no javascript enabled you would have a form with N radio-buttons with the values of the attendees which the user is only allowed to click once, hence the radio-button, then press submit. 

Otherwise going for the <form> with multiple buttons or anchors, would involve a hidden input that uses client-side scripting to get the button or anchor value and copy it to the hidden input, and only after that submitting it, isn't a straight forward pattern, and takes away the native features of browsers.

I agree with CRSF but don't think it's in the scope of this bug-fix, neither is the view refactor in any agreed form that sends POSTs instead of GETs which should be treated as a change request.

ztolley:
You could have one big long form with a button visible when no js, the CSS would say when js is set on the body class, hide them, and another bit of JS would add event handlers to the results which would set the radio button and submit the form. 

Best to talk this through on Slack or a hangout and invite Lee and Marco

yozzo:
https://trello.com/c/ZnsFf3u9/171-event-attendees-double-clicking-adds-the-same-attendee-twice

PR #1497, opened 2018-07-13
Avoid duplicate attendees

ztolley:
This ticket tackles the issue in 2 ways but is separate from the issue of users double clicking.

capture

When a user searches for contacts to add as event attendees, if a result has already been added, its style is modified to use the highlight background colour, a badge to indicate it has already been added and nothing happens if they try and click the result.

In addition, when a call is made to the 'create attendee' controller, it checks the attendee list for the event to see if it contains the contact passed to it. If that contact has previously been added then the user is shown an error.

### DIFF
--- a/src/apps/events/attendees/controllers/find.js
+++ b/src/apps/events/attendees/controllers/find.js
@@ -3,6 +3,7 @@ const { get, isEmpty } = require('lodash')
 const { search } = require('../../../search/services')
 const { transformApiResponseToSearchCollection } = require('../../../search/transformers')
 const { transformContactToListItem } = require('../../../contacts/transformers')
+const { createContactItemToAttendeeSearchResult } = require('../transformers')
 
 async function renderFindAttendee (req, res, next) {
   try {
@@ -25,6 +26,7 @@ async function renderFindAttendee (req, res, next) {
 async function findAttendee (req, res, next) {
   try {
     const event = res.locals.event
+    const token = req.session.token
 
     if (!event) {
       throw new Error('No event supplied')
@@ -37,9 +39,11 @@ async function findAttendee (req, res, next) {
       return next()
     }
 
-    const contacts = await search({
+    const transformListItemToAttendeeSearchResult = await createContactItemToAttendeeSearchResult(token, event)
+
+    const contactsResponse = await search({
+      token,
       searchEntity: 'contact',
-      token: req.session.token,
       page: query.page,
       requestBody: {
         archived: false,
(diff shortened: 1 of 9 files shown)


yozzo:
It all looks fine to me but this feature raises the question about how we're dealing the the attendee list as an anchor tag, doing a GET request instead of POST.

ztolley:
The aim of this change is to stop people from selecting results for people already added as the list an get very long. The issue of POST vs GET is a different one, hence a different ticket.

src/templates/_macros/entity/entity.njk
```diff
@@ -23,11 +23,15 @@
     {% set metaBadges = props.meta | filter(['type', 'badge']) %}
     {% set metaItems = props.meta | reject(['type', 'badge']) %}
     {% set highlightedName = props.name | highlight(props.highlightTerm) %}
-    {% set containerElement = 'a' if props.isBlockLink else 'div' %}
+    {% set containerElement = 'a' if props.isBlockLink and not props.disableLink else 'div' %}
```
yeahfro: seems to be a mix of naming conventions here.. + or - `is`. Are we okay with this?

src/apps/events/attendees/repos.js
```diff
@@ -10,12 +12,13 @@ async function fetchEventAttendees ({ token, eventId, page = 1, sortby, limit =
 
   const eventAttendees = await authorisedRequest(token, {
     url: `${config.apiRoot}/v3/interaction`,
-    qs: {
+    qs: pickBy({
```
yeahfro: if you spread this object rather than `pickBy` wouldn't it have the same desired effect?

yeahfro: and doesn't `pickBy` without the second argument do exactly the name?

src/apps/events/attendees/controllers/find.js
```diff
@@ -25,6 +26,7 @@ async function renderFindAttendee (req, res, next) {
 async function findAttendee (req, res, next) {
```
yeahfro: This commit could have been split, just saying

ztolley: Yeah I know, but at least I split it out to it's own commit

src/apps/events/attendees/repos.js
ztolley: no, it strips out things with a value of 0, which is required in some instances.

src/templates/_macros/entity/entity.njk
ztolley: hmm.. you think `isLinkDisabled` would be better?

PR #1498, opened 2018-07-16
Warning and hide add attendee for disabled event

ztolley:
* If an event is disabled then hide the add button and show a message explaining why you cannot add.
* Shows any attendees there may be if the event is incomplete as people can add attendees through another route.
* Shows indicator in search results if an event is disabled.
* Show indicator in event details if disabled.

### DIFF
--- a/src/apps/events/attendees/views/list.njk
+++ b/src/apps/events/attendees/views/list.njk
@@ -2,17 +2,22 @@
 
 {% block main_grid_right_column %}
   <h2 class="heading-medium">Event Attendees</h2>
-  {% if incompleteEvent %}
+  {% if event.disabled_on %}
+    <p class="c-message c-message--info">
+      You cannot add an event attendee because the event
+      has been disabled.
+    </p>
+  {% elif incompleteEvent %}
     <p class="c-message c-message--info">
       You cannot add an event attendee on this page because the event
       record is incomplete. You can create a service delivery interaction
       from the company record to add an event attendee.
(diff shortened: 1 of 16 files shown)


src/apps/events/attendees/views/list.njk
```diff
@@ -2,17 +2,22 @@
 
 {% block main_grid_right_column %}
   <h2 class="heading-medium">Event Attendees</h2>
-  {% if incompleteEvent %}
+  {% if event.disabled_on %}
+    <p class="c-message c-message--info">
+      You cannot add an event attendee to this page because the event
```
reupen: 'to this page' doesn't make much sense – the incomplete message below says 'on this page'.

ztolley: Yeah it should be some wording about event, not page.

ztolley:
@yeahfro hmm gonna be a pain to do that, there's no way to mark an event as disabled in the frontend so I'll have to add that in as test data... which highlights the issue that we don't do anything to handle disabled events too.. rats.... ok I may have to do pr's for adding test data and another one for showing the event is disabled in results, detail and disable the edit button...

src/apps/events/attendees/views/list.njk
ztolley: The incomplete text has previously been signed off. I'll remove it from this one as it doesn't make much sense, though technically you could add one through another route.

PR #1499, opened 2018-07-16
Only show sort options when there are more than 1 results

ztolley:


ztolley:
er yeah I guess acceptance tests just.. it's such a simple change

ztolley:
Added acceptance tests

PR #1513, opened 2018-07-31
Integrate with companies core-team endpoint

yeahfro:
This change integrates with the new Leeloo companies `core-team` endpoint, as follows:
- integration
- updated unit tests
- acceptance test

As the feature evolves, I will add more tests. Expecting the appearance to change as we get more involvement from design.

screen shot 2018-07-31 at 15 31 20

### DIFF
--- a/src/apps/companies/repos.js
+++ b/src/apps/companies/repos.js
@@ -105,6 +105,12 @@ function getCompanySubsidiaries (token, companyId, page = 1) {
   })
 }
 
+function getCoreTeam (token, companyId) {
+  return authorisedRequest(token, {
+    url: `${config.apiRoot}/v3/company/${companyId}/core-team`,
+  })
+}
+
 module.exports = {
   saveCompany,
   getDitCompany,
@@ -115,4 +121,5 @@ module.exports = {
   getCompanyAuditLog,
   getCompanyTimeline,
   getCompanySubsidiaries,
+  getCoreTeam,
 }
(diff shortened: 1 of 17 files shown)


src/apps/companies/transformers/core-team-to-collection.js
```diff
@@ -0,0 +1,23 @@
...
+      region ? { label: coreTeamLabels.region, value: region } : null,
+      { label: coreTeamLabels.country, value: get(adviser, 'dit_team.country.name') },
+      { label: coreTeamLabels.team, value: get(adviser, 'dit_team.name') },
+    ]
+
+    return {
+      name: adviser.name,
+      type: 'adviser',
+      subTitle: {
+        value: isGlobalAccountManager ? 'Global Account Manager' : null,
```
reupen: As discussed – when isGlobalAccountManager is false it seems to be displaying `[object Object]`

Screenshot with test data:
image

src/apps/companies/repos.js
```diff
@@ -105,6 +105,12 @@ function getCompanySubsidiaries (token, companyId, page = 1) {
   })
 }
 
+function getCoreTeam (token, companyId) {
+  return authorisedRequest(token, {
+    url: `${config.apiRoot}/v3/company/${companyId}/core-team`,
```
reupen: Something for another day, but is authorisedRequest() ever called with a URL not starting with `${config.apiRoot}`? I don't understand why we pass it all the time (instead of just passing the path).

PR #1522, opened 2018-08-08
Use one_list_account_owner instead of account_manager for project team

yeahfro:
https://trello.com/c/XjkJQA0D/111-investment-project-use-onelistaccountowner-instead-of-accountmanager

# Problem
1. The `Project team` tab for an investment project is displaying the `account_manager` whereas it should be displaying `one_list_account_owner`
2. The label should be `Global account manager`

# Solution
The overall change is straightforward but it was also discovered that there is no acceptance tests around the `Project team` tab. Changes as follows:
- Refactor of existing acceptance table assertion code to support `data-table`
- Transformation updates
- Unit test updates

# Further development
- Further tidy up but this change was already getting large

# Before
screen shot 2018-08-08 at 13 44 20

# After
screen shot 2018-08-08 at 13 38 56

### DIFF
--- a/src/apps/investment-projects/transformers/team.js
+++ b/src/apps/investment-projects/transformers/team.js
@@ -30,15 +30,17 @@ function transformClientRelationshipManagementForView (investmentData) {
   const result = [{
     role: 'Client relationship manager',
     adviser: getAdviserName(investmentData, 'client_relationship_manager'),
-    team: get(investmentData, 'client_relationship_manager.dit_team.name', null),
+    team: get(investmentData, 'client_relationship_manager.dit_team.name'),
   }]
-
-  const accountManager = get(investmentData, 'investor_company.account_manager.id', null)
-  if (accountManager) {
+  const globalAccountManager = get(investmentData, 'investor_company.one_list_account_owner.id')
+  if (globalAccountManager) {
+    const firstName = get(investmentData, 'investor_company.one_list_account_owner.first_name')
+    const lastName = get(investmentData, 'investor_company.one_list_account_owner.last_name')
+    const team = get(investmentData, 'investor_company.one_list_account_owner.dit_team.name')
     result.push({
-      adviser: get(investmentData, 'investor_company.account_manager.name', null),
-      role: 'Account manager',
-      team: get(investmentData, 'investor_company.account_manager.dit_team.name', null),
+      team,
+      adviser: `${firstName} ${lastName}`,
+      role: 'Global account manager',
     })
   }
 
--- a/test/acceptance/features/step_definitions/details.js
+++ b/test/acceptance/features/step_definitions/details.js
@@ -1,25 +1,35 @@
 const { client } = require('nightwatch-cucumber')
 const { Then } = require('cucumber')
-const { get, includes, startsWith } = require('lodash')
+const { get, includes, startsWith, keys } = require('lodash')
 
-const { getKeyValueTableRowValueCell } = require('../../helpers/selectors')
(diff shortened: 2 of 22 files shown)


test/acceptance/pages/investments/project.js
```diff
@@ -259,17 +259,17 @@ module.exports = {
...
-            projectDescription: getDetailsTableRowValue('Project description'),
-            anonDescription: getDetailsTableRowValue('Anonymised description'),
-            estimatedLandDate: getDetailsTableRowValue('Estimated land date'),
-            actualLandDate: getDetailsTableRowValue('Actual land date'),
-            newOrExistingInvestor: getDetailsTableRowValue('New or existing investor'),
-            levelOfInvolvement: getDetailsTableRowValue('Level of involvement'),
-            specificInvestmentProgramme: getDetailsTableRowValue('Specific investment programme'),
+            typeOfInvestment: getKeyValueTableRowValueCell('Type of investment'),
+            primarySector: getKeyValueTableRowValueCell('Primary sector'),
+            businessActivity: getKeyValueTableRowValueCell('Business activity'),
```
yozzo: usage of `getKeyValueTableRowValueCell` method here and the previous file seems a bit repetitive, wouldn't a function that takes an object with the key value pairs as an argument, which looping through each element would then call the `getKeyValueTableRowValueCell` method?

src/apps/investment-projects/transformers/team.js
```diff
@@ -30,15 +30,17 @@ function transformClientRelationshipManagementForView (investmentData) {
...
+  const globalAccountManager = get(investmentData, 'investor_company.one_list_account_owner.id')
+  if (globalAccountManager) {
+    const firstName = get(investmentData, 'investor_company.one_list_account_owner.first_name')
+    const lastName = get(investmentData, 'investor_company.one_list_account_owner.last_name')
+    const team = get(investmentData, 'investor_company.one_list_account_owner.dit_team.name')
     result.push({
-      adviser: get(investmentData, 'investor_company.account_manager.name', null),
-      role: 'Account manager',
-      team: get(investmentData, 'investor_company.account_manager.dit_team.name', null),
+      team,
```
reupen: Any reason for changing the order of the properties here?

test/acceptance/features/step_definitions/details.js
```diff
@@ -32,36 +42,42 @@ const removeFalsey = (details, state) => {
...
-      continue
-    }
-    if (row.key === 'Sector') {
-      // todo: https://uktrade.atlassian.net/browse/DH-1086
-      continue
+const assertTableContent = async function (tableSelector, expectedData, tableType) {
+  for (const row of expectedData) {
+    const rowKeys = keys(row)
+
+    for (var i = 1; i < rowKeys.length; i++) {
```
yozzo: //todo replace with lodash or ES6 loop forEach, map, etc

test/acceptance/pages/investments/project.js
yeahfro: @yozzo it is a fair suggestion but:
1. It is okay to be explicit rather than creating another abstraction
2. It would be a separate change as this PR is already getting big

I will have a play with the code after this and see how it plays out.

test/acceptance/features/step_definitions/details.js
yeahfro: Thanks for pointing this one out, but because it's using `async` I will need to resort to `for of` as in the parent loop

src/apps/investment-projects/transformers/team.js
yeahfro: Cheers @reupen, it's an agreed standard that we put implicit property values at the top of object definitions.

PR #1524, opened 2018-08-10
Rename parent to global headquarters to avoid disambiguation

marcofucci:
A parent/child relationship between companies assumes that there can be more levels (e.g. a child company with its own child).

The hierarchy we have at the moment is Global HQ and subsidiaries so it's better to avoid the word 'parent' and use global headquarters instead.

Also note that a Global HQ might not be the direct parent of a subsidiary in real life so using 'parent' is not 100% correct. We don't want to map the real world relationships between companies as it's too unnecessarily complex.

### DIFF
--- a/test/acceptance/features/investment-projects/team.feature
+++ b/test/acceptance/features/investment-projects/team.feature
@@ -7,7 +7,7 @@ Feature: View team for an investment project
     Then the Client relationship management data details are displayed
       | Role                        | Adviser                                          | Team                                             |
       | Client relationship manager | investmentProject.clientRelationshipManager.name | investmentProject.clientRelationshipManager.team |
-      | Global Account Manager      | investmentProject.globalAccountManager.name      |                                                  |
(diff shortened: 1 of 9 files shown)


test/acceptance/features/investment-projects/team.feature
```diff
@@ -7,7 +7,7 @@ Feature: View team for an investment project
     Then the Client relationship management data details are displayed
       | Role                        | Adviser                                          | Team                                             |
       | Client relationship manager | investmentProject.clientRelationshipManager.name | investmentProject.clientRelationshipManager.team |
-      | Global Account Manager      | investmentProject.globalAccountManager.name      |                                                  |
+      | Global Account Manager      | investmentProject.globalAccountManager.name      | IST - Sector Advisory Services                   |
```
yeahfro: 👍 hard coding here is better than burying it away in the `fixtures.js` - I felt like I wanted to change this earlier so I will update the others anyway

PR #1529, opened 2018-08-15
Change in wording for email marketing

jim68000:
Claire's asked to have the display of email marketing info changed to make it clearer whether someone is opted in or not. This turns that on.


ztolley:
Don't forget your fixup!

PR #1537, opened 2018-08-29
Investment Evidence Feature

yozzo:
The FDI Rulebook requires a section called `Evidence` where users can upload supporting documents for an investment project.
The Evidence feature itself is only a tags dropdown a file upload and a comments textarea. For documents upload we're reusing the `Documents` module,
and intsantiating it investment-projects/router by calling `postUpload` and binding the api configs necessary for building the request urls.

As Evidence is only required for Investment projects, we've created it inside an `apps` folder within the investment-projects app. This is where we will also need move the `propositions` app as it's not shared across data-hub.


yozzo:
clean PR here: https://github.com/uktrade/data-hub-frontend/pull/1544

PR #1539, opened 2018-08-31
Feature/email permissions

jim68000:
For such a simple change there’s a lot going on here, so I will explain.

The business need this change represents is to make it clearer to users that we can email market to contacts that have not explicitly opted out of our emails, under the GDPR public body exemption. It also means that contacts are defaulted to accept email marketing.

I therefore needed to create a new Boolean to make the negative version of ‘accepts email marketing’ work. Because the API hasn’t changed (because in turn the people that consume the contact list spreadsheet need it to be true for accepts marketing) this needs to be the inverse of the API value, and this needs to switch on loading from the API and on saving to the API. 

We have however simplified the API - there’s only one variable that controls marketing permissions now - `accepts_dit_email_marketing` when there used to be five, which the JS layer bundled into an object called `marketing_permissions`. This is relevant because of what happens in the next paragraph.

The control this Boolean is attached to is a single checkbox in the create/edit screen. Because we used to have the five marketing states this was controlled with a multiple choice macro which took the `marketing_permissions` Object and displayed a bunch of checkboxes. As there’s only one now, and I wanted to clean up the code by removing the `marketing_permissions` object, the multiple choice macro stopped working properly, because it was expecting an object and not a primitive. I’ve therefore created a standalone piece of code to control the checkbox in `edit.njk` - as this is the only instance of a standalone checkbox toggle in the code I haven’t converted it into a macro. I would suggest this as a todo if we ever create another.

Lastly there was an unexpected piece of brittleness in the property_helpers library. The function that converts strings to Boolean didn’t like it if a passed value was already a boolean.  This now typechecks to stop it trying to call toLocaleString() on a Boolean.

### DIFF
--- a/src/apps/contacts/views/edit.njk
+++ b/src/apps/contacts/views/edit.njk
@@ -83,19 +83,24 @@
       error: errors.messages.email
     }) }}
 
-    {{ MultipleChoiceField({
-      type: 'checkbox',
-      name: 'marketing_preferences',
-      label: 'Marketing preferences',
-      isLabelHidden: true,
-      value: formData.marketing_preferences,
-      options: [
-        {
-          label: 'Accepts email marketing from DIT',
-          value: 'accepts_dit_email_marketing'
-        }
-      ]
-    }) }}
+    <fieldset id="group-field-rejects_dit_email_marketing" class="c-form-group">
+        <legend class="c-form-group__label u-visually-hidden">
+            <span class="c-form-group__label-text">
+                Marketing preferences
+            </span>
+        </legend>
+        <div class="c-form-group__inner">
+            <div class="c-multiple-choice">
+                <input class="c-multiple-choice__input" type="checkbox" name="rejects_dit_email_marketing" value="rejects_dit_email_marketing" id="field-rejects_dit_email_marketing-1" {% if formData.rejects_dit_email_marketing %} checked {% endif %}>
+                    <label class="c-multiple-choice__label" for="field-rejects_dit_email_marketing-1">
+                        <span class="c-multiple-choice__label-text">
+                            Does not accept email marketing
+                        </span>
+                    </label>
+            </div>
+         </div>
+    </fieldset>
+
 
     {{ MultipleChoiceField({
       name: 'address_same_as_company',
--- a/src/lib/property-helpers.js
+++ b/src/lib/property-helpers.js
@@ -70,11 +70,13 @@ function convertYesNoToBoolean (object) {
   for (const key of keys) {
     const value = convertedObject[key]
 
-    if (value.toLocaleLowerCase() === 'yes') {
-      convertedObject[key] = true
-    }
-    if (value.toLocaleLowerCase() === 'no') {
-      convertedObject[key] = false
+    if (typeof value !== 'boolean') {
+      if (value.toLocaleLowerCase() === 'yes') {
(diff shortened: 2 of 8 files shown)


src/apps/contacts/views/edit.njk
```diff
@@ -83,19 +83,24 @@
...
-    }) }}
+    <fieldset id="group-field-rejects_dit_email_marketing" class="c-form-group">
+        <legend class="c-form-group__label u-visually-hidden">
+            <span class="c-form-group__label-text">
+                Marketing preferences
+            </span>
+        </legend>
+        <div class="c-form-group__inner">
+            <div class="c-multiple-choice">
+                <input class="c-multiple-choice__input" type="checkbox" name="rejects_dit_email_marketing" value="rejects_dit_email_marketing" id="field-rejects_dit_email_marketing-1" {% if formData.rejects_dit_email_marketing %} checked {% endif %}>
```
jim68000: Is there another style I can use here? Feels wrong to reuse the multiple choice one

jim68000: It's not a multiple choice anymore...

yeahfro:
Hi Jim,

Unless I am missing something, the following macro does work:

```
    {{ MultipleChoiceField({
      type: 'checkbox',
      name: 'marketing_preferences',
      label: 'Marketing preferences',
      isLabelHidden: true,
      value: formData.marketing_preferences,
      options: [
        {
          label: 'Does not accept email marketing',
          value: 'rejects_dit_email_marketing'
        }
      ]
    }) }}
```

The benefit of using this is it simplifies the changes to the transformation code, as follows:

```
marketing_preferences: contact.accepts_dit_email_marketing ? null : 'rejects_dit_email_marketing',
```

And saving will be:

```
accepts_dit_email_marketing: contactForm.marketing_preferences !== 'rejects_dit_email_marketing'
```

jim68000:
Yeah, I ran into problems because I got rid of the `marketing_preferences` object entirely, so instead of `formData.marketing_preferences.accepts_dit_email_marketing` I'm asking it to look at `formData.accepts_dit_email_marketing` - it seemed like overkill to keep the `marketing_preferences` object around if it only ever had one member, but if you think that's a better way to do it I'm happy to revert.

src/lib/property-helpers.js
```diff
@@ -70,11 +70,13 @@ function convertYesNoToBoolean (object) {
   for (const key of keys) {
     const value = convertedObject[key]
 
-    if (value.toLocaleLowerCase() === 'yes') {
-      convertedObject[key] = true
-    }
-    if (value.toLocaleLowerCase() === 'no') {
-      convertedObject[key] = false
+    if (typeof value !== 'boolean') {
+      if (value.toLocaleLowerCase() === 'yes') {
```
yeahfro: Would it be safe to assume if `value !== 'yes'` would it be `false` hence `convertedObject[key] = value.toLocaleLowerCase() === 'yes'` rather than lines 74-79?

jim68000: That's a good point. Will tidy.

jim68000: Looking at the usage that isn't the case - this function gets all sorts of inputs, most of which can't be turned into booleans. There's something of smell coming from this function, but I can't quite identify what it is...

PR #1542, opened 2018-09-04
Build download button, based on filter logic.

stuaxo:
Work in progress - do not merge !

First cut -

Creates a form consisting of a download button, hidden fields populated by the current filters.


stuaxo:
Moving PR

PR #1543, opened 2018-09-04
Propositions Documents Feature

yozzo:
For Propositions there is a user need to have a documents upload feature. We're using the `documents` module recently created,
and we're instantiating it in the `router.sub-app` section of Propositions by calling the proprietary `getDocumentsOptions` which contains the string params for the API urls, and then calling the Document's  `postUpload` function

Complete Proposition Journey

Because now we're using documents upload we no longer need the work-around view with a text input for a sharepoint link,
so we've removed it.

### DIFF
--- a/src/apps/documents/middleware/upload.js
+++ b/src/apps/documents/middleware/upload.js
@@ -15,21 +15,23 @@ function parseForm (req, res) {
       return res.status(500).json({ error: err })
     }
 
-    files.forEach(async (file, value, collection) => {
+    Object.keys(files).forEach(async (key, index, collection) => {
       try {
         await chainUploadSequence(req.session.token, {
-          file,
           fields,
+          file: files[key],
           url: res.locals.documents.url,
         })
       } catch (e) {
         req.flash('error', e.message)
         res.redirect(req.originalUrl)
       }
-    })
 
-    req.flash('success', `${filter(files).length} File(s) uploaded`)
-    res.redirect(res.locals.returnLink)
+      if (collection.length === index + 1) {
+        req.flash('success', `${filter(files).length} File(s) uploaded`)
+        res.redirect(res.locals.returnLink)
+      }
+    })
   })
 }
 
--- a/src/apps/propositions/controllers/upload.js
+++ b/src/apps/propositions/controllers/upload.js
@@ -1,0 +1,43 @@
+/* eslint camelcase: 0 */
+const { assign, get } = require('lodash')
+const { notFound } = require('../../../middleware/errors')
+
+const { buildFormWithStateAndErrors } = require('../../builders')
+const { uploadForm } = require('../macros')
+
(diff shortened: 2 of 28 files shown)


src/apps/documents/middleware/upload.js
```diff
@@ -15,21 +15,23 @@ function parseForm (req, res) {
...
-          file,
+          file: files[key],
           fields,
           url: res.locals.documents.url,
         })
       } catch (e) {
         req.flash('error', e.message)
         res.redirect(req.originalUrl)
+      } finally {
+        if (collection.length === index + 1) {
```
yeahfro: There is no need to put this inside `finally` as it will attempt to fire even if there is an error. This is ideal if you want to dispose of resources but in this instance it is not required.

src/apps/documents/middleware/upload.js
```diff
@@ -15,21 +15,23 @@ function parseForm (req, res) {
       return res.status(500).json({ error: err })
     }
 
-    files.forEach(async (file, value, collection) => {
+    Object.keys(files).forEach(async (key, index, collection) => {
       try {
         await chainUploadSequence(req.session.token, {
-          file,
+          file: files[key],
           fields,
```
yeahfro: Move this up one line

src/apps/documents/middleware/upload.js
```diff
@@ -15,21 +15,23 @@ function parseForm (req, res) {
       return res.status(500).json({ error: err })
     }
 
-    files.forEach(async (file, value, collection) => {
+    Object.keys(files).forEach(async (key, index, collection) => {
```
yeahfro: In the past we have agreed to favour Lodash over native. Let's continue with that.

src/apps/propositions/controllers/upload.js
```diff
@@ -0,0 +1,48 @@
...
+  const investment = get(res.locals, 'investmentData.id')
+  const selectUploadForm = buildFormWithStateAndErrors(uploadForm(
+    assign({}, res.locals.options, res.locals.conditions, {
+      returnLink: res.locals.returnLink,
+      hiddenFields: {
+        investment,
+        proposition,
+      },
+    })),
+  get(res.locals, 'form.errors.messages'),
```
yeahfro: Spaces don't look right here

src/apps/propositions/controllers/upload.js
```diff
@@ -0,0 +1,48 @@
...
+
+  res
+    .breadcrumb('Choose files')
+    .title('Choose files')
+    .render('propositions/views/upload', {
+      selectUploadForm,
+    })
+}
+
+function getDocumentsOptions (req, res, next) {
```
yeahfro: This should be in a middleware, see `setInteractionsDetails`. Note `set`

PR #1544, opened 2018-09-04
Investment Evidence Feature

yozzo:
The FDI Rulebook requires a section called `Evidence` where users can upload supporting documents for an investment project.
The Evidence feature itself is only a tags dropdown a file upload and a comments textarea. For documents upload we're reusing the `Documents` module,
and intsantiating it investment-projects/router by calling `postUpload` and binding the api configs necessary for building the request urls.

As Evidence is only required for Investment projects, we've created it inside an `apps` folder within the investment-projects app. This is where we will also need move the `propositions` app as it's not shared across data-hub.


src/apps/investment-projects/apps/evidence/controllers/create.js
```diff
@@ -0,0 +1,35 @@
+/* eslint camelcase: 0 */
+const { assign, get } = require('lodash')
+const { getOptions } = require('../../../../../lib/options')
```
yeahfro: This needs to move down

src/apps/investment-projects/apps/evidence/controllers/create.js
```diff
@@ -0,0 +1,35 @@
...
+    assign({}, res.locals.options, res.locals.conditions, {
+      returnLink: res.locals.returnLink,
+      returnText: 'Cancel',
+      buttonText: 'Upload',
+      tags,
+      hiddenFields: {
+        investment,
+      },
+    })),
+  get(res.locals, 'form.errors.messages'),
```
yeahfro: Funny spacing

PR #1545, opened 2018-09-06
Hide feature flagged investment local nav items

yeahfro:
Recent document upload developments need to be hidden by default until they are ready to be enabled. This will allow us to deploy to production.


yeahfro:
We decided against flagging the feature at this stage of the development.

PR #1546, opened 2018-09-07
CSV downloads for Investment projects.

stuaxo:
WIP - do not merge

A form consisting of a download button, hidden fields populated by the current filters.

- [x] Form that posts to the endpoint with hidden fields populated by filters
- [x] Make a successful download
- [x] Enable streaming from the django API
- [x] Gate functionality behind permission
- [x] Test new functionality
... could always use more? - stu
- [x] Look and feel

### DIFF
--- a/assets/stylesheets/components/_collection.scss
+++ b/assets/stylesheets/components/_collection.scss
@@ -37,6 +37,12 @@
   line-height: 1;
 }
 
+// export action
+
+.c-collection__export-action {
+  float: right;
+}
+
 // Filter summary
 
 .c-collection__filter-summary {
--- a/src/apps/builders.js
+++ b/src/apps/builders.js
@@ -147,6 +147,14 @@ async function buildFieldsWithSelectedEntities (token, fields = [], query = {}) 
   return processedFields
 }
 
+async function buildFiltersValues (filtersFieldsWithSelectedOptions) {
+  const filtersValues = filtersFieldsWithSelectedOptions
+    .filter(filter => filter.value !== undefined)
+    .map(filter => { return [ filter.name, filter.value ] })
+
+  return filtersValues
+}
+
 module.exports = {
   getDeepObjectValuesForKey,
   assignPropsIfFoundInObject,
@@ -155,4 +163,5 @@ module.exports = {
   buildFormWithStateAndErrors,
   buildSelectedFiltersSummary,
   buildFieldsWithSelectedEntities,
+  buildFiltersValues,
 }
--- a/src/apps/investment-projects/controllers/list.js
+++ b/src/apps/investment-projects/controllers/list.js
@@ -9,6 +9,32 @@ const QUERY_STRING = FILTER_CONSTANTS.INVESTMENT_PROJECTS.SECTOR.PRIMARY.QUERY_S
 const SECTOR = FILTER_CONSTANTS.INVESTMENT_PROJECTS.SECTOR.NAME
 const MAX_EXPORT_ITEMS = FILTER_CONSTANTS.INVESTMENT_PROJECTS.SECTOR.MAX_EXPORT_ITEMS
 
+function hasExportPermission (userPermissions) {
+  return userPermissions.includes('investment.export_investmentproject')
+}
+
+function buildExportAction (userPermissions, exportFiltersFields, queryString) {
+  if (hasExportPermission(userPermissions)) {
+    return {
+      messages: {
+        tooManyItems: 'Filter to ' + MAX_EXPORT_ITEMS + ' items to download',
+        default: 'You can now download these records',
+      },
+      maxItems: MAX_EXPORT_ITEMS,
+      enable: true,
+      form: {
+        action: 'investment-projects/export',
+        hiddenFields: exportFiltersFields,
+        query: queryString,
+      },
+    }
+  } else {
+    return {
+      enable: false,
+    }
+  }
+}
+
 async function renderInvestmentList (req, res, next) {
   try {
     const token = req.session.token
@@ -33,23 +59,7 @@ async function renderInvestmentList (req, res, next) {
     const exportFiltersFields = await buildFiltersValues(filtersFieldsWithSelectedOptions)
 
     const userPermissions = get(res, 'locals.user.permissions')
-    const hasExportPermission = userPermissions.includes('investment.export_investmentproject')
-
-    const exportAction = hasExportPermission ? {
-      messages: {
-        tooManyItems: 'Filter to ' + MAX_EXPORT_ITEMS + ' items to download',
-        default: 'You can now download these records',
-      },
-      maxItems: MAX_EXPORT_ITEMS,
-      enable: true,
-      form: {
-        action: 'investment-projects/export',
-        hiddenFields: exportFiltersFields,
-        query: queryString,
-      },
-    } : {
-      enable: false,
-    }
+    const exportAction = await buildExportAction(userPermissions, exportFiltersFields, queryString)
 
     res.render('_layouts/collection', {
       sortForm,
--- a/src/apps/investment-projects/controllers/list.js
+++ b/src/apps/investment-projects/controllers/list.js
@@ -1,12 +1,13 @@
 const { get, merge, omit } = require('lodash')
 
-const { buildSelectedFiltersSummary, buildFieldsWithSelectedEntities } = require('../../builders')
+const { buildSelectedFiltersSummary, buildFieldsWithSelectedEntities, buildFiltersValues } = require('../../builders')
 const { getOptions } = require('../../../lib/options')
 const { investmentFiltersFields, investmentSortForm } = require('../macros')
 
 const FILTER_CONSTANTS = require('../../../lib/filter-constants')
 const QUERY_STRING = FILTER_CONSTANTS.INVESTMENT_PROJECTS.SECTOR.PRIMARY.QUERY_STRING
 const SECTOR = FILTER_CONSTANTS.INVESTMENT_PROJECTS.SECTOR.NAME
+const MAX_EXPORT_ITEMS = FILTER_CONSTANTS.INVESTMENT_PROJECTS.SECTOR.MAX_EXPORT_ITEMS
 
 async function renderInvestmentList (req, res, next) {
   try {
@@ -29,10 +30,31 @@ async function renderInvestmentList (req, res, next) {
 
     const filtersFieldsWithSelectedOptions = await buildFieldsWithSelectedEntities(token, filtersFields, req.query)
     const selectedFilters = await buildSelectedFiltersSummary(filtersFieldsWithSelectedOptions, req.query)
+    const exportFiltersFields = await buildFiltersValues(filtersFieldsWithSelectedOptions)
+
+    const userPermissions = get(res, 'locals.user.permissions')
+    const hasExportPermission = userPermissions.includes('investment.export_investmentproject')
+
+    const exportAction = hasExportPermission ? {
+      messages: {
+        tooManyItems: 'Filter to ' + MAX_EXPORT_ITEMS + ' items to download',
+        default: 'You can now download these records',
+      },
+      maxItems: MAX_EXPORT_ITEMS,
+      enable: true,
+      form: {
+        action: 'investment-projects/export',
+        hiddenFields: exportFiltersFields,
+        query: queryString,
+      },
+    } : {
+      enable: false,
+    }
 
     res.render('_layouts/collection', {
       sortForm,
       selectedFilters,
+      exportAction,
       filtersFields: filtersFieldsWithSelectedOptions,
       title: 'Investment Projects',
       countLabel: 'project',
--- a/src/lib/authorised-raw-request.js
+++ b/src/lib/authorised-raw-request.js
@@ -1,0 +1,53 @@
+const { isNil, pickBy, set } = require('lodash')
+const request = require('request')
+
+const config = require('../../config')
+const logger = require('../../config/logger')
+
+function hasValue (value) {
+  return !isNil(value)
+}
+
+function isString (s) {
+  return typeof (s) === 'string' || s instanceof String
+}
+
+// Accepts either options in a kashmap or a string with a url
+// Combines the options or url with the given token to create a
+// call to the API server
+// The response is not parsed
+module.exports = (token, opts) => {
+  let requestOptions = (isString(opts))
+    ? {
+      json: true,
+      method: 'GET',
+      url: opts,
+    }
+    : {
+      body: opts.body,
+      headers: pickBy(opts.headers, hasValue), // remove empty headers
+      json: true,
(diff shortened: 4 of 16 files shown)


src/apps/builders.js
```diff
@@ -147,6 +157,23 @@ async function buildFieldsWithSelectedEntities (token, fields = [], query = {})
   return processedFields
 }
 
+async function buildDownloadFilterFields (token, filtersFieldsWithSelectedOptions) {
+  const processedFields = {}
+
+  for (let key in filtersFieldsWithSelectedOptions) {
```
yozzo: you might want to use lodash [https://lodash.com] `forEach` or `mapKeys` for looping

src/apps/builders.js
```diff
@@ -147,6 +157,23 @@ async function buildFieldsWithSelectedEntities (token, fields = [], query = {})
...
+    const value = filtersFieldsWithSelectedOptions[key]
+    if (value.value === undefined) {
```
yozzo:
```
if (item.value) {
    processedFields[item.name] = item.value
}
```

src/apps/builders.js
```diff
@@ -147,6 +157,23 @@ async function buildFieldsWithSelectedEntities (token, fields = [], query = {})
...
+      continue
+    }
+    processedFields[value.name] = value.value
+  }
+
+  return {
+    token: token,
```
yozzo: in ES6 and above when you have identical key value pairs, i.e. `token: token` you can declare it like so:
```
return {
    token,
    hiddenFields: processedFiels
}
```

src/apps/investment-projects/controllers/list.js
```diff
@@ -29,11 +29,13 @@ async function renderInvestmentList (req, res, next) {
 
     const filtersFieldsWithSelectedOptions = await buildFieldsWithSelectedEntities(token, filtersFields, req.query)
     const selectedFilters = await buildSelectedFiltersSummary(filtersFieldsWithSelectedOptions, req.query)
+    const downloadFiltersFields = await buildDownloadFilterFields(token, filtersFieldsWithSelectedOptions)
 
     res.render('_layouts/collection', {
       sortForm,
       selectedFilters,
       filtersFields: filtersFieldsWithSelectedOptions,
+      downloadFiltersFields: downloadFiltersFields,
```
yozzo: just `downloadFiltersFields` instead of `downloadFiltersFields: downloadFiltersFields`

src/templates/_layouts/collection.njk
```diff
@@ -14,6 +14,15 @@
 
     {% block xhr_content %}
       <article class="column-two-thirds" id="xhr-outlet">
+        {% block collection_download %}
+          {{ CollectionDownload({
+            query: QUERY,
+            downloadFiltersFields: downloadFiltersFields,
```
yozzo:
```
{ 
  //...
  downloadFiltersFields,
  filtersFields
  //...
}
```

src/modules/search/services.js
```diff
@@ -30,6 +30,31 @@ function search ({ token, searchTerm = '', searchEntity, requestBody, isAggregat
...
+  let options = {
+    url: isAggregation ? searchUrl : `${searchUrl}/${searchEntity}/export`,
+    method: 'POST',
+  }
+  requestBody = assign({}, requestBody, {
+    term: searchTerm,
+  })
+
+  if (isAggregation) {
+    options.qs = assign(requestBody, {
```
yozzo:
```
options.qs = {
  ...requestBody,
  entity: earchEntity,
}
```

src/modules/search/services.js
```diff
@@ -30,6 +30,31 @@ function search ({ token, searchTerm = '', searchEntity, requestBody, isAggregat
...
+      entity: searchEntity,
+    })
+  } else {
+    options.body = requestBody
+  }
+
+  return authorisedRequest(token, options)
+    .then(result => {
```
yozzo: `then(result => result)` one line arrow functions don't need `return`

src/modules/search/services.js
```diff
@@ -30,6 +30,31 @@ function search ({ token, searchTerm = '', searchEntity, requestBody, isAggregat
...
+      return result
+    })
+}
+
 module.exports = {
+  exportSearch: exportSearch,
```
yozzo: `exportSearch,` instead of `exportSearch: exportSearch,` (and all other similar instances in this PR)

src/templates/_macros/collection/collection-content.njk
```diff
@@ -2,6 +2,7 @@
 {% from '../entity/entity-list.njk' import EntityList %}
 {% from './collection-header.njk' import CollectionHeader %}
 
+
```
yozzo: remove extra line

src/apps/investment-projects/controllers/list.js
```diff
@@ -29,16 +30,30 @@ async function renderInvestmentList (req, res, next) {
...
     const filtersFieldsWithSelectedOptions = await buildFieldsWithSelectedEntities(token, filtersFields, req.query)
     const selectedFilters = await buildSelectedFiltersSummary(filtersFieldsWithSelectedOptions, req.query)
-    const downloadFiltersFields = await buildFiltersValues(filtersFieldsWithSelectedOptions)
+    const exportFiltersFields = await buildFiltersValues(filtersFieldsWithSelectedOptions)
 
     const userPermissions = get(res, 'locals.user.permissions')
-    const hasExportPermission = userPermissions.includes('interaction.export_interaction')
+    const hasExportPermission = userPermissions.includes('investment.export_investmentproject')
+
+    const exportAction = hasExportPermission ? {
```
yozzo: I think this ternary operator can sit in its own function as a conditional for more readability

assets/stylesheets/components/_collection.scss
```diff
@@ -37,6 +37,12 @@
   line-height: 1;
 }
 
+// export button
+
+.c-collection__export-form {
```
yozzo: comment says `export button` but selector says `export-form`. If it's a form, I'm not sure `float: right` is the best property to use as it could affect its siblings position and also cascade to children. If it's a button, that shouldn't be a problem, but isn't this something that `text-align: right` could solve? I just know that `float` properties are to be avoided if we're not doing any column layout styling.

src/modules/search/services.js
```diff
@@ -30,6 +30,31 @@ function search ({ token, searchTerm = '', searchEntity, requestBody, isAggregat
     })
 }
 
+function exportSearch ({ token, searchTerm = '', searchEntity, requestBody, isAggregation = true }) {
+  const searchUrl = `${config.apiRoot}/v3/search`
+  let options = {
+    url: isAggregation ? searchUrl : `${searchUrl}/${searchEntity}/export`,
+    method: 'POST',
+  }
+  requestBody = assign({}, requestBody, {
```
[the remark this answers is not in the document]
stuaxo: Will do :)

src/apps/builders.js
stuaxo: I'll have another try at this, initially I did try with lodash, but ended up with code that was more obfuscated than this, will have a poke.

Also, noticed this seems to not be the newest code, have just pushed which has changed some names..

src/apps/builders.js
stuaxo: Ah, this was in the older version - but I'm fairly sure the current version could be nicer than 

    ```if (typeof item.value === 'undefined') {``` 

:(

stuaxo: This is much better now :)  turned out it could just return key/values which simplified things too :D

src/apps/builders.js
stuaxo: Pretty sure this is resolved in newer version

src/apps/investment-projects/controllers/list.js
stuaxo: :+1: - was resolved in newer version

assets/stylesheets/components/_collection.scss
```diff
@@ -37,6 +37,12 @@
   line-height: 1;
 }
 
+// export action
+
+.c-collection__export-action {
+  float: right;
+}
+
 // Filter summary
```
stuaxo: @yozzo 

Got a bit stuck trying to work out how to do this without a float.. it works the same as the badges that use a float, there.. I guess I could make the <span>.

src/apps/investment-projects/controllers/list.js
stuaxo: Will do

test/unit/macros/collection/collection-export.test.js
```diff
@@ -0,0 +1,54 @@
+const { getMacros } = require('~/test/unit/macro-helper')
+const entitiesMacros = getMacros('collection')
+
+describe('CollectionExport macro', () => {
+  describe('invalid props', () => {
+    it('should not render if there is no form', () => {
+      const component = entitiesMacros.renderToDom('CollectionExport')
```
yeahfro: We are putting the `entitiesMacros.renderToDom` inside the `beforeEach`

test/unit/macros/collection/collection-export.test.js
```diff
@@ -0,0 +1,54 @@
...
+          'favourite-shape': 'hexagon',
+          'favourite-tea': 'assam',
+        },
+        action: 'example_action',
+        queryString: '?example_query=123',
+      }
+    })
+
+    it('should render collection export component', () => {
+      const component = entitiesMacros.renderToDom('CollectionExport', this.props)
```
yeahfro: We are putting the `entitiesMacros.renderToDom('CollectionExport', this.props)` inside the `beforeEach`

test/unit/macros/collection/collection-export.test.js
```diff
@@ -0,0 +1,54 @@
...
+
+describe('CollectionExport macro', () => {
+  describe('invalid props', () => {
+    it('should not render if there is no form', () => {
+      const component = entitiesMacros.renderToDom('CollectionExport')
+      expect(component).to.be.null
+    })
+  })
+
+  describe('valid props', () => {
```
yeahfro: The `describe` should be a `context`

test/unit/macros/collection/collection-export.test.js
```diff
@@ -0,0 +1,54 @@
+const { getMacros } = require('~/test/unit/macro-helper')
+const entitiesMacros = getMacros('collection')
+
+describe('CollectionExport macro', () => {
+  describe('invalid props', () => {
```
yeahfro: The `describe` should be a `context`

test/unit/macros/collection/collection-export.test.js
```diff
@@ -0,0 +1,54 @@
...
+      expect(renderInputFilter1.name).to.equal('favourite-shape')
+      expect(renderInputFilter1.value).to.equal('hexagon')
+
+      expect(renderInputFilter2).to.exist
+      expect(renderInputFilter2.name).to.equal('favourite-tea')
+      expect(renderInputFilter2.value).to.equal('assam')
+    })
+
+    it('is possible to disable the button', () => {
+      const component = entitiesMacros.renderToDom('CollectionExport', this.props, true)
```
yeahfro: This line will be inside the `beforeEach`

src/apps/builders.js
```diff
@@ -147,6 +147,14 @@ async function buildFieldsWithSelectedEntities (token, fields = [], query = {})
   return processedFields
 }
 
+async function buildFiltersValues (filtersFieldsWithSelectedOptions) {
+  const filtersValues = filtersFieldsWithSelectedOptions
```
yeahfro: We are favouring Lodash over native

yeahfro: It might read better to just use `map` and then `compact` for when there is not a `filter.value`

src/lib/authorised-raw-request.js
```diff
@@ -0,0 +1,53 @@
+const { isNil, pickBy, set } = require('lodash')
```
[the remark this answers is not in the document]
stuaxo: Yeah it's meant to be.  Basically provides the same API, but using an actual request.

I should probably factor out the shared functions now you mention it though.

src/apps/investment-projects/controllers/list.js
```diff
@@ -1,12 +1,39 @@
...
+      },
+      maxItems: MAX_EXPORT_ITEMS,
+      enable: true,
+      form: {
+        action: 'investment-projects/export',
+        hiddenFields: exportFiltersFields,
+        query: queryString,
+      },
+    }
+  } else {
```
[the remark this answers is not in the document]
stuaxo: Hm .. maybe I should just use the ternary operator here.

src/apps/builders.js
stuaxo: > We are favouring Lodash over native

Is there a list of these preferences somewhere?   I'm don't normally touch JS much, so am going to make quite a few of these n00b type errors.

test/unit/macros/collection/collection-export.test.js
stuaxo: What is the rule to decide what goes in beforeEach ?

src/modules/search/services.js
```diff
@@ -30,6 +31,37 @@ function search ({ token, searchTerm = '', searchEntity, requestBody, isAggregat
     })
 }
 
+function exportSearch ({ res, token, searchTerm = '', searchEntity, requestBody, isAggregation = true }) {
```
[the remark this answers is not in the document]
stuaxo: Though I guess I still don't know what a lot of this actually does :(

test/unit/macros/collection/collection-export.test.js
stuaxo: Heh, just saw where I got describe inside a describe from ... https://mochajs.org/#synchronous-code

src/apps/investment-projects/controllers/list.js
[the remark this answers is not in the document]
stuaxo: OK, inverted this, prefer to have early bailout when it's a simple condition, and then the bigger code after.

yeahfro:
Morning @stuaxo,

The main point was that this can be drastically simplified by it being a `GET` rather than a `POST`. This will be less code and therefore negate the need for the more fiddly bits of code you have implemented.

If you want me to pair or at the very least have a face-to-face conversation on this then let's do that.

Cheers

src/apps/builders.js
stuaxo: > Morning @stuaxo,
> 
> The main point was that this can be drastically simplified by it being a `GET` rather than a `POST`. This will be less code and therefore negate the need for the more fiddly bits of code you have implemented.
> 
> If you want me to pair or at the very least have a face-to-face conversation on this then let's do that.
> 
> Cheers

Sounds good, it actually is a GET (wasn't always)... so totally makes sense to rip out all this stuff, definitely up for pairing, can come into the office after lunch if you're about.

stuaxo:
Supeseded by PR #1556

PR #1549, opened 2018-09-14
Client-side fix for the name attribute suffix issue

yozzo:
Because form ecription type `multipart/form-data` required for document upload doesn't take any input attributes other than name,
to be able to submit multiple documents, we need a unique `name` for each input. Initially we reused the same functionality as for creating multiple inputs, but this was generating unique `id`'s for each input.
With this fix we added some logic that checks for the form type and changes the attribute that needs the unique suffix, while not breaking the functionality of other components sharing this script.

### DIFF
--- a/assets/javascripts/lib/helpers.js
+++ b/assets/javascripts/lib/helpers.js
@@ -163,23 +163,21 @@ function resetFieldValues (element) {
  *
  * @returns {nodeElement}
  */
-function regenIds (wrapper) {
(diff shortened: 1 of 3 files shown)


marcofucci:
It would be better to have the commit message in present-tense and rebase develop to avoid the merge commit if you can

marcofucci:
@yozzo for next time, the commit message is still _Fixed_ plus it'd better to wait 2 approvals before merging ;-)

PR #1556, opened 2018-09-27
Feature/investment csv download link

stuaxo:
Users with permission on the investments projects page can download a CSV of results if there are 5000 or less results.

This PR supersedes #1546 

Notes on review of the previous PR:
 - [x] Make download a simple link
 - [x] Change wording to show amount of records that will be downloaded
 ~- [ ] Clean up authorised-request and simplify authorised-raw-request
Currently having a look at this.~
 This has been moved to it's own branch
- [x] Tests

**Screenshot before**
image

**Screenshot after**
image

PR #1557, opened 2018-09-28
Feature/refactor authorised request

stuaxo:
This separates out the refactoring of authorised_request / authorised_request_raw from PR #1556 

- [x] Fix two remaining broken tests
- [x] requestOptions:  break option parsing out into function. 
  ~- [ ] Look at whats going on with .qs~ 
not entirely sure what this was about...
  - [x] Move everything to immutable structures.
  ~- [ ] Move from custom hasValue to something lodash based, e.g. isPresent~
This is a bit annoying, using isPresent causes all sorts of failures in the tests


stuaxo:
> 
> src/lib/authorised-raw-request.js
> // call to the API server
> // The response is not parsed
> module.exports = (token, opts) => {
>   let requestOptions = (isString(opts))
> @yozzo
> yozzo 4 hours ago Member
> 
> this variable declaration will have to sit in its own dedicated function with normal conditionals, rather than a multi-line ternary operator. Also we tend to use const for most of the time, instead of let
>

stuaxo:
Taken WIP tag off, but probably needs a bit of thought around hasValue vs isPresent.

isPresent isn't exactly the same has hasValue, and using isPresent causes a bunch of tests to fail.

src/lib/authorised-request.js
```diff
@@ -18,93 +18,67 @@ function stripScript (text) {
...
 
-function parseOptions (opts) {
+function headerOptions (opts, token) {
+  return {
+    ...opts,
+    ...(Boolean(token)) && { 'Authorization': `Bearer ${token}` },
+  }
+}
+
+function parseOptions (opts, token, jsonReviver = false) {
```
yeahfro: I like the ideas here but maybe some of it could be simplified. I have not tested this, but something like:

```function parseOptions (opts, token, jsonReviver = false) {
  const defaults = {
    jsonReviver,
    json: true,
    proxy: process.env.PROXY,
    headers: {
      ...headerOptions(opts.headers, token)
    },
    method: 'GET',
  }

  if (isString(opts)) {
    return {
      ...defaults,
      url: opts,
    }
  }
  return {
    ...defaults,
    body: opts.body,
    method: opts.method || 'GET',
    qs: pickBy(opts.qs, hasValue),
    url: opts.url,
  }
}```

File: src/lib/authorised-request.js [hunk 2]
```diff
@@ -18,93 +18,67 @@ function stripScript (text) {
...
-
-  if (token) {
-    set(requestOptions, 'headers.Authorization', `Bearer ${token}`)
+  if (process.env.PROXY && config.isDev) {
+    process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0'
   }
 
   logger.debug('Send authorised raw request: ', requestOptions)
-
   return Promise.resolve(request(requestOptions))
```
yeahfro: Why can't we use `request-promise` here?

src/lib/authorised-request.js
```diff
@@ -18,93 +18,67 @@ function stripScript (text) {
...
-    }
-    : {
-      body: opts.body,
-      headers: pickBy(opts.headers, hasValue),
-      json: true,
-      method: opts.method || 'GET',
-      qs: pickBy(opts.qs, hasValue),
-      url: opts.url,
-    }
+  const requestOptions = parseOptions(opts, token, stripScripts)
```
yeahfro: I would say these two functions can be refactored further as it's only the options that are changing

src/lib/authorised-request.js
stuaxo: Splitting out defaults is a big improvement in readability, +1.

src/lib/authorised-request.js
stuaxo: And how they are returned.

I know you were pro creating something that takes different kinds of request, if you do, then the main thing is not to expose that as an API on the outside:
 - There should never be a time where you need to switch between them at runtime.
 - Raw request, by design can't do any checks on the returned data (the jsonReviver)

src/lib/authorised-request.js
stuaxo: request-promise doesn't work well with pipe - somewhere on their github is a ticket for it.
The conversation starts saying they will implement it, but by the end, the recommendation is not to use it... 

It looks like something they may work out in future though -

https://github.com/request/request-promise/issues/90

^ I should probably have linked to this ticket in the code.

OT:

Underneath request all this, request streaming should be using something like the sendfile API in linux, pretty much handing off implementation from CPU, which is great... however:

It would be great to know that this is in fact happening, from what I've read under some circumstances request can end up using the CPU.

It would be great if it was possible to write a test to find out what's being used here, but I'd need to work with JS tech for longer to have an idea how this could work.

src/lib/authorised-request.js
stuaxo: I guess it comes down to two two line functions:

```
raw:
    call options with no json reviver
    return request(...) wrapped in promise

promise
    call options with json reviver
    return request-promise(...)
```

They look superficially similar, but adding an abstraction here costs:
 - 1 . extra level of indirection
 - zero (but probably) or more lines.

It's important that nobody accidentally replaces request with request-promise (I actually did this a couple of times through copy and paste earlier in development, and wasted an hour or two wondering why tests were not passing).

stuaxo: I missed out this thing:

  if (process.env.PROXY && config.isDev) {
    process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0'
  }

I'm not sure whether it's worth pushing that into a function or not, afaict it's ugly but necessary :)

PR #1559, opened 2018-10-01
Add multi step form builder with address module skeleton

yeahfro:
https://trello.com/c/h3I90SsV/234-add-address-front-end-module

# Description

Before tackling the work to introduce an address module, it is essential to work on implementing multiple step forms. This has been a source of our over complicated code.

All forms follow the same pattern:
1. Present a form
2. POST/PATCH to server
3. If there are errors then present errors, otherwise present success message

This is the first in a series of pull requests to introduce multiple step forms. A basic skeleton implementation of the address module has been included to show how it will work.

address-module

# Run local

To run this locally you could add create address functionality to `companies` by adding the following to `router.js`:

```
const addressesRouter = require('../addresses/router.sub-app')
...
router.use('/:companyId', addressesRouter)
```

# Outstanding

Key pieces that are still outstanding:
- client-side validation (partial validation is required for each step)
- POST/PATCH to API
- handle API errors
- persist state across multiple steps
- further development of address module

### DIFF
--- a/src/apps/addresses/apps/create/steps.js
+++ b/src/apps/addresses/apps/create/steps.js
@@ -1,0 +1,58 @@
+const {
+  overseasAddressForm,
+  postcodeLookupForm,
+  selectUkAddressForm,
+  ukOrOverseasForm,
+} = require('../../macros')
+
+module.exports = {
+  steps: [
+    {
+      path: '/create/step-1',
+      type: 'form',
+      heading: 'Is this a UK or overseas address?',
+      breadcrumbs: [
+        { name: 'Add an address' },
+      ],
+      macro: ukOrOverseasForm,
+      nextPath: ({ uk_or_overseas: ukOrOverseas }) => {
+        const path = {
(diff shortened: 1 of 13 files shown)


src/apps/addresses/apps/create/steps.js
```diff
@@ -0,0 +1,58 @@
...
+  steps: [
+    {
+      path: '/create/step-1',
+      type: 'form',
+      heading: 'Is this a UK or overseas address?',
+      breadcrumbs: [
+        { name: 'Add an address' },
+      ],
+      macro: ukOrOverseasForm,
+      nextPath: ({ uk_or_overseas: ukOrOverseas }) => {
```
jim68000: Why the switch from camel to snake case? Is this one of our idiosyncrasies I’m not aware of?

yozzo: where I encountered snake case was when referencing the (python) API request and response keys, I'm guessing this is a similar situation.

PR #1567, opened 2018-10-15
add csv download button to companies collection

jim68000:
Straight copy of Stu's work into Companies


yeahfro:
Hi Jim,

The changes look fine. I expect you are expecting my feedback to be:
- no screenshot on PR
- are we going to copy the code for all entities? (nice to be in this position now)
- no unit tests (understood there are none for list now)

jim68000:
WIP while I add a test to the company acceptance flow

src/apps/companies/router.js
```diff
@@ -65,6 +65,12 @@ router.get('/',
   renderCompanyList,
 )
 
+router.get('/export',
+  setDefaultQuery(DEFAULT_COLLECTION_QUERY),
+  getRequestBody(QUERY_FIELDS),
+  exportCollection('company', 'Data Hub - Companies'),
```
reupen: I see it was already like that, but I’m a bit confused about the filename stuff (`'Data Hub - Companies'`) – it seems to pass the filename and content type from the back end through, so not sure why it's setting them as well (in `streamToFile()`).

jim68000: Lemme see if I can figure that out

jim68000: OK, so it appears Express would normally like to receive a filename so it can figure out the content disposition header, however that's not necessary here as that's generated by the back end - along with the actual filename. As this requires refactoring the download code for investment properties as well I think I'd rather get this merged in and then raise that as a separate PR.

stuaxo: As far as I can tell the may be a browser dependant thing happening wrt filename, but would require more testing to work out.
 content disposition should be set by the code that sets the type.
I found it a bit confusing that the downstream API wanted "txt" for the type instead of a mimetype, but that's just how it seems to work.

PR #1568, opened 2018-10-16
refactor csv export functions into library

jim68000:
This moves the common functions from Stu's CSV export work into a library as preparation for adding export to other entities. As Lee likes screenshots...

image

### DIFF
--- a/src/lib/export-helper.js
+++ b/src/lib/export-helper.js
@@ -1,0 +1,40 @@
+const { pluralise } = require('../../config/nunjucks/filters')
+
+function hasExportPermission (userPermissions, targetPermission) {
+  return userPermissions.includes(targetPermission)
+}
+
+function invalidNumberOfItems (resultCount, maxItems) {
+  return resultCount === 0 || resultCount >= maxItems
+}
+
+function buildExportMessage (resultCount, exportOptions) {
+  if (resultCount >= exportOptions.maxItems) {
+    return `Filter to fewer than ${exportOptions.maxItems} ${pluralise(exportOptions.entityName)} to download`
+  }
+  if (resultCount === 1) {
+    return `You can now download this ${exportOptions.entityName}`
+  }
+  if (resultCount === 0) {
+    return `There are no ${pluralise(exportOptions.entityName)} to download`
+  }
+  return `You can now download these ${resultCount} ${pluralise(exportOptions.entityName)}`
+}
+
+function buildExportAction (queryString, userPermissions, exportOptions) {
+  if (!hasExportPermission(userPermissions, exportOptions.targetPermission)) {
+    return {
+      enabled: false,
+    }
+  }
+  return {
+    enabled: true,
+    buildMessage: (count) => buildExportMessage(count, exportOptions),
+    url: `${exportOptions.urlFragment}/export?${queryString}`,
+    invalidNumberOfItems,
+  }
+}
+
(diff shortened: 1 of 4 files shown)


src/lib/export-helper.js
```diff
@@ -0,0 +1,32 @@
...
+  return userPermissions.includes(targetPermission)
+}
+
+function tooManyItems (resultCount, maxItems) {
+  return resultCount >= maxItems
+}
+
+function buildExportMessage (resultCount, maxItems) {
+  if (tooManyItems(resultCount, maxItems)) {
+    return `Filter to less than ${maxItems} projects to download`
```
reupen: Shouldn't the correct noun (projects) be passed in as well?

jim68000: Good spot, well fix

src/lib/export-helper.js
```diff
@@ -0,0 +1,32 @@
+function hasExportPermission (userPermissions, targetPermission) {
+  return userPermissions.includes(targetPermission)
+}
+
+function tooManyItems (resultCount, maxItems) {
+  return resultCount >= maxItems
```
reupen: It was > rather than >= before, > looked right to me?

jim68000: Might need to change the wording then. As it says you need to have fewer than 5000 items, so 5000 is actually not acceptable. Also, I've just noticed it says 'less than 5000 items' which is grammatically incorrect as well. I will try and grab Anthony for clarification.

jim68000:
I’m going to WIP this as I’ve realised I need to deal with multiple possible values for the message rather than two

src/lib/export-helper.js
```diff
@@ -0,0 +1,32 @@
...
+function buildExportAction (queryString, userPermissions, exportOptions) {
+  if (!hasExportPermission(userPermissions, exportOptions.targetPermission)) {
+    return {
+      enabled: false,
+    }
+  }
+  return {
+    enabled: true,
+    buildMessage: (count) => buildExportMessage(count, exportOptions.maxItems),
+    url: exportOptions.urlFragment + '/export?' + queryString,
```
yeahfro: Here we could use string interpolation

src/lib/export-helper.js
```diff
@@ -1,16 +1,25 @@
...
-function tooManyItems (resultCount, maxItems) {
-  return resultCount >= maxItems
+function invalidNumberOfItems (resultCount, maxItems) {
+  return resultCount === 0 || resultCount >= maxItems
 }
 
 function buildExportMessage (resultCount, exportOptions) {
-  if (tooManyItems(resultCount, exportOptions.maxItems)) {
-    return `Filter to less than ${exportOptions.maxItems} ${exportOptions.entityPlural} to download`
+  const entityPlural = get(exportOptions, 'entityPlural', exportOptions.entityName + 's')
```
yeahfro: Looks like this `entityPlural` will never be set and the `+ 's'` will not work for `company`. We use `pluralise`.

jim68000: Actually, yes, I remembered pluralise on the way home last night.... the unset var here is me thinking ahead to the company/companies issue, but I will refactor to use pluralise instead

src/lib/export-helper.js
jim68000: WIll change

PR #1573, opened 2018-10-19
Add middleware to post details to API

yeahfro:
https://trello.com/c/h3I90SsV/234-add-address-front-end-module

This is a critical middleware component, hence exhaustive unit testing.

Middleware checks partial validation and then checks if there is a send function. If there is a send function then it is assumed this is the final step of the journey.

If the send function is successful then the user is redirected to the next path.

If the send fails then the errors are presented to the user.

### DIFF
--- a/src/modules/form/middleware.js
+++ b/src/modules/form/middleware.js
@@ -1,0 +1,42 @@
+const { isEmpty, set } = require('lodash')
+
+const { getNextPath } = require('./helpers')
+const state = require('./state/current')
+const { getErrors } = require('./errors')
+
+const postDetails = async (req, res, next) => {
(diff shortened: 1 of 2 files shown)


src/modules/form/middleware.js
```diff
@@ -0,0 +1,42 @@
...
+    state.update(req.session, key, currentStep.path, { completed: false })
+    return next()
+  }
+
+  if (currentStep.send) {
+    const data = state.reduceSteps(req.session, key)
+    await currentStep.send(data, (err) => {
+      if (err) {
+        state.update(req.session, key, currentStep.path, { completed: false })
+        if (err.statusCode === 400) {
```
jim68000: Only 400?

PR #1575, opened 2018-10-19
remove unused params from stream to file

jim68000:
The back end actually provides the file name and mime type for CSV downloads, so these parameters can be removed from the library, along with their tests


src/lib/stream-to-file.js
```diff
@@ -1,7 +1,6 @@
 // req must be request not request-promise
-function streamToFile (req, res, filename, type = 'txt') {
-  res.attachment(filename)
-  res.type(type)
+function streamToFile (req, res) {
+  res.attachment()
```
reupen: Is this line needed (`res.attachment()`)?

jim68000: I knew you'd only ask that if you knew the answer ;-)

No, no it is not

reupen:
I think `streamToFile()` can be got rid of now (in favour of calling `req.pipe()` directly in `exportCollection()`).

jim68000:
Pfft. Consider it done.

src/modules/search/middleware/collection.js
```diff
@@ -20,19 +19,15 @@ function getCollection (searchEntity, entityDetails, ...itemTransformers) {
...
-
-function exportCollection (searchEntity, filenamePrefix) {
+function exportCollection (searchEntity) {
   return async function (req, res, next) {
     try {
       exportSearch({
         searchEntity,
         requestBody: req.body,
         token: req.session.token,
       }).then(req => {
```
reupen: I realise these don't relate to the specific changes being made, but a couple of comments and questions:

- it seems a bit unfortunate to reuse `req` here (additionally, isn't it a response?)
- is it not possible to use `await` with exportSearch() (and therefore a single catch handler)?

test/unit/modules/search/middleware/collection-export.test.js
```diff
@@ -3,44 +3,40 @@ const config = require('~/config')
...
 
         this.reqMock.query = {
           sortby: 'name:asc',
         }
-        await this.exportCollection('entity', 'test-file')(this.reqMock, this.resMock, this.nextSpy).catch(e => { throw e })
-      })
-
-      it('should set the attachment filename', () => {
-        expect(this.streamToFileStub.args[0][2]).to.equal('test-file.csv')
+        await this.exportCollection('entity')(this.reqMock, this.resMock, this.nextSpy).catch(e => { throw e })
```
yeahfro: Is this `catch` required?

src/modules/search/middleware/collection.js
stuaxo: This should probably have a different name, but is a request object at this point.

Since this is setting up the .pipe, a response may not have been received at this point, the idea being we hand off (er.. handling?) using the pipe stuff.

At some point I used ```_req``` to show it was different not the same ```req``` object passed to the function.

Maybe it should be rawReq - since it's an ordinary request (not the request promise we use elsewhere) to show that something special is happening here.

test/unit/modules/search/middleware/collection-export.test.js
stuaxo: Good catch, it appears not to be.

stuaxo:
Hey @jim68000 following our chat earlier, I'm going to go ahead and do the patches suggested by the reviewers above

src/modules/search/middleware/collection.js
stuaxo: @reupen fixed naming (and verified this is indeed a request).

re: using a single await, do you mean something like this:

```javascript
    await exportSearch({
      searchEntity,
      requestBody: req.body,
      token: req.session.token,
    }).then(apiReq => {
      return apiReq.pipe(res)
    }).catch(error => {
      return next(error)
    })

```

?

test/unit/modules/search/middleware/collection-export.test.js
stuaxo: Fixed.

src/modules/search/middleware/collection.js
stuaxo: Implemented as above.

reupen: I was thinking something more like:

```javascript
function exportCollection (searchEntity) {
  return async function (req, res, next) {
    try {
      await apiReq = exportSearch({
        searchEntity,
        requestBody: req.body,
        token: req.session.token,
      })
      
      return apiReq.pipe(res)
    } catch (error) {
      next(error)
    }
  }
}
```

stuaxo:
Soo.. .can I merge this ?

stuaxo:
image

PR #1576, opened 2018-10-19
Progress bar stage (e.g. node on the bar) colour change

paulgain:
At present when a particular stage is active the background colour is
white despite the previous completed stages being blue. We can improve
the UX by changing the active state to blue which indicates where you
are in the process more clearly.

https://uktrade.atlassian.net/browse/IPBETA-8

### DIFF
--- a/assets/stylesheets/components/_progress.scss
+++ b/assets/stylesheets/components/_progress.scss
@@ -1,6 +1,11 @@
 @import "../settings";
 @import "../tools";
 
+@mixin progress-bar-node {
+  background-color: $govuk-blue;
+  border: 2px solid $govuk-blue;
+}
+
 $_section-width: 25%;
 
 .c-progress {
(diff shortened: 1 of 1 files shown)


assets/stylesheets/components/_progress.scss
```diff
@@ -67,7 +67,7 @@ $_section-width: 25%;
 
   &.is-active {
     &::before {
-      background-color: $white;
```
reupen: Would it make sense to share the `::before` styles with `.is-complete` (rather than repeating them)?

paulgain: Good spot.

https://github.com/uktrade/data-hub-frontend/blob/feature/progress-bar-styling/assets/stylesheets/components/_progress.scss#L4

paulgain: Good spot.

https://github.com/uktrade/data-hub-frontend/blob/feature/progress-bar-styling/assets/stylesheets/components/_progress.scss#L4

paulgain: Good spot.

paulgain: Good spot.

paulgain: Good spot.

PR #1577, opened 2018-10-22
CSV export for contacts.

stuaxo:
Based off @jim68000's company CSV export, so everything is pretty streamlined by now

Before company export feature
image

After export company feature
image

### DIFF
--- a/test/acceptance/features/companies/step_definitions/collection.js
+++ b/test/acceptance/features/companies/step_definitions/collection.js
@@ -85,7 +85,7 @@ Then(/^the companies should be filtered by company name/, async function () {
     .assert.containsText('@header', expected)
 })
 
-Then(/^the CSV download button should be visible/, async function () {
+Then(/^the companies CSV download button should be visible/, async function () {
   await CompanyList
     .section.collectionHeader
     .waitForElementVisible('@download')
--- a/test/acceptance/features/contacts/step_definitions/collection.js
+++ b/test/acceptance/features/contacts/step_definitions/collection.js
@@ -190,6 +190,13 @@ Then(/^the contacts should be filtered by ([A-Za-z]+) ([A-Za-z]+)$/, async funct
     .assert.containsText(selector, expected)
 })
 
+Then(/^the contacts CSV download button should be visible/, async function () {
+  await ContactList
+    .section.collectionHeader
+    .waitForElementVisible('@download')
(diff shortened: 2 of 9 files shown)


stuaxo:
> Looks like the acceptance test failed...

Doh ... #hubris

stuaxo:
Will these tests ever finish .. ?

Makes me think, you could have a really evil pattern where each run of the CI takes 2x the amount of time ...

test/acceptance/features/contacts/step_definitions/collection.js
```diff
@@ -190,6 +190,13 @@ Then(/^the contacts should be filtered by ([A-Za-z]+) ([A-Za-z]+)$/, async funct
     .assert.containsText(selector, expected)
 })
 
+Then(/^the CSV download button should be visible/, async function () {
```
reupen: If the one for companies has 'companies' in it (`'the companies CSV download button should be visible'`) then shouldn't we put 'contacts' in this one?

test/acceptance/features/companies/step_definitions/collection.js
```diff
@@ -85,7 +85,7 @@ Then(/^the companies should be filtered by company name/, async function () {
     .assert.containsText('@header', expected)
 })
 
-Then(/^the CSV download button should be visible/, async function () {
+Then(/^the companies CSV download button should be visible/, async function () {
```
yeahfro: Yeah, let's make this a reusable step as @reupen suggests

test/acceptance/features/contacts/step_definitions/collection.js
yeahfro: Or maybe the entity type is irrelevant

jim68000: What is being used as the fixture here?

stuaxo: When I missed this out, I got an error because the description was the same as the companies test.

test/acceptance/features/companies/step_definitions/collection.js
stuaxo: This sounds good, what is involved?

jim68000: There's a standard reusable test in the step definitions for all features - I've changed the test name so this gets picked up

PR #1579, opened 2018-10-23
Remove obsolete feature flag checks

paulgain:
Currently we are checking for 2 feature flags:

 * Investment evidence
 * Proposition documents

Both of these features are now live so we can safely remove both of
these checks.

https://uktrade.atlassian.net/browse/IPBETA-50


paulgain:
Before this PR can be merged we need to remove any feature flags from the DB @marcuspp.

marcuspp:
@paulgain I've tested this from the point of view of the backend and this is safe to be merged without removing the flags from the database. Once this is merged and released the flags will then be removed from the db.

I've removed the `Do not merge` label.

PR #1590, opened 2018-10-25
Use event fixtures for unreliable tests

yeahfro:
Because there is a delay between page load and ES update, it meant the newly created event was sometimes not visible.

Creating the event was unnecessary as we are not testing event creation. We have suitable event fixtures.

Note there are other opportunities to apply this pattern.


test/acceptance/features/events/collection.feature
```diff
@@ -11,15 +11,12 @@ Feature: View a list of events
     And I click the "Add event" link
     Then I am taken to the "Add event" page
 
-  @events-collection--view
+  @events-collection--view--without-uk-region
```
reupen: Any reason this tag has `--view` in the middle, but the tag for the next test doesn't?

yeahfro: Nice fine - will update

PR #1594, opened 2018-10-25
Update next path when redirecting to the next step

yeahfro:
Maintain next path in state when redirecting to the next step. This is required for invalidating state when next path changes.

### DIFF
--- a/src/modules/form/middleware.js
+++ b/src/modules/form/middleware.js
@@ -45,8 +45,9 @@ const postDetails = async (req, res, next) => {
       res.redirect(req.baseUrl + getNextPath(currentStep, req.body))
     })
   } else {
-    state.update(req.session, key, currentStep.path, { completed: true })
-    res.redirect(req.baseUrl + getNextPath(currentStep, req.body))
+    const nextPath = getNextPath(currentStep, req.body)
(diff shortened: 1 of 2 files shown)


src/modules/form/middleware.js
```diff
@@ -45,8 +45,9 @@ const postDetails = async (req, res, next) => {
       res.redirect(req.baseUrl + getNextPath(currentStep, req.body))
     })
   } else {
-    state.update(req.session, key, currentStep.path, { completed: true })
-    res.redirect(req.baseUrl + getNextPath(currentStep, req.body))
+    const nextPath = getNextPath(currentStep, req.body)
+    state.update(req.session, key, currentStep.path, { completed: true, nextPath })
+    res.redirect(req.baseUrl + nextPath)
```
jim68000: Do we have a standard pattern on where slashes go in the path? I see in the test that the assumption is that the nextPath var starts with a slash - do we protect against this being wrong anywhere, or document our expectations?

yeahfro: It is on my radar - thinking about something similar to the [trimming slashes solution](https://github.com/uktrade/data-hub-frontend/blob/fffb8a23558f32781cdc90511cc5dd903ddc4f0d/src/modules/form/helpers.js#L4).

PR #1596, opened 2018-10-29
Remove hard coded timeline data source labels.

ian-leggett:
Previously the company source labels were hard coded. These are now available in the backend and are being pulled through.

timeline

### DIFF
--- a/test/unit/data/companies/timeline.json
+++ b/test/unit/data/companies/timeline.json
@@ -5,26 +5,31 @@
   "results": [
     {
       "data_source": "companies_house.companies",
+      "data_source_label": "Companies House (Companies)",
       "datetime": "2018-10-31T00:00:00Z",
       "description": "Accounts next due date"
     },
     {
       "data_source": "companies_house.companies",
+      "data_source_label": "Companies House (Companies)",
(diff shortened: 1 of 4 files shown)


test/unit/data/companies/timeline.json
```diff
@@ -4,27 +4,27 @@
   "previous": null,
   "results": [
     {
-      "data_source": "companies_house.companies",
```
elcct: The backend also sends `data_source` field, so I would keep that in the test.

ian-leggett: Just adding this back to the JSON.

yeahfro:
@Ian, the commit messages should read like an instruction, ie "Remove", "Update", "Replace"

Also, as we are going through the PR process we use fixups and then once approved, the fixups are squashed

ian-leggett:
@yeahfro just updated the commit comments to "Remove", "Update", "Replace", I was trying to --fixup one of the commits for squashing but it will merge into the commit for the test update rather than the one it relates too.

yeahfro:
I am on the cusp of signing this off but the commits don't really make sense as you would apply the first and the test would fail until the second is applied.

PR #1598, opened 2018-10-30
update config for PaaS London

nusnewob:

### DIFF
--- a/config/index.js
+++ b/config/index.js
@@ -4,6 +4,8 @@ const isDev = process.env.NODE_ENV !== 'production'
 const isProd = process.env.NODE_ENV === 'production'
 const root = path.normalize(`${__dirname}/..`)
 
+var vcap_services = JSON.parse(process.env.VCAP_SERVICES)
+
 const config = {
   root,
   buildDir: path.join(root, '.build'),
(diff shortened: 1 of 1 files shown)


config/index.js
```diff
@@ -4,6 +4,8 @@ const isDev = process.env.NODE_ENV !== 'production'
 const isProd = process.env.NODE_ENV === 'production'
 const root = path.normalize(`${__dirname}/..`)
 
+var vcap = process.env.VCAP_SERVICES ? JSON.parse(process.env.VCAP_SERVICES) : JSON.parse('{"redis": [{}]}')
```
yeahfro: @nusnewob 

Rather than setting this `var` and then doing the ternary if, how about we have a function:

```javascript
const buildRedisConfig = () => {
  const metadataTtl = (process.env.METADATA_TTL || (15 * 60))
  if (process.env.VCAP_SERVICES) {
    const vcap = JSON.parse(process.env.VCAP_SERVICES)

    return {
      metadataTtl,
      url: vcap.redis[0].credentials.uri,
      port: vcap.redis[0].credentials.port,
      host: vcap.redis[0].credentials.host,
      useTLS: vcap.redis[0].credentials.tls_enabled,
    }
  }

  return {
    metadataTtl,
    url: process.env.REDIS_URL || process.env.REDISTOGO_URL,
    port: process.env.REDIS_PORT || 6379,
    host: process.env.REDIS_HOST || 'redis',
    useTLS: process.env.REDIS_USE_TLS,
  }
}

and then the `redis` field changes to `buildRedisConfig()`

File: config/index.js [hunk 2]
```diff
@@ -21,11 +23,11 @@ const config = {
     baseUrl: 'https://api.getAddress.io/v2/uk/{postcode}?api-key={api-key}',
   },
   redis: {
```
yeahfro: This would be `redis: buildRedisConfig()`

config/index.js
nusnewob: Yes, I agree, the config is becoming more unreadable, I only wanted to make minimal changes, I'm happy what you proposed.

nusnewob: @yeahfro updated the branch with your function.

PR #1599, opened 2018-10-30
Bug - Business type changes when trying to add a foreign company without completing mandatory fields

ian-leggett:
## Problem

When creating a new Foreign Organisation, (type Company)

1. If user does not complete a mandatory field and attempts to save, the error message appears, but editable Foreign Org form changes to editable UK Company creation form.

2. Form keeps completed data, but if user attempts to change business type to foreign, then data is lost and user has to start again

## Solution
I have now added the "business_type" and the "country" params to the post action on the form as without these the form redirects to "**Add UK company**" with errors rather than "**Add foreign company**" with errors.

### DIFF
--- a/src/templates/_macros/form/form.njk
+++ b/src/templates/_macros/form/form.njk
@@ -29,13 +29,17 @@
   {% set returnText = props.returnText or 'Back' %}
   {% set hideFormActions = props.hideFormActions or false %}
   {% set children = caller() if caller else renderAsMacro(props.children) %}
-
   <form
       method="{{ method }}"
       {% if props.enctype %}
         enctype="{{ props.enctype }}"
       {% endif %}
-      action="{{ props.action }}?_csrf={{csrfToken}}"
+      action="{{ props.action }}?_csrf={{csrfToken}}
+      {%- if props.actionExtension -%}
+        {%- for key, value in  props.actionExtension -%}
+           &{{ key }}={{ value }}
+        {%- endfor -%}
+      {%- endif -%}"
       {% if props.class %}class="{{ props.class }}"{% endif %}
       {% if props.role %}role="{{ props.role }}"{% endif %}
   >
--- a/src/apps/companies/views/edit.njk
+++ b/src/apps/companies/views/edit.njk
@@ -29,6 +29,9 @@
     buttonText: 'Save and return' if company.id else 'Add company',
     returnLink: '/companies/' + company.id if company.id else '/companies/add-step-1',
     returnText: 'Return without saving' if company.id else 'Back',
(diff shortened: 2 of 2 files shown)


jim68000:
Think you've got a git problem - you appear to have picked up one of Lee's commits

ian-leggett:
Just removed Lee's commit that got added in the original PR and updated commit message.

src/templates/_macros/form/form.njk
```diff
@@ -29,13 +29,23 @@
   {% set returnText = props.returnText or 'Back' %}
   {% set hideFormActions = props.hideFormActions or false %}
   {% set children = caller() if caller else renderAsMacro(props.children) %}
-
+  {% if props.hiddenFields.uk_based === 'no' %}
+    {% set ukBased = 'non-uk' %}
```
yeahfro: This template cannot know about specific fields.

I would say that the best option to fix this is with routing.

ian-leggett: Do you mean the values 'no' and 'uk_based'? Should these values be passed in via a route '/companies/add'.?

ian-leggett: Ah I see what you mean, specific like props.hiddenFields.business_type

ian-leggett: @yeahfro I have removed the specific prop passed to the view and added a new prop to the edit.njk view. Any thoughts?

src/templates/_macros/form/form.njk
```diff
@@ -29,13 +29,15 @@
...
   {% set children = caller() if caller else renderAsMacro(props.children) %}
-
   <form
       method="{{ method }}"
       {% if props.enctype %}
         enctype="{{ props.enctype }}"
       {% endif %}
-      action="{{ props.action }}?_csrf={{csrfToken}}"
+      action="{{ props.action }}?_csrf={{csrfToken}}
+      {%- if props.isForeign -%}
```
jim68000: This still feels like the wrong place to be putting business logic specific to one flow. Form.njk is supposed to be a neutral renderer of forms that shouldn’t know specifics. I think it would be more correct to create a props.actionExtension in form.njk, and then get the company edit to pass both key and value through because that allows the form macro to be ignorant if the company flow, and is potentially reusable.  That make sense? Happy to talk it through

ian-leggett: @jim68000 I see what you mean. I have now removed the business logic out of the form macro. You can now pass down key value pairs in an object called `actionExtension` which lives in the edit.njk so if this object exists the form macro will loop through and add the key value pairs as params for the form action.

PR #1602, opened 2018-11-01
Filter investment project stages

paulgain:
Filter investment project stage names
    
    At present the progress bar has the following stages:
    
    * Prospect
    * Assign PM
    * Active
    * Verify win
    * Won
    
    This piece of work allows us to filter out any given stage name by
    setting the exclude_from_investment_flow property to true, data coming
    from the API endpoint /metadata/investment-project-stage.
    
    The implementation is hidden behind an exclude_from_investment_flow
    feature flag.
    
    https://uktrade.atlassian.net/browse/IPBETA-73

### DIFF
--- a/src/apps/investment-projects/middleware/shared.js
+++ b/src/apps/investment-projects/middleware/shared.js
@@ -25,6 +25,13 @@ function getCompanyDetails (req, res, next) {
     .catch(next)
 }
 
+function getInvestmentProjectStages (features) {
+  if (features && features['streamlined-investment-flow']) {
+    return metadata.investmentProjectStage.filter(stage => stage.exclude_from_investment_flow !== true)
+  }
+  return metadata.investmentProjectStage
+}
+
 async function getInvestmentDetails (req, res, next) {
   const investmentId = req.params.investmentId
 
@@ -37,7 +44,7 @@ async function getInvestmentDetails (req, res, next) {
     const ukCompanyId = get(investmentData, 'uk_company.id')
     const clientRelationshipManagerId = get(investmentData, 'client_relationship_manager.id')
     const stageName = investmentData.stage.name
-    const investmentProjectStages = metadata.investmentProjectStage
+    const investmentProjectStages = getInvestmentProjectStages(res.locals.features)
 
     investmentData.investor_company = Object.assign({}, investmentData.investor_company, investorCompany)
 
--- a/test/unit/apps/investment-projects/middleware/shared.js
+++ b/test/unit/apps/investment-projects/middleware/shared.js
@@ -1,6 +1,7 @@
-const { merge } = require('lodash')
+const { merge, cloneDeep } = require('lodash')
 
 const investmentData = require('~/test/unit/data/investment/investment-data.json')
+const investmentProjectStages = require('~/test/unit/data/investment/investment-project-stages.json')
 
 const companyData = {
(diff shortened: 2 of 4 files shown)


paulgain:
This piece of work requires some backend work from @marcuspp which is currently being reviewed. Feel free to remove the 'Do Not Merge' label when it is sensible to do so.

src/apps/investment-projects/middleware/shared.js
```diff
@@ -25,6 +25,14 @@ function getCompanyDetails (req, res, next) {
     .catch(next)
 }
 
+function getInvestmentProjectStages (features) {
```
yeahfro: There are a couple of options which might be cleaner here:

Basically avoiding the mutation of `stages`

```javascript
function getInvestmentProjectStages (features) {
  if (features && features['streamlined-investment-flow']) {
    return metadata.investmentProjectStage.filter(stage => stage.exclude_from_investment_flow !== true)
  }
  return metadata.investmentProjectStage
}

function getInvestmentProjectStages (features) {
  return compact(map(metadata.investmentProjectStage, (stage) => {
    if (features && features['streamlined-investment-flow'] && stage.exclude_from_investment_flow) {
      return stage
    }
  }))
}

File: test/unit/apps/investment-projects/middleware/shared.js [hunk 2]
```diff
@@ -274,5 +248,34 @@ describe('Investment shared middleware', () => {
         expect(this.nextSpy).to.have.been.calledWith(this.error)
       })
     })
+
+    context('when the streamlined-investment-flow feature flag is set to true', () => {
+      beforeEach(async () => {
+        const stages = cloneDeep(investmentProjectStages)
```
yeahfro: 👍

src/apps/investment-projects/middleware/shared.js
paulgain: Yeah, the first option is one line cleaner :-) 

I don't believe the `metadata.investmentProjectStage` is being mutated due to how the filter function works on Array. Taking a look at the docs: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter it says:

**"filter() does not mutate the array on which it is called."**

paulgain: I also favour `const` over `let`. In this context `let` was more appropriate as I was reassigning an array reference which you cannot do when using `const`. This is now a moot point as I prefer your first option where there are neither of them present :-)

paulgain: One last point, the stage list coming back from the API:

 https://github.com/uktrade/data-hub-frontend/blob/feature/filter-investment-project-stages/test/unit/data/investment/investment-project-stages.json

I think the `null` values are going, the `exclude_from_investment_flow` properties we be set to either `true` or `false` based on @marcuspp feedback - so this maybe open to change hence the `Do Not Merge` flag being present.

marcuspp: The backend change to `exclude_from_investment_flow` so that it only returns `true` or `false` has just been completed, It will be released on thursday. I will remove the `Do Not Merge` when the release has gone out.

marcuspp:
@yeahfro @paulgain This can be merged now. Will this automatically get deployed to staging so we can test turning the feature flag on or is that a manual process?

PR #1615, opened 2018-11-07
Front end for Interactions CSV export

ian-leggett:
## Problem
As an investment project team member, I want to download a subset of interactions from the Interactions section of Data Hub so that I can monitor my / my team's performance.

## Solution
* Add a button to enable a user to download the interaction list

* The button should only be enabled once filters have been applied, such that there are fewer than 5,000 results

* The button should not be shown at all to LEP and DA users

screen shot 2018-11-07 at 09 50 46

### DIFF
--- a/test/unit/apps/interactions/controllers/list.test.js
+++ b/test/unit/apps/interactions/controllers/list.test.js
@@ -13,6 +13,9 @@ describe('interaction list', () => {
           permissions: [],
         },
       },
+      query: {
+        sortby: 'date:desc',
+      },
     }
 
     this.res = {
--- a/src/apps/interactions/constants.js
+++ b/src/apps/interactions/constants.js
@@ -20,7 +20,7 @@ const QUERY_FIELDS = [
   'date_before',
   'sortby',
   'dit_team',
-  'service'
+  'service',
 ]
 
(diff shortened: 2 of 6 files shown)


test/unit/apps/interactions/controllers/list.test.js
```diff
@@ -85,6 +89,42 @@ describe('interaction list', () => {
...
+
+      it('should return a function that can check if items have been filtered enough to export', () => {
+        expect(this.exportAction.invalidNumberOfItems).to.be.a('function')
+      })
+
+      it('should return a url to download the csv', () => {
+        expect(this.exportAction.url).to.equal('interactions/export?sortby=date%3Adesc')
+      })
+
+      it('should output a message if further filtering is required', () => {
```
yeahfro: These three tests I would each put inside a `context`.

e.g. `context('when only one item is downloadable', () => {})`

yeahfro: Hmm.. having given it more thought I would say that we need more unit tests around `buildExportAction` which would then negate the need for these tests here. We basically don't need to test the message for every entity type. We can have one set of tests for `buildExportAction` and then each entity would just test the presence of the export action.

PR #1627, opened 2018-11-22
Update typeahead component to search teams

ian-leggett:
## Problem

Trello - https://trello.com/c/HUCNAUIX/397-team-filter-component-fe

The team search (service providers) is currently a select list of over a thousand entries which means searching a team is difficult and time consuming. Scrolling a list of this size is currently not a pleasant experience.

## Solution

As the users needs to scroll through so many entries in the current filter component it makes sense to use the existing typeahead Vue component and add a method that uses a fuzzy search which will enable the user to find the team faster by typing rather then scrolling.  We already use a component for searching advisers (Vue Typeahead/multiselect). 

Vue multiselect - https://vue-multiselect.js.org/ (component used for typeahead)

For this use case we don’t have the same API endpoints that allow you to return individual teams as we do advisers so we will filter the data that is already loaded into the browser (as it currently does). As this search behaviour is slightly different due to the way we expose the data I now detect inside the Vue instance wether we are passing a model to it, if we do, it uses a method ‘find’ which searches data that is already loaded, if we don’t pass a model then the original ‘asyncFind’ method fires and it calls out to the API. 

screen shot 2018-11-22 at 11 38 49

### DIFF
--- a/assets/javascripts/lib/helpers.js
+++ b/assets/javascripts/lib/helpers.js
@@ -223,13 +223,14 @@ function closest (element, selector) {
 
 function matchWords (str, words) {
   const queryWords = words.split(' ')
-  let matched = 0
-  queryWords.forEach((word) => {
+  const count = queryWords.reduce((allWords, word) => {
     if (str.search(new RegExp(word, 'i')) !== -1) {
-      matched += 1
+      allWords++
     }
-  })
-  return queryWords.length === matched
+    return allWords
+  }, 0)
+
+  return queryWords.length === count
 }
 
 module.exports = {
--- a/assets/javascripts/lib/helpers.js
+++ b/assets/javascripts/lib/helpers.js
@@ -207,6 +207,31 @@ function closest (element, selector) {
   return null
 }
 
+/**
+ * matchWords
+ *
+ * breaks your search query into an array containing a word/character or words/characters
+ * loops through your array and increments a count whenever it gets a match
+ * returns true when the count value matches the array length
+ * Useful when doing a fuzzy search on a string
+ *
+ * @param {string} str is the data to search on
+ * @param {string} words is your query
+ *
+ * @returns {boolean}
+ */
+
+function matchWords (str, words) {
+  const queryWords = words.split(' ')
+  let matched = 0
+  queryWords.forEach((word) => {
+    if (str.search(new RegExp(word, 'i')) !== -1) {
+      matched += 1
+    }
+  })
+  return queryWords.length === matched
+}
+
 module.exports = {
   addClass,
   removeClass,
@@ -223,4 +248,5 @@ module.exports = {
   regenIds,
   resetFieldValues,
   closest,
+  matchWords,
 }
--- a/assets/javascripts/vue/filters.js
+++ b/assets/javascripts/vue/filters.js
@@ -1,9 +1,17 @@
-function highlight (words, query) {
-  if (!query) { return words }
-  const iQuery = new RegExp(query, 'ig')
-  return words.toString().replace(iQuery, (matchedTxt, a, b) => {
-    return ('<span class=\'highlight\'>' + matchedTxt + '</span>')
+function highlight (str, words) {
+  if (!words) { return str }
+
+  const queryWords = words.split(' ').filter((word) => word.length >= 1)
+  const openTag = '<span class=\'highlight\'>'
+  const closeTag = '</span>'
+
+  queryWords.forEach((word) => {
+    const iQuery = new RegExp(`(${word})(?![^<]+?>)`, 'ig')
+    str = str.replace(iQuery, (matchedTxt) => {
+      return (openTag + matchedTxt + closeTag)
+    })
   })
+  return str
 }
 
 module.exports = {
--- a/assets/javascripts/vue/typeahead.vue
+++ b/assets/javascripts/vue/typeahead.vue
@@ -1,14 +1,16 @@
 <template>
-  <div class="c-form-group c-form-group--light c-form-group--smaller c-form-group--filter">
+  <div class="c-form-group c-form-group--light c-form-group--smaller c-form-group--filter"
+       v-bind:id="name+'__typeahead'">
     <label class="c-form-group__label" :for="id">
       <span class="c-form-group__label-text">{{ label }}</span>
     </label>
     <multiselect
       label="label"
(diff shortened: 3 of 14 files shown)


assets/javascripts/lib/helpers.js
```diff
@@ -207,6 +207,31 @@ function closest (element, selector) {
...
+ * @param {string} str is the data to search on
+ * @param {string} words is your query
+ *
+ * @returns {boolean}
+ */
+
+function matchWords (str, words) {
+  const queryWords = words.split(' ')
+  let matched = 0
+  queryWords.forEach((word) => {
```
yeahfro: It feels like this could be tidied up with `reduce`

assets/javascripts/vue/filters.js
```diff
@@ -1,9 +1,21 @@
...
+
+  const queryWords = words.split(' ').filter((word) => word.length >= 1)
+  const start = '~'
+  const end = '§'
+  const openTag = '<span class=\'highlight\'>'
+  const closeTag = '</span>'
+
+  queryWords.forEach((word) => {
+    const iQuery = new RegExp(word, 'ig')
+    str = str.replace(iQuery, (matchedTxt) => {
```
yeahfro: Surely this can be simplified by doing the `replace` once rather than replacing with special characters and then replacing again on line 17

assets/javascripts/vue/typeahead.vue
```diff
@@ -1,51 +1,52 @@
 <template>
-  <div class="c-form-group c-form-group--light c-form-group--smaller c-form-group--filter">
+  <div class="c-form-group c-form-group--light c-form-group--smaller c-form-group--filter" v-bind:id="name+'__typeahead'">
```
yeahfro: All of these indentation changes should really be on the previous commit as these changes do not reflect the commit message

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "data-hub-frontend",
-  "version": "4.43.0",
+  "version": "4.42.0",
```
yeahfro: ??

test/acceptance/features/interactions/collection.feature
```diff
@@ -61,9 +61,9 @@ Feature: View collection of interactions
     Then I see the success message
     When I navigate to the `interactions.list` page
     Then I confirm I am on the Interactions page
-    And the results summary for a interaction collection is present
-    Then I filter the interactions list by service provider
-    Then the interactions should be filtered by service provider
+#    And the results summary for a interaction collection is present
```
yeahfro: Rather than commenting out, can you add `@ignore` with a comment?

test/acceptance/features/interactions/step_definitions/collection.js
```diff
@@ -1,13 +1,15 @@
+// Todo The acceptance tests with the typeahead Vue component needs looking into as setting values with nightwatch in the search field does not work
```
yeahfro: Move this comment to where the `@ignore` will be

test/vue/typeahead.test.js
```diff
@@ -206,6 +210,101 @@ describe('Typeahead', () => {
         })
       })
     })
+
+    describe('#find', () => {
+      let textInput
+      let listItem
```
yeahfro: Rather than declaring this here, each test can `const listItem ...`

assets/javascripts/vue/typeahead.vue
```diff
@@ -83,56 +97,67 @@
...
+      return
+    }
+    this.isLoading = true
+    axios.get(`/api/options/${this.entity}?term=${query}`)
+      .then((response) => {
+        this.options = response.data
+        this.isLoading = false
+      })
+      .catch(function (error) {
+        console.error(error)
```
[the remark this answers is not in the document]
ian-leggett: I don't think we do as its set to false when the component is mounted.

package.json
ian-leggett: So I think this happened when I accidentally merged develop into my branch. This should be on version 4.42.0.

ian-leggett: Thinking about this now, should I bump this to latest version?

assets/javascripts/vue/filters.js
ian-leggett: Yeah this does feel a bit nasty I admit. The problem I have here is that once one highlight has been done HTML then gets searchable which then gets replaced and you have nested span's. It goes a bit crazy, must be an easier way to do this. Graham helped me with this and did mention adding him to the PR to see if he had any better ideas.

package.json
yeahfro: You shouldn't need to change this in this PR

assets/javascripts/vue/typeahead.vue
```diff
@@ -37,19 +41,21 @@
...
+           :key="index" :value="option.value">
   </div>
 </template>
-
 <script>
+
   const axios = require('axios')
   const Multiselect = require('vue-multiselect').default
   const getFormData = require('get-form-data').default
+  const { debounce } = require('lodash')
```
yeahfro: This will be better as `const debounce = require('lodash/debounce')`

package.json
yeahfro: This change needs removing

assets/javascripts/vue/typeahead.vue
```diff
@@ -129,8 +134,26 @@
           })
       }, 500),
     },
+    watch: {
+      selectedOptions: function (newOptions) {
+        if (!this.autoSubmit || this.isSubmitting) { return }
+        this.isSubmitting = true
```
yeahfro: is this right that `isSubmitting` is set to `true` when on line 143 we could be returning if `form` is falsy?

ian-leggett: 🤔 this is odd, it was there from the previous implementation and to be honest I don't think it does anything. Removed it and all is ok, `isSubmitting` isn't used anywhere

assets/javascripts/vue/filters.js
web-bert: @ian-leggett are there any tests for this to cover the expected cases of the new RegEx? I have run some locally and simplified it to `new RegExp( '(' + word + ')(?![^<]+?>)', 'ig' )` what are the other cases that this implementation is covering? At the very least I think it should be lazy matching?

PR #1629, opened 2018-11-22
Check One List group tier for presenting section

yeahfro:
https://trello.com/c/p0V2zAPm/414-subsidiary-of-one-list-company-only-show-global-account-manager-section-if-a-company-is-on-the-one-list

## Description
Subsidiaries of parent companies that are on the One List need to display the `Global Account Manager - One List` section.

## Before
screenshot 2018-11-23 at 11 27 00

## After
screenshot 2018-11-23 at 11 25 34


yeahfro:
😆

PR #1638, opened 2018-12-04
Change HTML date fields to use type=['date'] rather than text

ian-leggett:
## Problem
The format of the date expected by the user needs changing as we currently ask for `YYYY-MM-DD`, we now need to ask for `DD/MM/YYYY` which is a format thats more recognised by users in the UK.
We also would like to remove the "text" (type="text") input type and use the "date" (type="date") so that users using modern browsers and smartphones will be able to use the native date pickers from their browsers. As IE does not support this value (type="date") we need to fall back to a text input field.
screen shot 2018-12-04 at 12 09 11

## Solution
Change all the input date types to use (type="date"), detect the user agent on the server side and if Internet Explorer reverse the date format for the user so they can type in the correct format of `DD/MM/YYYY` then reverse this again for the API request as the API expects `YYYY-MM-DD`
screen shot 2018-12-04 at 12 09 47

### DIFF
--- a/config/nunjucks/filters.js
+++ b/config/nunjucks/filters.js
@@ -80,6 +80,11 @@ const filters = {
   pluralise,
   sentenceCase: Case.sentence,
 
+  reverseDate (dateString) {
+    if (isNil(dateString)) { return }
+    return dateString.split('/').reverse().join('/')
+  },
+
   encodeQueryString (value) {
     return encodeURIComponent(queryString.stringify(value))
   },
--- a/src/apps/events/controllers/list.js
+++ b/src/apps/events/controllers/list.js
@@ -7,7 +7,7 @@ async function renderEventList (req, res, next) {
   try {
     const token = req.session.token
     const advisers = await getAdvisers(req.session.token)
-    const filtersFields = eventFiltersFields({ advisers: advisers.results })
+    const filtersFields = eventFiltersFields({ advisers: advisers.results, userAgent: res.locals.userAgent })
     const filtersFieldsWithSelectedOptions = await buildFieldsWithSelectedEntities(token, filtersFields, req.query)
     const selectedFilters = await buildSelectedFiltersSummary(filtersFieldsWithSelectedOptions, req.query)
 
--- a/src/apps/events/macros/event-filters-fields.js
+++ b/src/apps/events/macros/event-filters-fields.js
@@ -1,10 +1,9 @@
 const { assign, flatten } = require('lodash')
 const { globalFields } = require('../../macros')
 const { collectionFilterLabels } = require('../labels')
-const currentYear = (new Date()).getFullYear()
 const { transformObjectToOption } = require('../../transformers')
 
-const eventFiltersFields = ({ advisers }) => {
+const eventFiltersFields = ({ advisers, userAgent }) => {
   return [
     {
       macroName: 'TextField',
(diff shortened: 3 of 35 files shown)


config/nunjucks/filters.js
```diff
@@ -80,6 +80,11 @@ const filters = {
   pluralise,
   sentenceCase: Case.sentence,
 
+  reverseDate (dateString) {
+    if (dateString === undefined) { return }
```
yeahfro: The preferred approach here would be to use `isNil` from Lodash.

src/apps/events/controllers/list.js
```diff
@@ -7,10 +7,12 @@ async function renderEventList (req, res, next) {
   try {
     const token = req.session.token
     const advisers = await getAdvisers(req.session.token)
-    const filtersFields = eventFiltersFields({ advisers: advisers.results })
+    const filtersFields = eventFiltersFields({ advisers: advisers.results, userAgent: res.locals.userAgent })
     const filtersFieldsWithSelectedOptions = await buildFieldsWithSelectedEntities(token, filtersFields, req.query)
     const selectedFilters = await buildSelectedFiltersSummary(filtersFieldsWithSelectedOptions, req.query)
 
+    // console.log(selectedFilters)
```
yeahfro: 🔴

src/apps/events/macros/event-filters-fields.js
```diff
@@ -17,16 +16,18 @@ const eventFiltersFields = ({ advisers }) => {
       options: advisers.map(transformObjectToOption),
     },
     {
-      macroName: 'TextField',
+      macroName: 'DateField',
+      type: 'date',
       name: 'start_date_after',
-      hint: 'YYYY-MM-DD',
-      placeholder: `e.g. ${currentYear}-07-18`,
+      hint: userAgent.isIE ? 'DD/MM/YYYY' : '',
```
yeahfro: Does this have to be an empty string or can it be `null`?

src/apps/events/macros/event-filters-fields.js
```diff
@@ -17,16 +16,18 @@ const eventFiltersFields = ({ advisers }) => {
...
+      placeholder: '',
```
yeahfro: Does this need to be an empty string or does it just not need to be set at all?

src/apps/events/macros/event-filters-fields.js
```diff
@@ -17,16 +16,18 @@ const eventFiltersFields = ({ advisers }) => {
...
     },
     {
-      macroName: 'TextField',
+      macroName: 'DateField',
+      type: 'date',
       name: 'start_date_before',
-      hint: 'YYYY-MM-DD',
-      placeholder: `e.g. ${currentYear}-07-21`,
+      hint: userAgent.isIE ? 'DD/MM/YYYY' : '',
```
yeahfro: See above

src/apps/events/macros/event-filters-fields.js
```diff
@@ -17,16 +16,18 @@ const eventFiltersFields = ({ advisers }) => {
...
+      placeholder: '',
```
yeahfro: See above

src/lib/if-ie-reverse-date-value.js
```diff
@@ -0,0 +1,10 @@
+function reverseDateIfIE (dateRequest, userAgent) {
+  try {
+    if (!userAgent.isIE || (dateRequest === undefined)) { return dateRequest }
```
yeahfro: `isNil` would be better here

yeahfro: In addition I would consider destructuring `userAgent` when it is passed as a parameter

ian-leggett: Do you mean something like `const {isIE} = userAgent` before the "try catch"

test/acceptance/features/step_definitions/list.js
```diff
@@ -51,7 +51,8 @@ Then(/^I can view the (.+) in the collection$/, async function (entityType, data
 
   await Collection
     .section.firstCollectionItem
-    .waitForElementPresent('@header')
+    .waitForElementPresent('@header', 20000)
```
yeahfro: Is this actually required? It seems like if the header is not present after `2000` then it will not be present after `20000`.

PippoRaimondi: there are times it could take more than 2s... if you'd be happier I can drop it to 10s as it most likely won't exceed 10s...

PippoRaimondi: regarding waitForCOntainsText... it was extremely flaky running locally... this additional wait makes it more robust and will help give consistent results when running locally...

yeahfro: `waitForContainsText ` - I am confused as when the element is present then it will contain the text - it's not as if anything is happening asynchronously once the element is present. I have not seen this failing locally.

yeahfro:
👍 nice commits

PR #1641, opened 2018-12-06
Retrofitting the GOV.UK frontend

paulgain:
https://github.com/alphagov/govuk-frontend

* Added the govuk-frontend dependency
* Removed the fast-sass-loader dependency and replaced it with a new
  dependency called sass-loader as it was incompatible with the
  govuk-frontend dependency - it incorrectly rewrote the govuk-frontend
  font Sass URLs when used in conjuction with resolve-url-loader meaning
  the govuk fonts failed to load.
* Merged dit-base.njk into datahub-base.njk and extended the GOV.UK
  frontend [template](https://github.com/alphagov/govuk-frontend/blob/master/src/template.njk)
* Updated the README file illustrating the new nunjucks blocks.

https://uktrade.atlassian.net/browse/IPBETA-158


webpack.config.js
```diff
@@ -68,7 +69,7 @@ const common = {
             },
             'resolve-url-loader',
             {
-              loader: 'fast-sass-loader',
+              loader: 'sass-loader',
```
yozzo: this is fantastic!

paulgain: I don't follow, how is this fantastic?

yeahfro: @paulgain I get the impression you are not the first person to experience pain from this module

paulgain: @yeahfro "pain" is the correct word. The weekly downloads for sass-loader and fast-sass-loader are 2,600,638 and 13,308 respectively - I rest my case.

yozzo: it's fantastic because loading external assets in sass is a known issue, we did find painful work-arounds but haven't thought of looking at the tools we're using, nice one!

PR #1645, opened 2018-12-11
WIP Feature/deprecate companies

yeahfro:

TODO


src/apps/companies/views/_deprecated/details.njk
```diff
@@ -22,7 +22,16 @@
 
   {% if company.id %}
     {% if not company.archived %}
```
retex73: Could this not be written {% if  company.id and not company.archived %} ? 
I also wonder if the {% if company.id %} is really required

src/apps/companies/views/_deprecated/interactions.njk
```diff
@@ -0,0 +1,51 @@
+{% extends "./_layout-view.njk" %}
+
+{% block main_grid_right_column %}
+<h2 class="heading-medium">Company interactions</h2>
+
+{% if company.contacts.length %}
```
retex73: Would {% if company.contacts %} not be enough here?

yeahfro:
@retex73 the views are just direct copies of what is already there. I definitely agree with one of your suggestions but it's beyond the scope of this change. We can get the views tidied up over the next few weeks.

PR #1650, opened 2018-12-13
Feature/likelihood to land

paulgain:
* Filter projects based on their `likelihood to land` by selecting a filter of which there are 3: Low, Medium and High.
    
* Edit a projects `likelihood to land` setting via a dropdown in the detail/edit page, again Low, Medium and High.
    
* Added a nunjucks `json` filter function to enable easy debugging within our views, for example: 
```<pre>{{ data | json }}</pre>```
    
https://uktrade.atlassian.net/browse/IPBETA-156
https://uktrade.atlassian.net/browse/IPBETA-163

### DIFF
--- a/src/apps/companies/controllers/subsidiaries.js
+++ b/src/apps/companies/controllers/subsidiaries.js
@@ -9,8 +9,8 @@ async function renderSubsidiaries (req, res, next) {
     const token = req.session.token
     const query = req.query
     const page = query.page || '1'
-    const { id: companyId, name: companyName, archived: companyArchived, duns_number: DnB } = res.locals.company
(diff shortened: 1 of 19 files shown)

### CODE REVIEW

File: config/nunjucks/filters.js [hunk 1]
```diff
@@ -210,6 +210,10 @@ const filters = {
 
     return `${className} ${classModifier}`.trim()
   },
+
+  json: obj => {
```
yeahfro: Is this required? Would `dump` have done the trick?

paulgain: Yes it would of. I didn't realise you could format dump to 2 spaces. I'll remove my `json` filter. Thanks ;-)

PR #1654, opened 2018-12-13
Present different audit view for dnb company

yeahfro:

If a company does not have a DUNS number then present the _deprecated view

This is similar to #1652 .


test/unit/apps/companies/controllers/audit.test.js
```diff
@@ -61,47 +38,80 @@ describe('Company audit controller', () => {
       })
```
ian-leggett: Should this be empty?

PR #1656, opened 2018-12-13
Present different contacts view for dnb company

yeahfro:

If a company does not have a DUNS number then present the _deprecated view

This is similar to #1652 .


src/apps/companies/controllers/contacts.js
```diff
@@ -16,14 +16,17 @@ function renderContacts (req, res) {
     ],
   })
 
-  const actionButtons = companyArchived ? undefined : [{
+  const actionButtons = company.archived ? undefined : [{
```
jim68000: Nice change

PR #1661, opened 2018-12-18
Add company known as to business details view

yeahfro:
https://trello.com/c/0hu6YBSW/551-implement-full-business-details-page

This change adds the `{companyName} is known as` to a new `Business details` view for the company.

## Included in this change
- Updates to `key-value-table` component so that cell lists can include links
- Add new `Business details`
- Start acceptance tests

## More to come
- Evolve acceptance tests to cover more scenarios
- More `Business details` sections

## One List Corp business details
screenshot 2018-12-18 at 07 15 10

## Companies House number
At this stage there is not an example with a Companies House number, but if there was it would look like:

screenshot 2018-12-18 at 07 19 42

### DIFF
--- a/src/apps/companies/controllers/business-details.js
+++ b/src/apps/companies/controllers/business-details.js
@@ -1,0 +1,20 @@
+/* eslint-disable camelcase */
+const {
+  transformCompanyToKnownAsView,
+} = require('../transformers')
+
+async function renderBusinessDetails (req, res) {
+  const company = res.locals.company
+
+  res
+    .breadcrumb(company.name, `/companies/${company.id}`)
+    .breadcrumb('Business details')
+    .render('companies/views/business-details', {
+      heading: 'Business details',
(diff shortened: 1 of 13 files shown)


paulgain:
FYI: I don't have access to Trello.

Looks good to me. However, when playing with this locally I don't see a link to view the company's `Business details`. Looking at the [markup](https://github.com/uktrade/data-hub-frontend/blob/7f7b77f8f128b4183577dca22ad7f72dd94cf860/src/templates/_components/key-value-table.njk#L52) it appears the link only appears when the `url` is defined which makes sense but I'm unable to view the `Business details` based on the default API data.

yeahfro:
Thank you @paulgain - in the description I wrote `At this stage there is not an example with a Companies House number, but if there was it would look like`.

I will be adding more scenarios to the test fixtures.

Thank you for the review

paulgain:
Sorry @yeahfro that wasn't clear enough for me.

PR #1673, opened 2018-12-20
Add mi database to acceptance tests configuration.

elcct:
This adds to acceptance tests configuration mi postgres database and environment variable needed by Leeloo.


elcct:
The `setup-uat.sh` and other related stuff was not working correctly so I updated it now in my branch on Leeloo.

PR #1683, opened 2019-01-07
Remove support for legacy policy feedback interactions

reupen:
Having policy feedback as a separate kind of interaction was superseded by adding policy feedback to normal interactions and service deliveries. This PR removes code relating to the old form of policy feedback interactions.

This includes removing:

- the 'Capture policy feedback' kind option when creating an interaction
- code related to policy feedback permissions
- the 'Policy feedback' kind filter option
- anything related to the `policy_issue_type` field (this was superseded by `policy_issue_types`)
- anything else related to the 'policy_feedback' interaction kind
- the separate acceptance test run for a policy feedback user

Some tests were updated rather than removed as they still appeared to be useful.


reupen:
The enforced `policy_staff_at` and `master-policy_staff_at` checks will need to be turned off here on GitHub before this can be merged.

PR #1684, opened 2019-01-08
'Trading name' changes to 'Trading names' in FE 

retex73:
https://trello.com/c/fJDmbjyk
https://trello.com/c/ZwkN6ECX
###  Edit company page
screenshot 2019-01-08 at 10 43 16
### Company listing page
screenshot 2019-01-08 at 10 42 43
### Company search page
screenshot 2019-01-08 at 10 42 14


src/apps/companies/transformers/company-to-list-item.js
```diff
@@ -36,7 +36,7 @@ module.exports = function transformCompanyToListItem ({
 
   if (trading_name) {
     meta.push({
-      label: 'Trading name',
+      label: hqLabels.trading_names,
       value: trading_name,
```
yeahfro: This should be `trading_names`?

marcofucci:
@retex73 sorry if it wasn't clear but the edit should still allow editing just one trading_name as it is at the moment so the label should stay the same. I'm going to allow changing trading_names in the API so that you can change this edit form in the frontend as well but for now let's keep it this way.

marcofucci:
@retex73 Also, the view company page seems to still be using the trading_name field, it should be using the trading_names field instead. `trading_name` should only be using as a PATCH data field for now.

marcofucci:
Can you also check this, I get a styling problem in local but it could be just my setup:

screenshot 2019-01-08 at 12 42 37

yeahfro:
I am picking up this change in @retex73's absence.

Note that it is clear that the edit company code is difficult to work with. We are carrying technical debt in this area. Once the ability to add/edit companies is no longer required then we can remove the code.

PR #1685, opened 2019-01-08
Adviser typeahead for Investments and Events

ian-leggett:
## Problem
The typeahead needs to replace select dropdown options in investments and events. We also need a no-js fallback added for the typeahead.

screen shot 2019-01-08 at 13 29 05

## Solution
Update the investments and event forms/filters with the typeahead and offer a `<noscript>` option for  users that have JS switched off.

## Note: testing issue with typeahead
From: Fillipo
We are unable to interact with `.multiselect__input` via webdriver api... mainly due to the way this component is built (the input field is actually hidden)... so unless we refactor the component so it exposes the input field, we can't use selenium to set a value on it, at least I couldn't figure a way to do so.


test/acceptance/features/step_definitions/details.js
```diff
@@ -74,6 +74,9 @@ const assertTableContent = async function (tableSelector, expectedData, tableTyp
       const rowCellSelector = tableType.rowCellSelector(row[rowFirstCellKey], columnIndex)
       const tableRowCellXPathSelector = tableSelector.selector + rowCellSelector.selector
       const expected = getExpected(row[columnKey], this.state)
+      if (row[columnKey] === 'interaction.contact') {
```
yeahfro: Can we add this quirk to `getExpected` next to other quirks?

test/unit/apps/events/controllers/edit.test.js
```diff
@@ -43,6 +43,11 @@ const metadataMock = {
     { id: '2', name: 'r2', disabled_on: yesterday },
     { id: '3', name: 'r3', disabled_on: null },
   ],
+  adviserOptions: {
```
yeahfro: Technically this does not make the first commit atomic

test/vue/typeahead.test.js
```diff
@@ -151,19 +151,21 @@ describe('Typeahead', () => {
...
           name: 'adviser',
           entity: 'adviser',
+          multipleSelectOptions: false,
+          options: [],
         }
       })
 
       beforeEach(() => {
-        asyncFind = Typeahead.methods.asyncFind.bind(instance)
+        asyncSearch = Typeahead.methods.asyncSearch.bind(instance)
```
yeahfro: Up until now the pattern has been to set `this.x`

assets/javascripts/vue/typeahead.vue
```diff
@@ -94,9 +108,74 @@
...
+      } else {
+        this.isCloseOnSelect = false
+        this.multiSelectModel = null
+      }
+
+      if (!this.isAsync) {
+        this.multiSelectModel = this.model
+      }
+    },
+    created () {
```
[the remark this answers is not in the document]
ian-leggett: Wow odd will look at this not sure what happened! Thanks

package.json
```diff
@@ -25,7 +25,7 @@
     "coverage": "nyc --reporter=html --reporter=text --reporter=lcov npm run test:unit",
     "circle:unit": "nyc --reporter=lcov npm run test:unit -- --reporter mocha-circleci-reporter && codeclimate-test-reporter < coverage/lcov.info",
     "circle:vue": "mocha-webpack --webpack-config webpack.config.js --opts ./test/vue/mocha.opts --reporter mocha-circleci-reporter ./test/vue/**/*.test.js",
-    "circle:acceptance": "nightwatch --config test/acceptance/nightwatch.conf.js --verbose --env circleci",
+    "circle:acceptance": "nightwatch --config test/acceptance/nightwatch.conf.js --env circleci",
```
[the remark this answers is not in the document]
ian-leggett: So this is one for @PippoRaimondi I think it was getting too noisy to debug

assets/javascripts/vue/typeahead.vue
```diff
@@ -94,9 +108,74 @@
...
+      classes: {
+        type: String,
+        required: false,
+      },
+      isAsync: {
+        type: Boolean,
+        required: false,
+        default: true,
+      },
+      closeOnSelect: {
```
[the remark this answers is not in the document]
ian-leggett: Wow odd will look at this not sure what happened! Thanks. Looks like a merge went wrong here 🤔

test/vue/typeahead.test.js
ian-leggett: At this point the `this` context is undefined. Can you show me where an example is as I am having to reference the method off the component directly.

PR #1712, opened 2019-01-23
LH Nav Has Documents Removed For Legacy Projects

jaranF:
IPBETA-9

### DIFF
--- a/src/apps/middleware.js
+++ b/src/apps/middleware.js
@@ -49,7 +49,6 @@ function handleRoutePermissions (routes) {
 function setLocalNav (items = []) {
   return function buildLocalNav (req, res, next) {
     const userPermissions = get(res, 'locals.user.permissions')
-
     res.locals.localNavItems = items
       .filter(filterNonPermittedItem(userPermissions))
       .map((item) => {
--- a/test/unit/apps/middleware.test.js
+++ b/test/unit/apps/middleware.test.js
@@ -378,4 +378,5 @@ describe('Apps middleware', () => {
       })
     })
   })
+  // (isLegacyProject, documentsLabel, filterOnPermissions, item) {
 })
--- a/src/apps/investment-projects/constants.js
+++ b/src/apps/investment-projects/constants.js
@@ -46,7 +46,7 @@ const LOCAL_NAV = [
   },
   {
     path: 'documents',
-    label: 'Documents',
+    label: 'CDMS documents',
     permissions: [
       'investment.view_investmentproject_document',
     ],
--- a/src/apps/investment-projects/middleware/local-navigation.js
+++ b/src/apps/investment-projects/middleware/local-navigation.js
@@ -1,0 +1,17 @@
+const { startsWith, find, get, filter } = require('lodash')
+const { setLocalNav } = require('../../middleware')
+
+function setInvestmentsLocalNav (req, res, next, LOCAL_NAV, filterNonPermittedItem) {
+  const { project_code: projectCode } = res.locals.investment
+  const isLegacyProject = startsWith(projectCode, 'DHP')
+  const documentsLabel = get(find(LOCAL_NAV, { 'path': 'documents' }), 'label')
+  const userPermissions = get(res, 'locals.user.permissions')
+  const navItems = filter(LOCAL_NAV, (item) => item.label === documentsLabel && isLegacyProject
+    ? false
+    : filterNonPermittedItem(userPermissions)(item)
+  ) // _.filter() does not mutate the array as opposed to _.remove()
+  setLocalNav(navItems)(req, res, next)
+  return navItems
+}
+
+module.exports = setInvestmentsLocalNav
--- a/src/apps/investment-projects/router.js
+++ b/src/apps/investment-projects/router.js
@@ -1,13 +1,13 @@
 const router = require('express').Router()
-
+const { partialRight } = require('lodash')
 const { ENTITIES } = require('../search/constants')
 const { DEFAULT_COLLECTION_QUERY, LOCAL_NAV, APP_PERMISSIONS, QUERY_FIELDS, QUERY_DATE_FIELDS } = require('./constants')
 
 const { getRequestBody } = require('../../middleware/collection')
 const { detectUserAgent } = require('../../middleware/detect-useragent')
 const { getCollection, exportCollection } = require('../../modules/search/middleware/collection')
-
-const { setLocalNav, setDefaultQuery, redirectToFirstNavItem, handleRoutePermissions } = require('../middleware')
+const { setDefaultQuery, redirectToFirstNavItem, handleRoutePermissions } = require('../middleware')
+const { filterNonPermittedItem } = require('../../modules/permissions/filters')
 const { shared } = require('./middleware')
 const {
   getBriefInvestmentSummary,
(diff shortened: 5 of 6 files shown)


src/apps/middleware.js
```diff
@@ -46,12 +47,20 @@ function handleRoutePermissions (routes) {
   }
 }
 
+function removeDocumentsFromLHNav (isLegacyProject, documentsLabel, filterOnPermissions, item) {
```
yeahfro: This `middleware.js` should not know about investment projects ie `isLegacyProject`

src/apps/middleware.js
```diff
@@ -46,12 +47,20 @@ function handleRoutePermissions (routes) {
...
+  return item.label === documentsLabel && isLegacyProject
+    ? false
+    : filterOnPermissions(item)
+}
+
 function setLocalNav (items = []) {
   return function buildLocalNav (req, res, next) {
     const userPermissions = get(res, 'locals.user.permissions')
-
+    const isLegacyProject = startsWith(get(res, 'locals.investment.project_code'), 'DHP')
```
yeahfro: As this is specific to investment projects then it belongs in the investment project app. An example of how this was done for `companies` and be found in [local-navigation.js](https://github.com/uktrade/data-hub-frontend/blob/develop/src/apps/companies/middleware/local-navigation.js)

test/acceptance/features/step_definitions/location.js
```diff
@@ -21,14 +21,11 @@ function getExpectedValue (row, state) {
   return row.value
 }
 
-const assertNavLinkCount = async (navLinkClassName, expectedData) => {
```
yeahfro: We should keep this

test/acceptance/features/step_definitions/location.js
```diff
@@ -21,14 +21,11 @@ function getExpectedValue (row, state) {
...
-  await Location.api.elements('css selector', `.${navLinkClassName}`, (result) => {
-    client.expect(result.value.length).to.equal(expectedData.length)
-  })
-}
-
 const assertNavLinks = async (navLinkClassName, expectedData) => {
   for (const row of expectedData) {
+    if (row.text.toLowerCase() === 'documents') {
```
yeahfro: This isn't right as it's a step shared across all apps so as an example, `companies` needs to assert the link (for now anyway)

You need to set up a scenario specific to your investment project with and without the link.

test/unit/apps/middleware.test.js
```diff
@@ -378,4 +378,66 @@ describe('Apps middleware', () => {
       })
     })
   })
```
yeahfro: The pattern we normally follow for testing is as follows:

```javascript
describe('module', () => {
  context('scenario', () => {
    beforeEach(() => {
      // arrange
      
      // act
    })
    
    it('should', () => {
      // assert
    })
  })
})
```

yeahfro: Based upon moving the code from the shared `middleware.js`, this would not be the right place for this

test/unit/apps/middleware.test.js
```diff
@@ -378,4 +378,66 @@ describe('Apps middleware', () => {
...
+  // (isLegacyProject, documentsLabel, filterOnPermissions, item) {
+  describe('removeDocumentsFromLHNav', () => {
```
yeahfro: If you were keeping `removeDocumentsFromLHNav` my concern about this is that you are testing internals or implementation details which has negative implications for refactoring and maintainability.

The preference would be to extend the tests around `setLocalNav`

src/apps/middleware.js
```diff
@@ -1,6 +1,7 @@
-const { get, isEmpty, assign, intersection, isUndefined } = require('lodash')
+const { get, isEmpty, assign, intersection, isUndefined, find, startsWith, partial } = require('lodash')
```
yeahfro: With eight imports, this would have been better over multiple lines

src/apps/middleware.js
jaranF: To be honest I have my serious doubts. Can you give me the business justification for embarking on what seems to have been someone's pet project of introducing Acceptance Tests without said person being an actual tester? Also, why was JavaScript chosen as the Webdriver binding not something more industry standard such as Python / Java / C#. Natively, JavaScript isn't a go-to choice because it is single threaded (Java etc are not). Also, I don't see evidence that end clients / users inform the writing of these acceptance tests as opposed to developers adding acceptance tests as see fit. Again far, far away from industry standard. Thinks may be better if there was a document explaining the set-up of the tests (for instance I am not sure that by installing ImageMagick using Homebrew was the correct approach); anway I cannot run the AT locally and there is nowhere to go to in terms of support. I will Slack you for your phone number.

jaranF:
proj_investments_lhn_20190130

src/apps/middleware.js
jim68000: Agreed, although the original code should have done that as well

src/apps/middleware.js
jim68000: Agreed - is there a logical place for this to live in Investment?

src/apps/middleware.js
jim68000: These are all fair points, and I can provide some background. JS was chosen over Python because the policy has always been for devs rather than QAs to write the tests, and no-one wanted to learn Python. This obviously leads to threading issues but we get around that by parallelising the tests when they run on CI.

The points about lack of QA oversight, and lack of involvement by business owners is entirely fair. I'm aware there is a structural debt issue around Acceptance Testing and we will be working with Pippo (who is the first QA we've had on the project since the beginning of 2018) over the next few months to restructure our testing into a more sensible unit/integration/acceptance pyramid. Once that's done we can move to getting tests reviewed by business owners - again we've had a lack of owners who can articulate these things, so have probably overcompensated. Again something we will be tackling in Data Hub live services.

For now though, I'd like us to maintain the same level of testing while we prep the new world. As for running TA locally - it is tricky and I tend to rely on WIP PRs and CircleCI to do it for me. CI uses a different back end anyway which causes some odd issues - again, Pippo is building a more comprehensive mocking server to get over this issue, but we're at the early stages of that project.

test/unit/apps/middleware.test.js
yeahfro: Although this is a small example, if we start doing it everywhere then we have a bunch of tests which are testing private functions. More code implies more maintenance points. In addition, every time we refactor or fix a bug, we need to update more tests.

As a rule of thumb, alarm bells should be ringing if you are exposing a function solely for testing.

If you feel the need to be exposing private functions then you are trying to test too much or the class/module is doing too much.

src/apps/middleware.js
yeahfro: The recommendation is to follow a pattern similar to [`setCompaniesLocalNav`](https://github.com/uktrade/data-hub-frontend/blob/develop/src/apps/companies/middleware/local-navigation.js)

jaranF: Yes, I agree and see what I did wrong here. Thanks for the pointer you gave me (apropos look at Companies); I've implemented that way and removed 'removeDocumentsFromLHNav()' entirely.

jaranF: Commit d95a7582 should user in the move of file

src/apps/middleware.js
jaranF: apropos 'local-navigation.js' following the technique as used by in 'companies'. yes, you should see this change in commit d95a7582

test/acceptance/features/step_definitions/location.js
jaranF: Despite getting Ian L. up to help me I never got the ATs to work... as it was - as per Jim's comment -  then only way I ever got to know if a AT test was broken because of my code was to see what happened in CircleCI

test/acceptance/features/step_definitions/location.js
jaranF: Pls see comment made a few seconds ago about not being able to set up AT's locally

test/unit/apps/middleware.test.js
jaranF: Yes, my tests should not have been in 'middleware.test.js' I have moved them. 
Also regarding the pattern mentioned above ...describe...context...it(should) ... I got Ian L. over to eyeball my tests and they read clearly to him and he would say that they fall into the prescribed pattern
  ...

yeahfro:
Closing as this work was completed in #1729

PR #1717, opened 2019-01-28
Show total value on OMIS orders

ian-leggett:
## Problem
Users would like to see the total value of orders on a collection list page.

## Solution
Show the total value at the top of the collection list.
screenshot 2019-01-28 at 15 24 50

### DIFF
--- a/src/modules/api/transformers/api-response-to-collection.js
+++ b/src/modules/api/transformers/api-response-to-collection.js
@@ -36,6 +36,7 @@ function transformApiResponseToCollection (options = {}, ...itemTransformers) {
     return assign({}, {
       items,
       count: response.count,
+      summary: response.summary,
       pagination: buildPagination(options.query, response),
     })
   }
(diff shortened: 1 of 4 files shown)


src/modules/api/transformers/api-response-to-collection.js
```diff
@@ -36,6 +36,7 @@ function transformApiResponseToCollection (options = {}, ...itemTransformers) {
     return assign({}, {
       items,
       count: response.count,
+      summary: response.summary,
```
yeahfro: Is this `summary` field specific to `Orders`? This transformer is being used across all entities and should not know about fields specific to particular entities.

src/templates/_macros/collection/collection-header.njk
```diff
@@ -13,6 +13,7 @@
 #}
 {% macro CollectionHeader(props) %}
   {% set buildQuery = buildQuery or props.buildQuery %}
+  {% set totalCost = props.summary.total_subtotal_cost %}
```
yeahfro: Again I feel uncomfortable about adding entity specific fields to a template used for all collections.

PR #1729, opened 2019-02-08
Removing the document nav item when a project code starts with 'DHP'

paulgain:
Currently there are 3 types of project codes, the oldest is 1 and
the latest is 3:

1. 016020 (really old)
2. P-30016857 (CDMS)
3. DHP-00000007 (Data Hub)

At present the document nav item is displayed for all 3 scenarios. This
work excludes the document nav item when the project is new, all new
projects have a code that start with 'DHP'. The documents nav item is
specifically and solely for the S3 migrated documents that are
associated with project codes such as 1 & 2.

https://uktrade.atlassian.net/browse/IPBETA-9

### DIFF
--- a/src/apps/investment-projects/constants.js
+++ b/src/apps/investment-projects/constants.js
@@ -46,7 +46,7 @@ const LOCAL_NAV = [
   },
   {
     path: 'documents',
-    label: 'Documents',
+    label: 'CDMS documents',
     permissions: [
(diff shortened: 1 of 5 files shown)


paulgain:
**This nav item will only be visible for older projects**
* 016020 (really old)
* P-30016857 (CDMS)

removed-documents-link

PR #1737, opened 2019-02-13
Adding three investment tabs/routes to investment-projects

paulgain:
Note: this is work in progress.

Adding three new tabs into investment projects
  - Projects
  - Profiles
  - Opportunities

Each one of these tabs has a corresponding route
  - /investment-projects/projects
  - /investment-projects/profiles
  - /investment-projects/opportunities

Before going live and once prudent to do so /investment-projects will be renamed
to /investments, all Data Hub links will be updated accordingly.

The current investment list has been moved into the the projects tab, however, the current
implementation remains unchanged apart from code reuse across both routes.

To be clear:
/investment-projects and /investment-projects/projects are identical except the latter is
within a tab. The former will be removed once we're good to go.

There is no feature flag as the new functionality is hidden behind new routes which the user
has no link to.

Acceptance tests to follow.

https://uktrade.atlassian.net/browse/IPBETA-226

### DIFF
--- a/assets/stylesheets/components/_pagination.scss
+++ b/assets/stylesheets/components/_pagination.scss
@@ -74,15 +74,6 @@
   color: $grey-1;
 }
 
-.c-pagination__navigation {
-  @include clearfix;
-}
-
-.c-pagination__summary {
-  margin-top: $default-spacing-unit;
-  color: $grey-1;
-}
-
 @include media(mobile) {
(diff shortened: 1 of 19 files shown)


paulgain:
new-investment-tabs

paulgain:
**Current implementation unchanged for users**.

current-implementation-unchanged

PR #1738, opened 2019-02-13
Update typeahead to use new params for search adviser API

ian-leggett:
## Problem
- The API search behaviour has changed and we can now search for advisors by team name as well as the first name/last name. 

- Long selected options in the typeahead need to be wrapped as they are currently overflowing their containers.
screenshot 2019-02-13 at 14 46 11

Trello - https://trello.com/c/HgyRp7Bu/506-search-behaviour-for-add-advisor-type-ahead

## Solution
- Update the search params in the typeahead to call the search API with:
`autocomplete`
`is_active`

instead of the old params `first_name__icontains` and `last_name__icontains`
- Update styles to wrap long content in selected options

### DIFF
--- a/src/apps/adviser/repos.js
+++ b/src/apps/adviser/repos.js
@@ -1,4 +1,4 @@
-const { get, isEmpty } = require('lodash')
+const { get } = require('lodash')
 const config = require('../../../config')
 const { authorisedRequest } = require('../../lib/authorised-request')
 
@@ -19,11 +19,9 @@ function getAdviser (token, id) {
 }
 
 async function fetchAdviserSearchResults (token, params) {
-  if (!isEmpty(params.term)) {
-    const url = `${config.apiRoot}/adviser/?autocomplete=${params.term}&is_active=${params.is_active}`
-    const adviserResults = await authorisedRequest(token, { url })
-    return adviserResults.results
-  }
+  const url = `${config.apiRoot}/adviser/?autocomplete=${params.term}&is_active=${params.is_active}`
+  const adviserResults = await authorisedRequest(token, { url })
+  return adviserResults.results
 }
 
 module.exports = {
--- a/src/apps/adviser/repos.js
+++ b/src/apps/adviser/repos.js
@@ -18,45 +18,16 @@ function getAdviser (token, id) {
   return authorisedRequest(token, `${config.apiRoot}/adviser/${id}/`)
 }
 
-function fetchAdviserSearchResults (token, firstName, lastName) {
-  let url = `${config.apiRoot}/adviser/?first_name__icontains=${firstName}`
-  if (!isEmpty(lastName)) url += `&last_name__icontains=${lastName}`
-
-  return authorisedRequest(token, { url })
-}
-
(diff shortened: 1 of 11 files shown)


src/apps/adviser/repos.js
```diff
@@ -18,41 +18,21 @@ function getAdviser (token, id) {
   return authorisedRequest(token, `${config.apiRoot}/adviser/${id}/`)
 }
 
-function fetchAdviserSearchResults (token, firstName, lastName) {
-  let url = `${config.apiRoot}/adviser/?first_name__icontains=${firstName}`
-  if (!isEmpty(lastName)) url += `&last_name__icontains=${lastName}`
+function fetchAdviserSearchResults (token, params) {
+  const isActive = params.is_active
```
yeahfro: It's a little out of place that you are declaring `isActive` here and then `params.term` is on the next line. `params.is_active` would also read fine on the next line

src/apps/adviser/repos.js
```diff
@@ -18,41 +18,21 @@ function getAdviser (token, id) {
...
-function adviserNameSort (a, b) {
-  if (a.name.toLowerCase() < b.name.toLowerCase()) return -1
-  if (a.name.toLowerCase() > b.name.toLowerCase()) return 1
-  return 0
-}
-
-async function adviserSearch (token, term) {
-  if (isEmpty(term)) {
+async function adviserSearch (token, params) {
+  if (isEmpty(params.term)) {
```
yeahfro: This could essentially be:

```
async function adviserSearch (token, params) {
  if (!isEmpty(params.term)) {
    return fetchAdviserSearchResults(token, params).results
  }
}
```

And then it makes you think that `adviserSearch` is just an unnecessary abstraction of `fetchAdviserSearchResults`

reupen:
@ian-leggett Just spotted that there's a JavaScript error if you type 'smi' on the preview app for this PR (in the adviser filter on the interaction collection page).

src/apps/adviser/repos.js
ian-leggett: Good spot, this should be fixed now

src/apps/adviser/repos.js
ian-leggett: Agreed I have now moved unnecessary abstraction

src/apps/adviser/repos.js
```diff
@@ -18,45 +18,16 @@ function getAdviser (token, id) {
...
-  if (a.name.toLowerCase() < b.name.toLowerCase()) return -1
-  if (a.name.toLowerCase() > b.name.toLowerCase()) return 1
-  return 0
-}
-
-async function adviserSearch (token, term) {
-  if (isEmpty(term)) {
-    return
+async function fetchAdviserSearchResults (token, params) {
+  if (!isEmpty(params.term)) {
```
yeahfro: This negative scenario is not being covered in unit tests

ian-leggett: I have now removed the guard so no test needed now :-)

PR #1743, opened 2019-02-14
Integrate Introspect endpoint with permitted applications

yozzo:
On top of regular user permissions there is a need to have permitted applications vetting, such that not all users should see all applications in the global header nav bar, but just the ones they have been given access to.


src/apps/transformers.js
```diff
@@ -78,6 +80,28 @@ function transformDateStringToDateObject (dateString) {
...
+
+  return collection
+}
+
+function transformAppsToPermittedApps (appsNamesAndPaths, permittedApplications) {
+  const appKeys = map(appsNamesAndPaths, (item) => item.key)
+  const permittedAppsKeys = map(permittedApplications, (item) => item.key)
+  const keys = intersection(appKeys, permittedAppsKeys)
+
+  return transformPermittedAppsToCollection(appsNamesAndPaths, keys)
```
yeahfro: I wonder if this code could be simplified as you're getting a collection of `keys` from and existing collection and then iterating over that collection.

Basically I am seeing four uses of `map` and one use of `intersection` which feels like the same could be achieved with less.

src/middleware/locals.js
```diff
@@ -17,6 +19,7 @@ try {
 module.exports = function locals (req, res, next) {
   const baseUrl = `${(req.encrypted ? 'https' : req.protocol)}://${req.get('host')}`
   const userPermissions = get(res, 'locals.user.permissions')
+  const introspect = req.session && !config.oauth.bypassSSO ? req.session.introspect : null
```
yeahfro: My thoughts with the name `introspect` is that it's a verb and doesn't really mean much. I don't feel like it's very self documenting.

src/middleware/locals.js
```diff
@@ -32,10 +35,25 @@ module.exports = function locals (req, res, next) {
...
-          label,
-          url,
-          isActive: req.path.startsWith(url),
+        const isActive = req.path.startsWith(url)
+
+        if (isObject(introspect)) {
+          const permittedApps = transformAppsToPermittedApps(GLOBAL_NAV_APP_KEYS, introspect.permitted_applications)
+          const appNames = map(permittedApps, (item) => item.name)
+
+          if (includes(appNames, navItem.label)) {
```
yeahfro: It seems like this can be simplified as the same `object` is being returned twice.

yeahfro: This will need some unit tests.

src/apps/transformers.js
yeahfro: There needs to be some unit tests in this area.

src/apps/global-nav-items.js
```diff
@@ -25,4 +25,38 @@ const GLOBAL_NAV_ITEMS = concat(SORTED_APP_GLOBAL_NAV_ITEMS, {
   label: 'Find exporters',
 })
 
-module.exports = GLOBAL_NAV_ITEMS
+/**
+ * Required for querying access to apps
+ */
+const GLOBAL_NAV_APP_KEYS = [
```
yeahfro: I wonder if actually each app should be given the key rather than maintaining this separate collection of keys. ie [add a key in `index.js`](https://github.com/uktrade/data-hub-frontend/blob/40108b839d0cd96442c606b91e476f8552348ecd/src/apps/companies/index.js#L4)

src/middleware/locals.js
yeahfro: Abstracting to a function might help

src/apps/global-nav-items.js
yozzo: that's a very good point, and it also simplifies the logic a lot

src/apps/transformers.js
yozzo: after adding the keys for each app individually, these functions no longer needed, thus removing them, yay!

src/middleware/locals.js
yozzo: refactored it to `userProfile`

src/middleware/locals.js
```diff
@@ -30,14 +34,7 @@ module.exports = function locals (req, res, next) {
...
       .filter(filterNonPermittedItem(userPermissions))
-      .map(navItem => {
-        const { path: url, label } = navItem
-        return {
-          label,
-          url,
-          isActive: req.path.startsWith(url),
-        }
-      }),
+      .map(navItem => buildNavObject(req, navItem, permittedApplications, config.oauth.bypassSSO)),
```
yeahfro: @yozzo this is going in the right direction but as you are essentially filtering in `buildNavObject` my suggestion would be to follow a similar pattern to `filterNonPermittedItem` which would looking something like (untested):

```
const filterNonSsoUserProfileItem = (permittedApps, bypassSSO) => {
  return (item) => {
    return bypassSSO || some(permittedApps, (key) => key === item.key)
  }
}
```

Then you can carry on building the `object`.

src/middleware/locals.js
```diff
@@ -17,6 +19,8 @@ try {
 module.exports = function locals (req, res, next) {
   const baseUrl = `${(req.encrypted ? 'https' : req.protocol)}://${req.get('host')}`
   const userPermissions = get(res, 'locals.user.permissions')
+  const userProfile = req.session && !config.oauth.bypassSSO ? req.session.userProfile : null
+  const permittedApplications = isObject(userProfile) ? userProfile.permitted_applications : []
```
yeahfro: For this I would use `get(userProfile, 'permitted_applications', [])`

src/middleware/locals.js
```diff
@@ -17,6 +19,8 @@ try {
 module.exports = function locals (req, res, next) {
   const baseUrl = `${(req.encrypted ? 'https' : req.protocol)}://${req.get('host')}`
   const userPermissions = get(res, 'locals.user.permissions')
+  const userProfile = req.session && !config.oauth.bypassSSO ? req.session.userProfile : null
```
yeahfro: This would probably read better as `const userProfile = config.oauth.bypassSSO ? null : get(req.session, 'userProfile')`

PR #1750, opened 2019-02-15
Feature/add interaction cypress tests

PippoRaimondi:


src/templates/_macros/common/global-nav.njk
```diff
@@ -15,7 +15,7 @@
         <div class="govuk-width-container">
           <ul class="global-nav__list">
             {% for item in props.items %}
-              <li class="global-nav__list-item">
+              <li class="global-nav__list-item" test-data-id="{{ item.label }}">
```
yeahfro: My preference would be for the naming convention to be `data-test-id`

yeahfro: I am dubious of `item.label` being used as the identifier as it will likely cause problems

test/functional/cypress/support/commands.js
```diff
@@ -0,0 +1,25 @@
+// ***********************************************
```
yeahfro: Lots of commented out code 🤔

src/apps/companies/views/_deprecated/interactions.njk
```diff
@@ -1,7 +1,7 @@
 {% extends "./template.njk" %}
 
 {% block main_grid_right_column %}
-<h2 class="govuk-heading-m">Company interactions</h2>
+<h2 class="govuk-heading-m" test-auto-id="companyInteractionHeader">Company interactions</h2>
```
yeahfro: This is another naming convention being used for the attribute ie `test-auto-id` vs originally using `test-data-id`

My vote would be for `data-test-id` or `data-auto-id`.

Can we put this to the team?

src/templates/_macros/collection-header/template.njk
```diff
@@ -23,7 +23,7 @@
     {% if actionButtons.length %}
       <div class="c-collection__header-actions">
         {% for actionButton in actionButtons %}
-          <a class="button" href="{{ actionButton.url }}">{{ actionButton.label }}</a>
+          <a class="button" href="{{ actionButton.url }}" test-auto-id="{{ actionButton.label }}">{{ actionButton.label }}</a>
```
yeahfro: Again, I am dubious of using the label to identify elements.

src/templates/_macros/common/global-nav.njk
PippoRaimondi: I think data-test-id is the standard so you are right! I will refactor it in the next couple of commits, good catch!

test/functional/cypress/support/commands.js
PippoRaimondi: that's generated code, I can get rid of it... although for people less familiar with the tool, it could be useful as a reminder that we have a way to introduce custom commands...

src/apps/companies/views/_deprecated/interactions.njk
PippoRaimondi: one of the above (can't remember whcih though) is the standard, let me try to figure out which one it is and we can stick with that one?

src/templates/_macros/collection-header/template.njk
PippoRaimondi: ok, I will think of using some other attribute, will refactor soon!

test/functional/cypress/specs/interaction/add-interaction-spec.js
```diff
@@ -0,0 +1,141 @@
+import selectors from '../../selectors'
+
+let subject
+
+describe('Add Interaction', () => {
+  beforeEach(() => {
+    subject = Math.random().toString(36).substring(7)
```
yeahfro: Can we do this in some kind of helper? [As an example it has been previously done](https://github.com/uktrade/data-hub-frontend/blob/develop/test/acceptance/helpers/uuid.js) in the existing acceptance tests.

PippoRaimondi: yep, that can be done!

PippoRaimondi:
Thanks a lot for the feedback @yeahfro:

The timings defers a lot when running the tests against the mocking service locally (35-40s) vs where it's currently deployed (75s+)... despite the mocking service being very light weighted, noticing the amount of times the mocking service is hit (300x) explains reason it takes double the time running against the server...

Also, while mocking things out and monitoring the application logs, I noticed a lot of duplicate calls being made too, without understanding properly the reason behind it... so maybe the underlying issue of the tests being slow is lack of caching + unnecessary calls made throughout the flow of the tests... An alternative to speed things up (other than improving the app itself) is have the mocking service run locally, but will get to that in time...

1:. Regarding commonjs vs es6, I was wondering if there is any reasoning behind using commonjs? consistency with the rest of the project?
2:. I didn't refactor the selectors, it could use a bit of renaming/shuffling... will work on it
3:. I was unhappy with the tr index reference too, will have it migrated to the selectors!
4:. I can move the config files to the parent dir, I just noticed we manage everything there instead and that's cool!

I am just waiting for this branch to get merged to include these tests in the pipeline along with reducing the tests run in nightwatch (out of the 13 tests, we should keep 2 and have similar coverage).

PR #1772, opened 2019-03-01
Bugfix/header spacing issues

yozzo:
Fix spacing issues on global header

Add a fallback for `inital` property which doesn't work in IE
screenshot 2019-03-01 at 13 46 07

### DIFF
--- a/assets/stylesheets/objects/base/_global-header.scss
+++ b/assets/stylesheets/objects/base/_global-header.scss
@@ -41,6 +41,7 @@
 
     @include govuk-media-query($from: tablet) {
       width: auto;
+      width: initial;
       display: inline-block;
       list-style: none;
     }
@@ -106,6 +107,7 @@
 
     @include govuk-media-query($from: tablet) {
       width: auto;
+      width: initial;
       display: inline;
       list-style: none;
       padding: 14px 0;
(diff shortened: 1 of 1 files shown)


assets/stylesheets/objects/base/_global-header.scss
```diff
@@ -135,8 +137,8 @@
     }
 
     @include govuk-media-query($from: tablet) {
-      display: inline-block;
-      width: auto;
+      display: inline;
+      display: initial;
```
ian-leggett: We can't use initial as its not supported in IE

assets/stylesheets/objects/base/_global-header.scss
```diff
@@ -106,6 +107,7 @@
 
     @include govuk-media-query($from: tablet) {
       width: auto;
+      width: initial;
```
ian-leggett: We can't use initial as its not supported in IE

assets/stylesheets/objects/base/_global-header.scss
```diff
@@ -41,6 +41,7 @@
...
```
ian-leggett: We can't use initial as its not supported in IE

assets/stylesheets/objects/base/_global-header.scss
yozzo: it's for non IE, so it falls-back to the previous one when it doesn't work.

PR #1773, opened 2019-03-01
Update File input styling

yozzo:
the styling for the file type input regressed so some properties had to
be adjusted


jim68000:
Code looks good, but have you tested on IE?

yozzo:
@jim68000 yes, on IE it falls-back to the native styling, so even now, without the changes it looks fine.

PR #1780, opened 2019-03-05
Bug - update contacts query param in events

ian-leggett:
## Problem
The events still query the contacts using the old param value `contact_id` when it should be `contacts__id`. Also when transforming the data events is still consuming `contact` in the transformer when it should be `contacts` as the contacts is now an array which was changed around the new use of multiple contacts.

## Solution
Update the query param and update the transformer to use the right contacts.

### DIFF
--- a/src/apps/events/attendees/transformers.js
+++ b/src/apps/events/attendees/transformers.js
@@ -4,10 +4,10 @@ const { compact, get, pickBy } = require('lodash')
 const { attendeeLabels } = require('./labels')
 const { fetchEventAttendees } = require('./repos')
 
-function transformServiceDeliveryToAttendeeListItem ({ contact, company, date, id }) {
+function transformServiceDeliveryToAttendeeListItem ({ contacts = [], company, date, id }) {
   const metaItems = [
     { key: 'company', value: get(company, 'name'), url: `/companies/${get(company, 'id')}` },
-    { key: 'job_title', value: contact ? contact.job_title : 'Not available' },
+    { key: 'job_title', value: contacts.length ? contacts[0].job_title : 'Not available' },
     { key: 'attended_date', value: date, type: 'date' },
     { key: 'service_delivery', value: 'View or edit service delivery', url: `/interactions/${id}` },
   ]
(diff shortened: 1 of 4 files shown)


src/apps/events/attendees/transformers.js
```diff
@@ -4,10 +4,10 @@ const { compact, get, pickBy } = require('lodash')
 const { attendeeLabels } = require('./labels')
 const { fetchEventAttendees } = require('./repos')
 
-function transformServiceDeliveryToAttendeeListItem ({ contact, company, date, id }) {
+function transformServiceDeliveryToAttendeeListItem ({ contacts, company, date, id }) {
   const metaItems = [
     { key: 'company', value: get(company, 'name'), url: `/companies/${get(company, 'id')}` },
-    { key: 'job_title', value: contact ? contact.job_title : 'Not available' },
+    { key: 'job_title', value: contacts ? contacts[0].job_title : 'Not available' },
```
paulgain: What happens here if contacts is null, undefined or an empty array? We should put some safeguards in place to avoid a TypeError.

src/apps/events/attendees/transformers.js
```diff
@@ -17,11 +17,11 @@ function transformServiceDeliveryToAttendeeListItem ({ contact, company, date, i
       label: attendeeLabels[key],
     }))
 
-  const listItem = contact
+  const listItem = contacts
```
paulgain: Same applies here.

src/apps/events/attendees/transformers.js
ian-leggett: Good point will add guard around this 👍

src/apps/events/attendees/transformers.js
ian-leggett: And this thanks Paul 👍

src/apps/events/attendees/transformers.js
```diff
@@ -63,7 +63,7 @@ async function createContactItemToAttendeeSearchResult (token, event) {
 }
 
 function existingAttendee ({ id }, attendees) {
-  return attendees.results.find(attendee => attendee.contact.id === id)
+  return attendees.results.find(attendee => attendee.contacts[0].id === id)
```
paulgain: There is still a potential Type Error here if the array is empty. How about something like this?
```
function existingAttendee ({ id }, attendees) {
  return attendees.results.find((attendee) => {
    const contact = attendee.contacts && attendee.contacts[0];
    return contact ? contact.id === id : false
  })
}
```

ian-leggett: Makes sense just updated this.

PR #1789, opened 2019-03-06
Fix proposition download link

yeahfro:
https://trello.com/c/LZQrc8eG/849-fix-proposition-download-link

## Problem
There was a regression when refactoring the key value table which means that `Download` links were not being rendered correctly.

## Fix
The transformation of the links now returns an `object` containing `name` and `url` fields. This removes the dependency on `download` specific types in the shared key value table component.

In addition more unit tests have been introduced to cover more scenarios.

## Before
screenshot 2019-03-06 at 22 15 24

## After
screenshot 2019-03-06 at 22 15 05

### DIFF
--- a/src/apps/documents/transformers.js
+++ b/src/apps/documents/transformers.js
@@ -17,54 +17,25 @@ function transformLabelsToShowFiles (key, labels) {
 }
 
 function getDownloadLinkOrState (file, proposition_id, investment_project_id) {
-  const output = {
-    type: 'document',
+  const items = {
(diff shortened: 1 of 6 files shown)


src/apps/documents/transformers.js
```diff
@@ -17,54 +17,25 @@ function transformLabelsToShowFiles (key, labels) {
 }
 
 function getDownloadLinkOrState (file, proposition_id, investment_project_id) {
```
paulgain: This function looks a lot cleaner and concise. 👍

PR #1798, opened 2019-03-08
Add cypress test for D&B GHQ on One List, viewing business details

yeahfro:
This test is intended as a replacement for the [`View business details for a Dun & Bradstreet GHQ company on the One List not in the UK` Nightwatch scenario](https://github.com/uktrade/data-hub-frontend/blob/8980f97f224d258e12b1a7fe7a0955735c4744ac/test/acceptance/features/companies/business-details.feature#L4).

Over time, I will replace all of the scenarios to remove `business-details.feature`.

### DIFF
--- a/test/functional/cypress/specs/companies/business-details-spec.js
+++ b/test/functional/cypress/specs/companies/business-details-spec.js
@@ -1,0 +1,134 @@
+const { keys, forEach } = require('lodash')
+
+const oneListCorp = require('../../fixtures/one-list-corp.json')
+const selectors = require('../../selectors/index.js')
+
+describe('Companies business details', () => {
+  context('when viewing business details for a Dun & Bradstreet GHQ company on the One List not in the UK', () => {
+    before(() => {
+      cy.visit(`/companies/${oneListCorp.id}/business-details`)
+    })
+
+    it('should display the "Business details" heading', () => {
+      cy.get(selectors.localHeader().heading).should('have.text', 'Business details')
+    })
+
+    it('should display the "Where does information on this page come from?" details summary', () => {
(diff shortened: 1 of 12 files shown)


test/functional/cypress/specs/companies/business-details-spec.js
```diff
@@ -0,0 +1,134 @@
...
+
+  const assertKeyValueTable = (dataAutoId, expected) => {
+    forEach(keys(expected), (key, i) => {
+      const rowNumber = i + 1
+      cy.get(selectors.keyValueTable(dataAutoId).keyCell(rowNumber)).should('have.text', key)
+      cy.get(selectors.keyValueTable(dataAutoId).valueCell(rowNumber)).should('have.text', expected[key])
+    })
+  }
+
+  const assertValueTable = (dataAutoId, expected) => {
```
PippoRaimondi: This is fine as an example of how we want to reduce code by having functions to assist in repetitive checks, similar to the `assertKeyValueTable` although in this particular case we only need to validate a single row, so a 1 liner get/assertion could have done it for now...

test/functional/cypress/specs/companies/business-details-spec.js
```diff
@@ -0,0 +1,134 @@
+const { keys, forEach } = require('lodash')
+
+const oneListCorp = require('../../fixtures/one-list-corp.json')
+const selectors = require('../../selectors/index.js')
+
+describe('Companies business details', () => {
+  context('when viewing business details for a Dun & Bradstreet GHQ company on the One List not in the UK', () => {
+    before(() => {
+      cy.visit(`/companies/${oneListCorp.id}/business-details`)
```
PippoRaimondi: didn't realise we could import a json and use it like an object, pretty cool!

PR #1808, opened 2019-03-13
Format currency as GBP from USD for D&B companies

yeahfro:
https://trello.com/c/qqNKNRST/733-convert-annual-turnover-from-usd-to-gbp

## Change
Currently the `turnover` of Dun & Bradstreet companies is being presented in USD without formatting. This change converts the USD to GBP using the average rate from 2018. The currency is presented with formatting.

A further iteration will be to remove some of the zeros. As an example `£1,000,000.00` could be presented as `£1 million`

## Before
Screenshot 2019-03-13 at 13 23 55

## After
Screenshot 2019-03-13 at 13 27 00

### DIFF
--- a/src/apps/companies/transformers/company-to-about-view.js
+++ b/src/apps/companies/transformers/company-to-about-view.js
@@ -4,11 +4,15 @@ const { get, pickBy, isEmpty } = require('lodash')
 const { aboutLabels } = require('../labels')
 const { getDataLabels } = require('../../../lib/controller-utils')
 const { NOT_SET_TEXT } = require('../constants')
+const { currencyRate } = require('../../../../config')
 
 function transformTurnover (turnover, turnover_range) {
   if (turnover) {
     return [
-      `USD ${turnover}`,
+      {
+        name: turnover * currencyRate.usdToGbp,
+        type: 'currency',
+      },
       {
         name: 'This is an estimated number',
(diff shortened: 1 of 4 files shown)


test/functional/cypress/specs/companies/business-details-spec.js
```diff
@@ -267,7 +267,7 @@ describe('Companies business details', () => {
     it('should display the "About" details', () => {
       assertKeyValueTable('aboutDetails', {
         'Trading names': 'DnBD&B',
-        'Annual turnover': 'USD 1000000This is an estimated numberWhat does that mean?This is an estimated number',
+        'Annual turnover': '£750,148.00This is an estimated numberWhat does that mean?This is an estimated number',
```
ian-leggett: Do we need a space after the value?

src/apps/companies/transformers/company-to-about-view.js
```diff
@@ -4,11 +4,15 @@ const { get, pickBy, isEmpty } = require('lodash')
 const { aboutLabels } = require('../labels')
 const { getDataLabels } = require('../../../lib/controller-utils')
 const { NOT_SET_TEXT } = require('../constants')
+const config = require('../../../../config')
```
paulgain: How about object destructuring?

PR #1818, opened 2019-03-15
Adding Companies/Investment Large Capital Profile tab and API call

paulgain:
When a user selects the "Large capital profile" tab we make an API to a new endpoint:
**/v4/large-investor-profile**. If the Company does not have a profile then show the "Create a profile" button.

https://uktrade.atlassian.net/browse/IPBETA-353

large-capital-profile-tab


src/apps/companies/apps/investments/large-capital-profile/views/main.njk
```diff
@@ -0,0 +1,11 @@
+{% from "companies/apps/investments/macros/navigation.njk" import subnavigation %}
+
+{{ subnavigation(company, 'large-capital-profile') }}
+
+<h2 class="govuk-heading-m" data-cy="investment-subheading">Large capital investor profile</h2>
+
+{% if companyProfile  %}
+  <pre>{{ companyProfile | dump(2) }}</pre>
```
paulgain: As all of this work is hidden from the user I wish this to remain. Thanks.

test/functional/cypress/specs/companies/investments/large-capital-profile-spec.js
```diff
@@ -0,0 +1,61 @@
+const fixtures = require('../../../fixtures/index.js')
+const selectors = require('../../../selectors/index.js')
+
+const baseUrl = Cypress.config().baseUrl
+const id = fixtures.company.oneListCorp.id
```
yeahfro: Would it be better to destructure `oneListCorp` and then reference `oneListCorp.id` in the tests? I feel that reads better

test/functional/cypress/selectors/company-investment.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
+  investmentProjectsTab: '[data-cy=investment-projects-tab]',
```
yeahfro: We have been using `data-auto-id`

test/functional/cypress/specs/companies/investments/large-capital-profile-spec.js
paulgain: Ha! Brillant. Touché.

test/functional/cypress/selectors/company-investment.js
paulgain: At least `cy` refers to cypress what does 'auto' refer to?

paulgain:
> Could the `sub-navigation` actually be a `TabbedLocalNav` and then `.govuk-tabs__list--full-sub` modifier class that will apply the styles you want?

To be honest I wasn't aware of the modifier. Our sub-tabs may (currently being designed) behave differently when on smaller devices to the TabbedLocalNav. I'm not redoing work at this stage as we have a lot to get through. I've noticed when resizing the browser using TabbedLocalNav some of the tabs start to stack on each other (the ones on the far right) with is not what we want.

src/apps/companies/apps/investments/large-capital-profile/styles.scss
```diff
@@ -0,0 +1,21 @@
+nav.sub-navigation {
```
ian-leggett: Do we want to hang the class on the HTML element? Could this just be `.sub-navigation` just thinking of possible reuse.

src/apps/companies/apps/investments/large-capital-profile/styles.scss
```diff
@@ -0,0 +1,21 @@
...
+  border-bottom: 1px solid #c6c8c8;
+  ol {
```
ian-leggett: Also I would use a class here as this could be a un-ordered list maybe.

src/apps/companies/apps/investments/large-capital-profile/styles.scss
paulgain: I'm not sure I follow @ian-leggett. This macro is reused an additional two times although these are in the pull requests that follow this.

ian-leggett: @paulgain no good point sorry I was thinking of something else here all good.

PR #1819, opened 2019-03-15
Adding a "Growth capital profile" tab into Companies/Investment

paulgain:
When a user selects the "Growth capital profile" tab we will make an
API call to a new endpoint although this is further down the line as
it's not a top priority at present.

For now we can just show the "Create a profile" button without any
API call or logic being applied and start to build out the functional
tests.

https://uktrade.atlassian.net/browse/IPBETA-373

Screenshot 2019-03-15 at 16 19 50

### DIFF
--- a/src/apps/companies/apps/investments/growth-capital-profile/views/main.njk
+++ b/src/apps/companies/apps/investments/growth-capital-profile/views/main.njk
@@ -1,0 +1,11 @@
+{% from "companies/apps/investments/macros/navigation.njk" import subnavigation %}
+
+{{ subnavigation(company, 'growth-capital-profile') }}
+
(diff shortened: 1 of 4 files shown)


src/apps/companies/apps/investments/growth-capital-profile/views/main.njk
```diff
@@ -0,0 +1,11 @@
+{% from "companies/apps/investments/macros/navigation.njk" import subnavigation %}
+
+{{ subnavigation(company, 'growth-capital-profile') }}
+
+<h2 class="govuk-heading-m" data-auto-id="investment-subheading">Growth capital investor profile</h2>
+
+{% if companyProfile  %}
+  <pre>{{ companyProfile | dump(2) }}</pre>
```
paulgain: Again, this is all hidden from the user and I want this to remain. Thanks.

PR #1826, opened 2019-03-21
Rework Dockerfile to be based on test base image

currycoder:
The Dockerfile used for setting up local dev copies was quite outdated.  This lead to local `docker-compose build` commands to take over 30 minutes to complete.

This change cleans up the Dockerfile and makes it depend on our dockerhub-hosted frontend base image, which keeps things more DRY.
Also fixed an issue where the app would frequently crash when modifying source files - this was due to an incompatibility between nodemon and docker.  The fix is a little bit of a hack, but there's not really a more elegant alternative until a fix is made to nodemon directly.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,56 +1,9 @@
-FROM ubuntu:16.04
+FROM ukti/docker-data-hub-base:latest
 
-ENV NODE_VERSION 8.11.3
-ENV YARN_VERSION 1.7.0
-ENV NPM_CONFIG_LOGLEVEL info
+ENV NODE_PATH $NVM_DIR/versions/node/v$NODE_VERSION/lib/node_modules
+ENV PATH      $NVM_DIR/versions/node/v$NODE_VERSION/bin:$PATH
 
-RUN apt-get update -qq
-RUN apt-get upgrade -y
-RUN apt-get install -y build-essential libpq-dev curl libpng-dev
-
-# Node js
-# gpg keys listed at https://github.com/nodejs/node#release-team
-RUN set -ex \
-  && for key in \
-    9554F04D7259F04124DE6B476D5A82AC7E37093B \
-    94AE36675C464D64BAFA68DD7434390BDBE9B9C5 \
-    FD3A5288F042B6850C66B31F09FE44734EB7990E \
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -1,68 +1,22 @@
-FROM ubuntu:16.04
+FROM ukti/docker-data-hub-base:latest
 
-ENV NODE_VERSION 8.11.3
-ENV YARN_VERSION 1.7.0
-ENV NPM_CONFIG_LOGLEVEL info
+ENV NODE_PATH $NVM_DIR/versions/node/v$NODE_VERSION/lib/node_modules
```
PippoRaimondi: is this necessary given it's set in the base image?

Dockerfile
```diff
@@ -1,68 +1,22 @@
...
-  done \
-  && curl -fSLO --compressed "https://yarnpkg.com/downloads/$YARN_VERSION/yarn-v$YARN_VERSION.tar.gz" \
-  && curl -fSLO --compressed "https://yarnpkg.com/downloads/$YARN_VERSION/yarn-v$YARN_VERSION.tar.gz.asc" \
-  && gpg --batch --verify yarn-v$YARN_VERSION.tar.gz.asc yarn-v$YARN_VERSION.tar.gz \
-  && mkdir -p /opt/yarn \
-  && tar -xzf yarn-v$YARN_VERSION.tar.gz -C /opt/yarn --strip-components=1 \
-  && ln -s /opt/yarn/bin/yarn /usr/local/bin/yarn \
-  && ln -s /opt/yarn/bin/yarn /usr/local/bin/yarnpkg \
-  && rm yarn-v$YARN_VERSION.tar.gz.asc yarn-v$YARN_VERSION.tar.gz
+RUN apt-get install -y psmisc
```
PippoRaimondi: Maybe leave a comment with the reasoning we need this installed? i.e `necessary for the application to run` or something like that

Dockerfile
currycoder: This combines some ENV vars provided in the base image with the established path pattern that NVM uses - as far as I can see looking at the dockerfile for the base image, there's not a full path to the node library provided in there.

Dockerfile
currycoder: Good point, added some detail here

PR #1803, opened 2019-03-11
Add multiple advisers to interaction forms

ian-leggett:
## Problem
We need to add multiple advisers to interactions.
Trello 
-  https://trello.com/c/aMohbehV/590-add-typeahead-to-add-another-pattern
- https://trello.com/c/aMohbehV/590-add-typeahead-to-add-another-pattern

Screenshot 2019-03-13 at 10 51 33

## Solution
Use "add another" pattern with the typeahead component so users can search for new advisers and add multiple advisers.
Screenshot 2019-03-13 at 10 52 02

PR #1831, opened 2019-03-25
Remove unnecessary input field on add another pattern

ian-leggett:
## Problem
As part of #1803 I am removing an input field that we don't need in the add another pattern as it just creates unnecessary space on the page. I have spoken with other teams and they agree we should remove it.
Screenshot 2019-03-25 at 09 44 18

## Solution
Remove the input field.
Screenshot 2019-03-25 at 09 46 45

### DIFF
--- a/assets/javascripts/vue/typeahead.vue
+++ b/assets/javascripts/vue/typeahead.vue
@@ -162,6 +162,7 @@
         this.setPlaceHolder = this.placeholder
       },
       getLabelFromValue: function (value, model) {
+        if(!value){ return }
         const activeValue = JSON.parse(model).filter((item) => {
           return item.value === value
         })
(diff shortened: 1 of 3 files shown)


assets/javascripts/vue/typeahead.vue
```diff
@@ -162,6 +162,7 @@
         this.setPlaceHolder = this.placeholder
       },
       getLabelFromValue: function (value, model) {
```
yozzo: in ES6 you can assign a default value where you declare the argument `function (value = '', model, model={})`

assets/javascripts/vue/typeahead.vue
```diff
@@ -162,6 +162,7 @@
...
+        if(!value){ return }
```
zakdit: Not anything big but in single line if statements you don't need the parenthesis, you can just do  `if(!value) return `

zakdit: Also consider doing `if(!value || !model) return` or is `model` always being passed through?

assets/javascripts/vue/typeahead.vue
ian-leggett: Good point let me update this

assets/javascripts/vue/typeahead.vue
ian-leggett: Thanks Zak looking at this now

PR #1837, opened 2019-03-26
Remove new layout flag check for investment apps

yeahfro:
https://trello.com/c/Ap3nl8af/917-remove-companies-new-layout-flag

## Change
Removes the check for the `companies-new-layout` flag as it is now safe to display the new layout for all companies without checking the flag or `duns_number`.

### DIFF
--- a/src/apps/companies/apps/investments/growth-capital-profile/controllers/list.js
+++ b/src/apps/companies/apps/investments/growth-capital-profile/controllers/list.js
@@ -1,11 +1,5 @@
-const list = 'companies/apps/investments/growth-capital-profile/views/list'
-
-const renderGrowthCapitalProfile = (req, res, next) => {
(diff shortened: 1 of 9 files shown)


test/unit/apps/companies/controllers/investments/large-capital-profile.js
```diff
@@ -45,40 +44,17 @@ describe('Company Investments - large capital profile', () => {
         )
       })
 
-      commonTests('companies/apps/investments/large-capital-profile/views/list-deprecated', companyMock, undefined)
+      commonTests(undefined)
```
ian-leggett: If this is removed the tests still pass, what do we need this line for?

ian-leggett: This runs 4 times, should we just run it once?

PR #1844, opened 2019-03-26
Improved autocomplete advisor selection performance

zakdit:
[616](https://trello.com/c/OhMVI2uW/616)

Autocomplete triggers a load of all metadata on advisor selection in the interactions section.

This has been refactored and tests have been added reducing the time it takes every request to be completed 

**Checklist**

- [x] Does PR title include the Trello ticket id and short description of the change?
- [x] Has the code be properly tested? (unit/functional/integration/E2E)
- [x] Is the implementation of this change clear and simple?

### DIFF
--- a/src/apps/interactions/controllers/list.js
+++ b/src/apps/interactions/controllers/list.js
@@ -1,7 +1,10 @@
 const qs = require('querystring')
 const { get } = require('lodash')
 const { collectionFilterFields } = require('../macros')
-const { buildSelectedFiltersSummary, buildFieldsWithSelectedEntities } = require('../../builders')
+const {
+  buildSelectedFiltersSummary,
+  buildFieldsWithSelectedEntities,
+} = require('../../builders')
 const { getOptions } = require('../../../lib/options')
 const { buildExportAction } = require('../../../lib/export-helper')
 const { getAdvisers } = require('../../adviser/repos')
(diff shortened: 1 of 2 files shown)


test/unit/apps/interactions/controllers/list.test.js
```diff
@@ -195,4 +210,49 @@ describe('interaction list', () => {
       })
     })
   })
+
+  context('getInteractionOptions()', () => {
```
yeahfro: Here I would go with `#getInteractionOptions` to be consistent with other tests

test/unit/apps/interactions/controllers/list.test.js
```diff
@@ -195,4 +210,49 @@ describe('interaction list', () => {
...
+      ],
+      teamOptions: [
+        { value: 'te1', label: 'te1' },
+        { value: 'te2', label: 'te2' },
+        { value: 'te', label: 'te3' },
+      ],
+      adviserOptions: [{ value: 'ad1', label: 'ad1', subLabel: 'ad1' }],
+      types: [{ value: '1', label: 'pt1' }, { value: '3', label: 'pt3' }],
+    }
+    it(`return all interaction options if not XHR request`, async () => {
```
yeahfro: Here you have three scenarios which would be better placed inside `context`

ie 
`context('when the request is XHR', () => ...`
`context('when the request is not XHR', () => ...`
`context('when the XHR is not set', () => ...`

In addition, if you look at some of the other unit tests patterns we have tended towards `beforeEach` within each `context` and then keeping the actual test, `it` as a single liner. This is not always the case but has a been a good general rule.

zakdit: Applied the suggested changes, however I think in this context, adding `beforeEach` might be unnecessary. Let me know your thoughts, thanks.

src/apps/interactions/controllers/list.js
```diff
@@ -23,7 +23,7 @@ const exportOptions = {
 }
 
 async function getInteractionOptions (token, req, res) {
-  const currentAdviser = get(res.locals, 'interaction.dit_adviser.id')
+  if (get(req, 'xhr')) return
```
yeahfro: `get` not really required

PR #1845, opened 2019-03-27
Integrate v4 company endpoint and formatAddress

yeahfro:
https://trello.com/c/urWUCYnD/884-consolidate-formatting-of-addresses

## Change
This change is to integrate the new `formatAddress` Nunjucks filter. As part of this change we will not be using the new `/v4/company/{id}` endpoint. I am temporarily adding the `registered_address_x` and `trading_address_x` fields to be backwards compatible with existing code. Once these fields are no longer depended on, I can remove the temporary addition.

There should be no difference in visual appearance.

### DIFF
--- a/src/apps/companies/repos.js
+++ b/src/apps/companies/repos.js
@@ -1,10 +1,30 @@
 /* eslint camelcase: 0, prefer-promise-reject-errors: 0 */
+const { get } = require('lodash')
+
 const config = require('../../../config')
 const { authorisedRequest } = require('../../lib/authorised-request')
 
-// Get a company and then pad out the interactions with related data
+// TODO remove this when no dependencies on company.registered_address_* and company.trading_address_*
+const mapDeprecatedAddressFields = (companyAddress, deprecatedAddressPrefix) => {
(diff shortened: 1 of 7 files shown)


src/apps/companies/repos.js
```diff
@@ -1,10 +1,30 @@
...
+    [`${deprecatedAddressPrefix}_address_town`]: get(companyAddress, 'town'),
+    [`${deprecatedAddressPrefix}_address_county`]: get(companyAddress, 'county'),
+    [`${deprecatedAddressPrefix}_address_postcode`]: get(companyAddress, 'postcode'),
+    [`${deprecatedAddressPrefix}_address_country`]: get(companyAddress, 'country'),
+  }
+}
+
 function getDitCompany (token, id) {
-  return authorisedRequest(token, `${config.apiRoot}/v3/company/${id}`)
+  return authorisedRequest(token, `${config.apiRoot}/v4/company/${id}`)
```
ian-leggett: Just a thought could the 'v4' come from a config?

PR #1852, opened 2019-03-29
Migrate collection features to cypress

PippoRaimondi:


test/functional/cypress/selectors/filter.js
```diff
@@ -0,0 +1,12 @@
+module.exports = {
```
yeahfro: These would read slightly better and be more consistent as:

```
module.exports = {
  name: '#field-name',
  statusActive: '[for="field-archived-1"]',
  statusInactive: '[for="field-archived-2"]',
  firstCountry: '[for="field-country-1"]',
  firstUkRegion: '[for="field-uk_region-1"]',
  contacts: {
    companyName: '#field-company_name',
    firstCountry: '[for="field-address_country-1"]',
    firstUkRegion: '[for="field-company_uk_region-1"]',
  },
}
```

PippoRaimondi: > Cheers @PippoRaimondi
> 
> This looks good. My main thought is what's the difference between `entity-list` and `entity-collection` selectors? Could `entity-collection.js` have a `firstEntity` field?

entity is a child of the collection, but I don't see both selectors growing too much so they could be merged!

PR #1859, opened 2019-04-02
Use lodash get for nullable country on address

yeahfro:
https://trello.com/c/pUVVgjT9/1003-bug-fix-formataddress-as-country-is-nullable

## Problem
It is possible for the country to be `null` on the `address` object. As a result, there are template render errors when searching or filtering companies.

e.g. https://sentry.ci.uktrade.io/dit/front-end/issues/8572

## Solution
Use `lodash` `get` to retrieve the company name. If there is no value then the company address will be presented without a country.

### DIFF
--- a/config/nunjucks/filters.js
+++ b/config/nunjucks/filters.js
@@ -12,6 +12,7 @@ const {
   compact,
   concat,
   escape,
+  get,
   isArray,
   isFunction,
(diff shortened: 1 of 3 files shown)


yeahfro:
Closed as this is being fixed at the source by fixing the data

PR #1862, opened 2019-04-03
Disable tests until issue is fixed

PippoRaimondi:


ian-leggett:
> I think the issue itself should be prioritised and fixed, rather than switching off tests _until_ it's fixed.

The problem here is that the issue has always been there, its just these tests exposed it.

PippoRaimondi:
> I think the issue itself should be prioritised and fixed, rather than switching off tests _until_ it's fixed.

We will discuss the issue in standup and prioritise it!

PR #1868, opened 2019-04-04
Change Abandoned Proposition details logic to distinguish urls from plain string and format accordingly

yozzo:
Users seem to add both urls and explanatory text to the details section when abandoning a proposition and they expect to see the formatting accordingly.

Create an html type for key-value-table template for rendering html strings

Screenshot 2019-04-04 at 15 33 59

### DIFF
--- a/src/apps/propositions/transformers.js
+++ b/src/apps/propositions/transformers.js
@@ -87,7 +87,7 @@ function addLineBreaks (string) {
   } else {
     return {
       type: 'word',
-      string,
+      value: string,
     }
   }
 }
@@ -110,7 +110,7 @@ function addAnchorsToUrl (s) {
 
   return {
     type: 'link',
-    string: {
+    value: {
       url: string,
       name: string,
     },
@@ -150,7 +150,7 @@ function transformPropositionResponseToViewRecord ({
     adviser,
     details: formattedDetails ? {
       type: 'paragraph',
-      string: formattedDetails,
+      value: formattedDetails,
     } : null,
     ...transformFilesResultsToDetails(files.results, id, investment_project.id),
   }
--- a/src/apps/propositions/transformers.js
+++ b/src/apps/propositions/transformers.js
@@ -1,5 +1,5 @@
 /* eslint-disable camelcase */
-const { assign, capitalize, compact, get, mapKeys, pickBy } = require('lodash')
+const { assign, capitalize, get, mapKeys, pickBy } = require('lodash')
 const { format, isValid } = require('date-fns')
 
 const { transformDateObjectToDateString } = require('../transformers')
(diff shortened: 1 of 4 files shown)


src/templates/_components/key-value-table.njk
```diff
@@ -60,6 +60,8 @@
     {{ data.address | formatAddress }}
   {% elif data.type === 'error' %}
     {{ error(data) }}
+  {% elif data.type === 'html' %}
```
yeahfro: My suggestion is potentially adding a type `paragraph`. This potentially opens the dates for a bunch of HTML to be transformed on the server.

yeahfro: Whatever is changed here needs a corresponding unit test

test/unit/apps/propositions/transformers.test.js
```diff
@@ -112,11 +112,8 @@ describe('Proposition transformers', () => {
       })
 
       it('should not transform it to a url', () => {
-        expect(this.transformed.Details).to.deep.equal('the world is a stage')
-        expect(this.transformed.Details).to.not.deep.equal({
-          name: 'http://the-world-is-a-stage',
-          url: 'http://the-world-is-a-stage',
-        })
+        expect(this.transformed.Details.string).to.equal('the world is a stage')
+        expect(this.transformed.Details.string).to.not.equal('<a href="http://the-world-is-a-stage">http://the-world-is-a-stage</a>')
```
yeahfro: The negative test on this line is implicit based upon the previous line passing

test/unit/apps/propositions/transformers.test.js
```diff
@@ -143,28 +163,61 @@ describe('Proposition transformers', () => {
...
+        this.transformed = transformPropositionResponseToViewRecord(mockProposition)
+      })
+
+      it('should transform the url to an anchor thus ignoring the normal text', () => {
+        expect(this.transformed.Details.string).to.equal('the world <a href="https://not.your.business.com/in-oui/Tower/BMI-slender/Heroes/Alpacas/AllTheThings.bruv?CarrotFruit=%2Bounce-ti%2Tennis%2FWD-atlethics%2BStuff%2Travel%2More%2L3ss%tim3ws6erz%20-%20Blah%2Cu8cumb34&Rap_Music=0x012000D998E6715Fr0GPF4881B0002C5764B8EF&EyeEye=%7B08EEDE07-DOOH-4B91-ASIA-BOYBAND48766CE%7D">https://not.your.business.com/in-oui/Tower/BMI-slender/Heroes/Alpacas/AllTheThings.bruv?CarrotFruit=%2Bounce-ti%2Tennis%2FWD-atlethics%2BStuff%2Travel%2More%2L3ss%tim3ws6erz%20-%20Blah%2Cu8cumb34&Rap_Music=0x012000D998E6715Fr0GPF4881B0002C5764B8EF&EyeEye=%7B08EEDE07-DOOH-4B91-ASIA-BOYBAND48766CE%7D</a> <br> The World is a stage and Susan from Accounts piloted the revolving bacon')
```
yeahfro: This is quite hairy, can the same be achieved with less?

yozzo: this is the kind of data users put in, I think it makes sense the have a test for it.

src/apps/propositions/transformers.js
```diff
@@ -73,6 +73,37 @@ function transformPropositionToListItem ({
   }
 }
 
+function formatDetails (details = '') {
```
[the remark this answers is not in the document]
yozzo: yeah, that makes sense - because we already have the array, and the strings are already formatted, instead of adding the tags here could store strings in a collection with objects with types `paragraph`, `url` or `line-break`  and add the according tags in the template.

src/apps/propositions/transformers.js
```diff
@@ -73,6 +71,116 @@ function transformPropositionToListItem ({
...
+
+/**
+ * Returns a list of indexes where the string matches the url pattern (http, www, ftp)
+ * @param list
+ * @returns {Array}
+ */
+function getUrlListIndexes (list = []) {
+  const indexes = []
+
+  forEach(list,
```
PanosParalakis: We are using lodash a lot, but personally I would have used native forEach :)

PR #1873, opened 2019-04-07
Added an Investor type select to the Investor details section of LCP

paulgain:
* Added a new GOV.UK select component to the Investor details edit
section enabling the user to select one of 12 investor types within
Large capital profile (LCP).

* Added a label to the profile page that matches the users selection
showing what they previously selected.

* Adding a PATCH API call to update the Company profile.

* Ensured the select component scales to 75% of the width on tablet and
desktop but takes the full 100% width when on mobile.

* Adding functional tests

* Adding unit tests with mocked data.

https://uktrade.atlassian.net/browse/IPBETA-270

**Before**

before-select

**After**

after-select

**Before**

before-label

**After**
after-label


src/apps/companies/apps/investments/large-capital-profile/controllers/profile.js
```diff
@@ -1,17 +1,38 @@
...
+  const { editing } = req.query
 
   try {
-    const profiles = await getCompanyProfiles(token, company.id)
-    const profile = profiles.results && profiles.results[0]
-    data.profile = profile ? transformProfile(profile) : data.profile
-    res.render('companies/apps/investments/large-capital-profile/views/profile', data)
+    const profile = await getCompanyProfile(token, company, editing)
+
+    if (profile) {
```
rafenden: This could be wrapped in a single `if` statement, e.g.
```
if (profile && profile.editing) {
```

paulgain: HI @rafenden thanks for the review it's much appreciated. 

The string `profile.editing` can be any one of these guys and represents the section the user is editing: 

https://github.com/uktrade/data-hub-frontend/blob/d6d3e411da2b4ee24bea467e53a59f032833f2d5/src/apps/companies/apps/investments/large-capital-profile/sections.js

Future pull requests the code will be expanded to look like the following which makes it a lot easier to read:
```
if (profile) {
  if (profile.editing === INVESTOR_DETAILS) {
    profile.investorDetails.investorType.items = await getInvestorTypes(token, profile)
     ...
  } else if (profile.editing === INVESTOR_REQUIREMENTS) {
      ...
  } else {
      ... // LOCATION
  }
}
```
It makes sense to add the additional check so we do not enter the outer `if` block when `profile.editing` is `undefined`, however, we still need the inner `if` statements when `profile.editing` is defined due to the 3 types.

```
if (profile && profile.editing) {
  ...
}
```

**I guess there are some drawbacks to smaller PRs.**

PR #1876, opened 2019-04-08
Hide the company name within the list of company interactions

rafenden:
## Motive

https://trello.com/c/7cporIpn/691-do-not-show-company-on-interactions-page

> Company meta item should not show for interactions collection when browsing from a company. When browsing to Interactions from the global navigation Company should show as it's a list of interactions across all companies.

## Solution

This PR is removing the **Company** field and the sorting option from the list of interactions on the company page.

It is done by introducing a new boolean parameter `showCompany` passed to `res.locals.interactions`.

Interactions - Companies - DIT Data Hub 2019-04-08 14-07-23

Interactions - Companies - DIT Data Hub 2019-04-08 14-08-55

### DIFF
--- a/src/apps/interactions/transformers/interaction-to-list-item.js
+++ b/src/apps/interactions/transformers/interaction-to-list-item.js
@@ -1,61 +1,50 @@
 /* eslint-disable camelcase */
-const { get } = require('lodash')
+const { compact, get } = require('lodash')
 
 const { INTERACTION_NAMES } = require('../constants')
-const { interaction } = require('../labels')
+const labels = require('../labels')
 
-function transformInteractionToListItem ({
-  id,
-  subject,
-  kind,
-  contacts,
-  company,
-  date,
-  dit_participants,
-  dit_team,
-  service,
-  was_policy_feedback_provided,
-}) {
-  const hasMultipleContacts = (contacts) => contacts.length > 1 ? interaction.multiple_contacts : contacts.map(contact => contact.name)
-  const hasTeam = (name) => get(name, 'team') ? `${name.adviser.name}, ${name.team.name}` : name.adviser.name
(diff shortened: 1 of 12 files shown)


src/apps/interactions/transformers/interaction-to-list-item.js
```diff
@@ -1,61 +1,55 @@
...
-  service,
-  was_policy_feedback_provided,
-}) {
-  const hasMultipleContacts = (contacts) => contacts.length > 1 ? interaction.multiple_contacts : contacts.map(contact => contact.name)
-  const hasTeam = (name) => get(name, 'team') ? `${name.adviser.name}, ${name.team.name}` : name.adviser.name
-  const hasMultipleAdvisers = (dit_participants) => dit_participants.length > 1 ? interaction.multiple_advisers : dit_participants.map(ditParticipant => hasTeam(ditParticipant))
-  return {
-    was_policy_feedback_provided,
+function transformInteractionToListItem (showCompany = true) {
+  return function ({
```
ian-leggett: I guess my only comment is why we don't return an object straight up like we do on `src/apps/interactions/transformers/interaction-response-to-view.js` or `src/apps/interactions/transformers/interaction-response-to-form.js` no problem for me. @yeahfro what do you think?

yeahfro: Cheers @ian-leggett 

I am comfortable with this as it's a Higher Order Function being passed to `transformApiResponseToCollection` and we know the value of `showCompany` when `transformApiResponseToCollection` is being called in the middleware. Noted that this is a slight shift in the current pattern.

ian-leggett: Thanks @yeahfro for your input. Its good from me 👍

PR #1879, opened 2019-04-08
Evaluate company.address to get company address when creating investment project

yeahfro:
https://trello.com/c/X9ob0P2O/1015-use-new-address-object-when-creating-investment

## Change
Removes dependency on `registered_address_*` when creating an investment project to display company information.

There should be no visual change on the FE.

### DIFF
--- a/src/apps/investments/views/create/equity-source.njk
+++ b/src/apps/investments/views/create/equity-source.njk
@@ -13,7 +13,7 @@
         <td>{{ company.name }}</td>
       </tr>
       <tr>
-        <td>{{ company.registered_address_country.name }}</td>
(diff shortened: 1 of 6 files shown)


src/apps/investments/views/create/equity-source.njk
```diff
@@ -13,7 +13,7 @@
         <td>{{ company.name }}</td>
       </tr>
       <tr>
-        <td>{{ company.registered_address_country.name }}</td>
+        <td>{{ company.address.country.name }}</td>
```
rafenden: I believe the registered address is in `company.registered_address` property.

PR #1880, opened 2019-04-09
Adding three new fields into Investor details, Large capital profile

paulgain:
The three new fields are:

1. Global assets under management
2. Investable capital
3. Investor description

https://uktrade.atlassian.net/browse/IPBETA-270

### Three new fields

three-new-fields

### Three new labels

three-new-labels


yeahfro:
Just a small comment is that the commit message could read as `Add...` rather than `Adding...` ie it is the imperative, giving a command

PR #1882, opened 2019-04-10
Hotfix/filter labels

zakdit:
##Problem 
Fixing the performance on interaction filters introduced an unintended consequence of selecting a filter and having it show the id instead of the filter name.

##Solution
To fix it, the meta data on load needs to be cached in a session key and retrieved on XHR requests.


rafenden:
I believe that hot-fixes should target the `master` branch, not `develop` as suggested by the git-flow strategy.

This way you can avoid releasing changes unrelated to the actual fix.

PR #1884, opened 2019-04-10
Retrieve node modules from cache instead of installing

PippoRaimondi:


package.json
```diff
@@ -128,7 +128,7 @@
     "chai": "^4.1.2",
     "chai-as-promised": "^7.1.1",
     "chai-subset": "^1.6.0",
-    "chromedriver": "^2.44.1",
+    "chromedriver": "^2.46.0",
```
rafenden: Looks like there is a missing change in `yarn.lock` file.

PR #1890, opened 2019-04-12
Use new address objects when transforming address for edit company display

rafenden:
https://trello.com/c/jU3otZrf/1017-use-new-address-objects-when-transforming-address-for-edit-company-display

**Implementation notes**

This change is removing dependency on the old `trading_address` and `registered_address` fields from the company v3 endpoint. 

Note that adding and editing a company is affected with this change and therefore should be given special attention when manually testing.

**Checklist**

### DIFF
--- a/src/apps/companies/repos.js
+++ b/src/apps/companies/repos.js
@@ -28,7 +28,7 @@ function getDitCompany (token, id) {
 }
 
 function getCHCompany (token, id) {
-  return authorisedRequest(token, `${config.apiRoot}/v3/ch-company/${id}`)
+  return authorisedRequest(token, `${config.apiRoot}/v4/ch-company/${id}`)
 }
 
 function saveCompany (token, company) {
--- a/src/apps/companies/transformers/company-to-form.js
+++ b/src/apps/companies/transformers/company-to-form.js
@@ -1,46 +1,59 @@
 /* eslint-disable camelcase */
-const {
-  assign,
-  get,
-  isPlainObject,
-  mapValues,
-} = require('lodash')
+const { get } = require('lodash')
 
-// TODO: This is a temporary transformer to transform an API response into
-// a format needed for the form view
-//
-// Will be replaced with newer form builder transformers once the form view is
-// replaced with newer form macros
-module.exports = function transformCompanyToForm (company) {
-  if (!isPlainObject(company)) { return }
+const ADDRESS_TYPE = {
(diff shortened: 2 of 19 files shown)


src/apps/companies/repos.js
```diff
@@ -1,30 +1,10 @@
 /* eslint camelcase: 0, prefer-promise-reject-errors: 0 */
-const { get } = require('lodash')
 
 const config = require('../../../config')
 const { authorisedRequest } = require('../../lib/authorised-request')
 
-// TODO remove this when no dependencies on company.registered_address_* and company.trading_address_*
```
yeahfro: We are probably not ready to do this as there are still dependencies on these fields, ie OMIS.

When we are ready there is [a ticket](https://trello.com/c/ImTjSMyT/1020-remove-augmentation-of-legacy-address-fields-in-getditcompany) to remove this.

src/apps/companies/transformers/company-to-form.js
```diff
@@ -1,39 +1,53 @@
 /* eslint-disable camelcase */
```
yeahfro: We can remove the vast majority of this code as we are only dependent on `sector` and `headquarter_type`

test/unit/apps/companies/transformers/company-to-form.test.js
```diff
@@ -0,0 +1,80 @@
+const datahubOnlyCompany = require('~/test/unit/data/companies/datahub-only-company.json')
```
yeahfro: 👍

test/unit/apps/companies/transformers/company-to-view.test.js
```diff
@@ -12,9 +12,9 @@ describe('transformCompanyToView', () => {
     it('should contain just the fields expected', () => {
```
yeahfro: Most of these tests can be removed

src/apps/companies/transformers/company-to-form.js
rafenden: Apparently, we are using all of these fields, see `edit.njk`

Some of the fields like `formData.name`, `formData.company_number`, are read-only so it might be a good idea to get them from `res.locals.company`?

yeahfro: You're right, I meant to write this comment against `company-to-view.js`

PR #1901, opened 2019-04-24
Feature/homepage redesign

zakdit:
**Implementation notes**
https://trello.com/c/iums91AN/660-add-feed-of-latest-zendesk-help-centre-articles-to-home-page
New homepage design which includes a Zen desk article feed

**Checklist**

### DIFF
--- a/src/apps/dashboard/controllers.js
+++ b/src/apps/dashboard/controllers.js
@@ -1,25 +1,47 @@
 const { get } = require('lodash')
+const rp = require('request-promise')
+const moment = require('moment')
 
 const GLOBAL_NAV_ITEMS = require('../global-nav-items')
 
 const { isPermittedRoute } = require('../middleware')
 const { fetchHomepageData } = require('./repos')
+const config = require('../../../config')
 
 async function renderDashboard (req, res, next) {
   try {
     const userPermissions = get(res, 'locals.user.permissions')
-    const {
+    const { contacts, interactions } = await fetchHomepageData(
+      req.session.token
+    )
+
+    const articleFeed = await rp({
+      uri: config.zen.announcementsURL,
+      json: true,
+      timeout: 1000,
+    })
+      .then(feed =>
+        feed.articles.map(item => ({
+          heading: item.title,
+          link: item.html_url,
+          date: moment(item.created_at)
+            .fromNow(),
+        }))
+      )
+      .catch(err => {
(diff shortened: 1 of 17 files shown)


test/unit/components/info-feed.test.js
```diff
@@ -0,0 +1,98 @@
...
+      this.component = renderComponentToDom('info-feed', data)
+      this.heading = this.component.querySelector('.dashboard-section-title')
+    })
+    it('show heading', () => {
+      expect(this.heading.textContent).to.equal(data.heading)
+    })
+  })
+
+  describe('Render correct sub-heading', () => {
+    beforeEach(() => {
```
ian-leggett: I think we could move the component to a higher level and then use one `beforeEach` against all the assertions as we only ever need the one component to test against. Removes dups

src/templates/_components/data-table.njk
```diff
@@ -44,9 +46,11 @@
     </thead>
     <tbody>
       {% for record in data %}
-        <tr>
+        <tr {% if record.outboundURL %} onclick="location.href='{{ record.outboundURL }}'" style="cursor:pointer"{% endif %}>
```
ian-leggett: This will be an issue without JS also should we wrap the cell values in links instead. 🤔

src/apps/dashboard/controllers.js
```diff
@@ -1,25 +1,47 @@
...
+      req.session.token
+    )
 
-    res
-      .title('Dashboard')
-      .render('dashboard/views/dashboard', {
-        contacts,
-        interactions,
-        interactionsPermitted: isPermittedRoute('/interactions', GLOBAL_NAV_ITEMS, userPermissions),
+    const articleFeed = await rp({
```
ian-leggett: I would update the test file for this '/dashboard/controllers.test.js'

test/unit/components/info-feed.test.js
zakdit: Done 👍

src/templates/_components/data-table.njk
zakdit: removed and introduced anchor solution as per discussion 👍

src/apps/dashboard/controllers.js
zakdit: moved out code to a new transformers file which includes it's own tests

PR #1902, opened 2019-04-24
Add Gross Value Added (GVA)

paulgain:
Gross Value Added or GVA is a calculation performed via the API. The
value is calculated providing all of these conditions are true:
1. The investment is FDI.
2. The Foreign Equity Investment field has been defined by the user.
3. The Primary Sector field has been defined by the user.

The user is shown a helpful message when either or both 2 & 3 are false.

https://uktrade.atlassian.net/browse/IPBETA-329

**The user needs to complete both fields before GVA can be calculated**

add-both-foreign-equity-value-and-primary-sector

**The user needs to complete "Foreign equity investment" before GVA can be calculated**

add-foreign-equity-investment-value

**The user needs to complete the "Primary sector" before GVA can be calculated**
add-primary-sector

**GVA has been calculated**
gva

gva-read-only

**Checklist**

### DIFF
--- a/src/apps/investments/transformers/value.js
+++ b/src/apps/investments/transformers/value.js
@@ -17,11 +17,23 @@ function transformInvestmentAmount (clientCannotProvideInvestment, investmentAmo
   }
 }
 
+function transformGrossValueAdded (grossValueAdded) {
+  if (!grossValueAdded) {
+    return null
+  }
+
+  return {
+    type: 'currency',
+    name: grossValueAdded,
+  }
+}
+
 function transformInvestmentValueForView ({
   client_cannot_provide_total_investment,
   total_investment,
   client_cannot_provide_foreign_investment,
   foreign_equity_investment,
+  gross_value_added,
   number_new_jobs,
   number_safeguarded_jobs,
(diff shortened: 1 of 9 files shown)


src/apps/investments/transformers/value.js
```diff
@@ -17,11 +17,23 @@ function transformInvestmentAmount (clientCannotProvideInvestment, investmentAmo
   }
 }
 
+function transformGrossValueAdded (grossValueAdded) {
```
ian-leggett: Any reason why we are not using arrow function here? I know some areas we are not due to `this` context etc

paulgain: The entire file defines functions in this way. I simply followed suite for consistency.

paulgain:
> We have a new condition around `src/apps/investments/views/value-edit.njk` should we get a test around this making sure it renders the new HTML when this condition is met?

This is temporary as the lot is coming out with all the new work we are doing. I've created solid unit tests around the new functionality. I'm unsure if it's wise to spend more time on something that eventually will be ripped out. What do you think?

PR #1927, opened 2019-05-07
Bug - filter active inactive advisers

ian-leggett:
Screenshot 2019-05-02 at 09 37 47

## Problem
Only the adviser filters should return inactive advisers, currently this behaviour has been applied globally.

Trello - https://trello.com/c/vbEYJKwl/720-forms-should-only-search-active-advisers-when-adding-advisers

## Solution
Set a global option for the typeahead to only return active advisers then add the option to return inactive advisers for the filter only.

## Manual test
1. Open Django and find an adviser that is inactive
2. Go on the interactions page and search for this user and you should find him/her
3. Now go to a form page(add interaction) that uses the adviser search and you shouldn't be able to see the inactive adviser when you search.

### DIFF
--- a/assets/javascripts/vue/typeahead.vue
+++ b/assets/javascripts/vue/typeahead.vue
@@ -143,6 +143,11 @@
         type: Boolean,
         required: false,
         default: true,
+      },
+      hideInactive: {
+        type: Boolean,
+        required: false,
+        default: true,
       }
     },
     created () {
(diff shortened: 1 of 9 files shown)


PanosParalakis:
LGTM 👍

paulgain:
Looks good @ian-leggett  
At some point in the future you should increase the coverage for repos.js within Interactions: https://43995-83823675-gh.circle-artifacts.com/0/root/data-hub-frontend/coverage/lcov-report/src/apps/interactions/index.html

PR #1937, opened 2019-05-10
Add nvmrc file to project

PanosParalakis:
**Implementation notes**
I have added nvmrc file for those of us who are using [nvm](https://github.com/nvm-sh/nvm) and I have updated readme file.

**Checklist**


README.md
```diff
@@ -240,6 +240,12 @@ Engines in `package.json` needs updating:
   },
 ```
 
+#### Version inside .nvmrc
+Default node version for `nvm` is defined inside .nvmrc. Update by running:
+```
+  echo "v8.15.1"  > .nvmrc
+```
```
PanosParalakis: Here is the new addition

PanosParalakis:
Thank you @yeahfro! You are right, unfortunately my editor auto strips white space, so it is difficult to locate the change. I have added a comment pointing to the actual change. I would prefer to keep the my editor setting for trimming white spaces on. I do not use any other formatting tools atm but maybe it is something that we can think about. Something like prettier or js-beautify.
 
> There's quite a few line changes here which detracts away from the actual change. My recommendation would be to isolate the change in a single commit. If you want to update formatting for everything else then that would make sense as a separate commit.

yeahfro:
> Thank you @yeahfro! You are right, unfortunately my editor auto strips white space, so it is difficult to locate the change. I have added a comment pointing to the actual change. I would prefer to keep the my editor setting for trimming white spaces on. I do not use any other formatting tools atm but maybe it is something that we can think about. Something like prettier or js-beautify.
> 
> > There's quite a few line changes here which detracts away from the actual change. My recommendation would be to isolate the change in a single commit. If you want to update formatting for everything else then that would make sense as a separate commit.

Okay, so please have two commits:
1. The specific lines with the change
2. White space changes

PanosParalakis:
I have create a new PR regarding: [1939](https://github.com/uktrade/data-hub-frontend/pull/1939)
so I am closing this now

PR #1969, opened 2019-05-24
Add 'Minimum equity percentage' radio buttons for Large capital profile  

paulgain:
Add 4 new radio buttons for 'Minimum equity percentage'

1. 0% - Not required
2. 1-19%
3. 20-49%
4. 50% +

Add a single label within the read-only view showing the users selection

https://uktrade.atlassian.net/browse/IPBETA-272

** Minimum equity percentage - edit**

equity-percentage-edit

** Minimum equity percentage - read-only**

equity-percentage-read-only

**Checklist**

### DIFF
--- a/src/apps/companies/apps/investments/utils/transformers.js
+++ b/src/apps/companies/apps/investments/utils/transformers.js
@@ -2,13 +2,23 @@ const { find, isString, castArray, compact } = require('lodash')
 
 const transformObjectToOption = ({ value, label }) => ({ value, text: label })
 
-const checkMatchingItemById = (items) => {
-  return obj => {
-    if (find(items, (item) => item.id === obj.value)) {
-      obj.checked = true
+const checkOptionByFindingMatchingId = (items) => {
+  return option => {
+    if (find(items, (item) => item.id === option.value)) {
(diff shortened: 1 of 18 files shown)


src/apps/companies/apps/investments/utils/transformers.js
```diff
@@ -2,13 +2,23 @@ const { find, isString, castArray, compact } = require('lodash')
...
+    if (find(items, (item) => item.id === option.value)) {
+      option.checked = true
     }
 
-    return obj
+    return option
+  }
+}
+
+const checkOptionByMatchingId = (id) => {
```
yeahfro: Is it possible to do the following or is `false` not good?

```
const checkOptionByMatchingId = (id) => {
  return option => {
    option.checked = option.value === id

    return option
  }
}

paulgain: I tried that during development as it simplifies the code but it doesn't work.

paulgain: Actually that does work, I must of tried it on the checkboxes. I'll add that into the next PR.

PR #1985, opened 2019-05-31
Replace in the Node layer axios with request-promise

yozzo:
**Implementation notes**

Due to a malfunction to the axios library, we decided to replace it with require on the node layer as only one instance was used.

**Checklist**

### DIFF
--- a/src/apps/oauth/controllers.js
+++ b/src/apps/oauth/controllers.js
@@ -1,7 +1,6 @@
 const request = require('request-promise')
 const queryString = require('qs')
 const uuid = require('uuid')
-const axios = require('axios')
 
 const { get, set, isUndefined } = require('lodash')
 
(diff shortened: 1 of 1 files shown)


src/apps/oauth/controllers.js
```diff
@@ -37,16 +36,17 @@ function handleMissingState (req, res, next) {
...
-    options,
-  ).then((response) => response.data)
-    .catch((error) => error)
-
-  return data
+  try {
+    const response = await request(options)
+    return response
+  } catch (error) {
+    return error
```
yeahfro: Are we definitely okay with this `try .. catch` here? ie if there is a problem then the consuming function will cache the error and it will carry on. Is that intended behaviour?

src/apps/oauth/controllers.js
```diff
@@ -37,16 +36,17 @@ function handleMissingState (req, res, next) {
...
-  const data = await axios.get(
-    config.oauth.userProfileUrl,
-    options,
-  ).then((response) => response.data)
-    .catch((error) => error)
-
-  return data
+  try {
+    const response = await request(options)
+    return response
```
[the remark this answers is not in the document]
yozzo: acceptance tests will fail though as it relies on mock-sso

src/apps/oauth/controllers.js
yozzo: it’s intended behaviour as if the call fails, it shouldn’t display to the users the links to the which the user queries access to, but shouldn't prevent the UI from rendering.

PR #1992, opened 2019-06-05
Bug: Filter is still applied when unselecting it while a filter request is still being loaded.

zakdit:
Fixed a bug that caused filter checkboxes to showing incorrect tick values when rapidly clicked 2 or more times.

Controversy: This section of code is all client side and had no unit tests prior. Adding them would require quite a bit of work for such a small change as all the previous code would need to be tested and the DOM structure set up.

### DIFF
--- a/assets/javascripts/modules/auto-submit.js
+++ b/assets/javascripts/modules/auto-submit.js
@@ -11,15 +11,35 @@ const AutoSubmit = {
     this.bindEvents()
   },
 
+  isCheckboxSynced (evt) {
+    // This is needed because checkboxes can go out of sync with posted values. We prevent
+    // this by reverting the checkbox to it's previous state if there is already a submit in progress.
+    // This intentionally affects all checkboxes that trigger form submits site-wide.
+
+    if (this.isSubmitting && evt.target.type === 'checkbox') {
+      evt.target.checked = !evt.target.checked
+      return
+    }
+    return true
+  },
+
   handleFormSubmit (evt) {
+    if (!this.isCheckboxSynced(evt)) return
+
     if (evt.target.classList.contains('ie-date-field')) {
-      if (!checkDateFormat(evt.target.value)) { return }
+      if (!checkDateFormat(evt.target.value)) {
+        return
+      }
     }
     const targetForm = evt.target.closest('form')
 
-    if (!targetForm) { return }
+    if (!targetForm) {
+      return
+    }
 
-    const shouldSubmit = targetForm.classList.contains(this.selector.substring(1))
+    const shouldSubmit = targetForm.classList.contains(
+      this.selector.substring(1)
+    )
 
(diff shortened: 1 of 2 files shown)


assets/javascripts/modules/auto-submit.js
```diff
@@ -12,14 +12,29 @@ const AutoSubmit = {
   },
 
   handleFormSubmit (evt) {
+    // This is needed because checkboxes can go out of sync with posted values. We prevent
+    // this by reverting the checkbox to it's previous state if there is already a submit being completed.
+    // This intentionally affects all checkboxes that trigger form submits site-wide.
+
+    if (this.isSubmitting && evt.target.type === 'checkbox') {
```
yeahfro: Is it possible to get a test around this change?

assets/javascripts/modules/auto-submit.js
```diff
@@ -12,14 +12,29 @@ const AutoSubmit = {
...
+      evt.target.checked = !evt.target.checked
```
rafenden: This seems counterintuitive, the checkbox will deselect itself if the user clicks multiple of them quickly?

If yes, I suggest to use [_.debounce](https://lodash.com/docs/4.17.11#debounce) which essentially will cancel the first AJAX request if a second one was triggered within a specific period of time.

> Debouncing means to coalesce several temporally close signals into one signal.

See this excellent article on the topic:
https://css-tricks.com/debouncing-throttling-explained-examples/

zakdit: The problem is that even if we use debounce it wouldn't solve the problem because the checkbox still falls out of sync as we do not use AJAX responses for UI state management. 

At first glance it might seem counter-intuitive but essentially it provides the exact result we're looking for. From the UI perspective they don't see it being ticked and unticked because the event handler function changes the checkbox value before it reaches the DOM.  Instead there is a limit between how fast they are able to submit the form (i.e. only when the previous submit is completed) and the UI stays in sync with that.

assets/javascripts/modules/auto-submit.js
zakdit: test added 👍

zakdit:
> The actual change is relatively minimal so I would have been tempted to make this two commits:
> 
> 1. Refactor
> 2. The actual change
> 
> Can you add the `Bug` label?

Will keep that in mind for future commits, added bug tag :+1

assets/javascripts/modules/auto-submit.js
```diff
@@ -1,5 +1,5 @@
 const getFormData = require('get-form-data').default
-const pickBy = require('lodash/pickBy')
+const { pickBy, debounce } = require('lodash')
```
yeahfro: It's better to reference `require('lodash/pickBy')` and `require('lodash/debounce')` specifically in client-side code

PR #1999, opened 2019-06-10
Replace dit_adviser references

zakdit:
There are still some references to `interaction.dit_adviser` and `interaction.dit_team` in the FE that needed to be updated to use the `dit_participants` array.

**Checklist**

### DIFF
--- a/src/apps/adviser/filters.js
+++ b/src/apps/adviser/filters.js
@@ -1,9 +1,13 @@
-const { get } = require('lodash')
+const { get, includes } = require('lodash')
 
 function filterActiveAdvisers ({ advisers = [], includeAdviser }) {
-  return advisers.filter((adviser) => {
+  return advisers.filter(adviser => {
     // This returns all active users plus the include advisor (whether they are active or not)
-    return get(adviser, 'is_active', true) || adviser.id === includeAdviser
+    return (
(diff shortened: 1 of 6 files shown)


src/apps/adviser/filters.js
```diff
@@ -1,9 +1,13 @@
-const { get } = require('lodash')
+const { get, includes } = require('lodash')
 
 function filterActiveAdvisers ({ advisers = [], includeAdviser }) {
```
reupen: I would consider changing the signature to be something like:

```javascript
function filterActiveAdvisers (advisers, ...includeAdvisers)
```

So that you can call it using `filterActiveAdvisers(advisers, currentAdviser)` (for one adviser) or `filterActiveAdvisers(advisers, ...currentAdvisers)` (for an array of advisers).

Reason being is to simplify the logic in the function to something like:

```javascript
  return advisers.filter(adviser => {
    return get(adviser, 'is_active', true) || includes(includeAdvisers, adviser.id)
  })
```

(and avoid calling `includes()` with a string for the first parameter when there is only one adviser to explicitly include).

zakdit: Thanks, I have adjusted it to something along these lines now.

PR #2007, opened 2019-06-11
Change all buttons to match GOV.UK Design System

rafenden:
**Implementation notes**

Currently our buttons doesn't match the GOV.UK design system:
https://design-system.service.gov.uk/components/button/

This PR is meant to fix this.

https://trello.com/c/coevqkWf/1245-change-all-buttons-to-match-govuk-design-system

## Before

image

## After

image

**Checklist**

### DIFF
--- a/src/apps/components/views/form.njk
+++ b/src/apps/components/views/form.njk
@@ -4,18 +4,29 @@
   <h2 class="heading-large">Buttons</h2>
 
   {% call Example('Primary button') %}
-    <button class="button">Primary button</button>
-    <button class="button" disabled="disabled">Disabled primary button</button>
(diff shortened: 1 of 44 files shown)


src/apps/components/views/form.njk
```diff
@@ -4,18 +4,29 @@
   <h2 class="heading-large">Buttons</h2>
```
yeahfro: I wonder if we actually need these examples when you're essentially documenting GDS Design System.

rafenden: It might be still worth to remind users they can use them, could be also helpful to see how they look next to other data-hub related components.

PR #2045, opened 2019-07-04
Fix sorting on the Activity Feed

rafenden:
## Description of change

https://trello.com/c/5rX6Jygn/1372-fix-sorting-of-the-activities

Current sorting on the activity feed is only using `object.startDate` field which doesn't exists on some of the activities.

This PR is introducing a dynamic sorting which uses `object.startDate` as the default sorting field and `published` as a secondary field in case if the `startDate` doesn't exists in the document.

## Test instructions

Check if activities are correctly sorted, a good example can be found on the DEV environment:
https://www.datahub.dev.uktrade.io/companies/b2c34b41-1d5a-4b4b-9249-7c53ff2868dd/activity/data

## Checklist

### DIFF
--- a/src/apps/companies/apps/activity-feed/repos.js
+++ b/src/apps/companies/apps/activity-feed/repos.js
@@ -11,7 +11,14 @@ function fetchActivityFeed ({
     size,
     from,
     sort: {
-      'object.startTime': 'desc',
+      _script: {
+        type: 'string',
+        script: {
+          lang: 'painless',
+          source: "return !doc['object.startTime'].empty ? doc['object.startTime'].value : doc['published'].value",
+        },
+        order: 'desc',
+      },
     },
     query: {
       bool: {
--- a/.gitignore
+++ b/.gitignore
@@ -65,3 +65,6 @@ local.log
 # Yarn version policy
 .yarnrc
 .yarn/
+
+# NVM version
+.nvmrc
--- a/test/unit/apps/companies/apps/activity-feed/repos.test.js
+++ b/test/unit/apps/companies/apps/activity-feed/repos.test.js
@@ -1,6 +1,17 @@
 const config = require('~/config')
 const activityFeedRawFixture = require('~/test/unit/data/activity-feed/activity-feed-from-es')
+
 const token = 'abcd'
+const sortQuery = {
+  _script: {
+    type: 'string',
+    script: {
+      lang: 'painless',
+      source: "return !doc['object.startTime'].empty ? doc['object.startTime'].value : doc['published'].value",
+    },
+    order: 'desc',
+  },
+}
 
 describe('Activity feed repos', () => {
   beforeEach(() => {
@@ -33,7 +44,7 @@ describe('Activity feed repos', () => {
             },
           },
           size: 21,
-          sort: { 'object.startTime': 'desc' },
+          sort: sortQuery,
         }
         expect(this.authorisedRequestStub).to.be.calledOnceWith(token, {
           body: expectedBody,
@@ -55,7 +66,7 @@ describe('Activity feed repos', () => {
         const expectedBody = {
           from: 0,
           size: 20,
-          sort: { 'object.startTime': 'desc' },
+          sort: sortQuery,
           query: {
             bool: {
               filter: [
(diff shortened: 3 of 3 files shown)


src/apps/companies/apps/activity-feed/repos.js
```diff
@@ -11,7 +11,14 @@ function fetchActivityFeed ({
     size,
     from,
     sort: {
-      'object.startTime': 'desc',
+      _script: {
+        type: 'string',
+        script: {
+          lang: 'painless',
+          source: "return !doc['object.startTime'].empty ? doc['object.startTime'].value : doc['published'].value",
```
currycoder: I am a little concerned about performance here.  Unless ES does something very clever, I am assuming that the mechanism here is to carry out the search first and then sort the result set using this script sort mechanism.

This may work fine for our current needs (I doubt that the average company has more than 100 activities) but it may prove problematic/slow if we are dealing with large result sets.  What do you think?

rafenden: Here's the performance comparison:

<details>
<summary>Using script - 77ms</summary>
<pre>
{
    "from": 0,
    "size": 2000,
    "sort" : {
        "_script" : {
            "type" : "string",
            "script" : {
                "lang": "painless",
                "source": "return !doc['object.startTime'].empty ? doc['object.startTime'].value : doc['published'].value"
            },
            "order" : "desc"
        }
    },
    "query": {
        "bool": {
            "filter": [
                {
                    "terms": {
                        "object.type": [
                            "dit:Interaction",
                            "dit:ServiceDelivery",
                            "dit:InvestmentProject",
                            "dit:OMISOrder"
                        ]
                    }
                }
            ]
        }
    }
}
</pre>

<pre>
{
    "took": 623,
    "timed_out": false,
    "_shards": {
        "total": 39,
        "successful": 39,
        "skipped": 0,
        "failed": 0
    },
...
</pre>
</details>


<details>
<summary>Using normal sorting - 63ms</summary>
<pre>
{
    "from": 0,
    "size": 2000,
    "sort" : {
        "published" : "desc"
    },
    "query": {
        "bool": {
            "filter": [
                {
                    "terms": {
                        "object.type": [
                            "dit:Interaction",
                            "dit:ServiceDelivery",
                            "dit:InvestmentProject",
                            "dit:OMISOrder"
                        ]
                    }
                }
            ]
        }
    }
}
</pre>

<pre>
{
    "took": 63,
    "timed_out": false,
    "_shards": {
        "total": 39,
        "successful": 39,
        "skipped": 0,
        "failed": 0
    },
...
</pre>
</details>

Queried agains DEV environment.

As you can see, the difference is very minor.

currycoder: Nice.  Just to qualify these comparisons, how large is the index that this is querying on dev?

rafenden: Looks like there are `161062` activities in the index and `39` shards.

currycoder: Cool.  For completeness, how do things look with 
```
...
    "from": 159000,
    "size": 2000,
...
```
?

rafenden: Getting an error in this case:
```
{
    "error": {
        "root_cause": [
            {
                "type": "query_phase_execution_exception",
                "reason": "Result window is too large, from + size must be less than or equal to: [10000] but was [161000]. See the scroll api for a more efficient way to request large data sets. This limit can be set by changing the [index.max_result_window] index level setting."
            },
            {
                "type": "query_phase_execution_exception",
                "reason": "Result window is too large, from + size must be less than or equal to: [10000] but was [161000]. See the scroll api for a more efficient way to request large data sets. This limit can be set by changing the [index.max_result_window] index level setting."
            },
```

Is there an auto-cleanup on dev?

rafenden: In total there is `161062` documents in the index, after filtering, it's `2242`

currycoder: Ah I've run in to this before with ES.  TBH I doubt we will see queries like this in practice for activity stream (massive sorted result sets), but it is worth double checking that the "tail" of pagination is performant IMO.

rafenden:
We decided to add a new field to each activity which will be used for a cross-activity sorting.

PR #2052, opened 2019-07-08
Bugfix/prevet unvalidated redirect from omis controller edit lead assignee

debitan:
## Description of change

Replace `returnUrl` with `res.locals.ORIGINAL_URL` in documents upload middleware in order to prevent unvalidated redirect as per this lgtm alert (https://lgtm.com/projects/g/uktrade/data-hub-frontend/snapshot/450ce6fc03cea40cac751f5a7ba955bd7579341d/files/src/apps/omis/apps/edit/controllers/edit-lead-assignee.js?sort=name&dir=ASC&mode=heatmap)

Using res.locals.ORIGINAL_URL sanitises the url as the middleware takes the baseUrl and checks if it has the https prefix and then adds the host path. This prevents the route redirecting the user off of the site. It then appends the baseUrl with the req.originalUrl.

This is existing middleware and so there is no new code required.

This is slightly different to the other two fixes to similar alerts as this controller had a variable `returnUrl`. However, `res.locals.ORIGINAL_URL` should do the same job with less code. 

## Test instructions

The user should still be redirected back to where they came from if there is no advisor or no order, and when the user has been successful in updating the lead adviser in the market.

## Checklist

### DIFF
--- a/src/apps/omis/apps/edit/controllers/edit-lead-assignee.js
+++ b/src/apps/omis/apps/edit/controllers/edit-lead-assignee.js
@@ -6,7 +6,9 @@ async function editLeadAssignee (req, res, next) {
   const adviserId = req.body.adviserId
   const orderId = req.body.orderId
 
-  if (!adviserId || !orderId) { return res.redirect(res.locals.ORIGINAL_URL) }
+  if (!adviserId || !orderId) {
+    return res.redirect(res.locals.ORIGINAL_URL)
+  }
 
   try {
     const allAssignees = await Order.getAssignees(req.session.token, orderId)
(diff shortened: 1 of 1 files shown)


src/apps/omis/apps/edit/controllers/edit-lead-assignee.js
```diff
@@ -5,11 +5,8 @@ const { Order } = require('../../../models')
 async function editLeadAssignee (req, res, next) {
   const adviserId = req.body.adviserId
   const orderId = req.body.orderId
-  const returnUrl = req.body.returnUrl || req.header('Referer')
 
-  if (!adviserId || !orderId) {
-    return res.redirect(returnUrl)
-  }
+  if (!adviserId || !orderId) { return res.redirect(res.locals.ORIGINAL_URL) }
```
rafenden: Could be more readable if return was placed on a separate line.

debitan: done

PR #2058, opened 2019-07-11
Add dynamic visual test screenshot name

PippoRaimondi:
## Description of change

## Test instructions

_What should I see?_
 
### Before

_Add a screenshot_

### After 

_Add a screenshot_

## Checklist


test/visual/wdio.conf.js
```diff
@@ -1,14 +1,15 @@
...
+const { addAttachment } = require('@wdio/allure-reporter').default
 
 const IMPLICIT_TIMEOUT = process.env.WDIO_IMPLICIT_TIMEOUT || 90000
 const BASE_URL = process.env.BASE_URL || 'http://localhost:3000'
 
 const browserStackUser = process.env.BROWSERSTACK_USERNAME || ''
 const browserStackKey = process.env.BROWSERSTACK_ACCESS_KEY || ''
 const isRemote = !!process.env.IS_REMOTE
-
+let testName = 'undefined'
```
yozzo: is this initialisation here necessary? You could get the same output from a conditional statement, as I don't see it being require anywhere else than the `beforeTest` block

PippoRaimondi: you are right, it's not necessary here as it's given a default value in the wrapper itself... anyways,  that default value at this point would always be overwritten...fixed!

PR #2065, opened 2019-07-19
Guard against no service being submitted

ian-leggett:
## Description of change

When submitting a interaction with no service the form errors

## Test instructions
Add interaction without a service selected.
 
### Before 
Screenshot 2019-07-19 at 11 09 24

### After 
Screenshot 2019-07-19 at 11 37 57

### DIFF
--- a/src/apps/interactions/transformers/interaction-form-body-to-api.js
+++ b/src/apps/interactions/transformers/interaction-form-body-to-api.js
@@ -14,28 +14,30 @@ function transformInteractionFormBodyToApiRequest (props, services) {
   const selectedServiceOption = services.find(
     service => service.value === props.service
   )
+
   const serviceAnswers = {}
+  let service = null
+  if (selectedServiceOption) {
+    const serviceHasSecondaryOptions = !!selectedServiceOption.secondaryOptions.length
 
-  const serviceHasSecondaryOptions = !!selectedServiceOption.secondaryOptions
-    .length
+    service = serviceHasSecondaryOptions
+      ? props.subService.find(id => id.length)
+      : selectedServiceOption.value
 
-  const service = serviceHasSecondaryOptions
-    ? props.subService.find(id => id.length)
-    : selectedServiceOption.value
+    const serviceOptionStore = serviceHasSecondaryOptions
+      ? selectedServiceOption.secondaryOptions[0]
+      : selectedServiceOption
 
-  const serviceOptionStore = serviceHasSecondaryOptions
-    ? selectedServiceOption.secondaryOptions[0]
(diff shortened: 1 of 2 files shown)


test/unit/apps/interactions/transformers/interaction-form-body-to-api.test.js
```diff
@@ -141,4 +141,23 @@ describe('#transformInteractionFormBodyToApiRequest', () => {
...
+      beforeEach(() => {
+        this.transformed = transformInteractionFormBodyToApiRequest(
+          {
+            service: undefined,
+          },
+          transformedServices
+        )
+      })
+
+      it('should pass empty string as service value', () => {
```
mcbhenwood: Maybe you should extend the test description to cover the empty `service_answers` too (or move the `service_answers` check to a separate text).

src/apps/interactions/transformers/interaction-form-body-to-api.js
```diff
@@ -14,28 +14,31 @@ function transformInteractionFormBodyToApiRequest (props, services) {
...
-  serviceOptionStore.interactionQuestions.map(interactionQuestion => {
-    for (const [key, value] of Object.entries(props)) {
-      if (key === interactionQuestion.value) {
-        serviceAnswers[key] = {
-          [value]: {},
+  const serviceAnswers = {}
+  let service = ''
+  if (selectedServiceOption) {
+    const serviceHasSecondaryOptions = !!selectedServiceOption.secondaryOptions
+      .length
```
mcbhenwood: I'm all for short lines but at first read this looks like we are casting `selectedServiceOption.secondaryOptions` to a boolean before checking the length property. I appreciate that doesn't make sense (and obviously on second glance isn't the case) but maybe the `selectedServiceOption.secondaryOptions.length` oughtn't be broken up in the case? Would reduce cognitive load when quickly parsing the code visually.

I'm not particularly wedded to this point; just observing.

test/unit/apps/interactions/transformers/interaction-form-body-to-api.test.js
yeahfro: > move the service_answers check to a separate text

👍

ian-leggett: No worries I will split the test into two as Lee mentioned

src/apps/interactions/transformers/interaction-form-body-to-api.js
ian-leggett: Good point this does read better on one line too

PR #2070, opened 2019-07-24
Add company to user list.

ian-leggett:
## Description of change
Trello - https://trello.com/c/kHaS9RTG/838-companies-list-add-and-remove-company

Users need to be able to "add" and "remove" companies to their own personal list which will soon be surfaced on the home page of Data Hub. 

## Test instructions

Note: You will need to set up a feature flag when running Leeloo locally, the flag is called `add_company_to_list`

1. Go to companies
2. Choose company
3. Click"add to my list" > You should see a success message
4. Click "remove from my list" > you should see a "add to my list" button re-appear
5. Archive the company and you should still see a "remove from my list" button if its in your list but not a "add to my list" button if it isn't.

 
### Before
Screenshot 2019-07-31 at 10 30 39

#### Company not added
Screenshot 2019-07-31 at 10 31 07

#### Company added
Screenshot 2019-07-31 at 10 31 16

#### Company removed
Screenshot 2019-07-31 at 10 31 24

#### Company added but archived
Screenshot 2019-07-31 at 10 31 51

## Checklist


test/unit/apps/companies/middleware/params.test.js
```diff
@@ -44,4 +50,75 @@ describe('Companies form middleware', () => {
...
+        }
+        this.reqMock = {
+          ...this.reqMock,
+          body: {
+            action: 'remove',
+          },
+        }
+        await this.middleware.addCompanyOrRemoveFromList(this.reqMock, this.resMock, this.nextSpy)
+      })
+      it('should add a company to the user list', () => {
```
mcbhenwood: "remove a company from" :)

src/apps/companies/middleware/params.js
```diff
@@ -1,4 +1,4 @@
-const { getDitCompany, getCHCompany } = require('../repos')
+const { getDitCompany, addDitCompanyToList, isDitCompanyAdded, removeDitCompanyFromList, getCHCompany } = require('../repos')
```
yeahfro: Now that five functions are being imported, this would read nice split over multiple lines.

src/apps/companies/middleware/params.js
```diff
@@ -9,6 +9,37 @@ async function getCompany (req, res, next, id) {
...
+    await isDitCompanyAdded(req.session.token, id)
+    res.locals.isCompanyAlreadyAdded = true
+    next()
+  } catch (error) {
+    res.locals.isCompanyAlreadyAdded = false
+    next()
+  }
+}
+
+async function addCompanyOrRemoveFromList (req, res, next) {
```
yeahfro: An alternative for this `function` might be to have an `object`:

```
const someName = {
  add: async (req, companyId) => {
    await addDitCompanyToList(req.session.token, companyId)
    req.flash('success', `This company has been added to your list of companies. You can find this list on your Data Hub Homepage. Only you can see this list.`)
  },
  remove: async (req, companyId) => {
    await removeDitCompanyFromList(req.session.token, companyId)
  },
}
```

And then:
```
async function addCompanyOrRemoveFromList (req, res, next) {
  const companyId = res.locals.company.id

  try {
    await someName[req.body.action](req, companyId)
  } catch (error) {
    next(error)
  }
}
```

src/apps/companies/middleware/params.js
```diff
@@ -9,6 +9,37 @@ async function getCompany (req, res, next, id) {
...
+    try {
+      await addDitCompanyToList(req.session.token, companyId)
+      req.flash('success', `This company has been added to your list of companies. You can find this list on your Data Hub Homepage. Only you can see this list.`)
+      res.redirect(`${req.baseUrl}/${companyId}`)
+    } catch (error) {
+      next(error)
+    }
+  } else if (req.body.action === 'remove') {
+    try {
+      await removeDitCompanyFromList(req.session.token, companyId)
```
yeahfro: Does this not need a `req.flash` for user feedback?

src/apps/companies/middleware/params.js
```diff
@@ -9,6 +9,37 @@ async function getCompany (req, res, next, id) {
...
+  }
+}
+
+async function addCompanyOrRemoveFromList (req, res, next) {
+  const companyId = res.locals.company.id
+  if (req.body.action === 'add') {
+    try {
+      await addDitCompanyToList(req.session.token, companyId)
+      req.flash('success', `This company has been added to your list of companies. You can find this list on your Data Hub Homepage. Only you can see this list.`)
+      res.redirect(`${req.baseUrl}/${companyId}`)
```
yeahfro: I noticed that this redirect is not being tested

src/apps/companies/views/template.njk
```diff
@@ -1,9 +1,28 @@
 {% extends "_layouts/template.njk" %}
 
 {% block local_header %}
-  {% call LocalHeader({ heading: company.name, modifier: 'light-banner' }) %}
-    <p class="c-local-header__heading-after">{{ company.address | formatAddress }}</p>
+  {% set action %}
+    <form action="/companies/{{ company.id }}/manage-company-list?_csrf={{ csrfToken }}" method="post">
+      <input type="hidden" value="{{ company.id }}"/>
+      {% if isCompanyAlreadyAdded %}
+        <button type="submit" name="action" value="remove" class="govuk-button govuk-button--secondary">Remove from my
```
yeahfro: It might be personal preference but I feel the `Remove from my list` would read better on a single line. Perhaps you can break the attributes over multiple lines?

src/apps/companies/views/template.njk
```diff
@@ -1,9 +1,28 @@
...
+          list
+        </button>
+      {% else %}
+        {% if not company.archived %}
+          <button type="submit" name="action" value="add" class="govuk-button govuk-button--secondary">Add to my list
```
yeahfro: As above

test/functional/cypress/specs/companies/add-company-to-list.spec.js
```diff
@@ -0,0 +1,28 @@
+const selectors = require('../../selectors')
+
+describe('Add company to list', () => {
+  context('when a company is already added to the users company list', () => {
+    before(() => {
+      cy.visit(`/companies/0fb3379c-341c-4da4-b825-bf8d47b26baa/activity`)
```
yeahfro: Only minor, I have been referencing the company ID from the `~/fixtures/company/..` constants.

src/apps/companies/middleware/params.js
```diff
@@ -9,6 +9,37 @@ async function getCompany (req, res, next, id) {
   }
 }
 
+async function isCompanyAddedToList (req, res, next, id) {
```
mcbhenwood: The name of this function suggests that it returns a boolean (and is probably pure) whereas in fact it returns nothing and actually sets a boolean in the locals. Maybe a better name would be something like `setIsCompanyAlreadyAdded` or similar.

src/apps/companies/repos.js
```diff
@@ -6,6 +6,27 @@ function getDitCompany (token, id) {
   return authorisedRequest(token, `${config.apiRoot}/v4/company/${id}`)
 }
 
+function isDitCompanyAdded (token, id) {
```
mcbhenwood: I get that this function can help determine whether to not the company is added, on the basis that it returns 'something' (truthy) if the company exists and throws an exception if not. But I kind of feel that a function whose name begins "is" should return pure booleans.

Plus, the whole "it throws an exception if the company is not there" thing really is a contract between this function and the API. It shouldn't befall any code which uses this function to have to know about that implementation.

So maybe move the try / catch block into this function and make it return one of only `true` or `false` under all conditions.

Then the code in the (renamed) 'isCompanyAddedToList' function could just read:
```javascript
res.locals.isCompanyAlreadyAdded = await isDitCompanyAdded(req.session.token, id)
next()
```

mcbhenwood: If for some reason we can't introduce this structure into repos.js, then I would suggest instead renaming 'isDitCompanyAdded' to something like 'getDitCompanyFromList` (because that is actually what it does) because that would push the onus for try / catch logic into any consuming code.

test/unit/apps/companies/middleware/params.test.js
ian-leggett: Good spot 👍

src/apps/companies/middleware/params.js
ian-leggett: Much better thanks 👍

src/apps/companies/middleware/params.js
ian-leggett: Agree, this is much tidier 👍

src/apps/companies/middleware/params.js
ian-leggett: Good spot thanks

src/apps/companies/middleware/params.js
ian-leggett: Tests are in now

src/apps/companies/views/template.njk
ian-leggett: Formatted it as per comment looks much better

src/apps/companies/views/template.njk
ian-leggett: All done

test/functional/cypress/specs/companies/add-company-to-list.spec.js
ian-leggett: Much nicer thanks

src/apps/companies/repos.js
ian-leggett: Thanks totally see your point and it does read as if it would return a boolean. I have updated the name of the function but assigning the boolean to the `res.locals.isCompanyAlreadyAdded` is a bit tricky as that function `isDitCompanyAdded` just return a function

src/apps/companies/middleware/params.js
ian-leggett: Much better thanks 👍

src/apps/companies/middleware/params.js
```diff
@@ -9,6 +15,43 @@ async function getCompany (req, res, next, id) {
...
+    next()
+  } catch (error) {
+    res.locals.isCompanyAlreadyAdded = false
+    next()
+  }
+}
+
+const companyListActions = {
+  add: async (req, res, next, companyId) => {
+    try {
```
yeahfro: I feel that this `try .. catch` and `res.redirect` could just be used once so that it is around `companyListActions[req.body.action](req, res, next, companyId)`

ie you need this:

```
const companyListActions = {
  add: async (req, companyId) => {
    await addDitCompanyToList(req.session.token, companyId)
    req.flash('success', 'This company has been added to your list of companies. You can find this list on your Data Hub Home page. Only you can see this list.')
  },
  remove: async (req, companyId) => {
    await removeDitCompanyFromList(req.session.token, companyId)
    req.flash('success', 'This company has been removed from your list.')
  },
}
```

And then just:

```
async function addCompanyOrRemoveFromList (req, res, next) {
  const companyId = res.locals.company.id
  try {
    await companyListActions[req.body.action](req, companyId)
    res.redirect(`${req.baseUrl}/${companyId}`)
  } catch (e) {
    next(e)
  }
}
```

ian-leggett: This is much nicer thanks 👍

src/apps/companies/repos.js
ian-leggett: After playing around with this one I did get a refactor done, @mcbhenwood see what you think. I think it's a bit tidier and removed the `next(error)` as we don't actually need it.

src/apps/companies/repos.js
```diff
@@ -6,6 +6,27 @@ function getDitCompany (token, id) {
   return authorisedRequest(token, `${config.apiRoot}/v4/company/${id}`)
 }
 
+function getDitCompanyFromList (token, id) {
```
rafenden: Is that list ID or company ID? It would be nice to have either `companyId` or `listId` here.

test/functional/cypress/specs/companies/add-company-to-list.spec.js
```diff
@@ -0,0 +1,29 @@
+const selectors = require('../../selectors')
+const fixtures = require('../../fixtures')
```
rafenden: I suggest to create `test/functional/cypress/fixtures/company/index.js` exporting all of the companies and then importing theme here like that:

```
const { lambdaPlc, marsExportsLtd } = require('../../fixtures/company')
```

PR #2073, opened 2019-07-26
React Boilerplate/Sandbox

stackzak:
## Description of change

This PR is a boilerplate that intends to put forward a way of working with React. I'm keen to get everyone's feedback on this approach.

It also provides a React sandbox on the dev environment by going to /react-sandbox on your local env. 

### What's included
- There's now react component entry file that expects our react component ID's to start with 'react-mount-...'
- A small Nunjucks react slot component
- State management using React Hooks & Context with some syntactic sugar using [Constate](https://github.com/diegohaz/constate)
- A sandbox URL
- A Sandbox.jsx app that is intended to help when creating your own components
- A flow of data from the Node controller async requests that then get injected it into our front end JSX files without the use of data attributes 

## Test instructions
-Clear your node directory and re-install modules
-Browse to localhost:3000/react-sandbox
 
## Checklist

### DIFF
--- a/assets/javascripts/app.js
+++ b/assets/javascripts/app.js
@@ -3,6 +3,7 @@ require('core-js/fn/array/from')
 require('core-js/fn/promise')
 require('element-closest')
 require('classlist.js')
+require('./react/index.js')
 
 const ConditionalSubfields = require('./modules/conditional-subfields')
 const SortableTable = require('./modules/sortable-table')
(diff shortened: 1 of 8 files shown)


mcbhenwood:
@rafenden -

How does this compare with the method you used to get React components rendering in Activity Feed? Before today I wasn't previously aware that we had already covered the ground which @stackzak is trying to cover here. [I don't think he was, either.]

It would be good to use a common (identical) approach in this case as well. Can you look at this PR ahead of Zak's return tomorrow, with a view to aligning these efforts in the Dashboard Company List feature in play here? Thanks.

PR #2079, opened 2019-08-02
Use prod environment var when building client-side code on CircleCI

rafenden:
## Description of change

CircleCI will now run tests using code compiled same way as on prod.

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -290,7 +290,7 @@ jobs:
       - checkout
       - *restore_yarn_cache
       - run: yarn install --pure-lockfile
-      - run: yarn build
+      - run: yarn build:prod
       - run:
           name: Run client-side unit tests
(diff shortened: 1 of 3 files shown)


stackzak:
I think this will cause problems both with some of our tests and some of our code, as there's a few different areas that we're relying on the test flag (Redis for example). 

Another suggestion to pick up these issues in future would be to have the our dev environment automatically build from the latest develop branch and alert accordingly.

rafenden:
> I think this will cause problems both with some of our tests and some of our code, as there's a few different areas that we're relying on the test flag (Redis for example).
> 
> Another suggestion to pick up these issues in future would be to have the our dev environment automatically build from the latest develop branch and alert accordingly.

The `NODE_ENV` flag was used only to build the code, tests run as before.

PR #2080, opened 2019-08-02
Fix failing build 

stackzak:
Fixes the Uglify issues that are causing build failures.

### DIFF
--- a/package.json
+++ b/package.json
@@ -123,6 +123,7 @@
     "sniffr": "^1.1.4",
     "style-loader": "^0.21.0",
     "tar": "^4.4.8",
+    "uglifyjs-webpack-plugin": "^1.3.0",
     "uuid": "^3.1.0",
     "vue": "^2.5.13",
     "vue-loader": "^15.2.2",
--- a/webpack.config.prod.js
+++ b/webpack.config.prod.js
@@ -1,5 +1,6 @@
 const webpack = require('webpack')
 const ExtractTextPlugin = require('extract-text-webpack-plugin')
+const UglifyJsPlugin = require('uglifyjs-webpack-plugin')
 
 module.exports = {
   devtool: false,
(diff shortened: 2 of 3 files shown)


webpack.config.prod.js
```diff
@@ -9,18 +10,18 @@ module.exports = {
   plugins: [
     new webpack.DefinePlugin({
       'process.env': {
-        'NODE_ENV': JSON.stringify('production'),
+        NODE_ENV: JSON.stringify('production'),
```
web-bert: No need for the `JSON.stringify` here?

stackzak: I have no idea why that's there or what it's doing but it was there before and I'm afraid to touch it and have some other build problem pop up, we can look at it outside of this PR perhaps?

paulgain: `console.log(JSON.stringify('production'))` yields `""production""`. I would say it does absolutely nothing.

web-bert: Seems you are correct to be concerned: https://stackoverflow.com/questions/50454341/why-json-stringifyproduction Although it would probably be better to use the backtick template to include the quotes - to make it clear that is the intention. But as you say, maybe do that as a separate PR 👍

package.json
```diff
@@ -123,6 +123,7 @@
     "sniffr": "^1.1.4",
     "style-loader": "^0.21.0",
     "tar": "^4.4.8",
+    "uglifyjs-webpack-plugin": "^1.3.0",
```
rafenden: You are missing `yarn.lock` update.

paulgain: 👍

webpack.config.prod.js
paulgain: So it does do something 👍

package.json
stackzak: Added

PR #2100, opened 2019-08-27
Fix running FE/BE together on docker - ensure everything works with mock SSO

currycoder:
## Description of change

This ensures that the frontend and backend can work together over docker.  This ensures that `SESSION_SECRET` has a default value, that mock-sso's token is a consistent value and that a username can be specified through an environment variable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,6 +136,10 @@ and be provided with a back end server to provide the API, data storage and sear
     ```bash
     API_ROOT=example.com docker-compose up frontend
     ```
+
+    If using the mock-sso instance specified in `docker-compose.yml`, you will 
+    also need to specify the `MOCK_SSO_USERNAME` environment variable in `.env`.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -139,6 +139,10 @@ and be provided with a back end server to provide the API, data storage and sear
     ```bash
     API_ROOT=example.com docker-compose up frontend
     ```
+
+    If using the mock-sso instance specified in `docker-compose.yml`, you will 
+    also need to specify the `MOCK_SSO_USERNAME` environment variable in `.env`.
+    This should wire up with an Adviser in the data hub backend.
```
mcbhenwood: Might be better to state explicitly what's needed, e.g. "This value should correspond with an `Advisor.email` value in the Data Hub backend" (if indeed that is what is meant).

PR #2118, opened 2019-09-05
Update styles for the Help centre feed on homepage

ian-leggett:
## Description of change
The Help centre feed count needs reducing to one item and styles need updating to tighten up the layout.

## Test instructions
One feed item showing and a tighter layout.
 
### Before

Screenshot 2019-09-05 at 12 15 07

### After 

Screenshot 2019-09-05 at 12 15 00

## Checklist


src/templates/_components/info-feed.njk
```diff
@@ -30,7 +30,7 @@
 {% if subHeading %} <p class="dashboard-section__subheading">{{subHeading}}</p> {% endif %}
```
yeahfro: Is this `subHeading` option redundant now?

src/templates/_components/info-feed.njk
```diff
@@ -30,7 +30,7 @@
...
 {% if dataFeed.length %}
   {% for item in dataFeed.slice(0, feedLimit) %}
-      <a href="{{ item.link }}" class="dashboard-section__info-feed-link" target="_blank">{{ item.heading }}</a>
+    <a href="{{ item.link }}" class="dashboard-section__info-feed-link" target="_blank">{{ item.heading }}</a> <span class="dashboard-section__info-feed-link-note">(Link opens in a new window)</span>
```
yeahfro: It'd read nicer if this was broken over multiple lines as it's quite a long line with multiple elements as it stands.

src/templates/_components/info-feed.njk
```diff
@@ -30,7 +30,7 @@
...
       <p class="dashboard-section__info-feed-date">{{ item.date }}</p>
```
yeahfro: Thinking if this was the right thing to do, date inside <p>

yeahfro:
Also there doesn't appear to be any test updates in `info-feed.test.js`. Would expect there to be something based upon change.

src/templates/_components/info-feed.njk
ian-leggett: I am going to leave it in there as if we have more than one item showing then this can be a useful message.

src/templates/_components/info-feed.njk
ian-leggett: 👍

ian-leggett:
> Also there doesn't appear to be any test updates in `info-feed.test.js`. Would expect there to be something based upon change.

Just added a test for the negative scenario around no sub heading 👍

PR #2158, opened 2019-09-24
Make use of new /v4/metadata endpoints

elcct:
## Description of change

This makes use of new `/v4/metadata` endpoints in the backend which use Hawk authorisation.

Existing `/metadata` endpoints are deprecated and will be removed on or after 17th October.

~Currently the Hawk implementation in the backend also performs IP check that is specific to PaaS, so without setting up a reverse proxy, the tests will not work locally. There is a PR that adds an option to disable this check.~ It has been merged.

## Test instructions

The front end will not work if metadata can't be requested.
 
Front end needs following test environment variables:
```
DATA_HUB_BACKEND_ACCESS_KEY_ID=frontend-key-id
DATA_HUB_BACKEND_SECRET_ACCESS_KEY=frontend-key
```

and corresponding variables in the backend running locally:

```
DATA_HUB_FRONTEND_ACCESS_KEY_ID=frontend-key-id
DATA_HUB_FRONTEND_SECRET_ACCESS_KEY=frontend-key
DISABLE_PAAS_IP_CHECK=True
```

### Before

_Add a screenshot_

### After 

_Add a screenshot_

## Checklist


src/lib/hawk-request.js
```diff
@@ -0,0 +1,86 @@
+const request = require('request')
+const hawk = require('@hapi/hawk')
+const config = require('../../config')
+
+function getHawkHeader (credentials, requestOptions) {
```
rafenden: We need unit tests for this file.

elcct: I have added unit tests now.

PR #2167, opened 2019-10-01
Use latest sandbox image

PippoRaimondi:
## Description of change

Use latest sandbox rather than tag now that master has the latest code

## Test instructions

N/A

## Checklist


web-bert:
Should we not just add a new tag and use that? What if something gets merged into master that is a breaking change, i.e. we go to 4.0.0, then all the tests will break?

PR #2191, opened 2019-10-17
Move unit tests to the same level as the tested file

rafenden:
## Description of change

Move unit tests to a location closer to the tested file so humans are happy.

We should also move the `/test/unit/data` directory to relevant `__fixtures__` directories relative to the testing file.

## Checklist


test/unit/nunjucks.js
```diff
@@ -1,5 +1,6 @@
 const nunjucks = require('nunjucks')
```
PippoRaimondi: should this be moved to `/src` as well?

rafenden: Not sure, it's just a helper file.

src/apps/interactions/transformers/__test__/interaction-to-options.test.js
```diff
@@ -1,4 +1,4 @@
-const serviceOptionData = require('../../../data/interactions/service-options-data.json')
+const serviceOptionData = require('~/test/unit/data/interactions/service-options-data.json')
```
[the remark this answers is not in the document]
ian-leggett: To be fair this is more about moving the tests. All good :-)

rafenden:
> In general good, but weren't we moving to drop the use of wavy and the tilde `~` because it introduces unwelcome ambiguity?

I would prefer to do this in two separate PRs as this will be a huge amount of work.

PR #2196, opened 2019-10-21
Remove hard coded apiRoot and use from config

web-bert:
## Description of change

Remove hard coded apiRoot so the tests don't break when you are not running the BE on port 8000
Also refactor tests to be simpler

### DIFF
--- a/src/lib/__test__/metadata.test.js
+++ b/src/lib/__test__/metadata.test.js
@@ -1,57 +1,60 @@
 const rewire = require('rewire')
+const config = require('../../../config')
 
-describe('#metadata: check getMetadata', () => {
+const modulePath = '../metadata'
+
+describe('metadata', () => {
+  let metadata
+  let hawkRequest
   beforeEach(() => {
-    this.metadata = rewire('~/src/lib/metadata')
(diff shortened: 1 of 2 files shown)


src/lib/__test__/metadata.test.js
```diff
@@ -1,57 +1,60 @@
...
-
-    let hawkRequest = sinon.stub().rejects()
-    this.metadata.__set__('hawkRequest', hawkRequest)
-    await expect(getMetadataItem('fake', 2)).to.be.rejectedWith(Error)
-    expect(hawkRequest).to.have.been.calledOnceWith('http://localhost:8000/v4/metadata/fake')
+  describe('#getMetadataItem', () => {
+    it('gets correct metadata item', async () => {
+      const getMetadataItem = metadata.__get__('exports').getMetadataItem
+
+      hawkRequest = sinon.stub().resolves(
```
rafenden: We should probably use `nock` in for network mocking.

web-bert: I agree, I just refactored these tests without modifying so that they pass for me locally

PR #2204, opened 2019-10-23
Ensure accepting untrusted certificates only happens once

web-bert:
## Description of change

An unnecessary check was happening for each request when it only needs to happen once when the server starts


web-bert:
@rafenden tests added for the refactor but I'm not adding retrospective tests now as this was meant to be a quick fix

PR #2215, opened 2019-10-29
Feature/fex 211 enable activity feed filters

yozzo:
## Description of change

Integration of Activity Feed filters in Data-hub. The way we're filtering is by querying the API with the filter params that come from the component.

## Test instructions

On Activity Feed we should see a dropdown with types of values we should be filtering by

 
### Before

Screenshot 2019-10-29 at 17 00 20

### After 

Screenshot 2019-10-29 at 17 00 37

## Checklist


src/apps/companies/apps/activity-feed/builders.js
```diff
@@ -0,0 +1,27 @@
...
+  }
+
+  return query
+}
+
+function buildFilter (item = '') {
+  const terms = JSON.parse(item)
+
+  return { terms }
+}
```
peterhudec: All of this can be simplified to:
```js
const buildActivityFeedFilters = (companyId, filter) => [
  { term: { 'object.attributedTo.id': `dit:DataHubCompany:${companyId}` } },
  ..._.flatten([filter]).map(filter => ({ terms: JSON.parse(filter) })),
]
```

src/apps/companies/apps/activity-feed/controllers.js
```diff
@@ -1,11 +1,28 @@
...
+    ],
+    default: ACTIVITY_TYPE_FILTERS.dataHubActivity,
+  }
+
   try {
     const addContentProps = company.archived ? {} : {
       addContentText: 'Add interaction',
       addContentLink: `/companies/${company.id}/interactions/create`,
+      addActivityTypeFilter: activityTypeFilter,
+      isFilterEnabled: activityTypeFilter.values && features['activity-feed-filters'],
```
peterhudec: What's the point of having `activityTypeFilter.values &&` here? We just defined it a couple of lines earlier. Also an array is always _truthy_, you need to check its `length`.

src/apps/companies/apps/activity-feed/controllers.js
```diff
@@ -1,11 +1,28 @@
...
+      ACTIVITY_TYPE_FILTERS.externalActivity,
+    ],
+    default: ACTIVITY_TYPE_FILTERS.dataHubActivity,
+  }
+
   try {
     const addContentProps = company.archived ? {} : {
       addContentText: 'Add interaction',
       addContentLink: `/companies/${company.id}/interactions/create`,
+      addActivityTypeFilter: activityTypeFilter,
```
peterhudec: You could have inlined the `activityTypeFilter` literal here. But if you want to have it declared explicitly, then perhaps name it `addActivityTypeFilter` so you can use the shorthand here.

peterhudec:
Oh, sorry @yozzo I haven't realized that this is a WIP.

config/index.js
```diff
@@ -115,12 +116,14 @@ const config = {
...
-      'dit:InvestmentProject',
-      'dit:OMISOrder',
-    ],
+    supportedActivityTypes: [].concat(
+      ...[
+        ACTIVITY_TYPE.Interaction,
+        ACTIVITY_TYPE.InvestmentProject,
+        ACTIVITY_TYPE.Omis,
+      ]
+    ),
```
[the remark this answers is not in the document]
yozzo: `[ACTIVITY_TYPE.Interaction]` is an array itself and we need a flat array, rather than an array of arrays, hence `concat` does that for us in this case

src/apps/companies/apps/activity-feed/builders.js
```diff
@@ -0,0 +1,27 @@
+const { forEach, isArray } = require('lodash')
```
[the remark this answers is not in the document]
yozzo: `lodash` is preferred to native JS on this project when it comes to utility functions. Also lodash is _allegedly_ faster than native

src/apps/companies/apps/activity-feed/builders.js
yozzo: the `buildActivityFeedFilters` function might require further amends for an upcoming feature so it'd be better to read the way it is right now. @paulgain to confirm

src/apps/companies/constants.js
```diff
@@ -66,13 +66,67 @@ const LOCAL_NAV = [
   },
 ]
 
+const ACTIVITY_TYPE = {
+  CompaniesHouseAccount: ['dit:Accounts'],
+  CompaniesHouseCompany: ['dit:Company'],
+  HmrcExporter: ['dit:Export'],
+  Interaction: ['dit:Interaction', 'dit:ServiceDelivery'],
```
rafenden: We should split Interaction into two filters and this will simplify the code and would make `Array.concat` calls redundant.

e.g.

```
Interaction: 'dit:Interaction',
ServiceDelivery: 'dit:ServiceDelivery',
```

and then
```
const ACTIVITY_TYPE_FILTERS = {
  all: {
    label: 'All Data Hub & external activity',
    value: [
        ACTIVITY_TYPE.CompaniesHouseAccount,
        ACTIVITY_TYPE.CompaniesHouseCompany,
        ACTIVITY_TYPE.HmrcExporter,
        ACTIVITY_TYPE.Interaction,
        ACTIVITY_TYPE.ServiceDelivery,
        ACTIVITY_TYPE.InvestmentProject,
        ACTIVITY_TYPE.Omis,
    ],
  },
...
```

src/apps/companies/apps/activity-feed/builders.js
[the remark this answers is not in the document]
rafenden: It's worth discussing this as a native-first approach should be the preferred one (in my opinion).

src/apps/companies/constants.js
yozzo: I think that's a product decision rather than code, but happy to amend if the filter is to be split

yarn.lock
```diff
@@ -6551,12 +6568,7 @@ got@^8.3.1:
     url-parse-lax "^3.0.0"
     url-to-options "^1.0.1"
 
-govuk-colours@^1.0.3:
-  version "1.0.3"
-  resolved "https://registry.yarnpkg.com/govuk-colours/-/govuk-colours-1.0.3.tgz#41d983693626dbaadb05594c5931a51696f79488"
-  integrity sha512-xdVD7YzHG+KEwCs8Rl+MgVfgEagjANVau4dJNTe30Xi6KySJ1LaI1vzBRRgThBjmJC8YFTwk4CshRtRfSbXKcA==
-
-govuk-colours@latest:
+govuk-colours@^1.0.3, govuk-colours@latest:
```
rafenden: Looks like this change made outline of React components brighter than the rest of Data Hub.

image

src/apps/companies/apps/activity-feed/builders.js
paulgain: Let's keep it as is for now.

yozzo:
waiting for https://github.com/uktrade/data-hub-components/pull/216 to be merged

src/apps/companies/apps/activity-feed/builders.js
```diff
@@ -0,0 +1,25 @@
+function buildActivityFeedFilters (companyId, filter) {
+  const query = [
+    { term: { 'object.attributedTo.id': `dit:DataHubCompany:${companyId}` } },
+  ]
+
+  if (Array.isArray(filter)) {
+    Array.prototype.forEach(filter, (item) => {
+      query.push(buildFilter(item))
```
mcbhenwood: This looks like we are taking the filter names from the front end and building an ES query using these verbatim without validation or checks. If that's the case then we need to stop doing that because it's basically an ES version of a SQL Injection exposure and as such is a undesirable attack surface. This is probably quite closely aligned with the comments made by @paulgain in the Components repo.

However to my mind the problem is _this_ layer which appears to pass on those arguments without validating their suitability.

yozzo: we removed the ES logic outside of the browser side and it really simplified everything 🎉  
and _this_ layer is now checking for the right values otherwise returns and empty object:
https://github.com/uktrade/data-hub-frontend/pull/2215/files#diff-d7cb0675d9c27fc0cf8cc958d5f55feaR19-R34

src/apps/companies/apps/activity-feed/builders.js
```diff
@@ -0,0 +1,39 @@
...
+    case 'my-activity':
+      return { 'terms': { 'object.attributedTo.id': [`dit:DataHubAdviser:${user.id}`] } }
+    case 'datahub-activity':
+      return { 'terms': { 'object.type': dataHubActivity } }
+    case 'external-activity':
+      return { 'terms': { 'object.type': externalActivity } }
+    default:
+      return {}
+  }
+}
```
paulgain: Rather than use magic strings we should put these into constants, where they can be exported and reused elsewhere especially within our tests.

```

const FILTERS: {
   ALL: 'all',
   MY_ACTIVITY: 'my-activity',
   DATA_HUB_ACTIVITY: 'data-hub-activity',
   EXTERNAL_ACTIVITY: 'external-activity'
}

const OBJECT: {
    TYPE: 'object.type'
    ATTRIBUTED_TO_ID: 'object.attributedTo.id':  
}

const terms = (key, value) => {
  return {
     terms: {
       [key]: value
     }
  }
}
```

```
  switch (queryParams) {
    case Filters.ALL: 
        return terms(OBJECT.TYPE, allActivity)
    case Filters.MY_ACTIVITY: 
        return terms(OBJECT.ATTRIBUTED_TO_ID, ...etc)
    case Filters.DATA_HUB_ACTIVITY:
        return terms(OBJECT.TYPE, dataHubActivity)
    case Filters.EXTERNAL_ACTIVITY:
        return terms(OBJECT.TYPE, externalActivity)
  }
```

src/apps/companies/constants.js
```diff
@@ -66,13 +66,76 @@ const LOCAL_NAV = [
...
+  allActivity: [].concat(
+    ...[
+      ACTIVITY_TYPE.CompaniesHouseAccount,
+      ACTIVITY_TYPE.CompaniesHouseCompany,
+      ACTIVITY_TYPE.HmrcExporter,
+      ACTIVITY_TYPE.Interaction,
+      ACTIVITY_TYPE.InvestmentProject,
+      ACTIVITY_TYPE.Omis,
+    ]
+  ),
```
paulgain: There's is easier read:
```
  allActivity: [
    ...ACTIVITY_TYPE.CompaniesHouseAccount,
    ...ACTIVITY_TYPE.CompaniesHouseCompany,
    ...ACTIVITY_TYPE.HmrcExporter,
    ...ACTIVITY_TYPE.Interaction,
    ...ACTIVITY_TYPE.InvestmentProject,
    ...ACTIVITY_TYPE.Omis,
  ]
```

src/apps/companies/constants.js
```diff
@@ -66,13 +66,76 @@ const LOCAL_NAV = [
...
+
+  externalActivity: [].concat(
+    ...[
+      ACTIVITY_TYPE.CompaniesHouseAccount,
+      ACTIVITY_TYPE.CompaniesHouseCompany,
+      ACTIVITY_TYPE.HmrcExporter,
+    ]
+  ),
```
paulgain:
```
  externalActivity: [
    ...ACTIVITY_TYPE.CompaniesHouseAccount,
    ...ACTIVITY_TYPE.CompaniesHouseCompany,
    ...ACTIVITY_TYPE.HmrcExporter,
  ]
```

src/apps/companies/constants.js
```diff
@@ -66,13 +66,76 @@ const LOCAL_NAV = [
...
+
+  dataHubActivity: [].concat(
+    ...[
+      ACTIVITY_TYPE.Interaction,
+      ACTIVITY_TYPE.InvestmentProject,
+      ACTIVITY_TYPE.Omis,
+    ]
+  ),
+}
+
```
paulgain: Same applies here.

src/apps/companies/constants.js
yozzo: `Interaction` is composed by an array of `['dit:Interaction', 'dit:ServiceDelivery']`,  it's be only `externalActivity` that could be composited without concatenation, but then we'll have arrays and strings, I'd rather keep them all as arrays, in case requirements might change and more filter params will be needed.

src/apps/companies/apps/activity-feed/__test__/repos.test.js
```diff
@@ -1,6 +1,10 @@
 const config = require('~/src/config')
 const activityFeedRawFixture = require('~/test/unit/data/activity-feed/activity-feed-from-es')
 
+const { ACTIVITY_TYPE_FILTER_KEYS, ACTIVITY_TYPE_FILTER_OBJECT } = require('../../../constants')
+const KEYS = ACTIVITY_TYPE_FILTER_KEYS
```
rafenden: I would prefer to use the full name to avoid confusion.

src/apps/companies/apps/activity-feed/__test__/repos.test.js
```diff
@@ -1,6 +1,10 @@
...
+const OBJECT = ACTIVITY_TYPE_FILTER_OBJECT
```
rafenden: Same here.

src/apps/companies/apps/activity-feed/controllers.js
```diff
@@ -1,11 +1,24 @@
...
+    externalActivity,
+    dataHubActivity,
+  }
+
   try {
     const addContentProps = company.archived ? {} : {
       addContentText: 'Add interaction',
       addContentLink: `/companies/${company.id}/interactions/create`,
+      addActivityTypeFilter,
+      isTypeFilterEnabled: features['activity-feed-filters'],
```
rafenden: The flag should match the prop name, i.e. `activity-feed-type-filter-enabled`.

src/apps/companies/constants.js
paulgain: The `concat` needs to go, it's not an elegant solution.

package.json
```diff
@@ -96,7 +96,7 @@
     "file-loader": "^1.1.6",
     "formidable": "^1.2.1",
     "get-form-data": "^2.0.0",
-    "govuk-colours": "latest",
+    "govuk-colours": "1.0.3",
```
rafenden: Good catch!

src/apps/companies/apps/activity-feed/builders.js
```diff
@@ -0,0 +1,47 @@
...
+    case ACTIVITY_TYPE_FILTERS.allActivity.value:
+      return terms(ACTIVITY_TYPE_FILTER_OBJECT.TYPE, allActivity)
+    case ACTIVITY_TYPE_FILTERS.myActivity.value:
+      return terms(ACTIVITY_TYPE_FILTER_OBJECT.ATTRIBUTED_TO_ID, [`dit:DataHubAdviser:${user.id}`])
+    case ACTIVITY_TYPE_FILTERS.dataHubActivity.value:
+      return terms(ACTIVITY_TYPE_FILTER_OBJECT.TYPE, dataHubActivity)
+    case ACTIVITY_TYPE_FILTERS.externalActivity.value:
+      return terms(ACTIVITY_TYPE_FILTER_OBJECT.TYPE, externalActivity)
+    default:
+      return {}
```
rafenden: We should probably return here `config.activityFeed.supportedActivityTypes`.

yozzo: good call

src/apps/companies/apps/activity-feed/__test__/repos.test.js
```diff
@@ -14,22 +16,27 @@ describe('Activity feed repos', () => {
...
     context('when called with companyId', () => {
       beforeEach(async () => {
-        this.results = await this.repos.fetchActivityFeed({ token, from: 1, size: 21, companyId: '123' })
+        this.results = await this.repos.fetchActivityFeed(
+          { token,
+            from: 1,
+            size: 21,
+            filter: {
+              terms: {
+                [ACTIVITY_TYPE_FILTER_OBJECT.TYPE]: ACTIVITY_TYPE_FILTER_KEYS.dataHubActivity.value,
```
paulgain: This is absolute nuts, the constant names are way too long. We should move all Activity Feed constants into its own file and then drop `ACTIVITY_TYPE `.

config/index.js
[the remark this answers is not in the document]
peterhudec: Yeah, I actually didn't know, that you can use `concat()` to flatten an array.

src/apps/companies/apps/activity-feed/__test__/repos.test.js
yozzo: I feel too much attention is being given to constants.

PR #2228, opened 2019-11-05
Chore/company lists permission cleanup

peterhudec:
## Description of change

Cleaned up company lists permissions related code.

## Test instructions

Nothing should have changed.

### DIFF
--- a/src/apps/companies/views/template.njk
+++ b/src/apps/companies/views/template.njk
@@ -10,7 +10,7 @@
 {% block local_header %}
   {% call LocalHeader({
     heading: company.name,
-    actions: action if canViewCompanyList,
+    actions: action if user.permissions.includes('company_list.add_companylist'),
     modifier: 'light-banner',
     fullWidthContent: true
   }) %}
--- a/src/apps/dashboard/controllers.js
+++ b/src/apps/dashboard/controllers.js
@@ -31,11 +31,8 @@ async function renderDashboard (req, res, next) {
       articleFeed = []
     }
 
-    const companyLists = userPermissions.includes('company_list.view_companylistitem') &&
-      await fetchCompanyLists(req.session.token)
-
     res.title('Dashboard').render('dashboard/views/dashboard', {
-      companyLists,
+      companyLists: await fetchCompanyLists(req.session.token),
       articleFeed,
       interactionsPermitted: isPermittedRoute(
         '/interactions',
--- a/src/apps/dashboard/controllers.js
+++ b/src/apps/dashboard/controllers.js
@@ -32,7 +32,8 @@ async function renderDashboard (req, res, next) {
     }
 
     res.title('Dashboard').render('dashboard/views/dashboard', {
-      companyLists: await fetchCompanyLists(req.session.token),
+      companyLists: userPermissions.includes('company_list.view_companylist') &&
+        await fetchCompanyLists(req.session.token),
       articleFeed,
       interactionsPermitted: isPermittedRoute(
         '/interactions',
--- a/src/apps/companies/middleware/__test__/params.test.js
+++ b/src/apps/companies/middleware/__test__/params.test.js
@@ -25,11 +25,7 @@ describe('Companies form middleware', () => {
     redirectSpy = sinon.spy()
     reqMock = { query: {}, session: { token: 2 } }
     resMock = {
-      locals: {
-        user: {
-          permissions: ['company_list.view_companylist'],
-        },
-      },
+      locals: {},
     }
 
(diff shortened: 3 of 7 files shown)


src/apps/companies/views/template.njk
```diff
@@ -10,7 +10,7 @@
 {% block local_header %}
   {% call LocalHeader({
     heading: company.name,
-    actions: action if canViewCompanyList,
+    actions: action if user.permissions.includes('company_list.add_companylist'),
```
web-bert: I feel like this should be set somewhere else, doing that check in the template seem logic it should not be doing? Also it would be good to add the permissions as constants somewhere instead of using strings.

src/apps/dashboard/controllers.js
```diff
@@ -34,11 +34,9 @@ async function renderDashboard (req, res, next) {
       articleFeed = []
     }
 
-    const companyLists = userPermissions.includes('company_list.view_companylistitem') &&
-      await fetchCompanyLists(req.session.token)
-
     res.title('Dashboard').render('dashboard/views/dashboard', {
-      companyLists,
+      companyLists: userPermissions.includes('company_list.view_companylist') &&
```
web-bert: same here would prefer this as a constant instead of a string for the permissions

src/apps/companies/views/template.njk
paulgain: It could be set in the controller and passed into the view.

peterhudec: It was in the controller before, but I noticed we have the whole user available in the template context so I thought it kind of redundant to assign it to a variable.

peterhudec: But I'll move both to the controller and put the string to a constant.

peterhudec: After searching for quite a while for the controller which is rendering this template, I have come to the conclusion that (correct me if I'm wrong) there is none. Most of the data seems to be added to `res.locals` in all sorts of middleware. I didn't feel like adding yet another one which would just add the `canViewCompanyList`, so I solved it in a different way, which still involves a new `permissions` middleware:

The problem with having permissions as an array of strings is that we then need those constants. I've added a middleware, which converts the array of permissions into a map of camel-cased permission names to `true` and assigns it to `res.user.can`, so that in a template you can do:
```
{% if user.can.companyListAddCompanylist %}
  ...
{% endif %}
```
This way we would never agian have to assign a `canSeeSomething` in a controller and test whether the controller passed it to render. We would also not need to define any constants. It does't involve any logic other than the semantics of a property access.

src/apps/dashboard/controllers.js
peterhudec: Replaced it with `res.locals.user.can.companyListVewCompanylist`.

src/middleware/permissions.js
```diff
@@ -0,0 +1,9 @@
+const { camelCase } = require('lodash')
+
+module.exports = (req, res, next) => {
+  res.locals.user.can = res.locals.user.permissions.reduce(
+    (a, x) => ({ ...a, [camelCase(x)]: true }),
```
web-bert: Looks good, can you use less cryptic variables names?

src/apps/companies/views/template.njk
peterhudec: Ok, so this template is only rendered when extended by other templates, so there's no controller to add it to, and I don't wanna create yet another middleware which attaches some arbitrary value to `req` or `res`, which is actually already there and the whole thing is accessible in any template anyway.

The real problem is that having permissions as a list of strings is not very practical. If it was an object e.g. `user.hasPermission.companyListAddCompanylist` it would be so much easier to work with and we would never again need to define `canDoSomething` variables in controlers just to be passed to a template.

I've implemented a middleware which does this, but will push it through a separate PR.

src/apps/dashboard/controllers.js
peterhudec: But if you think about it, having it in a constant would just move the problem to the constant, and you could still make a typo there.

Constants only make sense, when you need to share the same _flag_ in several places of your codebase and you don't really care about the actual value of the constant. But they won't help you at all when refering to external values. There's no way around this.

```js
const ENABLED = 'enbaled' // Oops, there is a typo here, but it still works

setSomeState(ENABLED)
// later on somewhere else
if getState() === ENABLED {
   doSomething()
}
```

src/middleware/permissions.js
peterhudec: I moved it to a separate PR. We can discuss the names there 😉

PR #2242, opened 2019-11-08
Added the /api proxy endpoint

peterhudec:
## Description of change

Added the `/api` endpoint, which proxies all requests to `config.apiRoot` with added `authorization: Bearer <session.token>` header. The `/api` prefix is stripped from the requests e.g a request to `localhost:3000/api/v4/company-list` will be proxied to `<apiRoot>/v4/company-list`.

## Test instructions

* You should be able to access the API from `/api` without the need to use the access token.
* You should not be able to access the API from `/api` if you are not logged in.

- [ x] Has the branch been rebased to develop?

### DIFF
--- a/package.json
+++ b/package.json
@@ -106,6 +106,7 @@
     "history": "^4.7.2",
     "hmpo-form-wizard": "^9.9.0",
     "html5shiv": "^3.7.3",
+    "http-proxy-middleware": "^0.20.0",
     "image-webpack-loader": "^4.0.0",
     "lodash": "^4.17.13",
     "markdown-it": "^8.4.0",
--- a/src/server.js
+++ b/src/server.js
@@ -110,10 +110,10 @@ app.use(store())
 app.use(csrf())
 app.use(csrfToken())
 
-app.use(proxy('/api', {
+app.use(proxy('/api-proxy', {
   target: config.apiRoot,
   pathRewrite: {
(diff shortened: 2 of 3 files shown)


rafenden:
This potentially might expose some data we don't want to be publicly accessible, could some BE devs have a look as well? @alixedi @currycoder @reupen @elcct

reupen:
I think, if we do this, we may want a whitelist of URL paths. E.g. we shouldn't expose `/admin/` through this.

peterhudec:
It turned out that there already is an `/api` endpoint, only accomodating `/api/options` and `api/postcodelookup`. I tried to move them to `/options` and `/postcodelookup`, but the tests are so tightly coupled to the folder structure with _proxyquire_ that I gave upa and just renamed my new endpoint to `/api-proxy`.

mcbhenwood:
We should resurrect this PR really.

I think the approach is sound but we also need a way to stop this leaking API endpoints. How about something like this:

- This endpoint adds an HTTP header to the API request, something like `X-Http-Datahub-Proxy-Version: 1`
- Create a decorator for API View methods, something like `@AllowFrontendProxiedCall`
- Add some middleware which rejects requests which contain that header but which are directed to views not wrapped in that decorator

That way you can switch on the backend methods one by one and by default they are off.

PR #2250, opened 2019-11-13
Remove the redundant company region and sector edit forms

rafenden:
## Description of change

Remove the redundant company's region and sector edit forms.

## Test instructions

Everything should work as before.


rafenden:
>All seems to still work. Looks like the project's overall code coverage takes a hit though blocking the PR, is this just because of the removal of tests bringing the % of tested lines down?

I think you are right here.

PR #2252, opened 2019-11-13
Show last interaction date

web-bert:
## Description of change

Show the last interaction date in the companies collection (and subsidiaries list)

## Test instructions

When a last interaction date is available
 it should come out under the address in the companies collection (or subsidiaries list)
 
### Before

Screen Shot 2019-11-13 at 19 23 46

### After 

Screen Shot 2019-11-13 at 19 23 12

## Checklist

### DIFF
--- a/src/apps/companies/transformers/__test__/company-to-list-item.test.js
+++ b/src/apps/companies/transformers/__test__/company-to-list-item.test.js
@@ -1,62 +1,64 @@
-const { assign, find } = require('lodash')
+const { find } = require('lodash')
 
-const companyData = require('~/test/unit/data/company')
-const transformCompanyToListItem = require('~/src/apps/companies/transformers/company-to-list-item')
+const companyData = require('../../../../../test/unit/data/company')
+const urls = require('../../../../lib/urls')
+const transformCompanyToListItem = require('../company-to-list-item')
 
 describe('transformCompanyToListItem', () => {
+  let listItem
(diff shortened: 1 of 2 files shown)


src/apps/companies/transformers/__test__/company-to-list-item.test.js
```diff
@@ -1,62 +1,64 @@
...
     })
 
     it('should return undefined', () => {
-      expect(this.listItem).to.be.undefined
+      expect(listItem).to.be.undefined
     })
   })
 
   context('when called with a fully populated company', () => {
     beforeEach(() => {
```
debitan: I think this can just be `before()`. `beforeEach()` would run `listItem = transformCompanyToListItem(companyData)` before each it statement. If I'm reading this right you can just run it once and use the same listItem for each of the its. 

https://stackoverflow.com/questions/21418580/what-is-the-difference-between-before-and-beforeeach

debitan: If you agree this is the case, it might be good to refactor the other `beforeEach()`s in the file. But appreciate this isn't your code and you are just removing the `this.` in this instance.

web-bert: @debitan I agree, I was more focussed on removing `this` instead of assessing the rest of the tests... I have now replaced `beforeEach` with `before`

PR #2253, opened 2019-11-14
Show new local nav in investments and add project breadcrumbs

debitan:
## Description of change

Currently breaks visual tests. Will work with Pippo to update these accordingly. 

The PR shows the local investment navigation tab on the investments index page. It also adds the `Projects` breadcrumb to projects. 

All tests that were touched are also refactored to remove unnecessary `beforeEach()`s and the like and the new urls module is used everywhere possible. 

## Test instructions

A local navigation tab on the investment projects page allowing you to switch between viewing projects and profiles. A new `Projects` breadcrumb is also visible on all projects and the projects index page. 
 
### Before

FireShot Capture 011 - Investments - DIT Data Hub - localhost

### After 

FireShot Capture 019 - Projects - Investments - DIT Data Hub - localhost

## Checklist


src/apps/investments/controllers/__test__/projects.test.js
```diff
@@ -14,8 +17,8 @@ describe('Investment project controller', () => {
     ],
   }
 
-  beforeEach(() => {
-    this.req = {
+  before(() => {
```
peterhudec: Changing `beforeEach` to `before` is nut much of an improvement. All the tests are still entangled by the `let req, res, next`. Here's an example how it can be written with no _befores_ and _lets_. Notice how I moved all the `it(msg)` messages to `expect(expression, msg)`, which is where it really belongs, because all the `it()` calls were actually just assertions about the same state. This reduces 11 tests to just 2.
```js
const config = require('../../../../config')
const { renderProjectsView } = require('../../controllers/projects')

const fixture = statusCode => {
  const metadataMock = {
    sectorOptions: [{ id: 's1', name: 's1', disabled_on: null }],
    adviserOptions: {
      results: [
        { id: 'ad1', name: 'ad1', is_active: true, dit_team: { name: 'ad1' } },
      ],
    },
    countryOptions: [
      { id: '9999', name: 'United Kingdom' },
      { id: '8888', name: 'Test' },
    ],
  }

  nock(config.apiRoot)
    .get('/v4/metadata/sector?level__lte=0')
    .reply(statusCode, metadataMock.sectorOptions)
    .get('/v4/metadata/country')
    .reply(statusCode, metadataMock.countryOptions)
    .get('/adviser/?limit=100000&offset=0')
    .reply(statusCode, metadataMock.adviserOptions)

  return {
    req: {
      session: {
        user: {
          id: '1234',
          name: 'Fred Smith',
          permissions: [],
        },
        token: 'abcd',
      },
      query: {
        sortby: 'estimated_land_date:asc',
      },
    },
    res: {
      render: sinon.spy(),
      breadcrumb: sinon.stub().returnsThis(),
      query: {},
      locals: {
        userAgent: {
          isIE: false,
        },
      },
    },
    next: sinon.spy(),
  }
} 

describe('#renderProjectsView', () => {
  context('when the list renders successfully', () => {
    it('Should render the page', async () => {
      const { req, res, next } = fixture(200)

      await renderProjectsView(req, res, next)

      expect(res.render, 'should render').to.be.calledOnce
      expect(
        res.breadcrumb,
        'should render the breadcrumbs',
      ).to.be.calledWith('Projects')
      expect(
        res.render.firstCall.args[0],
        'should render the collection template',
      ).to.equal('investments/views/projects')
      expect(
        res.render.firstCall.args[1].title,
        'should render the view with a title',
      ).to.equal('Investment Projects')
      expect(
        res.render.firstCall.args[1].countLabel,
        'should render the view with a count label',
      ).to.equal('project')
      expect(
        res.render.firstCall.args[1].sortForm,
        'should render the view with a sort form',
      ).to.exist
      expect(
        res.render.firstCall.args[1].selectedFilters,
        'should render the view with selected filters',
      ).to.exist
      expect(
        res.render.firstCall.args[1].exportAction,
        'should render the view with an export action',
      ).to.deep.equal({ enabled: false })
      expect(
        res.render.firstCall.args[1].filtersFields,
        'should render the view with filter fields',
      ).to.exist
    })
  })

  context('When there is an error', () => {
    it('Should propagate the error', async () => {
      const { req, res, next } = fixture(500)

      await renderProjectsView(req, res, next)

      expect(res.render).to.be.thrown
      expect(next, 'should call next with an error').to.have.been.calledOnce
    })
  })
})

```

test/functional/cypress/specs/interaction/add-interaction-spec.js
```diff
@@ -2,20 +2,21 @@ const fixtures = require('../../fixtures')
 const selectors = require('../../../../selectors')
```
peterhudec: Replacing `beforeEach` to `before` would make for quite a performance gain especially in this file. It is fine to use `before` in Cypress tests, as we don't need to share variables between assertions, so we still can use `it()`, because the state is the actual DOM.

src/apps/investments/views/projects.njk
```diff
@@ -1,6 +1,13 @@
 {% extends "_layouts/template.njk" %}
 
 {% block content %}
+  <div id="investors-profiles-tab">
+    <div class="govuk-grid-row">
+      <div class="govuk-grid-column-full">
+        {{ TabbedLocalNav({ items: investmentTabItems }) }}
+      </div>
+  </div>
+
```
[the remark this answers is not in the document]
debitan: You're right that this component does need to go into React at some point, but I haven't added or changed anything here, the work was already done and I'm just exposing it.

test/functional/cypress/specs/interaction/add-interaction-spec.js
debitan: I've tried to do this but the tests break if I replace `beforeEach` with `before`. Something about the form being tampered with.

PR #2254, opened 2019-11-14
Add global tests helpers

web-bert:
To save long paths back to the test folder add some helpers to the global see files for example

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -31,7 +31,8 @@
         "nock": true,
         "rootPath": true,
         "globalReq": true,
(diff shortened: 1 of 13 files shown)


web-bert:
@rafenden We can't use anything relating to webpack to solve this problem, so that removes them from the equation.
Part of the point of this is to remove a dependency, replacing it with another one doesn't seem to make sense.

If the goal is to have short paths then I don't see how using anything to get paths from the root solves this problem, if you look at the changes on this PR it shortens all the paths...

PR #2264, opened 2019-11-19
Permissions as a map

peterhudec:
## Description of change

Added the `permissions` middleware, which adds the `hasPermission` property to `res.locals.user`. The value is a map of camel cased permissions to `true`:

```js
res.user.hasPermission = {
  investmentViewAllInvestmentproject: true,
  orderExportOrder: true,
  companyViewCompanieshousecompany: true,
  // ...
}
```

The point is that `res.locals` is merged to nunjucks context and thus is available in all templates. So we don't actually need to define a `canDoThisAndThat` variable everytime we implement a controller just to pass it to the template, let alone test the various states. It's just that the form of an array of strings makes it a bit unwieldy and verbose and makes one think that there is some logic involved which doesn't belong to a template.
```nunjucks
{% if user.permissions.includes('order.change_order') %}
   <h1>Change Order</h1>
{% endif %}
```
It would be much more practical to represent the permissions as a map of names to `true`. The only problem is that our permission names contain dots, we can't really access them with the dot notation unless we camel case them:
```nunjucks
{% if user.hasPermission.orderChangeOrder %}
   <h1>Change Order</h1>
{% endif %}
```
It is worth noting, that there really is no semantic difference whatsoever between `['something'].includes('something')` and `({something: true}).something`. Both are forgiving if you misspell the value/name. The main difference is expressiveness, especially when using the dot notation access in Nunjucks.


src/apps/companies/views/template.njk
```diff
@@ -10,7 +10,11 @@
 {% block local_header %}
   {% call LocalHeader({
     heading: company.name,
+<<<<<<< HEAD
     actions: action if canViewCompanyList,
+=======
+    actions: action if user.can.companyListAddCompanylist,
+>>>>>>> 948488a7... Added the permissions middleware
     modifier: 'light-banner',
```
paulgain: You have merged conflicts that haven't been resolved.

peterhudec: Damit

peterhudec:
What's wrong with the codecow? It's all covered as far as I know.

src/middleware/permissions.js
```diff
@@ -0,0 +1,14 @@
+const { camelCase, get } = require('lodash')
+
+module.exports = (req, res, next) => {
+  if (get(res, 'locals.user.permissions')) {
```
rafenden: You could use `has` instead.

peterhudec:
> I have a mixed feeling about this change as it obfuscates the real permission name with a camelCase version.

That's a valid point. The conversion wouldn't be needed if the permissions didn't have a dot in their name, which makes working with them a pain in the ass in JS and Nunjucks. Maybe a better API would be `user.hasPermission('foo.bar')`? In any case, I don't think we shoud pollute the controllers with them.

web-bert:
@rafenden @peterhudec what about splitting on the `.` and creating an object for each one, so it could be `if (user.hasPermission.foo.bar)`?

rafenden:
@peterhudec `user.hasPermission('foo.bar')` seems to be a better option as it uses the original permission name used across the app.

PR #2265, opened 2019-11-19
Add Lead adviser information in company header

ian-leggett:
## Description of change
If a company has an allocated Lead adviser this information should be present in the company header.

## Test instructions
View a company that has an allocated account manager
Screenshot 2019-11-19 at 15 41 52

## Checklist

### DIFF
--- a/src/apps/companies/middleware/__test__/params.test.js
+++ b/src/apps/companies/middleware/__test__/params.test.js
@@ -1,5 +1,6 @@
 const proxyquire = require('proxyquire')
 const company = require('../../../../../test/unit/data/companies/company-v4.json')
+const oneListTypeDItaCompany = require('../../../../../test/unit/data/companies/one-list-group-tier-d-ita.json')
 
 describe('Companies form middleware', () => {
   let nextSpy
@@ -14,7 +15,7 @@ describe('Companies form middleware', () => {
   let resMock
   let middleware
 
-  beforeEach(() => {
+  before(() => {
     nextSpy = sinon.spy()
     getDitCompanyStub = sinon.stub()
     getDitCompanyFromListStub = sinon.stub()
@@ -42,19 +43,45 @@ describe('Companies form middleware', () => {
   })
 
   describe('getCompany', () => {
-    beforeEach(async () => {
-      getDitCompanyStub.resolves(company)
-      await middleware.getCompany(reqMock, resMock, nextSpy, 2)
-    })
+    context('when the company is not a One list Tier D ITA', () => {
+      before(async () => {
+        getDitCompanyStub.resolves(company)
+        await middleware.getCompany(reqMock, resMock, nextSpy, 2)
+      })
 
-    it('should return the company', () => {
-      expect(resMock.locals).to.have.deep.property('company', company)
+      it('should return the company', () => {
+        expect(resMock.locals).to.have.deep.property('company', {
+          ...company,
+          isItaTierDAccount: false,
+          hasAllocatedLeadIta: true,
+          hasManagedAccountDetails: true,
+          isGlobalHQ: null,
+          isUltimate: false,
+        })
+      })
+    })
+    context('when the company is a One list Tier D ITA', () => {
+      before(async () => {
+        getDitCompanyStub.resolves(oneListTypeDItaCompany)
+        await middleware.getCompany(reqMock, resMock, nextSpy, 2)
+      })
+
+      it('should return the company', () => {
+        expect(resMock.locals).to.have.deep.property('company', {
+          ...oneListTypeDItaCompany,
+          isItaTierDAccount: true,
+          hasAllocatedLeadIta: true,
+          hasManagedAccountDetails: true,
+          isGlobalHQ: null,
+          isUltimate: false,
+        })
+      })
     })
   })
 
   describe('setIsCompanyAlreadyAdded', () => {
     context('when the company is already added to the user company list', async () => {
-      beforeEach(async () => {
+      before(async () => {
         getDitCompanyFromListStub.resolves()
         await middleware.setIsCompanyAlreadyAdded(reqMock, resMock, nextSpy, 2)
       })
@@ -65,7 +92,7 @@ describe('Companies form middleware', () => {
     })
 
     context('when the company is not added to the user company list', async () => {
-      beforeEach(async () => {
+      before(async () => {
         getDitCompanyFromListStub.rejects()
         await middleware.setIsCompanyAlreadyAdded(reqMock, resMock, nextSpy, 2)
       })
@@ -78,7 +105,7 @@ describe('Companies form middleware', () => {
 
   describe('addCompanyOrRemoveFromList', () => {
     context('when adding a company to the user company list', async () => {
-      beforeEach(async () => {
+      before(async () => {
         resMock = {
           ...resMock,
           locals: {
@@ -106,7 +133,7 @@ describe('Companies form middleware', () => {
       })
     })
     context('when removing a company from the user company list', async () => {
-      beforeEach(async () => {
+      before(async () => {
         resMock = {
           ...resMock,
           locals: {
--- a/src/apps/companies/middleware/params.js
+++ b/src/apps/companies/middleware/params.js
@@ -5,9 +5,18 @@ const {
   removeDitCompanyFromList,
 } = require('../repos')
 
+const { isItaTierDAccount } = require('../../../lib/is-tier-type-company')
+
 async function getCompany (req, res, next, id) {
   try {
-    res.locals.company = await getDitCompany(req.session.token, id)
+    const company = await getDitCompany(req.session.token, id)
+    company.isItaTierDAccount = isItaTierDAccount(company)
+    company.hasAllocatedLeadIta = company.one_list_group_global_account_manager != null
+    company.hasManagedAccountDetails = company.one_list_group_tier && company.hasAllocatedLeadIta
+    company.isUltimate = !!company.is_global_ultimate && res.locals.features['companies-ultimate-hq']
+    company.isGlobalHQ = company.headquarter_type && company.headquarter_type.name === 'ghq'
+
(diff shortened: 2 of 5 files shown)


src/apps/companies/middleware/__test__/params.test.js
```diff
@@ -46,13 +47,31 @@ describe('Companies form middleware', () => {
   })
 
   describe('getCompany', () => {
-    beforeEach(async () => {
-      getDitCompanyStub.resolves(company)
-      await middleware.getCompany(reqMock, resMock, nextSpy, 2)
+    context('when the company is not a One list Tier D ITA', () => {
+      beforeEach(async () => {
```
debitan: could this be just `before()`?

And I notice there are a number of other `beforeEach()` in the other tests in this file that you haven't touched. Appreciate it might be quite a lot of extra work to refactor the whole file.

src/apps/companies/middleware/__test__/params.test.js
```diff
@@ -46,13 +47,31 @@ describe('Companies form middleware', () => {
...
+
+      it('should return the company', () => {
+        expect(resMock.locals).to.have.deep.property('company', {
+          ...company,
+          isItaTierDAccount: false,
+        })
+      })
     })
+    context('when the company is a One list Tier D ITA', () => {
+      beforeEach(async () => {
```
debitan: same here

src/apps/companies/views/template.njk
```diff
@@ -26,20 +29,19 @@
...
         {% endif %}
-        {% if isOneListTier %}
+
+
           <p>This is an account managed company (One List {{ company.one_list_group_tier.name }})</p>
           <p>
-            Global Account Manager: {{ company.one_list_group_global_account_manager.name }}
-            <a href={{ urls.companies.advisers.index(company.id) }}>View core team</a>
+            {{ "Lead ITA" if company.isItaTierDAccount else "Global Account Manager" }}: {{ company.one_list_group_global_account_manager.name }}
+            <a href={{ urls.companies.advisers.index(company.id) }}>{{ "View Lead adviser" if company.isItaTierDAccount else "View core team" }}</a>
```
debitan: This is going to be a really annoying comment, as I'm sure you've already considered it, but I just thought that maybe we should be refactoring this `local-header__description` into a React component? I guess by just adding to the nunjucks template we are creating future work for ourselves in moving to a SPA.

src/apps/companies/middleware/__test__/params.test.js
```diff
@@ -1,5 +1,6 @@
 const proxyquire = require('proxyquire')
 const company = require('../../../../../test/unit/data/companies/company-v4.json')
+const oneListTypeDItaCompany = require('../../../../../test/unit/data/companies/one-list-group-tier-d-ita.json')
```
debitan: I don't know as I haven't tried replacing it locally, but we might be able to just pass a string instead of all this mock data as the test, as I understand it, is just testing if the thing you pass it is the thing that you passed it. (Thinking back to my lightbulb moment the other day!)

src/apps/companies/views/template.njk
ian-leggett: Good point I am happy to do this refactor if I get the go-ahead. This is going to take more time so I need to talk to Billy on this one. @mcbhenwood do you have any input on this? Ideally the whole header needs splitting up into React components.

src/apps/companies/middleware/__test__/params.test.js
ian-leggett: Before I modified this test I would say you are right however there is logic around certain props in these files such as `one_list_group_tier` and `one_list_group_global_account_manager` and depending on what values these have they get decorated with a `isItaTierDAccount: false` or a `isItaTierDAccount: true`

src/apps/companies/middleware/__test__/params.test.js
ian-leggett: No worries good spot, I need to get into the habit of removing the unnecessary code we have dotted all around the place. All refactored 👍

src/apps/companies/middleware/__test__/params.test.js
debitan: No, that's fair enough 👍

src/apps/companies/views/template.njk
```diff
@@ -1,4 +1,6 @@
 {% extends "_layouts/template.njk" %}
+{% set hasAllocatedLeadIta = company.one_list_group_global_account_manager != null %}
+{% set hasManagedAccountDetails = company.one_list_group_tier and hasAllocatedLeadIta %}
```
web-bert: Can you move these to `getCompany` in `src/apps/companies/middleware/params.js` ?

src/apps/companies/views/template.njk
```diff
@@ -17,6 +19,7 @@
     <p class="c-local-header__heading-after">{{ company.address | formatAddress }}</p>
     {% set isUltimate = company.is_global_ultimate and features["companies-ultimate-hq"] %}
     {% set isGlobalHQ = company.headquarter_type and company.headquarter_type.name == 'ghq' %}
+
     {% set isOneListTier = company.one_list_group_tier %}
```
web-bert: Same here, can they go in `getCompany`?

test/functional/cypress/specs/companies/lead-advisers.spec.js
```diff
@@ -57,6 +57,12 @@ describe('Lead advisers', () => {
         'Lead adviser': null,
       })
     })
+    it('should show the allocated Lead adviser in the company header', () => {
+      cy.get(selectors.localHeader().description.paragraph(2)).should('contain', 'Lead ITA: Travis Greene')
+    })
+    it('should have a link to the Lead adviser tab', () => {
+      cy.contains('View Lead adviser').invoke('attr', 'href').should('eq', `/companies/${fixtures.company.oneListTierDita.id}/advisers`)
```
web-bert: Use `urls`

src/apps/companies/middleware/params.js
```diff
@@ -8,10 +8,15 @@ const {
...
+const { isItaTierDAccount } = require('../../../lib/is-tier-type-company')
 
 async function getCompany (req, res, next, id) {
   try {
-    res.locals.company = await getDitCompany(req.session.token, id)
+    const company = await getDitCompany(req.session.token, id)
+    res.locals.company = {
+      ...company,
+      isItaTierDAccount: isItaTierDAccount(company),
+    }
```
web-bert: I would question the need for the spread here, you can just add the property to the `company` and then assign it to `res.locals.company`

src/apps/companies/views/template.njk
ian-leggett: We also need to think about how we deal with the global variables in React. That file above has access to anything on `res.locals` so this works but we need to look at how we do the above in React 🤔

mcbhenwood: I think we should put the work to convert this onto the tech backlog and then look to do that as a priority in an upcoming sprint.

The mandate to "put new stuff in React" is sensible because we don't want to create new items that we know we will have to rework. However... When we are amending the operation of an existing component to some degree (for some value of "some" _[super-meta]_ ) then I don't think it's fair on the feature work to weigh it down with the React conversion.

Also there will be deeper considerations (`res.locals` is a good example), the resolution of which don't belong in feature work either.

I know some people like to apply a kind of "Common Law" approach to PR comments and may be tempted to take this as a green light to throw out requests to do React conversion work _ever_ as part of a feature PR but I don't think that's realistic either. We need to consider these on a case-by-case basis [80% of tech strategy is an ability to prioritise clearly] and _in this case_ this conversion does not need to be done.

src/apps/companies/middleware/params.js
ian-leggett: All refactored 👍

src/apps/companies/middleware/params.js
```diff
@@ -8,10 +8,14 @@ const {
...
 
 const { getAllCompanyLists } = require('../../company-lists/repos')
+const { isItaTierDAccount } = require('../../../lib/is-tier-type-company')
 
 async function getCompany (req, res, next, id) {
   try {
     res.locals.company = await getDitCompany(req.session.token, id)
+    res.locals.company.isItaTierDAccount = isItaTierDAccount(res.locals.company)
+    res.locals.company.hasAllocatedLeadIta = res.locals.company.one_list_group_global_account_manager != null
+    res.locals.company.hasManagedAccountDetails = res.locals.company.one_list_group_tier && res.locals.company.hasAllocatedLeadIta
```
web-bert: I would assign the response to `company` and then have one statement at the end to assign that to `res.locals.copmany` it would make this more DRY

src/apps/companies/views/template.njk
web-bert: It doesn't look as if this has changed?

src/apps/companies/middleware/__test__/params.test.js
```diff
@@ -141,8 +164,18 @@ describe('Companies form middleware', () => {
...
   describe('setDoAnyListsExist', () => {
+    before(() => {
+      reqMock = { query: {}, session: { token: 2 } }
+      resMock = {
+        locals: {
+          user: {
+            permissions: ['company_list.view_companylist'],
+          },
+        },
+      }
```
rafenden: You could use `test/unit/helpers/middleware-parameters-builder.js` here to make the code more compact, e.g.

```
const buildMiddlewareParameters = require('~/test/unit/helpers/middleware-parameters-builder')
const middlewareParameters = buildMiddlewareParameters({
  user: { permissions: ['company_list.view_companylist'] },
})
```

ian-leggett: I did refactor this but after rebasing its been removed by Peter anyway.

src/apps/companies/middleware/params.js
ian-leggett: Refactored 👍

PR #2271, opened 2019-11-22
Move Sandbox from its own repo to the frontend codebase

rafenden:
## Description of change

Moving sandbox code from https://github.com/uktrade/data-hub-sandbox to this repo to avoid context switching.


rafenden:
> hey @rafenden won't this stop us from being able to merge without code reviews? This was one of the advantages of having it externally, allowing us to move quickly in putting up fixtures etc.

@debitan You can still create and tests the fixtures locally when working on a new functionality, do you have any other use case on mind?

debitan:
> > hey @rafenden won't this stop us from being able to merge without code reviews? This was one of the advantages of having it externally, allowing us to move quickly in putting up fixtures etc.
> 
> @debitan You can still create and tests the fixtures locally when working on a new functionality, do you have any other use case on mind?

not as such. I just remember @PippoRaimondi telling me that the no reviews thing was done on purpose to make the sandbox lightweight. I presume you've discussed the move of sandbox into frontend with him though right?

rafenden:
> not as such. I just remember @PippoRaimondi telling me that the no reviews thing was done on purpose to make the sandbox lightweight. I presume you've discussed the move of sandbox into frontend with him though right?

@debitan We use sandbox solely for functional tests, so it makes sense to keep these in the same repo, benefits are clear:
* Functional tests will never break because someone changed a fixture without telling anyone.
* Sandbox fixtures could be shared with different tests, e.g. unit tests.
* Sandbox fixtures will now included in the review process which will increase transparency and will bundle the tests with the fixtures in a single PR.
* There will be no need to change context as the fixtures will be in the same codebase as the related tests.
* CI will be slightly faster as it doesn't need to spin up additional docker instance.

Haven't spoked to @PippoRaimondi yet as he seems to be off today.

mcbhenwood:
So let's not go ahead with this until at least @PippoRaimondi is back next week.

Also I don't know where this idea was discussed (if it was discussed) but I should very much like to be involved in future discussions of this nature before they get to PR.

PR #2272, opened 2019-11-22
Add support for testing JSX files using Mocha

rafenden:
## Description of change

Enable testing of JSX files using Mocha runner.


PippoRaimondi:
@rafenden looks good but there are a couple of unit tests failing with some odd errors, which I assume it could be related to some of the package updates?

rafenden:
Marking it as blocked for now.

PR #2273, opened 2019-11-22
Replace wavy with module-alias

rafenden:
## Description of change

Use `module-alias` in place of `wavy` package which is causing us a lot of trouble as it relies on symlinks.

## Test instructions

Everything should work as before.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -100,7 +100,6 @@ const common = {
     ],
     alias: {
       'vue$': 'vue/dist/vue.common.js',
-      '~': __dirname,
     },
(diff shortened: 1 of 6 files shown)


webpack.config.js
```diff
@@ -100,6 +100,7 @@ const common = {
     ],
     alias: {
       'vue$': 'vue/dist/vue.common.js',
+      '~': __dirname,
```
web-bert: Should we use this for webpack?

rafenden: We don't need indeed. This change was removed, thanks!

rafenden:
Superseded by https://github.com/uktrade/data-hub-frontend/pull/2286

PR #2276, opened 2019-11-25
Add "Strict-Transport-Security" and "X-XSS-Protection" headers

rafenden:
## Description of change

Increase security by adding two additional HTTP headers.

## Test instructions

Check if the headers are included in the response.


rafenden:
>We should consider using https://helmetjs.github.io/

@paulgain That's a great solution but I would rather prefer (for now) to implement these two small changes and consider upgrading to helmet in the near future.

PR #2278, opened 2019-11-28
Add the 'add interaction' button to the companies table

ian-leggett:
## Description of change

Updating Data Hub Components which will add the "add interaction" button to each company listed in your companies list.

## Test instructions

Follow a company by adding it to a list then the button "add interaction" should show to the right of each company on the home page..

## Screenshots
Screenshot 2019-11-28 at 09 58 37

## Checklist


src/apps/dashboard/client/MyCompanies.jsx
```diff
@@ -6,6 +6,9 @@ const MyCompanies = (props) =>
     deleteListPropsAccessor={list =>
       ({ href: `/company-lists/${list.id}/delete` })
     }
+    addInteractionPropsAccessor={(company) => ({
+      href: `/companies/${company.id}/interactions/create`,
```
rafenden: Non-critical: For symmetry, I would suggest keeping the same formatting as `deleteListPropsAccessor`.

test/functional/cypress/specs/company-lists/view-spec.js
```diff
@@ -8,7 +8,7 @@ const expectedRows = {
       text: '14 Aug 2019',
     },
     {
-      text: 'Here is a long interaction title some more text some more text some more text almost finished some more text nearly there more text finished',
+      text: 'Here is a long interaction',
```
rafenden: Are we not asserting for existence of ellipsis?

ian-leggett: I was speaking with @PippoRaimondi about this and feel that this is very hard to test, we now have more cells in the table so depending on screen size the ellipses will be in different places so its very hard to test. I think that CSS should deal with ellipses rather than React, if it was CSS doing the ellipses I could make an assertion around all the text, as its React the text is removed from the DOM and it can be quite random. So what I have done is just check for some of the text for now.

src/apps/dashboard/client/MyCompanies.jsx
ian-leggett: Good spot worth doing 👍

test/functional/cypress/specs/company-lists/view-spec.js
rafenden: So perhaps, we should replace the mechanism for ellipsis in this PR?

ian-leggett: I agree I do think it needs updating as the ellipse React wrapper doesn't update on window re-size too. Can this be iterated from this PR? I would need to create another PR for the ellipse fix in Data Hub components.

ian-leggett: As for the assertion for the `...` I can't as its in a different place in CI than local (different resolution)

rafenden: I'm fine with a separate PR, regarding the assertion, you could add one more test case just for the existence of the `...` string, i.e.

```
.should('contain', '...')
```

mcbhenwood: From what @ian-leggett was saying, it's not possible to test for the dots because these are injected by CSS and not available in the DOM.

I therefore recommend that we leave the ellipsis untested since really we would only be testing an imported implementation which we oughtn't do.

When what we're really testing is "did we use a particular externally-tested component" then if it is _hard_ to test then I favour the practice of knowing that we implemented the plugin / component / whatever and _if it presents buggy behaviour_ then only then do we invest effort in putting in a test to detect its operation. Otherwise, fire and forget. What's the worst that can happen? As long as we have tests that verify that the text renders under normal circumstances, that seems to me to be an acceptable level of risk.

Let's get some ticks on this and get it through. It's a week old and only ~50 lines.

mcbhenwood: I agree @rafenden - Just add a test which puts in a big long string and then verifies that it ends in `...`. That tests that "some component" has been deployed which manipulates the provided string but does not test exactly how that component works (which is correct since it's an imported component).

So in short we don't care about _where_ the string is truncated; we just care that it is truncated at some point and ends in dots.

ian-leggett: So just added tests around whats been discussed above. We really need to revisit the way we do truncation of long text content within the table cells, this should be done with CSS and cover this type of thing with visual regression tests. I have added a "TODO" in the test so when this is done it is clear what part of the test needs removing.

test/functional/cypress/specs/company-lists/view-spec.js
```diff
@@ -195,28 +180,24 @@ const describeSortListBy = ({ option, rows }) =>
...
     describeTable(rows)
   })
 
-const describeSortList = (rows) => {
+const describeSortList = rows => {
   const recentRows = [...rows].sort(([, { text: a }], [, { text: b }]) =>
-    b === '-'
-      ? -1
-      : new Date(b) - new Date(a)
+    b === '-' ? -1 : new Date(b) - new Date(a)
```
peterhudec: I always line-break ternaries even if they fit into a single line, as they are much more readable that way.

ian-leggett: Just reformatted 👍

PR #2285, opened 2019-11-29
Fix for when a user edits a Client Relationship Manager

paulgain:
## Fixes a known issue within Investment Projects

Previously this bug manifested itself when attempting to change a Client Relationship Manager (CRM) within Investment Project / teams. To reproduce the issue you have to remove any existing CRM within the admin tool.

https://datahub-api-staging.london.cloudapps.digital/admin/investment/investmentproject/b30dee70-b2d6-48cf-9ce4-b9264854470c/change/

Attempt to change the CRM here:

https://www.datahub.staging.uktrade.io/investments/projects/b30dee70-b2d6-48cf-9ce4-b9264854470c/team

error

## Worth noting
In JavaScript, function parameters default to `undefined`. However, it's often useful to set a different default value. For example an array:

`const method = (a=[]) => { return a }`

It's important to point out that `null` like other falsey values are honoured:

| Function call  | Returns  |
| ------------- |:--------:|
| method(undefined) | [] |
| method('Hello') |   "Hello"  |
| method(['hi', 'again'])  |  ["hi", "again"]  |
| method('') |   ''  |
| method(null) |   null  |

You cannot rely that the argument `a` will always default to an empty array. 

That was the root cause of this bug where a developer depended on `includeAdviser ` never being `null`.

## Checklist

### DIFF
--- a/src/apps/adviser/filters.js
+++ b/src/apps/adviser/filters.js
@@ -1,12 +1,17 @@
-const { get, includes } = require('lodash')
+const {
+  get,
+  includes,
+  castArray,
+  compact,
+} = require('lodash')
 
-function filterActiveAdvisers ({ advisers = [], includeAdviser = [] }) {
-  includeAdviser = [...includeAdviser]
+function filterActiveAdvisers ({ advisers = [], includeAdviser }) {
+  let adviserIds = compact(castArray(includeAdviser))
   return advisers.filter(adviser => {
     // This returns all active users plus the include advisor (whether they are active or not)
     return (
       get(adviser, 'is_active', true) ||
-      includes(includeAdviser, adviser.id)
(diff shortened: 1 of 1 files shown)


src/apps/adviser/filters.js
```diff
@@ -1,12 +1,21 @@
...
+const {
+  get,
+  includes,
+  isString,
+  castArray,
+  compact,
+} = require('lodash')
+
+function filterActiveAdvisers ({ advisers = [], includeAdviser }) {
+  let adviserIds = isString(includeAdviser)
```
rafenden: We could slightly simplify this line, i.e.
```
const adviserIds = compact(castArray(includeAdviser))
```

paulgain: 👍

PR #2281, opened 2019-11-28
Add countries when adding an interaction - Part 1

web-bert:
## Description of change

**Hidden behind a feature flag called "interaction-add-countries"** so this PR should have no changes, unless this feature flag is present - this is to help make the PRs for this feature smaller.

To capture more up to date information about future countries of interest, countries currently exporting to or countries a company is not interested in exporting to, this adds a new question when adding an interaction for Export and Other "themes" to ask if any countries were discussed. If countries were discussed then three additional fields are revealed to ask for a list of countries for 

- Future countries of interest
- Countries currently exporting to
- Countries not interested in

Each of these fields is a Typeahead and there were a few modifications required to make the typeahead work in this scenario.

## Test instructions

Add a new interaction with a theme/type of "Export" or "Other" with any sub type selected, you should see a new question at the bottom asking "Were any countries discussed?"

_What should I see?_
 
### Before

Screenshot 2019-11-28 at 19 18 24

### After 

Screenshot 2019-11-28 at 19 18 34

## Checklist

[//]: # "When submitting a PR make sure the code review guidelines have been satisfied. 
https://github.com/uktrade/data-hub-frontend/blob/develop/docs/Code%20review%20guidelines.md"

- [x] Has the branch been rebased to develop?
- [x] Automated tests (Any of the following when applicable: Unit, Functional or Acceptance)
- [ ] Manual compatibility testing (Browsers: Chrome, Firefox, IE11, Safari)

PR #2307, opened 2019-12-05
Add countries when adding an interaction - Part 2 (save)

web-bert:
## Description of change

Building on the [part 1 PR](https://github.com/uktrade/data-hub-frontend/pull/2281) this now sends the data to the API and handles error scenarios - **still behind a feature flag**

We have three fields for the countries that are combined into one for the API, so when an error is returned in the 400 response we have to map it to a real field. I have tried several solutions to inject the error message between the "Were countries discussed" question and the three country fields, but each attempt has failed. Therefore I have now just mapped the `export_countries` errors to `were_countries_discussed` so the error gets displayed. This will mean I can implement the rest of the feature and I will come back to the error message when time permits.

**I have not refactored the tests to remove `this` as I want to reduce the noise in this PR, I will make an additional PR to refactor tests**

## Test instructions

When the feature flag is enabled an extra question for "Export" and "Other" interactions to ask "Were countries discussed"
 
### Before

Screenshot 2019-12-05 at 12 18 03

### After 

Screenshot 2019-12-05 at 12 17 30

## Checklist

### DIFF
--- a/src/apps/interactions/constants.js
+++ b/src/apps/interactions/constants.js
@@ -118,6 +118,12 @@ const KINDS = {
   SERVICE_DELIVERY: 'service-delivery',
 }
 
+const EXPORT_INTEREST_STATUS = {
+  FUTURE_INTEREST: 'future_interest',
+  EXPORTING_TO: 'currently_exporting',
+  NOT_INTERESTED: 'not_interested',
+}
+
 module.exports = {
   QUERY_FIELDS,
   QUERY_DATE_FIELDS,
@@ -132,4 +138,6 @@ module.exports = {
   INTERACTION_CONTEXTS,
   THEMES,
   KINDS,
+  EXPORT_INTEREST_STATUS,
+  EXPORT_INTEREST_STATUS_VALUES: Object.values(EXPORT_INTEREST_STATUS),
 }
--- a/src/apps/interactions/controllers/edit.js
+++ b/src/apps/interactions/controllers/edit.js
@@ -11,7 +11,7 @@ const { joinPaths } = require('../../../lib/path')
 const isValidTheme = require('../macros/is-valid-theme')
 const canAddCountries = require('../macros/can-add-countries')
 
-const { KINDS, THEMES } = require('../constants')
+const { KINDS, THEMES, EXPORT_INTEREST_STATUS_VALUES } = require('../constants')
 
 const formConfigs = {
   [KINDS.INTERACTION]: interactionForm,
@@ -146,11 +146,7 @@ async function renderEditPage (req, res, next) {
       : `Add ${kindName + forEntityName}`
 
     if (canAddCountries(theme, res.locals.features)) {
-      [
-        'future_countries',
-        'export_countries',
-        'no_interest_countries',
-      ].forEach(addSelectedOptions(interactionForm.children))
+      EXPORT_INTEREST_STATUS_VALUES.forEach(addSelectedOptions(interactionForm.children))
     }
 
     res
--- a/src/apps/interactions/controllers/edit.js
+++ b/src/apps/interactions/controllers/edit.js
@@ -146,6 +146,7 @@ async function renderEditPage (req, res, next) {
       ? `Edit ${kindName}`
       : `Add ${kindName + forEntityName}`
 
+    // istanbul ignore next: Covered by functional tests
     if (canAddCountries(theme, res.locals.interaction, res.locals.features)) {
       EXPORT_INTEREST_STATUS_VALUES.forEach(addSelectedOptions(interactionForm.children))
     }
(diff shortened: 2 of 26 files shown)


src/apps/interactions/controllers/edit.js
```diff
@@ -46,6 +46,7 @@ async function buildForm (req, res, params) {
       : `Add ${lowerCase(kind)}`,
     company: get(res.locals, 'company.name'),
     theme,
+    interaction: res.locals.interaction,
```
ian-leggett: Optional: Is it worth deconstructing `interaction` and `features` here. You could have then omit the interaction value and just have something like
```
{
interaction,
featureFlags: features,
}

File: src/apps/interactions/macros/can-add-countries.js [hunk 2]
```diff
@@ -1,8 +1,9 @@
 const { THEMES } = require('../constants')
 
-module.exports = (theme, featureFlags) => {
+module.exports = (theme, interaction, featureFlags) => {
   const featureEnabled = featureFlags[ 'interaction-add-countries' ]
 
+  if (interaction) { return false }
```
ian-leggett: Optional: Could you do
```if (interaction || !featureEnabled) return false```

File: src/apps/interactions/macros/fields.js [hunk 3]
```diff
@@ -230,15 +232,15 @@ module.exports = {
       options: channels,
     }
   },
-  countriesDiscussed: (theme, featureFlags) => {
-    if (!canAddCountries(theme, featureFlags)) {
+  countriesDiscussed: (theme, interaction, featureFlags) => {
+    if (!canAddCountries(theme, interaction, featureFlags)) {
```
ian-leggett: Optional: could this be a ternary 
```
return !canAddCountries(theme, interaction, featureFlags) ? [] : [
      {
        type: 'radio',
        name: 'were_countries_discussed',
        modifier: 'inline',
        optional: false,
        options: [
          {
            label: 'Yes',
            value: 'true',
          },
          {
            label: 'No',
            value: 'false',
          },
        ],
      },
      ...[
        EXPORT_INTEREST_STATUS.FUTURE_INTEREST,
        EXPORT_INTEREST_STATUS.EXPORTING_TO,
        EXPORT_INTEREST_STATUS.NOT_INTERESTED,
      ].map((name) => ({
        macroName: 'Typeahead',
        name,
        hint: 'Add all that you discussed',
        modifier: 'subfield',
        condition: {
          name: 'were_countries_discussed',
          value: 'true',
        },
        isAsync: false,
        isLabelHidden: false,
        useSubLabel: false,
        placeholder: 'Search countries',
        classes: 'c-form-group--no-filter',
        multipleSelect: true,
        options: globalFields.countries.options(),
        target: 'metadata',
        autoSubmit: false,
      })),
    ]
```

src/apps/interactions/controllers/edit.js
web-bert: As the value is only used once, I don't see any benefit in it?

src/apps/interactions/macros/can-add-countries.js
web-bert: Updated 👍

src/apps/interactions/macros/fields.js
web-bert: Updated 👍

web-bert:
@ian-leggett addressed your comments

src/apps/interactions/constants.js
```diff
@@ -118,6 +118,12 @@ const KINDS = {
   SERVICE_DELIVERY: 'service-delivery',
 }
 
+const EXPORT_INTEREST_STATUS = {
```
rafenden: Optional: Consider using a plural form as the rest the constants in this file.

src/apps/interactions/macros/get-export-countries.js
```diff
@@ -0,0 +1,17 @@
+const castCompactArray = require('../../../lib/cast-compact-array')
+
+const { EXPORT_INTEREST_STATUS_VALUES } = require('../constants')
+
+module.exports = (body) => {
```
rafenden: Optional: We should decide where we define function exports and make it a convention, currently we are using the following forms:

At the end of the file:
```
function foo() {
}

module.exports = foo
```

Inline with the function:
```
module.exports = () => {
}
```

I would suggest using the first option and stick to it in the entire project as it's easier to find what it's public and what isn't when a file has many lines.

src/apps/interactions/constants.js
web-bert: There is a mix in this file and so I copied `INTERACTION_STATUS`

src/apps/interactions/macros/get-export-countries.js
web-bert: I like the first option when there is going to be more than one export (so each one is named), but the second when there is only one, as there is no indirection

PR #2312, opened 2019-12-06
Add Lead adviser filter

ian-leggett:
## Description of change

As we can now allocate a Lead ITA to a company we need a filter that will filter companies by Lead ITA or Global Account manager.

## Test instructions

Go to the companies page, find the filter on the bottom left hand side of the screen and search for a Lead ITA/Global Account manager and you should see the results of all companies that adviser has been allocated too. 
 
## Screenshots
Screenshot 2019-12-06 at 16 18 09

## Checklist


test/functional/cypress/specs/companies/filter-spec.js
```diff
@@ -181,4 +181,23 @@ describe('Company Collections Filter', () => {
...
+
+  it('should filter by Lead ITA and Global Account Manager', () => {
+    const leadIta = selectors.filter.leadIta
+    const { typeahead } = selectors.filter
+    cy
+      .get(typeahead(leadIta).selectedOption)
+      .click()
+      .get(typeahead(leadIta).textInput)
+      .type('Shawn Cohen')
+      .wait(2000)
```
rafenden: To make this test deterministic, we should define an alias for the typeahead endpoint and use it here instead of time, e.g.
```
cy.route('/url-to-ita-search?*').as('findITA')

cy
...
.wait('@findITA')
...
```

Disclaimer: Haven't tried this myself so not sure it this will work.

test/functional/cypress/specs/companies/filter-spec.js
```diff
@@ -181,4 +181,23 @@ describe('Company Collections Filter', () => {
...
+      .type('{enter}')
+      .type('{esc}')
+
+    cy.wait('@filterResults').then(xhr => {
```
rafenden: Optional: Consider asserting on the elements actually displayed to the user, not the HTTP response.

test/functional/cypress/specs/companies/filter-spec.js
rafenden: One more option would be to add assertion for the number of items returned to the typeahead, see the Cypress docs:
https://docs.cypress.io/guides/references/best-practices.html#Unnecessary-wait-for-cy-get

ian-leggett: Thanks much better, removed the `wait` and split into two assertions. 👍

PR #2313, opened 2019-12-09
Corrections and enhancements to the Healthcheck endpoints

mcbhenwood:
## Description of change

This PR addresses a few things:

1. Pingdom needs to be able to call _two_ healthcheck points per service. One for critical failures (P1); the other for less critical "warning only" level problems (P2)
2. The recurrent "getaddress postcode lookup" check should be relegated to a P2 check
3. The basic microservice healthcheck _should not_ check dependencies. It is there only to allow CF components to verify that _this_ service has mounted its router and is available to serve requests, regardless of the health of its dependencies.
4. Removal of `this` and the `beforeEach()` antipattern from unit tests

- [X] Has the branch been rebased to develop?

### DIFF
--- a/src/apps/healthcheck/__test__/controllers.test.js
+++ b/src/apps/healthcheck/__test__/controllers.test.js
@@ -58,12 +58,12 @@ const getMicroserviceHealthcheck = async (dependencies) => {
   return callNamedHandler('getMicroserviceHealthcheck', dependencies)
 }
 
-const renderPingdomXml = async (dependencies) => {
-  return callNamedHandler('renderPingdomXml', dependencies)
+const getPingdomFailures = async (dependencies) => {
+  return callNamedHandler('getPingdomFailures', dependencies)
 }
 
-const renderPingdomWarningXml = async (dependencies) => {
-  return callNamedHandler('renderPingdomWarningXml', dependencies)
+const getPingdomWarnings = async (dependencies) => {
+  return callNamedHandler('getPingdomWarnings', dependencies)
 }
 
 describe('Health check controller', () => {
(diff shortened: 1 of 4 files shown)


src/apps/healthcheck/__test__/controllers.test.js
```diff
@@ -1,231 +1,298 @@
 const proxyquire = require('proxyquire')
 
+const _successDependencies = [
```
web-bert: Optional: using _ notation for private is not something I have seen across the rest of the codebase

src/apps/healthcheck/controllers.js
```diff
@@ -50,30 +50,25 @@ function renderPingdomXml (req, res, next) {
...
-    .then((results) => results.filter((result) => result.statusText !== 'OK'))
-    .then((errors) => {
-      res.set('Cache-Control', 'no-cache, no-store, must-revalidate')
-
-      if (errors.length) {
-        errors.forEach((dependency) => {
-          logger.error(`${dependency.name} health check failed`, dependency.error)
-        })
+function getPingdomFailures (req, res, next) {
+  const failureDependencies = serviceDependencies.filter((dependency) => (!dependency.warningOnly))
```
web-bert: This should be done once at the top of the module, as this will be called once a second it's a waste of resources.

src/apps/healthcheck/controllers.js
```diff
@@ -50,30 +50,25 @@ function renderPingdomXml (req, res, next) {
...
+  return renderPingdomXml(req, res, next, failureDependencies)
+}
 
-        return res
-          .status(503)
-          .send('Service Unavailable')
-      }
+function getPingdomWarnings (req, res, next) {
+  const warningDependencies = serviceDependencies.filter((dependency) => (dependency.warningOnly))
```
web-bert: This should be done once at the top of the module, as this will be called once a second it's a waste of resources.

PR #2316, opened 2019-12-09
Simplify the release process by removing the "develop" branch

rafenden:
## Description of change

* Branch `develop` will no longer be used to simplify the workflow.
* End to end tests will be running now against the `master` branch of the API instead of `develop`.
* No version number in `package.json` file, env var `GIT_BRANCH` will be used instead, which will contain the Git tag name. See https://github.com/uktrade/ci-pipeline/blob/f10d914745083e738a6384af858089dc644840eb/Jenkinsfile#L299


docs/Release process.md
```diff
@@ -0,0 +1,15 @@
+# Release process
```
peterhudec: Seems like you accidentally renamed the `release-process.md`.

rafenden: It was renamed on purpose 🙂

peterhudec: I see, but how about the step 14? Is the new step 6 sufficient replacement for it?

docs/Release process.md
```diff
@@ -0,0 +1,15 @@
...
+
+1. Create a GIT tag `v5.X.X` pointing to the latest `master` (X refers to the next version number) and push it to the remote.
```
elcct: If you are using semantic versioning it could be helpful to reflect that here, so instead of `v5.X.X` have `v<major>.<minor>.<patch>' or if that is not the scheme, explain a bit how the next version number should be worked out. Also another thing is that `v5.X.X` may suggest the version number should be the same in both places e.g. `v5.2.2` or `v5.15.15`, so this could be changed to `v5.X.Y`.

docs/Release process.md
rafenden: Well, we don't use the GH releases for anything so what's the point of creating them?

.circleci/config.yml
```diff
@@ -7,7 +7,6 @@ aliases:
   - &oauth2_dit_staff_token     ditStaffToken
   - &oauth2_da_staff_token      daStaffToken
   - &oauth2_lep_staff_token     lepStaffToken
-  - &api_develop_image          quay.io/uktrade/data-hub-api:develop
   - &api_master_image           quay.io/uktrade/data-hub-api:master
```
PippoRaimondi: The backend still has 2 branches right? master and develop? in that case we want to use the develop api image which is continuously pushed when code is merged to dev... just in case there is a backend dependency that we need in the FE and we wouldn't have to wait for them to release or merge it to master...

rafenden: I think we should test against a real production environment, otherwise there is a risk to deploy code which will not work.

docs/Release process.md
rafenden: Very good point, description was added.

.circleci/config.yml
PippoRaimondi: you are right, thinkin about CD we should test against prod backend!

docs/Deployments.md
```diff
@@ -0,0 +1,29 @@
...
+
+1. Create a GIT tag `v<MAJOR>.<MINOR>.<PATCH>`, e.g. `v5.1.2` pointing to the latest `master` and push it to the remote.
+   
+    | Release type | When to increase |
+    | ------------- | ------------- |
+    | Major (**1**.0.0) | When a change requires modifications to the infrastructure, e.g. NodeJS version upgrade. |
+    | Minor (0.**1**.0) | When a release contains at least one new feature. |
+    | Patch (0.0.**1**) | When a release contains only fixes. | 
+
+2. Go to Jenkins and open the job used for FE deployments.
```
peterhudec: It would be handy to have Jenkins as a link here.

docs/Deployments.md
```diff
@@ -0,0 +1,29 @@
...
+
+Deployments to production are done manually through Jenkins where a Git tag can be used.
+
+## Deploying to production
+
+1. Create a GIT tag `v<MAJOR>.<MINOR>.<PATCH>`, e.g. `v5.1.2` pointing to the latest `master` and push it to the remote.
+   
+    | Release type | When to increase |
+    | ------------- | ------------- |
+    | Major (**1**.0.0) | When a change requires modifications to the infrastructure, e.g. NodeJS version upgrade. |
```
peterhudec: Maybe also mention breaking changes here?

docs/Deployments.md
rafenden: For security reasons we shouldn't include links to our internal infra.

docs/Deployments.md
rafenden: Can you clarify? I'm hesitant to use `breaking` as this library is not a dependency to anything else.

rafenden:
> Only suggested small improvements to the Deployments.md, but seems that a lot of stuff leaked into this PR.
@peterhudec it's fixed now

docs/Deployments.md
peterhudec: Yeah, I have just realized that.

PR #2321, opened 2019-12-11
Migrate investment features to cypress

PippoRaimondi:
## Description of change

This PR will be responsible to migrate all investment project features to cypress

## Test instructions

yarn test:e2e:dit
 

src/lib/urls.js
```diff
@@ -101,6 +101,7 @@ module.exports = {
     index: url('/companies'),
     interactions: createInteractionsSubApp('/companies', '/:companyId'),
     orders: url('/companies', '/:companyId/orders'),
+    investment: url('/companies', '/:companyId/investments'),
```
mcbhenwood: Fiddly though it is, I do think that this property should have a plural name to match the names ("interactions", "orders", etc) of other properties.

That may mean that you call it `company_investments` or (preferably) that you assign a new name to the existing `investments` property which follows.

test/end-to-end/cypress/specs/DIT/collection-spec.js
```diff
@@ -25,4 +27,24 @@ describe('Collection', () => {
       checkCollection()
     })
   })
+
+  describe('investment interaction', () => {
+    before(() => {
+      cy.visit(investments.projects.interactionCollection('721e2a04-21c3-4172-a321-4368463a4b2d'))
```
mcbhenwood: Is it worth us collecting these fixture UUIDs in a central module somewhere then referring to them like:
```
const { interactionCollectionUUID } = require ('superTestFixtureUUIDs')
...
  cy.visit(investments.projects.interactionCollection(interactionCollectuionUUID))
```

(optional)

PippoRaimondi: yes, this is coming as part of the refactor I want to do... I will move the fixtures folder up a level and update functionals and e2e to use these same fixtures (want to do this in a single PR)...

src/lib/urls.js
PippoRaimondi: agreed, will sort this out!

PR #2327, opened 2019-12-16
Companies list - edit list name

ian-leggett:
## Description of change
This enables the user to edit a company list.

## Test instructions
As there is no button to edit a list (this is a separate PR) you will need to select a list from the home page, click the "delete this list" like then change the `/delete` route in the browser to `/rename`
 
## Screenshots
Screenshot 2019-12-16 at 09 36 25

## Checklist

[//]: # "When submitting a PR make sure the code review guidelines have been satisfied. 
https://github.com/uktrade/data-hub-frontend/blob/develop/docs/Code%20review%20guidelines.md"

- [x] Has the branch been rebased to develop?
- [x] Automated tests (Any of the following when applicable: Unit, Functional or Acceptance)
- [x] Manual compatibility testing (Browsers: Chrome, Firefox, IE11, Safari)

PR #2329, opened 2019-12-16
Updated layout of my companies dashboard section and added the rename button

peterhudec:
## Description of change

This PR adds the _Edit list name_ button to, and updates the layout of the _my companies_ dashboard section. The layout update was a good opportunity move it here from `data-hub-components` and refactor it.
- Fixed a bug when text from the search disappeared after switching lists, but the filter stayed applied and couldn't be removed.
- All the styling is now done with [`typestyle`](https://typestyle.github.io/)
- Context is gone. The component now only uses the `useReducer` hook, which makes it Redux-ready.

https://trello.com/c/wobMbHdS/1035-change-homepage-table-layout

## Test instructions
On the homepage `/` you should see a new layout resembling the [proposed design](https://departmentforinternationaltrade.invisionapp.com/share/3GUWQ9ZPVRF#/screens/394146818), with the _Edit list name_ button alongside the _Delete list_ button. The button should link to `/company-lists/:id/rename`.
 
### Before
Screen Shot 2019-12-16 at 11 27 06 AM

### After
Screen Shot 2019-12-16 at 11 27 19 AM

## Checklist


package.json
```diff
@@ -138,6 +138,7 @@
     "style-loader": "^0.21.0",
     "styled-components": "^4.3.2",
     "tar": "^4.4.8",
+    "typestyle": "^2.0.4",
```
rafenden: Can you please explain why you didn't use `styled-components`? What problems does this lib solve?

test/functional/cypress/specs/company-lists/view-spec.js
```diff
@@ -44,6 +44,7 @@ const expectedRows = {
 const expectedLists = {
   'List B': {
     deleteLinksTo: '/company-lists/75e14e32-292e-4d1b-a361-992d548251f7/delete',
```
ian-leggett: I have started to add these list ID's into a fixtures folder. Have a look at https://github.com/uktrade/data-hub-frontend/pull/2327 this will make it more DRY. Happy for comments on my PR as I have only just started to add fixtures for the lists.

package.json
peterhudec: I needed a _secondary_ button, which `govuk-react` doesn't provide out of the box, so I had to tweak the button style manually. But I needed the button to work as a link: `<Button as={Link}/>` or `<Link as={button}/>` . Unfortunately, styling the buttons/links with `styled(Button)` or `styled(Link)` breaks their styles. Styling it with `typestyle` works as a charm.

Here is a CodeSandbox demonstrating the problem:
https://codesandbox.io/s/govuk-react-styled-components-vs-typestyle-h5e77

Most importantly though, `styled-components` don't allow any other way of applying styles than by defining new components, which forces you to invent meaningless names like `StyledDiv`, `StyledUnorderedList`, `StyledTableRow`, `StyledCheckbox` etc. It's a shame, because under the hood, it must have a mechanism to take a style definition, convert it to a _namespaced_ string and assign it to the `className` prop, but it doesn't expose it. I was looking for it in their source, but the whole mechanism is so tightly coupled to the concept of components that it's probably impossible for them to isolate it.

[`typestyle`](https://github.com/typestyle/typestyle) is one of the simplest JSS libraries out there. It does only one thing and it does it right and leaves it up to you to invent ingenious high level APIs.

I know we can't migrate completely from `styled-components` because of our dependency on `govuk-react`, but then again, if you look at my [button example](https://codesandbox.io/s/govuk-react-styled-components-vs-typestyle-h5e77), the fragile `styled-components` mechanism just doesn't work.

src/apps/dashboard/client/CompanyLists/index.jsx
```diff
@@ -0,0 +1,287 @@
+import { GREY_2, GREY_3 } from 'govuk-colours'
```
ian-leggett: This file is quite massive, I think we need to chop this up at some time.

src/apps/dashboard/client/CompanyLists/SecondaryButton.jsx
```diff
@@ -0,0 +1,9 @@
+import { TEXT_COLOUR, GREY_3 } from 'govuk-colours'
+import Button from '@govuk-react/button'
+import React from 'react'
+
+// TODO: Move to data-hub-components
+const SecondaryButton = props =>
```
ian-leggett: Can't we get this from Gov React? `import Button from '@govuk-react/button'`

ian-leggett: Optional: I would remove this and just use `import Button from '@govuk-react/button'` in the index.jsx file

peterhudec: The `@govuk-react/button` is _primary_ (green) by default, but we are using the secondary variant quite a lot.

ian-leggett: Fair play 👍

peterhudec: If you look at the line #7 below, it's exactly what I'm doing. I abstracted the button away to a `SecondaryButton` because just in this feature we are using it twice. This is exactly the type of component that should go to `data-hub-components`. But ideally a `secondary` flag would be added to `@govuk-react/button`.

peterhudec: If you look at the line #7 below, it's exactly what I'm doing. I abstracted the button away to a `SecondaryButton` because just in this feature we are using it twice. This is exactly the type of component that should go to `data-hub-components`. But ideally a `secondary` flag would be added to `@govuk-react/button`.

package.json
rafenden: This can be easily achieved using styled components:
https://codesandbox.io/s/govuk-react-styled-components-vs-typestyle-0gb3x

Please do not introduce alternative styling library without discussing this first with the rest.

PR #2337, opened 2019-12-18
Update docs to include step for creating a GH release

rafenden:
## Description of change

Looks like some people in the organization relies on the GH releases (solely for the notes) so let's keep creating them until semantic release is implemented which will automate the process.

### DIFF
--- a/docs/Deployments.md
+++ b/docs/Deployments.md
@@ -23,7 +23,10 @@ Deployments to production are done manually through Jenkins where a Git tag can 
 4. Type your version number created in step `1` to the `Git_Commit` text field.
 
 5. Press the `Build` button.
-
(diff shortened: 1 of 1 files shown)


docs/Deployments.md
```diff
@@ -23,7 +23,10 @@ Deployments to production are done manually through Jenkins where a Git tag can
 4. Type your version number created in step `1` to the `Git_Commit` text field.
 
 5. Press the `Build` button.
-
-6. Release notes can be found using the tag-based comparison provided by GitHub, `github.com/uktrade/data-hub-frontend/compare/OLD_TAG…NEW_TAG`
    
+6. [Create a GH release](https://github.com/uktrade/data-hub-frontend/releases).
```
mcbhenwood: Somewhat pickily, I might suggest that the hyperlink is only around the words "GH release". Otherwise the link suggests that clicking on it _will create a GH release_ when in fact it's just a link to the GH releases themselves

rafenden: Good point.

PR #2345, opened 2020-01-07
Redux

peterhudec:
## Description of change

* Added Redux state management to the codebase
* Migrated the `CompanyLists` component on the _dashboard_ to Redux
* `src/client.jsx` moved to `src/client/index.jsx` to make place for `src/client/actions.js`

The preloaded data is still passed around through the `data-props` attribute. This will be addressed in another pull request.

## Test instructions

Everything should work exactly as before.

## Checklist


package.json
```diff
@@ -125,8 +125,11 @@
     "react": "^16.8.6",
     "react-app-polyfill": "^1.0.1",
     "react-dom": "^16.8.6",
+    "react-redux": "^7.1.3",
     "react-use": "^13.2.1",
     "redis": "^2.8.0",
+    "redux": "^4.0.5",
+    "redux-devtools-extension": "^2.13.8",
```
mcbhenwood: Ought this go in `devDependencies` instead?

peterhudec: Good point. It depends whether we want to hide it in production, or show it off as AirBnB does it. It's up for discussion.

mcbhenwood: I think from a DIT security perspective the steer is that if it's not _needed_ in prod, then we should leave it out of prod.

src/apps/dashboard/client/CompanyLists/index.jsx
```diff
@@ -5,11 +5,16 @@ import {
...
-import React, { useReducer } from 'react'
+import PropTypes from 'prop-types'
+import React from 'react'
+import { connect } from 'react-redux'
 import styled from 'styled-components'
 
-import * as actions from './actions'
-import reducer, { initialState } from './reducer'
+import {
+  COMPANY_LIST_VIEWER__FILTER,
```
rafenden: It's worth documenting this convention.

src/apps/dashboard/client/CompanyLists/index.jsx
```diff
@@ -100,7 +89,29 @@ function CompanyLists (props) {
 }
 
 CompanyLists.propTypes = {
-  lists: propTypes.lists,
+  lists: propTypes.lists.isRequired,
+  selectedIdx: PropTypes.oneOfType([PropTypes.number, PropTypes.string]).isRequired,
+  sortBy: PropTypes.oneOf(['alphabetical', 'recent', 'least-recent']).isRequired,
```
rafenden: We should use constants here.

src/apps/dashboard/client/CompanyLists/index.jsx
```diff
@@ -100,7 +89,29 @@ function CompanyLists (props) {
...
+  filter: PropTypes.string.isRequired,
+  onChange: PropTypes.func,
```
rafenden: I suggest adding a more descriptive name. Same applies for `Header`, it's hard to remember what exactly is changing here, a better name could be `onListChange`.

src/apps/dashboard/client/CompanyLists/index.jsx
```diff
@@ -100,7 +89,29 @@ function CompanyLists (props) {
...
+  onOrderChange: PropTypes.func,
+  onSearch: PropTypes.func,
 }
 
-export default CompanyLists
+export default connect(
+  state => state.companyLists,
+  dispatch => ({
+    onChange: idx => dispatch({
```
rafenden: As above, I suggest naming it `onListChange`.

src/apps/dashboard/client/CompanyLists/index.jsx
```diff
@@ -100,7 +89,29 @@ function CompanyLists (props) {
...
+      type: COMPANY_LIST_VIEWER__CHANGE,
```
rafenden: As above, I suggest naming it`COMPANY_LIST_VIEWER__LIST_CHANGE`.

package.json
rafenden: Worth noting: Our CI uses `dependencies` for all environments so these tools won't be available on any of our remote servers (including the review apps).

PR #2347, opened 2020-01-08
Show non field errors

web-bert:
## Description of change

Add ability to show non field errors and do so when adding an interaction. Due to the new feature of adding countries discussed with an interaction there is a validation case where a non field error can be returned, this passes that error to the controller which then gets passed to the error-summary component/macro

## Test instructions

**Using the API branch** `feature/interaction_journey_add_countries_validation`

Add an interaction and answer "Were countries discussed" as "Yes", then put the same country in each field and it will return a `non_field_error` in the API which will get displayed in the UI
 
### Before

No error shown

### After 

Screenshot 2020-01-08 at 07 21 25

## Checklist

### DIFF
--- a/test/functional/cypress/specs/interaction/add-interaction-spec.js
+++ b/test/functional/cypress/specs/interaction/add-interaction-spec.js
@@ -30,6 +30,11 @@ describe('Add Export', () => {
 
             cy.get(selectors.interactionForm.add).click()
 
+            cy.get(formSelectors.form).contains(
+              'The export_countries field may not be null.'
+            )
+            cy.get(formSelectors.form).contains('A non field error message')
+
             cy.get(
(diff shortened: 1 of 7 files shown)


test/functional/cypress/specs/interaction/add-interaction-spec.js
```diff
@@ -19,6 +19,9 @@ describe('Add Export', () => {
 
             cy.get(selectors.interactionForm.add).click()
 
+            cy.get(formSelectors.form).contains('The export_countries field may not be null.')
```
rafenden: Could we make this error more user-friendly by using a label and not the field name from the DB?

web-bert: This is only the mock error message, the real error is user friendly, I wanted it unique for this assertion

PR #2354, opened 2020-01-09
Update eslint config to include JSX files

rafenden:
## Description of change

JSX files are currently not included when linting code, this PR is fixing that problem.


paulgain:
Now works for me. Thanks.

package.json
```diff
@@ -15,10 +15,9 @@
     "watch:test": "npm run test:unit -- -w",
     "watch:js:client": "webpack --watch --progress",
     "watch:js:server": "nodemon $NODE_DEBUG_OPTION --inspect --ignore 'src/**/__test__/**/*'",
-    "lint": "npm-run-all lint:all:*",
-    "lint:fix": "npm run lint:all:js --fix",
-    "lint:all:js": "eslint ./test ./src ./assets ./.circleci",
-    "lint:all:styles": "sass-lint -v -q -c .sass-lint.yml assets/stylesheets/**/*.scss",
+    "lint": "npm run lint:js && npm run lint:sass",
```
web-bert: This can be `run-p lint:*` to run them in parallel and will automatically run any additional lint tasks that are added in the future

web-bert: @rafenden ☝️

rafenden: Good point, the code was updated!

package.json
```diff
@@ -11,14 +11,13 @@
     "start": "node --use-strict src/server.js",
     "build": "npm run clean && webpack",
     "build:prod": "NODE_ENV=production npm run build",
-    "develop": "run-p watch:js:*",
+    "develop": "npm-run-all watch:js:*",
```
web-bert: Do we want to change this to sequential? use `--parallel` or `run-p`

package.json
```diff
@@ -11,14 +11,13 @@
...
     "watch:test": "npm run test:unit -- -w",
     "watch:js:client": "webpack --watch --progress",
     "watch:js:server": "nodemon $NODE_DEBUG_OPTION --inspect --ignore 'src/**/__test__/**/*'",
-    "lint": "npm-run-all lint:all:*",
-    "lint:fix": "npm run lint:all:js --fix",
-    "lint:all:js": "eslint ./test ./src ./assets ./.circleci",
-    "lint:all:styles": "sass-lint -v -q -c .sass-lint.yml assets/stylesheets/**/*.scss",
+    "lint": "npm-run-all lint:*",
```
web-bert: Do we want this to be sequential? use `--parallel` or `run-p`

rafenden: Good point, I check the `npm-run-all` package and it indeed defines `run-p` binary: https://github.com/mysticatea/npm-run-all/tree/master/bin/run-p

The code was updated to use `run-s`.

package.json
rafenden: This change was reverted, thanks.

PR #2365, opened 2020-01-15
Add find a matching D&B company record

paulgain:
## Find a matching D&B company record

Allow users to match a Data Hub record with a Dunn & Bradstreet record by performing a search. This PR only deals with the searching of companies. The next PR includes company selection from the list of results. 

## Test instructions
_/companies/id/match_
Define a search term and click **_Find company_**

### Before
find-a-company

### After -  still hidden from the user
match-company-record

## Checklist

### DIFF
--- a/src/apps/companies/apps/match-company/client/FindCompany.jsx
+++ b/src/apps/companies/apps/match-company/client/FindCompany.jsx
@@ -1,8 +1,32 @@
 import React from 'react'
+import PropTypes from 'prop-types'
 import { H3 } from '@govuk-react/heading'
+import { companies } from '../../../../../lib/urls'
 
-function FindCompany() {
-  return <H3>Hidden from the user</H3>
+import { Form, FieldDnbCompany } from 'data-hub-components'
+
+function FindCompany({ csrfToken, company }) {
+  function onSubmit(values) {
+    console.log(values) // Intentional
+  }
+
+  return (
+    <Form onSubmit={onSubmit}>
+      <H3>Find the company record in the Dun & Bradstreet database</H3>
+      <FieldDnbCompany
+        apiEndpoint={`${companies.match.index(company.id)}?_csrf=${csrfToken}`}
+        queryParams={{ address_country: 'GB' }}
+        name="dnbCompany"
+      />
+    </Form>
+  )
+}
+
+FindCompany.propTypes = {
+  csrfToken: PropTypes.string.isRequired,
+  company: PropTypes.shape({
+    id: PropTypes.string.isRequired,
+  }),
 }
 
 export default FindCompany
--- a/src/apps/companies/apps/match-company/controllers.js
+++ b/src/apps/companies/apps/match-company/controllers.js
@@ -9,21 +9,6 @@ const { searchDnbCompanies } = require('../../../../modules/search/services')
 const ZENDESK_TICKET_TAG_MATCH_REQUEST = 'dnb_match_request'
 const ZENDESK_TICKET_TYPE_TASK = 'task'
 
-async function renderFindCompanyForm(req, res, next) {
-  try {
-    const { company } = res.locals
-
-    res
-      .breadcrumb(company.name, urls.companies.detail(company.id))
-      .breadcrumb('Find this company record')
-      .render('companies/apps/match-company/views/find-company', {
-        props: {},
-      })
-  } catch (error) {
-    next(error)
-  }
-}
-
 function parseCountry(country, countries) {
   return get(
     typeof country === 'string'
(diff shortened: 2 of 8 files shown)


src/apps/companies/apps/match-company/client/FindCompany.jsx
```diff
@@ -1,8 +1,30 @@
...
+import { Form, Step, FieldDnbCompany } from 'data-hub-components'
+
+function FindCompany({ csrfToken, companyId }) {
+  function onSubmit(values) {
+    console.log(values) // Intentional
+  }
+
+  return (
+    <Form onSubmit={onSubmit}>
+      <div>
```
rafenden: Probably, there is no need for this DIV.

src/apps/companies/apps/match-company/client/FindCompany.jsx
```diff
@@ -1,8 +1,30 @@
...
+        <H3>Find the company record in the Dun & Bradstreet database</H3>
+        <FieldDnbCompany
+          apiEndpoint={`/companies/${companyId}/match?_csrf=${csrfToken}`}
```
rafenden: We should use the urls module.

src/apps/companies/apps/match-company/client/FindCompany.jsx
```diff
@@ -1,8 +1,30 @@
...
+          queryParams={{ address_country: 'GB' }}
+          name="dnbCompany"
+        />
+      </div>
+    </Form>
+  )
+}
+
+FindCompany.propTypes = {
```
rafenden: `companyId` is missing here.

src/apps/companies/apps/match-company/controllers.js
```diff
@@ -8,13 +9,29 @@ async function renderFindCompanyForm(req, res, next) {
       .breadcrumb(company.name, companies.detail(company.id))
       .breadcrumb('Find this company record')
       .render('companies/apps/match-company/views/find-company', {
-        props: {},
+        props: {
+          companyId: company.id,
```
rafenden: You might want to pass the company object here which will be useful in next PR when you will display the existing company details above the search form.

test/functional/cypress/specs/companies/match-company-spec.js
```diff
@@ -1,11 +1,16 @@
...
 const {
   assertLocalHeader,
   assertBreadcrumbs,
 } = require('../../support/assertions')
-const fixtures = require('../../fixtures')
-const urls = require('../../../../../src/lib/urls')
 
 describe('Match a company', () => {
+  beforeEach(function() {
+    Cypress.Cookies.preserveOnce('datahub.sid')
```
rafenden: Could you please explain why this was required?

src/apps/companies/apps/match-company/client/FindCompany.jsx
paulgain: 👍

src/apps/companies/apps/match-company/client/FindCompany.jsx
paulgain: 👍

src/apps/companies/apps/match-company/controllers.js
paulgain: It's a good point. However, do we really need to pass the entire company from the server to the browser when all we need is the company name, id and address? In addition, defining the `propTypes` for a company would be ridiculous although we could be vague, for example: `company: PropTypes.object.isRequired,`. I think the best approach would be to create a new object and pass just what we need, the propTypes would be more concise and readable. What do you think?

test/functional/cypress/specs/companies/match-company-spec.js
paulgain: Strictly speaking it's not required if we add one of these to each `context`:

```js
before(() => {
  cy.visit(urls.companies.match.index(fixtures.company.venusLtd.id))
})
```

src/apps/companies/apps/match-company/client/FindCompany.jsx
paulgain: Take a look at the discussion further down. Thanks.

src/apps/companies/apps/match-company/controllers.js
rafenden: That's correct, you don't need to pass the entire object, this could be enough (for now):
```
company: {
  ...pick(company, [
      'id',
    ])
}
```

test/functional/cypress/specs/companies/match-company-spec.js
rafenden: Shall we remove it as it's not required then? I assume each context opens a new URL.

paulgain: Happy to remove it.

src/apps/companies/apps/match-company/controllers.js
paulgain: 👍

test/functional/cypress/specs/companies/match-company-spec.js
```diff
@@ -40,19 +104,19 @@ describe('Match a company', () => {
       cy.contains('Send update request').click()
     })
 
-    it('should redirect to the company page', () => {
-      cy.location('pathname').should(
-        'eq',
-        urls.companies.activity.index(fixtures.company.venusLtd.id)
-      )
-    })
+    // it('should redirect to the company page', () => {
```
ian-leggett: Can we remove this?

paulgain: After a rebase on top of master these tests failed, they are Raf's tests. I've spoken to Raf about them, I'm missing an environment variable which now fixes both tests. They have now been uncommented - we're now good to go.

PR #2368, opened 2020-01-16
Rename the Audit trail HTML link to Edit history

paulgain:
## Update HTML link name & URL
This pull request simply renames the **Audit trail page** link to **Edit history page** and updates the URL exposing the new implementation to users.

<dl>
  <dt>URL from:</dt>
  <dd>/companies/id/audit</dd>

  <dt>URL to:</dt>
  <dd>/companies/id/edit-history</dd>
</dl>

## Test instructions
Click the link to view the new Edit history page.

### Before

audit-history-link

### After
edit-history-link

## Checklist

### DIFF
--- a/src/apps/companies/apps/business-details/__test__/controllers.test.js
+++ b/src/apps/companies/apps/business-details/__test__/controllers.test.js
@@ -87,7 +87,9 @@ describe('Company business details', () => {
                   companyMock.id
                 )}?_csrf=csrf`,
                 companyAdvisers: urls.companies.advisers.index(companyMock.id),
-                companyAudit: urls.companies.audit(companyMock.id),
+                companyEditHistory: urls.companies.editHistory.index(
+                  companyMock.id
+                ),
                 support: urls.support(),
                 subsidiaries: urls.companies.subsidiaries.index(companyMock.id),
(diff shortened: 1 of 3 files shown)


web-bert:
@paulgain does the url need to be changed? If someone has it bookmarked it will break, so if it does need to be changed a redirect from the old url to the new one should be added?

src/apps/companies/apps/business-details/__test__/controllers.test.js
```diff
@@ -87,7 +87,9 @@ describe('Company business details', () => {
                   companyMock.id
                 )}?_csrf=csrf`,
                 companyAdvisers: urls.companies.advisers.index(companyMock.id),
-                companyAudit: urls.companies.audit(companyMock.id),
```
rafenden: I assume we will remove the old audit page and url in the next PR?

paulgain:
> @paulgain does the url need to be changed? If someone has it bookmarked it will break, so if it does need to be changed a redirect from the old url to the new one should be added?

We should be free to update our URLs as we see fit. If it breaks a bookmark it shouldn't matter, the user can simply update it. I don't think it's a good idea to pepper our code base with redirects and the corresponding tests which would occur over time.

PR #2370, opened 2020-01-16
Refactor transformCompanyToDnbHierarchyList

debitan:
## Description of change

Refactor transformer for dnb hierarchy list to provide badge props as an object with the key of `text`. The allows future passing of a `borderColour` key. Without this change the list would not render because of breaking changes in Data-Hub-Components release 3.5.0

## Test instructions

The dnb hierarchy page should render the same. 

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -85,7 +85,7 @@
     "core-js": "^2.5.7",
     "css-loader": "^1.0.0",
     "csurf": "^1.9.0",
-    "data-hub-components": "^3.2.0",
+    "data-hub-components": "^3.5.0",
     "date-fns": "^1.29.0",
     "del-cli": "^2.0.0",
     "dotenv": "^5.0.0",
--- a/src/apps/companies/apps/dnb-hierarchy/__test__/controllers.test.js
+++ b/src/apps/companies/apps/dnb-hierarchy/__test__/controllers.test.js
@@ -157,7 +157,7 @@ describe('D&B Company hierarchy', () => {
           count: 1,
           results: [
             {
-              badges: ['United Kingdom', 'North West'],
+              badges: [{ text: 'United Kingdom' }, { text: 'North West' }],
               headingText: 'Mars Components Ltd',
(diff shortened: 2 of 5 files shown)


debitan:
> I realise this was here beforehand, however, there's no value is calling `get` here after performing the checks:
> 
> ```js
>   if (address && address.country) {
>     badges.push({
>       text: get(address, 'country.name'),
>  })
> ```
> 
> You called do this instead:
> 
> ```js
>   if (address && address.country) {
>     badges.push({
>       text: address.country.name,
>  })
> ```
> 
> Or the preferred approach (current code base):
> 
> ```js
> const countryName = get(address, 'country.name')
> if (countryName) {
>   badges.push({ text: countryName })
> }
> ```

@paulgain oh if I can that is great. Tbh, I don't like using lodash as some of their functions are easier to understand in plain JS. Was just keeping it consistent with the rest of the file, but I'll go and change that now.

paulgain:
> > I realise this was here beforehand, however, there's no value is calling `get` here after performing the checks:
> > ```js
> >   if (address && address.country) {
> >     badges.push({
> >       text: get(address, 'country.name'),
> >  })
> > ```
> > 
> > 
> > You called do this instead:
> > ```js
> >   if (address && address.country) {
> >     badges.push({
> >       text: address.country.name,
> >  })
> > ```
> > 
> > 
> > Or the preferred approach (current code base):
> > ```js
> > const countryName = get(address, 'country.name')
> > if (countryName) {
> >   badges.push({ text: countryName })
> > }
> > ```
> 
> @paulgain oh if I can that is great. Tbh, I don't like using lodash as some of their functions are easier to understand in plain JS. Was just keeping it consistent with the rest of the file, but I'll go and change that now, using the second option if that's ok so I can remove lodash from it entirely.

@debitan The benefit of the last approach is that you will never create a badge when the country name is `undefined` or `null`.

debitan:
discussed further with @paulgain and I have made the code more defensive to account for any rouge undefined fields in the data received from the api. combines the benefit of removing lodash from option 2 with the security of option 3.

PR #2373, opened 2020-01-16
Exp/api endpoint

peterhudec:
This is a resurection of PR #2242, which has been closed, probably when we moved from `develop` to `master` and which can't be reopened because the target branch `develop` doesn't exist anymore.

## Description of change

Added the `/api-proxy` endpoint, which proxies all requests to **whitelisted** routes to `config.apiRoot` with added `authorization: Bearer <session.token>` header. The `/api-proxy` prefix is stripped from the requests, e.g a request to `localhost:3000/api-proxy/v4/company-list` will be proxied to `<apiRoot>/v4/company-list`.

## Test instructions

* You should be able to access the API from `/api-proxy` without the need to use the access token.
* You should not be able to access the API from `/api-proxy` if you are not logged in.
* You should only be able to access whitelisted routes, which at the moment are `/v4/company-list` and `/v4/company-list/:id/item`

### DIFF
--- a/src/server.js
+++ b/src/server.js
@@ -120,10 +120,10 @@ app.use(store())
 app.use(csrf())
 app.use(csrfToken())
 
-app.use(proxy('/api', {
+app.use(proxy('/api-proxy', {
   target: config.apiRoot,
   pathRewrite: {
-    '^/api': '/',
+    '^/api-proxy': '/',
   },
   onProxyReq: (req, res) =>
     req.setHeader('authorization', `Bearer ${res.session.token}`),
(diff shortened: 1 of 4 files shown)


src/server.js
```diff
@@ -12,6 +12,7 @@ const enforce = require('express-sslify')
 const favicon = require('serve-favicon')
 const cookieParser = require('cookie-parser')
 const minifyHTML = require('express-minify-html')
+var proxy = require('http-proxy-middleware')
```
rafenden: Should be `const`.

PR #2379, opened 2020-01-20
Task component (Redux-Saga)

peterhudec:
## Description of change

Added the redux-saga based `Task` component, an abstraction of an asynchronous task which can be in progress, possibly fail and be retried.

I couldn't write unit tests for this because it blows up on `import` statements.

## Test instructions

The component is not used anywhere yet, but the plan is to use it first in the _my companies list_ in the dashboard. To see it in action you can replace the contents of `src/client/index.jsx` with the following snippet, which is roughly the way the _my companies list_ will be done:

```jsx
// Paste me to src/client/index.jsx
import React from 'react'
import ReactDOM from 'react-dom'
import { Provider } from 'react-redux'
import { createStore, combineReducers, applyMiddleware } from 'redux'
import { composeWithDevTools } from 'redux-devtools-extension/developmentOnly'
import createSagaMiddleware from 'redux-saga'

import Task from './components/Task'
import tasksSagaFactory from './components/Task/saga'
import tasksReducer from './components/Task/reducer'
import { connect } from 'react-redux'

const companyListsReducer = (
  state = { listDetails: {} },
  { type, ...action }
) => {
  switch (type) {
    case 'COMPANY_LISTS__LISTS_LOADED':
      return {
        ...state,
        lists: action.result,
        selectedId: action.result[0].id,
      }
    case 'COMPANY_LISTS__COMPANIES_LOADED':
      return {
        ...state,
        listDetails: {
          ...state.listDetails,
          [action.payload]: action.result,
        },
      }
    case 'COMPANY_LIST__SELECT':
      return { ...state, selectedId: action.id }
    default:
      return state
  }
}

const dummyFetchCompanyistsTask = () =>
  new Promise((resolve) =>
    setTimeout(resolve, 1000, [
      {
        name: 'Foo',
        id: 'foo',
      },
      {
        name: 'Bar',
        id: 'bar',
      },
      {
        name: 'Baz',
        id: 'baz',
      },
    ])
  )

const dummyFetchSingleCompanyListTask = (id) =>
  new Promise((resolve, reject) =>
    id == 'bar'
      ? setTimeout(reject, 1000, Error('foo failed'))
      : setTimeout(
          resolve,
          1000,
          {
            foo: ["Ian's campervans", "Peter's RVs"],
            bar: ['Blah blah'],
            baz: ['Aaaa', 'Bbbbb', 'Ccccc'],
          }[id]
        )
  )

const CompanyLists = connect((state) => state.companyLists)(
  ({ lists, selectedId, listDetails, dispatch }) =>
    lists ? (
      <>
        Company lists
        <select
          onChange={(e) =>
            dispatch({
              type: 'COMPANY_LIST__SELECT',
              id: e.target.value,
            })
          }
        >
          {lists.map(({ id, name }) => (
            <option key={id} value={id}>
              {name}
            </option>
          ))}
        </select>
        {listDetails[selectedId] ? (
          <ul>
            {listDetails[selectedId].map((company) => (
              <li key={company} id={company}>
                {company}
              </li>
            ))}
          </ul>
        ) : (
          <Task.Status
            name="Company list"
            id={selectedId}
            progressMessage="Loading companies"
            startOnRender={{
              payload: selectedId,
              onSuccessDispatch: 'COMPANY_LISTS__COMPANIES_LOADED',
            }}
          />
        )}
      </>
    ) : (
      <Task.Status
        name="Company lists"
        id="dashboard"
        progressMessage="Loading company lists"
        startOnRender={{
          onSuccessDispatch: 'COMPANY_LISTS__LISTS_LOADED',
        }}
      />
    )
)

const sagaMiddleware = createSagaMiddleware()

const store = createStore(
  combineReducers({
    companyLists: companyListsReducer,
    tasks: tasksReducer,
  }),
  composeWithDevTools(applyMiddleware(sagaMiddleware))
)

sagaMiddleware.run(
  tasksSagaFactory({
    // Register tasks
    'Company lists': dummyFetchCompanyistsTask,
    'Company list': dummyFetchSingleCompanyListTask,
  })
)

ReactDOM.render(
  <Provider store={store}>
    <CompanyLists />
  </Provider>,
  document.getElementById('react-app'),
)

```

## Checklist


src/client/components/ProgressIndicator.jsx
```diff
@@ -0,0 +1,20 @@
...
+})
+
+const ProgressIndicator = ({ message }) => (
+  <StyledRoot>
+    <StyledLoadingBox loading={true} />
+    {message && <p>{message}</p>}
+  </StyledRoot>
+)
+
+export default ProgressIndicator
```
debitan: are you planning on also moving this to data-hub-components?

src/client/components/Task/index.jsx
```diff
@@ -0,0 +1,235 @@
...
+}
+
+/**
+ * @typedef SuccessAction
+ * @property {string} type - The value of the {onSuccessDispatch} option
+ * @property {string} name - The _task_ name
+ * @property {string} id - The _task_ id
+ * @property {any} payload - The payload with which the _task_ was started
+ * @property {any} result - The result with which the _task_ succeeded
+ */
```
debitan: do we need all of the comments in this file? The convention has been to not use comments, but I'm guessing the logic here is to explain how the saga works?

src/client/components/ProgressIndicator.jsx
peterhudec: This one could actually go there, but honestly, I don't thing there's currently any value in keeping components in a separate repo if they are only used in this project and there's no prospect of them being used anywhere else in forseeable future. It only adds a lot of friction to the process.

src/client/components/Task/index.jsx
peterhudec: If there is such a convention, then it is a very bad one, but maybe it only meant inline explanation comments. These however are not comments, but documentation, which is rather important. It would be very hard to decipher how this abstraction should be used by just looking at its implementation.

src/client/components/Task/index.jsx
```diff
@@ -0,0 +1,235 @@
+import _ from 'lodash'
```
paulgain:
```js
import { get } from 'lodash'
```

src/client/components/Task/index.jsx
```diff
@@ -0,0 +1,235 @@
...
+import PropTypes from 'prop-types'
+import React from 'react'
+import { connect } from 'react-redux'
+
+import { TASK__START } from '../../actions'
+import Err from './Error'
+import ProgressIndicator from '../ProgressIndicator'
+import { useEffect } from 'react'
```
paulgain: This line needs moving up  to the other `node_module` imports.

src/client/components/Task/reducer.js
```diff
@@ -0,0 +1,37 @@
+import _ from 'lodash'
```
paulgain:
```js
import { omit } from 'lodash'
```

src/client/components/Task/saga.js
```diff
@@ -0,0 +1,62 @@
+import _ from 'lodash'
```
paulgain:
```js
import { get } from 'lodash'
```

src/client/components/Task/Error.jsx
```diff
@@ -0,0 +1,51 @@
...
+  FOCUS_WIDTH,
+  MEDIA_QUERIES,
+  RESPONSIVE_4,
+} from '@govuk-react/constants'
+import { spacing } from '@govuk-react/lib'
+
+import SecondaryButton from '../SecondaryButton'
+
+// Taken from https://github.com/govuk-react/govuk-react/blob/master/components/error-summary/src/index.js
+const StyledRoot = styled.div(
```
rafenden: Please use `StatusMessage` instead:
https://uktrade.github.io/data-hub-components/?path=/story/statusmessage--default

src/client/components/Task/index.jsx
peterhudec: At some point we should configure webpack for dead code elimination / tree shaking.

src/client/components/Task/index.jsx
rafenden: We are sparingly using comments (or documentation in this case) because the code should be self-explanatory.

For example, one could argue that it's clear what `payload` variable is without the long description `The payload with which the _task_ was started`.

Having comments that spread across 100+ lines makes it hard to navigate across the source code.

A quote from Uncle Bob:
> A comment is a failure to express yourself in code.  If you fail, then write a comment; but try not to fail. — https://twitter.com/unclebobmartin/status/870311898545258497

src/client/components/Task/Error.jsx
peterhudec: `StatusMessage` is still quit specific in that it seems to be intended to only show a simple message, and it aplies the border color to the text as well. Here however, I have richer content which should not be colored:

Screen Shot 2020-01-22 at 4 32 53 PM

Ideally `StatusMessage` would be built on a lower lever `BorderBox` which I could reuse, but it is not. The same applies to the [`govuk-react/errror-summary`](https://govuk-react.github.io/govuk-react/?path=/story/form-errorsummary--component-default), which is very close to what I need, but it's [not composed of lower level components ](https://github.com/govuk-react/govuk-react/blob/3bc73f23480558e2a34a181361555d2cc3b9ae21/components/error-summary/src/index.js#L116)which I could reuse, but built from scratch.

I could probably make it work with `StatusMessage`, but it would feel like misusing it for something it was not inteded for. I would for example have to override the text color.

The right thing to do perhaps would be to refactor `StatusMessage` to be built on the hypothetical `BorderBox` which I could reuse here, but I'm not very keen to go down that rabbit hole.

src/client/components/Task/index.jsx
peterhudec: I'm pretty sure Uncle Bob meant an inline comment and not documentation, but I agree with your `payload` example.

I'm not sure what do you meen by 100+ lines as none of the lines in this whole module exceed 80 characters.

peterhudec: I don't really need those docstrings. But I was always considering not having documentation a pretty bad practice, especially in an open source project.

src/client/components/Task/Error.jsx
peterhudec: Had a look at `StatusMessage`, but unfortunately restyling it was not taken into consideration when it was implemented, so I can't do:
```js
const StyledStatusMessage = styled(StatusMessage)({
  color: 'initial'
})
```
It's because [the `StatusMessage` wraps `StyledStatusMessage` but doesn't propagate `className`](https://github.com/uktrade/data-hub-components/blob/master/src/status-message/StatusMessage.jsx#L18). Not only is the wrapper completely unnecessary, but also makes this component not further customizable with `styled-componets`.

The only way I can get rid of the color is to nest another styled component, which is approach I loathe:
```jsx
const ResetColor = styled.div({color: 'initial'})
const Err = ({ children }) =>
  <StatusMessage>
    <ResetColor>
      {children}
    </ResetColor>
  </StatusMessage>
```

src/client/components/Task/index.jsx
paulgain: Yes, it's always good to reduce the size of our Webpack bundle. 👍

src/client/components/Task/Error.jsx
rafenden: > I could probably make it work with StatusMessage, but it would feel like misusing it for something it was not inteded for. I would for example have to override the text color.

In my opinion, we should modify the `StatusMessage` component to meet your needs, it's fine to do so.

Another option would be to simplify the rich error message so the `StatusMessage` can be used without the required changes, example:

image


Creating new components very similar to each other will introduce maintenance nightmare.

src/apps/dashboard/client/CompanyLists/Table.jsx
```diff
@@ -7,7 +7,7 @@ import LinesEllipsis from 'react-lines-ellipsis'
 import styled from 'styled-components'
 
 import urls from '../../../../lib/urls'
-import SecondaryButton from './SecondaryButton'
+import SecondaryButton from '../../../../client/components/SecondaryButton'
```
rafenden: This should have been moved to the component library as other components could reuse this.

src/client/components/ProgressIndicator.jsx
rafenden: > I don't thing there's currently any value in keeping components in a separate repo

We decided to keep reusable components in the component library, so we should stick to that.

> there's no prospect of them being used anywhere else in forseeable future

https://github.com/uktrade/data-science-frontend is already using some of our components (Activity Feed).

Hopeful this will spread across other projects in DIT, as that was the purpose of creating a component library in the first place.

src/apps/dashboard/client/CompanyLists/Table.jsx
peterhudec: Maybe, but we need to discuss this. Anyway, I don't want to do it in this PR as it is only using it.

PR #2381, opened 2020-01-20
Refactor implementation of FieldDnbCompany on the "Add company form"

rafenden:
## Description of change

Update the implementation of `FieldDnbCompany` on the "Add company form" to use the new `entityRenderer` and `onCannotFind` props.

Depends on https://github.com/uktrade/data-hub-components/pull/245

## Test instructions

Check if the "Add company form" works as before.

### DIFF
--- a/src/apps/companies/apps/add-company/client/AddCompanyForm.jsx
+++ b/src/apps/companies/apps/add-company/client/AddCompanyForm.jsx
@@ -8,16 +8,11 @@ import axios from 'axios'
 import { get } from 'lodash'
 import { H3 } from '@govuk-react/heading'
 
-import {
-  FieldDnbCompany,
-  FieldRadios,
-  FieldSelect,
-  Form,
-  Step,
-} from 'data-hub-components'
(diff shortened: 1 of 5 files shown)


src/apps/companies/apps/add-company/client/AddCompanyForm.jsx
```diff
@@ -49,6 +51,48 @@ function AddCompanyForm({
...
+    function onEntityClick({ dnb_company }) {
+      setFieldValue('cannotFind', false)
+      setFieldValue('dnbCompany', dnb_company)
+      goForward()
+    }
+
+    return function(props) {
+      const { data } = props
+      const isClickable =
+        !data.datahub_company && !data.dnb_company.is_out_of_business
```
paulgain: Are we assuming that both `data` and `dnb_company ` will always be defined otherwise this line will through an error.

rafenden: Good point, the code was updated.

PR #2385, opened 2020-01-21
Move urls from nunjucks config

debitan:
## Description of change

Move the find exporters and support link from the global config and instead pass through the controller where needed using the urls module. This is in preparation for replacing the export page with react, which won't have access to nunjucks globals. 

## Test instructions

Nothing should change.

## Checklist

### DIFF
--- a/src/apps/global-nav-items.js
+++ b/src/apps/global-nav-items.js
@@ -3,6 +3,7 @@ const path = require('path')
 const { compact, sortBy, concat, includes } = require('lodash')
 
 const config = require('../config')
+const urls = require('../lib/urls')
 
 const subApps = fs.readdirSync(__dirname)
 
@@ -33,7 +34,7 @@ const GLOBAL_NAV_ITEMS = concat(
     key: 'datahub-mi',
   },
   {
-    path: config.findExportersUrl,
+    path: urls.external.findExporters(),
     label: 'Find exporters',
     key: 'find-exporters',
   }
--- a/src/config/nunjucks/globals.js
+++ b/src/config/nunjucks/globals.js
@@ -9,9 +9,6 @@ module.exports = {
   projectPhase: config.projectPhase,
   description:
     'Data Hub is a customer relationship, project management and analytical tool for Department for International Trade.',
-  feedbackLink: '/support',
-  findExportersUrl: config.findExportersUrl,
-
   assign,
   urls,
 
--- a/src/lib/urls.js
+++ b/src/lib/urls.js
@@ -68,6 +68,7 @@ module.exports = {
     greatProfile: (id) => config.greatProfileUrl.replace('{id}', id),
     companiesHouse: (companyNumber) =>
       `https://beta.companieshouse.gov.uk/company/${companyNumber}`,
+    findExporters: () => 'https://find-exporters.datahub.trade.gov.uk/',
   },
   dashboard: url('/'),
   companies: {
--- a/src/apps/companies/apps/exports/controller.js
+++ b/src/apps/companies/apps/exports/controller.js
@@ -44,6 +44,8 @@ function renderExports(req, res) {
     .render('companies/apps/exports/views/exports-view', {
       exportDetails,
       exportPotentials: Object.values(exportPotentialLabels),
+      findExportersUrl: urls.external.findExporters(),
+      feedbackLink: urls.support(),
     })
 }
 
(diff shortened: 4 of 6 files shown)


web-bert:
@debitan I think you need to `git rebase --onto` as this must be branched from your move exports branch instead of `master`

debitan:
> @debitan I think you need to `git rebase --onto` as this must be branched from your move exports branch instead of `master`

@web-bert think I've fixed that now

src/lib/urls.js
```diff
@@ -68,6 +68,7 @@ module.exports = {
     greatProfile: (id) => config.greatProfileUrl.replace('{id}', id),
     companiesHouse: (companyNumber) =>
       `https://beta.companieshouse.gov.uk/company/${companyNumber}`,
+    findExporters: 'https://find-exporters.datahub.trade.gov.uk/',
```
web-bert: This should be a function as that is the pattern for all urls

src/config/nunjucks/globals.js
```diff
@@ -9,8 +9,7 @@ module.exports = {
   projectPhase: config.projectPhase,
   description:
     'Data Hub is a customer relationship, project management and analytical tool for Department for International Trade.',
-  feedbackLink: '/support',
-  findExportersUrl: config.findExportersUrl,
+  findExportersUrl: urls.external.findExporters,
```
web-bert: Is this needed?

debitan: no it's not. good shout.

src/lib/urls.js
web-bert: Just to keep them consistent, I shouldn't have to check whether I call one or use the property value directly - so they should all be functions that you call to get a url regardless of if it just returns a string or generates a string with or without input

web-bert: Also we may change this in the future to generate a value so a function gives us the flexibility to do what we want

src/apps/companies/apps/exports/exports.js
```diff
@@ -44,6 +44,8 @@ function renderExports(req, res) {
     .render('companies/apps/exports/views/exports-view', {
       exportDetails,
       exportPotentials: Object.values(exportPotentialLabels),
+      findExportersUrl: urls.external.findExporters(),
+      feedbackLink: urls.support(),
```
rafenden: I suggest keeping this variable consistent and call it `feedbackUrl`.

debitan: I agree. Will change it.

debitan: @rafenden  Actually, I am already changing this in an upcoming PR as I'm redoing this page in React. So will leave it for now to save refactoring the nunjucks template.

src/apps/global-nav-items.js
```diff
@@ -33,7 +34,7 @@ const GLOBAL_NAV_ITEMS = concat(
     key: 'datahub-mi',
   },
   {
-    path: config.findExportersUrl,
+    path: urls.external.findExporters,
```
web-bert: This needs to be a function call?

src/config/nunjucks/globals.js
```diff
@@ -9,9 +9,6 @@ module.exports = {
   projectPhase: config.projectPhase,
   description:
     'Data Hub is a customer relationship, project management and analytical tool for Department for International Trade.',
-  feedbackLink: '/support',
```
web-bert: This is removed but there are no files updated to call the new url function?

src/apps/global-nav-items.js
debitan: changed

src/config/nunjucks/globals.js
debitan: it was only used in the export-view.njk and I'm now passing it explicitly through the exports controller.

web-bert: I thought it was used by the header, but it was built into the component 👍

src/lib/urls.js
```diff
@@ -68,6 +68,7 @@ module.exports = {
     greatProfile: (id) => config.greatProfileUrl.replace('{id}', id),
     companiesHouse: (companyNumber) =>
       `https://beta.companieshouse.gov.uk/company/${companyNumber}`,
+    findExporters: () => 'https://find-exporters.datahub.trade.gov.uk/',
```
web-bert: Missing a test for this

debitan: will add in a bit. forgot we have tests for this file. was thinking it was just covered by tests when it's implemented.

PR #2387, opened 2020-01-21
Redirect to the matching confirmation page after a DNB company is clicked

rafenden:
## Description of change

Redirect user to the matching confirmation page after a D&B company is clicked from the search results list.

## Test instructions

1. Go to the company match page `companies/COMPANY_ID/match`
2. Search for a company
3. Click one of the results
4. Check if you are on the company match confirmation page


src/apps/companies/apps/match-company/client/FindCompany.jsx
```diff
@@ -1,22 +1,28 @@
...
+        entityRenderer={(props) => (
+          <EntityListItem
+            onEntityClick={({ dnb_company }) =>
+              window.location.assign(
+                `/companies/${company.id}/match/${dnb_company.duns_number}`
+              )
+            }
+            {...props}
+          />
+        )}
```
paulgain: Don't forget to add the company `countryCode` to the `propTypes`.

rafenden: Prop added, thanks!

PR #2400, opened 2020-01-27
Move great.gov url into urls module and remove config from that file

debitan:
## Description of change

Move great.gov url into urls module and remove config from that file. The means you no longer have to have the env variable `GREAT_PROFILE_URL`. 

If the great.gov.uk url changes in future, rather than change the env var, we will need to edit it in the urls module and do a release. However, we now no longer need to import the config file into the urls module and no longer need proxyrequire in unit tests when testing the great profile url. 

## Test instructions

Nothing should change. 

## Checklist

### DIFF
--- a/src/lib/__test__/urls.test.js
+++ b/src/lib/__test__/urls.test.js
@@ -1,23 +1,12 @@
 const faker = require('faker')
-const proxyquire = require('proxyquire')
-
-const modulePath = '../urls'
+const urls = require('../urls')
 
 describe('urls', () => {
-  let urls
-  before(() => {
-    urls = proxyquire(modulePath, {
-      '../config': {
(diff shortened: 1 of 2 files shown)


src/lib/__test__/urls.test.js
```diff
@@ -1,23 +1,12 @@
...
-      },
-    })
-  })
-
   describe('external', () => {
     it('should have the correct urls', () => {
       const companyNumber = faker.random.alphaNumeric(8)
       expect(urls.external.greatProfile(companyNumber)).to.equal(
-        `http://a.b.c.com/path/${companyNumber}`
+        `https://www.great.gov.uk/international/trade/suppliers/${companyNumber}`
```
paulgain: I think we should centralise all external URLs (like the urls module) so we can easily reuse them throughout the app?

PR #2401, opened 2020-01-27
Update Matching Interface Flash Message

paulgain:
## Update of the Flash message
Provide more information to the user once that have matched a company. 
### Before
before
### After
after
## Checklist

### DIFF
--- a/src/apps/companies/apps/match-company/controllers.js
+++ b/src/apps/companies/apps/match-company/controllers.js
@@ -151,7 +151,12 @@ async function submitMatchRequest(req, res) {
     const ticket = createMatchRequestMessage(req, res)
     const result = await postToZenDesk(ticket)
 
-    req.flash('success', 'Company record update request sent')
+    req.flash(
+      'success',
+      'Your request to update this company record has been sent.\n' +
+        'In the near future the Support Team will update this company ' +
+        'record and enable automatic updates for future changes to the company information.'
+    )
     res.json({ message: 'OK', ticket: get(result, 'data.ticket.id') })
   } catch (error) {
     const statusCode = get(error, 'response.status', 500)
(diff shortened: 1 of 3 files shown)


test/functional/cypress/specs/companies/match-company-spec.js
```diff
@@ -130,7 +130,8 @@ describe('Match a company', () => {
     it('displays the "Company record update request sent" flash message', () => {
       cy.get(localHeader().flash).should(
         'contain.text',
```
rafenden: We could be more strict here and use `have.text` instead.

paulgain: Initially I did try that, however, the test failed due to the new line.

```html
<li class="c-message c-message--success">
  Your request to match this company record has been sent.
  <br>
  In the near future the Support Team will update this company record and enable automatic updates for future changes to 
  the company information.
  <a class="c-message__close" href="#">Dismiss</a>
</li>
```

web-bert:
@paulgain we had a similar request to put additional text in the flash message. Should we update it to have a heading and body text instead? The all bold text seems a bit too much...

paulgain:
> @paulgain we had a similar request to put additional text in the flash message. Should we update it to have a heading and body text instead? The all bold text seems a bit too much...

@web-bert Good idea, I've just spoken to Roger and he also agrees - let's do it!

src/apps/companies/apps/match-company/controllers.js
```diff
@@ -151,7 +151,11 @@ async function submitMatchRequest(req, res) {
     const ticket = createMatchRequestMessage(req, res)
     const result = await postToZenDesk(ticket)
 
-    req.flash('success', 'Company record update request sent')
+    req.flash(
+      'success',
+      `Your request to match this company record has been sent.
+       In the near future the Support Team will update this company record and enable automatic updates for future changes to the company information.`
```
peterhudec: I would line break it so that it fits into 80 chractaers.

paulgain: 👍

PR #2403, opened 2020-01-27
Update sandbox docs

debitan:
Update sandbox docs to reflect the new, easier way, to run functional tests with docker now that the sandbox is in the same repo as the frontend.

Remove the exposure of redis to a port to allow the simultaneous running of the api and sandbox.

### DIFF
--- a/docs/Running tests.md
+++ b/docs/Running tests.md
@@ -154,40 +154,23 @@ Updating the baseline consists in 2 steps:
 
 - 2:. Run `$ yarn test:visual:update` to update the failed tests with updated images of how the page in test should look like.
 
-
 ## Sandbox (API mocks)
 
 Sandbox is as a light replacement for API backend and it's used only by functional tests.
 
 ### Using sandbox within docker (preferred method)
 
-1. Clone this repo
+1. cd into `data-hub-frontend` and run `docker-compose up`. This will start up the sandbox api in conjunction with the frontend. Make sure nothing else is running at ports 8001 and 3000.
 
-2. `cd` into the folder you just cloned
+2. If you make changes to sandbox you will need to rebuild docker with `docker-compose up -d --no-deps --build mock-backend` before running it again with `docker-compose up`. If you've also made changes to the front end and need to rebuild the full stack, use `docker-compose up --build --force-recreate`.
 
-3. Build the image:
+### Using sandbox within docker with local frontend
 
-   ```bash
-   docker build -t data-hub-sandbox .
-   ```
+1. cd into `test/sandbox` and run `docker run --rm --name data-hub-sandbox -it -p 8001:8001 data-hub-sandbox`.
 
-4. Start sandbox server on port `8001`:
+2. Change your `API_ROOT` in your env file to point to `http://mock-backend:8001` and then run the frontend locally with `yarn develop`.
 
-   ```bash
-   docker run -it -p 8001:8001 data-hub-sandbox
-   ```
-
-   Stop the server by pressing `CTRL`+`C`
-
-### Additional steps if you're running sandbox with Data Hub Frontend for functional testing
-
-1. You'll also have to run redis as the sandbox does not come with this and will throw an error on the frontend without it. So open a new terminal window and run `docker run -it -p 6379:6379 redis:3.2`
-
-2. `cd` into your datahub-front-end folder
-
-3. Run `yarn run develop` and you should now be all set up
-
-4. When you make changes to the fixtures, you'll need to rebuild the docker image and then run the server again.
+3. If you make changes to sandbox rebuild docker with `docker build -t data-hub-sandbox`.
 
 ### Using sandbox on host machine
 
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,6 +1,5 @@
-version: "3"
+version: '3'
 services:
-
   frontend:
     build:
       context: .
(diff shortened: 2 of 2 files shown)


docs/Running tests.md
```diff
@@ -166,33 +166,7 @@ Sandbox is as a light replacement for API backend and it's used only by function
...
-### Additional steps if you're running sandbox with Data Hub Frontend for functional testing
-
-1. You'll also have to run redis as the sandbox does not come with this and will throw an error on the frontend without it. So open a new terminal window and run `docker run -it -p 6379:6379 redis:3.2`
-
-2. `cd` into your datahub-front-end folder
-
-3. Run `yarn run develop` and you should now be all set up
-
-4. When you make changes to the fixtures, you'll need to rebuild the docker image and then run the server again.
+1. cd into `test/sandbox/` and run `docker-compose up`. This will start up the sandbox api in conjunction with the frontend. Make sure nothing else is running at ports 6379, 8001 and 3000.
```
rafenden: This is not quite right, we don't need to enter the `test/sandbox/` dir when using `docker-compose`.

debitan:
@PippoRaimondi do we need to use the yaml file from sandbox on running it. @rafenden reckons just running `docker-compose up` would suffice. And does this mean that running docker in the frontend you can only connect to the sandbox api and not to any other backends environments?

debitan:
I've added instructions for rebuilding the docker image when changes to sandbox are made

docs/Running tests.md
```diff
@@ -166,33 +166,9 @@ Sandbox is as a light replacement for API backend and it's used only by function
...
-### Additional steps if you're running sandbox with Data Hub Frontend for functional testing
-
-1. You'll also have to run redis as the sandbox does not come with this and will throw an error on the frontend without it. So open a new terminal window and run `docker run -it -p 6379:6379 redis:3.2`
-
-2. `cd` into your datahub-front-end folder
-
-3. Run `yarn run develop` and you should now be all set up
-
-4. When you make changes to the fixtures, you'll need to rebuild the docker image and then run the server again.
+2. If you make changes to sandbox you will need to rebuild docker with `docker build -t data-hub-sandbox .` before running it again with `docker-compose up`.
```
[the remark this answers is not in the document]
debitan: @PippoRaimondi I would welcome your thoughts on this if you get a chance? No rush though.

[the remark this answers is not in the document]
rafenden: Have you tried `docker-compose up --build --force-recreate`?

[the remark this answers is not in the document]
PippoRaimondi: @debitan I missed this comment... reinforcing what @rafenden said above, when we are not sure sandbox image is up to date we should run the command above, but maybe with an additional parameter to prevent it rebuilding all services:

`$docker-compose up -d --no-deps --build mock-backend`

debitan:
@PippoRaimondi @rafenden have updated in line with your comments. all good?

docs/Running tests.md
```diff
@@ -166,33 +166,9 @@ Sandbox is as a light replacement for API backend and it's used only by function
...
-### Additional steps if you're running sandbox with Data Hub Frontend for functional testing
-
-1. You'll also have to run redis as the sandbox does not come with this and will throw an error on the frontend without it. So open a new terminal window and run `docker run -it -p 6379:6379 redis:3.2`
-
-2. `cd` into your datahub-front-end folder
-
-3. Run `yarn run develop` and you should now be all set up
-
-4. When you make changes to the fixtures, you'll need to rebuild the docker image and then run the server again.
+2. If you make changes to sandbox you will need to rebuild docker with `docker-compose up -d --no-deps --build mock-backend` before running it again with `docker-compose up`.
```
PippoRaimondi: you won't need to run `docker-compose up` again, you are already bringing the stack up with the below command but ensuring it rebuilds sandbox before.

`docker-compose up -d --no-deps --build mock-backend`

PippoRaimondi: the above is incorrect, it won't bring the stack up, only rebuild a specific service. so documentation is ok. Will resolve this conversation

docs/Running tests.md
```diff
@@ -166,33 +166,17 @@ Sandbox is as a light replacement for API backend and it's used only by function
...
-
-### Additional steps if you're running sandbox with Data Hub Frontend for functional testing
-
-1. You'll also have to run redis as the sandbox does not come with this and will throw an error on the frontend without it. So open a new terminal window and run `docker run -it -p 6379:6379 redis:3.2`
-
-2. `cd` into your datahub-front-end folder
+1. cd into `test/sandbox` and run `docker run --rm --name data-hub-sandbox -it -p 8001:8001 data-hub-sandbox`. 
 
-3. Run `yarn run develop` and you should now be all set up
+2. Change your `API_ROOT` in your env file to point to `http://mock-backend:8001` and then run the frontend locally with `yarn develop`.
```
PippoRaimondi: I don't think you need to change the API_ROOT: using service names is a way to access containers within a stack... since you are starting the container using `docker run` and exposing the port 8001 `-p 8001:8001` your localhost has access to this containers port hence the API_ROOT could be left alone with `http://localhost:8001`

debitan: By this I mean that the default API_ROOT is `localhost:8000` so you would need to point it to 8001 to get the sandbox api.

PippoRaimondi: got it! I am ok with that!

PR #2406, opened 2020-01-28
Remove Export tab for non DIT users

web-bert:
## Description of change

As we enrich the data on the Export tab, we need to ensure we only show it to the relevant users. Therefore this tab is only available to users with the DIT permissions.

## Test instructions

As a DIT user you should see the Export tab, as a DA or LEP user you should not

### Before

Screenshot 2020-01-28 at 20 45 21

### After
Screenshot 2020-01-28 at 20 44 44

## Checklist

### DIFF
--- a/src/apps/companies/constants.js
+++ b/src/apps/companies/constants.js
@@ -53,6 +53,7 @@ const LOCAL_NAV = [
   {
     path: 'exports',
     label: 'Export',
+    permissions: ['company.view_companyexportcountry'],
   },
   {
     path: 'orders',
--- a/src/apps/companies/middleware/__test__/local-navigation.test.js
+++ b/src/apps/companies/middleware/__test__/local-navigation.test.js
@@ -31,6 +31,7 @@ describe('Companies local navigation', () => {
             'company.view_contact',
             'investment.view_all_investmentproject',
             'order.view_order',
+            'company.view_companyexportcountry',
           ],
         },
       })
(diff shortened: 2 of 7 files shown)


src/apps/companies/constants.js
```diff
@@ -53,6 +53,7 @@ const LOCAL_NAV = [
   {
     path: 'exports',
     label: 'Export',
+    permissions: ['interaction.view_all_interaction'],
```
rafenden: Shall we consider adding `exports.view_all_exports` as new permission?

web-bert: I was hoping someone would pick this up, I don't know what the correct thing is here. We want to match the LEP and DA permissions currently and this seemed to provide that - and the information to hide from them is interactions data so it seemed to make sense to use that

web-bert: We have discussed the permissions here and I will be updating this PR once they are in place

web-bert:
We are going to use different permissions for this so this PR will be updated in due course

src/apps/companies/constants.js
```diff
@@ -53,7 +53,7 @@ const LOCAL_NAV = [
   {
     path: 'exports',
     label: 'Export',
-    permissions: ['interaction.view_all_interaction'],
+    permissions: ['company.view_companyexportcountry'],
```
rafenden: Sorry for being picky, but shouldn't we use underscores to separate words like we did elsewhere?

web-bert: @rafenden these are auto generated and there are similar ones here: https://github.com/uktrade/data-hub-frontend/blob/master/test/sandbox/fixtures/whoami.json

PR #2407, opened 2020-01-29
Replace export page with React

debitan:
The release of this PR is dependent on yellow turning on a feature flag.

The unit test coverage also drops ever so slightly in the controller tests. With the new controller testing the one line using mocha was extremely difficult, instead it is covered in the new functional tests. 

Ignoring whitespace the PR is +536 −229. Sorry for the slightly large size. Part of this is down to the fact that there were no functional tests in the area and I've had to write new ones from scratch.

## Description of change

Render a new export page design with React. This rearranges the existing data and adds the 'Countries of no interest' field. 

## Test instructions

When you go to the export page of a company you will see the new React page as shown in the 'after' screenshot below. 

### Before

FireShot Capture 030 - Exports - ONEBUTTON LIMITED - Companies - DIT Data Hub_ - www datahub trade gov uk

### After

FireShot Capture 031 - Exports - One List Corp - Companies - DIT Data Hub - localhost

## Checklist


src/apps/companies/apps/exports/client/ExportIndex.jsx
```diff
@@ -0,0 +1,111 @@
...
+  findExportersUrl,
+  supportUrl,
+  exportWinsUrl,
+  editCountriesUrl,
+}) => {
+  return (
+    <>
+      <SummaryTable caption="Exports">
+        {exportDetails.map((row) =>
+          row.value.url ? (
```
rafenden: This condition could be within `SummaryTable.Row`, e.g.

```suggestion
          <SummaryTable.Row heading={row.heading} key={row.heading}>
            {row.value.url ? <span><Link>...</Link></span> : row.value}
          </SummaryTable.Row>
```

src/apps/companies/apps/exports/client/ExportIndex.jsx
```diff
@@ -0,0 +1,111 @@
...
+            </SummaryTable.Row>
+          )
+        )}
+      </SummaryTable>
+
+      <Details summary="What is export potential">
+        <p>
+          The export potential score is a prediction of a company's likelihood
+          of exporting, and was originally created for the{' '}
+          <Link href={findExportersUrl} target="_blank">
```
rafenden: You could reuse the existing component which also adds some useful aria attributes:
https://github.com/uktrade/data-hub-components/blob/master/src/new-window-link/NewWindowLink.jsx

src/apps/companies/apps/exports/client/ExportIndex.jsx
```diff
@@ -0,0 +1,111 @@
...
+        <p>
+          <br />
+          We are continuing to improve the algorithm so please do share your
+          feedback or let us know of any anomalies through the{' '}
+          <a href={supportUrl}>support channel</a>.
+        </p>
+      </Details>
+
+      <H3>Export countries information</H3>
+      <NoTopMarginSummaryTable>
```
rafenden: We usually prefix styled components with `Styled`, e.g. `StyledNoTopMarginSummaryTable`

test/functional/cypress/specs/companies/export-spec.js
```diff
@@ -1,18 +1,192 @@
 const fixtures = require('../../fixtures')
-const selectors = require('../../../../selectors')
+const { assertBreadcrumbs } = require('../../support/assertions')
 
 describe('Companies Export', () => {
+  context(
+    'when viewing the export tab for an active company with no export information',
+    () => {
+      before(() => {
+        cy.visit(`/companies/${fixtures.company.dnbCorp.id}/exports`)
```
rafenden: I think we should also use the urls module in our tests.

src/apps/companies/apps/exports/client/ExportIndex.jsx
```diff
@@ -0,0 +1,111 @@
...
+              <strong>{category.text}</strong> - {category.description}
+            </li>
+          ))}
+        </ol>
+
+        <p>
+          <br />
+          We are continuing to improve the algorithm so please do share your
+          feedback or let us know of any anomalies through the{' '}
+          <a href={supportUrl}>support channel</a>.
```
rafenden: We could use the urls module directly here.

debitan: I am doing in the controller and then passing it in as a prop. Suppose I could import the urls module directly into the component now that it no longer imports config. However, I was thinking it was cleaner to show what data was available to the component by explicitly listing it in the controller.

src/apps/companies/apps/exports/client/ExportIndex.jsx
```diff
@@ -0,0 +1,111 @@
...
+      {isArchived ? null : (
+        <Button
+          as={Link}
+          href={editCountriesUrl}
+          buttonColour={GREY_3}
+          buttonTextColour={BLACK}
+          data-test-id="editExportCountriesButton"
+        >
+          Edit export countries
+        </Button>
```
peterhudec: Use the [`src/client/components/SecondaryButton`](https://github.com/uktrade/data-hub-frontend/blob/master/src/client/components/SecondaryButton.jsx) component.

src/apps/companies/apps/exports/client/ExportIndex.jsx
```diff
@@ -0,0 +1,111 @@
...
+  exportPotentials,
+  findExportersUrl,
+  supportUrl,
+  exportWinsUrl,
+  editCountriesUrl,
+}) => {
+  return (
+    <>
+      <SummaryTable caption="Exports">
+        {exportDetails.map((row) =>
```
peterhudec: I would call it `exportDetail` instead of `row`, even better, I would destructure `({heading, value})` here.

src/apps/companies/apps/exports/transformer.js
```diff
@@ -21,61 +20,72 @@ function getGreatProfileValue(profileStatus, companiesHouseNumber) {
...
 
-  // Whilst we have the feature flag we have to delete the new label
-  // Otherwise a new field will be displayed with no value
-  // @TODO remove this when the feature flag is removed
-  if (!useNewCountries) {
-    delete labels.noInterestCountries
-  }
+  const exportDetails = [
+    {
+      heading: exportDetailsLabels.exportExperienceCategory,
```
peterhudec: I would name it `name` instead of `heading`. `heading` implies some notion of UX, but this part of code shouldn't care about that. `name` is more consistent with the rest of the code and the API.

src/apps/companies/apps/exports/transformer.js
```diff
@@ -21,61 +20,72 @@ function getGreatProfileValue(profileStatus, companiesHouseNumber) {
...
+    noInterestCountries,
+  } = getCountriesFields(company)
 
-  // Whilst we have the feature flag we have to delete the new label
-  // Otherwise a new field will be displayed with no value
-  // @TODO remove this when the feature flag is removed
-  if (!useNewCountries) {
-    delete labels.noInterestCountries
-  }
+  const exportDetails = [
```
peterhudec: An object is perhaps a better data structure for mapping a name to a value:
```js
const exportDetails = {
  [exportDetailsLabels.exportExperienceCategory]: company.export_experience_category
    && company.export_experience_category.name,
  [exportDetailsLabels.greatProfile]: getGreatProfileValue(
    company.great_profile_status,
    company.company_number,
  ),
  [exportDetailsLabels.exportPotential]: getExportPotentialLabel(company.export_potential),
}
```

src/apps/companies/apps/exports/transformer.js
```diff
@@ -21,61 +20,72 @@ function getGreatProfileValue(profileStatus, companiesHouseNumber) {
...
+    {
+      heading: exportDetailsLabels.exportExperienceCategory,
+      value: company.export_experience_category
+        ? company.export_experience_category.name
+        : 'None',
```
peterhudec: Falling back to `'None'` is a concern of the presentation layer and should be done there. You should fallback to any _falsy_ value here, which you can do with `_.get(company, 'export_experience_category.name')`.

src/apps/companies/apps/exports/transformer.js
```diff
@@ -21,61 +20,72 @@ function getGreatProfileValue(profileStatus, companiesHouseNumber) {
...
+      heading: exportDetailsLabels.exportToCountries,
+      value: exportToCountries || 'None',
+    },
+    {
+      heading: exportDetailsLabels.futureInterestCountries,
+      value: futureInterestCountries || 'None',
+    },
+    {
+      heading: exportDetailsLabels.noInterestCountries,
+      value: noInterestCountries || 'None',
```
peterhudec: Same as above, you don't need to fall back to anything here. Do it in the view.

src/apps/companies/apps/exports/transformer.js
```diff
@@ -21,61 +20,72 @@ function getGreatProfileValue(profileStatus, companiesHouseNumber) {
...
+        company.company_number
+      ),
+    },
+    {
+      heading: exportDetailsLabels.exportPotential,
+      value: getExportPotentialLabel(company.export_potential),
+    },
+  ]
+
+  const exportCountriesInformation = [
```
peterhudec: Same as above, an object is a better data structure to represent a mapping from name to value.

test/functional/cypress/specs/companies/export-spec.js
```diff
@@ -1,18 +1,192 @@
...
+        cy.get('td')
+          .eq(2)
+          .should('have.text', 'No score given')
+      })
+
+      it('should render the "What is export potential" dropdown', () => {
+        cy.get('details summary span').should(
+          'have.text',
+          'What is export potential'
+        )
```
peterhudec: You are tightly coupling the tests to the markup, which can be considered an implementation detail. If you removed the redundand `<span>` as I suggested before, this test would break although it still pretty much works.

I try to write cypress tests from the perspective of the user, so I look for an element containing some expected text and then make some assertions about it, but in this case you can just do:
```js
cy.contains('What is export potential')
```

test/functional/cypress/specs/companies/export-spec.js
```diff
@@ -1,18 +1,192 @@
...
+        cy.get('td')
+          .eq(5)
+          .should('have.text', 'None')
+      })
+
+      it('should render the "Edit export countries" button', () => {
+        cy.get('[data-test-id="editExportCountriesButton"]').should(
+          'have.text',
+          'Edit export countries'
+        )
```
peterhudec: If written from the user's perspective, it would read:
```js
cy.contains('Edit export countries').should('have.prop', 'tagName', 'BUTTON')
```

test/functional/cypress/specs/companies/export-spec.js
```diff
@@ -1,18 +1,192 @@
...
+      })
+
+      it('should render the "Edit export countries" button with the correct link', () => {
+        cy.get('[data-test-id="editExportCountriesButton"]')
+          .invoke('attr', 'href')
+          .should(
+            'eq',
+            '/companies/cc7e2f19-7251-4a41-a27a-f98437720531/exports/edit'
+          )
```
peterhudec: You can do `.should('have.attr', 'href', '/companies/.../exports/edit')`.

test/functional/cypress/specs/companies/export-spec.js
```diff
@@ -1,18 +1,192 @@
...
+        cy.get('h3')
+          .eq(1)
+          .siblings('p')
+          .should('have.text', 'Record your win on our Export Wins site')
+      })
+
+      it('should render the link to Export Wins', () => {
+        cy.get('[data-test-id="exportWinsLink"]')
+          .invoke('attr', 'href')
+          .should('eq', 'https://www.exportwins.service.trade.gov.uk/')
```
peterhudec: Use `.should('have.attr', 'href', ...)`

test/functional/cypress/specs/companies/export-spec.js
```diff
@@ -1,18 +1,192 @@
...
+          .should('have.text', '"Find a supplier" profile (Opens in a new tab)')
+        cy.get('td a')
+          .invoke('attr', 'href')
+          .should(
+            'eq',
+            'https://www.great.gov.uk/international/trade/suppliers/12345678'
+          )
+        cy.get('td a')
+          .invoke('attr', 'target')
+          .should('eq', '_blank')
```
peterhudec: Use `.should('have.attr', 'target', ...)`

src/apps/companies/apps/exports/client/ExportIndex.jsx
rafenden: That's a valid point, currently we are passing the URLs in both ways so probably would be better chose one going forward and make it a convention.

src/apps/companies/apps/exports/client/ExportIndex.jsx
```diff
@@ -0,0 +1,111 @@
...
+          row.value.url ? (
+            <SummaryTable.Row heading={row.heading} key={row.heading}>
+              <span>
```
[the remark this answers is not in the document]
debitan: they're needed because of the space `{'  '}`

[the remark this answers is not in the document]
debitan: although I could use a fragment if that is preferable?

src/apps/companies/apps/exports/client/ExportIndex.jsx
debitan: we discussed. I'll use the urls module directly in the component and not pass them through the controller.

src/apps/companies/apps/exports/transformer.js
debitan: I think that would be a good approach if all objects in this array were treated the same by the presentation layer. However, each object has it's own text for a falsy value, be that `No profile` or `No score given`. Therefore, in this instance, it is simpler to handle this in the transformer rather shifting the weight of this to the client. Additionally, much of the code to do this transformation already exists.

src/apps/companies/apps/exports/transformer.js
debitan: If it's ok I'm going to retain the current structure as it makes it easier and cleaner to map over in the component.

src/apps/companies/apps/exports/transformer.js
debitan: while the logic I referenced above doesn't apply here as they are all treated in the same way, I would propose keeping it this way for consistency with the above.

src/apps/companies/apps/exports/client/ExportIndex.jsx
debitan: Actually, I've discussed this with Farhad. We're not going to show 'Opens in a new window' on this page, so I'll just use the `Link` component, but with added `aria-label`. The `NewWindowLink` comes with the text '(Opens in a new window)'. This isn't great in the middle of a sentence - which is the case in some places on this page -, the 'O' shouldn't be capitalised, and it is actually opening it in a new tab, not a new window.

debitan:
just realised I hadn't used propTypes at all. will add this in now.

src/apps/companies/apps/exports/controller.js
```diff
@@ -53,20 +53,26 @@ function getPostedFormData(body) {
...
+
+  const {
+    exportDetails,
+    exportCountriesInformation,
+  } = transformCompanyToExportDetailsView(company)
 
   res
     .breadcrumb(company.name, urls.companies.detail(company.id))
     .breadcrumb('Exports')
     .render('companies/apps/exports/views/exports-view', {
```
web-bert: I think the template should be called `index.njk` ( as in: `.render('companies/apps/exports/views/index`), as that is what is called in `urls` and the components name - or it should match the components name (`exports-index.njk`)

src/apps/companies/apps/exports/client/ExportIndex.jsx
```diff
@@ -0,0 +1,135 @@
+import React from 'react'
```
web-bert: I think this name should be plural: `ExportsIndex.jsx`

PR #2410, opened 2020-01-30
Fix: use new api for company exports

web-bert:
## Description of change

- API has changed based on feedback on the BE PR, this is to reflect those changes
- refactor the controller to stop sending the whole company back to the API
- refactor the tests to remove `this`

## Test instructions

Feature flag: `interaction-add-countries`

### Feature flag off:
No change

### Feature flag on
Edit the export countries of a company and save. The changes should be reflected after a save and no errors should be thrown, apart from when you try and add the same county to more than one category

## Checklist

### DIFF
--- a/src/apps/companies/apps/exports/controller.js
+++ b/src/apps/companies/apps/exports/controller.js
@@ -6,7 +6,7 @@ const urls = require('../../../../lib/urls')
 const groupExportCountries = require('../../../../lib/group-export-countries')
 const getExportCountries = require('../../../../lib/get-export-countries')
 
-const { saveCompany } = require('../../repos')
+const { updateCompany, saveCompanyExportDetails } = require('../../repos')
 const { transformObjectToOption } = require('../../../transformers')
 const transformCompanyToExportDetailsView = require('./transformer')
 const { exportDetailsLabels, exportPotentialLabels } = require('../../labels')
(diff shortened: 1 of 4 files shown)


src/apps/companies/repos.js
```diff
@@ -122,6 +122,14 @@ function saveDnbCompanyInvestigation(token, company) {
...
 }
 
+function saveCompanyExportDetails(token, companyId, body) {
+  return authorisedRequest(token, {
+    body,
+    url: `${config.apiRoot}/v4/company/${companyId}/export-detail`,
+    method: 'PATCH',
+  })
+}
+
```
peterhudec: Now that we have the `/api-proxy` endpoint, we don't need to deal with the token like this anymore. You can just make a request to `/api-proxy/v4/company/:id/export-detail`, which adds the token to the request automatically. You only need to add `'/v4/company/:id/export-detail'` to the [whitelisted API endpoints in the config](https://github.com/uktrade/data-hub-frontend/blob/master/src/config/index.js#L46).

src/apps/companies/apps/exports/controller.js
```diff
@@ -127,34 +127,36 @@ function renderExportEdit(req, res) {
...
+      ])
+
+      Object.assign(data, {
+        export_to_countries: filter(exportToCountries),
+        future_interest_countries: filter(futureInterestCountries),
+      })
+    }
 
-    res.redirect(urls.companies.exports.index(save.id))
+    await updateCompany(token, companyId, data)
```
peterhudec: You can call ``axios.patch(`/api-proxy/v4/company/${companyId}/export-detail`, data)`` directly here, but I guess the `authorisedRequest` also does some logging?

src/apps/companies/repos.js
web-bert: This is a Node request to the API not via AJAX, so it can't make a request to itself to use the proxy?

src/apps/companies/apps/exports/controller.js
web-bert: Same here, this is Node making the request directly

src/apps/companies/repos.js
peterhudec: off course 🤦‍♂

PR #2415, opened 2020-02-06
Change OMIS UK region filter to typeahead

peterhudec:
## Description of change

The UK Region filter on the Orders (OMIS) page is now a _typeahead_.

## Test instructions

1. Go to the Orders section.
2. On the very bottom of the filters sidebar, you should see that the UK region filter is now a _typeahead_
3. Focus the field, type the name of your region of choice, select it by either clicking on it or hit enter
4. You should see that the search results reflect the filter's setting

### Before

Screen Shot 2020-02-06 at 11 27 56 AM

### After
Screen Shot 2020-02-06 at 11 26 13 AM

## Checklist


test/sandbox/routes/v3/search/order.js
```diff
@@ -1,9 +1,19 @@
...
 var order = require('../../../fixtures/v3/search/order.json')
+var filteredOrders = require('../../../fixtures/v3/search/filter/order-filter')
+
+var NORTH_WEST_REGION = '824cd12a-6095-e211-a939-e4115bead28a'
 
 exports.order = function(req, res) {
   if (req.body.company === '375094ac-f79a-43e5-9c88-059a7caa17f0') {
     return res.json(order)
   }
+  if (
```
rafenden: Why not to use a simple condition?
```suggestion
  if (req.body.uk_region === '824cd12a-6095-e211-a939-e4115bead28a') {
```

peterhudec: The simple condition doesn't work if multiple regions are selected, which the test doesn't do anyway, but it could be quite confusing for anyone who is playing with it while it's connected to the sandbox, to see the results disappear when you add another region to `Nort West`.

peterhudec: So I prefer to keep it as it is.

rafenden: Good point, perhaps it's worth adding a comment.

PR #2416, opened 2020-02-06
Replace "subsidiarys" with "subsidiaries" on the company business details page

rafenden:
## Description of change

Fix incorrect content on the company business details page and replace `pluralise` NPM package with `pluralize`.


web-bert:
@rafenden is it worth adding a test here to check the output, I assume by changing the package used it has fixed the spelling mistake?

rafenden:
> @rafenden is it worth adding a test here to check the output, I assume by changing the package used it has fixed the spelling mistake?

@web-bert Yes, I used a different package which is more powerful, the tests were updated as requested.

PR #2417, opened 2020-02-07
Update flash messages to have a body

web-bert:
## Description of change

Add a new bit of middleware `req.flashWithBody` that takes `type`, `header` and `body` params. These params then get stringified into JSON and the normal `req.flash` middleware is called, appending `:with-body` to the `type`.

To handle displaying of these richer flash messages the `getMessages` method in the `user-locals.js` file has been updated to `JSON.parse` these special flash messages. This ensures that the `messages.njk` macro then has an object to pull the properties from, which has now been updated to check for body text and display it correctly.

## Test instructions

Add an interaction and answer "Were countries discussed" as "Yes", this will then show the new richer message on the interaction detail page upon successful creation of the interaction

### Before

Screenshot 2020-02-07 at 15 17 17

### After

Screenshot 2020-02-07 at 15 18 10

## Checklist


web-bert:
@paulgain this can be used for the flash message you added last week, have a look and see if it works as you would expect

PR #2420, opened 2020-02-10
Add full export history page (no pagination)

debitan:
## Description of change

This adds the full export history page. In this first iteration there is no pagination or axios calls. It simply fetches all the data and displays it in one go. Pagination will be in a later PR. 

This is also a hidden URL at the moment. I will create a seperate PR to add the link to this page to the export page. 

All additions and deletions are currently shown as individual events. There is a ticket for the backend to group these together to reflect when countries were added and deleted in a single user event. I will further iterate this front end feature to incorporate this when it is ready. 

## Test instructions

You should see the export countries history using the collectionList component when going to /companies/:companyId/exports/history

There is only history from when we merged the history PR last week - so you may have to add and delete some export countries to see the feature in action. 

## Screenshots

FireShot Capture 034 - Full export history - Exports - One List Subsidiary Ltd - Companies -_ - localhost

## Checklist


src/apps/companies/apps/exports/client/ExportsFullHistory.jsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import styled from 'styled-components'
+
+import { H3 } from '@govuk-react/heading'
+
+import { CollectionList } from 'data-hub-components'
+
+const Wrapper = styled('div')`
+  margin-top: 16px;
```
ian-leggett: We should use spacing units here - `import { SPACING } from '@govuk-react/constants'`

src/apps/companies/apps/exports/client/ExportsFullHistory.jsx
```diff
@@ -0,0 +1,30 @@
...
+`
+
+function ExportFullHistory({ data }) {
+  return (
+    <Wrapper>
+      <H3>Export markets history</H3>
```
ian-leggett: `<h2>` as mentioned above

src/apps/companies/apps/exports/client/ExportsFullHistory.jsx
```diff
@@ -0,0 +1,30 @@
...
+      <CollectionList
+        items={data.results}
```
ian-leggett: I think `data.results` should be populated via a saga and injected into the redux store.

src/apps/companies/apps/exports/controller.js
```diff
@@ -81,6 +85,22 @@ function renderExports(req, res) {
     })
 }
 
+async function renderFullExportHistory(req, res) {
+  const { token } = req.session
+  const { company } = res.locals
+  const data = await getFullExportHistory(token, company.id)
```
ian-leggett: This should be a Redux saga

src/apps/companies/apps/exports/client/ExportsFullHistory.jsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import styled from 'styled-components'
+
+import { H3 } from '@govuk-react/heading'
```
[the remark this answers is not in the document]
paulgain: All `node_module` imports should be grouped together.

src/apps/companies/apps/exports/repos.js
```diff
@@ -0,0 +1,14 @@
+const config = require('../../../../config')
+const { authorisedRequest } = require('../../../../lib/authorised-request')
+
+module.exports = {
+  getFullExportHistory: (token, companyId) => {
```
ian-leggett: We don't actually need this now, this should be a Redux saga which populates state in the Redux store

src/apps/companies/apps/exports/repos.js
```diff
@@ -0,0 +1,14 @@
...
+    return authorisedRequest(token, {
+      url: `${config.apiRoot}/v4/search/export-country-history`,
+      method: 'POST',
+      body: {
+        company: `${companyId}`,
```
paulgain: `companyId ` will suffice.

src/apps/companies/apps/exports/repos.js
debitan: even for this iteration that doesn't have any state? is this maybe a consideration for the next PR - https://github.com/uktrade/data-hub-frontend/pull/2428

src/apps/companies/apps/exports/transformer.js
```diff
@@ -25,51 +36,78 @@ function getCountriesFields(company) {
...
-      company.export_experience_category &&
-      company.export_experience_category.name,
-  }
+  return [country, historyTypeText, typeText].join(' ')
+}
 
-  const greatProfile = {
-    name: exportDetailsLabels.greatProfile,
-    value: company.great_profile_status,
+function createCountry(item) {
```
paulgain: You could destructure `item` here.

src/apps/companies/apps/exports/transformer.js
```diff
@@ -25,51 +36,78 @@ function getCountriesFields(company) {
...
+}
 
-  const exportPotential = {
-    name: exportDetailsLabels.exportPotential,
-    value:
-      exportPotentialLabels[company.export_potential] &&
-      exportPotentialLabels[company.export_potential].text,
-  }
+module.exports = {
+  transformFullExportHistory: (data) => ({
```
paulgain: You could destructure `data` here.

src/apps/companies/apps/exports/views/full-history.njk
```diff
@@ -0,0 +1,15 @@
...
+
+{% block body_main_content %}
+
+  <div class="govuk-grid-column-full">
+            {% component 'react-slot', {
+              id: 'company-export-full-history',
+              props: props
+            }
+            %}
+  </div>
```
paulgain: The content of this `div` needs aligning correctly.

test/functional/cypress/specs/companies/export-spec.js
```diff
@@ -202,8 +203,96 @@ describe('Companies Export', () => {
...
     })
   })
 })
+
+describe('Companies Export Countries', () => {
+  const countrySelectors = exportSelectors.countries
+
+  context('when there is no history', () => {
+    before(() => {
+      cy.visit(`/companies/${fixtures.company.lambdaPlc.id}/exports/history`)
```
paulgain: urls module

src/apps/companies/apps/exports/repos.js
ian-leggett: Fair enough lets tackle the Redux bits in the second PR

test/functional/cypress/specs/companies/export-spec.js
```diff
@@ -202,8 +203,96 @@ describe('Companies Export', () => {
...
+
+    it('renders the collection list with the 7 results', () => {
+      cy.contains('7 results').should('exist')
+      cy.get(countrySelectors.listItemHeadings).should('have.length', 7)
+
+      cy.contains('Argentina added to currently exporting').should('exist')
+      cy.contains('Argentina added to currently exporting')
+        .siblings()
+        .should('contain', 'By DIT Staff')
+        .should('contain', 'Date 6 Feb 2020, 4:06pm')
```
paulgain: L255 and L256 do the same thing, you can remove L255 as it duplicating the next line.

`Contains`: 
Get the DOM element containing the text
https://docs.cypress.io/api/commands/contains.html#Syntax

`should`:
Asserts that the target is neither null nor undefined (Chai: https://www.chaijs.com/api/assert/#method_exists)

The same applies to everything below.

test/functional/cypress/specs/companies/export-spec.js
```diff
@@ -202,8 +203,96 @@ describe('Companies Export', () => {
...
+        .should('contain', 'By DIT Staff')
+        .should('contain', 'Date 6 Feb 2020, 3:40pm')
+      cy.contains('Albania added to currently exporting').should('exist')
+      cy.contains('Albania added to currently exporting')
+        .siblings()
+        .should('contain', 'By DIT Staff')
+        .should('contain', 'Date 6 Feb 2020, 3:40pm')
+    })
+  })
+})
```
paulgain: I think we're over using contains here. The problem with contains is that it queries the entire document which is not performant (I don't have any stats to back this up). Personally, we are better off using contains sparingly, we could use it to initially to find a DOM element (even better, be specific with `get`) and then cycle though by chaining all the commands/asserts like @peterhudec does. This also ensures all DOM elements are in the correct order.
https://github.com/uktrade/data-hub-frontend/blob/master/test/functional/cypress/specs/companies/dnb-hierarchy-spec.js#L49-L87

src/apps/companies/apps/exports/client/ExportsFullHistory.jsx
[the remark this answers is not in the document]
debitan: group the imports. 

H3 is how it is in our design and it is also consistent with the headers within the tab navigation.

src/apps/companies/apps/exports/client/ExportsFullHistory.jsx
debitan: I'll use SCALE_3 which is 15px and so close enough to the original 16px I chose.

src/apps/companies/apps/exports/client/ExportsFullHistory.jsx
debitan: see comment above

test/functional/cypress/specs/companies/export-spec.js
debitan: We can talk about this more away from review, but I think as this is quite a dynamic page I would have to essentially reimplement the transformer function here to be able to cycle programmatically through and assert what should appear. This feels a bit too much like mocking. All I want to check here is that the text is appearing in the way expected for the user and that the outcome of the transformer is correct. The order of the items is dictated by the backend and is tested there. 
As for speed, these tests execute almost instantaneously, so I don't think we have much to worry about there. 
I don't believe this should be a blocker for this PR.

src/apps/companies/apps/exports/client/ExportsFullHistory.jsx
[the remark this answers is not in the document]
ian-leggett: The size of the header (look and feel) is fine I was just thinking accessibility. It looks fine but I would use `<H2 size={LEVEL_SIZE[3]}>` Ideally you can't have a H3 following a H1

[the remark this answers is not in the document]
debitan: Gotcha! Will fix that.

PR #2429, opened 2020-02-13
Add docker compose and env file for visual tests

PippoRaimondi:
## Description of change

The intent of this PR is add docker compose to run the visual tests and add env support to deal with browserstack.

## Test instructions

`$ ./.bin.sample/visual`

## Checklist


test/visual/.template.browserstack.env
```diff
@@ -0,0 +1,2 @@
+BROWSERSTACK_USERNAME=user
```
rafenden: Probably, it's a good idea to consider a unified approach with the way we run functional tests and sandbox.

@web-bert recently added a new `.bin.sample` directory.

In this case we could create a new file there, e.g. `.bin.sample/visual-test` and place the code there, e.g.
```
BROWSERSTACK_USERNAME=user /
BROWSERSTACK_ACCESS_KEY=key /
yarn run test:visual
```

web-bert: I agree with @rafenden as this could just be a PR to add a new sample script which we can all benefit from

PippoRaimondi: sorry, this was work in progress.. it should be done now... the idea is simplify the visual test run as everyone seems to struggle setting it up... so I created a compose file which should do the work for us... downside is that every time we modify code we need to rebuild, which takse 2-3minutes

PippoRaimondi: So for now we will continue with the compose file but simplifying run by adding a bin sample for visual tests

test/visual/Dockerfile
```diff
@@ -0,0 +1,23 @@
+# E2E docker image used to run tests locally
+FROM node:12.15.0
+
+ENV NPM_CONFIG_LOGLEVEL    warn
+ENV NPM_CONFIG_UNSAFE_PERM true
+
+RUN mkdir -p /project
```
mcbhenwood: Technically you don't need the `-p` here. In fact you don't even need the `mkdir` because `WORKDIR` creates the path if it doesn't already exist: https://meet.google.com/kmp-cnkq-hwm?hs=122

test/visual/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+RUN apt-get update
+
+# Install dockerize https://github.com/jwilder/dockerize
+ENV DOCKERIZE_VERSION v0.2.0
+RUN wget https://github.com/jwilder/dockerize/releases/download/$DOCKERIZE_VERSION/dockerize-linux-amd64-$DOCKERIZE_VERSION.tar.gz \
+    && tar -C /usr/local/bin -xzvf dockerize-linux-amd64-$DOCKERIZE_VERSION.tar.gz
+
+# Install visual test dependencies
+RUN apt-get install -y imagemagick
+RUN yarn install
```
mcbhenwood: It's probably better to do the yarn install much earlier on and after that to copy in the rest of the project. Reason being that you will better leverage the caching nature of Docker builds. Under my suggestion, it would avoid the need for the `yarn install` to run again if only the app code changes. With the current structure here, `yarn install` will run on every build of the Docker image, if anything at all changed.

So what I'm suggesting is:
```
RUN apt-get update
RUN apt-get install -y imagemagick
ENV DOCKERIZE_VERSION v0.2.0
RUN wget https://github.com/jwilder/dockerize/releases/download/$DOCKERIZE_VERSION/dockerize-linux-amd64-$DOCKERIZE_VERSION.tar.gz \
    && tar -C /usr/local/bin -xzvf dockerize-linux-amd64-$DOCKERIZE_VERSION.tar.gz

WORKDIR /project
COPY package.json yarn.lock /project
RUN yarn install

COPY . /project
CMD yarn test:visual
```
or something like that. Basically do the slowest-changing build steps first

mcbhenwood: Plus.... `install` is redundant as it's the default action for `yarn`

.bin.sample/visual
```diff
@@ -0,0 +1,3 @@
+export BROWSERSTACK_USERNAME=user
+export BROWSERSTACK_ACCESS_KEY=key
```
web-bert: I don't think the `export` keywords are required here

visual-regression.yml
```diff
@@ -0,0 +1,59 @@
...
+    entrypoint: dockerize -wait tcp://frontend:3000 -timeout 120s
+    command: yarn test:visual
+    depends_on:
+      - frontend
+
+  sandbox:
+    build:
+      context: ./test/sandbox
+    ports:
+      - "8001:8001"
```
web-bert: Do we need to expose this to the host? It will conflict if we already have sandbox running...

visual-regression.yml
```diff
@@ -0,0 +1,59 @@
...
+
+  redis:
+    image: redis:3.2
+    ports:
+      - 6379:6379
```
web-bert: Do we need to expose this to the host? It will conflict if we already have redis running...

visual-regression.yml
```diff
@@ -0,0 +1,59 @@
...
+
+  frontend:
+    build: .
+    ports:
+      - "3000:3000"
```
web-bert: This should probably map to a different port so we can run the tests with the FE already running locally?

.gitignore
```diff
@@ -52,6 +52,7 @@ cypress/**
 visual-screenshots/comparison
 visual-screenshots/diff
 visual-report
+.browserstack.env
```
web-bert: I don't think this is needed anymore

docs/Running tests.md
```diff
@@ -137,19 +137,16 @@ Screenshots will be stored in the root of the project. We commit the baselines a
 
 ### Browserstack environment variables
 
-to run in browserstack, ensure you have the following environment variables set:
+In order to run in browserstack, ensure you have the following variables in env file named `.browserstack.env`:
```
web-bert: This needs to change to give instructions to copy the `.bin.sample/visual` script into something like `.bin/visual` and update the env vars

docs/Running tests.md
```diff
@@ -137,19 +137,16 @@ Screenshots will be stored in the root of the project. We commit the baselines a
...
 export BROWSERSTACK_ACCESS_KEY=xxx
-export IS_REMOTE=true
 ```
 
 ### Running the tests
 
-After setting up the environment variables, run the following command to execute the tests:
-
-`$ yarn test:visual`
+`$ ./.bin.sample/visual`
```
web-bert: This should be whatever is used in the instructions above (`.bin/visual`)

visual-regression.yml
```diff
@@ -0,0 +1,54 @@
...
+      OAUTH2_BYPASS_SSO: "true"
+      OAUTH2_CLIENT_SECRET: youAintSeenMyRight
+      OAUTH2_CLIENT_ID: randomClientId
+      OAUTH2_DEV_TOKEN: ditStaffToken
+      CACHE_ASSETS: "true"
+      HELP_CENTRE_URL: https://datahub-helpcentre.london.cloudapps.digital
+      HELP_CENTRE_ANNOUNCMENTS_URL: https://datahub-helpcentre.london.cloudapps.digital/updates/announcements
+      HELP_CENTRE_API_FEED: http://sandbox:8001/help-centre/announcement
+      HELP_CENTRE_FEED_API_TOKEN: apiToken
+      SESSION_SECRET: foobarbaz
```
web-bert: I wonder if it is worth overriding the log level with `LOG_LEVEL=info`, as my default is set to debug so I see a lot of noise and I can't change it easily?

.bin.sample/visual
```diff
@@ -0,0 +1,3 @@
+BROWSERSTACK_USERNAME=user
+BROWSERSTACK_ACCESS_KEY=key
+docker-compose -f visual-regression.yml up
```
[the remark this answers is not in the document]
PippoRaimondi: agreed!

PR #2432, opened 2020-02-17
Referral details page

ian-leggett:
## Description of change

Trello -https://trello.com/c/owy0yoIf/1113-fe-referral-details-page-6

As part of the referral work this is the referral details page which is currently hidden behind a route that is not exposed anywhere in the app `/referral/{id}`. 

## Test instructions

You will need to post a referral to the back end. Method `POST` `/v4/company-referral/{id}`

For more info see the API docs - https://api.datahub.dev.uktrade.io/docs

Once you have a referral posted just access the referral via `/referral/{id}`

Note: The "Complete referral" button currently does not post anything to the back end, this will be done in another PR.

_What should I see?_

## Screenshots
Screenshot 2020-02-17 at 19 19 47

## Checklist


src/apps/referrals/client/referral-details.jsx
```diff
@@ -0,0 +1,108 @@
...
+import Link from '@govuk-react/link'
+import { SummaryTable, FormActions } from 'data-hub-components'
+
+import Task from '../../../client/components/Task'
+import { state2props } from '../state'
+
+import { REFERRAL_DETAILS } from '../../../client/actions'
+
+const AdviserDetails = ({ adviser }) => {
+  const { subject, name, email, team } = adviser
```
mforner13: @ian-leggett  I think you can remove subject here as it's not used in the AdviserDetails component :)

ian-leggett: Good spot just cleaned up this and some other unused variables and imports.

src/apps/referrals/client/referral-details.jsx
```diff
@@ -0,0 +1,116 @@
...
+    id,
+    company,
+    contact,
+    sendingAdviser,
+    receivingAdviser,
+    date,
+    notes,
+  }) => {
+    return (
+      <>
```
peterhudec: You don't need the fragment

src/apps/referrals/client/referral-details.jsx
```diff
@@ -0,0 +1,116 @@
...
+    subject,
+    id,
+    company,
+    contact,
+    sendingAdviser,
+    receivingAdviser,
+    date,
+    notes,
+  }) => {
+    return (
```
peterhudec: You are not definig any constants so you could just use implicit return

src/apps/referrals/reducer.js
```diff
@@ -0,0 +1,28 @@
...
+        contact,
+        sendingAdviser,
+        receivingAdviser,
+        date,
+        notes,
+      }
+    default:
+      return state
+  }
+}
```
peterhudec: As you only check for a single action type, you don't need a whole switch statement. Also, you can jus spread the result on top of the state, or actually replace it as the state doesn't hold anything else than all the props of result so you can do:
```js
export default (state = {}, { type, result }) =>
   type === REFERRAL_DETAILS
     ? result
     : state
```
It is actually safer to replace the state as opposed to extending it, because if the referal is loaded the second time and it for example doesn't have the `notes` property, the `notes` from the previously loaded referral would still be there.

src/apps/referrals/transformer.js
```diff
@@ -0,0 +1,33 @@
...
+      name: created_by.name,
+      email: get(created_by, 'contact_email', ''),
+      team: get(created_by.dit_team, 'name', ''),
+    },
+    receivingAdviser: {
+      name: recipient.name,
+      email: get(recipient, 'contact_email', ''),
+      team: get(recipient.dit_team, 'name', ''),
+    },
+    date: moment(created_on).format('DD MMMM YYYY'),
```
peterhudec: Formatting of any kind should be done in the presentation layer and you should use the [`data-hub-components/DateUtils`](https://github.com/uktrade/data-hub-components/blob/e760f46a567819ec1cef3834bab9e8e826424fb1/src/utils/DateUtils.js)

src/apps/referrals/transformer.js
```diff
@@ -0,0 +1,33 @@
...
+    contact: contact.name,
+    sendingAdviser: {
+      name: created_by.name,
+      email: get(created_by, 'contact_email', ''),
+      team: get(created_by.dit_team, 'name', ''),
+    },
+    receivingAdviser: {
+      name: recipient.name,
+      email: get(recipient, 'contact_email', ''),
+      team: get(recipient.dit_team, 'name', ''),
```
peterhudec: You don't need to fall back to empty strings. The implicit `undefined` is just fine.

src/apps/referrals/__tests__/transformer.test.js
```diff
@@ -0,0 +1,76 @@
...
+      },
+      receivingAdviser: {
+        name: 'Barry Oling',
+        email: 'barry@barry.com',
+        team: 'Aberdeen City Council',
+      },
+      date: '16 February 2020',
+      notes:
+        'Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.',
+    }
```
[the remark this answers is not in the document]
ian-leggett: I have removed two of the tests as I don't really need them, I now only test the output of the transformed results. `let` has been removed and used `const` instead.

test/sandbox/main.js
```diff
@@ -28,7 +28,18 @@ var v4Metadata = require('./routes/v4/metadata/index.js')
 var v4SearchCompany = require('./routes/v4/search/company.js')
 var v4SearchCompanyWithCountry = require('./routes/v4/search/company/autocomplete.js')
 var v4SearchLargeInvestorProfiles = require('./routes/v4/search/large-investor-profile/results.js')
+<<<<<<< HEAD
```
PippoRaimondi: You need to fix this conflict I guess =D

src/apps/referrals/tasks.js
```diff
@@ -0,0 +1,10 @@
+const axios = require('axios')
```
rafenden: I think we should keep client-side code in the `apps/referrals/client` dir, this applies to `tasks.js`, `reducer.js` and `state.js`.

test/selectors/referrals/referral-details.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
```
rafenden: Optional: We should stop using selectors and rely on Cypress chain.

src/apps/referrals/client/referral-details.jsx
```diff
@@ -0,0 +1,81 @@
...
+            <Details summary="Why can I not edit the referral?">
+              <p>
+                This referral has been placed in the "My referrals" section on
+                the Homepage of both the recipient and sender. If necessary
+                contact the receiving adviser directly if any of the information
+                has changed.
+              </p>
+            </Details>
+            <FormActions>
+              <Button as={Link} href="/">
```
rafenden: We should use `a` instead of `Link` as this will not inherit styles from the `Link` component and will not change the font colours when the link is visited.
```suggestion
              <Button as="a" href="/">
```

test/sandbox/main.js
ian-leggett: All fixed, my mistake

src/apps/referrals/client/referral-details.jsx
```diff
@@ -0,0 +1,81 @@
+import React from 'react'
```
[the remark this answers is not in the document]
ian-leggett: All done

src/apps/referrals/tasks.js
ian-leggett: Yeah I like this makes sense going forward

test/selectors/referrals/referral-details.js
ian-leggett: Will look at removing these in the next iteration

PR #2444, opened 2020-02-21
Better task error handling & apiTaskFactory

peterhudec:
## Description of change

Improved the way `Task` component handles errors. Because of the nature of error handling in promisses, JavaScript errors were caught in the underlying saga, leaving no stack trace in the console and being displayed in the UI.

Now only promises that reject with ~string~ an instance of Error are treated like failed state, all other rejections will be rethrown. So if you want the task to end up in a UI error message, the promise must reject with a string which will be used as the error message in the UI:

```js
// These won't be caught and will log a stack trace in the console
const errorTask1 = () => undefined.foo
const errorTask2 = () => {throw Error()}
const errorTask3 = () => Promise.reject(Error())

// These will be caught in the saga and "Error message" will be rendered in the UI
const failingTask1 = () => {throw 'Error message'}
const failingTask2 = () => Promise.reject('Error message')
const failingTask2 = () => Promise.reject(123)
const failingTask2 = () => Promise.reject({toString: () => 'Error message'})
```
Screen Shot 2020-02-21 at 12 25 46 PM

Axios requests should thus transform the axios error with:
```js
const axiosTask = () => axios.get('/api-endpoint/v4/company-lists')
  .catch(e => Promise.reject(e.response.data.detail))
```

NOTE: The `apiTaskFactory` mentioned in some of the comments was removed from this PR. Instead there is a custom `apiProxyAxios` Axios instance added which has the error handling built in.

### DIFF
--- a/src/client/components/Task/utils.js
+++ b/src/client/components/Task/utils.js
@@ -1,4 +1,5 @@
-import { curry } from 'lodash'
+import axios from 'axios'
+import { curry, identity } from 'lodash'
 
 /**
  * A curried function which Decorates a task so, that it's progress will take
@@ -16,4 +17,23 @@ export const delay = curry((duration, task, payload) =>
     task(payload),
     new Promise((resolve) => setTimeout(resolve, duration)),
   ]).then(([result]) => result)
+)
+
+export const catchApiError = ({ response: { data } }) =>
+  Promise.reject(data.detail || data)
+
+/**
+ * A custom Axios instance for easy access to the `/api-proxy` endpoint.
+ * It has its {baseURL} set to `/api-proxy/` so you should omit that part of the
+ * path from the {url} of your request. It also handles API errors so that
+ * if used in a task of the {Task} component, the correct error message will be
+ * displayed in the {Task.Status}.
+ */
+export const apiProxyAxios = axios.create()
+apiProxyAxios.interceptors.request.use(({ url, ...config }) => ({
+  ...config,
+  url: url.startsWith('/api-proxy') ? url : url.replace(/^\/?/, '/api-proxy/'),
+}))
+apiProxyAxios.interceptors.response.use(identity, ({ response }) =>
+  Promise.reject(response.data.detail || response.statusText)
 )
--- a/src/apps/companies/apps/add-company/client/tasks.js
+++ b/src/apps/companies/apps/add-company/client/tasks.js
@@ -1,9 +1,8 @@
 import axios from 'axios'
-
-const handleError = (e) => Promise.reject(Error(e.response.data.detail))
+import { catchApiError } from '../../../../../client/components/Task/utils'
 
 export default (postcode) =>
   axios
     .get(`/api/postcode-to-region-lookup/${postcode}`)
-    .catch(handleError)
+    .catch(catchApiError)
     .then(({ data }) => data)
--- a/src/apps/companies/apps/edit-one-list/client/tasks.js
+++ b/src/apps/companies/apps/edit-one-list/client/tasks.js
@@ -1,5 +1,4 @@
-import axios from 'axios'
-
+import { apiProxyAxios } from '../../../../../client/components/Task/utils'
 import { SAVED, API_ERROR } from './state'
 import {
   NONE,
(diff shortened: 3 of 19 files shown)


src/client/components/Task/utils.js
```diff
@@ -17,3 +18,28 @@ export const delay = curry((duration, task, payload) =>
...
+ */
+export const apiTaskFactory = (path, transformer = (x) => x, axiosOptions) => (
+  payload
+) =>
+  axios({
+    ...(typeof axiosOptions === 'function'
+      ? axiosOptions(payload)
+      : axiosOptions),
+    url:
+      '/api-proxy/' +
```
web-bert: I'd rather just pass in the full string here, requiring the url to drop the initial slash in counter intuitive and makes reading the code using this harder to reason about

src/client/components/Task/utils.js
```diff
@@ -17,3 +18,28 @@ export const delay = curry((duration, task, payload) =>
...
+/**
+ * Creates a task for requests to the `/api-proxy` endpoint
+ * @param {string | payload => string} path - The API endpoint path e.g.
+ * `'v4/company-list'`. If the path is dependent on the payload, you can pass
+ * a function which will receive the payload and shuld return the path.
+ * @param {(result, payload) => any} [transformer=identity] - Use this to
+ * transform the data received from the API.
+ * @param {Object} [axiosOptions=] - Axios options
+ */
+export const apiTaskFactory = (path, transformer = (x) => x, axiosOptions) => (
```
web-bert: I wonder whether this should follow the axios interface and return an object with `post` and `get` and use the returned promise for the transformer - it would make this simpler to implement and keep code somewhat consistent?

src/client/components/Task/utils.js
```diff
@@ -17,3 +18,28 @@ export const delay = curry((duration, task, payload) =>
...
   ]).then(([result]) => result)
 )
+
+export const catchApiError = (e) => Promise.reject(e.response.data.detail)
+
+/**
+ * Creates a task for requests to the `/api-proxy` endpoint
+ * @param {string | payload => string} path - The API endpoint path e.g.
+ * `'v4/company-list'`. If the path is dependent on the payload, you can pass
+ * a function which will receive the payload and shuld return the path.
```
debitan: 'should'

src/client/components/Task/utils.js
```diff
@@ -17,3 +18,28 @@ export const delay = curry((duration, task, payload) =>
...
+ * @param {(result, payload) => any} [transformer=identity] - Use this to
+ * transform the data received from the API.
+ * @param {Object} [axiosOptions=] - Axios options
+ */
```
debitan: I would put an empty line between the comment and the function, as it is at the top of the comment block.

src/client/components/Task/utils.js
debitan: If it is kept the same I would move the transformer argument after the `axiosOptions` as that makes more chronological sense to me, i.e. you provide the path, you say how you're going to get the data, now you've got it - decide how you want to transform it.

rafenden: It's also a good practise to place the optional arguments (`transformer` in this case) at the end.

src/apps/companies/apps/exports/tasks.js
```diff
@@ -38,14 +39,15 @@ const transformFullExportHistory = ({ count, results }) => ({
...
-
-export const fetchExportsHistory = ({ companyId, activePage }) =>
-  axios
-    .post('/api-proxy/v4/search/export-country-history', {
+export const fetchExportsHistory = apiTaskFactory(
+  'v4/search/export-country-history',
+  transformFullExportHistory,
+  ({ companyId, activePage }) => ({
+    method: 'post',
+    data: {
```
rafenden: I would be easier to understand (and less magical)  if `fetchExportsHistory` was a simple function.
```suggestion
export const fetchExportsHistory = ({ companyId, activePage }) => apiTaskFactory(
  'v4/search/export-country-history',
  {
    method: 'post',
    data: {
      company: companyId,
      limit: 10,
      offset: activePage * 10 - 10,
    },
  },
  transformFullExportHistory,
)
```

src/client/components/Task/utils.js
rafenden: Agree

debitan:
hey @peterhudec is this still alive?

PR #2447, opened 2020-02-24
Create validated input component with 'multiInstance' util for reducer

mforner13:
## Description of change

The multiinstance util makes updating the state of similar components much simpler. When two components of the same type are used, multiinstance creates an id for each instance of that component and adds their values to state under that ID. 

The multiinstance util is used by the validated input component, of which we envisage there often being multiple instances.

The validated input component wraps around the gov uk InputField and adds validation to it. The component takes a 'validator' function which takes a value and should return a string describing the  error if the value if invalid. The error is displayed on the component using the structure of the gov.uk inputField: meta={touched: true, error: 'string'}.

## Test instructions

No visual changes yet - both the multiinstance util and the validated input component will be used in an upcoming PR for the create referral form.

## Checklist


src/client/utils/multiinstance.js
```diff
@@ -0,0 +1,76 @@
...
+  dispatch({
+    ...action,
+    id,
+  })
+
+export const connect = (
+  componentState2props = identity,
+  dispatch2props,
+  ...rest
+) => (component) => (name) => {
```
web-bert: Why not return a single function here that takes `component, name` ?

peterhudec: I wanted to follow the way they do it in `react-redux.connect`. I'll just wrap it in `_.curry` and it will work both ways.

src/client/components/ValidatedInput.jsx
```diff
@@ -0,0 +1,43 @@
...
+ * @description wraps the govuk-react InputField component, accepts the same
+ * props and adds two more.
+ * @param {Object} props - Props
+ * @param {string} props.id - Id under which the state of the component is
+ * stored.validatedInput.
+ * @param {(value: string) => falsy | string} props.validator
+ * - A function which takes a value and should return a string describing the
+ * error if the value if invalid.
+ */
+export default multiinstance({
```
rafenden: I truly believe that we should reuse the existing form elements from the component library. Recreating all the form components is a big waste.

src/client/components/ValidatedInput.jsx
```diff
@@ -0,0 +1,43 @@
+import React from 'react'
+
+import { InputField } from 'govuk-react'
+import multiinstance from '../utils/multiinstance'
```
web-bert: shouldn't `multiinstance` be `multiInstance` ?

src/client/components/ValidatedInput.jsx
mforner13: So this was originally part of my PR for the create referral form. We separated it out to reduce the PR, and as it was decided to not use the DH components for the form, the multiInstance util will be useful in the form. I know that's not as clear without the context of the form though.

debitan: we discussed this as a three. this won't block this PR, but as Raf is working to lift state out of the forms component this should be considered for future work.

peterhudec: If @rafenden ever succeeds in lifting the forms to use redux, we then can get rid of this component, but we need this now.

rafenden: In the future we should discuss these things BEFORE the PR stage.

peterhudec: We actually did discuss it as early as we felt a need for such a component

PR #2453, opened 2020-02-26
Add query string to the url after a company is linked for use in GA

rafenden:
## Description of change

In order to track number of linked companies, we need to add a query string to the page after the form is submitted.


web-bert:
There must be a better way than this, for one we could implement the data layer for analytics. Also this will not give an accurate number because of page refreshes, browser history or sharing links etc

I think we need to discuss our solution as a dev team, we should use something like the flash messages so that this only gets fired as an event once.

paulgain:
The whole GTM/GA needs to be addressed properly. My experience here is not good as we're not allowed to touch GTM and GA will only take us so far. We are not in a position to do what we want, especially creating a data layer. We are working under constraints and to some extent our hands are tide.

web-bert:
We are totally in control of adding a Data Layer and it would be as simple as adding:
```
<script>
  dataLayer = [{
    'pageCategory': 'signup',
    'visitorType': 'high-value'
  }];
</script>
```
It would be up to the analysts to define what we put in there and for it to actually work, I would explore this option first as I know they are keen for us to start using the Data Layer

rafenden:
@web-bert 

> There must be a better way than this, for one we could implement the data layer for analytics. Also this will not give an accurate number because of page refreshes, browser history or sharing links etc

For now I added ID to the status message, in the future we should use Redux middleware.

PR #2454, opened 2020-02-26
Hotfix: defend against update items in export history

debitan:
## Description of change

This is a hotfix to filter out export history items that have a `history_type` of `update`. These are not meant to appear in the api and there is a ticket to fix that root problem. 

I can't fix the count number, however. That will still show the total number of items across all pages, including the updates. As I only get a maximum of ten results back at a time I can't accurately work out how many to reduce the count by. This will only be rectified when the backend fix is put in place. 

## Test instructions

You should no longer see any export history items that do not say either 'added to' or 'removed from'. 

Previously `update` items were leaking through to the frontend would appear as, for example, `Andorra currently exporting to`, i.e. without one of the verbs above. 

## Checklist

### DIFF
--- a/src/apps/companies/apps/exports/tasks.js
+++ b/src/apps/companies/apps/exports/tasks.js
@@ -35,7 +35,9 @@ const createCountry = (item) => ({
 
 const transformFullExportHistory = ({ count, results }) => ({
   count,
-  results: results.map(createCountry),
+  results: results
+    .filter((result) => result.history_type === 'insert' || 'delete')
+    .map(createCountry),
 })
 
 const handleError = (e) => Promise.reject(Error(e.response.data.detail))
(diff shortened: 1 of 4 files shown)


src/apps/companies/apps/exports/tasks.js
```diff
@@ -35,7 +35,9 @@ const createCountry = (item) => ({
 
 const transformFullExportHistory = ({ count, results }) => ({
   count,
-  results: results.map(createCountry),
+  results: results
+    .filter((result) => result.history_type === 'insert' || 'delete')
```
rafenden: In general, it's a good practise to avoid magic strings, this could be refactored into:
```suggestion
    .filter((result) => WHITELISTED_HISTORY_TYPES.includes(result.history_type)
```

Also, the above will not work as expected, here's the working example:
```suggestion
    .filter((result) => result.history_type === 'insert' || result.history_type ===  'delete')
```

PR #2455, opened 2020-02-26
Hotfix to export history count

debitan:
This is a follow up PR to https://github.com/uktrade/data-hub-frontend/pull/2454

I didn't think I could fix the count while waiting for the backend fix, but there is this slightly convoluted way to do it. 

This code is temporary and will be removed when the backend fix is ready.

### DIFF
--- a/src/apps/companies/apps/exports/tasks.js
+++ b/src/apps/companies/apps/exports/tasks.js
@@ -35,21 +35,27 @@ const createCountry = (item) => ({
   ],
 })
 
-const transformFullExportHistory = ({ count, results }) => ({
-  count,
-  results: results
-    .filter((result) => WHITELISTED_HISTORY_TYPES.includes(result.history_type))
-    .map(createCountry),
-})
+const transformFullExportHistory = ({ results }, offset) => {
+  const cleanedResults = results.filter((result) =>
+    WHITELISTED_HISTORY_TYPES.includes(result.history_type)
+  )
+  const count = cleanedResults.length + offset
+
+  return {
+    count,
+    results: cleanedResults.slice(0, 10).map(createCountry),
+  }
+}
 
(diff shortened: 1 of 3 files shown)


src/apps/companies/apps/exports/tasks.js
```diff
@@ -35,21 +35,27 @@ const createCountry = (item) => ({
...
+    results: cleanedResults.slice(0, 10).map(createCountry),
+  }
+}
 
 const handleError = (e) => Promise.reject(Error(e.response.data.detail))
 
-export const fetchExportsHistory = ({ companyId, activePage }) =>
-  axios
+export const fetchExportsHistory = ({ companyId, activePage }) => {
+  const offset = activePage * 10 - 10
```
adamledwards: Does offset always start from 1? 
If it is 0 is there a possibility you could send -10 to the server

src/apps/companies/apps/exports/tasks.js
```diff
@@ -35,21 +35,27 @@ const createCountry = (item) => ({
...
+  return axios
     .post('/api-proxy/v4/search/export-country-history', {
       company: companyId,
-      limit: 10,
-      offset: activePage * 10 - 10,
+      offset,
```
adamledwards: With no limit is there a chance of duplicate data?
Sorry I'm not familiar with the whole implementation on data hub yet.

src/apps/companies/apps/exports/tasks.js
debitan: it always starts at zero

src/apps/companies/apps/exports/tasks.js
debitan: yeah, so will be throwing away any data beyond 10 items. It is inefficient, but the only way to calculate the correct total number of items while this bug exists.

PR #2460, opened 2020-02-28
Add the TabNav component

peterhudec:
## Description of change

Added the `TabNav` component which is a stateful wrapper of [`govuk-react/tabs`](https://govuk-react.github.io/govuk-react/?path=/story/tabs--component-default).

## Test instructions

There's nothing to see as the component is not used anywhere yet. It will be covered with functional tests once it is used on the homepage to switch between _my companies lists_ and _my referrals_.


src/client/components/TabNav/index.jsx
```diff
@@ -0,0 +1,71 @@
...
+            key={idx}
+            onClick={() => onChange(idx)}
+            selected={idx === selectedIdx}
+          >
+            {label}
+          </Tabs.Tab>
+        ))}
+      </StyledTabsList>
+    </Tabs>
+    {get(tabs, [selectedIdx, 'content'])}
```
debitan: You can use `?.` here to avoid using lodash and make it, IMO, easier to read.

src/client/components/TabNav/reducer.js
```diff
@@ -0,0 +1,4 @@
+import { TAB_NAV__SELECT } from '../../actions'
+
+export default (state = {}, { type, ...action }) =>
+  type === 'TAB_NAV__SELECT' ? action : state
```
rafenden: Please use the imported constant here.

src/client/components/TabNav/index.jsx
```diff
@@ -0,0 +1,71 @@
...
+import reducer from './reducer'
+
+const StyledTabsList = styled(Tabs.List)({
+  a: {
+    cursor: 'pointer',
+  },
+})
+
+const TabNav = ({
+  defaultSelected,
```
rafenden: What's the point of having both `defaultSelected` and `selectedIdx`?

src/client/components/TabNav/index.jsx
```diff
@@ -0,0 +1,71 @@
...
+      label: PropTypes.oneOfType([PropTypes.string, PropTypes.element])
+        .isRequired,
+      content: PropTypes.oneOfType([
+        PropTypes.string,
+        PropTypes.element,
+        PropTypes.number,
+      ]),
+    })
+  ),
+  defaultSelected: PropTypes.number,
```
rafenden: It's missing a default value, e.g. `0`

src/client/components/TabNav/index.jsx
peterhudec: You right, `selectedIdx` is enough to pre-set the selected tab from the outside.

src/client/components/TabNav/index.jsx
peterhudec: `defaultSelected` will go as to your previous suggestion, so this one is irrelevant

rafenden: `selectedIdx` is missing in propTypes, also I think it should have a default value of `0`.

peterhudec: Yeah, it needs to be added to `propTypes`, but default value `0` would set the first tab as active. The default value is just `undefined`.

src/client/components/TabNav/index.jsx
```diff
@@ -0,0 +1,72 @@
...
+import reducer from './reducer'
+
+const StyledTabsList = styled(Tabs.List)({
+  a: {
+    cursor: 'pointer',
+  },
+})
+
+const TabNav = ({
+  selectedIdx,
```
web-bert: Remove shorthand here, it's not in keeping with the rest of the variable names: `selectedIdx => selectedIndex`

src/client/components/TabNav/index.jsx
```diff
@@ -0,0 +1,72 @@
...
+  // We are only destructuring dispatch from the props so that <div> won't
+  // complain about it being an unexpected prop.
+  dispatch,
+  ...props
+}) => (
+  <div {...props}>
+    <Tabs>
+      <Tabs.Title>{title}</Tabs.Title>
+      <StyledTabsList>
+        {tabs.map(({ label }, idx) => (
```
web-bert: Remove shorthand here too: `idx => index`

PR #2462, opened 2020-02-28
ReferralList component

peterhudec:
## Description of change

Adds the `ReferralList` component, which lists all referrals related to the logged in user.

## Test instructions

The component is not used anywhere yet, so the only way to see it is to check out this branch and put the `ReferallList` component somewhere.

## Screenshots
Screen Shot 2020-02-28 at 2 08 44 PM

## Checklist


src/client/components/ReferralList/Referral.jsx
```diff
@@ -0,0 +1,92 @@
+import { capitalize } from 'lodash'
+import React from 'react'
+import { H3, Link } from 'govuk-react'
+import { SPACING } from '@govuk-react/constants'
+import * as colours from 'govuk-colours'
```
debitan: I think it would be nicer to explicitly state which colours you're getting here

src/client/components/ReferralList/Referral.jsx
```diff
@@ -0,0 +1,92 @@
...
+      </StyledH3>
+      <div>{DateUtils.format(date)}</div>
+    </StyledRow>
+    <StyledRow>
+      {/* We need to wrap SummaryList as it can't be restyled */}
+      <StyledSummaryListWrapper>
+        <SummaryList
+          rows={[
+            {
+              label: 'Sending adviser(s)',
```
debitan: Could we not infer whether to display the plural 's' depending on the number of items in the value?

You have done a similar thing in the index file:

`pluralize('Referral', referrals.length, true)}`

src/client/components/ReferralList/Referral.jsx
```diff
@@ -0,0 +1,92 @@
...
+              value: <AdviserDetails {...sender} />,
+            },
+            {
+              label: 'Receiving adviser(s)',
```
debitan: Same here

src/client/components/ReferralList/Referral.jsx
peterhudec: I had it that way before and it was a developer's nightmare experience. I'ts way more practical this way.

src/client/components/ReferralList/Referral.jsx
peterhudec: This is how it was specified in the [mockup](https://trello-attachments.s3.amazonaws.com/5adf5c41e7427d44fb726c25/5e286e0394a9a225961e4fea/7c0b7dfed13aac86d92b6b4b6ea1573c/Screenshot_2020-02-25_at_13.20.31.png).

src/client/components/ReferralList/Referral.jsx
peterhudec: See my previous reply

src/client/components/ReferralList/Referral.jsx
paulgain: @peterhudec I agree with @debitan. I don't see how there is a "developer's nightmare experience". Prefixing all colours with `colours.` is verbose and not needed it's also inconsistent with how we require `govuk-colours` throughout Data Hub. This smells very similar to how you wanted to import `lodash` using `_` instead of the functions.

src/client/components/ReferralList/task.js
```diff
@@ -0,0 +1,35 @@
+import axios from 'axios'
+
+const handleError = (e) => Promise.reject(Error(e.response.data.detail))
```
paulgain: Is `response` and `data` always guaranteed to be defined?

src/client/components/ReferralList/Referral.jsx
peterhudec: With the whole `colours` imported code completion would suggest possible colors where I wanted to use them. With direct import, I need to know exactly which variant of blue I need when importing it. Of the two approaches I found the direct import extremely annoying, but I appreciate for you it might be the other way. Anyway, I can still write my programs with the whole namespace imported and then once everything is done, refactor the import to make you guys happy.

src/client/components/ReferralList/task.js
peterhudec: Yes, `response.data` is part of Axios API and `detail` is part of `data-hub-api` error response. So if they are not defined, this will raise an error, which is what we want.

src/client/components/ReferralList/Referral.jsx
peterhudec: BTW how aobut importing the whole `SPACING` object? We don't seem to be very consistent there.

src/client/components/ReferralList/task.js
```diff
@@ -0,0 +1,35 @@
...
+        }) => ({
+          id,
+          status,
+          subject,
+          companyName: company.name,
+          date: created_on,
+          sender: convertAdviser(created_by),
+          recipient: convertAdviser(recipient),
+        })
+      )
```
web-bert: Wouldn't this be clearer as:
```javascript
results.map(
   (referral) => ({
      id: referral.id,
      status: referral.status,
      subject: referral.subject,
      companyName: referral.company.name,
      date: referral.created_on,
      sender: convertAdviser(referral.created_by),
      recipient: convertAdviser(referral.recipient),
   })
)
```

src/client/components/ReferralList/Referral.jsx
```diff
@@ -0,0 +1,92 @@
...
+  padding: SPACING.SCALE_2,
+})
+const StyledLink = styled(Link)({
+  textDecoration: 'none',
+})
+const StyledH3 = styled(H3)({
+  margin: 0,
+})
+
+const Referral = ({
```
rafenden: Please reuse the existing activity card component as it looks exactly the same:
image

rafenden:
@peterhudec would be nice to see some tests.

src/client/components/ReferralList/Referral.jsx
debitan: I guessed that was the case, but maybe it wasn't realised that we can dynamically change the pluralisation.

peterhudec:
> @peterhudec would be nice to see some tests.

@rafenden tests will come in subsequent PR where `ReferralList` and `TabNav` will be integrated into dashboard.

src/client/components/ReferralList/Referral.jsx
```diff
@@ -0,0 +1,79 @@
...
+const STATUS_COLOURS = {
+  outstanding: BLUE,
+  completed: GREEN,
+}
+
+const StyledSummaryListWrapper = styled.div({
+  flexGrow: 1,
+  marginRight: SPACING.SCALE_2,
+  dt: {
+    // Prevent he left column from shrinking,
```
mforner13: Prevent *the

peterhudec:
Add ReferralList component

PR #2418, opened 2020-02-08
Send a referral form

mforner13:
## Description of change

This is a form for a user to fill in the details of a referral, including who they're referring the company to, a subject, notes and a suggested contact at that company. 

The values of the fields are added to the redux store so that they can be carried between the form and the confirmation component. The confirmation component will come in a future PR, so for now a simple div is rendered with a back button.

An explanation of the architecture using redux for the form can be found in [this document](https://docs.google.com/document/d/1LSqHOLdCrWbZGFHKobvc7-Ny81iRwDnKEpfB_OcMnzc/edit).

The form will be behind behind a url which isn't linked to anywhere in Data Hub. So users will not be able to access the form once the PR is merged.

TODO
- Add 'Send referral' button which points to the form to company page next to 'Add interaction' button
- Add in error handling for the notes field being required. This is in a separate ticket so as not to slow down release of the form.

## Test instructions

Go to companies/:companyId/send-referral.

### After

ReferralFormScreenshot

## Checklist

[//]: # "When submitting a PR make sure the code review guidelines have been satisfied.
https://github.com/uktrade/data-hub-frontend/blob/master/docs/Code%20review%20guidelines.md"

- [x] Has the branch been rebased to master?
- [x] Automated tests (Any of the following when applicable: Unit, Functional or Acceptance)
- [ ] Manual compatibility testing (Browsers: Chrome, Firefox, IE11, Safari)

PR #2472, opened 2020-03-04
Confirm referral details and send referral

mforner13:
## Description of change
Following a PR which captures referral details in a form (#2418), this change adds a confirmation component which displays the user's input data. They can then go back and edit the referral in the form, or click 'Send referral' which will post that referral to the API. 

They will then be redirected back to the company page they started from, and a flash message with a link to see 'My referrals' on the homepage appears. The 'My referrals' tab will be added to the homepage in an upcoming PR.
 
The confirmation page is a react component, rendered when the 'confirm' prop is true. This is toggled with an onSubmit action on the form component. 

The API post request is made using the traditional express router, rather than the task component. This is because a redirect after the request promise is fulfilled is not yet possible using the task component.

## Test instructions

- Go to companies/:companyId/send-referral
- Enter an adviser and a subject (currently only required fields)
- Click continue to see the confirmation page

### After

Screenshot 2020-03-04 at 12 16 16

Screenshot 2020-03-04 at 12 17 22

## Checklist

### DIFF
--- a/src/apps/referrals/apps/send-referral/client/SendReferralConfirmation.jsx
+++ b/src/apps/referrals/apps/send-referral/client/SendReferralConfirmation.jsx
@@ -1,0 +1,74 @@
+import React from 'react'
+import styled from 'styled-components'
+import PropTypes from 'prop-types'
+
+import { SummaryTable, FormActions } from 'data-hub-components'
+import { H4, Button, Link } from 'govuk-react'
+import SecondaryButton from '../../../../../client/components/SecondaryButton'
+
+const StyledParagraph = styled('p')`
+  font-size: 16px;
+`
+
+const SendReferralConfirmation = ({
+  companyName,
+  companyId,
+  adviser,
+  subject,
+  contact,
+  notes,
+  cancelUrl,
+  onBack,
+  csrfToken,
+}) => {
+  return (
+    <>
+      <SummaryTable caption="Check referral details">
+        <SummaryTable.Row heading="Send this company record to">
+          {adviser.name}
+        </SummaryTable.Row>
+        <SummaryTable.Row heading="Subject">{subject}</SummaryTable.Row>
+        <SummaryTable.Row heading="Notes">
+          {notes || 'No notes added'}
+        </SummaryTable.Row>
+        <SummaryTable.Row heading="Company contact">
+          {contact?.name || 'No contact added'}
+        </SummaryTable.Row>
+      </SummaryTable>
+      <SecondaryButton onClick={onBack}>Edit referral</SecondaryButton>
+      <H4>What happens next</H4>
+      <StyledParagraph>
+        <p>
+          <p>
+            Clicking “Send referral” will show the referral in the activity of{' '}
+            {companyName}, as well as in the Referrals section on both your Data
+            Hub Homepage and the Homepage of the recipient.
+          </p>
+          <p>It might take up to 24 hours for the referral to appear.</p>
+          <p>You will not be able to edit the referral after this point.</p>
+        </p>
+      </StyledParagraph>
+      <form method="post">
+        <input name="recipient" value={adviser.id} type="hidden"></input>
+        <input name="subject" value={subject} type="hidden"></input>
+        <input name="notes" value={notes} type="hidden"></input>
+        <input name="contact" value={contact?.id} type="hidden"></input>
+        <input name="company" value={companyId} type="hidden"></input>
+        <input name="_csrf" value={csrfToken} type="hidden"></input>
+        <FormActions>
+          <Button>Send referral</Button>
+          <Link href={cancelUrl}>Cancel</Link>
+        </FormActions>
+      </form>
+    </>
+  )
+}
+
+SendReferralConfirmation.propTypes = {
+  companyId: PropTypes.string.isRequired,
+  companyName: PropTypes.string.isRequired,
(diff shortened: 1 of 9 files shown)


src/apps/referrals/apps/send-referral/client/SendReferralConfirmation.jsx
```diff
@@ -0,0 +1,76 @@
...
+  adviser,
+  subject,
+  contact,
+  notes,
+  cancelUrl,
+  onBack,
+  csrfToken,
+}) => {
+  return (
+    <div>
```
rafenden: No need for div here, fragment should be sufficient, i.e. `<>`.

src/apps/referrals/apps/send-referral/client/SendReferralConfirmation.jsx
```diff
@@ -0,0 +1,76 @@
...
+        <SummaryTable.Row heading="Subject">{subject}</SummaryTable.Row>
+        <SummaryTable.Row heading="Notes">
+          {notes || 'No notes added'}
+        </SummaryTable.Row>
+        <SummaryTable.Row heading="Company contact">
+          {contact?.name || 'No contact added'}
+        </SummaryTable.Row>
+      </SummaryTable>
+      <SecondaryButton onClick={onBack}>Edit referral</SecondaryButton>
+      <br></br>
```
rafenden: `<br>` shouldn't be used to create margins between elements:

> Note: Do not use `<br>` to create margins between paragraphs; wrap them in `<p>` elements and use the CSS margin property to control their size. — https://developer.mozilla.org/en-US/docs/Web/HTML/Element/br

src/client/index.jsx
```diff
@@ -26,7 +26,7 @@ import ExportsFullHistory from '../apps/companies/apps/exports/client/ExportsFul
 import ReferralDetails from '../apps/referrals/apps/details/client/ReferralDetails'
 import ReferralHelp from '../apps/referrals/apps/help/client/ReferralHelp'
 import ValidatedInput from './components/ValidatedInput'
-import SendReferralForm from '../apps/referrals/apps/send-referral/client/SendReferralForm'
+import SendReferral from '../apps/referrals/apps/send-referral/client/SendReferralForm'
```
rafenden: Optional: I suggest keeping `SendReferralForm` to reflect the filename.

test/functional/cypress/specs/companies/send-referral-spec.js
```diff
@@ -152,55 +152,137 @@ describe('Send a referral form', () => {
...
-    )
   })
+
+  context(
+    'when "Continue" button is clicked after all fields filled in',
+    () => {
+      it('should display the confirmation component with the values just input', () => {
+        cy.visit(urls.companies.sendReferral(fixtures.company.withContacts.id))
+        selectTypeahead('Adviser', 'S', 3)
+        cy.get(selectors.companySendReferral.subjectField)
```
rafenden: Optional: We should avoid using selectors as they add an extra overhead.

src/apps/referrals/apps/send-referral/client/SendReferralConfirmation.jsx
mforner13: Ahh it felt wrong when I was doing it as well! Have fixed with <p>s :)

src/apps/referrals/apps/send-referral/client/SendReferralConfirmation.jsx
```diff
@@ -0,0 +1,74 @@
...
+          {contact?.name || 'No contact added'}
+        </SummaryTable.Row>
+      </SummaryTable>
+      <SecondaryButton onClick={onBack}>Edit referral</SecondaryButton>
+      <H4>What happens next</H4>
+      <StyledParagraph>
+        <p>
+          <p>
+            Clicking “Send referral” will show the referral in the activity of{' '}
+            {` ${companyName}`}, as well as in the Referrals section on both
```
rafenden: I think we will have two empty spaces here, i.e.
```
Clicking “Send referral” will show the referral in the activity of__{companyName}
```

```suggestion
            Clicking “Send referral” will show the referral in the activity of{' '}
            {companyName}, as well as in the Referrals section on both
```

mforner13: Good catch! Have pushed the change :)

src/apps/referrals/apps/send-referral/client/SendReferralConfirmation.jsx
```diff
@@ -0,0 +1,74 @@
+import React from 'react'
+import styled from 'styled-components'
+import PropTypes from 'prop-types'
+
+import { SummaryTable, FormActions } from 'data-hub-components'
+import { H4, Button, Link } from 'govuk-react'
+import SecondaryButton from '../../../../../client/components/SecondaryButton'
+
+const StyledParagraph = styled('p')`
+  font-size: 16px;
```
debitan: OPTIONAL: you can use `FONT_SIZE.SIZE_16` from the govuk react constants here

mforner13: If it's cool I think I'll keep it as 16px, the constant feels a bit unnecessary as i don't think its value would change from 16.

PR #2474, opened 2020-03-05
Check the value of a global account manager before it is transformed

cgsunkel:
## Description of change

This PR fixes a bug that causes Data Hub to hang and eventually leads to a 504 error when a user clicks on ‘View full business details’ on affected one list companies (this also appears in Sentry). An example of this can be found [here](https://www.datahub.staging.uktrade.io/companies/f1c4ae9e-2ec3-4f3e-b4d2-3e82c0707058/activity) - an error message should appear in the 'data-hub-sentry' channel after the link is clicked.

The bug is caused if someone using Django Admin changes a company to the one list tier but does not set the one list account owner. In the code, the translator always expects `one_list_global_account_manager` to be defined and throws an unexpected error when encountering 'null'

This PR allows the transformer to check the value of `one_list_global_account_manager` before the transformation process starts, and returns a value of ‘not set’ if the variable hasn’t been set beforehand.

## Test instructions

Go the the company identified above in the Heroku app and the business details page should appear normally

### Before

Screenshot 2020-03-05 at 09 49 28

### After

A normal business details page should appear

## Checklist

### DIFF
--- a/src/apps/companies/apps/business-details/transformers.js
+++ b/src/apps/companies/apps/business-details/transformers.js
@@ -3,11 +3,14 @@ const { get, pick, compact, omitBy } = require('lodash')
 const { hqLabels } = require('../../labels')
 const { convertUsdToGbp } = require('../../../../common/currency')
 
-const transformGlobalAccountManager = ({ dit_team, name }) => {
+const transformGlobalAccountManager = (globalAccountManager) => {
+  if (!globalAccountManager) {
+    return null
+  }
+  const { dit_team, name } = globalAccountManager
   const region = get(dit_team, 'uk_region.name')
   const country = get(dit_team, 'country.name')
   const items = compact([name, get(dit_team, 'name'), region || country])
-
   return items.length ? items : 'Not set'
 }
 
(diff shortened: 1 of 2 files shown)


src/apps/companies/apps/business-details/transformers.js
```diff
@@ -3,11 +3,14 @@ const { get, pick, compact, omitBy } = require('lodash')
 const { hqLabels } = require('../../labels')
 const { convertUsdToGbp } = require('../../../../common/currency')
 
-const transformGlobalAccountManager = ({ dit_team, name }) => {
+const transformGlobalAccountManager = (globalAccountManager) => {
+  if (!globalAccountManager) {
+    return null
```
web-bert: Should this return `'Not set'` to match the return on line 14 or should line 14 return `null` instead?

cgsunkel: This is to satisfy the unit test for passing in an empty object, if I use `Not set` here the returned object will no longer be empty and the test will fail
Screenshot 2020-03-05 at 11 31 35

cgsunkel: I have changed the return of line 14 to null

PR #2478, opened 2020-03-06
Group similar export history items

web-bert:
## Description of change

This is a temporary FE fix to group the API data to reduce items in the list. The API is expected to do this in the future at which point this code will be removed.

It will group export history country changes to show the same actions on the same line. Currently if you add two countries they will show as two line items when it should be one.

## Test instructions

Edit the export countries either on the export tab or by adding an interaction with "Were countries discussed" as "Yes" and then click on "View full export countries history" and then countries added to the same category should appear in the same line

### Before

exports_before

### After

export_after

## Checklist

### DIFF
--- a/src/apps/companies/apps/exports/group-history-items.js
+++ b/src/apps/companies/apps/exports/group-history-items.js
@@ -1,0 +1,43 @@
+const GROUP_TIME_DELTA = 1000
+
+const isInteraction = (item) => item.hasOwnProperty('kind')
+
+function isTimeMatch(timestampA, timestampB) {
+  const dateA = Date.parse(timestampA)
+  const dateB = Date.parse(timestampB)
+
+  return Math.abs(dateA - dateB) < GROUP_TIME_DELTA
+}
+
+function isHistoryItemMatch(a, b) {
+  const userMatch = a.history_user.id === b.history_user.id
+  const typeMatch = a.history_type === b.history_type
+  const statusMatch = a.status === b.status
+
+  return userMatch && typeMatch && statusMatch && isTimeMatch(a.date, b.date)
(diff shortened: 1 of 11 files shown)


src/apps/companies/apps/exports/group-history-items.js
```diff
@@ -0,0 +1,45 @@
...
+}
+
+function isTimeMatch(timestampA, timestampB) {
+  const dateA = Date.parse(timestampA)
+  const dateB = Date.parse(timestampB)
+
+  return Math.abs(dateA - dateB) < GROUP_TIME_DELTA
+}
+
+function historyItemsMatch(a, b) {
```
adamledwards: Optional
I notice you used `isTimeMatch ` and  `isInteraction ` for comparison would be nice `isHistoryItemsMatch` to show it also used as a comparison function

web-bert: Good point, thanks I will update 👍

src/apps/companies/apps/exports/group-history-items.js
```diff
@@ -0,0 +1,45 @@
+const GROUP_TIME_DELTA = 1000
+
+const isInteraction = (item) => item.hasOwnProperty('kind')
+
+function createCountries(item) {
```
adamledwards: Optional would it be better not to modify the original value 
```suggestion
function createCountries(item) {
  const { country, ...rest } = item
  return {
    ...rest,
    countries: country
  }
  
}
```

web-bert: I have refactored to remove the function, thanks 👍

PR #2479, opened 2020-03-06
Add cypress recording capability to functionals and e2e

PippoRaimondi:
## Description of change

Add video capability for failing tests and enhance current flaky tests due to chrome upgrade

## Checklist

### DIFF
--- a/test/end-to-end/cypress/specs/DIT/interaction-spec.js
+++ b/test/end-to-end/cypress/specs/DIT/interaction-spec.js
@@ -61,7 +61,7 @@ describe('Service delivery', () => {
     )
   })
 
-  it('should add the service delivery', () => {
+  it('should create the service delivery', () => {
     const subject = 'Some interesting service delivery'
 
     const formSelectors = selectors.interactionForm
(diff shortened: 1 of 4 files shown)


test/functional/cypress/specs/companies/send-referral-spec.js
```diff
@@ -8,14 +8,13 @@ const {
 
 const selectTypeahead = (fieldName, input, spanNumber) =>
```
rafenden: `spanNumber` can be now removed.

test/end-to-end/cypress/specs/DIT/interaction-spec.js
```diff
@@ -79,11 +79,14 @@ describe('Service delivery', () => {
...
       'Service delivery created'
     )
+  })
 
+  it('should display newly created service delivery', () => {
     cy.visit(interactions.index())
     cy.get(selectors.filter.interaction.myInteractions).click()
 
-    cy.get(selectors.collection.items)
+    cy.reload()
```
rafenden: Could you please explain why we need to reload?

PippoRaimondi: Reading about cypress retry mechanism, when a chain has an assertion and an element is not found within that assertion, it retries the chain until the timeout set in our config (5s). But rereading it to send you the docs, just noticed this: `Cypress only retries commands that query the DOM` so the above might make the test pass because gives the backend that extra time to sync the newly created service delivery... but reload is not the right solution for this now... I will rethink and push a change on this line

test/functional/cypress/specs/companies/send-referral-spec.js
PippoRaimondi: good catch, thanks!

PR #2484, opened 2020-03-09
Make TabNav accessible

peterhudec:
## Description of change

It turned out that [`govuk-react/tabs`](https://github.com/govuk-react/govuk-react/tree/master/components/tabs) on which the `TabNav` component was implemented is not accessible at all. It doesn't use any ARIA labels whatsoever. Alas, it doesn't even allow you to add the missing ARIA labels, because it just ignores all the props it doesn't use. What it actually does is some focus management, which is completely at odds with the [MDN tab role recommendation](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Tab_Role).

What's worse is that it looks like `govuk-react` is a pretty much a [dead project](https://github.com/govuk-react/govuk-react/graphs/commit-activity) as it's most recent commit is from 4th July 2019, which is about 8 months ago. There are hovewer 111 open issues and 12 dangling PRs.

This PR reimplements `TabNav` based on the [example from the aforementioned MDN page](https://mdn.mozillademos.org/en-US/docs/Web/Accessibility/ARIA/Roles/Tab_Role$samples/Example?revision=1573222) and it implements all the best practices it mentions. Styling was written complete from scratch without any usage of `govuk-react`.

## Test instructions

As the component is not used anywhere yet, you only can test it by copying and pasting the following code somewhere. To actually test it, first read the [MDN article](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Tab_Role) and then try to use the component firs only with the keyboard, but then also with a screen reader.

```js
// Edit the import path depending where you paste this
import TabNav from 'src/client/components/TabNav'

<TabNav
  id="abc"
  tabs={[
    {
      label: 'Foo',
      content: 'foo content',
    },
    {
      label: 'Bar',
      content: 'bar content',
    },
    {
      label: 'Baz',
      content: 'baz content',
    },
  ]}
/>
```

## Checklist


src/client/components/TabNav/index.jsx
```diff
@@ -1,63 +1,115 @@
...
-        {tabs.map(({ label }, index) => (
-          <Tabs.Tab
+    <StyledTablist
+      tabindex={0}
+      role="tablist"
+      aria-label={label}
+      onKeyUp={({ keyCode }) => {
+        const fi = focusIndex === undefined ? selectedIndex : focusIndex
+        keyCode === 39 && onFocusChange((fi + 1) % tabs.length)
+        keyCode === 37 &&
```
paulgain: Can we put both of these values (`39` and `37`) into constants and name them appropriately. There's no way of knowing what they refer to without looking at an ASCII table.

PR #2488, opened 2020-03-10
Add new lightweight dockerfile, add docker-compose for dev stack and mock stack

debitan:
Add new lightweight dockerfile, add docker-compose for dev stack and mock stack (improved from the previous iteration). 

The motivation for doing this is twofold:
1. You can simultaneously run the dev stack and mock/sandbox stack. 
2. Simplify onboarding for new devs. 

I've also added package.json scripts to allow easy running of these stacks. 
Either can be run with a simple comand - `yarn dockermock` or `yarn dockerdev`
This will build the images and then run the containers. 

You will need env vars in a file called (provisionally) `dev-stack.env`. I've provided a `dev-stack.sample.env` file listing the keys required.


Dockerfile.dev
```diff
@@ -0,0 +1,11 @@
+FROM node:12.16.1
+
+WORKDIR /usr/src/app
+
+COPY package*.json ./
+COPY yarn.lock ./
+RUN yarn
```
adamledwards: How does that work is yarn already installed?

docker-compose.dev.yml
```diff
@@ -0,0 +1,58 @@
...
+      - .:/usr/src/app
+      - /usr/src/app/node_modules
+    ports:
+      - 3000:3000
+      - 9229:9229
+    depends_on:
+      - redis
+      - api
+    env_file: dev-stack.env
+    command: sh -c './wait-for-it.sh api:8000 -t 240 -- yarn develop'
```
adamledwards: I don't think you need to wait for the back end to start. 
I don't think it resembles a real work scenario as the api and front end don't wait for each other. 
Also, It could slow down start up time.

docker-compose.dev.yml
```diff
@@ -0,0 +1,58 @@
...
+      context: .
+      dockerfile: Dockerfile.dev
+    volumes:
+      - .:/usr/src/app
+      - /usr/src/app/node_modules
+    ports:
+      - 3000:3000
+      - 9229:9229
+    depends_on:
+      - redis
```
adamledwards: Is it dependant on redis? I thought that the api or celery service would be dependant

docker-compose.dev.yml
debitan: it does because the current start up process for the frontend involves loading metadata from the api before the frontend will come up. without it it will fail

Dockerfile.dev
debitan: good point, I probably need that. npm comes with node, so I could just use npm. means diverging from the use of yarn though...

debitan: saying that though, it does work for me without installing yarn. wouldn't have thought it uses my global install

docker-compose.dev.yml
debitan: true. I don't need that here as previous services will have waited for it

Dockerfile.dev
web-bert: The base image has yarn installed: https://github.com/nodejs/docker-node/blob/5351774b970cd32b80fc0c47a8abff5ba155fccc/12/stretch/Dockerfile

docker-compose.dev.yml
web-bert: The FE is dependant on Redis, so I would leave it

dev-stack.sample.env
```diff
@@ -0,0 +1,38 @@
+# frontend env vars
```
rafenden: Are we sure there are no overlapping vars between FE and BE?

docker-compose.dev.yml
rafenden: Could we use `dockerize` instead? e.g.
```
dockerize -wait http://api:8000/ping.xml -timeout 600s
```

dev-stack.sample.env
debitan: there doesn't appear to be

docker-compose.dev.yml
debitan: dockerize involves installing a package and is quite a bit heavier. this simple bash script does exactly what we need to. my aim was to keep it as lean as possible and quick to spin up

debitan:
So I've tested this with @LaurenQurashi on her totally new clean mac. 7.5 mins to build the images from scratch - which is pretty good. Then c 2mins to bring up the containers first time round. So totally set up in ten minutes 👍 and of course subsequent runs are much faster thanks to caching. 

Did find though that the frontend container is reading from the `.env` file, but the `docker-dev.env` file trumps it if there are any clashes. However, that means I need to add some more env vars to the `docker-dev.env` as the joi env schema checking fails with what is there at the moment if you don't have a fully fledged `.env` file already. 

I will also add the docker-dev envs into vault as they can't all be inferred from the current dev env vars on vault. This should make onboarding new devs really easy. 

The only sticking point to this being a one stop shop for running the stack after you have access to the env config and docker installed if setting up the access token on the django admin. Would be keen to hear from BE folks about how this could possibly be automated?

docker-compose.dev.yml
rafenden: We are using `dockerize` in every docker service so it makes sense to use it for frontend too.

It's also less code to maintain.

rafenden: Perhaps, it's something we could change and call metadata when it's actually used. It's cached anyway.

debitan: I would suggest then that we replace `dockerize` as this is a lighter approach as we are only using it to wait for other services to come up.

debitan: We were speaking about this at lunch and we definitely should. But that isn't in scope of this PR.

docker-compose.dev.yml
debitan: it is, but if we say the api is dependent on redis and then the frontend is dependent on the api, it is inherent.

web-bert: I'd rather see an explicit dependency link as the FE cannot start without it

docker-compose.dev.yml
rafenden: That's the exact purpose of `dockerize`.

How slower the `dockerize` script is when compared to a custom bash script?

wait-for-it.sh
```diff
@@ -0,0 +1,182 @@
+#!/usr/bin/env bash
+# Use this script to test if a given TCP host/port are available
```
rafenden: We should give some credit to the author.

```suggesiton
# Use this script to test if a given TCP host/port are available
# Original version by Giles Hall: https://github.com/vishnubob/wait-for-it
```

debitan:
> So I've tested this with @LaurenQurashi on her totally new clean mac. 7.5 mins to build the images from scratch - which is pretty good. Then c 2mins to bring up the containers first time round. So totally set up in ten minutes 👍 and of course subsequent runs are much faster thanks to caching.
> 
> Did find though that the frontend container is reading from the `.env` file, but the `docker-dev.env` file trumps it if there are any clashes. However, that means I need to add some more env vars to the `docker-dev.env` as the joi env schema checking fails with what is there at the moment if you don't have a fully fledged `.env` file already.
> 
> I will also add the docker-dev envs into vault as they can't all be inferred from the current dev env vars on vault. This should make onboarding new devs really easy.
> 
> The only sticking point to this being a one stop shop for running the stack after you have access to the env config and docker installed if setting up the access token on the django admin. Would be keen to hear from BE folks about how this could possibly be automated?

@currycoder  the question in the last paragraph here is the one I mentioned to you yesterday on the way out.

docker-compose.dev.yml
currycoder: I'm with Raf here... Looks like dockerize is just 2.9MB and it avoids us absorbing a static `wait-for-it.sh` script (which is unlikely to be updated) in to the codebase(s).

debitan: Ok. Will add dockerize to the dockerfile.

docker-compose.dev.yml
```diff
@@ -0,0 +1,57 @@
...
+      - /usr/src/app/node_modules
+    ports:
+      - 3000:3000
+      - 9229:9229
+    depends_on:
+      - api
+    env_file: dev-stack.env
+    command: sh -c './wait-for-it.sh api:8000 -t 240 -- yarn develop'
+
+  api:
```
currycoder: There's some potential here for things to go out of sync between FE and BE...
I'll have a go at running the stack this way for dev work next week (might also be good to invite other BE devs to try)..  If it works for everybody I think we should probably look at how we can point to this one way for bringing up dev copies.

docker-compose.dev.yml
```diff
@@ -0,0 +1,57 @@
...
+    build: ../data-hub-api
+    volumes:
+      - ../data-hub-api:/app
+    env_file: dev-stack.env
```
currycoder: I don't think that the FE should have responsibility for managing BE's dev environment variables.  IMO this is asking for breakages (e.g. when we add a new environment variable to BE and don't reflect it in this repo).
Could this point to BE's .env file instead?

dev-stack.sample.env
```diff
@@ -0,0 +1,50 @@
...
+PERFORMANCE_DASHBOARDS_URL=
+ONE_LIST_EMAIL=
+HELP_CENTRE_URL=
+HELP_CENTRE_FEED_API_TOKEN=
+HELP_CENTRE_ANNOUNCMENTS_URL=
+FIND_EXPORTERS_URL=
+ARCHIVED_DOCUMENTS_BASE_URL=
+
+# api env vars
+ACTIVITY_STREAM_ACCESS_KEY_ID=
```
currycoder: I find it odd that this env file is totally un-opinionated about values.  Where do you expect new devs to populate the values from?

docker-compose.dev.yml
debitan: I tried, but unfortunately I couldn't get the `env_file` to reach outside of the repo. Could do some more research on it.

dev-stack.sample.env
debitan: I will need to add a new section to vault with all the values. And then point to this in the readme.

debitan:
> This is a little difficult for me to review right now... Some things look fine, but I'm unable to test this out myself as there are no instructions for a new dev on where to get env var variables and how to run everything - I think there should be a readme entry which assumes that somebody is starting totally fresh.
> 
> In fact, it looks like the instructions in the README.md are now broken: https://github.com/uktrade/data-hub-frontend/blob/master/README.md#running-project-within-docker

Yep, we definitely need a readme. I was going to do it in a follow up PR, but I guess it makes sense to do it here. Just want to make sure that the approach is sound before writing it all down. I'll look to do this though and add the env vars into vault so that this is easier to review.

docker-compose.dev.yml
debitan: I've tried doing this, but dockerize doesn't work without the installation of a full linux distro image. In the `Dockerfile.dev` file I'm only pulling down node to decrease the weight. 

Can I suggest we retain the `wait-for-ti.sh` script for the time being and look to refactor the start up of the app so the frontend doesn't need to wait for the api in order to get the metadata?

debitan:
@currycoder you might have seen I added docs and the env vars to vault over the weekend. I also ran through this with @rachaelcodes yesterday, asking her to read the docs and try and follow it on her own. This resulted in a few changes I will make: 
- add instructions around how to log into vault and get github key
- rename vault secrets to match env file names
- add slash to admin url
- add time for token into docs

I'm just working on the dependabot stuff at the moment and won't get to this for a little while, so just laying it out here for now. 

Still keen to hear from you if you think automating the creation of an access token for the backend is possible? As this is would make the whole experience much smoother and quicker.

docker-compose.dev.yml
rafenden: Node image is based on debian distro so this command should install dockerize (`Docker.dev`):
```
RUN apt-get update && apt-get install -y wget

ENV DOCKERIZE_VERSION v0.6.1
RUN wget https://github.com/jwilder/dockerize/releases/download/$DOCKERIZE_VERSION/dockerize-linux-amd64-$DOCKERIZE_VERSION.tar.gz \
    && tar -C /usr/local/bin -xzvf dockerize-linux-amd64-$DOCKERIZE_VERSION.tar.gz \
    && rm dockerize-linux-amd64-$DOCKERIZE_VERSION.tar.gz
```

docker-compose.dev.yml
rafenden: The FE is also using redis to cache responses from the API.

debitan: it is. but, the frontend is dependent on the api, which in turn is dependent on redis starting up first, so it will work no problem.

debitan:
> @currycoder you might have seen I added docs and the env vars to vault over the weekend. I also ran through this with @rachaelcodes yesterday, asking her to read the docs and try and follow it on her own. This resulted in a few changes I will make:
> 
> * add instructions around how to log into vault and get github key
> * rename vault secrets to match env file names
> * add slash to admin url
> * add time for token into docs
> 
> I'm just working on the dependabot stuff at the moment and won't get to this for a little while, so just laying it out here for now.
> 
> Still keen to hear from you if you think automating the creation of an access token for the backend is possible? As this is would make the whole experience much smoother and quicker.

This is done now.

reupen:
Have we had a look at [passing multiple docker-compose files to docker-compose](https://docs.docker.com/compose/extends/) rather than defining the API services here? I think that could easily be aliased and would avoid having to define and duplicate too many API things here.

docker-compose.dev.yml
currycoder: I'm fairly sold on the idea of putting these in vault, but the question I think still stands:

How do we get devs (both FE and BE) to keep that file up to date as part of a PR process?

README.md
```diff
@@ -35,46 +35,43 @@ and be provided with a back end server to provide the API, data storage and sear
...
-    that the containers share a network. The following snippet in `.env` should then
-    do the trick:
+3. Install docker if you don't have it already.
 
-    ```
-    API_ROOT=http://api:8000
-    ```
-  
-5.  Start the server:
+4. Run `docker-compose -p dev -f docker-compose.dev.yml build --parallel && docker-compose -p api -f docker-compose.dev.yml up --abort-on-container-exit`. This `docker-dev` script is also in the `.bin.sample` file. Feel free to move it to your own `.bin` file. Alternatively, if you already have yarn installed globally on your machine, you can run `yarn docker:dev`.
```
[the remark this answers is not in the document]
debitan: the reason for this was to avoid clashes between instances in different docker compose containers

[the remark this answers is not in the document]
debitan: we already have a `docker-compose.yml` so didn't want to delete that one at the moment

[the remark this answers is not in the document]
debitan: yes, I also probably need to make it clear that front end and api need to be within the same parent folder for this to work

README.md
```diff
@@ -35,46 +35,43 @@ and be provided with a back end server to provide the API, data storage and sear
...
-3.  By default, the frontend container will use the mock-sso instance specified
-    in `docker-compose.yml`.  To use this, you will also need to specify the `MOCK_SSO_USERNAME`
-    environment variable in `.env`.
-    This value should correspond with an `Advisor.email` value in the Data Hub backend.
+1. Go to the project root
 
-4.  Optionally, you can point the frontend to a different backend by specifying
-    an `API_ROOT` in `.env`. e.g. To connect the frontend docker container to the
-    shared dev environment backend.
+2. Create a `dev-stack.env` and populate it with the variables from vault. You will probably have to do some formatting to convert the values from json. To log into vault you'll need to be on the VPN and create a personal access token from the developer settings in your GitHub account.
```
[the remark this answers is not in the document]
debitan: I've had push back before on including links to our more sensitive materials in docs before. 
I guess the premise is that you won't be embarking down this path until you have been given access to the vpn to access vault anyway, so you will have to speak to someone.

README.md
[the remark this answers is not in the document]
debitan: Isn't this approach a bit more robust though? And it isn't any extra typing as it is a copy and paste job here and then it is also provided as part of a script to run using bash or yarn.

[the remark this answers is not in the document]
debitan: changed it in vault

docker-compose.dev.yml
debitan: I think that's a valid concern, but I can't think of a solution to it at the moment that is better than a) please remember and b) we'll find out when it breaks

README.md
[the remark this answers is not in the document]
currycoder: In that case, can you perhaps make a private gist or github page of some kind to link to here?  And then that has a link to the sensitive link?  I just think as much as possible a README should reduce the need to have to get somebody to help you.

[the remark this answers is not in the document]
debitan: nice idea. I'll look into that.

README.md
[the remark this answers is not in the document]
currycoder: I guess that this is another area where devs can customise things to their own preferences.  For me personally, I would rather run the whole command instead of depending on aliases.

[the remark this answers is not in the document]
debitan: I think so. Let's aim for keeping it as robust in the default, but it is totally customisable once you're up and running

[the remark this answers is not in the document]
currycoder: But this PR does delete `docker-compose.yml` I think!  TBH, I think it would be good to just delete it and go with a single approach for people do FE/docker.

debitan:
> Have we had a look at [passing multiple docker-compose files (i.e. the FE and API ones) to docker-compose](https://docs.docker.com/compose/extends/) rather than defining the API services here? I think a command like that could easily be aliased and would avoid having to define and duplicate too many API things here.

@reupen I'm keen on this idea. I think though that we should iterate. So it would be good to get this PR in first and then build off the back of that.

reupen:
Regarding the vault, I think requiring access to it in the instructions is a bit problematic for an open-source repository. It (or Rattic) is a probably a good option for secrets, but I think there should still be a working sample configuration in the repo so that only secrets have to be possibly obtained from the vault or Rattic.

README.md
```diff
@@ -43,31 +43,45 @@ and be provided with a back end server to provide the API, data storage and sear
...
+   **Note:** If you don't have access to vault to get the secrets you can copy the `dev-stack.sample.env` file to `dev-stack.env` and get up and running. Not everything will work right, but it should run.
 
-6. Navigate to access tokens, find your superuser account, add the access token which is equivalent to `OAUTH2_DEV_TOKEN` in your env file. Set an expiry time and date for your token. The easiest way to do this is to just click today and now and then increase the year by one. Finally, define the scope as `data-hub:internal-front-end`
+3. Install docker if you don't have it already - https://docs.docker.com/get-docker/
 
-7. You can now access the frontend at `http://localhost:3000` and the backend at `http://localhost:8000`.
+4. Run `docker-compose -f docker-compose.yml build --parallel && docker-compose -p dev -f docker-compose.yml up --abort-on-container-exit`. This is provided for you in the `dev-stack` file in the `.bin/sample` folder. Feel free to move it to your own `.bin` file. Alternatively, if you already have yarn installed globally on your machine, you can run `yarn dev-stack`.
```
currycoder: I think you can safely remove `-f docker-compose.yml` from this command now :raised_hands:

debitan: ah yes, good shout

PR #2498, opened 2020-03-15
Update dependabot config to update all packages and fix conflict

debitan:
There are currently two issues open regarding our dependabot config. 

https://github.com/uktrade/data-hub-frontend/issues/2361
https://github.com/uktrade/data-hub-frontend/issues/2496

One says it cannot find our default branch 'develop'. I've checked and the default branch is now master. Hopefully updating the conflict in the config will also resolve this issue. 

This new config _should_ also create pull requests to update all of our packages, helping us keep up to date. 

I have checked this config using the dependabot config validator - https://dependabot.com/docs/config-file/validator/

### DIFF
--- a/.dependabot/config.yml
+++ b/.dependabot/config.yml
@@ -7,16 +7,3 @@ update_configs:
     update_schedule: live
     default_reviewers:
       - datahub-fed
-    allowed_updates:
-      - match:
-          update_type: "security"
-
-  - package_manager: javascript
-    directory: "/"
-    update_schedule: weekly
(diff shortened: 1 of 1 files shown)


web-bert:
@debitan it might be a good idea to exclude `govuk-colours` from the list as we can't update until all colours match between `govuk-frontend` and `govuk-colours`. This is the snipped to exclude:
```yaml
ignored_updates:
      - match:
          dependency_name: "govuk-colours"
```

PR #2499, opened 2020-03-16
My referrals on dashboard

peterhudec:
## Description of change

The dashboard is now a tab navigation which contains the existing _My companies_ lists and also the new _My referrals_ list, which shows all the referrals related to the current user. 

## Test instructions

1. Go to the homepage
2. You should see a tab navigation with tabs _My companies lists_ and _My referrals_
3. Click on the _My referrals_ tab
4. After the progress indicator disappears, you should either see an error message or a list of referrals
5. Click on the _My companies lists_
6. After the progress indicator disappears, you should see the original _My companies lists_ section

### Before
Screen Shot 2020-03-16 at 9 39 12 AM

### After
Screen Shot 2020-03-16 at 9 39 42 AM


src/client/components/ContentWithHeading.jsx
```diff
@@ -0,0 +1,25 @@
...
+import SpacedSectionBreak from './SpacedSectionBreak'
+
+const ContentWithHeading = ({
+  heading,
+  children,
+  headingComponent: Heading = H3,
+}) => (
+  <>
+    <Heading>{heading}</Heading>
+    <SpacedSectionBreak />
```
ian-leggett: Could we not just style the `<SectionBreak>` here?

src/client/components/CompanyLists/index.jsx
```diff
@@ -70,22 +67,24 @@ const CompanyList = connect((state) => {
   </div>
 ))
 
+const EmptyState = () => (
+  <ContentWithHeading heading="My companies lists">
```
web-bert: I wonder whether this heading is required as it's just duplicate content: 
Screenshot 2020-03-17 at 19 01 41

src/client/components/ReferralList/index.jsx
```diff
@@ -1,44 +1,44 @@
...
+    </StyledOl>
+  </ContentWithHeading>
 )
 
 const EmptyState = () => (
-  <div>
-    <H3>My Referrals</H3>
-    <SectionBreak visible={true} />
-    <p>
+  <ContentWithHeading heading="My referrals">
```
web-bert: Same here, is this heading really needed as it's just duplicate content:
Screenshot 2020-03-17 at 19 02 16

src/client/components/ContentWithHeading.jsx
```diff
@@ -0,0 +1,25 @@
+import PropTypes from 'prop-types'
+import React from 'react'
+import { H3 } from 'govuk-react'
+
+import SpacedSectionBreak from './SpacedSectionBreak'
+
+const ContentWithHeading = ({
+  heading,
+  children,
+  headingComponent: Heading = H3,
```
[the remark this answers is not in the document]
peterhudec: This is to allow overriding the heading through props, but as currently we only use H3 I can just remove it. This is not an assignment though, I'm destrucuturing `headingComponent` under the name `Heading` and if it's undefined, falling back to `H3`.
```
<ContentWithHeading heading="Foo" headingComponent={H2}>
  Content
</ContentWithHeading>
```

src/client/components/CompanyLists/index.jsx
peterhudec: Good point, will ask Rogier about this.

src/client/components/ReferralList/index.jsx
peterhudec: The heading should be there, but it should read "2 referrals".

src/client/components/ContentWithHeading.jsx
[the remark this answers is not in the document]
web-bert: Ah yes sorry, it was the destructuring syntax that threw me off :-)

src/client/components/ContentWithHeading.jsx
peterhudec: We will pretty much always want to have the section break with margins. I already was doing it twice, but then I created the `ContentWithHeading` so now it's only used once.

src/client/components/ReferralList/index.jsx
web-bert: That is fine when there are referrals but when there are none it has the duplicate heading which I think could be improved

peterhudec: Actually this is the empty state. Will discuss this with Rogier too

src/client/components/ContentWithHeading.jsx
```diff
@@ -0,0 +1,25 @@
...
+}) => (
+  <>
+    <Heading>{heading}</Heading>
+    <SpacedSectionBreak />
+    {children}
+  </>
+)
+
+ContentWithHeading.propTypes = {
+  heading: PropTypes.oneOfType([PropTypes.string, PropTypes.element]),
```
rafenden: Why not to use`PropTypes.node`?

src/client/components/ContentWithHeading.jsx
```diff
@@ -0,0 +1,25 @@
...
+  children: PropTypes.oneOfType([PropTypes.string, PropTypes.element]),
+  headingComponent: PropTypes.func,
```
rafenden: Why not to pass custom component via `heading` instead?  
e.g.

```
<ContentWithHeading heading={<H2>Foo</H2>}>
  Content
</ContentWithHeading>
```

src/client/components/Dashboard.jsx
```diff
@@ -0,0 +1,30 @@
+import PropTypes from 'prop-types'
+import React from 'react'
+
+import CompanyLists from './CompanyLists'
+import ReferralList from './ReferralList'
+import TabNav from './TabNav'
+
+const Dashboard = ({ id }) => (
+  <TabNav
+    id={`${id}:TabNav`}
```
rafenden: I suggest using `.` to be compatible with HTML specification:
https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id#Specifications

src/client/components/TabNav/index.jsx
```diff
@@ -132,6 +133,17 @@ const TabNav = ({
   </div>
 )
 
+TabNav.propTypes = {
+  label: PropTypes.string.isRequired,
+  tabs: PropTypes.arrayOf(
+    PropTypes.shape({
+      label: PropTypes.oneOfType([PropTypes.string, PropTypes.element])
```
rafenden: Why not to use `PropTypes.node` instead?

test/functional/cypress/specs/company-lists/view-1-spec.js
```diff
@@ -6,6 +6,9 @@ describe('My companies lists', () => {
     cy.visit('/')
   })
 
+  it('My companies lists tab should be selected', () =>
```
rafenden: Can we use lowercase letter here to match the rest of the tests?

test/functional/cypress/support/commands.js
```diff
@@ -5,3 +5,157 @@ Cypress.Commands.add('selectFirstOption', (selector) => {
...
+ * @param {String} options.logName - The name of the log message e.g. 'GET'
+ * @param {(...args) => Promise} options.body - The actual implementation of the
+ * command It can take any number of arguments, but the last one is treated as
+ * options. It should return the Cypress promise, if it's a continuation command.
+ * @param {(...args) => String} options.getLogMessage - A function which takes
+ * all the command arguments an should return a string which will be used as
+ * the log message.
+ * @returns undefined
+ */
+const addLoggedCommand = ({
```
rafenden: We didn't agree to use commands, PR stage is too late for such discussion.

src/client/components/CompanyLists/index.jsx
rafenden: The heading is bigger when compared to the design, there's not enough spacing from the tabs.

image

test/functional/cypress/support/commands.js
```diff
@@ -5,3 +5,157 @@ Cypress.Commands.add('selectFirstOption', (selector) => {
...
+  name,
+  logName,
+  body,
+  getLogMessage,
+  options = {},
+}) =>
+  Cypress.Commands.add(name, options, (...args) => {
```
rafenden: That's a lot of code which could be simply done with `cy.log`

src/client/components/ContentWithHeading.jsx
peterhudec: The point of this component is abstracting away commonalities, i.e. there was a need for consistent heading size and spacing when switching the two tabs. Your suggestion is halfway to not having this component at all.

test/functional/cypress/support/commands.js
peterhudec: The `ariaTablistTabShouldBeSelected` command is the only one that makes an assertion. It doesn't really need to be a command.

src/client/components/CompanyLists/index.jsx
peterhudec: The designs are not really mockups. The heading has the same size as the one in _My company lists_, which is important. Also I'm using a vanilla `govuk-react` heading here which is part of the typography.

peterhudec: I agree with the spacing

test/functional/cypress/support/commands.js
peterhudec: I added this because just using `cy.log` created a snapshot where the selected DOM element was not highlighted. I should have mentioned it in the dosctring.

peterhudec:
> @peterhudec This is the current text & colour design for the 'completed referral' badge:
> Screenshot 2020-03-17 at 16 36 52



> @peterhudec This is the current text & colour design for the 'completed referral' badge:
> Screenshot 2020-03-17 at 16 36 52
> Screenshot 2020-03-17 at 16 49 27

@mforner13 The green color is the same we are using in activity feed badges. The two labels: _Referral accepted_ and _Outstanding referral_ sound a bit inconsistent. I will ask Rogier about it.

src/client/utils/multiinstance.js
```diff
@@ -1,3 +1,4 @@
+import React from 'react'
```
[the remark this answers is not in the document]
peterhudec: Agree, but perhaps it doesn't need to be in this PR

test/functional/cypress/specs/company-lists/view-1-spec.js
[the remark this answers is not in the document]
peterhudec: But the test asserts that it is selected, it doesn't select anything.

test/functional/cypress/specs/company-lists/view-2-spec.js
```diff
@@ -6,6 +6,9 @@ describe('My companies lists', () => {
     cy.visit('/')
   })
 
+  it('My companies lists tab should be selected', () =>
```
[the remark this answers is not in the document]
peterhudec: Same here, the test asserts that the tab is selected and doesn't select it.

src/client/components/Dashboard.jsx
peterhudec: Just to clarify that this `id` prop doesn't leak into the DOM.

test/functional/cypress/support/commands.js
peterhudec: I had this implemented before we even agreed to have the discussion about Cypress, I merely mentioned it there as something I've done.

rafenden: I don't see any advantages of using commands, but there are many disadvantages like:

* No intellisense (not everyone is using VS code)
* Defining and using these commands are like magic
* Increased complexity (a lot), have a look at https://github.com/uktrade/data-hub-frontend/pull/2499/files#diff-0b9b6ebf9848c47833cc49050e9bd38eR31

>I had this implemented before we even agreed to have the discussion about Cypress, I merely mentioned it there as something I've done.

You should discuss this with the rest of developers before doing that work as it's much different to the way helper assertions were defined so far.

 @web-bert @markhigham @debitan  any thoughts?

src/client/components/ContentWithHeading.jsx
rafenden: Well, perhaps we don't need that component at all as it only adds `SpacedSectionBreak`?

```
<Heading>{heading}</Heading>
<SpacedSectionBreak />
```

src/client/components/Dashboard.jsx
rafenden: I know, but it should still adhere to HTML specification as I need to check the source code of `TabNav` to make sure it doesn't.

test/functional/cypress/support/commands.js
markhigham: Interesting. But I'm 90% sure that you want @mcbhenwood

web-bert: I am not a fan of magic and as suggested in the Cypress meeting I think prefixing these commands or somehow differentiating them would help here.  Also I think `ariaTablistTabShouldBeSelected` is taking it too far as it could just be `cy.getAriaTabListSelectedTab(x).should('have.text', y)` so if we agree to keep these maybe the custom commands we define should be "getters only" and keep the assertions as regular Cypress syntax - which is also the main benefit of the commands: the "chainability"

It does seem there is support for intellisense in other editors, it's just not mentioned in their docs, https://intellij-support.jetbrains.com/hc/en-us/community/posts/360006839600-Getting-Cypress-IntelliSense-working so that may not be a blocker for this

From thinking about this I wonder if a "rule to follow" is custom commands are used to either get something from the DOM or do some interactivity (like type a value in the typeahead) and our assertions are used to assert the state?

src/client/components/CompanyLists/index.jsx
```diff
@@ -70,22 +67,24 @@ const CompanyList = connect((state) => {
...
+    <HintText>
+      You have not yet created any lists with companies.
+      <br />
+      You can add companies to lists from a company page, and only you can see
+      these lists.
```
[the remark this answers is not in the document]
peterhudec: That would add quite a big gap between the two. They are too short to be paragraphs anyway.

src/client/components/ContentWithHeading.jsx
peterhudec: It also adds the heading, hence the name

src/client/components/CompanyLists/index.jsx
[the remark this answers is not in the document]
web-bert: I'm not a fan of using a `<br>` as it's a tag for styling only, the gap can be fixed with CSS and a single sentence is a valid paragraph 

> A paragraph is typically a run of phrasing content that forms a block of text with one or more sentences

https://www.w3.org/TR/2013/CR-html5-20130806/dom.html#paragraph
https://www.w3.org/TR/2013/CR-html5-20130806/grouping-content.html#the-p-element

test/functional/cypress/support/commands.js
peterhudec: One advantage I saw was that they can make writing tests more expressive using the cypress idioms and the tab navigation seemed to be a good use case for them. But I'm not an advocate of them and have no problem converting them to functions if we agree that we don't want them.

BTW, the reason I tried it out was that I've stumbled upon the `selectFirstOption` command which was already there.

src/client/components/Dashboard.jsx
peterhudec: Changed it, but I still don't get it, the `id` has nothing to do with HTML.

test/functional/cypress/support/commands.js
web-bert: I saw `selectFirstOption` too, although a search does not yield any uses of it. I think this is a dicussion worth having as I can see a benefit to these custom commands, they just need to be easy to reason about

debitan: For clarity I have no strong feelings on this either way.

src/client/components/CompanyLists/Header.jsx
```diff
@@ -40,7 +41,9 @@ export const Header = connect(state2props, (dispatch) => ({
     }),
 }))(({ selectedId, lists, onChange }) => (
   <StyledRoot>
-    <StyledHeading size={LEVEL_SIZE[3]}>My companies lists</StyledHeading>
+    <StyledHeading size={LEVEL_SIZE[3]}>
+      {pluralize('My companies list', Object.keys(lists).length, true)}
```
web-bert: Optional: put the key length in a `const` to avoid unnecessary duplication

src/client/components/Dashboard.jsx
web-bert: This has not been updated to use a `.` like the other instance?

src/client/components/TabNav/index.jsx
```diff
@@ -55,10 +56,10 @@ const StyledTablist = styled.div({
 
 const StyledTabpanel = styled.div({
   ...focusStyle,
-  marginTop: 10,
+  marginTop: 30,
```
web-bert: This should use `SPACING`

peterhudec: The `TabNav` component is intentionally not using anything from `govuk-react`

PR #2531, opened 2020-03-19
Update packages with dependabot only when security fixes are released

rafenden:
Stop flooding our PR queue and update packages only when security fixes are released.

### DIFF
--- a/.dependabot/config.yml
+++ b/.dependabot/config.yml
@@ -2,13 +2,10 @@
 
 version: 1
 update_configs:
-  - package_manager: javascript
-    directory: '/'
-    update_schedule: live
-    default_reviewers:
-      - datahub-fed
-    ignored_updates:
+  - package_manager: "javascript"
+    directory: "/"
+    update_schedule: "live"
+    allowed_updates:
       - match:
-          dependency_name: 'govuk-frontend'
-      - match:
-          dependency_name: 'govuk-colours'
+          dependency_type: "production"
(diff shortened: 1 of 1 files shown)


.dependabot/config.yml
```diff
@@ -2,13 +2,10 @@
...
-    ignored_updates:
+  - package_manager: "javascript"
+    directory: "/"
+    update_schedule: "live"
+    allowed_updates:
       - match:
-          dependency_name: 'govuk-frontend'
-      - match:
-          dependency_name: 'govuk-colours'
+          dependency_type: "production"
```
mcbhenwood: can't find `dependency_type` in the docs?

debitan:
@rafenden the only reason we were just flooded with PRs is that dependabot hasn't been working for three months. I don't think we should restrict it as we want to be updating as frequently as possible. Once we clear this block of PRs it shouldn't be as bad. What do you think @mcbhenwood ?

.dependabot/config.yml
rafenden: I found it in one of the [examples](https://dependabot.com/docs/config-file/#allowed_updates):
image

rafenden: Some explanation:
https://github.com/dependabot/dependabot-core/issues/953

debitan:
@rafenden we're down to just 2 dependabot prs now so I would say that this is no longer needed and can be closed?

PR #2532, opened 2020-03-19
ignore .bin script folder

debitan:
As it says on the tin. 

I moved scripts from the `./bin/sample` folder into my own `./bin` folder and realised git wasn't set up to ignore this.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -76,3 +76,7 @@ local.log
 # Yalc packages
 .yalc
 yalc.lock
+
+# Ignore bin scripts
+.bin
+


web-bert:
@debitan I have a global git ignore for that folder so it always gets ignored regardless of which repo I am using it for. The benefit of this is it stops me having to edit the .gitignore  for each repo I have to work with - something to consider

debitan:
> @debitan I have a global git ignore for that folder so it always gets ignored regardless of which repo I am using it for. The benefit of this is it stops me having to edit the .gitignore for each repo I have to work with - something to consider

@web-bert smart! I'll still put this in for the plebs like me

PR #2548, opened 2020-03-20
Change dependabot to weekly

debitan:
Also fix the `default_reviewers`. I'm trying to add `datahub-front-end` twice as we need two reviews from the team.

### DIFF
--- a/.dependabot/config.yml
+++ b/.dependabot/config.yml
@@ -4,9 +4,7 @@ version: 1
 update_configs:
   - package_manager: javascript
     directory: '/'
-    update_schedule: live
-    default_reviewers:
-      - datahub-fed
+    update_schedule: weekly
     ignored_updates:
       - match:
           dependency_name: 'govuk-frontend'


debitan:
Don't know if the two default reviewers will work. We might be best trying to auto assign reviewers through pull panda?

.dependabot/config.yml
```diff
@@ -4,9 +4,10 @@ version: 1
 update_configs:
   - package_manager: javascript
     directory: '/'
-    update_schedule: live
+    update_schedule: weekly
     default_reviewers:
```
rafenden: This won't work as dependabot doesn't support teams:
https://github.com/dependabot/feedback/issues/546

debitan: it wasn't working through the dashboard so thought that might be the case. ok, I'll remove default reviewers, and try and add everyone individually through the dashboard. at least that way when people leave and join we don't have to make a code change.

rafenden: You won't be able to add people in dashboard when the config file exist within the project (dashboard is locked).

That was the reason why dependabot config was removed from the component library.

PR #2558, opened 2020-03-20
Make workspace links open in new tab

mforner13:
## Description of change

This PR makes the links to digital workspace open in new tabs and adds text to the links to inform users of this. 

## Test instructions

1. On send referral form

- Go to companies/:companyId/send-referral
- Check the digital workspace link opens in a tab

2) On help page

- Create a referral with the form above, it will appear in the activity feed
- click on the referral in the activity feed
- click 'I cannot complete this referral' 
- check the workspace link opens in a new tab

### DIFF
--- a/src/apps/companies/apps/referrals/help/client/ReferralHelp.jsx
+++ b/src/apps/companies/apps/referrals/help/client/ReferralHelp.jsx
@@ -3,6 +3,7 @@ import { connect } from 'react-redux'
 import Link from '@govuk-react/link'
 import { H2 } from '@govuk-react/heading'
 import { LEVEL_SIZE } from '@govuk-react/constants'
+import { NewWindowLink } from 'data-hub-components'
 import urls from '../../../../../../lib/urls'
 
 import Task from '../../../../../../client/components/Task'
@@ -54,9 +55,9 @@ export default connect(state2props)(
               </p>
               <p>
                 Or{' '}
-                <Link href={urls.external.digitalWorkspace.teams}>
+                <NewWindowLink href={urls.external.digitalWorkspace.teams}>
                   find their contact details on Digital Workspace
-                </Link>
+                </NewWindowLink>
               </p>
               <H2 size={LEVEL_SIZE[3]}>I'm not the right adviser for this</H2>
 
--- a/src/apps/companies/apps/referrals/send-referral/client/SendReferralForm.jsx
+++ b/src/apps/companies/apps/referrals/send-referral/client/SendReferralForm.jsx
@@ -4,6 +4,7 @@ import PropTypes from 'prop-types'
 import { throttle } from 'lodash'
 import axios from 'axios'
 
+import urls from '../../../../../../lib/urls'
 import SendReferralConfirmation from './SendReferralConfirmation.jsx'
 import ValidatedInput from '../../../../../../client/components/ValidatedInput.jsx'
 import styled from 'styled-components'
@@ -15,7 +16,7 @@ import {
   SEND_REFERRAL_FORM__ADVISER_CHANGE,
   SEND_REFERRAL_FORM__CONTACT_CHANGE,
 } from '../../../../../../client/actions'
-import { FormActions, Typeahead } from 'data-hub-components'
+import { FormActions, Typeahead, NewWindowLink } from 'data-hub-components'
 import {
   H4,
   TextArea,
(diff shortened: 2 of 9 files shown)


src/apps/companies/apps/referrals/help/client/ReferralHelp.jsx
```diff
@@ -50,10 +50,14 @@ export default connect(state2props)(
               <p>
                 If you need additional details or information about the
                 referral, contact the adviser who sent the referral by email:{' '}
-                <AdviserDetails {...sendingAdviser} />, Or find more contact
+                <AdviserDetails {...sendingAdviser} />, or find more contact
                 details on{' '}
-                <Link href={urls.external.digitalWorkspace.teams}>
-                  Digital Workspace
+                <Link
```
web-bert: As advised by @rafenden there is a component for this: https://github.com/uktrade/data-hub-components/blob/master/src/new-window-link/NewWindowLink.jsx

src/apps/companies/apps/referrals/send-referral/client/SendReferralForm.jsx
```diff
@@ -90,8 +91,13 @@ const SendReferralForm = ({
         <HintText>
           This can be a person in Post, a sector team or an ITA. If you are not
           sure who this might be, you can{' '}
-          <a href="https://people.trade.gov.uk/teams/department-for-international-trade">
-            find the relevant teams and people on Digital Workspace
+          <a
```
web-bert: Same here: https://github.com/uktrade/data-hub-components/blob/master/src/new-window-link/NewWindowLink.jsx

test/functional/cypress/specs/companies/referrals/referral-help-spec.js
```diff
@@ -62,5 +62,10 @@ describe('Referral help', () => {
         )
         .should('have.text', 'Go back to the referral')
     })
+    it("should link to digital workspace's people finder", () => {
+      cy.contains('Digital Workspace')
+        .should('have.attr', 'href', urls.external.digitalWorkspace.teams)
+        .should('have.attr', 'aria-label', 'opens in a new tab')
```
web-bert: I removed this assertion when switching to the NewWindowLink

src/apps/companies/apps/referrals/help/client/ReferralHelp.jsx
```diff
@@ -50,10 +50,14 @@ export default connect(state2props)(
...
+                  href={urls.external.digitalWorkspace.teams}
+                  target="_blank"
+                  aria-label="opens in a new tab"
```
rafenden: I agree with @web-bert on this, we could also update the `NewWindowLink` component to use lowercase message, i.e.
from `Opens in a new window` to `opens in a new tab`

debitan: I agree that the `NewWindowLink` component needs updating. If I recall correctly, I didn't use it in a previous PR because of its `Opens in a new window` message.

src/apps/companies/apps/referrals/send-referral/client/SendReferralForm.jsx
debitan: what is the difference between using `<a>` and `<Link>`?

src/apps/companies/apps/referrals/help/client/ReferralHelp.jsx
mforner13: I agree, I'll make a small PR in the component library to fix that and then will update this PR.

web-bert:
@debitan why should it be called a `NewTabLink`? This assumes that all user agents will render `target="_blank"` into a new tab which may not be the behaviour (and can be controlled by the user), so "Opens in new window" covers both scenarios  - as users of tabs should understand that it may open in a new tab. If we want to cover all it should be "Opens in a new window or tab" however you could even question the use of this at all... https://css-tricks.com/use-target_blank/

debitan:
> @debitan why should it be called a `NewTabLink`? This assumes that all user agents will render `target="_blank"` into a new tab which may not be the behaviour (and can be controlled by the user), so "Opens in new window" covers both scenarios - as users of tabs should understand that it may open in a new tab. If we want to cover all it should be "Opens in a new window or tab" however you could even question the use of this at all... https://css-tricks.com/use-target_blank/

@web-bert 
1. you're supposed to be on holiday today, so why are you here? :-p 
2. because that is the behaviour that we are aiming for, where a click opens it in a new tab rather than redirecting from the current page. I'm actually in favour of not showing the help text `opens in a new tab` afterwards and @rafenden @mforner13 and I were just discussing this in slack. Of course the user can override the behaviour, but I think the component should be called `NewTabLink` as that is the default behaviour enforced by it.

mforner13:
As a result of bumping the component library version, I have also updated three functional tests in companies to expect the new help text on the NewWindowLink component.

PR #2565, opened 2020-03-24
Update css spec to remove warning when starting the app

debitan:
We recently updated `postcss-loader` and as a result we were getting the below warning in the console when starting the app. I have therefore updated our css in line with the warning and now no warning shows. 

Some of the extra changes, like using single quotes are from prettier. 

Screenshot 2020-03-24 at 10 05 26

### DIFF
--- a/assets/stylesheets/elements/_base.scss
+++ b/assets/stylesheets/elements/_base.scss
@@ -1,6 +1,6 @@
-@import "../settings/colours";
-@import "../settings/typography";
-@import "../tools/mixins/addons";
+@import '../settings/colours';
+@import '../settings/typography';
+@import '../tools/mixins/addons';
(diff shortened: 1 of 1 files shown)


assets/stylesheets/elements/_base.scss
```diff
@@ -9,7 +9,7 @@ html {
 
 a {
   color: $link-colour;
-  text-decoration-skip: ink;
+  text-decoration-skip-ink: auto;
```
rachaelcodes: Quick question: Which browsers are we supporting? Can I Use suggests that Safari (both desktop and mobile) aren't yet using `text-decoration-skip-ink`: https://caniuse.com/#search=text-decoration-skip .

PR #2568, opened 2020-03-26
Rebuild edit company export countries form in React

web-bert:
## Description of change

As the title says

## Test instructions

Go to a company export tab and click "Edit export countries" the Typeahead should now be the React one

### Before
Screenshot 2020-03-26 at 13 42 41

### After
Screenshot 2020-03-26 at 13 43 15

## Checklist


test/sandbox/routes/v4/company/company.js
```diff
@@ -202,3 +202,15 @@ exports.referralDetails = function(req, res) {
   }
   return res.json(referralDetails)
 }
+
+exports.exportDetail = function(req, res) {
+  var companyId = req.params.companyId
+
+  if (companyId === companyLambdaPlc.id) {
+    res.status(500).send('')
```
ian-leggett: If there is a 500 server error do we need to send back an empty string?

test/functional/cypress/support/commands.js
```diff
@@ -5,3 +5,16 @@ Cypress.Commands.add('selectFirstOption', (selector) => {
     .parent()
     .trigger('change')
 })
+
+Cypress.Commands.add(
```
ian-leggett: I know there has been lots of talk about wether to use commands, is this ok?

src/apps/companies/apps/exports/client/ExportCoutriesEdit.jsx
```diff
@@ -0,0 +1,105 @@
...
+function saveExportCountries(companyId, setApiError) {
+  return async (values) => {
+    try {
+      await axios.patch(`/api-proxy/v4/company/${companyId}/export-detail`, {
+        export_countries: getExportCountries(convertFieldValues(values)) || [],
+      })
+
+      return urls.companies.exports.index(companyId)
+    } catch (e) {
+      const is400 = e?.response?.status === 400
```
ian-leggett: I wondering if we should do optional chaining here, if we do get an error back then there will be a status.

test/sandbox/routes/v4/company/company.js
web-bert: You have to call `send` so that it sends the response and according to the docs
> The body parameter can be a Buffer object, a String, an object, or an Array

So I think the answer is yes

test/functional/cypress/support/commands.js
web-bert: I believe we agreed it is ok now and without the `dh` prefix

src/apps/companies/apps/exports/client/ExportCoutriesEdit.jsx
web-bert: I wasn't sure what errors would come back here so played it safe, happy to remove

ian-leggett: No need just thought I would see if it was needed.

test/selectors/company/export.js
```diff
@@ -4,9 +4,9 @@ module.exports = {
   winCategory: '#export_experience_category',
   countries: {
     listItemHeadings: '#company-export-full-history > div > div > h3',
-    future: '#group-field-' + EXPORT_INTEREST_STATUS.FUTURE_INTEREST,
-    export: '#group-field-' + EXPORT_INTEREST_STATUS.EXPORTING_TO,
-    noInterest: '#group-field-' + EXPORT_INTEREST_STATUS.NOT_INTERESTED,
+    future: '#field-' + EXPORT_INTEREST_STATUS.FUTURE_INTEREST,
+    export: '#field-' + EXPORT_INTEREST_STATUS.EXPORTING_TO,
+    noInterest: '#field-' + EXPORT_INTEREST_STATUS.NOT_INTERESTED,
```
mforner13: I couldn't find anywhere these selectors are used anymore - do you still need them?

web-bert: The are used by the e2e tests here: https://github.com/uktrade/data-hub-frontend/pull/2568/files?file-filters%5B%5D=.js&file-filters%5B%5D=.json#diff-e2bc8c2464f602d61eee328ff5d5a2f8R195

src/apps/companies/apps/exports/controllers.js
```diff
@@ -138,23 +138,54 @@ function renderExportEdit(req, res) {
...
+  return { value: country.id, label: country.name }
+}
+
+function getCountryValues(countries) {
+  if (countries) {
+    return countries.map(createCountryOption)
+  }
+
+  return []
+}
```
peterhudec: * I would prefer the `createCountryOption` to be inlined as it's not doing much, it's only used once and you wouldn't need to come up with an awkward name for it
* You can get rid of the if statement by using default value for the argument
* I would name it `countryToTypeaheadOption` as that seems to be the function's purpose
```js
const getCountryValues = (countries = []) =>
  countries.map(({ id, name }) =>
    ({ value: id, label: name })
```

src/apps/companies/apps/exports/controllers.js
```diff
@@ -138,23 +138,54 @@ function renderExportEdit(req, res) {
...
+            name: FUTURE_INTEREST,
+            label: exportDetailsLabels.futureInterestCountries,
+            values: getCountryValues(exportCountries[FUTURE_INTEREST]),
+          },
+          {
+            name: NOT_INTERESTED,
+            label: exportDetailsLabels.noInterestCountries,
+            values: getCountryValues(exportCountries[NOT_INTERESTED]),
+          },
+        ],
```
peterhudec: The `ExportCountriesEdit` signature is unnecessarily convoluted:
* The `fields` prop is set to a hardcoded list, so you could just as well, and preferrably, allocate a separate prop for each of its items
* The `name` and `label` are internal details of `ExportCountriesEdit`, moreover the `label` belongs to the presentation layer

```jsx
<ExportCountriesEdit
  // These are fine
  companyId={/* ... */}
  countryOptions={/* ... */}
  // To me it would be much more readable this way
  exportingTo={getCountryValues(exportCountries[EXPORTING_TO])}
  futureInterest={getCountryValues(exportCountries[FUTURE_INTEREST])}
  notInterested={getCountryValues(exportCountries[NOT_INTERESTED])}
/>
```

src/apps/companies/apps/exports/views/edit-countries.njk
```diff
@@ -6,32 +6,10 @@
...
 
-  {% call Form({
-    errors: errors,
-    buttonText: 'Save and return',
-    returnText: 'Return without saving',
-    returnLink: urls.companies.exports.index(company.id),
-    hiddenFields: {
-      id: company.id
+  {% component 'react-slot', {
+      id: 'company-exports-edit-countries',
```
peterhudec: The word order could be more consistent e.g. `ExportCountriesEdit`

test/end-to-end/cypress/specs/DIT/companies-spec.js
```diff
@@ -129,7 +129,7 @@ describe('Export', () => {
           cy.visit(
             companies.exports.editCountries(fixtures.company.lambdaPlc.id)
           )
-          cy.get(selectors.companyForm.save).click()
+          cy.contains('button', 'Save and return').click()
```
peterhudec: Nice trick combining the content with the selector. Didn't know about it 👍

test/functional/cypress/specs/companies/export/edit-countries-spec.js
```diff
@@ -0,0 +1,184 @@
...
+
+const lambdaPlc = require('../../../../../sandbox/fixtures/v4/company/company-lambda-plc.json')
+const dnbCorp = require('../../../../../sandbox/fixtures/v4/company/company-dnb-corp.json')
+const oneListCorp = require('../../../../../end-to-end/cypress/fixtures/company/one-list-corp.json')
+const metadataCountries = require('../../../../../sandbox/fixtures/v4/metadata/country.json')
+
+const selectors = {
+  currentlyExporting: '#field-currently_exporting',
+  futureInterest: '#field-future_interest',
+  notInterested: '#field-not_interested',
```
peterhudec: We should avoid using HTML ids for various reasons

src/apps/companies/apps/exports/controllers.js
web-bert: As discussed, this is the MVP of this feature, ideally there will be a generic `ExportCountries` component that will be used between interactions and the export tab, so these concerns will be fixed as part of future work

test/functional/cypress/specs/companies/export/edit-countries-spec.js
web-bert: These are generated by the `<FieldTypeahed>`

web-bert:
@peterhudec comments have been addressed

src/apps/companies/apps/exports/client/ExportCountriesEdit/tasks.js
```diff
@@ -0,0 +1,37 @@
...
+      if (nonFieldMessages?.length) {
+        return Promise.reject({
+          message: { [API_ERROR]: nonFieldMessages.join(', ') },
+        })
+      } else {
+        return Promise.reject({ message: { [API_WARN]: e.message } })
+      }
+    })
+    .then((response) =>
+      response?.API_ERROR || response?.API_WARN
```
peterhudec, line 33: You sure you didn't wanna use these as constants?

web-bert: Good spot, yes I do!

src/client/components/Task/index.jsx
```diff
@@ -110,8 +110,8 @@ const Task = connect(
         onSuccessDispatch,
       }),
   })
-)(({ start, children, ...props }) =>
-  children((name, id) => {
+)(function Task({ start, children, ...props }) {
+  return children((name, id) => {
```
peterhudec, line 114: Why is it necessary to declare a named function?

web-bert: Without this name the name in the component tree in dev tools is Anonymous - which is not very helpful

test/functional/cypress/specs/companies/export/edit-countries-spec.js
peterhudec: I see. That's even worse though. Because there should only ever be one id in a HTML page and you already have three of them.

src/apps/companies/apps/exports/client/ExportCountriesEdit/tasks.js
```diff
@@ -0,0 +1,37 @@
...
+  return countryFields
+}
+
+export function saveExportCountries({ values, companyId }) {
+  return axios
+    .patch(`/api-proxy/v4/company/${companyId}/export-detail`, {
+      export_countries: getExportCountries(transformFieldValues(values)) || [],
+    })
+    .catch((e) => {
+      console.log('catch', e)
```
peterhudec, line 21: Forgotten log

src/apps/companies/apps/exports/client/ExportCountriesEdit/index.jsx
```diff
@@ -0,0 +1,115 @@
...
+  return (
+    <Task>
+      {(getTask) => (
+        <FormStateful
+          onSubmit={(values) => {
+            getTask(TASK_NAME, TASK_ID).start({
+              payload: { values, companyId },
+              onSuccessDispatch: EXPORT_COUNTRIES_EDIT__SAVE,
+            })
+            return null
```
peterhudec, line 77: You don't need to explicitly return here, a function by default returns `undefined` which will do the same job as `null` in the form.

test/functional/cypress/specs/companies/export/edit-countries-spec.js
web-bert: You can have many id's they just have to be unique across the whole document? These are generated with the `name` of the `<input>` so they will always be unique unless we have multiple `<input>`s with the same `name`?

PR #2587, opened 2020-03-31
Fix styles on dashboard company list button

ian-leggett:
## Description of change
The buttons in the company lists inside the dashboard are inheriting an active state CSS rule from a deprecated CSS file called `_dashboard.scss` which lives in `/assets/stylesheets/_deprecated/_dashboard.scss`. 

See below:
Screenshot 2020-03-31 at 16 30 29

We didn't notice this before as we hadn't seen any active states until after it was released. The solution was to remove the class that was on a containing div called `dashboard-section` so we don't inherit these style rules from the old deprecated file. After doing this I needed to style the font-sizes in the referral list cards as they were also inheriting rules from the old file.

 
## Test instructions

No active style states should show in the company lists buttons/links. Also referrals should look like it did before.

## Screenshots
Screenshot 2020-03-31 at 16 30 29

### After

Screenshot 2020-03-31 at 16 38 29
Screenshot 2020-03-31 at 16 39 34

## Checklist


cypress.json
```diff
@@ -1,6 +1,6 @@
 {
   "integrationFolder": "test/functional/cypress/specs",
-  "baseUrl": "http://localhost:3001",
+  "baseUrl": "http://localhost:3000",
```
web-bert, line 3: is this intentional?

src/client/components/Dashboard.jsx
```diff
@@ -1,26 +1,35 @@
...
+import styled from 'styled-components'
+import { GREY_2 } from 'govuk-colours'
 
 import CompanyLists from './CompanyLists'
 import ReferralList from './ReferralList'
 import TabNav from './TabNav'
 
+const StyledDiv = styled('div')`
+  border-top: 4px solid ${GREY_2};
+  padding-top: 16px;
```
web-bert, line 12: Should this be using `SPACING`?

cypress.json
ian-leggett: Good spot, sorry this was a mistake. I have put this back 👍

src/client/components/Dashboard.jsx
ian-leggett: I did start to use SPACING but then realised that the spacing from the container div and class I removed ie `dashboard-section` had a padding of 16px and this does not fall into the SPACING units available in constants. Maybe the dashboard section which uses legacy styles needs updating so both the dashboard section and the new React dashboard can use SPACING.

web-bert: Is a 1px difference going to be noticable?

PR #2595, opened 2020-04-02
Merge code coverage reports

PippoRaimondi:
## Description of change

The intent of this PR is to merge all functional reports together (from different circleci instances) and finally merge both functional and unit code coverage reports.

## Test instructions

CI should store the merged code coverage reports.

## Checklist


.circleci/config.yml
```diff
@@ -440,6 +450,30 @@ jobs:
...
       - *store_artifacts_cypress_video
 
+  merge-and-publish-coverage:
+    docker:
+      - <<: *docker_data_hub_base
+    steps:
+      - checkout
+      - attach_workspace:
+          at: ~/
+      - run: yarn global add nyc
```
web-bert, line 460: It seems either this is not needed or the lines below that are using `npx` don't need to be using `npx`?

PippoRaimondi: I am installing a package globally (instead of restoring npm package cache as it will take longer than installing this single package) and using it with npx. Alternatively I can install it locally and use `./node/.bin/nyc` or create a script for it in packages, just felt the npx was the least overhead option. Let me know if you prefer the alternatives I gave above.

PippoRaimondi:
> Just one comment, also does this mean that the codecov check can use the output of this to accurately report the code coverage for each PR?

Yes, the idea is that if the if functionals are successful then we should reliably capture both unit and functional codecov reports and it should give consistent results.

.circleci/config.yml
web-bert: Given that we know `nyc` has been installed globally you should not need to use `npx` to invoke it, have you tried removing the `npx` prefix from the commands below - I would expect them to work without `npx` or `./node_modules/.bin` as a prefix?

PippoRaimondi: hmmm, I haven't... I just assumed it would be needed... whoopsie... let me give it a go!

PippoRaimondi: @web-bert you were right like usual, thanks for pointing it out, fixed! =D

PR #2605, opened 2020-04-07
Fix visual test workflow

rachaelcodes:
Brings visual test docker files up to date with
frontend docker build process.

## Description of change

Recent changes to the frontend Docker build process means that dockerize is no longer being used. This updates the `docker-visual-regression.yml` to bring it in line with the new frontend Docker approach. Environment variables have been removed to a `.env` file.

## Test instructions

Make a copy of `visual-stack.sample.env` to `visual-stack.env` and add the correct environment variables. Then run the visual tests as before. 

## Checklist

- ~~[ ] Manual compatibility testing (Browsers: Chrome, Firefox, IE11, Safari)~~


.bin.sample/visual
```diff
@@ -1,3 +1,3 @@
-BROWSERSTACK_USERNAME=user
-BROWSERSTACK_ACCESS_KEY=key
+export BROWSERSTACK_USERNAME=user
+export BROWSERSTACK_ACCESS_KEY=key
```
web-bert, line 2: These env variables should not need the `export` as we don't want them hanging around when the script is done

rachaelcodes: Unfortunately, without including `export`, I'm getting this error: `Error: A "user" or "key" was provided but could not be connected to a known cloud service (SauceLabs, Browerstack or Testingbot).` Is there another way to add the env variables without it?

web-bert: I use [direnv](https://direnv.net/) to make the entries in my `.env` file available so I think it's already available to this script. Also, these scripts are only meant to be a sample and should be copied to a local version that you can modify to suit your local setup. Again using direnv I have all my helper scripts in a `.bin` folder and they are automatically added to my PATH when I `cd` into my directory, so this may also sort out the env variables for me too - it's worth taking a look at.

This is the only sample script that includes env variables, so maybe its best to have the `export` there if it works across more setups 👍

PR #2606, opened 2020-04-08
React router

peterhudec:
## Description of change

Adds [`react-router`](https://reacttraining.com/react-router/web/guides/quick-start) infrastructure.

## Test instructions

The PR should have no effect on how the application works.


src/middleware/spa-fallback.js
```diff
@@ -0,0 +1,37 @@
+/**
+ * An express middleware which adds `basePath` to `res.locals`.
+ * It must be applied in front of a request handler of a route
+ * which must have a trailing asterisk e.g. `/foo/bar/*`.
+ * You should use the `spaFallbackSpread` instead, which prevents the accidental
+ * omission of adding the asterisk.
+ */
+const spaFallback = (req, res, next) => {
```
rafenden, line 8: How are you planning to use this middle?

peterhudec: You should actually use the other one, the `spaFallbackSpread`, which uses this one, but it should be as follows:

Imagine you have a `/foo/bar` express route and you want to use react router for the rest of the path e.g. `/foo/bar/baz/bing` where `/baz/bing` are managed by react router. For this to work, react router needs to know that `/foo/bar` is the [`basename`](https://reacttraining.com/react-router/web/api/BrowserRouter/basename-string). To make this work, you need to add a trailing `*` to the route and use the `spaFallback` middleware in front of the controller. The `spaFallbackSpread` combines these two steps into a single one:

```js
// These two are equivalent
app.get('/foo/bar*', spaFallback, controller)
app.get(...spaFallbackSpread('/foo/bar'), controller)
```

This will result in a `<base href="/foo/bar"/>` tag to appear in the document's head, whose `href` value is then fed into react router's [`basename`](https://reacttraining.com/react-router/web/api/BrowserRouter/basename-string).

PR #2608, opened 2020-04-09
Recreate local header in React

LaurenQurashi:
## Description of change

This PR adds new LocalHeader and Body React components, to replace the old Nunjucks local header.
We've made a new Nunjucks template from which future pages which only use React can extend. This is a replica of the current template, just with the local header block removed. 

Currently, the LocalHeader contains a heading and breadcrumbs only, but when the more complex headers need converting to React, this component can be used as a base. 

The Body component replaces the styling and formatting brought in by the body_main_content Nunjucks block. When using the LocalHeader component, you will need to wrap all the content below it in the Body component in order to keep styling consistent. 

This PR applies the LocalHeader and Body components to the send referral form and confirmation pages only. 

We have also removed the grey background colour on the global html element as it didn't seem to be used on any pages.

## Test instructions

The tests have not been affected by our changes, meaning you shouldn't see any changes at all to the current test suit. 


test/functional/cypress/specs/companies/referrals/send-referral-spec.js
```diff
@@ -109,17 +113,20 @@ describe('Send a referral form', () => {
       'when "Continue" button is clicked without chosing an adviser',
       () => {
         it('should display error message', () => {
-          cy.get(selectors.companySendReferral.subjectField)
+          cy.get('label')
+            .eq(1)
```
web-bert, line 117: These might be better to get them via the label content so it's easier to follow, something like `cy.contains('label', 'The label text').find('input')...`

src/apps/companies/apps/referrals/send-referral/client/SendReferralConfirmation.jsx
```diff
@@ -6,8 +6,12 @@ import { H4, Button, Link } from 'govuk-react'
 import UnorderedList from '@govuk-react/unordered-list'
 import ListItem from '@govuk-react/list-item'
 import SecondaryButton from '../../../../../../client/components/SecondaryButton'
+import LocalHeader from '../../../../../../client/components/LocalHeader.jsx'
```
rafenden, line 9: We could remove `.jsx` file extension in imports as they are not required, and we don't include them elswhere.

src/client/components/LocalHeader.jsx
```diff
@@ -0,0 +1,59 @@
...
+    padding-bottom: ${SPACING.SCALE_5};
+    background-color: ${GREY_4};
+    padding-top: ${SPACING.SCALE_3};
+  }
+`
+
+const LocalHeader = ({ breadcrumbs, heading }) => {
+  const breadcrumbItems = breadcrumbs.map((breadcrumb) =>
+    breadcrumb.link ? (
+      <Breadcrumbs.Link key={breadcrumb.text} href={breadcrumb.link}>
```
rafenden, line 32: I would suggest to use `link` here instead of `text` are there might be cases when `text` is duplicated.

test/end-to-end/cypress/specs/DIT/referrals-spec.js
```diff
@@ -22,14 +22,18 @@ describe('Referrals', () => {
...
         .type('Example subject')
-        .get(selectors.companySendReferral.notesField)
+      cy.get('label')
+        .eq(2)
+        .find('textarea')
         .click()
         .type('Example notes')
-        .get(selectors.companySendReferral.continueButton)
+      cy.get('button')
+        .contains('Continue')
```
rafenden, line 36: Could be simplified with:
```
cy.contains('button', 'Continue').click()
```

test/functional/cypress/specs/companies/referrals/send-referral-spec.js
```diff
@@ -42,50 +42,51 @@ describe('Send a referral form', () => {
...
-        cy.get(selectors.companySendReferral.form)
-          .find('h4')
-          .should('contain', 'Who do you want to refer this company to?')
-          .next()
+        cy.contains('Who do you want to refer this company to?').should(
+          'be.visible'
+        )
+        cy.get('label')
+          .eq(0)
           .should('contain', 'Adviser')
```
rafenden, line 50: Could be simplified with:
```suggestion
        cy.contains('label', 'Adviser')
```

There is a few more tests that can be simplified with this method.

PR #2610, opened 2020-04-09
Add filter to ReferralList

peterhudec:
## Description of change

Adds a dropdown to _My referrals_ by which you can choose whether all, sent or received referrals will be listed.

## Test instructions

1. Go to the homepage
2. Select the _My referrals_ tab
3. The _All referrals_ option of the _View_ labeled dropdown should be selected by default and all referrals should be displayed in the list
4. Select one fo the other options of the dropdown and verify that the list reflects the selected filter
5. Repeat steps 3 and 4 for the remaining dropdown option

### Before
Screen Shot 2020-04-09 at 2 49 57 PM

### After
Screen Shot 2020-04-09 at 2 53 38 PM


src/client/components/ReferralList/index.jsx
```diff
@@ -34,14 +58,31 @@ export default multiInstance({
         referrals && (
           <ContentWithHeading
             heading={pluralize('Referral', referrals.length, true)}
```
mforner13, line 60: Should the number of referrals it displays here change to reflect how many are in the filtered list?

peterhudec:
All very good points @mforner13 I'll discuss them with Rogier.

src/client/components/ReferralList/index.jsx
```diff
@@ -34,14 +58,31 @@ export default multiInstance({
...
+            headingActions={
+              <StyledLabel>
+                View
+                <StyledSelectInput
+                  value={filter}
+                  onChange={(e) => onFilterChange(e.target.value)}
+                >
+                  <option value="all">All referrals</option>
```
rafenden, line 68: We should use a constant here as `all` and other values are referred across files.

package.json
```diff
@@ -58,6 +58,7 @@
   },
   "dependencies": {
     "@babel/core": "^7.9.0",
+    "@babel/helper-call-delegate": "^7.8.7",
```
rafenden, line 61: Why we need this package?

There are no Babel docs for this module.

peterhudec: My tests were failing on it not being installed. Installing it fixed the problem. But it seems to be gone now, probably by rebasing with master.

PR #2617, opened 2020-04-14
Feature/routed dashboard

peterhudec:
## Description of change

The dashboard tab navigation is now _routed_ i.e. it writes to and reads from the browser's address bar. 

## Test instructions

1. Open the homepage
2. The _My companies lists_ tab should be selected by default
3. Click the _My referrals_ tab
2. The path in the browser address bar should now be `/my-referrals`
3. Refresh the page with the `/my-referrals` path
4. After the page reloads, the _My referrals_ tab should be selected
5. Click on the _My companies lists_ tab
6. The path in the address barr should now read `/`


src/apps/dashboard/index.js
```diff
@@ -1,7 +1,11 @@
 const router = require('express').Router()
 const urls = require('../../lib/urls')
+const { spaFallbackSpread } = require('../../middleware/spa-fallback')
 const { renderDashboard } = require('./controllers')
 
 module.exports = {
-  router: router.get(urls.dashboard.route, renderDashboard),
+  router: router.get(
+    ...spaFallbackSpread(urls.dashboard.route),
```
rafenden, line 8: Did you explore `next.js` as I think this library is trying to achieve the same?

src/client/components/TabNav/index.jsx
```diff
@@ -70,87 +71,108 @@ const TabNav = ({
...
+                        ? // ...only the first tab participates in the tabindex
+                          0
+                        : // Otherwise, only the selected tab participates in tabindex
+                        selected
+                        ? 0
+                        : -1
+                    }
+                    onClick={() =>
+                      selected || routed
+                        ? (history.push('/' + key), onFocusChange(index))
```
rafenden, line 131: Isn't `history.push` considered a side effect?

peterhudec: Good point. If we didn't use [connected-router](https://github.com/supasate/connected-react-router) then it would, but as we are using it, `history.push()` should result in dispatching a router specific action. There is a router redux middleware hooked up, which is subscribed to that action and it's there where the side effect happens.

src/apps/dashboard/index.js
peterhudec: I'm not familiar with `next.js`, but from reading their [routing docs](https://nextjs.org/docs/routing/introduction) it looks like they have a very high level, folder structure based abstraction where you don't deal with Express routes directly. Did you mean something particular, can you post a link?

rafenden: I don't have anything specific on mind, just wanted to point out similarity of both approaches as we might be missing something.

PR #2633, opened 2020-04-22
Multiinstance action matching

peterhudec:
## Description of change

This PR fixes a bug when due to the fact, that in Redux each action passes through every reducer, states of _multiinstance_ components leaked to each other:

```js
const MultiinstanceFoo = multiinstance({
  name: 'Foo',
  reducer: fooReducer, // Subscribes to actions FOO__X and FOO__Y
  component Foo,
})

const MultiinstanceBar = multiinstance({
  name: 'Bar',
  reducer: fooReducer, // Subscribes to actions BAR__X and BAR__Y
  component Foo,
})

const store = createStore(combineReducers({
  ...MultiinstanceFoo.reducerSpread,
  ...MultiinstanceBar.reducerSpread,
}))

const App = () =>
  <ReduxProvider store={store}>
    <Foo id="a" />
    <Bar id="b" />
  </ReduxProvider>

// Now after a Foo dispatches the FOO__X action, the state will look like this
let state = {
  Foo: {
    a: 'Current Foo state',
  },
  Bar: {
    // Here the Foo state leaked to Bar
    a: 'Current Foo state',
  },
}

// And if Bar dispatches BAR__Y, it'll look like this
state = {
  Foo: {
    a: 'Current Foo state',
    // Now also Bar state leaked to Foo
    b: 'Current Bar state',
  },
  Bar: {
    a: 'Current Foo state',
    b: 'Current Bar state',
  },
}
```
This PR adds a required parameter `actionPattern` to `reducerDecorator` and thus also to `multiInstance`. The `actionPattern` should specify action types the _wrapped_ reducer will subscribe to. It can be a regex, a list of action types, or a string in which case it will subscribe to all action types starting with the string, which should work well with how we prefix our action types:

```js
const MultiinstanceFoo = multiinstance({
  name: 'Foo',
  actionPattern: ['FOO__X', 'FOO__Y'],
  reducer: fooReducer,
  component Foo,
})

const MultiinstanceBar = multiinstance({
  name: 'Bar',
  actionPattern: 'BAR__',
  reducer: fooReducer,
  component Foo,
})
```


peterhudec:
@debitan, the referral stuff accidentally leaked here, thanks for pointing it out.

PR #2635, opened 2020-04-22
Add contact loop to the interaction form

rafenden:
## Description of change

Add contact look to the interaction form allowing users for uninterrupted journey.

## Test instructions

1. Go to [the new interaction form](https://data-hub-fro-contact-lo-cyaipi.herokuapp.com/interactions/create?company=5bb239e4-b227-4a84-bc2e-3420fc129447)
2. Chose the interaction type
3. Fill some fields
4. Click `add a new contact` from the link within the `Contact(s)` field
5. Create new contact
6. You should be redirected to the interaction form with all the form values restored


src/apps/interactions/apps/add-interaction/client/tasks.js
```diff
@@ -48,6 +49,22 @@ function transformServiceAnswers(values) {
...
 
+export function openContactForm({ values, url }) {
+  window.localStorage.setItem(
+    STORE_ID,
+    JSON.stringify({
+      values,
+      currentStep: 1,
+    })
+  )
+  window.location.href = url
```
paulgain, line 60: Isn't this classified as a saga side-effect and should be handled by saga generator function?

rafenden: It is handled via saga generator created within the `Task` component.

paulgain: 👍

PR #2636, opened 2020-04-24
Refactor LocalHeader component to take children rather than props

debitan:
## Description of change

This PR refactors the LocalHeader component to take children rather than props. This allows the component to be more flexible in anticipation of its use on more complex pages. 

The reason for this is because I need to add a new action dropdown as per the screenshot below on the company page header. This is all currently rewritten in nunjucks and this PR is the first part of refactoring this to React. 

image

## Test instructions

Nothing should change. On navigating to the send referral page and the send referral confirmation page you should see the local header as before with no functional or styling changes. 

## Checklist

### DIFF
--- a/src/apps/companies/apps/referrals/send-referral/client/SendReferralConfirmation.jsx
+++ b/src/apps/companies/apps/referrals/send-referral/client/SendReferralConfirmation.jsx
@@ -5,6 +5,7 @@ import { SummaryTable, FormActions } from 'data-hub-components'
 import { H4, Button, Link, Main } from 'govuk-react'
 import UnorderedList from '@govuk-react/unordered-list'
 import ListItem from '@govuk-react/list-item'
+
 import SecondaryButton from '../../../../../../client/components/SecondaryButton'
 import LocalHeader from '../../../../../../client/components/LocalHeader'
 import { companies, dashboard } from '../../../../../../lib/urls'
--- a/src/client/components/LocalHeader.jsx
+++ b/src/client/components/LocalHeader.jsx
@@ -1,59 +1,66 @@
 import React from 'react'
-import Breadcrumbs from '@govuk-react/breadcrumbs'
 import PropTypes from 'prop-types'
 import styled from 'styled-components'
-import { typography } from '@govuk-react/lib'
-import { SPACING } from '@govuk-react/constants'
 import { GREY_4 } from 'govuk-colours'
-
-const StyledBreadcrumbs = styled(Breadcrumbs)`
-   {
-    margin-bottom: ${SPACING.SCALE_5};
-    margin-top: 0;
-  }
-`
-const StyledH1 = styled('h1')`
-   {
-    ${typography.font({ size: 36, weight: 'bold' })};
-  }
-`
+import { SPACING } from '@govuk-react/constants'
+import Main from '@govuk-react/main'
+import Breadcrumbs from '@govuk-react/breadcrumbs'
+import { typography } from '@govuk-react/lib'
 
 const StyledHeader = styled('header')`
-   {
-    padding-bottom: ${SPACING.SCALE_5};
-    background-color: ${GREY_4};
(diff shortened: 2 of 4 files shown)


src/apps/companies/apps/referrals/send-referral/client/SendReferralConfirmation.jsx
```diff
@@ -22,15 +26,19 @@ const SendReferralConfirmation = ({
...
+          <Breadcrumbs.Link href={companies.index()}>
+            Companies
+          </Breadcrumbs.Link>
+          <Breadcrumbs.Link href={companies.detail(companyId)}>
+            {companyName}
+          </Breadcrumbs.Link>
+          Send a referral
+        </BreadcrumbsWrapper>
+        <LocalHeaderHeading>Send a referral</LocalHeaderHeading>
+      </LocalHeader>
```
debitan, line 41: This is more verbose than the previous implementation. 

One idea to combat this would be to build wrappers that cater for different requirements. This would mean we could implement a wrapper with syntax similar to the old implementation, while still retaining the ability to build more complex or unique designs by dealing directly with the children. However, this is a future consideration for when this component is used more widely and isn't part of this PR.

src/client/components/LocalHeader/LocalHeaderHeading.jsx
```diff
@@ -0,0 +1,8 @@
+import styled from 'styled-components'
+import { typography } from '@govuk-react/lib'
+
+const LocalHeaderHeading = styled('h1')`
+  ${typography.font({ size: 36, weight: 'bold' })};
```
debitan, line 5: I'm not actually sure about whether this is the best syntax for this, it's not a style I've seen before. However, this is what was already there in the old LocalHeader component.

src/client/index.jsx
```diff
@@ -254,9 +253,6 @@ function App() {
         <Mount selector="#company-export-countries-edit">
           {(props) => <ExportCountriesEdit {...props} />}
         </Mount>
-        <Mount selector="#local-header">
-          {(props) => <LocalHeader {...props} />}
-        </Mount>
```
debitan, line 259: There was no need to add the LocalHeader to the index file as it isn't used through the React slot mechanism, but rather imported directly to other jsx component files.

src/apps/companies/apps/referrals/send-referral/client/SendReferralConfirmation.jsx
web-bert: This is way too verbose for me, given this is on so many pages, I think `BreadcrumbsWrapper` should just be `Breadcrumbs` which takes a prop of `items` which is the same as the previous implementation. We need something easy to use, we could even leave the `breadcrumbs` prop on `LocalHeader` as it will always be the first item in the header and the `children` can come after them - it should be an optional prop so you can remove them completely.

debitan: Cool. Yeah I think that is a good idea. I'll add the breadcrumbs to the `LocalHeader`.

src/client/components/LocalHeader.jsx
```diff
@@ -1,59 +1,66 @@
...
-}
+const StyledH1 = styled('h1')`
+  ${typography.font({ size: 36, weight: 'bold' })};
+`
+
+const LocalHeader = ({ breadcrumbs, heading, children }) => (
+  <StyledHeader aria-label="local header" data-auto-id="localHeader">
+    <StyledMain>
+      <BreadcrumbsWrapper>
+        {breadcrumbs?.map((breadcrumb) =>
```
web-bert, line 33: Optional: Given that `breadcrumbs` is required you can drop the `?.map` for just `.map`, otherwise there should be a `breadcrumbs?.length && ` in front of `<BreadcrumbsWrapper>`

PR #2641, opened 2020-04-27
Only show project value field for projects landing before 01/04/2020

mforner13:
## Description of change

Due to the way DIT measures the value of an investment project changing from 01/04/2020 onwards, the 'Project Value' field is no longer required for projects with an estimated or actual land date after this date. 

This PR adds a check in the edit-value controller to determine whether the field should be displayed or not, and passes a boolean to the 'edit-value' template. If the project is both FDI and has an actual or estimated land date before 01/04/2020, the field will be displayed. 

**Future work**
Once the investment verification team has carried out all checks on projects which landed before 01/04/2020, this field can be removed completely from the front end. 

## Test instructions

1. Go to an investment project with an estimated or actual land date before 01/04/200 (can use the filters on the investments tab)
2. Click 'Edit/Add value'
3. Check that the project value field appears

1. Go to a project with an estimated or actual land date after 01/04/2020
2. Click 'Edit/Add value'
3. Check that the project value field doe not appear

### Before

Screenshot 2020-04-27 at 16 58 53

### After

Screenshot 2020-04-27 at 16 59 40

## Checklist

### DIFF
--- a/src/apps/investments/controllers/edit.js
+++ b/src/apps/investments/controllers/edit.js
@@ -1,9 +1,21 @@
+// For projects landing after 01/04/2020, the project value field is not needed
+const valueCutOffDate = new Date('2020-04-01')
+
 function editDetailsGet(req, res) {
   res.breadcrumb('Edit details').render('investments/views/details-edit')
 }
 
 function editValueGet(req, res) {
-  res.breadcrumb('Edit value').render('investments/views/value-edit')
+  const {
+    estimated_land_date: estimatedLandDate,
+    actual_land_date: actualLandDate,
+  } = res.locals.form.state
+
+  res.breadcrumb('Edit value').render('investments/views/value-edit', {
+    projectValueNeeded:
+      (estimatedLandDate || actualLandDate) &&
+      new Date(estimatedLandDate || actualLandDate) < valueCutOffDate,
+  })
 }
 
 function renderRequirementsForm(req, res) {
--- a/src/apps/investments/views/value-edit.njk
+++ b/src/apps/investments/views/value-edit.njk
@@ -99,7 +99,7 @@
       error: errors.messages.number_safeguarded_jobs
     }) }}
 
-    {% if investment.investment_type.name == 'FDI' %}
+    {% if projectValueNeeded and investment.investment_type.name == 'FDI' %}
       {{ MultipleChoiceField({
         name: 'fdi_value',
         label: form.labels.fdi_value,
--- a/test/functional/cypress/fixtures/investment/investment-land-date-after-April-2020.json
+++ b/test/functional/cypress/fixtures/investment/investment-land-date-after-April-2020.json
@@ -1,0 +1,188 @@
+{
+  "id": "6e441b34-1fc8-4638-b4b1-a0922ecf403e",
+  "incomplete_fields": [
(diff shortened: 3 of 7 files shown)


src/apps/investments/controllers/edit.js
```diff
@@ -3,7 +3,13 @@ function editDetailsGet(req, res) {
 }
 
 function editValueGet(req, res) {
-  res.breadcrumb('Edit value').render('investments/views/value-edit')
+  const actualLandDate = res.locals.form.state.actual_land_date
+  const estimatedLandDate = res.locals.form.state.estimated_land_date
```
rachaelcodes, line 7: will one of these always be present?

mforner13: At the earlier stages of a project there is often neither value, and in this case the field does not show. Project Value is only relevant for projects that were 'won' before 1st April, and to mark it as 'won' it has to have an actual land date, in which case the field would appear.

rachaelcodes: I'm not sure if `new Date(estimatedLandDate || actualLandDate)` will work though - would this become - `new Date(undefined)`, which I think would not be valid?

rachaelcodes: ah, it's ok if it's `new Date(null)`

mforner13: @rachaelcodes you were right! new Date(null) returned a date of 1970. I've just committed an extra check if actual or estimatedLandDate is truthy before setting the date objects 👍

src/apps/investments/controllers/edit.js
```diff
@@ -3,7 +3,14 @@ function editDetailsGet(req, res) {
...
```
web-bert, line 7: Optional: To make this a little more DRY it could be:
```javascript
const { state } = res.locals.form
const actualLandDate = state.actual_land_date
const estimatedLandDate = state.estimated_land_date
```

src/apps/investments/controllers/edit.js
```diff
@@ -3,7 +3,14 @@ function editDetailsGet(req, res) {
...
+
+  res.breadcrumb('Edit value').render('investments/views/value-edit', {
+    projectValueNeeded:
+      (estimatedLandDate || actualLandDate) &&
+      new Date(estimatedLandDate || actualLandDate) < new Date('2020-04-01'),
```
web-bert, line 12: Optional: The checking date can be a `const` to save re-creating it for each request and this could be a little more DRY too:
```javascript
// in the controller module
const valueCutoffDate = new Date('2020-04-01')
// in the method
const projectDate = estimatedLandDate || actualLandDate
res.breadcrumb('Edit value').render('investments/views/value-edit', {
    projectValueNeeded: !!projectDate && new Date(projectDate) < valueCutoffDate,
```

src/apps/investments/views/value-edit.njk
```diff
@@ -99,7 +99,7 @@
       error: errors.messages.number_safeguarded_jobs
     }) }}
 
-    {% if investment.investment_type.name == 'FDI' %}
+    {% if investment.investment_type.name == 'FDI' and projectValueNeeded %}
```
web-bert, line 102: Optional: Switch these around as the `projectValueNeeded` has already been evaluated

PR #2645, opened 2020-04-30
Feature/refactor company local header into react

debitan:
## Description of change

## Test instructions

_What should I see?_

### Before

_Add a screenshot_

### After

_Add a screenshot_

## Checklist


src/apps/companies/apps/activity-feed/controllers.js
```diff
@@ -1,22 +1,44 @@
...
+        }
       : {
-          companyId: company.id,
+          company,
+          breadcrumbs: [
+            { link: dashboard(), text: 'Home' },
+            { link: companies.index(), text: 'Companies' },
+            { link: companies.detail(company.id), text: company.name },
+            { text: 'Activity Feed' },
+          ],
```
web-bert, line 35: This could be more DRY by being defined higher up

src/apps/companies/apps/advisers/controllers/advisers.js
```diff
@@ -19,32 +18,41 @@ function renderLeadAdvisers(req, res) {
...
+      company,
+      breadcrumbs: [
+        { link: dashboard(), text: 'Home' },
+        { link: companies.index(), text: 'Companies' },
+        { link: companies.detail(company.id), text: company.name },
+        { text: 'Lead adviser' },
+      ],
+      returnUrl,
+      dnbRelatedCompaniesCount,
+      flashMessages: { ...res.locals.getMessages(), ...req.flashWithBody },
```
web-bert, line 48: I don't think `...req.flashWithBody` is needed there? `flashWithBody` is a function that gets called to set a flash message

src/apps/companies/apps/advisers/controllers/advisers.js
```diff
@@ -67,15 +75,24 @@ async function renderCoreTeamAdvisers(req, res, next) {
...
+        company,
+        breadcrumbs: [
+          { link: dashboard(), text: 'Home' },
+          { link: companies.index(), text: 'Companies' },
+          { link: companies.detail(company.id), text: company.name },
+          { text: 'Core Team' },
+        ],
+        returnUrl,
+        dnbRelatedCompaniesCount,
+        flashMessages: { ...res.locals.getMessages(), ...req.flashWithBody },
```
web-bert, line 93: Same about the `...req.flashWithBody`

src/apps/companies/apps/advisers/controllers/advisers.js
debitan: Yeah I realised this earlier. Will remove it.

src/apps/companies/apps/exports/controllers.js
```diff
@@ -2,14 +2,14 @@
...
 const metadataRepo = require('../../../../lib/metadata')
-const urls = require('../../../../lib/urls')
 const groupExportCountries = require('../../../../lib/group-export-countries')
 const getExportCountries = require('../../../../lib/get-export-countries')
 
 const { saveCompanyExportDetails } = require('../../repos')
 const { transformObjectToOption } = require('../../../transformers')
 const { transformCompanyToExportDetailsView } = require('./transformer')
 const { exportDetailsLabels, exportPotentialLabels } = require('../../labels')
+const { companies, dashboard } = require('../../../../lib/urls')
```
web-bert, line 12: I agree with a comment from @rafenden on another PR that it is better to read a url as `urls.companies.detail(company.id)` rather than `companies.detail(company.id)` so it makes it clear this is a url - maybe one for discussion!

debitan: I'm not massively tied either way. I got feedback in another PR to do it this way and then have just settled on it as it is shorter.

src/client/components/LocalHeader/FlashMessages.jsx
```diff
@@ -31,22 +32,34 @@ const messageColours = {
...
+const FlashMessages = ({ flashMessages }) => (
+  <UnorderedList listStyleType="none" data-auto-id="flash">
+    {Object.entries(flashMessages).map(([type, message]) => {
+      const parts = String(type).split(':')
+      return parts.length > 1
+        ? message.map((message) => (
+            <li>
+              <StatusMessage
+                colour={messageColours[parts[0]]}
+                key={message.body}
```
web-bert, line 44: I think the key should be on the `<li>`?

debitan: It should. I only just added in the li and forgot to move the key. Good spot!

src/client/components/LocalHeader/LocalHeader.jsx
```diff
@@ -69,6 +69,7 @@ LocalHeader.propTypes = {
 
 LocalHeader.defaultProps = {
   breadcrumbs: null,
+  flash: null,
```
web-bert, line 72: This should be `flashMessages`

src/client/components/LocalHeader/FlashMessages.jsx
```diff
@@ -31,22 +32,34 @@ const messageColours = {
...
-        </StatusMessage>
-      ))
-    ) : (
-      <StatusMessage colour={messageColours[type]} key={message}>
-        <StyledMessage>{message}</StyledMessage>
-      </StatusMessage>
-    )
-  })
+const FlashMessages = ({ flashMessages }) => (
+  <UnorderedList listStyleType="none" data-auto-id="flash">
```
web-bert, line 36: The `data-auto-id` should be more descriptive, something like `flash-messages` or `flash-message-list`

test/end-to-end/cypress/specs/DIT/referrals-spec.js
```diff
@@ -47,7 +46,7 @@ describe('Referrals', () => {
     })
 
     it('should display the new referral on the homepage', () => {
-      cy.get(selectors.localHeader().flash)
+      cy.get(selectors.companyLocalHeader().flash)
```
web-bert, line 49: this `.flash` should be more descriptive so maybe consistent with the other uses of `.flashMessages` or describe the element it is getting so `.flashMessageList`

test/functional/cypress/specs/companies/activity-feed-spec.js
```diff
@@ -107,39 +46,28 @@ describe('Company activity feed', () => {
...
-      expectedHeading: fixtures.company.archivedLtd.name,
-      expectedAddress: '16 Getabergsvagen, Geta, 22340, Malta',
-      expectedCompanyId: fixtures.company.archivedLtd.id,
-      expectedActivitiesHeading: '1 activity',
+      cy.visit(companies.activity.index(fixtures.company.archivedLtd.id))
     })
 
     it('should display the badge', () => {
-      cy.get(selectors.localHeader().badge(1)).should('have.text', 'Global HQ')
+      cy.get(selectors.companyLocalHeader().badge).should(
```
web-bert, line 53: I'd cache the selectors at the top of the test file `const localHeaderSelectors = selectors.companyLocalHeader()` as it's unnecessary to continually call a function that returns static content

test/functional/cypress/support/assertions.js
```diff
@@ -310,6 +310,53 @@ const assertAriaTablistTabSelected = (tabListLabel, tabLabel) =>
     .getDhTablistTab(tabListLabel, tabLabel)
     .should('have.attr', 'aria-selected', 'true')
 
+const commonCompanyTests = ({
```
web-bert, line 313: I haven't seen another use of this pattern of re-using tests, it feels like this should not be in `assertions.js` but another file? Or is there even a need for the `it` wrappers, it could just be `assertCompanyHeader()`?

debitan: I need to remove this. I put it here when I thought I would reuse it. It already lives in the activity feed tests.

test/functional/cypress/support/assertions.js
```diff
@@ -310,6 +310,53 @@ const assertAriaTablistTabSelected = (tabListLabel, tabLabel) =>
...
+    cy.get(selectors.tabbedLocalNav().item(1)).should('have.text', 'Activity')
+    cy.get(selectors.tabbedLocalNav().item(2)).should(
+      'have.text',
+      'Company contacts'
+    )
+    cy.get(selectors.tabbedLocalNav().item(3)).should('have.text', 'Core team')
+    cy.get(selectors.tabbedLocalNav().item(4)).should('have.text', 'Investment')
+    cy.get(selectors.tabbedLocalNav().item(5)).should('have.text', 'Export')
+    cy.get(selectors.tabbedLocalNav().item(6)).should('have.text', 'Orders')
+  })
```
web-bert, line 350: Maybe one for the future but it seems like this could be refactored:
```
function assertTabbedLocalNav(labels){
  items.forEach((label, index) => cy.get(selectors.tabbedLocalNav().item(index).should('have.text', label))
}

assertTabbedLocalNav([
  'Activity',
  'Company contacts',
  'Core team',
  'Investment',
  'Export',
  'Orders',
])
```

debitan: Good point. Think this PR might be big enough though haha!

src/client/components/LocalHeader/FlashMessages.jsx
rafenden: Value from `data-auto-id` was used in GA so it will break the tracking if it changes.

PR #2647, opened 2020-05-01
Add flash messages to react local header

debitan:
## Description of change

Add flash messages to the react local header in preparation for its use in the company local header.

The flash messages are generated by `req.flash` and `req.flashWithBody` and get passed through to the component as an object. 

Functional tests covering an implementation of this will be in another PR when this new functionality is used for the first time. 

There will be another PR following this one showing how the flash messages work in the company header.

## Test instructions

Open storybook, `yarn start-storybook`, and you will see an example of the different flash message types.

## Screenshots

image

## Checklist

### DIFF
--- a/src/client/components/LocalHeader/FlashMessages.jsx
+++ b/src/client/components/LocalHeader/FlashMessages.jsx
@@ -1,0 +1,66 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import styled from 'styled-components'
+
+import { ERROR_COLOUR, BLUE, BUTTON_COLOUR, ORANGE, BLACK } from 'govuk-colours'
+import { FONT_WEIGHTS } from '@govuk-react/constants/lib/typography'
+import { FONT_SIZE } from '@govuk-react/constants'
+import { StatusMessage } from 'data-hub-components'
+
+const StyledBody = styled('p')`
+  margin-bottom: 0;
+  font-weight: ${FONT_WEIGHTS.bold};
+`
+
+const StyledHeading = styled('h2')`
+  font-size: ${FONT_SIZE.SIZE_20};
+  font-weight: ${FONT_WEIGHTS.bold};
+  margin-top: 0;
+`
+
+const StyledMessage = styled('p')`
+  margin: 0;
+  font-size: ${FONT_SIZE.SIZE_20};
+`
+
+const messageColours = {
+  info: BLUE,
+  success: BUTTON_COLOUR,
+  warning: ORANGE,
+  error: ERROR_COLOUR,
+  muted: BLACK,
+}
+
+const FlashMessages = ({ flashMessages }) =>
+  Object.entries(flashMessages).map(([type, message]) => {
+    const parts = String(type).split(':')
+    return parts.length > 1 ? (
+      message.map((message) => (
+        <StatusMessage colour={messageColours[parts[0]]} key={message.body}>
+          <StyledHeading>{message.heading}</StyledHeading>
+          <StyledBody>{message.body}</StyledBody>
+        </StatusMessage>
+      ))
+    ) : (
+      <StatusMessage colour={messageColours[type]} key={message}>
+        <StyledMessage>{message}</StyledMessage>
+      </StatusMessage>
+    )
+  })
+
+FlashMessages.propTypes = {
+  flashMessages: PropTypes.shape({
+    type: PropTypes.oneOfType([
+      PropTypes.arrayOf(
+        PropTypes.shape({
+          body: PropTypes.string.isRequired,
+          heading: PropTypes.string.isRequired,
+          id: PropTypes.string,
+        })
+      ),
+      PropTypes.arrayOf(PropTypes.string).isRequired,
+    ]),
+  }).isRequired,
+}
+
+export default FlashMessages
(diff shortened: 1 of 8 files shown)


src/client/components/LocalHeader/LocalHeaderHeading.jsx
```diff
@@ -0,0 +1,8 @@
+import styled from 'styled-components'
+import { typography } from '@govuk-react/lib'
+
+const LocalHeaderHeading = styled('h1')`
+  ${typography.font({ size: 36, weight: 'bold' })};
+`
+
+export default LocalHeaderHeading
```
debitan, line 8: Have simply split out the heading component here.

src/client/components/LocalHeader/__stories__/FlashMessages.stories.jsx
```diff
@@ -0,0 +1,36 @@
...
+        'info:with-body': [
+          {
+            heading: 'Info message heading',
+            body: 'Info message body',
+          },
+        ],
+        error: ['Error test message'],
+        warning: ['Warning test message'],
+        muted: ['Muted test message'],
+      }}
```
debitan, line 34: This is the shape of the data that is generated by the express `connect-flash` plugin.

src/client/components/LocalHeader/FlashMessages.jsx
```diff
@@ -0,0 +1,65 @@
...
+
+const messageColours = {
+  info: BLUE,
+  success: BUTTON_COLOUR,
+  warning: ORANGE,
+  error: ERROR_COLOUR,
+  muted: BLACK,
+}
+
+const FlashMessages = ({ flashMessages }) =>
```
web-bert, line 33: Seeing as the component is called `FlashMessages` can't the prop just be `messages` or `items` ?

src/client/components/LocalHeader/LocalHeaderHeading.jsx
web-bert: What is the point in doing this, for an upcoming PR?

src/client/components/LocalHeader/FlashMessages.jsx
```diff
@@ -0,0 +1,65 @@
...
+  Object.entries(flashMessages).map(([type, message]) => {
```
web-bert, line 34: This is a list of messages and so semantically it should be marked up so.  There should be a wrapping element here, ideally a `ul`, with a list of `li` items that are the messages. This wrapping element should have `role="alert"` on it too.

src/client/components/LocalHeader/FlashMessages.jsx
```diff
@@ -0,0 +1,65 @@
...
+  font-weight: ${FONT_WEIGHTS.bold};
+`
+
+const StyledHeading = styled('h2')`
+  font-size: ${FONT_SIZE.SIZE_20};
+  font-weight: ${FONT_WEIGHTS.bold};
+  margin-top: 0;
+`
+
+const StyledMessage = styled('span')`
```
web-bert, line 21: This should probably be a `p` tag for the semantics

src/client/components/LocalHeader/FlashMessages.jsx
debitan: Yep. I thought about this as the current flash implementation is ul and li. However, the statusMessage component I'm reusing is a div so didn't know whether it would be good to wrap it in an li. Happy to do so though.

src/client/components/LocalHeader/LocalHeaderHeading.jsx
debitan: Yep, it's in preparation for the next PR and has kind of leaked here. Happy to remove if you'd like, but it's also causing no harm.

src/client/components/LocalHeader/FlashMessages.jsx
web-bert: It is valid to have a `div` inside an `li` so I don't see why not

src/client/components/LocalHeader/FlashMessages.jsx
debitan: Agreed. Will just need to adjust the margin.

src/client/components/LocalHeader/LocalHeaderHeading.jsx
web-bert: Yeah no harm it so fine to leave it 👍

src/client/components/LocalHeader/FlashMessages.jsx
debitan: Cool. That is what I was unsure about, so thanks for clarifying. I'll make the change on Monday.

debitan: To be properly semantic maybe we should update the `StatusMessage` component to be a `<dialog>` element? Obviously out of scope of this PR.

src/client/components/LocalHeader/FlashMessages.jsx
debitan: It could, but I was thinking as the `flashMessages` will be passed all the way down from the local header implementation layer it would be clearer as to what these messages are if they are called `flashMessages`. We could alter the props name as it enters this component, but I was thinking consistency was better.

PR #2655, opened 2020-05-05
Add new SubTabNav component

web-bert:
## Description of change

As part of the upcoming pipeline to the dashboard, this component is needed to provide sub tabs for the main `TabNav` tab. This is currently a simple component and will be fleshed out when used as part of the `TavNav` tab in an upcoming PR - along with functional tests.

## Test instructions

Run story book with `yarn storybook`

## Screenshots
Screenshot 2020-05-05 at 15 43 15

## Checklist

### DIFF
--- a/src/client/components/SubTabNav/index.jsx
+++ b/src/client/components/SubTabNav/index.jsx
@@ -1,0 +1,101 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import styled from 'styled-components'
+import { WHITE, BLUE, GREY_2 } from 'govuk-colours'
+import { SPACING, FONT_SIZE } from '@govuk-react/constants'
+import Link from '@govuk-react/link'
+
+const StyledTabList = styled.ol`
+  padding: 0;
+  margin: 0 0 2px 0;
+  color: ${BLUE};
+  background-color: ${WHITE};
+  display: flex;
+  flex-direction: row;
+  flext-wrap: wrap;
+  justify-content: center;
+  justify-content: space-between;
+`
+const StyledTabItem = styled.li`
+  list-style: none;
+  font-size: ${FONT_SIZE.SIZE_24};
+  font-weight: normal;
+  display: inline-block;
+  line-height: 2;
+`
+
+const StyledTabPanel = styled.div`
+  margin: 0;
+  padding ${SPACING.SCALE_4} 0 0 0;
+  border-top: 1px solid ${GREY_2};
+`
+
+const StyledSpan = styled.span`
+  border-bottom: 8px solid ${BLUE};
+  padding-bottom: ${SPACING.SCALE_1};
+`
+
+const StyledLink = styled(Link)`
+  text-decoration: none;
+`
+
+function createId(id, key) {
+  return `sub-nav.${id}.${key}`
+}
+
+function SubTabNav({ id, label, selected, tabs, ...rest }) {
+  const tabList = Object.entries(tabs)
+  const selectedKey = selected || tabList[0][0]
+  const panelId = createId(id, selectedKey)
+  return (
+    <>
+      <StyledTabList {...rest} role="tablist" aria-label={label}>
+        {tabList.map(([key, { label, href }]) => {
+          const isSelected = selectedKey === key
+          const tabId = createId(id, key) + '.tab'
+          return (
+            <StyledTabItem
+              role="tab"
+              aria-selected={isSelected ? 'true' : 'false'}
+              id={tabId}
+              key={tabId}
+            >
+              {isSelected ? (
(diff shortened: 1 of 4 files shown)


src/client/components/SubTabNav/index.jsx
```diff
@@ -0,0 +1,101 @@
...
+      <StyledTabList {...rest} role="tablist" aria-label={label}>
+        {tabList.map(([key, { label, href }]) => {
+          const isSelected = selectedKey === key
+          const tabId = createId(id, key) + '.tab'
+          return (
+            <StyledTabItem
+              role="tab"
+              aria-selected={isSelected ? 'true' : 'false'}
+              id={tabId}
+              key={tabId}
```
adamledwards, line 61:
```suggestion
              key={tabId}
              onClick={() => onTabNavClicked(key)}
```
Should this have an onTabNavClicked event that passes the selected key to the subscribing function.

web-bert: I was trying to get this PR through as a bare bones implementation, there are real world uses of this yet so I was going to add more features as they are needed.

web-bert:
@adamledwards thanks for the suggestion of making it interactive. I thought about whether to make it interactive or not, but at the moment I don't know how we are going to use it so I thought I would leave that to get a natural answer once this component is used.

web-bert:
@peterhudec The `TabNav` component is not using the ARIA tab role correctly, as we don't have multiple panels in the page that are shown and hidden, therefore I decided to use links as that is what they are. The `TabNav` is is also breaking built in behaviour of the browser by not supporting ctrl/cmd clicking a link to open in a new window/tab. I would even question if there is any benefit of adding the ARIA attributes and whether a `<nav>` would actually be better, as they are essentially a list of links.

After thinking about it we might be better to make two components for the different use cases, `TabNav` that is a list of links in a `<nav>` and `TabPanels` which follows the AIRA tab role standard and has multiple panels of content that are dynamically switched with JS

peterhudec:
@web-bert I'm not a a11y expert, but from [what I read](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Navigation_Role), it seems that the _navigation_ role is a _landmark_ role and as such should basically only be used as the main navigation of the application.

> Landmark roles are intended to be used sparingly, to identify larger overall sections of the document. Using too many landmark roles can create "noise" in screen readers, making it difficult to understand the overall layout of the page. - [MDN ARIA: Navigation Role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Navigation_Role#Accessibility_concerns)

Using a _navigation_ hierarchically under a _tab nav_ thus doesn't seem to be the best idea. And your `SubTabNav` even admits in its name that it is a _tab nav_ and it's intended to be used underneath another _tab nav_ (an assumption a component should not make).

The `TabNav` component implements the [ARIA: tab](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Tab_Role) including the best practices to the point, with the only exception that there's only ever a single panel rendered at a time, which however has no effect on accessibility.

Anyway, I'm not sure we are in the position to reinvent the _tab nav_ accessibility and should probably stick to the standard, unless it turns out our users have a real problem with it like [these guys found out](http://simplyaccessible.com/article/danger-aria-tabs/).

web-bert:
The important bit is

> navigation landmarks provide a way to identify groups (e.g. lists) of links that are intended to be used for website or page content navigation. If a page includes more than one navigation landmark, each should have a unique label.

It is fine to have more than one, and in the context of our main `TabNav` that is what they are - so labelling them as navigation would make more sense than a `tablist` with one `tabpanel` - we have taken the visual representation of a `tablist` and mixed it with the behaviour of links. The suggested markup is not a standard only a suggestion, given the fact we are not hiding and revealing content (as also suggested) but essentially changing all the content below the tab (and updating the URL) like a link, the tab should definitely be an `<a>` tag with all the default behaviour that comes with them (tab focus, keyboard interaction) and we should not be breaking basic built in browser behaviour of ctrl/cmd click to open in new tab/window - especially when we are updating the URL to mimic what a normal link does.

What I am suggesting is that `TabNav` should be a `<nav>` list of links (which can still act the same with fetching the content and updating the URL) and then as you previously suggested our sub tab can just restyle that component. Then if we ever have the need for a `tablist` we can create a new `TabPanels` component.

That way the page structure would be `<nav>` for the top level navigation, one for the tab navigation and then, if required, one for the sub tab navigation. Therefore from a semantic markup point of view all the navigation can be  found easily and when going from tab to tab you know that you can find `<nav>` element two and select one of the links in there to navigate to another tab.

web-bert:
Reworked into #2677

PR #2660, opened 2020-05-07
Fix logic to show project value if either date before cut off

mforner13:
## Description of change

This fix follows a recent PR which added logic to only display the Project Value field for an investment project if the project's actual or estimated land date was before 01/04/2020. 

There is a bug in that when the estimated land date is after 01/04/2020 but the actual land date is before, this expression returns false: `Date(estimatedLandDate || actualLandDate) < valueCutOffDate`.

This PR fixes the logic in the editValue controller by first eliminating the null fields (because 
`new Date(null)` returns 01/01/1970) and then filtering the results by whether they're before the cut off date. 

I have also added more test cases to cover all possible combinations of estimated and actual land dates.

## Test instructions

Go to an investment project.
Change the estimated land date to after 01/04/200.
Change the actual land date to before 01/04/2020.
Click Add/Edit Value on the project details page.
Check the Project Value field appears.

## Checklist


src/apps/investments/controllers/edit.js
```diff
@@ -10,11 +10,14 @@ function editValueGet(req, res) {
...
+    (date) => date !== null
+  )
+  const datesBeforeCutOff = notNullDates.filter(
+    (date) => new Date(date) < valueCutOffDate
+  )
   res.breadcrumb('Edit value').render('investments/views/value-edit', {
-    projectValueNeeded:
-      (estimatedLandDate || actualLandDate) &&
-      new Date(estimatedLandDate || actualLandDate) < valueCutOffDate,
+    projectValueNeeded: datesBeforeCutOff.length > 0,
```
web-bert, line 20: I think this can be refactored slightly:
```js
function editValueGet(req, res) {
  const {
    estimated_land_date: estimatedLandDate,
    actual_land_date: actualLandDate,
  } = res.locals.form.state

  const datesBeforeCutOff = [estimatedLandDate, actualLandDate].some(
    (date) => date !== null && new Date(date) < valueCutOffDate
  )
  
  res.breadcrumb('Edit value').render('investments/views/value-edit', {
    projectValueNeeded: datesBeforeCutOff
  })
}
```

PR #2670, opened 2020-05-13
Change the content around asking for policy feedback

ian-leggett:
## Description of change

This is a content change to the question around policy feedback.

## Test instructions

Go to add an interaction, scroll down to the option to add policy feedback and you should see the new content (refer to screenshots below)

### Before
Screenshot 2020-05-13 at 09 29 24

### After
Screenshot 2020-05-18 at 10 48 32

## Checklist

### DIFF
--- a/src/apps/interactions/apps/details-form/client/StepInteractionDetails.jsx
+++ b/src/apps/interactions/apps/details-form/client/StepInteractionDetails.jsx
@@ -9,6 +9,10 @@ import axios from 'axios'
 import { GREY_1 } from 'govuk-colours'
 import styled from 'styled-components'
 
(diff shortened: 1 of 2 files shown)


web-bert:
I have to say I am not a fan of the way that the content is written, the label does not make sense without the hint text, and for visually impaired users they would not know about the hint text from interacting with the input - they would get "Did the contact provide (any of the the following)" and then two options of "Yes or "No", leaving them confused as to what the options are.

Either the radio options should be changed to "Government policy", "Business Intelligence" and "Other" or the content should be rewritten to something like "Did the contact provide any additional information?" and the hint text can be "This includes feedback on government policy or business intelligence"

PR #2672, opened 2020-05-14
Do not remove the Form fieldName when it's an empty String

LaurenQurashi:
## Description of change
This PR removes code that prevents an empty string field value from being added to the form state. This code would cause a bug where if an optional field had a value entered into it and saved, then later the user edited that field again and cleared the value, this input would not be updated. 

Similar changes have been made [in the components library](https://github.com/uktrade/data-hub-components/pull/299). Due to the components library being updated too, this PR also includes a version bump of the data-hub-components dependency. 

A visual example of what I'm on about: 

Before
```javascript
// Form values
const values: { a: '', b: 'foo' }
// Strip out any object key when the value is an empty string.
const values: { b: 'foo' }
```
After
```javascript
// Form values
const values: { a: '', b: 'foo' } 
// Keep all keys where the corresponding value is an empty string
const values: { a: '', b: 'foo' }

``` 

## Test instructions
On master - navigate to the interactions tab, then click on an interaction and edit it. Fill out all the fields including the notes text area. Click save, and you'll notice all is fine. Edit that same interaction again and clear out the notes text area. Click save, and you'll see it does not update. 

On this branch - do the same as above, however you should notice that once you clear the notes text area and save, the notes do not appear in the table row (as there's nothing to show cause it's updated your empty value).

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -91,7 +91,7 @@
     "core-js": "^2.5.7",
     "css-loader": "^1.0.0",
     "csurf": "^1.9.0",
-    "data-hub-components": "5.5.1",
+    "data-hub-components": "5.5.2",
     "date-fns": "^1.29.0",
     "del-cli": "^3.0.0",
     "dotenv": "^8.2.0",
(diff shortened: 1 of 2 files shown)


paulgain:
@LaurenQurashi there is no validation going on here. We should change the title to:
`Do not remove the Form fieldName when it's an empty String`

PR #2674, opened 2020-05-14
Add investment project admin page

mforner13:
## Description of change

This PR allows users from the FDI verification team to move a project between stages in the front end rather than in Django admin. Users with the correct 'can_move_to_any_stage' permission will be able to see the 'Admin' tab on an investment project. On this page they will be able to select any other stage to move the project to. A user is then redirected to the project details page if the request is successful. 

This PR is linked to another ( #2675 ) which sets a flash message in session storage which is then retrieved by the FlashMessages react component. They have been separated for clarity. 

## Test instructions

- Ensure you have the `investment.change_to_any_stage_investmentproject` permission
- Go to an investment project and click on the 'Admin' tab
- Select a new stage or click save without selecting a stage to view validation error
- Check that the new stage is displayed on the project details page

### After

Screenshot 2020-05-14 at 14 16 12

## Checklist


src/client/components/Form/index.jsx
```diff
@@ -69,7 +69,7 @@ const Form = ({
       return index !== -1 ? index : null
     },
     isFirstStep: () => props?.currentStep === 0,
-    isLastStep: () => props?.currentStep === steps?.length - 1,
+    isLastStep: () => !steps.length || props?.currentStep === steps?.length - 1,
```
adamledwards, line 72: if there is only one step is step.length == 0 ? I would imagine that if `isLastStep: steps.length == 1 || props?.currentStep === steps?.length - 1,`

src/apps/investments/views/admin/client/InvestmentProjectAdmin.jsx
```diff
@@ -21,7 +21,12 @@ const InvestmentProjectAdmin = ({
   projectName,
   projectStage,
   stages,
+  stageUpdated,
 }) => {
+  useEffect(() => {
+    stageUpdated &&
+      (window.location.href = urls.investments.projects.project(projectId))
+  })
```
adamledwards, line 29: you could add 

js```
useEffect(() => {
    stageUpdated &&
      (window.location.href = urls.investments.projects.project(projectId))
  }, [stageUpdated])
```
to prevent this running on each re-render this only runs when the stageUpdated value has changed

File: src/apps/investments/controllers/admin.js [hunk 3]
```diff
@@ -0,0 +1,27 @@
...
+
+async function renderAdminView(req, res) {
+  const { token } = req.session
+  const { id, name, stage } = res.locals.investment
+  const orderStages = (stages) => {
+    const activeStage = stages[0]
+    stages[0] = stages[2]
+    stages[2] = activeStage
+    return stages
+  }
```
adamledwards, line 11: I believe  that this should be a BE change and the FE shouldn't have this logic. 
However sometime we have to do this due to lack of resources, fast deployment etc.

Could you place a comment here to explain why you are swapping the order of the array.

src/client/components/Form/index.jsx
mforner13: @adamledwards I just tried that and it doesn't work unfortunately. @rafenden actually made this change, but I think that it needs to be !steps.length because there are no step components used in this form, so steps.length is 0 not 1.

src/apps/investments/controllers/admin.js
mforner13: @adamledwards The stages come from the API in alphabetical order. For the radio buttons we want to have them in chronological order (as they appear in the timeline on a project details page). I don't think we'd want the API to return them in chronological order by default as that wouldn't be what users of the API would expect. 

Have added a comment: https://github.com/uktrade/data-hub-frontend/pull/2674/commits/c2ed94d08be7934599d678aba954e585a53eb6db

src/client/components/Form/index.jsx
adamledwards: 🤯

src/apps/investments/views/admin/client/InvestmentProjectAdmin.jsx
```diff
@@ -0,0 +1,93 @@
...
+                { link: urls.investments.projects.index(), text: 'Projects' },
+                {
+                  link: urls.investments.projects.project(projectId),
+                  text: projectName,
+                },
+                { text: 'Admin' },
+              ]}
+            />
+            <Main>
+              <H4>Project details</H4>
```
ian-leggett, line 57: I think these should be rendered as a`<h2>` but styled to the size of a `<h4>` as this is the next header in the HTML document.

src/apps/investments/views/admin/client/InvestmentProjectAdmin.jsx
```diff
@@ -0,0 +1,93 @@
...
+                id={STATE_ID}
+                onSubmit={(values) => {
+                  updateStageTask.start({
+                    payload: { values, projectId },
+                    onSuccessDispatch: INVESTMENT_PROJECT_ADMIN__UPDATE_STAGE,
+                  })
+                }}
+                submissionError={updateStageTask.errorMessage}
+              >
+                <H4>Change the stage to</H4>
```
ian-leggett, line 72: Same as above

src/apps/investments/views/admin/client/InvestmentProjectAdmin.jsx
mforner13: Done :)

PR #2675, opened 2020-05-14
Set flash message in session storage

mforner13:
## Description of change

This PR proposes a solution for setting flash messages client-side when redirecting from a react page to a nunjucks page.

**Logic**

- I have added a slot for the flashMessages react component to the nunjucks local-header template. 
- I have also created a set of flash message util functions, which set, retrieve or clear flash messages in `window.sessionStorage`
- The FlashMessages component has been updated to either receive flashMessages as a prop, or get the flashMessages object from sessionStorage.
- The object from sessionStorage is saved as a const and then removed from localStorage to avoid the flash message persisting after refresh/change of url.
- If no flash message is coming from props or sessionStorage, the component returns null, avoiding there being an error in the local-header template when no flash message is present. 

**Implementation in the investment admin page**

- When the update stage task is successful, a flash message object is added to sessionStorage from the InvestmentAdminPage component and the user is redirected to the project details page. 
- The FlashMessages component gets this object from sessionStorage and it is rendered as part of the local-header nunjucks template. 

When the flash message is set through sessionStorage it's not possible to type check it as it's not passed as a prop to the FlashMessages component, which is a drawback of this solution. For now the error will be logged in the console.

This PR is linked to #2674 and will be merged into that branch when approved.

## Test instructions

- Ensure you have the `investment.change_to_any_stage_investmentproject` permission
- Go to an investment project and click on the 'Admin' tab
- Select a new stage and click save
- Check that the flash message appears on the project details page
- Refresh the page or go to another page to check that the flash message is removed

### After
Screenshot 2020-05-14 at 14 50 47

_Add a screenshot_

## Checklist

### DIFF
--- a/src/apps/investments/views/admin/client/InvestmentProjectAdmin.jsx
+++ b/src/apps/investments/views/admin/client/InvestmentProjectAdmin.jsx
@@ -11,6 +11,7 @@ import Task from '../../../../../client/components/Task'
 import { ID as STATE_ID, TASK_UPDATE_STAGE, state2props } from './state'
 import { INVESTMENT_PROJECT_ADMIN__UPDATE_STAGE } from '../../../../../client/actions'
 import urls from '../../../../../lib/urls'
+import flashUtils from '../../../../../client/utils/flash-messages'
 
 const StyledP = styled('p')`
   margin-bottom: ${SPACING.SCALE_2};
@@ -24,8 +25,10 @@ const InvestmentProjectAdmin = ({
   stageUpdated,
 }) => {
   useEffect(() => {
-    stageUpdated &&
-      (window.location.href = urls.investments.projects.project(projectId))
+    if (stageUpdated) {
+      flashUtils.addSuccessMessage('Project stage saved')
+      window.location.href = urls.investments.projects.project(projectId)
+    }
   }, [stageUpdated])
   const newStageOptions = stages.filter(
     (stage) => stage.value != projectStage.id
--- a/src/client/components/LocalHeader/FlashMessages.jsx
+++ b/src/client/components/LocalHeader/FlashMessages.jsx
@@ -7,6 +7,7 @@ import { FONT_WEIGHTS } from '@govuk-react/constants/lib/typography'
 import { FONT_SIZE } from '@govuk-react/constants'
 import UnorderedList from '@govuk-react/unordered-list'
 import { StatusMessage } from 'data-hub-components'
+import flashUtils from '../../utils/flash-messages'
 
 const StyledBody = styled('p')`
   margin-bottom: 0;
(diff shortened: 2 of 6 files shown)


src/apps/investments/views/admin/client/InvestmentProjectAdmin.jsx
```diff
@@ -24,8 +24,15 @@ const InvestmentProjectAdmin = ({
   stageUpdated,
 }) => {
   useEffect(() => {
-    stageUpdated &&
-      (window.location.href = urls.investments.projects.project(projectId))
+    if (stageUpdated) {
+      window.localStorage.setItem(
```
web-bert, line 28: I think this should be using `window.sessionStorage` so that it will be wiped when the tab is closed.

Also, this should be logic contained in a util so it is easily shared, so this should be something like:
```js
if (stageUpdated) {
  flashUtils.addSuccessMessage('Project stage saved')
  window.location.href = urls.investments.projects.project(projectId)
}
```

with the util something like this:
```js
const KEY = 'flash-messages'
function getMessages(){
  const items = window.sessionStorage.getItem(KEY)
 if( items ){
    try {
      return JSON.parse(items)
    } catch ( e ) {
      //log error here?
    }
  return {}
}

function addMessage(state, message){
  const messages = getMessages()
  messages[state] = messages[state] || []
  messages[state].push(message)
  try {
    window.sessionStorage.setItem(KEY, JSON.stringify(messages))
  } catch( e ) {
    //log error?
  }
}

function addSuccessMessage(message){
  addMessage('success', message)
}
```

src/client/components/LocalHeader/FlashMessages.jsx
```diff
@@ -32,31 +32,45 @@ const messageColours = {
...
-              </StatusMessage>
-            </li>
-          ))
-    })}
-  </UnorderedList>
-)
+const FlashMessages = ({ flashMessages }) => {
+  const flashMessagesFromStorage = JSON.parse(
+    window.localStorage.getItem('flashMessage')
+  )
```
web-bert, line 38: This can be from utils in previous comment and can be:
```js
const flashMessagesFromStorage = flashUtils.getMessages()
```

src/client/components/LocalHeader/FlashMessages.jsx
```diff
@@ -32,31 +32,45 @@ const messageColours = {
...
+  window.localStorage.removeItem('flashMessage')
```
web-bert, line 39: This should probably also be in a utils file so:
```js
flashUtils.clearMessages()
```

src/apps/investments/views/admin/client/InvestmentProjectAdmin.jsx
mforner13: Thanks Graham, will make this change 👍

src/client/components/LocalHeader/FlashMessages.jsx
```diff
@@ -32,31 +32,45 @@ const messageColours = {
...
+
+  if (flashMessages || flashMessagesFromStorage) {
+    return (
+      <UnorderedList listStyleType="none" data-auto-id="flash">
+        {Object.entries(flashMessages || flashMessagesFromStorage).map(
```
adamledwards, line 44: Just checking if the logic is correct here do we want to ignore the local flash messages if there are flash messages sent from the server or do we want to combine them?

mforner13: @adamledwards so the react flash message component doesn't handle ones coming from the server, they're still just handled on req.flash in the node layer. 

The flashMessages prop would come from when the localHeader of the page where you want to show the flash message is in react. So the two couldn't happen at the same time, as flashMessagesFromStorage would be used with the nunjucks local header.

src/apps/investments/views/admin/client/InvestmentProjectAdmin.jsx
```diff
@@ -11,6 +11,7 @@ import Task from '../../../../../client/components/Task'
 import { ID as STATE_ID, TASK_UPDATE_STAGE, state2props } from './state'
 import { INVESTMENT_PROJECT_ADMIN__UPDATE_STAGE } from '../../../../../client/actions'
 import urls from '../../../../../lib/urls'
+import flashUtils from '../../../../../client/utils/flashMessages'
```
web-bert, line 14: This file name should be all lower case: `flash-messages`

PR #2691, opened 2020-05-19
Pipeline Edit

adamledwards:
## Description of change
This PR add the ability to edit pipelines. It shares the form component for create and update so there was a small refactor to share the forms.
There was also a refactor to move the pipeline app out of the company subapp into its own app. The reasoning behind this was the company middleware was only used for adding a pipeline. For updating and viewing pipelines there is no need for the company middleware. 

There is also a fix to the form to submit when there is only one step. This fix is on master can not rebase off master because I've branched off feature/pipeline. This will be fixed in a rebase and the commit will be omitted. https://github.com/uktrade/data-hub-frontend/pull/2691/commits/cb56ddb21dd4361408b74cc69517bb306e134861

## Test instructions
on the deployment site go to /my-pipeline/876108b8-32ac-47b3-b641-c4089447a4bc/edit and change the values click on update and you should be redirected to /my-pipelines. Go back to the edit url /my-pipeline/876108b8-32ac-47b3-b641-c4089447a4bc/edit and check if the values have been updated.

## Screenshots
image


src/apps/my-pipeline/client/EditPipelineForm.jsx
```diff
@@ -0,0 +1,114 @@
...
+  csrfToken,
+  currentPipeline,
+  savedId,
+}) {
+  useEffect(() => {
+    if (savedId) {
+      /**
+       * TODO: Replace with react router navigation.
+       * As we move to SPA clear the saveId from the state before navigation.
+       */
```
debitan, line 67: I think we can do this now as the sub tab nav is hooked up to react router.

src/apps/my-pipeline/client/EditPipelineForm.jsx
```diff
@@ -0,0 +1,114 @@
...
+              >
+                <PipelineForm
+                  submissionError={editPipelineItem.errorMessage}
+                  onSubmit={(values) => {
+                    editPipelineItem.start({
+                      payload: { values, pipelineItemId, csrfToken },
+                      onSuccessDispatch: PIPELINE__EDIT_PIPELINE_SUCCESS,
+                    })
+                  }}
+                  cancelLink={urls.pipeline.index()}
```
debitan, line 93: Ideally this would take you back to the screen you were on before. So if you were on the active tab of the pipeline, then there. Currently this would take you to the prospects tab. We could use `window.history.back()`, but I guess that would be quite dirty. 
Another cleaner way would be to check the status of the item on load and set this `cancelLink` to take you back to that status screen.

src/apps/my-pipeline/client/EditPipelineForm.jsx
```diff
@@ -0,0 +1,114 @@
...
+    </Task>
+  )
+}
+
+EditPipelineForm.propTypes = {
+  pipelineItemId: PropTypes.string,
+  csrfToken: PropTypes.string,
+  currentPipeline: PropTypes.object,
+  savedId: PropTypes.string,
+}
```
debitan, line 112: Same comment here as above re propTypes

src/apps/my-pipeline/client/PipelineForm.jsx
```diff
@@ -0,0 +1,50 @@
...
+    </Form>
+  )
+}
+
+PipelineForm.propTypes = {
+  onSubmit: PropTypes.func.isRequired,
+  submissionError: PropTypes.string,
+  cancelLink: PropTypes.string.isRequired,
+  initialValue: PropTypes.objectOf(PropTypes.string),
+}
```
debitan, line 48: same here re `defaultProps`

src/apps/my-pipeline/client/index.jsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react'
+import { connect } from 'react-redux'
+import { state2props } from './state'
+import AddToPipelineForm from './AddToPipelineForm'
+import EditPipelineForm from './EditPipelineForm'
+
+function PipeLineFormView(props) {
+  const { pipelineItemId } = props
```
debitan, line 8: OPTIONAL: can this be de-structured inline? with something like this? 
```suggestion
function PipeLineFormView({ pipelineItemId }, props) {
```

src/apps/my-pipeline/client/tasks.js
```diff
@@ -0,0 +1,43 @@
...
+export async function editPipelineItem({ values, pipelineItemId, csrfToken }) {
+  await axios.post(
+    `${urls.pipeline.edit(pipelineItemId, { _csrf: csrfToken })}`,
+    {
+      name: values.name,
+      status: values.category,
+    }
+  )
+  return pipelineItemId
+}
```
debitan, line 43: we can now go through the `api-proxy` and set a flashMessage from the client(!). see @mforner13 's work re this. understand this is just you moving this file rather adding anything new here, so you might want it split out as a separate PR.

src/apps/my-pipeline/controllers/add.js
```diff
@@ -1,13 +1,13 @@
...
-const config = require('../../../../../config')
+const urls = require('../../../lib/urls')
+const { authorisedRequest } = require('../../../lib/authorised-request')
+const config = require('../../../config')
 
 function renderAddToPipeline(req, res) {
   const { company } = res.locals
   res
     .breadcrumb(company.name, urls.companies.detail(company.id))
     .breadcrumb('Add to your pipeline')
```
debitan, line 9: Another one we can come back to in another PR, but we've got the React local header now to use. Didn't when I wrote this first time round.

src/apps/my-pipeline/controllers/edit.js
```diff
@@ -0,0 +1,34 @@
+const { authorisedRequest } = require('../../../lib/authorised-request')
+const config = require('../../../config')
+
+function renderEditPipeline(req, res) {
+  const { pipelineItemId } = req.params
+  res
+    .breadcrumb('Edit your pipeline')
```
debitan, line 7: Once we use the React local header we won't need this line

src/apps/my-pipeline/controllers/edit.js
```diff
@@ -0,0 +1,34 @@
...
+    .render('my-pipeline/views/pipeline-form', {
+      heading: `Edit your pipeline`,
```
debitan, line 9: Nor this

src/apps/my-pipeline/controllers/edit.js
```diff
@@ -0,0 +1,34 @@
...
+      url: `${config.apiRoot}/v4/pipeline-item/${pipelineItemId}`,
+      method: 'PATCH',
+      body: req.body,
+    })
+    req.flash('success', 'Pipeline changes for this company have been saved')
+    return res.send(response)
+  } catch ({ error, statusCode }) {
+    return res.status(statusCode).send(error)
+  }
+}
```
debitan, line 29: So here we can go through the api-proxy and set the flash message on the client ala @mforner13

src/client/index.jsx
```diff
@@ -297,9 +301,9 @@ function App() {
         <Mount selector="#company-export-countries-edit">
           {(props) => <ExportCountriesEdit {...props} />}
         </Mount>
-        <Mount selector="#add-to-pipeline-form">
+        <Mount selector="#pipeline-form">
           {(props) => (
-            <AddToPipelineForm {...props} csrfToken={globalProps.csrfToken} />
+            <PipelineForm {...props} csrfToken={globalProps.csrfToken} />
```
debitan, line 306: I don't think we will need the `csrfToken` once we're going through the `api-proxy`

src/apps/my-pipeline/controllers/edit.js
adamledwards: That change is not on the my-feature branch yet shall we rebase?

debitan: Good point. Yeah, I'll rebase Graham's `my-pipeline` branch which should be fine as no one has started reviewing it yet. Then we can rebase this one onto that again.

src/apps/my-pipeline/client/AddToPipelineForm.jsx
```diff
@@ -155,4 +122,12 @@ function AddToPipelineForm({
...
 }
 
-export default connect(state2props)(AddToPipelineForm)
+AddToPipelineForm.propTypes = {
+  companyId: PropTypes.string,
+  companyName: PropTypes.string,
+  csrfToken: PropTypes.string,
+  pipelineStatus: PropTypes.object,
+  savedId: PropTypes.string,
+}
```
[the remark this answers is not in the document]
adamledwards: None of them are required for the component to load. On mount these are not available because the data is async.

src/apps/my-pipeline/client/EditPipelineForm.jsx
adamledwards: This is a note to upgrade path. Because the page is reloaded the state is cleared. In a SPA the state will remain so a clean up task is required.

src/apps/my-pipeline/client/EditPipelineForm.jsx
adamledwards: all the data is async so none of it is there on mount

src/apps/my-pipeline/client/AddToPipelineForm.jsx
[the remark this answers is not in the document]
adamledwards: I'll be setting the default prop to null or undefined so I don't think there is much point.

src/apps/my-pipeline/client/EditPipelineForm.jsx
adamledwards: Will do it the cleaner way. :)

src/apps/my-pipeline/client/index.jsx
adamledwards: Wouldn't that make props undefined?

src/apps/my-pipeline/client/constants.js
```diff
@@ -1,5 +1,11 @@
 import PropTypes from 'prop-types'
 
+export const URL_MAP = {
+  leads: 'index',
+  in_progress: 'active',
+  win: 'won',
+}
+
```
debitan, line 8: I've got a similar constants mapping in my PR. I can probably put up a PR to use this abstracted one once we are all merged in.

PR #2693, opened 2020-05-20
Fix setting a default theme when editing interactions

rafenden:
## Description of change

In some cases the `theme` field for interactions is not filled, if this happens we should use the `Other` theme by default.

## Test instructions

1. Edit interaction that doesn't have `theme` field filled.
2. The form should not display the `theme` and `kind` step.

### DIFF
--- a/docs/Running tests.md
+++ b/docs/Running tests.md
@@ -3,7 +3,7 @@
 ## Creating Docker container for CircleCI
 
 ```bash
-export VERSION=3.4.0 # Increment this version each time when you edit Dockerfile.
+export VERSION=3.5.0 # Increment this version each time when you edit Dockerfile.
 
 docker login # Ask webops for Docker Hub access to the ukti group.
 docker build -f test/Dockerfile -t data-hub-frontend-test .
@@ -60,9 +60,10 @@ Sandbox is as a light replacement for API backend and it's used only by function
 1. Install sandbox, for more info see [instructions](https://github.com/getsandbox/sandbox)
 
 ```bash
-wget https://s3.amazonaws.com/sandbox-binaries/runtime-binary.tar \
-  && tar -C /usr/local/bin -xzvf runtime-binary.tar \
-  && rm runtime-binary.tar
+wget https://github.com/getsandbox/sandbox/releases/download/1.1.0/sandbox-runtime.tar \
+  && tar -C /usr/local/bin -xzvf sandbox-runtime.tar \
+  && mv /usr/local/bin/sandbox-runtime /usr/local/bin/sandbox \
+  && rm sandbox-runtime.tar
 ```
 
 2. Start sandbox on port `8001`:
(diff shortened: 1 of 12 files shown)


test/sandbox/polyfills.js
```diff
@@ -0,0 +1,8 @@
+if (!Object.assign) {
+  Object.defineProperty(Object, 'assign', {
+    enumerable: false,
+    configurable: true,
+    writable: true,
+    value: _.assign,
+  })
+}
```
adamledwards, line 8: 👍

docs/Running tests.md
```diff
@@ -60,9 +60,10 @@ Sandbox is as a light replacement for API backend and it's used only by function
 1. Install sandbox, for more info see [instructions](https://github.com/getsandbox/sandbox)
 
 ```bash
-wget https://s3.amazonaws.com/sandbox-binaries/runtime-binary.tar \
-  && tar -C /usr/local/bin -xzvf runtime-binary.tar \
-  && rm runtime-binary.tar
+wget https://github.com/getsandbox/sandbox/releases/download/1.1.0/sandbox-runtime.tar \
```
adamledwards, line 63: We should update docker image with the same version at some point

test/functional/cypress/specs/interaction/interaction-details-form-spec.js
```diff
@@ -880,3 +881,19 @@ describe('Filtering services based on theme & kind', () => {
...
+
+describe('Editing an interaction without a theme', () => {
+  it('should set a default theme (Other)', () => {
+    cy.visit(urls.interactions.edit(interactionWithoutTheme.id))
+
+    cy.get('#field-subject').then((element) =>
+      assertFieldInput({
+        element,
+        label: 'Subject',
+        value: interactionWithoutTheme.subject,
```
adamledwards, line 893: I think you have a string vs int issue in your functional test.

docs/Running tests.md
rafenden: Good idea, the Dockerfile was updated in all places.

test/functional/cypress/specs/interaction/interaction-details-form-spec.js
rafenden: Can you elaborate on that?

docs/Running tests.md
rafenden: Thanks, done!

test/functional/cypress/specs/interaction/interaction-details-form-spec.js
adamledwards: Sorry the problem is not on your test it was a bug by a previous dev. you fixed the incorrect usage of the  nullish coalescing  here test/functional/cypress/support/assertions.js but I think we will need to add here 
https://github.com/uktrade/data-hub-frontend/blob/0043f11597b759e663cf348ced230ce038ff0540/test/functional/cypress/specs/companies/edit-spec.js#L272-L275

rafenden: This is now fixed, thanks!

PR #2709, opened 2020-05-29
Push referral info to Google Tag Manager when sending a referral

mforner13:
## Description of change

This PR pushes information about a referral to the Google Tag Manager 'dataLayer', to be used in Google Analytics, on submit of the send referral form. This will allow more in-depth analysis of the teams who are referring companies to each other and the content of referrals. 

A new saga has been created in order to do this, as the `Task` saga is designed for asynchronous calls and returns a promise which is not necessary for pushing to the data layer. As there was previously only one saga passed to `sagaMiddleware`, a `rootSaga` has been created which combines multiple sagas and then is passed to `sagaMiddleware`.

The component `TagManagerDataLayer` is designed to take components which trigger a push to the data layer as children, and passes its children a function which fires the appropriate action. This approach was chosen rather than having each component dispatch the `TAG_MANAGER_DATA_LAYER__PUSH` action itself, as it removes the need for these components to connect to the redux store unnecessarily. 

Only the team name of the sending and receiving advisers is pushed to Tag Manager to avoid any personal identifiable information being sent. In addition to the subject of the referral. 

## Test instructions

If you have access to Tag Manager:
- Go into preview mode
- Go to a company
- Fill out the send referral form
- Cmd or ctrl click on 'Send referral'
- Look in the 'Data Layer' tab of the Tag Manager preview window
- Check that the correct data is under the 'send_referral' event

## Screenshots

No visual changes


src/apps/companies/apps/referrals/send-referral/client/tasks.js
```diff
@@ -0,0 +1,17 @@
...
+  const recevingAdviserTeamName = adviser.name.split(',')[1]
+
+  window.dataLayer = window.dataLayer || []
+  window.dataLayer.push({
+    event: 'send_referral',
+    sendingAdviserTeam: sendingAdviserTeamName,
+    receivingAdviserTeam: recevingAdviserTeamName,
+    referralSubject: subject,
+    referralNotes: notes,
+  })
```
web-bert, line 16: Couldn't this just go in the `onSubmit` in `src/apps/companies/apps/referrals/send-referral/client/SendReferralConfirmation.jsx`

There is no async here so what is the need for a task?

mforner13: So my understanding was that all side effects should be handled by saga, so I moved it to a task.

web-bert: Good point, still getting used to thinking that way :-) 

I would rather see this task as `addToDataLayer` (or something generic) so that it can be re-used and then this logic can mostly be in `SendReferralConfirmation.jsx` and the logic here can simply be:
```js
export function addToDataLayer(data){
  window.dataLayer = window.dataLayer || []
  window.dataLayer.push(data)
}
```

As I assume events will rarely be re-used?

mforner13: A good idea @web-bert! And I'll move it to a new src/client/tasks folder 👍

src/client/tasks/addToDataLayerTask.js
```diff
@@ -0,0 +1,6 @@
+export const TASK_ADD_TO_DATA_LAYER = 'TASK_ADD_TO_DATA_LAYER'
+
+export function addToDataLayer(data) {
```
web-bert, line 3: Nice, the name of the file should be `add-to-data-layer.js` to follow conventions and remove the duplicate task (as it is already in the tasks folder)

src/apps/companies/apps/referrals/send-referral/client/SendReferralConfirmation.jsx
```diff
@@ -20,58 +23,84 @@ const SendReferralConfirmation = ({
...
+              <form
+                method="post"
+                onSubmit={() =>
+                  addToDataLayer.start({
+                    payload: {
+                      event: 'send_referral',
+                      sendingAdviserTeam: sendingAdviserTeamName,
+                      receivingAdviserTeam: recevingAdviserTeamName,
+                      referralSubject: subject,
+                      referralNotes: notes,
```
web-bert, line 79: I would question the need for the notes to be sent to GA, seems a bit of a privacy concern to me?

mforner13: @web-bert agreed by Billy, have removed notes from the payload.

src/apps/companies/apps/referrals/send-referral/client/SendReferralConfirmation.jsx
```diff
@@ -20,58 +23,83 @@ const SendReferralConfirmation = ({
...
+                onSubmit={() =>
+                  addToDataLayer.start({
+                    payload: {
+                      event: 'send_referral',
+                      sendingAdviserTeam: sendingAdviserTeamName,
+                      receivingAdviserTeam: recevingAdviserTeamName,
+                      referralSubject: subject,
+                    },
+                  })
+                }
```
rafenden, line 81: IMO, we should take a slightly different approach to analytics, instead of using tasks we should use Redux middleware

```
// We dispatch an event in our submit (replace `dispatch` with a named function in `dispatch2props`
onSubmit={()=> dispatch({ type: SEND_REFERRAL_CONFIRMATION, ...})}

// And then in our middleware /src/client/middleware/analytics.js we can send the tracking
function analyticsMiddleware(store, next, action) {
  switch (action.type) {
    case SEND_REFERRAL_CONFIRMATION:
      addToDataLayer(action)
  }
}
```

mforner13: Ahh interesting. What do you think the benefits of this approach over using a task is?

web-bert: A good point for discussion, my concern with this approach is the analytics becomes detached from the main feature. What if the feature is changed and you forget about the analytics, we would need to ensure both are kept in sync somehow?

rafenden: @web-bert [according to the Redux middleware docs](https://redux.js.org/advanced/middleware/):
> It provides a third-party extension point between dispatching an action, and the moment it reaches the reducer.

Which seems like exactly our use case as in the majority of places analytics will be just a side effect to the main action and should go through reducers as no state is changed.

mforner13: @web-bert @rafenden I have just spoken to Peter about this. So redux-saga is what we use to write middleware so I don't think we'd want to introduce the redux middleware mentioned above too. Peter did say however that a Task is unnecessary for adding data to tag manager, as it returns a promise and is designed more for API calls. He has advised creating a new saga for adding data to tag manager, which I will work on now.

web-bert: Sounds good!

src/apps/companies/apps/referrals/send-referral/client/SendReferralConfirmation.jsx
```diff
@@ -10,6 +10,7 @@ import SecondaryButton from '../../../../../../client/components/SecondaryButton
 import LocalHeader from '../../../../../../client/components/LocalHeader/LocalHeader'
 import { Main } from '../../../../../../client/components/'
 import { companies, dashboard } from '../../../../../../lib/urls'
+import TagManagerDataLayer from '../../../../../../client/components/TagManagerDataLayer/index.jsx'
```
web-bert, line 13: Optional: Rename this as it is actually for analytics not tag manager, so it should be `AnalyticsDataLayer` however I wonder if we even need to expose that it uses a "data layer" at all any maybe just `Analytics` would be enough here?

src/client/components/TagManagerDataLayer/index.jsx
```diff
@@ -0,0 +1,19 @@
+import { connect } from 'react-redux'
+import { TAG_MANAGER_DATA_LAYER__PUSH } from '../../actions'
+
+/* This component wraps around a child component from which you want to push
+data to the Google Tag Manager data layer. It passes an arrow function to the 
+child component which when called dispatches the TAG_MANAGER_DATA_LAYER__PUSH 
+action and passes the data intended for the GTM data layer to the 
+TagManagerDataLayer saga. */
+
+const TagManagerDataLayer = connect()(({ children, dispatch }) =>
```
rafenden, line 10: I'm not sure what's the advantage of having a component where just a simple function could do the trick? e.g.

```
function sendTracking(data) {
    dispatch({
      type: TAG_MANAGER_DATA_LAYER__PUSH,
      data,
    })
}
```

This way we don't need to make so many levels of nesting, i.e.

```
<TagManagerDataLayer>
  {(push) => (
    <button onClick={push}>Save</button>
  )}
</TagManagerDataLayer>
```

VS

```
<button onClick={sendTracking({ event: 'save_something' })}>Save</button>
```

mforner13: A good point @rafenden! Would that function sit in src/client/utils you think?

mforner13: Ignore that, I think it should go in the same folder as the saga.

mforner13: As you can only dispatch an action from a connected component, the component was decided to be the simpler way after all.

PR #2711, opened 2020-05-29
Additional fields for pipeline items

adamledwards:
## Description of change

This PR surfaces the following information in the pipeline lists.
- Project sector
- Company contact
- Potential export value
- Expected date for win

The meta fields will be hidden if the backend returns null for the fields. This is becuse the fields are optional.
## Test instructions
As the form is not ready to insert the additional fields the first test should be to see if the list will render correctly without the data. This can be down on the Heroku deployment env. To test the data would will need to use the sandbox environment. 

### Before
image

### After
image

_Add a screenshot_

## Checklist

### DIFF
--- a/src/client/components/Pipeline/PipelineItem.jsx
+++ b/src/client/components/Pipeline/PipelineItem.jsx
@@ -1,70 +1,143 @@
 import React from 'react'
+import moment from 'moment'
 import styled from 'styled-components'
-import PropTypes from 'prop-types'
+import { PipeLineItemPropType } from './types'
 import Button from '@govuk-react/button'
 import Link from '@govuk-react/link'
 import GridRow from '@govuk-react/grid-row'
 import GridCol from '@govuk-react/grid-col'
 import { SPACING, MEDIA_QUERIES } from '@govuk-react/constants'
-import { BLUE } from 'govuk-colours'
-import Card from 'data-hub-components/dist/activity-feed/activities/card/Card'
-import CardHeader from 'data-hub-components/dist/activity-feed/activities/card/CardHeader'
+import { BLUE, GREEN } from 'govuk-colours'
+import {
+  Card,
+  CardDetailsList,
+  CardHeader,
+} from 'data-hub-components/dist/activity-feed/activities/card'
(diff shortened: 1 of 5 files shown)


src/client/components/Pipeline/PipelineItem.jsx
```diff
@@ -1,70 +1,126 @@
...
+}) => (
   <Card>
     <CardHeader
-      company={{ name: projectName }}
+      badge={
+        LIKELIHOOD_TO_WIN[likelihood_to_win] && {
+          borderColour: GREEN,
+          text: `Likelihood to succeed - ${LIKELIHOOD_TO_WIN[likelihood_to_win]}`,
+        }
+      }
```
debitan, line 99: I think it might be better if we move the badge out of the header and into the `StyledGridRow`. Otherwise we get a chunk of whitespace between the header of the metadata.

debitan: I had code working like this when I was prototyping this out before. Happy to share if you'd like?

adamledwards: Do you mean like this so the Badge is in the same row as the metadata ?
image

debitan: Yep, exactly.

adamledwards: 👍  sure

test/functional/cypress/specs/pipeline/my-pipeline-spec.js
```diff
@@ -2,7 +2,72 @@ const urls = require('../../../../../src/lib/urls')
 const leads = require('../../../../sandbox/fixtures/v4/pipeline-item/leads.json')
 const inProgress = require('../../../../sandbox/fixtures/v4/pipeline-item/in-progress.json')
 const win = require('../../../../sandbox/fixtures/v4/pipeline-item/win.json')
+const { NumberUtils } = require('data-hub-components')
 
+function assetPipelineItem(
```
web-bert, line 7: Typo, this should be called `assertPipelineItem`

PR #2719, opened 2020-06-01
Add new fields to pipeline form

web-bert:
## Description of change

Add likelihood to export, export sector, company contact and potential export value.
Pass list of sectors from controller to add and edit components
The add controller already has access to the company contacts, so just pass them through.
The edit controller does not have access to contacts, so I updated the API proxy rules to allow
me to fetch them as part of the task that gets the pipeline data for editing.

Refactor tests to share selectors

## Test instructions

Four extra fields when adding a company to a pipeline and the same extra fields when editing a company on a pipeline.

### Before

Screenshot 2020-06-01 at 20 26 27

### After

Screenshot 2020-06-01 at 20 28 31

## Checklist


src/apps/my-pipeline/controllers/edit.js
```diff
@@ -1,13 +1,22 @@
...
+
+async function renderEditPipeline(req, res, next) {
+  try {
+    const { pipelineItemId } = req.params
+    const { token } = req.session
+
+    res
+      .breadcrumb('Edit your pipeline')
+      .render('my-pipeline/views/pipeline-form', {
+        heading: `Edit your pipeline`,
```
rafenden, line 11: Not a biggie but template literals are not needed here.

src/apps/my-pipeline/client/tasks.js
```diff
@@ -1,5 +1,26 @@
...
+import axios from 'axios'
+
+function transformValuesForApi(values, oldValues = {}) {
+  const data = {
+    name: values.name,
+    status: values.category,
+  }
+
+  function addValue(key, value) {
+    if (oldValues[key] || value) {
```
rafenden, line 12: What's the purpose of `oldValues` as it's not being used in the end?

test/functional/cypress/specs/pipeline/my-pipeline-edit.js
```diff
@@ -4,11 +4,14 @@ const {
...
   assertFieldInput,
+  assertFieldTypeahead,
 } = require('../../support/assertions')
 
 const selectors = require('../../../../selectors')
 
-const pipelineItem = pipelineItemLambdaPlc.results[0]
+const formSelectors = selectors.pipelineForm
+const firstItem = pipelineItemLambdaPlc.results[0]
+const secondItem = pipelineItemLambdaPlc.results[1]
```
rafenden, line 14: Optional: These could be defined in a single line.
```suggestion
const [firstItem, secondItem] = pipelineItemLambdaPlc.results
```

test/sandbox/routes/v4/pipeline-item/index.js
```diff
@@ -40,34 +40,46 @@ exports.createUpdatePipelineItem = function(req, res) {
...
     res.json(
       _.assign({}, pipelineCreate, {
         status: 'leads',
       })
     )
     return
   }
-  if (req.params.pipelineItemId === 'IN_PROGRESS') {
+
+  if (pipelineId === 'IN_PROGRESS') {
```
rafenden, line 66: We should use constants for pipeline IDs as they are used across files.

src/apps/my-pipeline/client/tasks.js
web-bert: When editing It is to signify if there was a value before, so that if there is not a new value it should still be added to the `body` so that the value can be removed, maybe this would be better as `const hasExistingValue = !!oldValues[key]`

web-bert:
@rafenden updates have been made, apart from the const for the tests as I'm not sure where these should go and seeing as they are tests I didn't deem it important enough to bother. The tests would fail if the match did not work

src/apps/my-pipeline/client/PipelineForm.jsx
```diff
@@ -2,38 +2,89 @@ import React from 'react'
...
-        name="category"
         label="Choose a status"
+        name="category"
         required="Choose a status"
-        options={options}
+        options={statusOptions}
         initialValue={initialValue.category}
       />
+      <FieldRadios
+        label="Likelihood to export (Optional)"
```
rafenden, line 47: Usually, the word `Optional` starts with lowercase letter

web-bert: Good spot, I just followed the design but I should have checked existing labels! Fixed now

PR #2724, opened 2020-06-03
Replace govuk-react Main component with FE version

ian-leggett:
## Description of change

As we now have a `<Main>` React component that outputs the correct `<main>` HTML element we should replace all use cases of the govuk-react version.

## Test instructions

1. Edit a One list and everything should look the same
2. Send a referral flow - everything should look the same
3. View the Investment project admin tab - everything should look the same

## Checklist


ian-leggett:
> It looks like `CompanyLocalHeader.jsx` and `LocalHeader.jsx` use govuk/main too. Do they need replacing as well?

I saw these, they do need replacing but not with a <main> HTML tag as there can only be one <main>on the page. These should be replaced by a generic container tag that also uses the grid with the right gutters and responsive behaviour.

PR #2725, opened 2020-06-04
Remove archived contacts from interaction form typeahead

mforner13:
## Description of change

Feedback has come through Zendesk that archived contacts are appearing in the Contacts typeahead options when adding an interaction. 

This PR adds a filter in the interactions controller to remove archived contacts before passing the prop to the form.

Whilst testing, i noticed that the 'Archive' button on the contacts page was using the class `govuk--button` rather than `govuk-button` meaning that it wasn't rendering the correct styling, so I have updated this as well. 

## Test instructions

- Go to a company
- Find a contact
- Archive that contact
- Check that the correct styling appears on the Archive button
- Go to the add interaction form
- Check that the archived contact does not appear in the Contacts typeahead

### Before
**Archive button**
Screenshot 2020-06-04 at 12 43 50

### After
**Archive button**
Screenshot 2020-06-04 at 12 54 46

## Checklist

### DIFF
--- a/test/functional/cypress/specs/interaction/interaction-details-form-spec.js
+++ b/test/functional/cypress/specs/interaction/interaction-details-form-spec.js
@@ -357,6 +357,15 @@ describe('Interaction theme', () => {
       ])
     })
 
+    it('should only show non-archived contacts in the contacts field', () => {
+      cy.contains(ELEMENT_CONTACT.label)
+        .next()
+        .next()
+        .click()
+        .find('> div > div > div')
+        .should('have.text', '-- Select contact --Johnny Cakeman')
+    })
+
     it('should validate the form', () => {
       cy.contains('button', 'Add interaction').click()
       cy.contains('h2', 'There is a problem')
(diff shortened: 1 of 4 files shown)


web-bert:
@mforner13 might be good to add a test for this too?

test/functional/cypress/specs/interaction/interaction-details-form-spec.js
```diff
@@ -357,6 +357,14 @@ describe('Interaction theme', () => {
       ])
     })
 
+    it('should not show archived contacts in the contacts field', () => {
+      cy.contains(ELEMENT_CONTACT.label)
+        .next()
+        .next()
+        .click()
+      cy.contains('div', 'Archived Adviser').should('not.be.visible')
```
web-bert, line 365: This should be `cy.contains('div', 'Archived Contact')`, if the list of contacts is small enough it might be best to assert that the list contains only the active contacts?

mforner13: Oo thanks, I kept getting contact and adviser mixed up and forgot to change this! But have changed the test to check what is there rather than what isn't as you said :)

web-bert: LGTM 👍

PR #2728, opened 2020-06-05
Add e2e test to cover missing UK Region

web-bert:
## Description of change

When working on the Export tab we noticed a scenario where a company can have the `uk_region` set to `null` and when trying to send a `PATCH` to `/v4/company/<company_id>` it would return an error. We stopped the error from happening and this PR adds a test to cover that scenario so that we do not regress.

## Test instructions

All e2e tests should pass

## Checklist


web-bert:
@mforner13 yes it needs the correct branch, which I had on the Draft PR to verify the tests pass. Once the API has a release done I will re-run the tests and they will pass.

PR #2735, opened 2020-06-08
Allow users with right permission to add lead ITA

mforner13:
## Description of change

Previously it was the purple team's intention that any adviser could add a Lead ITA to a company. It has now been decided that only those with the `company.can_add_regional_account_manager` permission can do this. 

This PR adds a condition so that only users with this permission can view the 'Add a lead ITA' button. It also updates the content which appears in the Lead Adviser tab to reflect these changes.

## Test instructions

- Go to the lead adviser tab of a non-One list tier company
- Check that the content is updated
- Check you can view the button if you have the correct permission in the API you're pointing to

### Before
Screenshot 2020-06-08 at 12 19 01

### After
image

## Checklist


mforner13:
> I feel there should be a test to cover this scenario, but that might be a discussion for the dev huddle on how we achieve that.

@web-bert Yeah I agree! It seems we'd need a way to point to another whoami.json.

PR #2736, opened 2020-06-08
Refactor flash messages to check session storage and set messages with body

ian-leggett:
## Description of change
There are two ways you can set flash messages, one is through the node layer and the other is via session storage. The later is done when we are using the likes of a saga tasks in React and submitting forms client side. Currently we are not checking session storage if the flash messages object is empty, this PR fixes this and simplifies the utility function used to set messages in React components.

## Test instructions
Nothing should change as we are not currently using these new methods but see Storybook for an example of what this looks like.

## Screenshots
Screenshot 2020-06-08 at 15 36 47

## Checklist

### DIFF
--- a/src/client/components/LocalHeader/FlashMessages.jsx
+++ b/src/client/components/LocalHeader/FlashMessages.jsx
@@ -55,9 +55,11 @@ const FlashMessages = ({ flashMessages }) => {
       {Object.entries(
         !isEmpty(flashMessages) ? flashMessages : flashMessagesFromStorage
       ).map(([type, messages]) => {
-        // Example "success:with-body" -  If the string you pass in the first argument "type"
-        // has a colon then the message argument accepts two props in an object, one for the heading
-        // and one for the body. The first part of the string is used to indicate colour, success - green, info - blue
+        /*
+        Example "success:with-body" -  If the string you pass in the first argument "type"
+        has a colon then the message argument accepts two props in an object, one for the heading
+        and one for the body. The first part of the string is used to indicate colour, success - green, info - blue
+        */
         const parts = String(type).split(':')
         return parts.length > 1
           ? messages.map(({ body, heading }) => (
--- a/src/client/components/LocalHeader/FlashMessages.jsx
+++ b/src/client/components/LocalHeader/FlashMessages.jsx
@@ -1,6 +1,7 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 import styled from 'styled-components'
+import { isEmpty } from 'lodash'
 
 import {
   ERROR_COLOUR,
@@ -14,7 +15,7 @@ import { FONT_WEIGHTS } from '@govuk-react/constants/lib/typography'
 import { FONT_SIZE, SPACING } from '@govuk-react/constants'
 import UnorderedList from '@govuk-react/unordered-list'
 import { StatusMessage } from 'data-hub-components'
-import flashUtils from '../../utils/flash-messages'
+import { getMessages, clearMessages } from '../../utils/flash-messages'
 
 const StyledBody = styled('p')`
   margin-bottom: 0;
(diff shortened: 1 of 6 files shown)


src/client/components/LocalHeader/FlashMessages.jsx
```diff
@@ -49,39 +50,35 @@ const messageColours = {
...
-  }
-  return null
+  return !isEmpty(flashMessages) || flashMessagesFromStorage ? (
+    <UnorderedList listStyleType="none" data-auto-id="flash">
+      {Object.entries(
+        !isEmpty(flashMessages) ? flashMessages : flashMessagesFromStorage
+      ).map(([type, message]) => {
+        const parts = String(type).split(':')
+        return parts.length > 1
+          ? message.map((message) => (
```
rachaelcodes, line 60: Might be good to change one of the `message`s here to make it a bit clearer to read?

src/client/components/LocalHeader/FlashMessages.jsx
```diff
@@ -49,39 +50,35 @@ const messageColours = {
...
-      </UnorderedList>
-    )
-  }
-  return null
+  return !isEmpty(flashMessages) || flashMessagesFromStorage ? (
+    <UnorderedList listStyleType="none" data-auto-id="flash">
+      {Object.entries(
+        !isEmpty(flashMessages) ? flashMessages : flashMessagesFromStorage
+      ).map(([type, message]) => {
+        const parts = String(type).split(':')
```
rachaelcodes, line 58: Is it worth adding a comment here to explain what is happening with the ':' (as it took us a bit of time to work out why it was there!)

src/client/utils/flash-messages.js
```diff
@@ -29,6 +29,18 @@ function addSuccessMessage(message) {
...
+}
+
+function addInfoMessage(message) {
+  addMessage('info', message)
+}
+
+function addInfoMessageWithBody(message) {
+  addMessage('info:with-body', message)
+}
+
```
debitan, line 43: we discussed the use of these helper functions. Really you can do everything with the `addMesssage` function and all these are doing is filling in the first argument for you. 

we also discussed a way to make these smarter by not requiring you to know the inner workings of the `FlashMessages` component. I.e. you shouldn't have to know to pass an array with `heading` and `body` as the `message` in order for it to show the header body style. Maybe the `addMessages` function can look to see if there is 2 or 3 arguments and treat them as `type` and `heading` or `type`, `heading` and `body` respectively.

src/client/components/LocalHeader/FlashMessages.jsx
ian-leggett: Agree, comment added

src/client/components/LocalHeader/FlashMessages.jsx
ian-leggett: I have now deconstructed both now and they are much clearer, see what you think?

ian-leggett:
> The main thing in this PR is the fix for checking if the objects coming into the FlashMessages component are empty or not so it knows to look in both places if the node layer is also passing an empty object through. So the description and title needs updating to reflect this. The `FlashMessages` component can already use header and body, so currently it is a bit confusing.

Yeah totally agree, I have changed the title and description as this wasn't clear on what I am trying to achieve here. 👍

src/client/utils/flash-messages.js
ian-leggett: I agree, I have refactored this file to use one function called `addMessage` which can be used for all messages. I have also updated the error messages it logs to help for clearer debugging.

src/client/components/LocalHeader/FlashMessages.jsx
```diff
@@ -1,6 +1,7 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 import styled from 'styled-components'
+import { isEmpty } from 'lodash'
```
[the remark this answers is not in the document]
ian-leggett: Following our conversation, I would prefer to remove this dependancy and go vanilla but this feels like a much better solution than the vanilla options that are out there. Using lodash's `isEmpty` feels like the way to go.

src/client/utils/flash-messages.js
```diff
@@ -1,41 +1,34 @@
...
       // eslint-disable-next-line no-console
-      console.log(e) // TODO: replace with more robust error logging when implemented
+      console.error('Cannot get messages from session storage', e)
     }
   }
   return {}
 }
 
-function addMessage(messageType, message) {
+const addMessage = (messageType, message) => {
```
mforner13, line 16: It feels strange that a user of this function needs to know that they have to pass 'success:with-body' in order to get the correct format. I think either this could be made clearer in the utils file (with a comment similar to in `Flashmessages.jsx`) or we could have a separate addMessageWithBody function which adds that suffix on.

mforner13, line 16: The issue with the user of the function needing to format the message in the correct way ([in Dave's comment](https://github.com/uktrade/data-hub-frontend/pull/2736#discussion_r436819027))seems to still persist too. Can we allow this function (or just the addMessageWithBody function if you chose to make that) to take 3 string arguments (status, heading and body) and then format them correctly before adding them to session storage?

ian-leggett: Following our discussion I have added `addMessageWithBody` this is much clearer 👍

PR #2739, opened 2020-06-08
Forward Zipkin headers for correlation tracing

rafenden:
## Description of change

Preserve Zipkin headers added by Cloud Foundry to allow for correlation tracing.

https://docs.cloudfoundry.org/concepts/http-routing.html#zipkin-headers

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "develop": "NODE_ENV=development run-p watch:js:*",
     "watch:test": "yarn test:unit -- -w",
     "watch:js:client": "webpack --watch --progress",
-    "watch:js:server": "nodemon $NODE_DEBUG_OPTION --inspect --ignore 'src/**/__test__/**/*'",
+    "watch:js:server": "nodemon $NODE_DEBUG_OPTION --inspect-brk=0.0.0.0 --ignore 'src/**/__test__/**/*'",
     "lint": "run-p lint:*",
     "lint:js": "eslint --ext .jsx,.js ./test ./src ./assets ./.circleci",
     "lint:sass": "sass-lint -v -q -c .sass-lint.yml assets/stylesheets/**/*.scss",
--- a/src/lib/get-zipkin-headers.js
+++ b/src/lib/get-zipkin-headers.js
@@ -1,0 +1,9 @@
+const { pickBy } = require('lodash')
+
+const ZIPKIN_HEADERS = ['x-b3-traceid', 'x-b3-spanid']
+
+function getZipkinHeaders(req) {
+  return pickBy(req.headers, (header) => header.toLowerCase() in ZIPKIN_HEADERS)
+}
+
+module.exports = getZipkinHeaders
--- a/src/middleware/__test__/api-proxy.test.js
+++ b/src/middleware/__test__/api-proxy.test.js
@@ -1,0 +1,34 @@
+const express = require('express')
+const request = require('supertest')
+
+const config = require('../../config')
+const apiProxy = require('../../middleware/api-proxy')
+
+describe('API proxy middleware', () => {
+  it('should forward requests to the backend', async () => {
+    const scope = nock(config.apiRoot)
+      .get('/whoami')
+      .reply(200, {})
+
+    const app = express()
+    app.use((req, res, next) => {
+      req.session = { token: '123' }
+      next()
+    })
+    apiProxy(app)
+
+    await request(app)
+      .get('/api-proxy/whoami')
+      .set({
+        'X-B3-TraceId': 'fake-trace-id',
+        'X-B3-SpanId': 'fake-span-id',
+      })
+      .expect(200)
+
+    expect(scope.interceptors[0].req.headers).to.contains({
+      authorization: 'Bearer 123',
+      'x-b3-spanid': 'fake-span-id',
+      'x-b3-traceid': 'fake-trace-id',
+    })
+  })
+})
--- a/src/middleware/api-proxy.js
+++ b/src/middleware/api-proxy.js
@@ -1,6 +1,7 @@
 const { createProxyMiddleware } = require('http-proxy-middleware')
 
-const config = require('../config/')
+const config = require('../config')
+const getZipkinHeaders = require('../lib/get-zipkin-headers')
 
 const API_PROXY_PATH = '/api-proxy'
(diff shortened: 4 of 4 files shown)


src/middleware/globals.js
```diff
@@ -0,0 +1,5 @@
+module.exports = function locals(req, res, next) {
+  global.currentRequest = req
```
web-bert, line 2: This is too much to put in the global scope, it should be the minimal amount possible which is the output of `getZipkinHeaders`. I'm not even sure this would work as a solution with multiple concurrent requests, we need some testing to confirm the validity of it, as we don't want to be sending mixed up headers as it would make log tracing unusable.

web-bert: Also I think it is preferred to pass through any header that starts `x-b3-` check with @munhitsu

rafenden: >This is too much to put in the global scope

Global vars are acceptable in some cases, could you suggest an alternative method? My first idea was to update `authorisedRequest` to accept additional `res` param but going forward I would have to edit hundreds of `authorisedRequest` implementations.

>Also I think it is preferred to pass through any header that starts x-b3- check with @munhitsu 

Not according to the [docs](https://docs.cloudfoundry.org/concepts/http-routing.html#zipkin-headers).

src/lib/get-zipkin-headers.js
```diff
@@ -0,0 +1,8 @@
+const { pickBy } = require('lodash')
+
+function getZipkinHeaders(req) {
+  const zipkinHeaders = ['x-b3-traceid', 'x-b3-spanid']
```
web-bert, line 4: `zipkinHeaders` should be defined outside of the function

src/lib/get-zipkin-headers.js
```diff
@@ -0,0 +1,8 @@
...
+  return pickBy(req.headers, (header) => header.toLowerCase() in zipkinHeaders)
```
web-bert, line 5: Given there are less zipkin headers than other ones, it would be more efficient to just pick out the zipkin headers and return them (given this will be used on a large proportion of requests).

src/middleware/api-proxy.js
```diff
@@ -33,6 +34,10 @@ module.exports = (app) => {
         ['^' + API_PROXY_PATH]: '',
       },
       onProxyReq: (proxyReq, req) => {
+        Object.entries(getZipkinHeaders(req)).map(([header, headerValue]) => {
+          proxyReq.setHeader(header, headerValue)
```
web-bert, line 38: This should be `forEach` instead of `map` and there is no return value

src/middleware/api-proxy.js
```diff
@@ -33,6 +34,10 @@ module.exports = (app) => {
         ['^' + API_PROXY_PATH]: '',
       },
       onProxyReq: (proxyReq, req) => {
+        Object.entries(getZipkinHeaders(req)).map(([header, headerValue]) => {
```
web-bert, line 37: Given that `getZipkinHeaders` is only used once, it may be easier to change the return format based on my previous comment so that this can be a more efficient function

src/lib/get-zipkin-headers.js
rafenden: That's what the `pickBy` function is doing, additionally it ensures the text case is ignored.

web-bert: pickBy` is looping and checking each header on the `req` where as you could just check for the existence of the two you need

src/middleware/api-proxy.js
rafenden: I prefer to return object of related key-value pairs because it's easier to reason about.

web-bert: Fair enough, although it might be worth considering prioritising efficiency here given the amount of calls this code will run on.

src/lib/get-zipkin-headers.js
```diff
@@ -0,0 +1,9 @@
+const { pickBy } = require('lodash')
+
+const ZIPKIN_HEADERS = ['x-b3-traceid', 'x-b3-spanid']
```
mcbhenwood, line 3: Shouldn't the downstream header names be cased as per `X-B3-TraceId`, `X-B3-SpanId` ? This solution using `pickBy()` will result in headers cased specifically as here in this const. I think it would be better to put the fully cased header names in here and update `getZipkinHeaders()` function to do ignore all casing in the comparison function.

mcbhenwood:
As per our chat earlier, this is just a partial solution for the overall requirement. We can implement it but the rest of the task will need to be addressed at some point soon.

Happy for merge once the above review point is addressed.

src/middleware/__test__/api-proxy.test.js
```diff
@@ -0,0 +1,34 @@
...
+      .set({
+        'X-B3-TraceId': 'fake-trace-id',
+        'X-B3-SpanId': 'fake-span-id',
+      })
+      .expect(200)
+
+    expect(scope.interceptors[0].req.headers).to.contains({
+      authorization: 'Bearer 123',
+      'x-b3-spanid': 'fake-span-id',
+      'x-b3-traceid': 'fake-trace-id',
```
mcbhenwood, line 31: My guess on the header casing is based on this test here

mcbhenwood, line 31: Not sure why these header names are not in the original? Does `setHeader()` alter the name to lower case?

rafenden: Headers are case-insensitive
https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2

package.json
```diff
@@ -17,7 +17,7 @@
     "develop": "NODE_ENV=development run-p watch:js:*",
     "watch:test": "yarn test:unit -- -w",
     "watch:js:client": "webpack --watch --progress",
-    "watch:js:server": "nodemon $NODE_DEBUG_OPTION --inspect --ignore 'src/**/__test__/**/*'",
+    "watch:js:server": "nodemon $NODE_DEBUG_OPTION --inspect-brk=0.0.0.0 --ignore 'src/**/__test__/**/*'",
```
web-bert, line 20: Is this mean to be in this PR?

PR #2744, opened 2020-06-12
Change dnb investigation endpoint

rafenden:
## Description of change

Replace the endpoint used to create D&B investigations when company cannot be found on when using the "Add company form".


rafenden:
> Looks good! Should the 4 `.bin.sample` files be part of this though?

I'm just updating the file permissions to allow execution of these files.

PR #2776, opened 2020-06-24
Add mechanism to show account plan link on companies

currycoder:
## Description of change

This PR adds a simple first iteration for surfacing Account Plans in Data Hub.

Our (Blue team) current thinking is that the *simplest thing* is to put a link to an account plan on company pages.  This has been achieved in the first instance without any backend work; there is a map of duns numbers to account plan URLs in the application configuration.  This can be effected by setting a JSON vault variable `ACCOUNT_PLAN_URLS`.

If it becomes clear that this experiment is a success that we want to further iterate, we can look at adding API support for account plan links.

**Note:** The copy here is not final - I expect to make some extra commits for minor amends, but the overall approach is ready for review.

## Screenshots

Screenshot from 2020-06-24 15-40-14

## Checklist

- [X] Has the branch been rebased to master?
- [X] Automated tests (Any of the following when applicable: Unit, Functional or Acceptance)

### DIFF
--- a/src/config/envSchema.js
+++ b/src/config/envSchema.js
@@ -22,6 +22,9 @@ const OAUTH2_URI = Joi.string().uri()
   .when('OAUTH2_BYPASS_SSO', { is: false, then: Joi.required() })
 
 const envSchema = Joi.object({
+  // A mapping of { duns_number: url, .. } for the companies that have account plans
+  ACCOUNT_PLAN_URLS: ExtendedJoi.json().default({}),
+
   // The url for a back end server instance for the service
   API_ROOT: Joi.string().uri().required(),
 
--- a/test/functional/cypress/specs/companies/local-header/account-plan-spec.js
+++ b/test/functional/cypress/specs/companies/local-header/account-plan-spec.js
@@ -1,0 +1,66 @@
+const selectors = require('../../../../../selectors')
+const fixtures = require('../../../fixtures')
+const urls = require('../../../../../../src/lib/urls')
+
+const companyLocalHeader = selectors.companyLocalHeader()
+
+describe('Local header for company with an account plan', () => {
+  context(
+    'when visting a company which has an account plan (no contact email)',
+    () => {
+      before(() => {
+        cy.visit(urls.companies.activity.index(fixtures.company.dnbCorp.id))
+      })
+
+      it('should display the company name', () => {
+        cy.get(companyLocalHeader.companyName).contains(
+          fixtures.company.dnbCorp.name
+        )
+      })
+
+      it('should display the account plan message', () => {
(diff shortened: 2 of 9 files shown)


src/config/envSchema.js
```diff
@@ -22,6 +22,9 @@ const OAUTH2_URI = Joi.string().uri()
   .when('OAUTH2_BYPASS_SSO', { is: false, then: Joi.required() })
 
 const envSchema = Joi.object({
+  // A mapping of { duns_number: url, .. } for the companies that have account plans
+  ACCOUNT_PLAN_URLS: ExtendedJoi.json().default({}),
```
paulgain, line 26: I'm guessing that `ACCOUNT_PLAN_URLS` is a required env var?

This only ensures the JSON provided is valid and errors if it's invalid 👍 , we need an error if the env var hasn't been set. For example, if you comment out an env var such as `DATA_STORE_SERVICE_ACCESS_KEY_ID` and then try to bring up the app you get this error: 
`"DATA_STORE_SERVICE_ACCESS_KEY_ID" [1]: -- missing --`.

test/functional/cypress/specs/companies/local-header/account-plan-spec.js
```diff
@@ -0,0 +1,40 @@
...
+
+    it('should display the company name', () => {
+      cy.get(companyLocalHeader.companyName).contains(
+        fixtures.company.dnbCorp.name
+      )
+    })
+
+    it('should display the account plan message', () => {
+      cy.get(companyLocalHeader.accountPlanMessage).contains(
+        'Go to Sharepoint to view the account plan for DnB Corp (opens in a new window or tab). You might have to request access for the file.'
```
paulgain, line 21: This string is too long and should be broken over two lines. Take a look at the [activity-feed-spec.js](https://github.com/uktrade/data-hub-frontend/blob/master/test/functional/cypress/specs/companies/activity-feed-spec.js) for examples.

src/config/envSchema.js
currycoder: I was hoping that using `.default({})` would mean that when this env var is undefined we would end up with an empty JS object as the default.  I was also hoping this would mean that e.g. for local copies, the app won't fall over on start when this is undefined and we don't need to set this in vault until we have the list together.

What do you think Paul?

paulgain: For the interim that's fine, once we have the list in place then it should be mandatory to define this env var. Historically we struggled bringing up the app (or something would fail in the app once it's up) due to one or more env vars not being set, `envSchema.js` addresses that short coming. We want the app to fall over on start, hence the helpful error message(s) when that happens.

currycoder: Ah great - that's really good context to know thank you.

PR #2762, opened 2020-06-18
Pipeline filter by archive

adamledwards:
## Description of change

This PR add checkbox to the my pipeline tabs to include archive results in the tabs.  This is state of the checkbox is stored in redux and is persistent  between tabs.
The default option is to hide the archive items, this is done by adding `archived=False` to the endpoint and to show all items we omit the filter.

## Test instructions
Add a company to your pipeline, click edit and repace /edit with /archive once a reason has been entered it should successfully POST to the API and then return you to the pipeline. Your pipeline should not show on the list by default. Clicking the `Show archived projects`  checkbox should make your archived item appear.

There will be styling added to differentiate between archived and non-archived items in future PRs 

### Before
image

### After

image

## Checklist

[//]: # "When submitting a PR make sure the code review guidelines have been satisfied.
https://github.com/uktrade/data-hub-frontend/blob/master/docs/Code%20review%20guidelines.md"

- [ ] Has the branch been rebased to master?
- [ ] Automated tests (Any of the following when applicable: Unit, Functional or Acceptance)
- [ ] Manual compatibility testing (Browsers: Chrome, Firefox, IE11, Safari)

PR #2778, opened 2020-06-25
Feature/pipeline sort

adamledwards:
## Description of change

Add Sort-by to my pipelines. This is part of the iteration 3 work for pipeline this works follows the work in https://github.com/uktrade/data-hub-frontend/pull/2762

## Test instructions
If you go on the deployment site you should be able to resort the list to the current sort. The sort by modified will not work with due to a bug caused modified_on record not being updated on solved. This PR solves that bug https://github.com/uktrade/data-hub-api/pull/2994#pullrequestreview-438085791 once this is merged in the sort by recently updated will work.

## Screenshots

image

_Add a screenshot_

### After

image

## Checklist


adamledwards:
> Looks really good and works well. My only question is do we need to do a call to the API every time you change the select? Would it be better to do one call on page load, then sort on the state?

@ian-leggett we could do the sort on the frontend but once we get into pagination its best to keep the sort on the backend.

PR #2779, opened 2020-06-26
Remove post to endpoint that doesn't get used anymore

ian-leggett:
## Description of change
We don't need to have the POST to `/assign-account-manager` end point in node/express anymore as this task has changed and is done in the client side using a Redux Saga task.

## Test instructions

Nothing should change

## Checklist


ian-leggett:
> Can the whole `submit` function not be removed if it's all happening in the task now? And then the `.post` from the router too?

That would be nice but unfortunately the "remove" relies on it. Ideally this needs to go in as a client side task but this would have bloated out the previous PR.

PR #2824, opened 2020-07-14
Mock postcode lookup call to reduce test flakiness

debitan:
## Description of change

This PR mocks out the call the postcode lookup service to avoid intermittent failing tests as shown here - https://dashboard.cypress.io/projects/w97se2/runs/4606/specs

This means that in your sandbox environment you will need to set the following env var:

`DATA_STORE_SERVICE_POSTCODE=http://localhost:8001/sandbox/postcodelookup/`

I have added the necessary env var into the vault so it is no longer hard coded. 

## Test instructions

Tests should all pass and you shouldn't see the add-company.spec failing every now and then because of a slow response from `https://api.getAddress.io/`

## Checklist


test/sandbox/server.js
```diff
@@ -58,6 +58,23 @@ var v4pipelineItem = require('./routes/v4/pipeline-item/index.js')
...
+      'D S D Ltd, 102 Petty France, Westminster, , , London, Greater London',
+      'DESIGN102, 102 Petty France, Westminster, , , London, Greater London',
+      'Her Majesty’s Courts and Tribunals Service, 102 Petty France, Westminster, , , London, Greater London',
+      'Judicial Appointments Commission, 102 Petty France, Westminster, , , London, Greater London',
+      'Legal Aid Agency (London), 102 Petty France, Westminster, , , London, Greater London',
+      'Ministry of Justice, 102 Petty France, , , , London, ',
+      'National Probation Service, 102 Petty France, Westminster, , , London, Greater London',
+    ],
+  })
+})
```
debitan, line 76: I actually intended to abstract this out into its own file as per the rest of the routes in this file, but I forgot. Is it ok like this?

ian-leggett: Yeah feels like it should live outside as a fixture but I don't think its a blocker.

PippoRaimondi: I would recommend abstracting it... if all of the endpoints wouldn't we would end up with a huge file

PR #2858, opened 2020-07-29
Move <FieldInput> element into FE and add filter style...

ian-leggett:
## Description of change

This is work around moving all components from DH Components to DH FE. As I have moving this I am also adding a new prop that can be passed to the component which will toggle styles for when using the form element as a filter in the left aside of the collection list pages.

## Test instructions
Run storybook and you should now see a form section and in there you see the input field examples with a new (reduced) version .

Screenshot 2020-07-29 at 09 28 12

## Checklist

### DIFF
--- a/src/client/components/Form/elements/FieldInput/__stories__/FieldInput.stories.jsx
+++ b/src/client/components/Form/elements/FieldInput/__stories__/FieldInput.stories.jsx
@@ -63,7 +63,6 @@ storiesOf('Forms/Input', module)
             hint="Some hint"
             name="testField"
             required="Enter text"
-            legend={() => <>mooo</>}
             type="text"
             reduced={true}
           />
--- a/src/client/components/Form/elements/FieldInput/__stories__/FieldInput.stories.jsx
+++ b/src/client/components/Form/elements/FieldInput/__stories__/FieldInput.stories.jsx
@@ -1,0 +1,73 @@
+import React from 'react'
+import { addDecorator, storiesOf } from '@storybook/react'
+import { action } from '@storybook/addon-actions/dist'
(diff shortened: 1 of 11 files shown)


src/client/components/Form/elements/FieldInput/__stories__/FieldInput.stories.jsx
```diff
@@ -0,0 +1,73 @@
...
+  .add('FieldInput (reduced)', () => (
+    <FormStateful onSubmit={action('onSubmit')}>
+      {() => (
+        <>
+          <FieldInput
+            label="Text"
+            hint="Some hint"
+            name="testField"
+            required="Enter text"
+            legend={() => <>mooo</>}
```
debitan, line 66: I'm not sure that this does anything here in this example.

src/client/components/Form/elements/FieldInput/__stories__/FieldInput.stories.jsx
```diff
@@ -0,0 +1,73 @@
...
+            type="text"
+            reduced={true}
```
debitan, line 68: I wonder if `reduced` is the right name for this? What is your thinking behind it? My initial feeling was that it should be called `filter` but I guess that doesn't make it that reusable as that is a description of where it is used rather than referring to the difference in styling of this component in isolation.

src/client/components/Form/elements/FieldInput/__stories__/FieldInput.stories.jsx
ian-leggett: Yeah you are right, I should have removed it and I have a typo here thanks

PR #2879, opened 2020-08-03
hotfix: show advisers in interaction form

debitan:
When getting the initial values the user also needs to be in an array for the interaction to save. Hotfix for #2862 

Testing instructions:

- create an interaction with just yourself as adviser, does it save?
- create an interaction with you and some other people as advisers, does it save?
- when clicking back into either of the interaction above do you see the correct advisers in the edit form or are any missing?

### DIFF
--- a/src/apps/interactions/apps/details-form/controllers.js
+++ b/src/apps/interactions/apps/details-form/controllers.js
@@ -150,10 +150,9 @@ const getInitialFormValues = (req, res) => {
       referral && referral.contact
         ? [transformObjectToOption(referral.contact)]
         : [],
-    dit_participants: [
-      (advisers &&
-        advisers.map((adviser) => transformObjectToOption(adviser))) ||
-        transformObjectToOption(user),
+    dit_participants: (advisers &&
+      advisers.map((adviser) => transformObjectToOption(adviser))) || [
+      transformObjectToOption(user),
     ],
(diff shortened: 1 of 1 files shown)


src/apps/interactions/apps/details-form/controllers.js
```diff
@@ -150,10 +150,9 @@ const getInitialFormValues = (req, res) => {
       referral && referral.contact
         ? [transformObjectToOption(referral.contact)]
         : [],
-    dit_participants: [
-      (advisers &&
-        advisers.map((adviser) => transformObjectToOption(adviser))) ||
-        transformObjectToOption(user),
+    dit_participants: (advisers &&
+      advisers.map((adviser) => transformObjectToOption(adviser))) || [
+      transformObjectToOption(user),
```
adamledwards, line 155: in the PR description, you said the user also needs to be in an array for the interaction to save. But it looks like the code is advisers or user, not advisers & users

debitan: yes, that's right. The user is only ever called upon when it is a new interaction and it defaults to putting your name in the adviser field. Otherwise, if you're editing the advisers are already present and you just need to map over them to get their details to display.

adamledwards: fair enough

PR #2880, opened 2020-08-03
Move <FieldDate> to Data Hub FE and create filter version

ian-leggett:
## Description of change

As part of the work around moving Data Components to Data Hub FE this one ports the <FieldDate> and creates a reduced version of it so we can use it as a filter in the collection lists page. As there is not support for the input type="date" IE11 has a fallback of the three input fields (same pattern used in standard forms).

## Test instructions

1. Run storybook and you should see the checkboxes available in the navigation.
2. Check that the reduced versions styles are inline with what we currently on Data Hub.

## Screenshots
Screenshot 2020-08-03 at 15 00 28

**IE11 version**
Screenshot 2020-08-03 at 15 00 35

## Checklist


ian-leggett:
@dependabot rebase

PR #2886, opened 2020-08-07
Redux and Saga documentation

debitan:
High level documentation of how to add a new saga task and accompanying redux actions/reducers


docs/redux-and-saga.md
```diff
@@ -0,0 +1,145 @@
...
+        .get('/api-proxy/v4/pipeline-item', { params: ... })
+        .then(handleSuccess, handleError)}
+```
+
+### 2. state.js
+
+This is where we will namespace our task and also export our state ID.
+
+```
+export const TASK_GET_PIPELINE_LIST = "TASK_GET_PIPELINE_LIST"
```
peterhudec, line 33: I don't think it's necessary to keep the _task names_ in constants and makes the code less readable. The `Task` component will throw a _task not registered_ error in case of a typo.

docs/redux-and-saga.md
```diff
@@ -0,0 +1,145 @@
...
+
+It should follow the pattern of [COMPONENT_NAME]\_\_[ACTION_VERB].
+
+```
+export const PIPELINE__LIST_LOADED = "PIPELINE__LIST_LOADED"
+```
+
+### 5. index.jsx
+
+Now it's time to hook up your task to your React component. You can do this in two ways with our `Task` wrapper - either with `Task.Status` or `getTask`. `Task.Status` gives you a built in loading box and is best used for loading data from the API before the component renders. `getTask` is more versitile, you can add a loading box if you want, but you don't have to. It is best used for tasks like submitting a post request when clicking a form button.
```
peterhudec, line 66: I would add a referrences to docstrings of all the task related components which explain all their options, e.g. the `Task.Status` takes the `renderStatus` and `renderError` which you can use to override its default progress and error views.

* [`Task` docs](https://github.com/uktrade/data-hub-frontend/blob/master/src/client/components/Task/index.jsx#L47-L100)
* [`Task.Status` docs](https://github.com/uktrade/data-hub-frontend/blob/master/src/client/components/Task/index.jsx#L164-L189)
* [`Task.StartOnRender` docs](https://github.com/uktrade/data-hub-frontend/blob/master/src/client/components/Task/index.jsx#L129-L140)

docs/redux-and-saga.md
```diff
@@ -0,0 +1,145 @@
...
+            }
+        default:
+            return state
+    }
+}
+```
+
+### 7. src/client/index.jsx
+
+This is the last step. We just need to hook up our reducer to the Redux store.
```
peterhudec, line 132: It would be more precise to say that "we need to compose our reducer into the root reducer".

docs/redux-and-saga.md
debitan: Ok. Will take it out. Just a pattern I've seen emerging from existing examples.

docs/redux-and-saga.md
debitan: I'll reference the doc strings. Was intending to keep this high level/an overview - but if people want to dive deeper they can in your doc strings.

docs/redux-and-saga.md
debitan: Will change.

docs/redux-and-saga.md
```diff
@@ -0,0 +1,145 @@
...
+
+<br />
+
+![redux-saga-flow](./redux-saga-flow.svg)
+
+<br />
+
+### 1. task.js
+
+This is where you will put your task. This is a standard JavaScript function that will make an axios call through the api-proxy (so be sure to add your api url to the api-proxy allowlist)
```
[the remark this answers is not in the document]
debitan: Will add.

docs/redux-and-saga.md
debitan: On second thoughts I will leave it in but make the point that it isn't necessary

docs/Redux and Saga.md
```diff
@@ -0,0 +1,147 @@
...
+
+![redux-saga-flow](./redux-saga-flow.svg)
+
+<br />
+
+### 1. task.js
+
+This is where you will put your task. The task function takes exactly one parameter - the payload - and returns a promise.
+
+```
```
peterhudec, line 21: Let Markdown know that it's TypeScript for better syntax highlighting by adding `ts` after the opening backticks: ```` ```ts ````

docs/Redux and Saga.md
```diff
@@ -0,0 +1,147 @@
+# Redux and Saga
+
+In Data Hub we use [Redux](https://redux.js.org/) for state management and [Saga](https://redux-saga.js.org/) for handling side effects (e.g. fetching data from the API).
+
+Wrapping our saga tasks we have a custom `Task` component and some other custom implementations with Redux. This can mean that just reading the Redux and Saga docs might not be enough to smoothly get started with writing new Data Hub code. So this high level guide goes over the flow of how to add a new Saga task and its accompanying Redux actions/reducers. From personal experience, it can get quite messy navigating all the different files you need to touch in order to simply load some data from the API and render it to a React component.
```
peterhudec, line 5: I would wrap the `` `Task` `` in a link to its source: ``[`Task`](https://github.com/uktrade/data-hub-frontend/blob/master/src/client/components/Task/index.jsx#L47-L101)``.

docs/Redux and Saga.md
```diff
@@ -0,0 +1,147 @@
...
+This is where you register your task with saga.
+
+```
+import { TASK_GET_PIPELINE_LIST } from './components/Pipeline/state'
+import * as pipelineListTasks from './components/Pipeline/tasks'
+
+export default function rootSaga() {
+    yield fork(
+        tasksSaga({
+            [TASK_GET_PIPELINE_LIST]: pipelineTasks.getPIpelineList,
```
peterhudec, line 55: Typo in `getP(I)pelineList`

docs/Redux and Saga.md
```diff
@@ -0,0 +1,147 @@
...
+```
+
+### 4. src/client/actions.js
+
+We also need to namespace our Redux action that will trigger once our Saga task has successfully completed. We do this so we don't give any two different actions the same name and accidentally trigger both actions.
+
+It should follow the pattern of [COMPONENT_NAME]__[ACTION_VERB].
+
+```
+export const PIPELINE**LIST_LOADED = "PIPELINE__LIST_LOADED"
```
peterhudec, line 66: It probably should be `export const PIPELINE__LIST_LOADED = "PIPELINE__LIST_LOADED"`

peterhudec (changes requested):
Almost there, but there are still some typos.

mcbhenwood:
While I think it can be enormously helpful having documentation to assist developers, publishing docs of this detail really does place a heavy burden on future developers to repeatedly parse these docs and make sure their own changes do not render the docs inaccurate or obsolete. Perhaps you need a comment in the actual JS code modules with 'links' to the location of these docs saying "if you materially alter this then please update the docs".

Of course, one could argue that the _need_ for such docs suggests an implementation which is in some way inscrutable or too "tightly wound" to be understood easily by simply reading. While conciseness and purity are excellent values to strive for, I feel that in the future when we are adding what is essentially our own framework tooling to a repo, it should be done as a joint effort including within that effort team members of lesser experience, so that we can build out solutions which can be understood simply by all, without the need for paragraphs of explanation.

peterhudec (approved):

ian-leggett (approved):

PR #2892, opened 2020-08-10
Add Elastic APM

ian-leggett:
## Description of change
This adds Elastic APM (Application Performance Monitoring) to the node layer so we can monitor transactions from the FE - https://www.elastic.co/guide/en/apm/agent/nodejs/3.x/configuring-the-agent.html

## Test instructions
This shouldn't effect anything on the FE, the only way you can see this take effect is by logging into our Kibana account and you will need access.

## Checklist

### DIFF
--- a/src/server.js
+++ b/src/server.js
@@ -1,5 +1,7 @@
 require('dotenv').config()
 
+require('elastic-apm-node').start()
+
 const path = require('path')
 const bodyParser = require('body-parser')
 const compression = require('compression')
--- a/package-lock.json
+++ b/package-lock.json
@@ -5375,6 +5375,11 @@
       "integrity": "sha1-/ts5T58OAqqXaOcCvaI7UF+ufh8=",
       "dev": true
     },
+    "after-all-results": {
+      "version": "2.0.0",
+      "resolved": "https://registry.npmjs.org/after-all-results/-/after-all-results-2.0.0.tgz",
+      "integrity": "sha1-asL8ICtQD4jaj09VMM+hAPTGotA="
+    },
     "agent-base": {
       "version": "5.1.1",
       "resolved": "https://registry.npmjs.org/agent-base/-/agent-base-5.1.1.tgz",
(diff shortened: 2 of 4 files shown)


src/server.js
```diff
@@ -1,5 +1,10 @@
 require('dotenv').config()
 
+require('elastic-apm-node').start({
+  serviceName: process.env.ES_APM_SERVICE_NAME,
+  secretToken: process.env.ES_APM_SECRET_TOKEN,
+  serverUrl: process.env.ES_APM_SERVER_URL,
```
ambrozic, line 6: do we need this variables added in vault?

paulgain, line 6: Shouldn't these also be defined in https://github.com/uktrade/data-hub-frontend/blob/master/src/config/envSchema.js?

mcbhenwood, line 6: According to [the docs](https://www.elastic.co/guide/en/apm/agent/nodejs/current/express.html#express-advanced-configuration), if you set the env vars `ELASTIC_APM_SERVICE_NAME`, `ELASTIC_APM_SECRET_TOKEN` and `ELASTIC_APM_SERVER_URL` then you can just start APM without any args like this:
```javascript
require('elastic-apm-node').start()
```

src/server.js
```diff
@@ -1,5 +1,10 @@
 require('dotenv').config()
 
+require('elastic-apm-node').start({
+  serviceName: process.env.ES_APM_SERVICE_NAME,
```
mcbhenwood, line 4: When configuring the env vars can we make sure the service name is 'datahub-fe' please. Thanks.

src/server.js
```diff
@@ -1,5 +1,10 @@
 require('dotenv').config()
 
+require('elastic-apm-node').start({
```
mcbhenwood, line 3: I note that in the [DIT docs](https://readme.trade.gov.uk/docs/howtos/app-metrics.html), Mateusz is asking for us to set the SERVER_TIMEOUT to 20 seconds as standard also

ian-leggett: I have now set this to 20 secs as per documentation

src/server.js
ian-leggett: All done

src/server.js
ian-leggett: All done

ian-leggett: Good spot I forgot about this.

ian-leggett: I have added new ones into the Vault (datahub-fe) as I was referencing the BE ones before. If you check Vault you should see these in all environments.

PR #2894, opened 2020-08-12
Add parallel capability to DIT e2e job

PippoRaimondi:
## Description of change

The intent of this PR is to run the e2e dit test in parallel in circleci to decrease the overall time of our build.

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -362,8 +362,8 @@ jobs:
       - *docker_mock_sso
       - *docker_data_hub_backend_api
       - *docker_data_hub_backend_celery
-      parallelism: 3
-      steps:
+    parallelism: 3
+    steps:
       - checkout
       - run: npm ci
       - run: npm run build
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -338,24 +338,25 @@ jobs:
       - *docker_mock_sso
       - *docker_data_hub_backend_api
       - *docker_data_hub_backend_celery
+    parallelism: 5
```
debitan, line 341: I played with this a little before and couldn't get any better overall performance because are already using up our max allocation of runners. So adding 5 here won't help too much I don't think and will get us in trouble with the other teams as we would block them.

But interested to see if you get a drop in overall test suite execution time. The DIT e2e run sometimes isn't the last one to finish, sometimes it is the coverage because that has to wait for the functional and unit test suites to finish and the functional takes about 5 mins anyway.

PippoRaimondi: good point... I might reduce this to 3 parallel runs actually... 

when running the tests against our docker setup the dit e2e job takes avg 15minutes (not entirely sure why), so I am predicting this will be needed once that is merged...

PippoRaimondi:
> Happy with the PR but one question, will this not be an issue as we don't setup and teardown properly?

@ian-leggett our e2e tests are all independent when it comes to data so they shouldn't have any impact between each other

.circleci/config.yml
PippoRaimondi: @debitan prior to the change the e2e suite was taking longer than the functionals, with the updates the suite is now running around 5minutes... I believe this is more balanced than before and gives us some allowance to add more e2e tests without slowing or build down...

PR #2898, opened 2020-08-12
feature: reduxify investment profile collection page

debitan:
## Description of change

This PR replaces the use of internal state with the use of Redux and Saga.

This is in preparation for this feature becoming the proof of concept for a unified redux/saga based collection list with filters.

## Test instructions

No visual changes. Everything should be as before with the exception that the url isn't updated with the page number. This isn't 'good' behaviour to force update this and we should look at this kind of things again when implementing react router on these pages. 

## Checklist


src/apps/investments/client/tasks.js
```diff
@@ -0,0 +1,26 @@
...
+  const offset = limit * (page - 1)
+  return axios
+    .get('/api-proxy/v4/large-investor-profile', {
+      params: {
+        limit,
+        offset,
+        sortby: 'modified_on',
+      },
+    })
+    .then(handleSuccess, handleError)
```
adamledwards, line 25: I prefer this pattern `.then(handleSuccess, handleError)` to `.then(handleSuccess).catch(handleError) `  👍

debitan: you can thank Graham for that :-)

I do think we should centralise the `handleSuccess` and `handleError` functions as we just repeated them in every tasks file

src/apps/investments/client/LargeCapitalProfileCollection.jsx
```diff
@@ -1,67 +1,50 @@
...
+      )}
+    </Task.Status>
   )
 }
 
-export default LargeCapitalProfileCollection
+export default connect(state2props, (dispatch) => ({
+  onPageClick: (page, event) => {
+    event.target.blur(),
+      event.preventDefault(),
```
adamledwards, line 44: Could you put the element logic in the component and keep this onPageClick to only dispatch.
The action shouldn't be aware of what is triggering the function.

src/apps/investments/client/reducer.js
```diff
@@ -0,0 +1,29 @@
...
+}
+
+export default (state = initialState, { type, result, page }) => {
+  switch (type) {
+    case INVESTMENTS__PROFILES_LOADED:
+      return {
+        ...state,
+        count: result.count,
+        results: result?.results?.map(transformLargeCapitalProfiles),
+      }
```
adamledwards, line 20: The initial value is an empty array but results could potentially be undefined. Can the CollectionList handle undefined values?

adamledwards:
Looks good just a question and a change request

src/apps/investments/client/reducer.js
debitan: It seems to work fine with no results - there is a functional test for an empty page. but I will double check exactly what the state looks like after returning no results.

debitan: Yup, it is fine. When there are no results the state remains as an empty array.

src/apps/investments/client/state.js
```diff
@@ -0,0 +1,5 @@
+export const TASK_GET_PROFILES_LIST = 'TASK_GET_PROFILES_LIST'
```
ian-leggett, line 1: By naming convention the "TASK" should have a double underscore so this will be `TASK__GET_PROFILES_LIST`. Personally I feel these should be in actions and not in this state.js file but for some reason we have been adding them here, I understand that the state ID should live here but not an action.

debitan: @peterhudec raised the question around this growing convention to put it in the state (I'm just copying that convention here). So would the best practice be to put it in actions,js? Is it an action?

debitan: also I'm not sure how strong the convention for the double underscore is with tasks. It is there for redux actions.

debitan: In fact, looking at the root_saga I don't think that is the convention that has developed:

```      [TASK_OPEN_REFERRALS_CONTACT_FORM]: referralsSendTasks.openContactForm,
      [TASK_SAVE_REFERRAL]: referralsSendTasks.saveReferral,
      [TASK_SAVE_ONE_LIST_DETAILS]: editOneListTasks.saveOneListDetails,
      [EXPORT_COUNTRIES_EDIT_NAME]:
        exportCountriesEditTasks.saveExportCountries,
      [TASK_GET_PIPELINE_BY_COMPANY]: pipelineTasks.getPipelineByCompany,
      [TASK_ADD_COMPANY_TO_PIPELINE]: pipelineTasks.addCompanyToPipeline,
      [TASK_GET_PIPELINE_LIST]: pipelineListTasks.getPipelineList,
      [TASK_GET_PIPELINE_ITEM]: pipelineTasks.getPipelineItem,
      [TASK_EDIT_PIPELINE_ITEM]: pipelineTasks.editPipelineItem,
      [TASK_ARCHIVE_PIPELINE_ITEM]: pipelineTasks.archivePipelineItem,
      [TASK_UNARCHIVE_PIPELINE_ITEM]: pipelineTasks.unarchivePipelineItem,
      [TASK_DELETE_PIPELINE_ITEM]: pipelineTasks.deletePipelineItem,
      [TASK_GET_PIPELINE_COMPANY_CONTACTS]: pipelineTasks.getCompanyContacts,
      [TASK_POSTCODE_TO_REGION]: addCompanyPostcodeToRegionTask,
      [TASK_GET_ACTIVE_EVENTS]: addInteractionFormTasks.fetchActiveEvents,
      [TASK_SAVE_INTERACTION]: addInteractionFormTasks.saveInteraction,
      [TASK_OPEN_CONTACT_FORM]: addInteractionFormTasks.openContactForm,
      [TASK_UPDATE_STAGE]: investmentAdminTasks.updateProjectStage,
      [TASK_UPDATE_ADVISER]: manageAdviser.updateAdviser,
      [DNB__CHECK_PENDING_REQUEST]: dnbCheck.checkIfPendingRequest,
      [TASK_GET_PROFILES_LIST]: investmentProfilesTasks.getLargeCapitalProfiles,
```

PR #2899, opened 2020-08-13
Use only a single Constate context

paulgain:
## Description of change

At present we have multiple contexts in use when there should only be one.

There are interdependencies among the following 3 files.

1. `useFormContext.js`
2. `useForm.js` 
3. `useField.js` 

Due to the migration of the data-hub-component library these files now live in two locations:

*data-hub-components*
https://github.com/uktrade/data-hub-components/tree/master/src/forms/hooks

*data-hub-frontend*
https://github.com/uktrade/data-hub-frontend/tree/master/src/client/components/Form/hooks

Components from the component library have a dependency on `useFormContext` and `useField`, when a component is moved to the FE it is important it continues to use the same context from the library. Once all components have been migrated we can then switch the context to the FE.

To be clear, we should leave the three files defined above (data-hub-frontend) in situ, however, they should only be used when the migration is complete, only then do we switch everything over.

Currently we are incorrectly mixing the context from the component library with the context from the FE causing unwanted side effects.

### DIFF
--- a/src/client/components/Form/elements/FieldCheckboxes/index.jsx
+++ b/src/client/components/Form/elements/FieldCheckboxes/index.jsx
@@ -9,7 +9,7 @@ import { BODY_SIZES } from '@govuk-react/constants'
 
 import useField from '../../hooks/useField'
 import FieldWrapper from '../FieldWrapper'
-import useFormContext from '../../hooks/useFormContext'
+import { useFormContext } from 'data-hub-components'
 
 const StyledDiv = styled('div')`
   ${(props) =>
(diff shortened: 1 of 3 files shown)


paulgain:
Closing out this PR now all React form components have been migrated from [data-hub-components](https://github.com/uktrade/data-hub-components) to [here](https://github.com/uktrade/data-hub-frontend/tree/master) and these [files](https://github.com/uktrade/data-hub-frontend/blob/master/src/client/components/Form/hooks/index.jsx) are the only set being used.

PR #2900, opened 2020-08-13
Mark tasks as finished

debitan:
## Description of change

Currently the child components of `Task.Status` will flash on the screen with no data before the task is started. Then when it is finished the child components re-render with data. This isn't ideal behaviour and so other developers have introduced checks against the state to see whether the task has finished or not. However, this doesn't work when you would still like to render a child component even if there are no results after the task has finished (unless you add another piece of state like `isFinished` that is updated in the reducer), and this has to be implemented each time a new component is built. This is easily missed and there are areas of the site that do flash before the Task start - this isn't great (but hands up, most of these are components I wrote...). 

Currently there is a `status` for the task. After it is started the `status` is set to `progress`. If it errors the `status` is set to `error`. However, when it finishes the task, and its `status`, is simply removed from the state entirely. 

This PR, rather than removing the task from state upon completion, leaves the task in state but marks its `status` as `finished`. This allows the `Task.Status` to check whether the status is `finished` before rendering the components. Currently it just checks to see if `status` is undefined. I think the problem we're seeing is that `status` is undefined momentarily before the Task begins. So checking to instead see whether it is `finished` should be more robust. 

This could lead to the state becomes clogged with all of the Tasks that a user has dispatched. However, currently in our express/React hybrid app, the state refreshes on every page change (barring the React routed dashboard sub tabs). So the state is wiped clean anyway as the user navigates around the site. Also, it probably isn't a big problem that they hang around in state anyway, there will never be enough to cause any performance problems. 

In summary, this change means that child components of `Task.Status` don't flash before the `Task` has finished. And you don't have to put a solution together to prevent this each time. 

## Test instructions

Navigate to referrals, make sure there are some in the database, and you will see there is no flash of the component before loading - even though I have removed the condition check that was there previously. You can also see that export history no longer flashes before loading. 

You may want to throttle your network speed to slow 3G to more easily see this in action. 

## Checklist

### DIFF
--- a/src/client/components/Task/reducer.js
+++ b/src/client/components/Task/reducer.js
@@ -1,4 +1,3 @@
-import { omit } from 'lodash'
 import { TASK__PROGRESS, TASK__ERROR, TASK__CLEAR } from '../../actions'
 
 const setTaskState = (state, { name, id, ...action }, status) => {
@@ -24,13 +23,7 @@ export default (state = {}, { type, ...action }) => {
     case TASK__ERROR:
       return setTaskState(state, action, 'error')
     case TASK__CLEAR:
-      return omit(
-        state,
-        Object.entries(state[action.name]).length > 1
-          ? `${action.name}.${action.id}`
-          : // Remove the whole task group if removing its last task
-            action.name
-      )
+      return setTaskState(state, action, 'finished')
     default:
       return state
   }
--- a/src/apps/companies/apps/referrals/details/client/ReferralDetails.jsx
+++ b/src/apps/companies/apps/referrals/details/client/ReferralDetails.jsx
@@ -63,89 +63,84 @@ export default connect(({ referrerUrl, ...state }) => ({
           onSuccessDispatch: REFERRAL_DETAILS,
         }}
       >
-        {() =>
-          company && (
-            <>
-              <SummaryTable caption={subject}>
-                <SummaryTable.Row heading="Company">
-                  <Link href={urls.companies.detail(company.id)}>
-                    {company.name}
+        {() => (
+          <>
+            <SummaryTable caption={subject}>
+              <SummaryTable.Row heading="Company">
+                <Link href={urls.companies.detail(company.id)}>
+                  {company.name}
+                </Link>
+              </SummaryTable.Row>
+              {contact && (
+                <SummaryTable.Row heading="Contact">
+                  <Link href={urls.contacts.contact(contact.id)}>
+                    {contact.name}
                   </Link>
                 </SummaryTable.Row>
-                {contact && (
-                  <SummaryTable.Row heading="Contact">
-                    <Link href={urls.contacts.contact(contact.id)}>
-                      {contact.name}
-                    </Link>
-                  </SummaryTable.Row>
-                )}
-                <SummaryTable.Row heading="Sending adviser">
-                  {sendingAdviser && <AdviserDetails {...sendingAdviser} />}
-                </SummaryTable.Row>
-                <SummaryTable.Row heading="Receiving adviser">
-                  {receivingAdviser && <AdviserDetails {...receivingAdviser} />}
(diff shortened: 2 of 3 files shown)


src/client/components/Task/reducer.js
```diff
@@ -24,13 +23,7 @@ export default (state = {}, { type, ...action }) => {
...
       return setTaskState(state, action, 'error')
     case TASK__CLEAR:
-      return omit(
-        state,
-        Object.entries(state[action.name]).length > 1
-          ? `${action.name}.${action.id}`
-          : // Remove the whole task group if removing its last task
-            action.name
-      )
+      return setTaskState(state, action, 'finished')
```
ian-leggett, line 26: This is nice, what about changing this to `setTaskState(state, action, true)` and then just check against status?

debitan: Because this can also have `error` there are 3 possibilities for it, so I'm not sure a boolean would work.

debitan: We could break them all out into booleans - true or false for progress, error and finished. Don't know if that is better?

debitan:
This actually looks to be causing a problem when firing a second request with a different payload, e.g. navigating to page 2 of the export history. I'll investigate further.

debitan:
> This actually looks to be causing a problem when firing a second request with a different payload, e.g. navigating to page 2 of the export history. I'll investigate further.

I can't figure this one out. Seems that the TASK__START won't trigger again if the task remains in state. @peterhudec any ideas? I presume this is the reason you were removing the task from state on TASK__CLEAR...

peterhudec:
> This actually looks to be causing a problem when firing a second request with a different payload, e.g. navigating to page 2 of the export history. I'll investigate further.

@debitan So the current behaviour was chosen precisely for this kind of reasons ☝️ There was actually a phase in the development of it where the success state was kept in the state, but I removed it in the end because it was behaving weirdly sometimes. I have implemented this concept previously in a different project where the task kept the success state around, including the resolved value so you didn't have to use the `onSuccessDispatch` and there was no need for a reducer to handle the result, because the `Task` would inject the result into its children:
```js
<Task.Status name="Load something" id="foo">
  {result => JSON.stringify(result)}
</Task.Status>
```
It was working pretty well, but I wanted to keep it simpler and lower level here, and also thought it would be better if the loaded data was managed by the component where the task is used. The `Task` would only manage the state of the phases of getting the data asynchrounously.

I had an idea about a `Resource` component which would work similarly to what I described above, and it even could actually be implemented in terms of `Task`. It could have a simple caching mechanism, and would map nicely to our API endpoints. But then I was moved to the Enquiry Management tool and didn't have time for this at all.

src/client/components/Task/reducer.js
peterhudec: You should also rename the `TASK__CLEAR` action. Also I would consider using `'success'` as the status name, because `'finished'` is a bit inconsistent with `'error'` the first being a past simple verb and the latter a noun.

peterhudec:
I would advise not to change the behaviour of `Task` but build a higher level abstraction on top of it instead, let's say a `Resource` which would also manage the result in its state.

debitan:
> I would advise not to change the behaviour of `Task` but build a higher level abstraction on top of it instead, let's say a `Resource` which would also manage the result in its state.

This is actually a solution I discussed with @ian-leggett - wrapping the Task in another function which could set a success state. Out of interest, why would you call it `resource`?

peterhudec:
> > I would advise not to change the behaviour of `Task` but build a higher level abstraction on top of it instead, let's say a `Resource` which would also manage the result in its state.
> 
> This is actually a solution I discussed with @ian-leggett - wrapping the Task in another function which could set a success state. Out of interest, why would you call it `resource`?

@debitan Yes I would call it `Resource`, but it depends on what exactly it would do.

PR #2903, opened 2020-08-14
Cleanup profiles task name

debitan:
This is a small cleanup PR that:

- changes `INVESTMENTS__PROFILE_SELECT_PAGE` to be `INVESTMENTS_PROFILES_SELECT_PAGE` so it is more consistently named with other actions.


ian-leggett (approved):

cgsunkel (approved):

src/client/components/CollectionList/CollectionSort.jsx
```diff
@@ -21,7 +21,7 @@ function CollectionSort({
       name="sortBy"
       label="Sort by"
       input={sortInput}
-      onChange={sortOnChange}
+      onChange={(event) => sortOnChange(event.target.value)}
```
paulgain, line 24: I would leave this how it was previously. It's best to pass the event, all consumers can then pick what they want from the `event.target` object - now it's limited to just `value.`

debitan: I think `value` is all that is needed for this. But I'll drop the commit with this change.

PR #2904, opened 2020-08-14
Update React form and hook references

ian-leggett:
## Description of change
After moving the form elements over from Data hub Components we need to update the references to these React elements and hooks so they no longer point to and depend on Data hub Components. 

As well as updating the component references I have updated the naming of the forms as it was getting confusing what does what. Now **BOTH** forms are in FE so we have the following
 
- `MultiInstanceForm` uses connect from react-redux
- `Form` only uses the context api. 

**Its clear that we need to merge these into one but this is out of scope for this PR.**

### Outstanding dependencies from Data Hub Components are:
1. Activity feed 

## Test instructions

Nothing should change as its all around renaming and changing references.

## Checklist


src/client/components/Form/elements/__stories__/FieldTypeahead.stories.jsx
```diff
@@ -4,7 +4,7 @@ import { action } from '@storybook/addon-actions'
 import { withKnobs } from '@storybook/addon-knobs'
 import Button from '@govuk-react/button'
 
-import { FormStateful } from 'data-hub-components'
+import { FormStateful } from '../../../'
```
paulgain, line 7: `import FormStateful from '../../FormStateful'`

I think this reads better and is consistent with other imports:

src/client/components/index.jsx
```diff
@@ -18,12 +18,17 @@ export { default as StatusMessage } from './StatusMessage'
...
 export { default as SummaryTable } from './SummaryTable'
 export { default as Tag } from './Tag'
+export { default as AdviserTypeAhead } from './AdviserTypeAhead'
 export { default as Typeahead } from './Typeahead/Typeahead'
 export { default as SmallTypeahead } from './Typeahead/SmallTypeahead'
 export { default as DateUtils } from './Utils/DateUtils'
 export { default as NumberUtils } from './Utils/NumberUtils'
 export { default as DropdownMenu } from './DropdownMenu'
-export { default as Form } from './Form/'
+export { default as Form } from './Form/Form'
```
paulgain, line 27: Do we really need to export `Form`? 

From my understanding `FormStateful` wraps `Form` and they both live in the same directory, `FormStateful` imports `Form` relatively: `import Form from './Form'`. I don't think `Form` is used in any other context so we should remove this export. In addition, we don't want anyone using `Form` directly as we should be using `MultiInstanceForm` instead.

ian-leggett:
> @ian-leggett looking good, just a few comments.
> 
> There are 4 occurrences of `NumberUtils` still being imported from the component library:
> `const { NumberUtils } = require('data-hub-components')`
> 
> 1. _src/apps/companies/apps/edit-company/transformers.js_
> 2. _test/functional/cypress/specs/companies/edit-spec.js_
> 3. _test/functional/cypress/specs/pipeline/my-pipeline-spec.js_
> 4. _test/functional/cypress/support/pipeline-assertions.js_ (`DateUtils` is also imported here)

These are utility classes which are not React components. We need to move these over separately as they are also exported using ES6 modules and imported using common JS which isn't quite right. I can do a PR afterwards to clean these up.

paulgain:
> > @ian-leggett looking good, just a few comments.
> > There are 4 occurrences of `NumberUtils` still being imported from the component library:
> > `const { NumberUtils } = require('data-hub-components')`
> > 
> > 1. _src/apps/companies/apps/edit-company/transformers.js_
> > 2. _test/functional/cypress/specs/companies/edit-spec.js_
> > 3. _test/functional/cypress/specs/pipeline/my-pipeline-spec.js_
> > 4. _test/functional/cypress/support/pipeline-assertions.js_ (`DateUtils` is also imported here)
> 
> These are utility classes which are not React components. We need to move these over separately as they are also exported using ES6 modules and imported using common JS which isn't quite right. I can do a PR afterwards to clean these up.

A follow up PR makes sense

PR #2921, opened 2020-08-28
Remove all data-hub-components references bar the Activity Feed

paulgain:
## Background

1. Delete `DateUtils` from `/src/client/components` as one already exists in `/src/client/utils`. 
2. Move `NumberUtils` from `/src/client/components` to `/src/client/utils`
3. Refactor `DateUtils` to `date-utils` and move all functions out of the class and import them like this:
    - ` import { format } from '/the/path/client/utils/date-utils'`
4. Refactor `NumberUtils` to `number-utils` and move all functions out of the class and import them like so:
    - `import { currencyGBP } from '/the/path/client/utils/number-utils'`
5. Remove the `data-hub-components` import for both `date-utils` and `number-utils` and use the local import instead
6. Refactor the files in `src/client/utils` to ensure they are all exported as ES6 modules, if they can't be refactored (i.e. they have to be CommonJS exports) then move them out to `src/common`

## Issues with date-utils.js
For `date-utils.js` to work we have to export all functions as an ES6 module as all consumers are React `.jsx` files.

If we use `module.exports` (i.e. CommonJS) the imports in the `.jsx` file cannot find the exported function.

**date.utils.js**
```js
const format = (foo) => {
  return foo
}

// Exported as a CommonJS module
module.exports = {
  format,
}
```

**InteractionReferralDetails.jsx**
```javascript
import { format } from '../../../../../../client/utils/date-utils'
```

Bringing up the app we get 21 Webpack warnings (one for each import), here's one of them:

```
WARNING in ./src/apps/companies/apps/referrals/details/client/InteractionReferralDetails.jsx 128:5-11
"export 'format' was not found in '../../../../../../client/utils/date-utils'
 @ ./src/client/index.jsx
 @ multi react-app-polyfill/ie9 react-app-polyfill/stable details-element-polyfill ./src/client/index.jsx
```
The only way this can work is by ensuring all functions are exported as ES6 modules

```js
export const format = (foo) => {
  return foo
}
```

All functions within [date-utils.js](https://github.com/uktrade/data-hub-frontend/blob/refactor/remove-data-hub-component-dependencies/src/client/utils/date-utils.js) have to be exported via ES6.

## Issues with number-utils.js
If we try to export `number-utils.js` as an ES6 module for consistency

```js
export const roundToSignificantDigits = (foo) => {
  return foo
}
```

On the server we have a transformer that requires it:

**transformers.js**
src/apps/companies/apps/edit-company/transformers.js

```js
const {
  roundToSignificantDigits,
} = require('../../../../client/utils/number-utils')
```

We get this error:

```
export const decimal = (number) => {
^^^^^^

SyntaxError: Unexpected token 'export'
    at wrapSafe (internal/modules/cjs/loader.js:1072:16)
    at Module._compile (internal/modules/cjs/loader.js:1122:27)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1178:10)
    at Module.load (internal/modules/cjs/loader.js:1002:32)
    at Function.Module._load (internal/modules/cjs/loader.js:901:14)
    at Module.require (internal/modules/cjs/loader.js:1044:19)
```

**This is because the CommonJS loader doesn't know how to load an ES6 module**

All functions within [number-utils.js](https://github.com/uktrade/data-hub-frontend/blob/refactor/remove-data-hub-component-dependencies/src/client/utils/number-utils.js) have to be exported via CommonJS.

## Conclusion
The issue has exposed itself by simply moving two utility files that incorrectly lived in `src/client/components` to `src/client/utils`. The problem is that we now have two files living in the same directory but each export their functions differently based on their consumers, one via CommonJS, the other via ES6. The functions were written for reusability but the type of consumer will now be limited.

I'm not sure what the solution is considering we have a very complex hybrid app with many technologies under the hood. We are also transitioning from one stack to another. I'm inclined to put things back as they were previously but felt it necessary to point this out to **elicit feedback** or maybe this is acceptable considering our circumstances.


ian-leggett (commented):
Maybe we duplicate the utility functions so we have client side versions and server side versions. It doesn't seem right that in the folder "client" we are exporting functions using commonJS. We should separate these concerns, what do you think? We are not duplicating code as they are being used in different contexts.

paulgain:
> Maybe we duplicate the utility functions so we have client side versions and server side versions. It doesn't seem right that in the folder "client" we are exporting functions using commonJS. We should separate these concerns, what do you think? We are not duplicating code as they are being used in different contexts.

I've just pushed some changes that addresses the issues. All files that live in `/src/client/utils` are now exported using ES6 syntax only. The files with CommonJS exports have been moved to `src/common` so we don't have any duplication (we should honour the DRY principle ;-))

ian-leggett:
> > Maybe we duplicate the utility functions so we have client side versions and server side versions. It doesn't seem right that in the folder "client" we are exporting functions using commonJS. We should separate these concerns, what do you think? We are not duplicating code as they are being used in different contexts.
> 
> @ian-leggett I've just pushed some changes that address the issues. All files that live in `/src/client/utils` are now exported using ES6 syntax only. The files with CommonJS exports have been moved to `src/common` so we don't have any duplication (we should honour the DRY principle ;-))

This looks much nicer 👍

ian-leggett (approved):

cgsunkel (approved):

PR #2939, opened 2020-09-09
Update date-fns

ian-leggett:
## Description of change
Before `v2` date-fns` used to apply more magic and accept dates as string such as "1/1/2020" or "1-1-2020" or timestamps and convert them to your desired format. After v2 they tightened up the rules on what you should pass to it for good reasons, ideally the only thing that should get passed to "date-fns" is a date object but throughout Data Hub this is not the case and to add to further confusion we have inconsistent date strings being passed to "date-fns".

This PR basically offers a patch over a bigger issue which is that we are passing various formatted human readable date strings at a level we shouldn't be. We should only offer "date-fns" a date object and that should come from the backend the formatting should only be done at the view layer. 

We have three options:

1. Accept this PR and address the bigger problem at a later date
2. Don't update `date-fns`
3. Ignore this and work on the bigger issue

## Checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1508,6 +1508,12 @@
             "restore-cursor": "^1.0.1"
           }
         },
+        "date-fns": {
+          "version": "1.30.1",
+          "resolved": "https://registry.npmjs.org/date-fns/-/date-fns-1.30.1.tgz",
+          "integrity": "sha512-hBSVCvSmWC+QypYObzwGOd9wqdDpOt+0wl0KbU+R+uuZBS1jN8VsD1ss3irQDknRj5NvxiTF6oj/nDRnN/UQNw==",
+          "dev": true
+        },
         "figures": {
           "version": "1.7.0",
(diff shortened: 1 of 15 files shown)


mcbhenwood:
Some thoughts on your options @ian-leggett:

"Accept this PR and address the bigger problem at a later date" We shouldn't avoid the bigger problem, we should fix it now. That's a general rule.

"Don't update date-fns" We should always patch packages as soon as we can (as you are here) - another general rule.

"Ignore this and work on the bigger issue" We should work on the Bigger Issue. BUT. I agree that human-readable strings ought not be passed around in the context of actual data but besides `Date` objects we can also legally pass around ISO-formatted dates at a programmatic level seeing as this is a valid format for these objects during transmission by JSON. They're timezone-aware and have sub-second granularity so they are OK as real data (in the way that '06 Jan 2020 12:15pm' is NOT)

If we accept ISO-formatted dates as data / variables does the problem diminish a little? I must confess I only glanced at the PR so I could be wrong.

PR #2942, opened 2020-09-10
Bugfix: omis quote edit

adamledwards:
## Description of change

We've updated our dependancy of [chrono-node](https://www.npmjs.com/package/chrono-node) that had breaking pages.
This PR add fixes for these changes and add simple test functional tests.

## Test instructions

Go to orders > press edit at the `Information of the quote` 
Edit the information and you should get a success message

### Before
image

### After

image

## Checklist

### DIFF
--- a/src/apps/omis/apps/edit/controllers/payment-reconciliation.js
+++ b/src/apps/omis/apps/edit/controllers/payment-reconciliation.js
@@ -28,7 +28,7 @@ class EditPaymentReconciliationController extends EditController {
 
   process(req, res, next) {
     const paymentDateStr = req.body.received_on
-    const parsedDate = chrono.en_GB.parseDate(paymentDateStr)
+    const parsedDate = chrono.en.GB.parseDate(paymentDateStr)
 
     if (parsedDate) {
       req.form.values.received_on = dateFns.format(parsedDate, 'YYYY-MM-DD')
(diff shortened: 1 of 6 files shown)


debitan:
I was just looking at the bug that prevents reconciling payments as well and this also seems to be related to this. So will double check it fixes that also.

src/apps/omis/apps/edit/controllers/payment-reconciliation.js
```diff
@@ -28,7 +28,7 @@ class EditPaymentReconciliationController extends EditController {
 
   process(req, res, next) {
     const paymentDateStr = req.body.received_on
-    const parsedDate = chrono.en_GB.parseDate(paymentDateStr)
+    const parsedDate = chrono.en.GB.parseDate(paymentDateStr)
```
debitan, line 31: Yup, this should solve my issue. Might be worth adding a cypress test around this also?

debitan: But this bug is pretty annoying so let's get it out like this and add a ticket to add the extra test afterwards. (Being pragmatic)

debitan (approved):

src/apps/omis/apps/edit/controllers/payment-reconciliation.js
debitan: I can't actually manually check that this solves the payment reconcilation bug because of the poor test data. I would need a `quote accepted` order to then reconcile. Can't force that myself because it relies on the customer accepting the quote and there is no way to do this from the frontend. So let's get this fix out and see if it fixes both problems. Not ideal, but.

PR #2972, opened 2020-09-22
Metadata API call failure discussion

paulgain:
### Background
During development some developers like myself prefer to develop with a native local frontend pointing to the dev API in PaaS.

### Recent problem
Currently when the app starts up it makes 33 metadata API calls, around two thirds of those calls are succeeding, the rest are failing. If you start and stop the app a number of times the ratio changes. There was one particular time when there was only a single failure, the response from the API is quite random. What is consistent is that the app doesn't come up when pointing to dev. There has been no frontend changes the would attribute to this change in app behaviour. I've gone back to the beginning of September (a point in time when everything was working) and checked out a random commit hash to see if we had introduced something during the previous three weeks, but to no avail, the problem still persists in that earlier version (at the time there was no such problem) - even last week everything was fine during application startup.

The calls that are failing are all timing out. During testing I temporarily swapped out our HTTP client for another called `superagent`, here  I was able to apply multiple retries for each call that failed, however, due to the retries the app took too long to start which is obviously unworkable.

If you make any of these failed calls in Postman they will succeed, that's because we are only making a single call, the problem exposes itself when making lots of API calls in quick succession where the results vary as mentioned.

Screenshot 2020-09-22 at 19 03 41

## Checklist

### DIFF
--- a/src/lib/hawk-request.js
+++ b/src/lib/hawk-request.js
@@ -1,6 +1,7 @@
 const request = require('request')
 const hawk = require('@hapi/hawk')
 const config = require('../config')
+const logger = require('../config/logger')
 
 function getHawkHeader(credentials, requestOptions) {
   if (config.isTest) {
@@ -18,6 +19,9 @@ function getHawkHeader(credentials, requestOptions) {
   })
 }
 
+let successCount = 0
+let failureCount = 0
+
 function createPromiseRequest(
   requestOptions,
   credentials,
@@ -25,7 +29,11 @@ function createPromiseRequest(
 ) {
   return new Promise((resolve, reject) => {
     request(requestOptions, (err, response, responseBody) => {
+      const uri = requestOptions.uri
+      const endpoint = uri.substring(uri.indexOf('/v4'))
+
       if (err) {
+        logger.error(`ERROR ${++failureCount} : ${endpoint} ${err}`)
         return reject(new Error(err))
       }
       if (!config.isTest) {
@@ -52,6 +60,7 @@ function createPromiseRequest(
 
       const statusCode = response.statusCode
       if (statusCode >= 200 && statusCode < 300) {
+        logger.debug(`SUCCESS ${++successCount} : ${endpoint}`)
         return resolve(body)
       }
       const error = new Error(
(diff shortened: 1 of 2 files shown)


src/lib/metadata.js
```diff
@@ -39,7 +39,7 @@ function getMetadata(path, key) {
       return responseData
     })
     .catch((err) => {
-      logger.error('Error fetching metadataRepository for url: %s', url)
+      // logger.error('Error fetching metadataRepository for url: %s', url)
```
ian-leggett, line 42: Do we need this comment?

src/lib/hawk-request.js
```diff
@@ -18,14 +19,21 @@ function getHawkHeader(credentials, requestOptions) {
   })
 }
 
+let successCount = 0
```
ian-leggett, line 22: I would pop these two `let`'s inside the function as the function below.

src/lib/hawk-request.js
```diff
@@ -18,14 +19,21 @@ function getHawkHeader(credentials, requestOptions) {
...
+let failureCount = 0
```
ian-leggett, line 23: I would pop these two `let`'s inside the function below.

paulgain:
@ian-leggett this is a discussion PR, it's not supposed to me merged - just like both tags say.

mforner13:
I just checked the dev API logs in cloud foundry and this error is appearing constantly:
`[APP/PROC/WEB/1] ERR 2020-09-23 10:23:38,704 [WARNING] [datahub.core.auth] PaaS IP check authentication is disabled.`. I will ask webops whether this could cause the errors we're getting.

paulgain:
Applying a delay to each API call yields the following results 

**50ms delay**
50ms

**500ms delay**
500ms

**1000ms delay**
1000ms

paulgain:
> I just checked the dev API logs in cloud foundry and this error is appearing constantly:
> `[APP/PROC/WEB/1] ERR 2020-09-23 10:23:38,704 [WARNING] [datahub.core.auth] PaaS IP check authentication is disabled.`. I will ask webops whether this could cause the errors we're getting.

👍

PR #2980, opened 2020-09-25
Toggle component

ian-leggett:
## Description of change
This toggle component just toggles visibility of any child elements you wrap it with, this is needed to wrap groups of filters in the new React collection list page that is currently under progress. The design for this can be seen here - https://trade.invisionapp.com/share/VYYV2GMAEW2#/screens/432558396_Investment_Default_Filters

Note: this is getting merged into the branch where we are iterating the React investments collection list feature not master.

## Test instructions
Run Storybook and select the `ToggleSection` component

## Screenshots
Screenshot 2020-10-29 at 17 17 55

Screenshot 2020-10-30 at 09 34 48

## Checklist


src/client/components/ToggleSection/index.jsx
```diff
@@ -0,0 +1,76 @@
...
+  font-size: ${FONT_SIZE.SIZE_19};
+  color: ${BLUE};
+  padding: 0 0 10px 0;
+  cursor: pointer;
+  &:hover {
+    text-decoration: underline;
+  }
+  &:focus {
+    outline: none;
+    text-decoration: underline;
```
rachaelcodes, line 32: What do you think about keeping this consistent with other focus styling (e.g. yellow outline/background, grey text: https://design-system.service.gov.uk/styles/colour/ )? I think it might be clearer, for example if people are tabbing through the page on a keyboard on high zoom, as well as being consistent.
(Maybe what we actually need is some sort of focus mixin, but that's maybe a problem for another day!)

src/client/components/ToggleSection/__stories__/Togglesection.stories.jsx
```diff
@@ -0,0 +1,85 @@
...
+        {() => (
+          <>
+            <FieldInput
+              label="Text"
+              hint="Some hint"
+              name="testField"
+              required="Enter text"
+              type="text"
+            />
+            <FieldInput
```
rachaelcodes, line 73: Just wondering - is there a reason to have two identical fields?

mforner13: Agreed - doesn't seem necessary.

src/client/components/ToggleSection/__stories__/Togglesection.stories.jsx
```diff
@@ -0,0 +1,85 @@
...
+      </ToggleWithState>
+      <ToggleWithState label="Toggle me two" isOpen={true}>
+        <p>
+          Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do
+          eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad
+          minim veniam, quis nostrud exercitation ullamco laboris nisi ut
+          aliquip ex ea commodo consequat.
+        </p>
+      </ToggleWithState>
+    </>
```
mforner13, line 59: I worry that having text as the example will be confusing as we are using the gov uk react `Details` component in the rest of DH for toggling text - which also comes with the grey line down the side - making it neater, e.g. on the companies export tab example below. 

Perhaps just having the form element example would prevent any confusion?

Screenshot 2020-10-14 at 14 30 57

mforner13 (changes requested):

src/client/components/ToggleSection/__stories__/Togglesection.stories.jsx
mforner13: Also adding something to the readme saying for toggling text use the gov uk `Details` component maybe?

ian-leggett: I agree this could be confusing and this shouldn't replace the use case for the details component, I am now only showing the form elements inside the toggle component. After speaking with Mark about the documentation in Storybook I have decided to remove it as I don't think its very useful at this point, its also very easy for this documentation to become stale and out of date as it doesn't get automagically updated when you update the component. Once we have all the new components in and we know exactly how they should be used maybe we can then add some documentation.

src/client/components/ToggleSection/index.jsx
ian-leggett: Good spot, I have updated this as per - https://www.gov.uk/search/all?keywords=search&order=relevance which should be using the correct focus styling

mforner13 (approved):

rachaelcodes (approved):

PR #2993, opened 2020-10-01
Add code scanning

cgsunkel:
## Description of change

Github have introduced a new feature that allows us to scan our code for security issues at the PR stage. It uses a Github action to check for vulnerabilities that runs when we create a new PR, and is a native alternative to our existing LGTM checks.

In order to demonstratre this more fully I have created a fork of the FE (in order to properly scan there needs to be a history of previous scans in `master`) - see [this PR](https://github.com/cgsunkel/data-hub-frontend/pull/2/files) in my fork for a better demonstration.

Vulnerabilities found are displayed in the Security tab:
Screenshot 2020-10-06 at 14 54 36

If we utilise this we will be able to use this to replace our existing LGTM checks.

## Test instructions

_What should I see?_

### Before

_Add a screenshot_

### After

_Add a screenshot_

## Checklist

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,60 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+name: "CodeQL"
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [master]
+  schedule:
+    - cron: '0 5 * * 2'
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
+        # Override automatic language detection by changing the below list
+        # Supported options are ['csharp', 'cpp', 'go', 'java', 'javascript', 'python']
+        language: ['javascript', 'python']
+        # Learn more...
+        # https://docs.github.com/en/github/finding-security-vulnerabilities-and-errors-in-your-code/configuring-code-scanning#overriding-automatic-language-detection
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v2
+      with:
+        # We must fetch at least the immediate parents so that if this is
+        # a pull request then we can checkout the head.
+        fetch-depth: 2
+
+    # If this run was triggered by a pull request event, then checkout
+    # the head of the pull request instead of the merge commit.
+    - run: git checkout HEAD^2
+      if: ${{ github.event_name == 'pull_request' }}
+
+    # Initializes the CodeQL tools for scanning.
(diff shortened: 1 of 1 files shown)


ambrozic (approved):

ambrozic:
this is great. can't wait to see it in action.

.github/workflows/codeql-analysis.yml
```diff
@@ -50,7 +50,7 @@ jobs:
         # By default, queries listed here will override any specified in a config file. 
         # Prefix the list here with "+" to use these queries and those in the config file.
         # queries: ./path/to/local/query, your-org/your-repo/queries@main
-        queries: +security-and-quality
+        queries: security-and-quality
```
mforner13, line 53: Do you know which config file the comment above this is referring to? If so, are we sure we want to override any queries in there?

cgsunkel: It's built-in to the action

mforner13:
Looks really useful! Do you foresee this being a check that must pass before a PR can be merged? 

Could you also link to some documentation which describes what will be checked for by this action?

.github/workflows/codeql-analysis.yml
mforner13: Ah okay and does it have any queries in which we are overriding?

cgsunkel:
I don't think this will need to be a required check, as there are some cases where an alert will be raised but we don't need to fix it (an example of this is the 'Disabling certificate validation' one, which is relating to something that is only activated in dev environments). I will see if I can find a list for everything that is checked for but I think it's mostly the same list that LGTM uses.

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,72 @@
...
+      uses: github/codeql-action/autobuild@v1
+
+    # ℹ️ Command-line programs to run using the OS shell.
+    # 📚 https://git.io/JvXDl
+
+    # ✏️ If the Autobuild fails above, remove it and uncomment the following three lines
+    #    and modify them (or add more) to build your code if your project
+    #    uses a compiled language
+
+    #- run: |
```
PippoRaimondi, line 67: this seems to be a command which we want to ignore? if so shall we delete this bit rather than having it commented out to keep it cleaner?

PippoRaimondi (approved):
looks cool, tiny comment...

.github/workflows/codeql-analysis.yml
cgsunkel: I don't think so as we don't have a separate config file, I think that queries would only be overwritten if we had a specific config file containing custom queries. The docs for this part are here - https://docs.github.com/en/free-pro-team@latest/github/finding-security-vulnerabilities-and-errors-in-your-code/configuring-code-scanning#running-additional-queries

.github/workflows/codeql-analysis.yml
cgsunkel: I will do that now, this file was generated automatically so I left most of it alone 🙂

mforner13 (approved):
Looks good 👍

PR #3030, opened 2020-10-23
Add telephone input validation

PippoRaimondi:
## Description of change

The intent of this PR is to enhance the validation around the telephone number when adding a company.

Validations added to the telephone field:
- non ascii characters
- no numeric values within the string provided for the telephone number
  - i.e `the company number is` will fail the validation, but `the company number is: 123123123` will pass the validation

In this PR we also enhanced the validation around website. Previously if a validation failed due to an invalid website input it would apply the same error message to the telephone number field.

## Checklist

### DIFF
--- a/src/apps/companies/apps/add-company/client/CompanyNotFoundStep.jsx
+++ b/src/apps/companies/apps/add-company/client/CompanyNotFoundStep.jsx
@@ -11,23 +11,40 @@ import {
   FieldAddress,
 } from '../../../../../client/components'
 
-import { WEBSITE_REGEX } from './constants'
+import {
+  WEBSITE_REGEX,
+  GENERIC_PHONE_NUMBER_REGEX,
+  NON_ASCII_REGEX,
+} from './constants'
 
-const requiredWebsiteAndOrPhoneValidator = (
+const websiteValidator = (
   value,
   name,
   { values: { website, telephone_number } }
 ) => {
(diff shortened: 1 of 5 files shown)


src/apps/companies/apps/add-company/client/CompanyNotFoundStep.jsx
```diff
@@ -11,23 +11,40 @@ import {
...
 
-import { WEBSITE_REGEX } from './constants'
+import {
+  WEBSITE_REGEX,
+  GENERIC_PHONE_NUMBER_REGEX,
+  NON_ASCII_REGEX,
+} from './constants'
 
-const requiredWebsiteAndOrPhoneValidator = (
+const websiteValidator = (
```
ian-leggett, line 20: Feels like this maybe needed in utils, I am also surprised we don't use a library for this. Just a thought.

ian-leggett (approved):

src/apps/companies/apps/add-company/client/CompanyNotFoundStep.jsx
PippoRaimondi: I had the same feeling but then realised this is probably the only page that the user can enter a telephone number or website, so maybe it's ok...

cgsunkel (approved):

PR #3042, opened 2020-10-30
SPIKE: replace request with axios

rachaelcodes:
## Description of change

## Test instructions

_What should I see?_

### Before

_Add a screenshot_

### After

_Add a screenshot_

## Checklist

### DIFF
--- a/src/apps/contacts/controllers/edit.js
+++ b/src/apps/contacts/controllers/edit.js
@@ -119,14 +119,14 @@ async function postDetails(req, res, next) {
         : urls.contacts.details(newContact.id)
     res.redirect(redirectUrl)
   } catch (errors) {
-    if (errors.error) {
-      if (errors.error.errors) {
+    if (errors.response) {
+      if (errors.response.data) {
         res.locals.errors = {
-          messages: errors.error.errors,
+          messages: errors.response.data,
         }
       } else {
         res.locals.errors = {
-          messages: errors.error,
+          messages: errors.response,
         }
       }
 
--- a/src/apps/events/middleware/details.js
+++ b/src/apps/events/middleware/details.js
@@ -20,10 +20,10 @@ async function postDetails(req, res, next) {
     }
     return res.redirect(`/events/${result.id}`)
   } catch (err) {
-    if (err.statusCode === 400) {
+    if (err.response.status === 400) {
       res.locals.form = assign({}, res.locals.form, {
         errors: {
-          messages: err.error,
+          messages: err.response.data,
         },
       })
       next()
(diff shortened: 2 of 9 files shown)


src/lib/authorised-request.js
```diff
@@ -82,7 +83,8 @@ function authorisedRequest(req, opts) {
   logRequest(requestOptions)
   requestOptions.jsonReviver = stripScripts
 
-  const r = requestPromise(requestOptions)
+  // const r = requestPromise(requestOptions)
+  const r = axios(requestOptions)
```
mcbhenwood, line 87: This may be needlessly picky _BUT_... in the terminology of the Axios library, the object passed is known as `config` rather than `/[a-z]+Options/`. If we'd adopted Axios first we'd probably have named this argument along those lines so maybe worth altering it now?

rachaelcodes: Sounds good - there's also possibly some unwanted noise in the object, so there's more work that needs doing to tidy it up into an axios shape.

(Although, to be honest, this is more of a 'what breaks if I brute-force this?' draft.)

src/apps/contacts/controllers/edit.js
```diff
@@ -119,14 +119,14 @@ async function postDetails(req, res, next) {
         : urls.contacts.details(newContact.id)
     res.redirect(redirectUrl)
   } catch (errors) {
-    if (errors.error) {
-      if (errors.error.errors) {
+    if (errors.response) {
```
rachaelcodes, line 122: Changes like this (that revolve around the differences between the axios & request error response interface) could be avoided if we decide to create an interface layer around our new axios requests. Whether or not that's a good idea going forward, or whether it's better to have the data structures updated to match the updated library is another matter.

src/apps/events/middleware/details.js
```diff
@@ -20,10 +20,10 @@ async function postDetails(req, res, next) {
     }
     return res.redirect(`/events/${result.id}`)
   } catch (err) {
-    if (err.statusCode === 400) {
+    if (err.response.status === 400) {
```
rachaelcodes, line 23: As above, this sort of change might be avoided if we instead create some kind of interface layer/transformation function around the new axios requests. It's up for debate whether we should.

I'll stop posting this now.

src/middleware/errors.js
```diff
@@ -31,17 +31,55 @@ function getStatusMessage(error) {
   return 'Page unavailable'
 }
 
+// TODO: remove this/integrate into getStatusMessage
```
rachaelcodes, line 34: The work I've done in this file is pretty exploratory: having changed one of the 'request' files to use 'axios' but not both, I wanted to quickly make sure that errors would be handled both for requests made by request and axios, while diagnosing issues caused by the change to axios. As raised earlier, we need to decide on a sensible strategy for handling the change of the data interface caused by the move to axios. 

TL:DR - these changes can probably be ignored and reverted.

PR #3043, opened 2020-10-30
Allow task cancelation and error abortion

peterhudec:
## Description of change

This PR allows _tasks_ to be cancelled in their progress and their errors to be aborted. This is a precursor for a `TaskTypeahead` component.

## Test instructions

The feature is not used anywhere yet, but can be seen in action in the storybook under Task -> Cancellation.


ian-leggett:
Unfortunately I couldn't run this :-( When I fire this up in Storybook I get console errors
Screenshot 2020-11-02 at 14 46 30
It looks like the demo task isn't registered in the root saga file

peterhudec:
> It looks like the demo task isn't registered in the root saga file. Is this all around allowing the user to retry a call to the backend or stop a current process in the UI?
Oh, sorry, @ian-leggett, seems I have forgot to commit the file where the task is defined.

peterhudec:
I have Enzyme tests ready for this, but they can't be used since we upgrade to React 17 as [there's not yet an Enzyme adapter for React 17](https://github.com/enzymejs/enzyme/issues/2429).

src/client/components/Task/reducer.js
```diff
@@ -17,20 +22,26 @@ const setTaskState = (state, { name, id, ...action }, status) => {
...
 }
 
+const remove = (state, { name, id }) => {
+  const taskState = state[name]
+  return taskState
+    ? _.omit(
+        state,
+        _.isEqual(_.keys(taskState), [name]) ? name : `${name}.${id}`
+      )
+    : state
```
paulgain, line 32: Tut-tut... `_.omit` and `_.isEqual`

LaurenQurashi (approved):
Looks good to me!

elcct (approved):

ian-leggett (changes requested):
Can we talk about this PR as I can't run it. Also there are no tests around this work so I am not confident that it won't break anything in the app. The code looks fine, I am just a bit nervous ticking this off without being able to run it.

peterhudec:
> Can we talk about this PR as I can't run it. Also there are no tests around this work so I am not confident that it won't break anything in the app. The code looks fine, I am just a bit nervous ticking this off without being able to run it.

@ian-leggett the Storybook example should work now that I added the missing file with the task definition. Yeah, the PR it's touching the `Task` mechanism, but we probably have all the code that uses it covered with unit and e2e tests and they are still passing. The cancellation mechanism itself is not tested though. The Enzyme adapter is still not around and probably [never will be](https://github.com/enzymejs/enzyme/issues/2429#issuecomment-820207891), but I could theoretically write tests with [React Testing Library](https://testing-library.com/docs/react-testing-library/intro/) as suggested by the omniscient Dan Abramov.

ian-leggett (approved):
After our good conversation around this I understand that although ideal, unit testing this would prove to be quite difficult at this stage. I have had a good manual test and as our current tests are green and we need this for a new feature that is being developed I am happy to approve. It would be really good to get some unit testing in place soon, we should get Cypress upgraded to 7 so we can take advantage of the component test runner.

PR #3055, opened 2020-11-02
New Chip component

ian-leggett:
## Description of change
This Chip component is to be used to indicate what filters are currently selected on the new react collection list page. Currently we only have one filter (Adviser typeahead filter) so this work has been built around this feature for now until we bring in more filters. 

Trello - https://trello.com/c/LuGgtZGp/533-add-filter-indicator-to-the-react-investments-collection-list

Note: This is being merged into a feature branch NOT master as its part of the React collection list work.

## Test instructions
1. Run the app and visit the investment project collection page
2. Run storybook and you should see the documentation for it.

## Screenshots
Screenshot 2020-11-02 at 12 45 16

## Checklist

### DIFF
--- a/src/client/components/AdviserFilterChips/index.jsx
+++ b/src/client/components/AdviserFilterChips/index.jsx
@@ -2,11 +2,10 @@ import React from 'react'
 import { Chip } from '../../components'
 
 const AdviserFilterChips = ({ selectedAdvisers }) => {
-  const advisers = selectedAdvisers.map(({ advisers }) => advisers)
   return (
     <>
-      {advisers.map(({ name, id }) => (
-        <Chip label={name} key={id} />
+      {selectedAdvisers.map(({ advisers: { name, id } }) => (
+        <Chip key={id}>{name}</Chip>
       ))}
     </>
   )
--- a/src/client/components/AdviserFilterChips/index.jsx
+++ b/src/client/components/AdviserFilterChips/index.jsx
@@ -1,0 +1,15 @@
+import React from 'react'
+import { Chip } from '../../components'
+
+const AdviserFilterChips = ({ selectedAdvisers }) => {
+  const advisers = selectedAdvisers.map(({ advisers }) => advisers)
+  return (
+    <>
+      {advisers.map(({ name, id }) => (
+        <Chip label={name} key={id} />
+      ))}
+    </>
+  )
+}
+
+export default AdviserFilterChips
--- a/src/client/components/Chip/index.jsx
+++ b/src/client/components/Chip/index.jsx
@@ -12,10 +12,10 @@ const StyledSpan = styled('span')`
   border-radius: ${SPACING.SCALE_2};
 `
 
-const Chip = ({ label }) => <StyledSpan>{label}</StyledSpan>
+const Chip = ({ children }) => <StyledSpan>{children}</StyledSpan>
 
 Chip.propTypes = {
-  label: PropTypes.string.isRequired,
+  children: PropTypes.string.isRequired,
 }
 
 export default Chip
--- a/src/client/components/Chip/index.jsx
+++ b/src/client/components/Chip/index.jsx
@@ -2,13 +2,14 @@ import React from 'react'
 import styled from 'styled-components'
 import PropTypes from 'prop-types'
 import { GREY_3 } from 'govuk-colours'
+import { SPACING } from '@govuk-react/constants'
 
 const StyledSpan = styled('span')`
   display: inline-block;
   padding: 12px;
   margin: 4px;
   background-color: ${GREY_3};
-  border-radius: 10px;
+  border-radius: ${SPACING.SCALE_2};
 `
 
 const Chip = ({ label }) => <StyledSpan>{label}</StyledSpan>
--- a/src/client/components/Chip/index.jsx
+++ b/src/client/components/Chip/index.jsx
@@ -1,0 +1,20 @@
+import React from 'react'
+import styled from 'styled-components'
+import PropTypes from 'prop-types'
+import { GREY_3 } from 'govuk-colours'
+
+const StyledSpan = styled('span')`
+  display: inline-block;
+  padding: 12px;
+  margin: 4px;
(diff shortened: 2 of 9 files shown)


mcbhenwood:
Yes, I think this is entirely the sort of simple, step-by-step, thin-layer iteration we need to adopt to build out these changes longterm. Thanks @ian-leggett. Not approving because that feels a bit unfair to others since we arrived at these decisions together.

src/client/components/AdviserFilterChips/index.jsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react'
+import { Chip } from '../../components'
+
+const AdviserFilterChips = ({ selectedAdvisers }) => {
+  const advisers = selectedAdvisers.map(({ advisers }) => advisers)
+  return (
+    <>
+      {advisers.map(({ name, id }) => (
+        <Chip label={name} key={id} />
+      ))}
```
peterhudec, line 10: You are mapping twice, which can be done in a single map.

src/client/components/Chip/index.jsx
```diff
@@ -0,0 +1,21 @@
...
+
+const StyledSpan = styled('span')`
+  display: inline-block;
+  padding: 12px;
+  margin: 4px;
+  background-color: ${GREY_3};
+  border-radius: ${SPACING.SCALE_2};
+`
+
+const Chip = ({ label }) => <StyledSpan>{label}</StyledSpan>
```
peterhudec, line 15: I would prefer `children` instead of `label`, it is more idiomatic React. Then you could just export the `StyledSpan`.

src/client/components/FilteredCollectionList/FilteredCollectionHeader.jsx
```diff
@@ -0,0 +1,85 @@
...
+      Add {collectionName}
+    </Button>
+  )
+
+  return (
+    <CollectionHeaderRowContainer>
+      <CollectionHeaderRow actions={actions}>
+        <StyledHeaderText>
+          <StyledResultCount>{formattedTotal}</StyledResultCount>
+          {` ${counterSuffix}`}
```
peterhudec, line 63: No need to wrap it in template string.

LaurenQurashi (approved):
Looks good! 🎉 🤠

src/client/components/FilteredCollectionList/index.jsx
```diff
@@ -24,6 +24,7 @@ const FilteredCollectionList = ({
   children,
   collectionName,
   activePage = 1,
+  selectedAdvisers,
```
peterhudec, line 27: Isn't this too specific? I meand shouldn't the `FilteredCollectionList` be a generic component?

peterhudec (changes requested):
Looks good, just a couple of things.

src/client/components/FilteredCollectionList/index.jsx
mcbhenwood: Yep but @peterhudec in this workstream we are employing the Xtreme tactic of only doing the very specific work required for this one story. At the end of this one story / PR there will be a filtered collection list whose only filter is one of Advisers. Hence the naming. If in later work we add other types then we will revisit the structure of the arguments etc.

This may seem crazy but this way of working yields more rapid iterations and makes the structural work an act of refactoring things which already function (rather than developing and architecting all in one move)

src/client/components/FilteredCollectionList/FilteredCollectionHeader.jsx
ian-leggett: This was done to give space between the number of results and text. I removed it and replaced it with a space in brackets.

src/client/components/AdviserFilterChips/index.jsx
ian-leggett: Good spot, not sure what I was thinking here

src/client/components/Chip/index.jsx
```diff
@@ -0,0 +1,21 @@
...
+  border-radius: ${SPACING.SCALE_2};
+`
+
+const Chip = ({ children }) => <StyledSpan>{children}</StyledSpan>
+
+Chip.propTypes = {
+  children: PropTypes.string.isRequired,
+}
+
+export default Chip
```
peterhudec, line 21: This is equivalent to:
```js
const Chip = StyledSpan
```
But worse, because you are not forwarding the `className`, which will prevent the `Chip` to be further stylable with `styled-components`.

You can just name the styled span `Chip` instead of `StyledSpan` and directly export it. You don't even need to define the prop types as every intrinsic component accepts `children`:

```js
const Chip = styled('span')`
  display: inline-block;
  padding: 12px;
  margin: 4px;
  background-color: ${GREY_3};
  border-radius: ${SPACING.SCALE_2};
`

export default Chip
```

src/client/components/FilteredCollectionList/FilteredCollectionHeader.jsx
```diff
@@ -0,0 +1,85 @@
...
+
+FilteredCollectionHeader.propTypes = {
+  totalItems: PropTypes.number.isRequired,
+  collectionName: PropTypes.string.isRequired,
+  addItemUrl: PropTypes.string,
+}
+
+FilteredCollectionHeader.defaultProps = {
+  addItemUrl: null,
+}
```
peterhudec, line 83: We should not use the prehistoric `defaultProps` now that we can use JS's param defaults.

peterhudec (changes requested):
Almost there, just two things.

src/client/components/Chip/index.jsx
ian-leggett: Much nicer thanks 👍

src/client/components/FilteredCollectionList/FilteredCollectionHeader.jsx
ian-leggett: Agreed, forgot that was in there!

peterhudec (approved):
Immaculate 🥇

PR #3100, opened 2020-11-27
Bug fix inputs not clearing

tomkingston:
## Description of change

Date Filter and Sort By Inputs were not clearing after clicking clear all or clearing any of the chips - they continued to show the value that the user selected previously. Interestingly the `TypeAhead` components are cleared as expected.

This is because the Date Filter and Sort By Inputs are uncontrolled components, so setting a new value via props will not override the user's input. React's [documentation on uncontrolled components](https://reactjs.org/docs/uncontrolled-components.html) explains:
> Changing the value of defaultValue attribute after a component has mounted will not cause any update of the value in the DOM.

To get around this, I have refactored the `DateField` and `Sort` (now `Select`) as controlled components. I can then use the `useEffect` hook to override the user input whenever the `initialValue` prop changes. This means that whenever the location is updated, these inputs take on the new value and overwrite the previous user input.

I believe this is also the reason `TypeAhead` component works - it manages its own state and updating `value` causes it to re-render. On that note, I could rename the `initialValue` prop to `value` for consistency with that component.

@ian-leggett, @mcbhenwood and I have discussed the issue and how it uses internal state and have agreed that this approach should be reasonable, however it would be good to hear any other opinions.

Note that the Sort component is also used on the dashboard where it is not routed.

## Test instructions

On `investments/projects` :

**Date filters**

* Apply date filters 
* Clear them by removing the chip or clicking clear all 
* This should clear the filter from the url and also clear the value in the date filter input.

**Sort by**

* Choose a sort by filter
* Clear it by clicking clear all
* The sort by should be cleared from the url and the input should reset to the first item in the dropdown

## Checklist


src/client/components/Select/index.jsx
```diff
@@ -0,0 +1,66 @@
+import React, { useState, useEffect } from 'react'
```
tomkingston, line 1: I'm happy to rename this back to `Sort` - I created this because I did have two versions at one point (one for the dashboard and one for the routed components). Select seemed to make more sense to me because it appears that there is nothing inherently sorty about this component.

src/client/components/DateField/index.jsx
```diff
@@ -20,17 +20,32 @@ const DateField = ({
   legend,
   hint,
   initialValue,
```
tomkingston, line 22: I am using `initialValue` as the prop input here, but perhaps `value` makes more sense (it would also be consistent with the `TypeAhead` component).

ian-leggett (approved):
After discussing this with yourself and Mark I agree that in the context of these filters the use of internal state seems appropriate.

PippoRaimondi (approved):

PR #3116, opened 2020-12-01
Refactored Routed Chips into a Single Component

tomkingston:
## Description of change

Refactors the three routed chip components (`RoutedOptionFilterChips`, `RoutedDateFilterChips` and `RoutedAdviserFilterChips`) into a single `RoutedFilterChips` component.

## Test instructions

As a refactor, everything should work as before - however, I have also made the mock adviser call a bit more clever so it should now return the specific adviser that you selected instead of always showing `Puck Head`. This was needed to ensure that the route and state were consistent with each other with regards to the selected adviser.

To test that the chips are working correctly, try adding various filters, and clearing them by clicking on the chip and the clear all filters option.

## Checklist


src/client/components/RoutedFilterChips/index.jsx
```diff
@@ -0,0 +1,66 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import qs from 'qs'
+import { Route } from 'react-router-dom'
+import { omit } from 'lodash'
+import dateFns from 'date-fns'
+
+import { Chip } from '..'
+
+const removeParamFromQs = (qsParams, queryParam, targetValue = null) => {
```
tomkingston, line 10: After saying to try and remove comments where possible, I've added quite a bit of a docstring to this one! However, I think this does help clarify what this function is doing a bit more as it is not that obvious.

src/client/components/RoutedFilterChips/index.jsx
```diff
@@ -0,0 +1,66 @@
...
+
+import { Chip } from '..'
+
+const removeParamFromQs = (qsParams, targetParam, targetValue = null) => {
+  /*
+   * Removes the targetParam with targetValue from qsParams
+   *
+   * When the targetParam corresponds to an array of values it removes the item
+   * with targetValue, otherwise it removes that param entirely.
+   */
```
tomkingston, line 16: After saying to remove comments on another PR, I've ended up adding some here! Just did this because it didn't seem immediately clear what this function does exactly!

src/apps/investments/client/projects/state.js
```diff
@@ -41,16 +41,26 @@ const listSelectedFilters = (metadataOptions, filterProp) =>
...
+  /*
+   * Convert both location and redux state to investment projects props
+   *
+   * Selected filters are built from the page's query props as well as the
+   * selected advisers in state.
+   */
   const queryProps = qs.parse(router.location.search.slice(1))
   const filteredQueryProps = collectionListPayload(queryProps)
-
+  const { selectedAdvisers } = state.projectsList
```
tomkingston, line 52: The previous chips were getting the selectedAdvisers from a prop, but it made sense to me that these should also be present in the selectedFilters object.

src/apps/investments/client/projects/state.js
```diff
@@ -40,27 +41,46 @@ const collectionListPayload = (paramProps) => {
...
+      value: advisers.id,
+    })),
     selectedSectors: listSelectedFilters(sectorOptions, sector_descends),
     selectedEstimatedLandDatesBefore: [
       {
-        label: estimatedLandDateBefore,
+        label: getDateLabel(
+          estimatedLandDateBeforeLabel,
+          estimated_land_date_before
+        ),
```
tomkingston, line 74: Rather than building the date label in the component, it made more sense to prepare it here - that way, there is no need to specify to the component how it should format the label

test/sandbox/routes/adviser.js
```diff
@@ -1,14 +1,26 @@
...
+   * Use the adviser id provided to find the adviser from the autocomplete list
+   */
+  const pathComponents = req.path.split('/')
+  const adviserId = pathComponents[pathComponents.length - 2]
+  let adviser = autoCompleteAdvisers.results.find(({ id }) => id === adviserId)
+  if (!adviser) {
+    adviser = Object.assign({}, singleAdviser)
+    adviser.id = adviserId
+  }
+  res.json(adviser)
```
tomkingston, line 25: This now gets the adviser that was actually specified rather than just `Puck Head` - this was needed because I started using the selectedAdvisers from the state.

src/apps/investments/client/projects/state.js
tomkingston: I'm wondering whether the date and sector filters should also be made present in the state?

ian-leggett: This will probably happen when we retrieve the metadata dynamically. The reason the "selectedAdvisers" are in the global state object is because we need to do a second call to the api and get the label values from the results and store them in state whereas with the others we already have access to the label as they are static. I think we will possible play this card soon.

ian-leggett (approved):
Great refactor 👍

test/sandbox/routes/adviser.js
```diff
@@ -1,14 +1,26 @@
...
+ */
 exports.singleAdviser = function (req, res) {
-  res.json(singleAdviser)
+  const pathComponents = req.path.split('/')
+  const adviserId = pathComponents[pathComponents.length - 2]
+  let adviser = autoCompleteAdvisers.results.find(({ id }) => id === adviserId)
+  if (!adviser) {
+    adviser = Object.assign({}, singleAdviser)
+    adviser.id = adviserId
+  }
```
paulgain, line 24: You can use the spread syntax here instead of the old `Object.assign`.
```js
if (!adviser) {
    adviser = { ...singleAdviser, id: adviserId }
}
```

rachaelcodes (approved):
This is nice and tidy :+1: Also appreciate what you've done adding the docstrings

paulgain (approved):
LGTM Tom 👍

PR #3117, opened 2020-12-01
Integrate Great.gov.uk form(s) into Data Hub

paulgain:
## Description of change

This piece of work integrates Great.gov.uk forms into Data Hub, initially it's just the `Export enquiry` form, more will be added over time. The new functionality is hidden behind a feature flag called `activity-feed-export-enquiry` which is currently turned on in `dev` and `staging` but off in `prod.`

To return `Export enquiry` documents from Elasticsearch four `terms` must match:

**1. Ensures we only match Great.gov.uk forms (which there are many)**
```js
{
   term: {
     'object.type': 'dit:directoryFormsApi:Submission'
   }
}
```

**2. Ensures the document returned is in JSON format (there are two other formats HTML and Text)**
```js
{
  term: {
    'object.attributedTo.type': 'dit:directoryFormsApi:SubmissionAction:gov-notify-email',
   }
}
````

**3. Ensures we only match Export enquires**
This is the great.gov.uk path of the form _(e.g. great.gov.uk/contact/export-advice/comment)_
```js
{
  term: {
    'object.url': '/contact/export-advice/comment/'
  }
}
```

**4. Ensures we match a Data Hub company contact to the user filling out the form at Great.gov.uk**
```js
{
  terms: {
    'actor.dit:emailAddress': [
      ...contacts.map((contact) => contact.email),
    ]
  }
}

```

It's far easier to read and maintain our own Elasticsearch queries compared to using `bodybuilder` (an ES query builder) which I've removed due to its inflexibility.

## Test instructions

Bring up the FE locally and point to the dev API on PaaS, then go to: `/companies/6df487c5-7c75-4672-8907-f74b49e6c635/activity` and apply the `All external activity` filter. 

You should see two external activities:

great gov uk-activity

### DIFF
--- a/src/apps/companies/apps/activity-feed/constants.js
+++ b/src/apps/companies/apps/activity-feed/constants.js
@@ -1,55 +1,14 @@
-const ES_KEYS = {
-  companiesHouseAccount: 'dit:Accounts',
-  companiesHouseCompany: 'dit:Company',
-  hmrcExporter: 'dit:Export',
-  interaction: 'dit:Interaction',
-  serviceDelivery: 'dit:ServiceDelivery',
-  investmentProject: 'dit:InvestmentProject',
-  omis: 'dit:OMISOrder',
-  type: 'object.type',
-  attributedTo: 'object.attributedTo.id',
-  companyReferral: 'dit:CompanyReferral',
-}
-
-const ES_KEYS_GROUPED = {
-  allActivity: [
-    ES_KEYS.companiesHouseAccount,
-    ES_KEYS.companiesHouseCompany,
-    ES_KEYS.hmrcExporter,
-    ES_KEYS.interaction, // Interaction
-    ES_KEYS.serviceDelivery, // Interaction
-    ES_KEYS.investmentProject,
-    ES_KEYS.omis,
-    ES_KEYS.companyReferral,
-  ],
-
-  externalActivity: [
-    ES_KEYS.companiesHouseAccount,
-    ES_KEYS.companiesHouseCompany,
-    ES_KEYS.hmrcExporter,
-  ],
-
-  dataHubActivity: [
-    ES_KEYS.interaction, // Interaction
-    ES_KEYS.serviceDelivery, // Interaction
-    ES_KEYS.investmentProject,
-    ES_KEYS.omis,
-    ES_KEYS.companyReferral,
-  ],
-}
-
 const FILTER_KEYS = {
-  allActivity: 'allActivity',
+  dataHubAndExternalActivity: 'dataHubAndExternalActivity',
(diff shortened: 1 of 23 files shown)


src/apps/companies/apps/activity-feed/constants.js
```diff
@@ -68,9 +27,29 @@ const FILTER_ITEMS = {
   },
 }
 
+const DATA_HUB_ACTIVITY = [
+  'dit:Interaction', // Interaction
+  'dit:ServiceDelivery', // Interaction
```
ambrozic, line 32: `ServiceDelivery` instead of `Interaction` or just remove both comments. are not adding anything to context.

ambrozic (changes requested):

src/apps/companies/apps/activity-feed/constants.js
paulgain: Both `dit:Interaction` and `dit:ServiceDelivery` are Interactions (hence the comment). 

**The context lies in the code**
[Here](https://github.com/uktrade/data-hub-frontend/blob/feature/great-gov-uk-forms/src/client/components/ActivityFeed/constants.js#L54) they're grouped. The reason they are grouped is because the [Interaction Component](https://github.com/uktrade/data-hub-frontend/blob/feature/great-gov-uk-forms/src/client/components/ActivityFeed/activities/Interaction.jsx#L27) renders both of them providing the activity type matches one of them.

If the comment is unclear, I can remove them both and add this:
```js
const INTERACTIONS = ['dit:Interaction', 'dit:ServiceDelivery']

const DATA_HUB_ACTIVITY = [
  ...INTERACTIONS,
  'dit:InvestmentProject',
  'dit:OMISOrder',
  'dit:CompanyReferral',
]
```
To me this is the same thing, however, it's just adding superfluous code.

src/client/components/ActivityFeed/__fixtures__/index.js
```diff
@@ -1,6 +1,7 @@
 import accountsAreDueFixture from './companies_house/accounts_are_due.json'
 import incorporatedFixture from './companies_house/incorporated.json'
 import exportOfGoodsFixture from './hmrc/export_of_goods.json'
+import exportAdviceFormFixture from './great_gov_uk_forms/export-enquiry.json'
```
mforner13, line 4: I think it could be good to refer to it as `exportEnquiryForm...` in-line with the UI to avoid confusion.

paulgain: Good spot 👍

ambrozic (approved):

mforner13 (approved):
Looks great!

PR #3130, opened 2020-12-04
Add actual land date filters

tomkingston:
## Description of change

Adds actual land date filters to the investment projects page.

I have also refactored the cypress tests for the filters to use ids instead of traversing the DOM tree. My justification for this is:

1. Tests should fail because the thing they are testing fails. If the structure of the page is changed, it will be confusing to see failed tests saying "it should filter the actual land date". I have added a new test 'it should contain filter fields in the right order' to ensure that the page is composed correctly without breaking other tests.
2. Related to number one, but using unusual descendant selectors can make the tests quite fragile, so a minor change to something unrelated could break all of the tests
3. Readability / maintainability and time taken to understand the code. It is not clear what long selectors like `main article div + div button` actually mean; by using ids, these selectors become quite a lot clearer, so the previous example would become `#filter-chips button`.

As part of the refactor, I have also changed the helper functions to use named props - this is to make it clearer what they are doing when you call them. I have also updated the names of a couple of 'assert' functions to be prefixed with 'test' because they actually perform actions that have side effects. Finally, I have also added a few docstrings to explain what the different functions do.

## Test instructions

There should now be filters for actual land date before and after. Selecting these filters should show a chip above the list of projects. Clicking on the chip should remove the filter, as should clicking 'remove all filters'.

## Screenshots

Screenshot from 2020-12-04 17-23-59

## Checklist

### DIFF
--- a/test/functional/cypress/specs/investments/filter-spec.js
+++ b/test/functional/cypress/specs/investments/filter-spec.js
@@ -5,8 +5,14 @@ const ADVANCED_ENGINEERING_SECTOR_ID = 'af959812-6095-e211-a939-e4115bead28a'
 const UK_COUNTRY_ID = '80756b9a-5d95-e211-a939-e4115bead28a'
 const SOUTH_EAST_UK_REGION_ID = '884cd12a-6095-e211-a939-e4115bead28a'
 
-const selectAdvisersTypeahead = (fieldName, input) =>
-  cy.get(fieldName).within(() => {
+/**
+ * Enter `input` into an advisers typeahead `element` and select the first result
+ *
+ * This waits for the adviser api request to complete before selecting the
+ * first option.
+ */
+const selectFirstAdvisersTypeaheadOption = ({ element, input }) =>
+  cy.get(element).within(() => {
     cy.server()
     cy.route('/api-proxy/adviser/?*').as('adviserResults')
     cy.get('div').eq(0).type(input)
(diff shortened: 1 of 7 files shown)


ian-leggett (approved):
Looks good and a great refactor to the tests 👍

test/functional/cypress/specs/investments/filter-spec.js
```diff
@@ -5,115 +5,197 @@ const ADVANCED_ENGINEERING_SECTOR_ID = 'af959812-6095-e211-a939-e4115bead28a'
...
+ */
+const assertTypeaheadOptionSelected = ({ element, expectedOption }) => {
+  cy.get(element).should('contain', expectedOption)
+}
+
+/**
+ * Asserts that a chip indicator exists in the specified position
+ */
+const assertChipExists = ({ label, position }) => {
+  cy.get(`#filter-chips button:nth-child(${position})`).should((el) => {
```
rachaelcodes, line 58: Nice - this is a really tidy selector that seems a lot less brittle 👍

test/functional/cypress/specs/investments/filter-spec.js
```diff
@@ -5,115 +5,197 @@ const ADVANCED_ENGINEERING_SECTOR_ID = 'af959812-6095-e211-a939-e4115bead28a'
...
 
     it('should filter by sector', () => {
-      assertTypehead(
-        '@sectorFilter',
-        'Sector',
-        'Search sectors',
-        'adv',
-        'Advanced Engineering'
-      )
+      testTypeahead({
```
rachaelcodes, line 151: I really like how readable this test has become :)

rachaelcodes (approved):
The tests in particular look really good in this - it'd be great to have this example shared

elcct (approved):

PR #3150, opened 2020-12-11
Replace dashboard unit tests with functional

tomkingston:
## Description of change

Replaces the old dashboard collections test which was coupled to axios with new tests to ensure that the correct tabs are displayed on the dashboard and that the help centre updates are displayed.

The lists and referrals are already covered by other tests.

Note: the failed checks on test coverage are related to me un-indenting some existing code which is now being picked up as an untested change.

## Checklist


src/client/components/TabNav/index.jsx
```diff
@@ -131,90 +131,85 @@ const TabNav = ({
...
   routed,
-  // We are only destructuring dispatch from the props so that <div> won't
-  // complain about it being an unexpected prop.
-  // eslint-disable-next-line no-unused-vars
-  dispatch,
-  ...props
 }) => {
   const tabKeys = Object.keys(tabs)
   return (
-    <div {...props}>
```
tomkingston, line 142: This was not needed - it passed on various object props that couldn't be parsed anyway  - take a look at dev and you'll see this:
`<div history="[object Object]" location="[object Object]" match="[object Object]">`

src/templates/_components/info-feed.njk
```diff
@@ -24,16 +24,24 @@
  #}
 
 {% set feedLimit =  feedLimit | default(5) %}
-<div class="dashboard-section">
-{% if outboundLinkText %} <a href="{{ outboundLinkURL }}" class="dashboard-section__top-link" target="_blank">{{ outboundLinkText }}</a> {% endif %}
-{% if heading %} <h1 class="govuk-heading-m">{{heading}}</h1> {% endif %}
+<div class="dashboard-section" data-cy="info-feed">
```
tomkingston, line 27: Using `data-cy` as suggested by the cypress docs. Would be good to make sure that this is consistent across the project, so let me know if you think another name would be better.

test/sandbox/fixtures/help-centre/announcement.js
```diff
@@ -1,6 +1,9 @@
 var now = new Date()
 now.setHours(now.getHours() - 1.1)
 
+var yesterday = new Date()
```
tomkingston, line 4: Set to yesterday so that the date in the feed shows as "a day ago" even when you started the sandbox server a number of hours ago (ie to make sure the test passes... unless you have been running the sandbox server for more than a day).

src/client/components/Dashboard/index.jsx
```diff
@@ -15,7 +15,7 @@ const StyledDiv = styled('div')`
 `
 
 const Dashboard = ({ id }) => (
-  <StyledDiv>
+  <StyledDiv data-cy="dashboard-tabs">
```
tomkingston, line 18: Using data-cy as suggested by the cypress docs. Would be good to make sure that this is consistent across the project, so let me know if you think another name would be better.

src/templates/_components/info-feed.njk
```diff
@@ -24,16 +24,24 @@
...
+{% endif %}
+{% if heading %} <h1 class="govuk-heading-m" data-cy="info-feed-heading">{{heading}}</h1> {% endif %}
 {% if subHeading %} <p class="dashboard-section__subheading">{{subHeading}}</p> {% endif %}
 {% if dataFeed.length %}
-  {% for item in dataFeed.slice(0, feedLimit) %}
-    <a href="{{ item.link }}" class="dashboard-section__info-feed-link" target="_blank">{{ item.heading }}</a>
-    <span class="dashboard-section__info-feed-link-note">(Link opens in a new window)</span>
-      <time class="dashboard-section__info-feed-date">{{ item.date }}</time>
-  {% endfor %}
+  <ul data-cy="info-feed-list">
```
tomkingston, line 36: Added `data-cy` for testing - this also seems to be more correct semantically

src/client/components/Dashboard/index.jsx
PippoRaimondi: I think we decided to use `data-auto-id`... not super fussed about what should come after 'data-' but if we want to keep it consistent to the rest of the selectors I would suggest sticking to `data-auto-id`

tomkingston: Thanks Pippo - yeah, good to keep it consistent. Just wondering what the thinking is behind `data-auto-id` since if anything it should be `data-manual-id`?

src/apps/dashboard/controllers.js
```diff
@@ -20,6 +20,7 @@ async function renderDashboard(req, res, next) {
         auth: {
           bearer: config.helpCentre.token,
         },
+        qs: { test: req.query.test },
```
tomkingston, line 23: Injecting a test parameter to test scenarios where the api is unavailable etc (as discussed).

rachaelcodes (approved):

PippoRaimondi (approved):

PR #3161, opened 2020-12-14
Change task error heading to div

peterhudec:
## Description of change

Changed the H2 heading in the default `Task` error view to a `div`.

## Test instructions

The task error view should not have a H2 element.

## Screenshots

There should be no visible change.

### DIFF
--- a/src/client/components/Task/Error.jsx
+++ b/src/client/components/Task/Error.jsx
@@ -39,7 +39,7 @@ const StyledSecondaryButton = styled(SecondaryButton)({
 
 const Err = ({ errorMessage, retry, noun }) => (
   <StyledRoot>
-    <H2 size="MEDIUM">Could not load {noun}</H2>
+    <H2 as="div">Could not load {noun}</H2>
     <p>Error: {errorMessage}</p>
     <StyledSecondaryButton onClick={retry}>Retry</StyledSecondaryButton>
   </StyledRoot>


src/client/components/Task/Error.jsx
```diff
@@ -38,7 +38,7 @@ const StyledSecondaryButton = styled(SecondaryButton)({
 
 const Err = ({ errorMessage, retry, noun }) => (
   <StyledRoot>
-    <H2 size="MEDIUM">Could not load {noun}</H2>
+    <H2 as="div">Could not load {noun}</H2>
```
ian-leggett, line 41: Why as `div`, can't we use a `p` and style it? Seems a little strange this one.

peterhudec: > it looks as if we are just keeping the `H2` component for styling only

We are in deed. It's the simplest way to keep the exact appearance. I just didn't know what element to replace it with, so I chose a div. @LaurenQurashi do you have any suggestion for how to structure the error message for a11y?

LaurenQurashi: This is a bit of a hard one! 

It's pretty clear that in the guidelines anything that is text should be labelled as such so that screen readers can get a good map of what the page looks like. We can't really be using a div to label text / headings - so it's gotta be somewhere in the range from H1-p tag.  

Looking through some articles (as it's all a bit of a grey area) I think that the styles of the text don't matter as long as you're not making a p tag try to look visually like a heading. (https://www.nomensa.com/blog/2017/how-structure-headings-web-accessibility) So my best guess would be to use a H3 and make it look like a H2... Or if this is a common problem, maybe we can make our own error message class of H3? H2's should be for new sections. I've not seen the task error visually, but if we're trying to make this text look anything other than a new section heading then maybe we need to rethink the design of the error message in general? I hope this helps but let me know if it doesn't make sense! 🐝

peterhudec: I see a couple of options:

* If the error box must have a heading (if it can't be a `<p/>`), we can allow to specify the level of the heading on `Task.Status`
* There can actually be multiple `H1`s on the page, but each one must be wrapped in a `<section/>` so the error box could be a `<section/>` with the heading being an H1
* How about using the `<heading/>` element?

ian-leggett: If it's a heading why can't it be a `<h2>`? If it has its own sectioning element then it can be a `<h1>` Do you have a screenshot of this error in the context of a page?

ian-leggett (approved):
This is currently flagging up as an accessibility issue so I think this is the safest and quickest fix.

peterhudec:
Closing this as it's a fix to an issue proactively raised by us and we don't really know what should be the solution from a11y point of view. Let's revisit this when someone is actually complaining about it.

peterhudec:
Closing this as it's a fix to an issue proactively raised by us and we don't really know what should be the solution from a11y point of view. Let's revisit this when someone is actually complaining about it.

PR #3169, opened 2020-12-17
Fix heading structure of ActivityFeed

peterhudec:
## Description of change

Fixes heading level inconsistency in the activity stream.

## Test instructions

* The activity stream item heading should now be a H3
* The _Filter by_ should not be a heading
* The item heading of the referral list on the dashboard should not be affected by this change

## Screenshots

There should be no visible change

### DIFF
--- a/src/client/components/ActivityFeed/activities/card/CardHeader.jsx
+++ b/src/client/components/ActivityFeed/activities/card/CardHeader.jsx
@@ -97,6 +97,7 @@ const CardHeader = ({
   heading,
   startTime,
   badge,
+  headingLevel = 3,
 }) => (
   <>
     {company && company.name && (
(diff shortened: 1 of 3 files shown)


src/client/components/ActivityFeed/activities/card/CardHeader.jsx
```diff
@@ -109,8 +110,9 @@ const CardHeader = ({
         )}
 
         {subHeading && <StyledSubHeading>{subHeading}</StyledSubHeading>}
-
-        {heading && <StyledHeading level={2}>{heading}</StyledHeading>}
+        {heading && (
+          <StyledHeading level={headingLevel}>{heading}</StyledHeading>
```
PippoRaimondi, line 114: I will leave this pre-approved but just wanted to confirm we are happy moving from headingLevel 3 vs level 2 prior to your change

PippoRaimondi (approved):

src/client/components/ActivityFeed/activities/card/CardHeader.jsx
PippoRaimondi: based on your PR comments I guess it is... I will resolve this thread.

elcct (approved):

PR #3196, opened 2020-12-29
Chore/replace static metadata

tomkingston:
## Description of change

Loads the options for the various investment project filters from the api so that we do not need to use the hardcoded values any more. The existing functionality should remain the same.

I actually started using a different approach which meant I added functionality for the `CheckboxGroupField` to load options from a dynamic source. This turned out to be unneeded but will hopefully be useful for future work.

The approach here is based on Ian's work here: https://github.com/uktrade/data-hub-frontend/compare/feature-merge/investment-project-collection-list...feature/remove-static-metadata

## Test instructions

The investments projects page should continue to show filter options which can be selected and cleared using the chips.

## Checklist


src/client/components/CheckboxGroupField/index.jsx
```diff
@@ -55,63 +55,84 @@ const StyledFieldWrapper = styled(FieldWrapper)`
  * @param {string} name - field name for the group
  * @param {string} hint - hint to display
  * @param {Array} options - the available options as objects with label and value
+ * @param {Func} loadOptions - function to load options
```
tomkingston, line 58: The work to add a `loadOptions` prop turned out to be unneeded, but I thought it might be useful for the future. On the other hand, there is a case for removing it on the basis that it is something extra to maintain and adds to code complexity.

ambrozic (approved):

src/client/components/CheckboxGroupField/index.jsx
ambrozic: Isn't `loadOptions` as a pattern used in most `typeahead` type use cases? I do think it is ok to leave it. I just used it in another project.

ian-leggett (approved):
Looks good to me 👍

PR #3203, opened 2020-12-31
Replace old collection data export test

rachaelcodes:
## Description of change

As part of the unit test removals to support axios, this replaces the old collection data export unit test.

## Checklist

- [ ] Has the branch been rebased to ~master~ chore-merge/requests-axios-replacement?


test/end-to-end/cypress/specs/DIT/export-data.js
```diff
@@ -0,0 +1,28 @@
...
+      expect(response.status).to.equal(200)
+      expect(response.headers['content-type']).to.equal(
+        'text/csv; charset=utf-8'
+      )
+    })
+  })
+}
+
+describe('exporting data', () => {
+  testDataExport({ name: 'companies' })
```
mcbhenwood, line 20: Does it work when you call a function synchronously like this during these tests? The call to `testDataExport()` doesn't happen when the tests are _run_ , it happens when the tests _compile_. I guess the `it()` calls in the `testDataExport` function would still enqueue a the right place in the compiled tests but I just wonder if you are assuming that these statements all unfold at test runtime, which they don't.

rachaelcodes: Screenshot 2020-12-31 at 14 15 55

It does seem to run them, but I can rewrite this so that it's clear when the calls happen.

tomkingston (approved):
Looks good (subject to Mark's comments)

test/end-to-end/cypress/specs/DIT/export-data.js
mcbhenwood: Maybe it would be clearer (i.e. possibly less ambiguous to others) if you renamed the function `queueDataExportTest`. Basically every time `it()` is called it just writes the enclosed anonymous function to the list of tests so far whereas the procedural syntax of JS makes it _look_ like `testDataExport` is called at test time. I just worry about it being a precedent which people adopt but which doesn't do what they think it is doing.

paulgain (approved):

PR #3211, opened 2021-01-05
Chore/react accessibility audit fixes

tomkingston:
## Description of change

Fixes accessibility issues:

- Wraps checkbox group fields in fieldset tag and uses legend for the group label
- Wraps typeaheads with fieldset tag
- Does not wrap date fields with fieldset tag since they do not have legends (just labels) - not 100% sure this is consistent with the accessibility guidance on forms which seems to suggest that absolutely everything should be wrapped in a fieldset: https://accessibility.18f.gov/forms/ - in the case of these dates, that would mean duplicating each label, so you would have "Estimated land date before" as the fieldset legend and "Estimated land date before" as the input label. These could be grouped, so the fieldset might be Estimated land date, with 2 inputs for before and after - I would suggest that this would be best played as a future story to build a "DateRangeInput" component that combines the two naturally.
- Landmarks: changes the filters `aside` to a `div` - aside landmarks should be at the root of the document and are for additional, non-critical content - if anything, the filters are better suited to a `nav` element
- Landmarks: removed the duplicate skip to content link which was not contained in an appropriate landmark
- Add aria-labels to date fields
- Storybook: add labels to typeahead examples

## Test instructions

- Run axe and wave accessibility checkers on the investments projects page
- Check the accessibility tab for storybook components
- Check accessibility checklist here: https://docs.google.com/spreadsheets/d/1ds5pBzEw-9AoHK6ouVptbRdB3f6Geru4yR1_R34sOS0/

Axe should only mention one item to review - the contrast of the watermark copyright crest in the bottom right.

## Checklist


ambrozic (approved):

rachaelcodes (approved):

src/templates/_layouts/template.njk
```diff
@@ -13,9 +13,13 @@
 {% if IS_XHR %}
   {% extends "./xhr.njk" %}
 {% else %}
+  {# This one: https://github.com/alphagov/govuk-frontend/blob/master/src/govuk/template.njk #}
```
tomkingston, line 16: In a previous commit, I actually copied the contents of that template and extended that. Not being in control of your own base template doesn't seem ideal!

ian-leggett (approved):

PR #3214, opened 2021-01-06
Restart express server on change rather than re-spawning it

tomkingston:
## Description of change

This is part of the work to enable hot reloading - this doesn't actually do hot reloading, but fixes a problem where webpack was spawning new express server instances each time a file changed. This should therefore make the dev server run a lot more quickly and there should be fewer issues with memory running out.

## Test instructions

Run the dev server and see that everything works as expected. Any changes you make should be reflected after refreshing the page (future work should make this happen without a refresh).

This needs to work through docker as well as natively.

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
     "start": "node --use-strict src/server.js",
     "start:coverage": "nyc --silent npm run start",
     "build": "npm run clean && webpack",
-    "develop": "NODE_ENV=development npm run watch:js:client -- --env backend=true",
+    "develop": "NODE_ENV=development npm run watch:js:client -- --env development",
     "watch:test": "npm run test:unit -- -w",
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -11,7 +11,7 @@ ifdef CI
 	start-command = up --build --force-recreate -d
 	cypress-args = -- --parallel --record --key $(CYPRESS_DASHBOARD_KEY) --ci-build-id $(CIRCLE_BUILD_NUM)
 else
-	start-command = up --build --force-recreate
```
tomkingston, line 14: Doing a full rebuild every time you run the make command is a bit of a pain, so this now only happens if you explicitly specify those flags yourself

rachaelcodes (approved):

paulgain (approved):
Works well Tom 👍

PR #3248, opened 2021-01-19
New dashboard layout for user testing...

ian-leggett:
## Description of change
We are currently designing a new dashboard and we need an alternative layout to test users against. We will use a piece of middleware called "layout-testing.js" which enables us to target a specific group of users (by team id) so we can gain feedback around the new layout and what components work well.

## Test instructions
To activate the new layout all you need to do is set a feature flag in the following format: `layoutTesting:1234` where `1234` is an id of a specific team. Once you set that team id you need to go into django and assign yourself to it, clear the session by logging out and then sign back in and you should see a new layout with the following text: "New dashboard starts here..." You should also see a query param in the url called `?layoutTesting=dashboard`, this is so we can track user behaviour around this specific layout.

## Checklist

### DIFF
--- a/src/apps/dashboard/views/dashboard-test.njk
+++ b/src/apps/dashboard/views/dashboard-test.njk
@@ -1,0 +1,14 @@
+{% extends "_layouts/template.njk" %}
+{# For the accessibility work, the search bar has been amended to have hint text and placeholder text.
+It seems to be that the label and hint are coupled up somewhere, so a label is required to show a hint.
+As nunjucks is on it's way out, a label of an empty space has been used as a work around for this issue. #}
+{% block local_header %}
+{% endblock %}
+
+{% block body_main_content %}
+<div class="grid-row dashboard">
+  <div class="govuk-grid-column-full" data-test="dashboard">
+    New dashboard starts here...
+  </div>
+</div>
+{% endblock %}
--- a/src/apps/dashboard/views/dashboard-test.njk
+++ b/src/apps/dashboard/views/dashboard-test.njk
@@ -1,7 +1,4 @@
 {% extends "_layouts/template.njk" %}
-{# For the accessibility work, the search bar has been amended to have hint text and placeholder text.
-It seems to be that the label and hint are coupled up somewhere, so a label is required to show a hint.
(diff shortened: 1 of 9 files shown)


cgsunkel (approved):

src/apps/dashboard/views/dashboard-test.njk
```diff
@@ -0,0 +1,14 @@
+{% extends "_layouts/template.njk" %}
+{# For the accessibility work, the search bar has been amended to have hint text and placeholder text.
+It seems to be that the label and hint are coupled up somewhere, so a label is required to show a hint.
+As nunjucks is on it's way out, a label of an empty space has been used as a work around for this issue. #}
```
tomkingston, line 4: Is this comment relevant here?

test/functional/cypress/specs/dashboard-new/feature-flag-spec.js
```diff
@@ -0,0 +1,55 @@
+describe('Dashboard - feature flag', () => {
+  context('when a feature flag is set and your are in the testing team', () => {
+    beforeEach(() => {
+      cy.setUserDitTeam('1234')
```
tomkingston, line 4: Looks useful - I wonder whether you should have a tear down to call `resetUserDitTeam` - otherwise, running this test will leave developers with the wrong team set on sandbox?

tomkingston (approved):
Looks good - just a couple of queries...

src/apps/dashboard/views/dashboard-test.njk
cgsunkel: This is an old comment from when we were doing the accessibility fixes

test/functional/cypress/specs/dashboard-new/feature-flag-spec.js
ian-leggett: I did have a tear down but I wasn't using it in the first context "afterEach" lifecycle hook, I was resetting it in the new context. I have moved this into the a lifecycle hook now in the first context 👍

src/apps/dashboard/views/dashboard-test.njk
ian-leggett: Good spot, all removed.

PR #3269, opened 2021-01-26
Re-work layout for new dashboard...

ian-leggett:
## Description of change
This PR just makes some slight changes to the layout, I now override the extended nunjucks template as I want to use an `aside` element in the layout which is now alongside the `main` container. I have also separated out the imports in the `PersonalisedDashboard` so they request re-usable components from the root components directory and ones specific to the dashboard will import using a relative path.

I have created `primary` and `secondary` container components too as these will wrap the components on the dashboard controlling things like, padding, margin, border etc etc.

## Checklist


elcct (approved):

src/client/components/PersonalisedDashboard/Aside.jsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react'
+
+import PrimaryContainer from './PrimaryContainer'
+
+const Aside = ({ children }) => (
+  <aside>
+    {children.map((child) => (
+      <PrimaryContainer>{child}</PrimaryContainer>
```
paulgain, line 8: You need to assign a unique `id` here for each container.

src/client/components/PersonalisedDashboard/Main.jsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react'
+
+import SecondaryContainer from './SecondaryContainer'
+
+const Main = ({ children }) => (
+  <main role="main" id="main-content" data-auto-id="bodyMainContent">
+    {children.map((child) => (
+      <SecondaryContainer>{child}</SecondaryContainer>
```
paulgain, line 8: Same as above.

src/client/components/PersonalisedDashboard/PrimaryContainer.jsx
```diff
@@ -0,0 +1,11 @@
+import styled from 'styled-components'
+import { SPACING } from '@govuk-react/constants'
+
+const PrimaryContainer = styled('div')`
+  padding: ${SPACING.SCALE_2};
+  margin-bottom: ${SPACING.SCALE_2};
+  border: 1px solid #454a4c;
```
paulgain, line 7: import { SOME_COLOUR } from 'govuk-colours'

src/client/components/PersonalisedDashboard/SecondaryContainer.jsx
```diff
@@ -0,0 +1,10 @@
+import styled from 'styled-components'
+import { SPACING } from '@govuk-react/constants'
+
+const SecondaryContainer = styled('div')`
+  padding: ${SPACING.SCALE_2};
+  margin-bottom: ${SPACING.SCALE_2};
+  border: 1px solid #454a4c;
```
paulgain, line 7: Same as above.

paulgain:
**The last tab is wrapping**
tab-is-wrapping

src/client/components/PersonalisedDashboard/PrimaryContainer.jsx
ian-leggett: I meant to remove this as it was just an example. The padding and margin is just made up, I don't know what it will be yet.

src/client/components/PersonalisedDashboard/SecondaryContainer.jsx
ian-leggett: Again I just added this as an example, I have removed it now.

src/client/components/PersonalisedDashboard/PrimaryContainer.jsx
paulgain: The border is fine, there's no need to remove it, I like the fact you can see the boundaries of each component. Can we just reduce the padding around the components so the last tab doesn't wrap?

src/client/components/PersonalisedDashboard/Aside.jsx
```diff
@@ -4,8 +4,8 @@ import PrimaryContainer from './PrimaryContainer'
 
 const Aside = ({ children }) => (
   <aside>
-    {children.map((child) => (
-      <PrimaryContainer>{child}</PrimaryContainer>
+    {children.map((child, i) => (
+      <PrimaryContainer key={i}>{child}</PrimaryContainer>
```
paulgain, line 8: I think adding the index as a key is OK in this scenario as the order of the items will never change.

paulgain: @ian-leggett you may find this useful: https://robinpokorny.medium.com/index-as-a-key-is-an-anti-pattern-e0349aece318

src/client/components/PersonalisedDashboard/Aside.jsx
ian-leggett: I just updated this, unfortunately I can't get an id or unique value from the child component so I used a template string with the first part being the component name and with the second part I concatenated the index value. So it will be `PrimaryContainer-1`

src/client/components/PersonalisedDashboard/PrimaryContainer.jsx
ian-leggett: I have removed all styles for now 👍

paulgain (approved):

PR #3300, opened 2021-02-04
Feature/investments profiles filters

peterhudec:
## Description of change

This PR adds the _Countries of origin_ typeahead filter to the `/investments/profiles` page. The feature is hidden behind the `capital-investments-filters` feature flag.

## Test instructions

1. Enable the `capital-investments-filters` feature flag in Django Admin
2. Go to the `/investments/profiles` page
3. There should be a sidebar with the _Countries of origin_ typeahead filter
4. Select one or multiple countries with the filter
5. The number of results should update and reflect the selected countries
6. There appear a _chip_ for each selected country in the header of the result list
7. Clicking on a chip should remove its corresponding country from the filter and the results should update
8. If the `capital-investments-filters` feature flag is disabled, the page should work as before this PR, i.e. should have no filter sidebar

### Before

Screen Shot 2021-02-04 at 5 31 28 PM

### After
Screen Shot 2021-02-04 at 5 30 29 PM

_Add a screenshot_


src/apps/investments/client/profiles/FilteredLargeCapitalProfileCollection.jsx
```diff
@@ -0,0 +1,88 @@
...
+  count,
+  results,
+  isComplete,
+  filterOptions,
+}) => (
+  <Route>
+    {({ location }) => {
+      const qsParams = qs.parse(location.search.slice(1))
+      const selectedCountries = (
+        qsParams.country_investment_originates_from || []
```
rachaelcodes, line 27: I don't think this is the right param: `country_investment_originates_from` belongs to investment projects. For investor profiles, I think the one you need is: `country_of_origin`

rachaelcodes (changes requested):
I don't think the filter is working - when I run this with `make dev`, I should only see one project listed if 'France' is selected, but I'm still seeing all 3. The filter does work if I run it in Postman though.

`/v4/search/large-investor-profile` with the body:
```
{
    "country_of_origin": [ID OF FRANCE HERE]
}
```
I think in some places the code refers to `country_investment_originates_from` when that's not a property of investor profiles.

src/apps/investments/client/profiles/FilteredLargeCapitalProfileCollection.jsx
rachaelcodes: Appreciate you might be using a different name in React than gets called in the back end, but I wonder if the confusion is contributing to the filter bug.

peterhudec:
> I don't think the filter is working - when I run this with `make dev`, I should only see one project listed if 'France' is selected, but I'm still seeing all 3. The filter does work if I run it in Postman though.
> 
> `/v4/search/large-investor-profile` with the body:
> 
> ```
> {
>     "country_of_origin": [ID OF FRANCE HERE]
> }
> ```
> 
> I think in some places the code refers to `country_investment_originates_from` when that's not a property of investor profiles.

@rachaelcodes this should work when I fix the wrong filter name

src/apps/investments/client/profiles/FilteredLargeCapitalProfileCollection.jsx
peterhudec: No, it was the way I was sending the param to the API. The body should be JSON, but I was sending it url-encoded.

rachaelcodes (approved):
Looks like it's working nicely :+1:

tomkingston (approved):
Are any of the `fixme`s easy to fix now, or do you think these would be good to do during the upcoming tech debt / maintenance sprint?

peterhudec:
> Are any of the `fixme`s easy to fix now, or do you think these would be good to do during the upcoming tech debt / maintenance sprint?

@tomkingston They are out of scope of our mission at the moment as we are running quite late with the delivery

PR #3301, opened 2021-02-05
Add pagination to the personalised dashboard investment list

paulgain:
## Description of change

Adding pagination to the adviser's Investment list

## Test instructions
1. You will need at least 11 (10 projects per page) investment projects that you have created yourself to test the pagination. Alternatively, you can temporarily hardcode my user ID into the [task](https://github.com/uktrade/data-hub-frontend/blob/126dbd191175124ea9ad410f079b274ef8f7b7cf/src/client/components/MyInvestmentProjects/tasks.js#L8) to fetch my investment projects. For this you will need to bring up the frontend locally and point to the dev API on PaaS. 
2. To view the new dashboard follow these [instructions](https://github.com/uktrade/data-hub-frontend/pull/3248).

### Before

before

### After (page 1)

after

### After (page 2)
page-2

## Checklist


src/client/components/MyInvestmentProjects/tasks.js
```diff
@@ -3,8 +3,9 @@ import { apiProxyAxios } from '../Task/utils'
 export const fetchMyInvestmentsList = ({ limit = 10, page, adviser }) => {
   return apiProxyAxios
     .post('/v3/search/investment_project', {
-      offset: limit * (parseInt(page, 10) - 1) || 0,
+      limit,
+      offset: limit * (page - 1),
       adviser: [adviser.id],
     })
-    .then(({ data }) => data.results)
+    .then(({ data }) => data)
```
ian-leggett, line 10: We should probably handle an error here if the post fails

src/client/components/MyInvestmentProjects/index.jsx
```diff
@@ -24,16 +35,36 @@ const MyInvestmentProjects = ({ page = 1, adviser, items = [] }) => (
...
+            onPageClick={onPageClick}
+          />
+        </>
+      )}
     </Task.Status>
   </article>
 )
 
 MyInvestmentProjects.propTypes = {
+  results: PropTypes.array.isRequired,
```
ian-leggett, line 53: Might be worth checking the shape of the objects in the array. As for results this will always be an array as its default is an empty array in the store. I would remove the `isRequired` on the array but type check the objects inside the array.

ian-leggett (commented):
Really good, just a couple of comments 👍

tomkingston (approved):

src/client/components/MyInvestmentProjects/tasks.js
paulgain: Error handling is built in to the Task :-)
error-handling

**Once we have some designs on what we should present to the user we can override this behaviour (i think).**

ian-leggett: > Error handling is built in to the Task :-)
> error-handling
> 
> **Once we have some designs on what we should present to the user we can override this behaviour (i think).**

Ah yes! forgot about this :-)

src/client/components/MyInvestmentProjects/index.jsx
paulgain: "_Might be worth checking the shape of the objects in the array_." 

We can do either of the following:
1. Map over the response array and create a smaller object based on the fields the list item requires and then define the shape at that level.
2. Leave it as is as the lower-level components have defined the exact props they require: https://github.com/uktrade/data-hub-frontend/blob/feature/dashboard-investments-pagination/src/client/components/MyInvestmentProjects/InvestmentListItem.jsx#L18

What do you think?

src/client/components/MyInvestmentProjects/tasks.js
ian-leggett: I think tailoring an error message for this is out of scope for this PR, lets add it as an iteration on the Task component 👍

src/client/components/MyInvestmentProjects/index.jsx
ian-leggett: After our discussion it makes sense to follow number 2. At the highest level on a connected component we shouldn't be too granular around type checking, its when we get into the child components we should be more granular as we will end up writing a lot of proptypes along with duplication of protypes as we go down the tree.

ian-leggett (approved):
All good 👍

PR #3314, opened 2021-02-08
Feature/hook up dash chart to api

tomkingston:
## Description of change

Hooks up the investment summary chart from the new dashboard page to the new API endpoint for investment summaries.

I have also refactored some of the chart component following the Single Responsibility Principal to make it easier to understand. 

This has meant splitting the chart into:

- A `BarChart` component that is purely responsible for rendering a chart
- A `DataTable` component that renders the data in a table
- A `DataSummary` which allows users to toggle between the `DataTable` and `BarChart` views for the data
- A `DataSummaryPicker` which allows the user to switch between data sets with the select drop down

Data Summary Hierarchy

## Test instructions

With the layoutTesting flag set (or just temporarily edit 'src/apps/dashboard/views/dashboard.njk' so that it points to the new dashboad), visit the dashboard page and take a look at the chart. 

Choose between the two different years in the drop down. The chart should look the same as it did previously, but with data loaded from the API instead.

I have also made some amendments around storybook, so also check that the chart is showing correctly there.

There are now a storybooks for:

- `Chart/BarChart`
- `DataTable`
- `DataSummaryPicker` (this is taken from the former `Chart` stories)

## Screenshots

Pretty much the same as before, but with the data coming from the API:

Screenshot from 2021-02-08 20-11-55

## Checklist


paulgain:
I've added some comments and thoughts on how this could be improved. 
It would be good to get both of your thoughts @tomkingston @ian-leggett  
```
const InvestmentProjectSummary = ({ adviser, investmentSummary = {} }) => {
  // Both of the following lines should be moved to task.js, we can map over the 
  // results and create new objects that can plug straight into the components
  // below. It’s not the responsibility of the component to transform the data
  // from the API. I feel this component’s arguments should only be the adviser
  // and the data ranges.

  // Move both lines to task.js
  const annualSummaries = get(investmentSummary, 'annual_summaries', [])
  const dataRanges = annualSummariesAsDataRanges(annualSummaries)

  return (
    <Task.Status
      name={TASK_GET_INVESTMENT_SUMMARY}
      id={ID}
      progressMessage="Loading your investment projects"
      startOnRender={{
        payload: { adviser },
        onSuccessDispatch: INVESTMENT_SUMMARY__LOADED,
      }}
    >
      // I think we need to keep things simple and on a single level - 
      // we're currently going three levels deep.
      // We need to apply logic here to only show a single chart/table but
      // you get the idea.
      {() => (
	<>
           <SummaryHeader>My Projects Summary<SummaryHeader/>
           <DateRangeSelect dataRanges={dataRanges} />
           <ProjectTotals>7 projects</ProjectTotals>
           <Button>Toggle Chart</Button>
           <BarChart dataRanges={dataRanges} />
           <PieChart dataRanges={dataRanges} />
           <AnyOtherChart dataRanges={dataRanges} />
           <Table dataRanges={dataRanges} />
	</>
      )}
    </Task.Status>
  )
}
```

tomkingston:
Hi Paul - thanks for the feedback. I have reworked the components to make things a bit simpler - here are my thoughts on your suggestions:

# Data Transformation

I see the responsibility of the `InvestmentProjectSummary` component as being to prepare the data for the chart component. So, if we ended up using another library for charting that required a different data structure, we would change how the data is transformed here. If the data transformation changed in the task, it would subsequently break any other components that depended on that data.

The `dataRanges` are actually quite a specific format that is useful for the chart but less useful for accessing specific bits of information. So for example, if we wanted to display the number of assigned projects in the most recent financial year, this could be accessed as follows:

```
// See below for example data formats of annualSummaries / dataRanges 
// With the unformatted annual summaries:
annualSummaries[0].totals.assign_pm.value

// With the annual summaries formatted as data ranges:
dataRanges[0].range.find((item) => item.id === 'c9864359-fb1a-4646-a4c1-97d10189fc03')
```

The first version is not only shorter and easier to understand, but also doesn't require much computation.

# Chart Component Composition

I have restructured and renamed some of the components to make it a bit clearer what they are actually doing now - I have also updated the PR description to explain the composition of the components. 

I don't see a big issue in having components that use other components that use other components - this is an important element of clean code composition and functional programming - in fact it is the recommended approach in the [React Docs](https://reactjs.org/docs/composition-vs-inheritance.html). Putting everything into one component would break [SOLID](https://en.wikipedia.org/wiki/SOLID) design principles, particularly the Single-Responsibility Principle.

These articles also explain the benefits of this sort of component composition in react: 
https://blog.usejournal.com/how-to-apply-solid-principles-in-react-applications-6c964091a982
https://medium.com/@roni.shabo/single-responsibility-in-reactjs-9c60e4163862

Hope that all makes sense!

```javascript
investmentSummary = {
	annual_summaries: [
		{
		    financial_year: {
			label: '2020-21',
			start: '2020-04-01',
			end: '2021-03-31'
		    },
		    totals: {
			...,
			assign_pm: {
			  label: 'Assigned',
			  id: 'c9864359-fb1a-4646-a4c1-97d10189fc03',
			  value: 3
			},
			...             
		    }
		},
		...
	],
	adviser_id: 'e83a608e-84a4-11e6-ae22-56b6b6499622'
}

dataRanges = [
	{
		label: 'Current financial year',
		name: '2020-04-01',
		range: [
			...,
			{
				label: 'assigned',
				id: 'c9864359-fb1a-4646-a4c1-97d10189fc03',
				value: 3,
			},
			...
		]
	},
	...
]
```

ian-leggett (commented):
I would lift all references to tasks and state management out of the clients/components folder, these should be left stateless(if we can). Can we follow the same pattern/structure as the collection list we recently built. I think all the tasks, state.js and reducer should live higher up like they do in `ProjectsCollection.jsx`. I think all this stuff should go into `src/apps/dashboard/client`

ian-leggett (approved):
After our conversation we agree that the all tasks and state management files should live as high up as possible and outside of the `client/components` folder. As long as @paulgain is happy that his comments have been address I am happy to accept this PR and we can refactor the structure afterwards in a separate PR.

paulgain:
@tomkingston do you mind holding off until @peterhudec has had a chance to review this PR. I feel we have a number of patterns and practices already in place, adding more to the mix is not always helpful. I'm not against what you're proposing perse but we need consistency throughout otherwise things will get confusing. My main concern is translating the data within the component, we don't do that in the old tech-stack or the new. Thanks.

src/client/components/DataSummaryPicker/DataSummary.jsx
```diff
@@ -0,0 +1,92 @@
...
+
+const StyledButton = styled(Button)`
+  margin-bottom: ${SPACING.SCALE_3};
+`
+
+const DataSummary = ({
+  title,
+  description,
+  subject,
+  accessible = false,
```
peterhudec, line 46: No need to fall back to `false` as `undefined` is just as good as `false`.

src/client/components/DataSummaryPicker/DataSummary.jsx
```diff
@@ -0,0 +1,92 @@
...
+  headers,
+  data = [],
+  children,
+}) => {
+  const [isAccessible, setAccessible] = useState(accessible)
```
peterhudec, line 51: We should use Redux instead of the internal component state.

src/client/components/DataSummaryPicker/DataSummary.jsx
```diff
@@ -0,0 +1,92 @@
...
+
+  const total = data.reduce((prev, curr) => prev + curr.value, 0)
+
+  return (
+    <>
+      <StyledHeader>{title}</StyledHeader>
+      {children}
+      <TotalHeader>
+        <span>{total}</span> <span>{pluralize(subject, total)}</span>
```
peterhudec, line 60: As I mentioned in the comment about the cascading selectors, it might seem that the spans here can just be removed, but they are actually styled. The problem is that they are styled by the `TotalHeader` that's wrapping them. If we agreed on this pattern, we could never be sure that an element is not relied on to be present by some styling done in any of the components up the hierarchy.

I would suggest to make each span a separately styled component.

src/client/components/DataSummaryPicker/index.jsx
```diff
@@ -0,0 +1,70 @@
...
+
+const StyledSelect = styled(Select)`
+  font-size: ${BODY_SIZES.XSMALL}px;
+  margin-bottom: ${SPACING.SCALE_5};
+  span {
+    font-size: ${BODY_SIZES.SMALL}px;
+  }
+  select {
+    font-size: ${BODY_SIZES.SMALL}px;
+  }
```
peterhudec, line 18: Cascading selectors

src/client/components/DataSummaryPicker/index.jsx
```diff
@@ -0,0 +1,70 @@
...
+`
+
+const DataSummaryPicker = ({ dataRanges, ...props }) => {
+  let defaultRangeName = null
+  if (dataRanges.length) {
+    defaultRangeName = dataRanges[0].name
+  }
```
peterhudec, line 25: Unnecessarily imperative, it could just be:
```js
const defaultRangeName = dataRanges[0]?.name
```

src/client/components/DataSummaryPicker/index.jsx
```diff
@@ -0,0 +1,70 @@
...
+  const [selectedRangeName, setSelectedRangeName] = useState(defaultRangeName)
```
peterhudec, line 26: Use redux instead of internal state.

src/client/components/DataSummaryPicker/index.jsx
```diff
@@ -0,0 +1,70 @@
...
+  const [selectedDataRange] = dataRanges.filter(
+    ({ name }) => name === selectedRangeName
+  )
+  props.data = get(selectedDataRange, 'range', [])
```
peterhudec, line 30: We should not be modifying props, they should be considered immutable.

src/client/components/DataSummaryPicker/index.jsx
```diff
@@ -0,0 +1,70 @@
...
+
+  return (
+    <DataSummary {...props}>
+      <StyledSelect
+        name="sortBy"
+        label="Date range"
+        input={{ onChange: (e) => setSelectedRangeName(e.target.value) }}
+      >
+        {dataRanges.map(({ label, name }, i) => (
+          <option value={name} key={i}>
```
peterhudec, line 40: Use `label` or `name` or they concatenation as key instead of the index.

src/client/components/InvestmentProjectSummary/index.jsx
```diff
@@ -1,98 +1,62 @@
...
-  const [projectsData, setProjectsData] = useState(data[0])
-
-  const onChange = (e) => {
-    const [selectedProjects] = data.filter(
-      ({ date }) => date === e.target.value
-    )
-    setProjectsData(selectedProjects)
-  }
+const InvestmentProjectSummary = ({ adviser, investmentSummary = {} }) => {
+  const dataRanges = investmentSummaryAsDataRanges(investmentSummary)
```
peterhudec, line 12: I agree with @paulgain here. Forwarding the whole API result is easy, but makes this component very tightly coupled to the API. A function should not make any assumptions about the outside world, it should only ask for information that it needs in their simplest possible form.

src/client/components/InvestmentProjectSummary/reducer.js
```diff
@@ -0,0 +1,13 @@
+import { INVESTMENT_SUMMARY__LOADED } from '../../actions'
+
+export default (state = {}, { type, result }) => {
+  switch (type) {
+    case INVESTMENT_SUMMARY__LOADED:
+      return {
+        ...state,
+        investmentSummary: { ...result },
```
peterhudec, line 8: No need for any of the spreads as the `investmentSummary` is the state's only property. It can just be:
```js
return {investmentSummary: result}
```

src/client/components/InvestmentProjectSummary/state.js
```diff
@@ -0,0 +1,3 @@
+export const TASK_GET_INVESTMENT_SUMMARY = 'TASK_GET_INVESTMENT_SUMMARY'
+export const ID = 'investmentSummary'
+export const state2props = ({ ...state }) => ({ ...state[ID] })
```
peterhudec, line 3: No need for the spreads. It can just be:
```js
export const state2props = (state) => state[ID]
```

src/client/components/InvestmentProjectSummary/tasks.js
```diff
@@ -0,0 +1,7 @@
+import { apiProxyAxios } from '../Task/utils'
+
+export const fetchInvestmentSummary = ({ adviser }) => {
+  return apiProxyAxios
+    .get(`/v4/adviser/${adviser.id}/investment-summary`)
+    .then(({ data }) => data)
```
peterhudec, line 6: By design, the _tasks_ are the place to translate data from the API (or wherever it comes from) to whatever the component that is using the task needs. It's the component that should dictate what the task should return, not the API. If we just forward the data like this, it makes the frontend tightly coupled to the API.

The `investmentSummaryAsDataRanges` should thus be used here.

src/client/components/InvestmentProjectSummary/utils.js
```diff
@@ -0,0 +1,26 @@
...
+export const investmentSummaryAsDataRanges = (investmentSummary) => {
+  const annualSummaries = get(investmentSummary, 'annual_summaries', [])
+  const adviserId = get(investmentSummary, 'adviser_id')
+
+  return annualSummaries.map(({ financial_year, totals }) => {
+    const now = new Date()
+    let { label, start, end } = financial_year
+    if (new Date(start) <= now && new Date(end) >= now) {
+      label = 'Current financial year'
+    }
```
peterhudec, line 16: Unnecessarily imperative. I'd prefer a ternary directly on assignment to the property of the returned object.

src/client/components/InvestmentProjectSummary/utils.js
```diff
@@ -0,0 +1,26 @@
...
+/**
+ * Converts investment summary received by the API into data ranges
+ * that the chart component can work with
+ */
+export const investmentSummaryAsDataRanges = (investmentSummary) => {
+  const annualSummaries = get(investmentSummary, 'annual_summaries', [])
+  const adviserId = get(investmentSummary, 'adviser_id')
+
+  return annualSummaries.map(({ financial_year, totals }) => {
+    const now = new Date()
```
peterhudec, line 12: This makes the `investmentSummaryAsDataRanges` function impure, which disqualifies it to be used inside of a React component (we aim for all of them to be pure functions). But it's fine to use it inside of a _task_. Tasks are the place to put all the dirty (input/output) loundry.

src/client/components/InvestmentProjectSummary/utils.js
```diff
@@ -0,0 +1,26 @@
...
+    let { label, start, end } = financial_year
+    if (new Date(start) <= now && new Date(end) >= now) {
+      label = 'Current financial year'
+    }
+    return {
+      label: label,
```
peterhudec, line 18: Use the shorthand assignment

peterhudec (changes requested):
Looks good, but I agree with what @paulgain said about where the data transformation should happen: The _tasks_ are basically the only interface of the application with the outside world (apart from the UI). But the interface should be dictated by the needs of the app, and not be forced by the outside world i.e. the API, even if it might look convenient and practical at the moment. It will turn into a nightmare when the shape of what the API returns changes.

I don't have anything against breaking down the markup into smaller components if it makes sense, but if they are too small and not reused elsewhere I find it better to just have the markup inline. I don't say that's the case here.

I'm more concerned about the usage of the internal React state instead of redux. Back when we decided to go with react, we also decided that we will use Redux exclusively for state management. The decision was based on good reasons after a lenghty discussion. We should perhaps reiterate them every now and then so we understand why we are using Redux in the first place.

The last thing is the usage of deep element type based CSS selectors used in styled components. Again, this might seem convenient at the moment, but it's really hard to follow for anyone who didn't author the code. Nested selectors should only be used to tweak the styling of components which we can't alter in any other way, e.g. third party components. But we should always wrap our own components with `styled` directly if we want to style them.

src/client/components/DataSummaryPicker/index.jsx
tomkingston: name is actually a unique identifier here, so that works well

src/client/components/DataSummaryPicker/DataSummary.jsx
```diff
@@ -0,0 +1,92 @@
...
+  span {
+    font-size: ${HEADING_SIZES.XLARGE}px;
+    display: table-cell;
+    padding: 0 ${SPACING.SCALE_1} 0 0;
+
+    + span {
+      vertical-align: top;
+      font-size: ${BODY_SIZES.SMALL}px;
+      line-height: 16px;
+    }
```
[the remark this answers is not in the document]
ian-leggett: I completely agree however this PR isn't about the styling of any components its about connecting the UI to a new API endpoint. I don't think this needs to be addressed in this PR (I know its confusing as these are new lines of code!). We have new tickets to play for all of this work.

src/client/components/DataSummaryPicker/DataSummary.jsx
ian-leggett: This will be corrected in another PR

paulgain:
@peterhudec thanks, much appreciated 👍

src/client/components/DataSummaryPicker/DataSummary.jsx
tomkingston: Actually, I've chucked in something to fix this, so we won't need to worry about it later

src/client/components/DataSummaryPicker/index.jsx
tomkingston: I didn't actually add this part, but these are used to style a third party component so can't really be changed

peterhudec (approved):
Well done

PR #3330, opened 2021-02-15
Feature/investments profiles filters

peterhudec:
## Description of change

Adds the _Asset class of interest_ and _Company name_ to the `/investments/profiles` page. The filters are hidden behind the `capital-investments-filters` feature flag.

## Test instructions

1. Enable the `capital-investments-filters` feature flag
1. Go to the `/investments/profiles` page
2. There should the two new filters
3. Use the filters
4. The results should reflect the filters


src/client/components/RoutedInputField/index.jsx
```diff
@@ -0,0 +1,115 @@
...
+    reset: () =>
+      dispatch({
+        type: ROUTED_INPUT_FIELD__RESET,
+      }),
+  }),
+  componentStateToProps: (cs, { router }, { qsParam }) => {
+    const qsValue = qs.parse(router.location.search.slice(1))[qsParam]
+    return { ...cs, qsValue }
+  },
+  reducer: (state, { type, value, selectedValue }) => {
```
ian-leggett, line 101: I thought we were lifting this out higher? This component shouldn't know about a reducer.

src/apps/investments/client/profiles/FilteredLargeCapitalProfileCollection.jsx
```diff
@@ -27,11 +32,19 @@ const LargeCapitalProfileCollection = ({
...
+        qsParams[QS_PARAMS.countryOfOrigin],
+        filterOptions.countries
+      )
+      const selectedAssetClassesOfInterest = resolveSelectedOptions(
+        qsParams[QS_PARAMS.assetClassesOfInterest],
+        filterOptions.assetClassesOfInterest
+      )
+
+      const resolveSelectedInvestorCompanyName = () => {
+        const x = qsParams[QS_PARAMS.investorCompanyName]
```
ian-leggett, line 45: Should we rename "x" to something more meaningful?

src/client/components/RoutedInputField/index.jsx
```diff
@@ -0,0 +1,115 @@
...
+  )
+}
+
+RoutedInputField.propTypes = {
+  name: PropTypes.string.isRequired,
+  qsParam: PropTypes.string.isRequired,
+  placeholder: PropTypes.string.isRequired,
+}
+
+export default multiInstance({
```
ian-leggett, line 77: Do we need this as a multinstance component?

ian-leggett (commented):
Looks good, just a couple questions around the multiInstance component

src/client/components/RoutedInputField/index.jsx
peterhudec: Yeah, that's right. But as we need to set the `value` prop of the input when the route changes, it becomes a _managed component_ and we are forced to keep the `value` in the state. But because of this, it is impossible to reset the `value` in the state when the param in the route is removed from outside the component. To make this work, we must also keep around the `selectedValue` in the state. We also have to use the `useEffect` hook 😠

src/client/components/RoutedInputField/index.jsx
peterhudec: Sure we do. Imagine a case when there would be two input filters on a collection page; they would have the same state if they were not _multiinstance_.

PippoRaimondi (approved):
Besides Ian's comments it looks good. I would suggest testing manually the profiles page when rebasing as it might have some conflict with my recent push.

ian-leggett (approved):

PR #3332, opened 2021-02-16
Refactor typeahead component to become AA compliant

ian-leggett:
## Description of change
Our current React Typeahead isn't working well against a screen reader and as a result was flagged up by the DAC (Digital Accessibility Centre) in the recent accessibility audit.

## Solution
To correct the typeahead I looked into some research and development that was done by Gov.uk a while ago. They spent a lot of time testing and developing a typeahead/autocomplete with actual users that rely on assistive technologies to browse the web. As a result they did manage to build a component that worked for the users:

Blog article - https://accessibility.blog.gov.uk/2018/05/15/what-we-learned-from-getting-our-autocomplete-tested-for-accessibility/

Result - https://alphagov.github.io/accessible-autocomplete/examples/

In order to get our typeahead working it was a case of updating mark up and adding the right aria attributes. I believe that the changes made in this PR has really improved how the screen readers works against this component.

Note: I have removed the smallTypeahead as we don't use it, this was something I introduced a while ago which was never consumed into the app.

## How to test

1. Run Storybook and look at the “SINGLE SELECT ONLY” (multi-select will be tackled in a slightly different way)
2. Compare the [GDS version](https://alphagov.github.io/accessible-autocomplete/examples/) to ours, look at the mark up and the aria attributes both static and dynamic. 
3. Most importantly run this with a screenreader, if you have a Mac use VoiceOver


LaurenQurashi (approved):
Looks great to me Ian! Well done 🎉

src/client/components/Typeahead/Typeahead.jsx
```diff
@@ -4,44 +4,107 @@ import AsyncSelect from 'react-select/async'
...
+          <Select {...customisedProps} options={options} />
+          <AssistiveText name={name} />
+        </>
       ) : (
-        <AsyncSelect {...customisedProps} inputId={`field-${name}-1`} />
+        <>
+          <AsyncSelect {...customisedProps} />
+          <AssistiveText name={name} />
+        </>
       )}
```
tomkingston, line 100: I suppose you only need to wrap the Select / AsyncSelect part in the conditional - then you could avoid repeating the AssistiveText Component

src/client/components/Typeahead/Typeahead.jsx
```diff
@@ -4,44 +4,107 @@ import AsyncSelect from 'react-select/async'
...
+  label.toLowerCase().includes(query.toLowerCase())
+
+const Option = ({
+  selectProps: { inputValue },
+  data: { label },
+  ariaProps,
+  ...props
+}) => (
+  <li {...ariaProps}>
+    <comps.Option {...props}>
```
tomkingston, line 17: This is existing, but the `{ components as comps }` feels a bit dirty to me - to avoid the 'components' name conflict you are using  an abbreviated version of the same name ('comps'). It might be a bit more verbose, but you could do something like this to make it clearer what each component is...

```
import Select from 'react-select'
import { Option, NoOptionsMessage, MenuList } from 'react-select/components'

const TypeaheadOption = ({
  selectProps: { inputValue },
  data: { label },
  ariaProps,
  ...props
}) => (
  <li {...ariaProps}>
    <comps.Option {...props}>
...

const Typeahead = ({ options, styles, error, name, ...props }) => {
  const customisedProps = {
    styles: {
      ...(error ? errorStyles : defaultStyles),
      ...styles,
    },
    components: { 
        Option: TypeaheadOption, 
        MenuList: TypeaheadMenuList, 
        Input: TypeaheadInput, 
        NoOptionsMessage: TypeaheadNoOptionsMessage 
    },
...

TypeaheadOption.propTypes = comps.Option.propTypes
...
```

tomkingston (approved):
Looks good, I've added a couple of minor comments, but up to you on what to do with them

src/client/components/Typeahead/Typeahead.jsx
ian-leggett: Much better thanks 👍

src/client/components/Typeahead/Typeahead.jsx
ian-leggett: Agree thanks 👍

PR #3336, opened 2021-02-18
Add banner alerting users to switch from IE to a modern browser

ian-leggett:
## Description of change
We are slowly deprecating the use of IE as soon it won't be available on the cirrus machines (EDGE will replace IE) so we are encouraging our users to use a modern browser for a better experience. To encourage users to stop using IE I have added a banner that will only show if you are on IE, this banner encourages the use of an alternative browser.

## Test instructions
- Open the app up in IE (use Browserstack if needed)
- Every route should have the banner showing at the top.

## Screenshot
Screenshot 2021-02-19 at 10 58 54

## Checklist


src/apps/dashboard/client/IEBanner.jsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react'
+import styled from 'styled-components'
+import { SPACING, FONT_WEIGHTS } from '@govuk-react/constants'
+
+import { StatusMessage } from '../../../client/components'
+
+const StyledStatusMessage = styled(StatusMessage)`
+  margin: ${SPACING.SCALE_4} 0;
+  p {
+    font-weight: ${FONT_WEIGHTS.bold};
```
peterhudec, line 10: This should not be necessary as the `StatusMessage` [already has `font-weight: bold`](https://github.com/uktrade/data-hub-frontend/blob/master/src/client/components/StatusMessage/index.jsx#L10), but maybe the `p` overrides that?

peterhudec (approved):
Looks good, just a small comment.

paulgain:
I've just played with this on BrowserStack and thought the banner looks odd sitting on top of the page, is that the requirement? All of our banners (that I can think of) sit within the page, but there again, this is not related to data within the app like our other banners.

paulgain (approved):

ian-leggett:
> I've just played with this on BrowserStack and thought the banner looks odd sitting on top of the page, is that the requirement? All of our banners (that I can think of) sit within the page, but there again, this is not related to data within the app like our other banners.

I know it does look a bit odd and I agree it is slightly off pattern but I think this is just a one off and it's a temporary thing. Miranda and Sally Torode (content) are happy with it so it's all okay.

paulgain:
One last thing, some users may find this banner annoying as they can't dispel it (it could be shown daily), after all they have until August 21 to move over to a new browser. For some users they will need time to move over (bookmarks etc) and displaying the banner encroaches on the screen real state.

src/apps/dashboard/client/IEBanner.jsx
ian-leggett: Yeah it was the `p` overriding it.

ian-leggett:
> One last thing, some users may find this banner annoying as they can't dispel it (it could be shown daily), after all they have until August '21 to move over to a new browser. For some users they will need time to move over (bookmarks etc) and displaying the banner encroaches on their screen real state.

We don't want them to remove it  or ignore it, I believe we are trying to politely say you have to use a different browser (thats what I understand anyway).

PR #3348, opened 2021-02-22
Added functional tests for oauth flow

tomkingston:
## Description of change

As part of the axios test replacement work, this adds functional tests for oauth happy path. 

It is very difficult / impossible to get the same level of testing as the existing unit tests with functional tests since we cannot mock errors very easily. For this reason, I have not deleted the existing unit tests, but they will need to be replaced when axios replaces requests.

## Checklist

### DIFF
--- a/test/functional/cypress/specs/oauth/oauth-spec.js
+++ b/test/functional/cypress/specs/oauth/oauth-spec.js
@@ -1,0 +1,105 @@
+const urls = require('../../../../../src/lib/urls')
+const qs = require('qs')
+
+/**
+ * Test that the OAuth redirect url redirects to the auth server with given state
+ */
+const testRedirectState = (expectedStatedId) => {
+  cy.request({
+    url: urls.oauth.redirect(),
+    followRedirect: false,
+  }).then((response) => {
+    expect(response.status).to.eq(302)
+    const location = new URL(response.redirectedToUrl)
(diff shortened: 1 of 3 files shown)


test/functional/cypress/specs/oauth/oauth-spec.js
```diff
@@ -0,0 +1,107 @@
...
+        url: urls.oauth.callback(),
+        followRedirect: false,
+      }).then((response) => {
+        expect(response.status).to.eq(302)
+        const location = new URL(response.redirectedToUrl)
+        expect(location.pathname).to.eq('/')
+      })
+
+      // Sign out should clear session cookie
+      cy.getCookie('datahub.sid').should('exist')
```
ian-leggett, line 103: If this clears the should we update the comment to something else?

ian-leggett (commented):
Looks really good to me, just one small comment 👍

test/functional/cypress/specs/oauth/oauth-spec.js
tomkingston: Ahh, I could probably amend the comment to make it a bit clearer. It is checking that the cookie exists before sign out, and then the sign out test function tests that it is cleared once the sign out url is visited.

rachaelcodes (approved):

ian-leggett (approved):

PR #3349, opened 2021-02-22
Fix TabNav keyboard interaction

peterhudec:
## Description of change

Fixes a bug in nested tab `TabNav` components when focus jumped to the inner `TabNav` when using arrow keys to move the focus of the outer `TabNav`.

## Test instructions

1. Go to the homepage `/` or to the [`TabNav` section of the Storybook](https://uktrade.github.io/data-hub-frontend/?path=/story/tabnav--default).
2. Focus the _tablist_ of the outer `TabNav` component
3. Press the right arrow key repeatedly
4. The focus should cycle through all the tabs in the _tablist_ of the outer `TabNav` and should not jump to the inner `TabNav`
5. Move the focus to the inner `TabNav` by hitting the _tab_ key.
6. Press the right arrow key repeatedly
7. The focus should cycle through all the tabs in the _tablist_ of the inner `TabNav` and should not jump to the outer `TabNav`
8. Repeat steps 4 to 7 with the left arrow key.

## Screenshots

There should be no visual change.


ian-leggett (changes requested):
This seems to work well but I need to know where in the DAC report they point out the issues and what it is they were not happy with. Could you point out where the issues are?

src/client/components/TabNav/index.jsx
```diff
@@ -142,13 +145,20 @@ const TabNav = ({
         return (
           <>
             <StyledTablist
-              tabindex={0}
+              tabIndex={-1}
```
ian-leggett, line 148: I think we need to remove this as you cannot access the child elements via tab

peterhudec: No, the `-1` here makes the focus skip the `tablist`, but jump straight to its tabs, which makes the navigation smoother and is exactly how it works in the [MDN example](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Tab_Role) this component is based on.

ian-leggett (approved):

tomkingston (approved):

PR #3361, opened 2021-03-01
Add the investor type filter

peterhudec:
## Description of change

Adds the _Investor type_ filter.

## Test instructions

1. Go to `/investments/profiles`
2. There should be a new _Investor type_ typeahead filter
3. If the filter is clear, it should not affect the search results
4. If the filter has options selected, it should filter the results accordingly


tomkingston (approved):

ian-leggett (approved):
Works well, LGTM

PR #3363, opened 2021-03-02
Feature/check clearance filter

peterhudec:
## Description of change

Adds the _Investor type_ filter.

## Test instructions

1. Go to `/investments/profiles`
2. There should be a new _Check clearance_ typeahead filter
3. If the filter is clear, it should not affect the search results
4. If the filter has options selected, it should filter the results accordingly


tomkingston (approved):

ian-leggett (approved):
My only comment around this is that the placeholder text is the same as the label so it looks a little strange and also the screen reader will sound odd. Looking closely I can see what should be labels are all fieldset legends.

peterhudec:
> My only comment around this is that the placeholder text is the same as the label so it looks a little strange and also the screen reader will sound odd. Looking closely I can see what should be labels are all fieldset legends.

@ian-leggett that's interesting, but does it make sense to have a fieldset with only one field, whose purpose is to group multiple elements?

PR #3364, opened 2021-03-03
Add project status tag to investment projects list

ian-leggett:
## Description of change
We need to add a status tag that indicates where the user is in the process of an Investment project lifecycle.

## Test instructions
- Run the new dashboard as per instructions from #3248
- You should see a status badge against each list item

## Checklist

### DIFF
--- a/src/client/components/MyInvestmentProjects/InvestmentListItem.jsx
+++ b/src/client/components/MyInvestmentProjects/InvestmentListItem.jsx
@@ -4,6 +4,7 @@ import { Details } from 'govuk-react'
 import styled from 'styled-components'
 import { MEDIA_QUERIES, SPACING } from '@govuk-react/constants'
 
+import { Tag } from '../../components'
 import InvestmentEstimatedLandDate from './InvestmentEstimatedLandDate'
 import InvestmentTimeline from './InvestmentTimeline'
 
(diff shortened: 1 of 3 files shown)


PippoRaimondi (approved):

paulgain:
@ian-leggett looks good, it's helpful to add a screenshot as before. Thanks.

src/client/components/MyInvestmentProjects/InvestmentListItem.jsx
```diff
@@ -40,8 +41,15 @@ const InvestmentListItem = ({
 }) => {
   return (
     <li>
+      <Tag
+        colour="grey"
+        data-test="project-status-tag"
+        aria-label="project status"
+      >
+        {stage?.name}
```
paulgain, line 49: Is the `?` really required here? A stage will always have a name.

paulgain (approved):

src/client/components/MyInvestmentProjects/InvestmentListItem.jsx
ian-leggett: To be honest I have just got into the habit of using "optional chaining".

ian-leggett: After looking at this a bit closer we should only use "optional chaining" on optional props so I have reverted this to just use dot notation to access the prop value.

PR #3381, opened 2021-03-08
New Data Hub Search Bar (New Dashboard)

paulgain:
## Description of change
A new React component for Data Hub searches.

## Test instructions
Read this #3248

## Screenshots
Screenshot 2021-03-08 at 16 54 45

## Checklist


rachaelcodes (approved):

src/client/components/PersonalisedDashboard/index.jsx
```diff
@@ -2,42 +2,67 @@ import PropTypes from 'prop-types'
...
   Container,
 } from '../../components'
 
-const PersonalisedDashboard = ({ id, adviser }) => {
-  return (
+const SearchBackground = styled('div')`
+  background-color: ${BLUE};
+`
+
+const SearchContainer = styled('div')`
```
ian-leggett, line 24: We can use the new `Container` component to do this. If we don't reuse the container we will keep repeating these rules.

src/client/components/SearchButton/index.jsx
```diff
@@ -0,0 +1,48 @@
...
+  height: ${({ size }) => `${size}px`};
+  background-color: ${({ backgroundColour }) => backgroundColour};
+`
+
+const SearchButton = ({
+  size = 38,
+  colour = WHITE,
+  backgroundColour = BLACK,
+}) => {
+  return (
```
ian-leggett, line 28: We can remove the return here.

src/client/index.jsx
```diff
@@ -222,7 +222,13 @@ function App() {
         <Dashboard id="homepage" />
       </Mount>
       <Mount selector="#dashboard">
-        {(props) => <PersonalisedDashboard id="dashboard" {...props} />}
+        {(props) => (
+          <PersonalisedDashboard
+            csrfToken={globalProps.csrfToken}
+            id="dashboard"
+            {...props}
```
ian-leggett, line 229: Are we able to remove the spread props here?

ian-leggett (commented):
Looking good, Just a thought, I think we could remove this `dashboard-new` folder in the tests dir or at least combine them. Not sure why I created that dir in the first place.

src/client/components/PersonalisedDashboard/index.jsx
paulgain: I wanted to speak to you about this, I did try reusing the `Container` but I ended up overwriting most of the values. We
can reuse the `Container` but it means passing in the values (`padding` and `margin` for each device breakpoint).
Yours is on the left, mine on the right. Don't forget that the `padding` and `margin` will be very different from page level to component level.

compare

src/client/index.jsx
paulgain: Not at the moment, we need to know who the user is, and that comes via the props. We have a card to make an API (e.g. `Task`) call to `/whoami`, once that's done we can remove the props.

ian-leggett (approved):

PR #3386, opened 2021-03-11
Large capital page framework

PippoRaimondi:
## Description of change

The intent of this PR is to add a dummy page for UK Opportunity so we could use it and add functionality to it.

### After

Screenshot 2021-03-17 at 13 34 54

## Checklist

### DIFF
--- a/src/apps/investments/client/opportunities/Opportunities.jsx
+++ b/src/apps/investments/client/opportunities/Opportunities.jsx
@@ -12,6 +12,8 @@ const StyledSpan = styled('span')`
   background: ${HIGHLIGHT_COLOUR};
 `
 
+// This is dummy data that will be removed once we hook up the API.
+// Data structure will most likely change.
 const detailsRows = [
   { label: 'Opportunity description', value: 'Example Ltd' },
   { label: 'UK location', value: 'Incomplete' },
@@ -26,6 +28,8 @@ const detailsRows = [
   { label: 'Construction risk', value: 'Incomplete' },
 ]
 
+// This is dummy data that will be removed once we hook up the API.
+// Data structure will most likely change.
 const requirementsRows = [
   { label: 'Total investment sought', value: 'Example Ltd' },
   { label: 'Investment secured so far', value: 'Secured Ltd' },
@@ -41,7 +45,7 @@ const incompleteRowsCount = (rows) => {
 const Opportunities = () => (
   <>
     <TabNav
-      id="TabNav"
+      id="Opportunity.tabnav"
       label="Dashboard"
       selectedIndex={'details'}
       tabs={{
@@ -51,7 +55,7 @@ const Opportunities = () => (
             <>
               <ToggleSection
                 label="Opportunity details"
-                id="toggle_details"
+                id="Opportunity.details.toggle"
                 fieldCount={incompleteRowsCount(detailsRows)}
                 showRequiredField={true}
               >
@@ -71,7 +75,7 @@ const Opportunities = () => (
               </ToggleSection>
               <ToggleSection
                 label="Opportunity requirements"
-                id="toggle_requirements"
+                id="Opportunity.requirements.toggle"
                 fieldCount={incompleteRowsCount(requirementsRows)}
                 showRequiredField={true}
               >
@@ -94,7 +98,6 @@ const Opportunities = () => (
         },
       }}
     />
-    &nbsp;
     <ToggleSection
       label="Need to delete this opportunity?"
       id="toggle_delete"
--- a/src/apps/investments/client/opportunities/Opportunities.jsx
+++ b/src/apps/investments/client/opportunities/Opportunities.jsx
@@ -2,33 +2,44 @@ import React from 'react'
 
 import TabNav from '../../../../client/components/TabNav'
 import ToggleSection from '../../../../client/components/ToggleSection'
-import SummaryList from '../../../../client/components/SummaryList'
-import ButtonLink from '../../../../client/components/ButtonLink'
+import SummaryTable from '../../../../client/components/SummaryTable'
 
-const rows = [
-  { label: 'Some Label 1', value: 'Example Ltd' },
-  { label: 'Some Label 2', value: 'Incomplete' },
+import styled from 'styled-components'
+import { HIGHLIGHT_COLOUR } from 'govuk-colours'
+import Link from '@govuk-react/link'
+
+const StyledSpan = styled('span')`
+  background: ${HIGHLIGHT_COLOUR};
+`
+
+const detailsRows = [
+  { label: 'Opportunity description', value: 'Example Ltd' },
+  { label: 'UK location', value: 'Incomplete' },
+  { label: 'Promoters', value: 'Incomplete' },
+  {
+    label: 'Has this opportunity cleared the required checks?',
+    value: 'Incomplete',
+  },
+  { label: 'Lead DIT relationship manager', value: 'Incomplete' },
+  { label: 'Asset classes', value: 'Incomplete' },
+  { label: 'Value', value: 'Incomplete' },
+  { label: 'Construction risk', value: 'Incomplete' },
 ]
 
-class WipButton extends React.Component {
-  onButtonPress = () => {
-    this.setState({
-      isOpen: true,
-    })
-  }
+const requirementsRows = [
+  { label: 'Total investment sought', value: 'Example Ltd' },
+  { label: 'Investment secured so far', value: 'Secured Ltd' },
+  { label: 'Types of investment', value: 'Investment Ltd' },
+  { label: 'Estimated return rate', value: 'Rate Ltd' },
+  { label: 'Timescales', value: 'Timescales Ltd' },
+]
 
-  render() {
-    return (
-      <ButtonLink onClick={this.onButtonPress} float="right">
-        Open All
-      </ButtonLink>
-    )
-  }
+const incompleteRowsCount = (rows) => {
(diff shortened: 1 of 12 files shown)


src/apps/investments/client/opportunities/Opportunities.jsx
```diff
@@ -0,0 +1,111 @@
...
+
+import styled from 'styled-components'
+import { HIGHLIGHT_COLOUR } from 'govuk-colours'
+import Link from '@govuk-react/link'
+
+const StyledSpan = styled('span')`
+  background: ${HIGHLIGHT_COLOUR};
+`
+
+const detailsRows = [
```
rachaelcodes, line 15: Is it worth adding a comment before this to note that it's going to change when the task is added?

src/apps/investments/client/opportunities/Opportunities.jsx
```diff
@@ -0,0 +1,111 @@
...
+                    </SummaryTable.Row>
+                  ))}
+                </SummaryTable>
+              </ToggleSection>
+            </>
+          ),
+        },
+      }}
+    />
+    &nbsp;
```
rachaelcodes, line 97: Elsewhere, I think we generally use `{' '}` in the React files

src/apps/investments/client/opportunities/Opportunities.jsx
```diff
@@ -0,0 +1,111 @@
...
+]
+
+const incompleteRowsCount = (rows) => {
+  return rows.filter(({ value }) => value === 'Incomplete').length
+}
+
+const Opportunities = () => (
+  <>
+    <TabNav
+      id="TabNav"
```
peterhudec, line 44: The `id` should be something which uniquely identifies this particular instance of the `TabNav`, ideally so that it would be obvious where the `TabNav` is used when looking at it's state in Redux dev-tools e.g `"Opportunities"`. Basically, the state of this particular instance of the `TabNav` is stored in the redux store under the ID. 
Two tab navs with the same id would work in sync.
```js
const state = {
  TabNav: {
    'some-id': {focusIndex: 'foo', selectedIndex: 'bar'},
    'some-other-id': {focusIndex: 'ham', selectedIndex: 'spam'},
    // Here it's obvious from the id,
    // that it's a state of the TabNav used in the Opportunities component
    Opportunities: {focusIndex: 'details', selectedIndex: 'details'},
   },
}
```

src/apps/investments/client/opportunities/Opportunities.jsx
```diff
@@ -0,0 +1,111 @@
...
+      label="Dashboard"
+      selectedIndex={'details'}
+      tabs={{
+        details: {
+          label: 'Details',
+          content: (
+            <>
+              <ToggleSection
+                label="Opportunity details"
+                id="toggle_details"
```
peterhudec, line 54: The same as in my comment above applies to the ids here. It should be something like `"Opportunities.details"`.

src/apps/investments/client/opportunities/Opportunities.jsx
```diff
@@ -0,0 +1,111 @@
...
+                  {detailsRows.map(({ label, value }) => (
+                    <SummaryTable.Row key={label} heading={label}>
+                      {value}
+                    </SummaryTable.Row>
+                  ))}
+                </SummaryTable>
+              </ToggleSection>
+              <ToggleSection
+                label="Opportunity requirements"
+                id="toggle_requirements"
```
peterhudec, line 74: Should be `id="Opportunities.requirements"`

src/apps/investments/client/opportunities/Opportunities.jsx
```diff
@@ -0,0 +1,111 @@
...
+              </ToggleSection>
+            </>
+          ),
+        },
+      }}
+    />
+    &nbsp;
+    <ToggleSection
+      label="Need to delete this opportunity?"
+      id="toggle_delete"
```
peterhudec, line 100: Should be `id="Opportunity.delete"`

src/apps/investments/client/opportunities/Opportunities.jsx
peterhudec, line 97: Not sure why `&nbsp;` is needed here. The `TabNav` and `ToggleSections` should be `display: block` elements. If it is for the spacing, we should use CSS.

src/apps/investments/controllers/opportunities.js
```diff
@@ -0,0 +1,13 @@
+const renderOpportunitiesView = async (req, res, next) => {
```
peterhudec, line 1: It probably doesn't need to be `async`, although it doesn't do any harm.

src/client/components/ButtonLink/index.jsx
```diff
@@ -12,6 +12,7 @@ const ButtonLink = styled(Button).attrs((props) => props)`
     color: ${LINK_COLOUR};
     cursor: pointer;
     text-decoration: underline;
+    ${(props) => props.float && `float: ${props.float};`}
```
peterhudec, line 15: We should not bake case specific logic int to this kind of generic components. A better approach is to re-style the `ButtonLink` in the component where the special case is needed:

```js
const FloatingButtonLink = styled(ButtonLink)({
  folat: 'right',
})
```

src/client/components/ToggleSection/index.jsx
```diff
@@ -48,11 +48,44 @@ const StyledButton = styled('button')`
...
+}
+
+const ToggleSection = ({
+  label,
+  open,
+  isOpen = false,
+  children,
+  showRequiredField = false,
+  fieldCount = 0,
+  isStatic = false,
```
peterhudec, line 77: These props are too coupled to the specific case of the `Opportunities` component. The `ToggleSection` is a pretty generic component used in at least two other places where there's no notion of a _field_. I would instead just add a single `subtitle` or `sublabel` prop which would accept any content and render it on the right side of the heading.

Also it is not clear what _static_ means here. It only seems to affect the font size. Instead of introducing this property, I would restyle the `ToggleSection` where the decreased font size is needed.

src/client/components/ToggleSection/index.jsx
```diff
@@ -48,11 +48,44 @@ const StyledButton = styled('button')`
...
+`
+
+const renderRequiredFields = (requiredFieldFlag, fieldCount) => {
+  if (requiredFieldFlag) {
+    if (fieldCount == 0) {
+      return <StyledLabel color={GREEN}>Completed</StyledLabel>
+    }
+    return <StyledLabel color={RED}>{fieldCount} fields required</StyledLabel>
+  }
+}
```
peterhudec, line 68: Simila as in the `ButtonLink` above, this is too specific to be in a generic component like this. I would move the whole thing out of here and pass it as a `subtitle` prop as explained in my comment below.

src/client/components/ToggleSection/index.jsx
```diff
@@ -48,11 +48,44 @@ const StyledButton = styled('button')`
...
   <ToggleContainer {...props}>
-    <StyledButton onClick={() => open(!isOpen)} isOpen={isOpen}>
+    <StyledButton
+      onClick={() => open(!isOpen)}
+      isOpen={isOpen}
+      isStatic={isStatic}
+    >
       <span>{label}</span>
     </StyledButton>
+    {renderRequiredFields(showRequiredField, fieldCount)}
```
peterhudec, line 88: It's slightly more performant (although negligible here) to shape the signature of functions like this as components and delegate calling of them to React by using them as components. e.g. `<RequiredFields showRequiredField={showRequiredField} fieldCount={fieldCount}/>`. That way, the function will only be called if one of the values changes.

peterhudec (changes requested):
Looks good, but I would not bake in the `Opportunity` specific functionality and terminology into the generic `ButtonLink` and `ToggleSection` components. Instead I would restyle them locally for the specific use cases.

src/apps/investments/client/opportunities/Opportunities.jsx
PippoRaimondi: good one, will do!

src/apps/investments/controllers/opportunities.js
PippoRaimondi: might as well remove so it doesn't mislead a possible debug on this feature.

rachaelcodes (approved):

src/apps/investments/client/opportunities/Opportunities.jsx
```diff
@@ -0,0 +1,140 @@
...
+                  }
+                >
+                  {detailsRows.map(({ label, value }) => (
+                    <SummaryTable.Row key={label} heading={label}>
+                      {value}
+                    </SummaryTable.Row>
+                  ))}
+                </SummaryTable>
+              </ToggleSection>
+              {renderRequiredFields(incompleteRowsCount(requirementsRows))}
```
peterhudec, line 101: It's better from the performance perspective to have the `renderRequiredFields` as a React component and not just a function call. That way it would only be called when `requirementsRows` changes.

```jsx

</ToggleSection>
{/* Will only be called if  requirementsRows changes */}
<RequiredFields rows={requirementsRows}/>
{/* Will be called on every render */}
{renderRequiredFields(incompleteRowsCount(requirementsRows))}
<ToggleSection
  variant={VARIANTS.SECONDARY}
  label="Opportunity requirements"
  id="opportunity_requirements_toggle"
>
```

peterhudec (approved):
Looks good, just one little suggestion about optimizing the `renderRequiredFields` for React performance.

PR #3403, opened 2021-03-15
New TabNav styles for the personalised dashboard

paulgain:
## Description of change

Applied new styles to the `TabNav` component inline with the new personalised dashboard design.

```js
// blue-theme.js
export default {
  tabNav: {
    selectedButton: {
      color: WHITE,
      ...
  },
  ...
}
```

```js
// PersonalisedDashboard.jsx
import blueTheme from './blue-theme'

<ThemeProvider theme={blueTheme}>
 ...
<ThemeProvider>

```

```js
// TabNav.jsx (or any styled component)
color: ({ theme }) => get(theme, 'tabNav.selectedButton.color', DEFAULT_COLOUR)
```

If there's a future requirement to add theming we can simply define additional files like `blue-theme.js` and then swap out the files based on user actions.

## Test instructions

Read this #3248

### Before

Screenshot 2021-03-15 at 09 34 12

### After

Screenshot 2021-03-15 at 09 29 26

## Checklist


src/client/components/MyInvestmentProjects/index.jsx
```diff
@@ -37,13 +37,15 @@ const MyInvestmentProjects = ({
...
-      </InvestmentListShowDetails>
+      {false && (
+        <InvestmentListShowDetails
+          onChange={(event) => onShowDetailsChange(event.target.checked)}
+          checked={showDetails}
+          disabled={!results.length}
+        >
+          Show details
+        </InvestmentListShowDetails>
+      )}
```
paulgain, line 48: Temporarily hiding this.

src/client/components/PersonalisedDashboard/theme.js
```diff
@@ -0,0 +1,19 @@
+import { BLUE, WHITE } from 'govuk-colours'
+
+export const DASHBOARD_THEME = {
+  dashboard: {
```
tomkingston, line 4: Having a theme named `DASHBOARD_THEME` and then `dashboard` within is a little confusing since you would also have to define dashboard in other themes (as below). 
```
DEFAULT_THEME = {
   dashboard: {
    tabNav: {
      ...
   }
}
```
Presumably, this new theme will be applied to all pages eventually, so perhaps this should be given a more generic name that isn't connected to a particular component (not quite sure what, `BLUE_THEME`???) and pushed up to a higher level.

tomkingston (commented):
This looks good and seems to work well, there are a few things that could possibly be looked at to make things better - let me know what you think?

- It would probably be good to include the new theme as another story in storybook, just to make it a bit easier to see what it will look like - I suppose this should probably need to be done wherever we have a different style applied for the new theme.

- Now that the tabs are wrapped in a blue border, the grey bottom border on the last list item doesn't look quite right when it touches the bottom - does this need to be removed with a last-child selector, or will the pagination take care of this? 
Screenshot from 2021-03-15 09-49-29

src/client/components/MyInvestmentProjects/index.jsx
tomkingston: Do you think passing a prop through to control the visibility might be nicer? At the moment this feels very similar to just commenting out the code.

paulgain: Whether I pass in a boolean or define a boolean in situ the end result is just the same and like the comment says it's just temporary.

src/client/components/PersonalisedDashboard/theme.js
paulgain: I'm thinking about this the other end when you reference these values within the styled component as that's where most of our time will be spent and `DEFAULT_THEME ` or `ANY_OTHER_THEME` doesn't even come in to it.

I could do this:


```js
// blue-theme.js
export default {
  blue: {
    tabNav: {
       ...
      },
    },
  },
}
```

```js
// Styled component
color: ({ theme }) => get(theme, 'blue.tabNav...', TEXT_COLOUR),
```

What do you think?

paulgain: _"It would probably be good to include the new theme as another story in storybook, just to make it a bit easier to see what it will look like - I suppose this should probably need to be done wherever we have a different style applied for the new theme."_

Personally, I think this will consume too much of our time, we don't want to be storybooking every style change.

paulgain: _"Now that the tabs are wrapped in a blue border, the grey bottom border on the last list item doesn't look quite right when it touches the bottom - does this need to be removed with a last-child selector, or will the pagination take care of this?"_

Good spot 👍

src/client/components/PersonalisedDashboard/blue-theme.js
```diff
@@ -0,0 +1,19 @@
+import { BLUE, WHITE } from 'govuk-colours'
+
+export default {
+  blue: {
```
tomkingston, line 4: From what I can tell, the `blue` namespacing is redundant since this is already the blue theme (or you could change line 4 back to `dashboard`) - I think the keys needs to be duck-typed so that it is possible to simply switch themes in and out. This should make some of the checks in the styled components a bit simpler (see below).

You could add `name` as one of the keys of the theme so that you could apply specific logic to the particular theme (they have done something similar [here](https://css-tricks.com/theming-and-theme-switching-with-react-and-styled-components/)).

src/client/components/TabNav/index.jsx
```diff
@@ -75,18 +76,32 @@ const StyledButton = styled(FocusableButton)({
...
     textDecoration: 'none',
-    border: `1px solid ${BORDER_COLOUR}`,
     borderBottom: 'none',
     // The negative margin is here so that it overcasts tabpane's top border
     marginBottom: -1,
     // The 19px include compensation for the 1px negative margin above
     padding: '14px 19px 16px',
-    background: WHITE,
+    color: ({ theme }) =>
+      get(theme, 'blue.tabNav.selectedButton.color', TEXT_COLOUR),
```
tomkingston, line 92: As mentioned above, the `blue` namespace means that this theme cannot be swapped out for another theme (I know we won;t actually be doing that though) - if it was namespaced with `dashboard` (as you had originally), or that top-level namespacing was just removed, it should be possible to swap themes in and out - ie, one of:
```
get(theme, 'tabNav.selectedButton.color', TEXT_COLOUR)
//...or...:
get(theme, 'dashboard.tabNav.selectedButton.color', TEXT_COLOUR)
```

src/client/components/TabNav/index.jsx
```diff
@@ -100,12 +115,21 @@ const StyledTablist = styled.div({
...
 
-const StyledTabpanel = styled.div({
-  marginTop: 15,
-  [MEDIA_QUERIES.TABLET]: {
-    marginTop: 30,
-  },
-})
+const StyledTabpanel = styled('div')`
+  ${({ theme }) =>
+    theme.blue
```
tomkingston, line 120: Ideally you wouldn't need a conditional here and could just apply the styles that are provided in the theme (falling back to defaults). 
Failing that, the following conditional would work with a duck-typed theme `theme.name === 'blue'`.

tomkingston (approved):
As you mentioned, I think it would be good to call the namespaced part 'dashboard' rather than just having it named after the colour 'blue' (or you could just remove that from the namespacing altogether), but I like this being contained within the 'blue-theme.js'.

I'm currently planning on working with this for the new `ToggleSections` too, so looking forward to seeing it in action!

src/client/components/PersonalisedDashboard/theme.js
tomkingston: On the storybook, I wasn't meaning to document every style change, just to show what the two different themes look like (default and blue) so you could quickly see if anything is broken in one theme, but as this is only temporary I think this will save us some time

paulgain: Yep, the scope of this work was to style the TabNav inline with the design, not to provide a global mechanism for swapping out one set of styles for another (AKA theming).

src/client/components/PersonalisedDashboard/blue-theme.js
paulgain: There's no requirement to switch themes in and out.

src/client/components/TabNav/index.jsx
paulgain: During the implementation I did both and settled on the latter because when a theme is not applied (the default) the theme object on the props object is empty, conversely, when a theme is applied you need an object to conditionally check as not all styles are simply overridden, some styles and completely swapped out for another set. 

Here's an example from the PR:

```js
const StyledTabpanel = styled('div')`
  ${({ theme }) =>
    theme.blue
      ? `
        border: ${get(theme.blue, 'tabNav.tabPanel.border')};
        padding: ${get(theme.blue, 'tabNav.tabPanel.padding')};
        margin-bottom: ${get(theme.blue, 'tabNav.tabPanel.marginBottom')};
      `
      : `
        margin-top: 15px;
        ${MEDIA_QUERIES.TABLET} {
          margin-top: 30px;
        }
      `}
`
```

src/client/components/TabNav/index.jsx
paulgain: That's what I've done elsewhere, this is a very different scenario though as both sets of styles are completely different.
The problem with `theme.name === 'blue' or theme.name === 'red' or theme.name === 'green'` is that you pepper your code with unwanted conditionals - the correct approach is to be style-agnostic. Not sure what that would look like when swapping like I've done here.

src/client/components/TabNav/index.jsx
paulgain: I've settled on this: `get(theme, 'tabNav.selectedButton.color', TEXT_COLOUR)`

paulgain:
In summarising we have this:


```js
// blue-theme.js
export default {
  tabNav: {
    selectedButton: {
      color: WHITE,
      ...
  },
  ...
}
```

```js
// PersonalisedDashboard.jsx
import blueTheme from './blue-theme'

<ThemeProvider theme={blueTheme}>
 ...
<ThemeProvider>

```

```js
// TabNav.jsx (or styled component)
color: ({ theme }) => get(theme, 'tabNav.selectedButton.color', TEXT_COLOUR)
```

ian-leggett (approved):

ian-leggett (commented):
In the future I really don't think we will need theming as this is more for white label sites or theme toggles on apps however this does give us what we want for now without duplicating components. Once we know more about the design and how to apply it throughout the site I think we will remove this theme provider.

paulgain:
> In the future I really don't think we will need theming as this is more for white label sites or theme toggles on apps however this does give us what we want for now without duplicating components. Once we know more about the design and how to apply it throughout the site I think we will remove this theme provider.

Agreed, it's an elegant temporary solution that solves the problem nicely.

PR #3407, opened 2021-03-16
Feature/dashboard toggle

tomkingston:
## Description of change

Adds a new toggle component for use on the new personalised dashboard page. There was an existing `ToggleSection` component, but it was not styled as needed for the new dashboard. The old style can now be chosen by specifying the appearance prop as "secondary", with the new (default) style as "primary". This allows us to use the primary style for the dashboard and use the more subtle secondary component on things like the investment projects list filters.

## Test instructions

Go to the new personalised dashboard page by toggling the feature flag. The Reminders component and the Investment Project Summary components should appear inside togglable sections. Note that the Reminders component has a red top border (to indicate high significance) above it, with the Project Summary showing a blue top border.

`ToggleSection` can also be seen on storybook with the primary and secondary appearances.

Please check that the new component is consistent with the [designs](https://glc39r.axshare.com/#id=ozm45m&p=spec_reminders&g=1) and toggles open and closed.

### Before

Screenshot from 2021-03-16 16-45-44

### After

Screenshot from 2021-03-16 16-34-49

## Checklist


src/client/components/PersonalisedDashboard/index.jsx
```diff
@@ -66,12 +68,34 @@ const PersonalisedDashboard = ({
...
+                  label="Reminders"
+                  id="investment-reminders-section"
+                  badge={
+                    <NotificationBadge
+                      label={`${outstandingPropositions.count}`}
+                    />
+                  }
+                  major={true}
+                  isOpen={true}
+                  appearance="primary"
```
paulgain, line 81: The string `primary` should be in a constant. See below for more details

src/client/components/PersonalisedDashboard/index.jsx
```diff
@@ -66,12 +68,34 @@ const PersonalisedDashboard = ({
...
+                </ToggleSection>
               )}
             </Task.Status>
-            <InvestmentProjectSummary adviser={adviser} />
+
+            <ToggleSection
+              label="Investment project summary"
+              id="investment-project-summary-section"
+              isOpen={true}
+              appearance="primary"
```
paulgain, line 95: Same applies here.

paulgain (approved):

paulgain (commented):
LGTM!

paulgain:
Your loading indicators are inconsistent. One is outside the toggle component, the other is inside. The Task needs to be inside for consistency. 
loading-indicators

src/client/components/PersonalisedDashboard/index.jsx
```diff
@@ -56,22 +60,46 @@ const PersonalisedDashboard = ({
...
-                payload: { adviser },
-                onSuccessDispatch: OUTSTANDING_PROPOSITIONS__LOADED,
-              }}
+            <ToggleSection
+              label="Reminders"
+              id="investment-reminders-section"
+              badge={
+                outstandingPropositions.loaded && (
+                  <NotificationBadge
+                    label={`${outstandingPropositions.count}`}
```
tomkingston, line 69: Interested on whether you think having a loaded prop is the way to go, or would it be better for the initial count value to be null and deal with it that way?

paulgain: What's the purpose of the `loaded` flag or put another way why can't `results`  and `count` be relied upon?

tomkingston: So the reason for having it is otherwise we end up showing 0 as the count for outstandingPropositions until the api has loaded (the heading is not inside the task, so is not hidden behind the loading gif). So the other option would be to set count as null or undefined as the initial value in redux - then just not show the badge when it is null or undefined.

tomkingston: Actually, I'm thinking if the count is 0 then there's no point of showing the badge anyway, update coming up....

src/client/components/PersonalisedDashboard/index.jsx
```diff
@@ -56,22 +60,46 @@ const PersonalisedDashboard = ({
...
+            <ToggleSection
+              label="Reminders"
+              id="investment-reminders-section"
+              badge={
+                !!outstandingPropositions.count && (
+                  <NotificationBadge
+                    label={`${outstandingPropositions.count}`}
+                  />
+                )
+              }
```
paulgain, line 72: No need for a double not `!!`. In the reducer `count` is initialised to 0 AKA falsy.

tomkingston: Ahh, yes, I did do that initially, but when it is used with the &&, it means that the 0 is actually rendered (when we want nothing to show)

paulgain: Ah ha, I've previously come across that as well.

src/client/components/ToggleSection/__stories__/ToggleSection.stories.jsx
```diff
@@ -0,0 +1,141 @@
+import React from 'react'
+import { storiesOf } from '@storybook/react'
+
+import { APPEARANCES } from '../../../../common/constants'
```
ian-leggett, line 4: Personally I would call this "Variants" and I would pass something like `<button variant="secondary">` it doesn't really matter as long as we reuse this idea of switching between primary and secondary using the "appearance" prop. I only like variant because I have used Material UI too much!

ian-leggett (approved):

src/client/components/ToggleSection/__stories__/ToggleSection.stories.jsx
tomkingston: Sounds good - I will make the change, I didn't really know what to call it

PR #3412, opened 2021-03-18
Lift up state for both MyInvestmentProject and InvestmentProjectSummary

paulgain:
## Description of change

Pass the props down from the `PersonalisedDashboard` to `MyInvestmentProjects` ensuring consistency. 

## Test instructions
Read this #3248

## Checklist

### DIFF
--- a/src/client/components/DashboardTabs/index.jsx
+++ b/src/client/components/DashboardTabs/index.jsx
@@ -15,32 +15,30 @@ const DashboardTabs = ({
   id,
   investmentProjects,
   myInvestmentProjectTaskProps,
-}) => {
-  return (
-    <StyledDiv data-cy="dashboard-tabs">
-      <TabNav
-        id={`${id}.TabNav`}
-        label="Dashboard"
-        routed={true}
-        tabs={{
-          [urls.personalisedDashboard.myInvestmentProjects()]: {
-            label: 'My projects',
-            content: (
-              <MyInvestmentProjects
-                myInvestmentProjectTaskProps={myInvestmentProjectTaskProps}
-                {...investmentProjects}
-              />
-            ),
-          },
-          [urls.dashboard()]: {
-            label: 'My companies lists',
-            content: <CompanyLists />,
-          },
-        }}
-      />
-    </StyledDiv>
-  )
-}
+}) => (
+  <StyledDiv data-cy="dashboard-tabs">
+    <TabNav
+      id={`${id}.TabNav`}
+      label="Dashboard"
+      routed={true}
+      tabs={{
+        [urls.personalisedDashboard.myInvestmentProjects()]: {
+          label: 'My projects',
+          content: (
+            <MyInvestmentProjects
+              myInvestmentProjectTaskProps={myInvestmentProjectTaskProps}
+              {...investmentProjects}
+            />
+          ),
+        },
+        [urls.dashboard()]: {
+          label: 'My companies lists',
+          content: <CompanyLists />,
+        },
+      }}
+    />
+  </StyledDiv>
+)
 
 DashboardTabs.propTypes = {
   id: PropTypes.string.isRequired,
--- a/src/client/components/DashboardTabs/index.jsx
+++ b/src/client/components/DashboardTabs/index.jsx
@@ -26,7 +26,7 @@ const DashboardTabs = ({
           label: 'My projects',
           content: (
             <MyInvestmentProjects
-              myInvestmentProjectTaskProps={myInvestmentProjectTaskProps}
+              taskProps={myInvestmentProjectTaskProps}
               {...investmentProjects}
             />
           ),
--- a/src/client/components/DashboardTabs/index.jsx
+++ b/src/client/components/DashboardTabs/index.jsx
@@ -11,29 +11,48 @@ const StyledDiv = styled('div')`
   padding-top: 16px;
 `
 
-const DashboardTabs = ({ id, adviser }) => (
-  <StyledDiv data-cy="dashboard-tabs">
-    <TabNav
-      id={`${id}.TabNav`}
-      label="Dashboard"
-      routed={true}
-      tabs={{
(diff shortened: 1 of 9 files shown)


src/client/components/DashboardTabs/index.jsx
```diff
@@ -11,29 +11,48 @@ const StyledDiv = styled('div')`
...
-      }}
-    />
-  </StyledDiv>
-)
+const DashboardTabs = ({
+  id,
+  investmentProjects,
+  myInvestmentProjectTaskProps,
+}) => {
+  return (
```
ian-leggett, line 19: We can remove the return

ian-leggett (commented):
Its looking really good, just a couple of comments:

1. I think we need to move the “Personalised dashboard/MyInvestmentProjects” components up to the app folder as they are not “presentational” components, they are both connected to the store. These components are also domain specific so they wouldn't be used anywhere else. I understand that there are other components like this in the client folder that need to be moved too.


2. When loading the “My projects” tab the contents flickers, it basically re-fetches the results when it already has it in the store, we need to guard against this as we don’t need to do a re-fetch. The task component should do this for free as the useEffect hook should only fire if the state has changed.

What you can do is toggle the "status" prop on the task component, try this and see how quick the results come back as it doesn't re-fetch

Try this in the PersonalisedDashboard.jsx and see what happens.
```
const myInvestmentProjectTaskProps = {
    name: TASK_GET_MY_INVESTMENTS_LIST,
    id: MY_INVESTMENT_PROJECTS_ID,
    progressMessage: 'Loading your investment projects',
    startOnRender: {
      payload: {
        adviser,
        page,
        filter,
        sort,
        results,
      },
      onSuccessDispatch: MY_INVESTMENTS__LIST_LOADED,
      status: Boolean(results.length), //This will stop the re-fetch
    },
```

paulgain:
> Its looking really good, just a couple of comments:
> 
> 1. I think we need to move the “Personalised dashboard/MyInvestmentProjects” components up to the app folder as they are not “presentational” components, they are both connected to the store. These components are also domain specific so they wouldn't be used anywhere else. I understand that there are other components like this in the client folder that need to be moved too.
> 2. When loading the “My projects” tab the contents flickers, it basically re-fetches the results when it already has it in the store, we need to guard against this as we don’t need to do a re-fetch. The task component should do this for free as the useEffect hook should only fire if the state has changed.
> 
> What you can do is toggle the "status" prop on the task component, try this and see how quick the results come back as it doesn't re-fetch
> 
> Try this in the PersonalisedDashboard.jsx and see what happens.
> 
> ```
> const myInvestmentProjectTaskProps = {
>     name: TASK_GET_MY_INVESTMENTS_LIST,
>     id: MY_INVESTMENT_PROJECTS_ID,
>     progressMessage: 'Loading your investment projects',
>     startOnRender: {
>       payload: {
>         adviser,
>         page,
>         filter,
>         sort,
>         results,
>       },
>       onSuccessDispatch: MY_INVESTMENTS__LIST_LOADED,
>       status: Boolean(results.length), //This will stop the re-fetch
>     },
> ```

1. When we all spoke yesterday at length this wasn't brought up at the time. As you know we already have a separate card (Move PersonalisedDashboard up to apps) so the work can be done in there, is that OK?
2. I wasn't aware of that `status` field so thanks for that. There's no need to cast the `number` to a `Boolean` as `results.length` is a `falsy` value.

paulgain:
@ian-leggett I've just noticed that applying the line `status: Boolean(results.length)` breaks the filtering, sorting and pagination. I've also noticed the other tab `My Companies list` also fetches the API more than once when tabbing between the two. I wonder if this is a short coming on the TabNav/Task components.

src/client/components/PersonalisedDashboard/index.jsx
```diff
@@ -47,81 +68,113 @@ const SearchContainer = styled(Container)`
...
-                id={ID}
-                progressMessage="Loading your reminders"
-                startOnRender={{
-                  payload: { adviser },
-                  onSuccessDispatch: OUTSTANDING_PROPOSITIONS__LOADED,
-                }}
+  investmentProjects,
+  outstandingPropositions,
+}) => {
+  const { page, filter, sort } = investmentProjects
```
tomkingston, line 75: These could end up with a name conflict if you introduced another task (like companyLists page etc), so might be good to extract them in the task payload props itself, then again that is probably premature abstraction at this stage (it can be updated if the companyLists are added later).

src/client/components/PersonalisedDashboard/state.js
```diff
@@ -1,6 +1,4 @@
-export const ID = 'personalisedDashboard'
+export const ID = 'outstandingPropositions'
```
tomkingston, line 1: This doesn't seem like the right name for the personalised dashboard state key. 
Look at https://github.com/uktrade/data-hub-frontend/blob/master/src/apps/investments/client/projects/state.js
It keeps the ID as `projectsList` and has three different tasks, the results of which are surfaced via state2props:
```
TASK_GET_PROJECTS_LIST 
TASK_GET_ADVISER_NAME 
TASK_GET_INVESTMENTS_PROJECTS_METADATA
```

The reducer, state and tasks for the MyInvestmentProjects component are still nested with the component, when I think the reason for doing this ticket was to move all of them up to here, where they will need to be combined with everything else.

tomkingston (commented):
As discussed, I'm starting to wonder what we get out of moving the state up... maybe if the state, tasks and reducer were also moved up from `MyInvestmentProjects` the advantages might become slightly clearer.... but then maybe not.

src/client/components/PersonalisedDashboard/index.jsx
paulgain: I tried to pass in the `investmentProjects ` object but that caused the Task component to make two API calls.

src/client/components/PersonalisedDashboard/state.js
paulgain: Yeah, that makes sense.

PR #3443, opened 2021-03-30
Feature/create investment opportunity

peterhudec:
## Description of change

Adds the _Create UK investment opportumity_ page at `/investments/opportunities/create`.

Note the form is using the vanila `<form/>` instead of any of the `Form` components that we have, all of which suffer from a [bug where they can't be used in conjunction with the `Task` component](https://trello.com/c/C0qlxm1N/542-fix-the-flawed-isloading-mechanism-of-forms).

This PR also introduces three new components: `HardRedirect`, `ReferrerLink` and `TaskLoadingBox`.

## Test instructions

1. Navigate to `/investments/opportunities/create`
2. You should see the _Create UK investment opportumity_ form
3. Submit the form without entering any value into the _ Opportunity name_ field
4. You should see a validation error saying "Enter an opportunity name"
5. Now fill out the field with some text
6. Submit the form
7. Depending on what version of the API you are running it (whether a change was incorporated to the `/v4/large-capital-opportunity` which reduces the number of required fields to the POST method to only the `name`) it should
  - Render a `Task` error message overlay on top of the form
  - Redirect to `/investments/opportunities/<id-of-the-newly-created-opportunity>`

## Screenshots

Screen Shot 2021-03-30 at 1 56 29 PM


paulgain (approved):
LGTM. Just some very minor comments.

src/apps/investments/views/create/uk-investment-opportunity.njk
```diff
@@ -0,0 +1,5 @@
+{% extends "_layouts/collection.njk" %}
+
+{% block content %}
+  {% component 'react-slot', {id: 'create-uk-investment-opportunity'} %}
+{% endblock %}
```
rachaelcodes, line 5: Should we have a new line at the end of the file?

src/client/components/CreateUKInvestmentOpportunity/index.jsx
```diff
@@ -0,0 +1,120 @@
...
+import { FormActions, ErrorSummary } from '..'
+import Task from '../Task'
+import TaskLoadingBox from '../Task/LoadingBox'
+import HardRedirect from '../HardRedirect'
+import ReferrerLink from '../ReferrerLink'
+
+const TASK_NAME = 'CREATE_INVESTMENT_OPPORTUNITY'
+
+const StyledInputField = styled(InputField)({
+  marginBottom: '2rem',
```
rachaelcodes, line 22: Is there a constant we can use for this?

src/client/components/CreateUKInvestmentOpportunity/index.jsx
```diff
@@ -0,0 +1,120 @@
...
+              <StyledInputField
+                meta={
+                  error && {
+                    touched: true,
+                    error: error,
+                  }
+                }
+                input={{
+                  value: name,
+                  name: 'name',
```
rachaelcodes, line 79: Does this create aria-labels and placeholder text automatically? (I couldn't see anything in the govuk-react repo, but might have missed something)

src/apps/investments/router-opportunities.js
```diff
@@ -9,4 +9,13 @@ router.get(
   renderOpportunitiesView
 )
 
+router.get('/create', (req, res) => {
+  const heading = 'Create UK investment opportumity'
```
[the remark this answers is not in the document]
peterhudec: Oh, it's actually used twice lines 15 and 17.

src/client/components/CreateUKInvestmentOpportunity/index.jsx
peterhudec: Not these two lines, but the resulting markup is an input wrapped in a label so there should not be any need for `aria-label` tags:
```html
<label class="sc-iJuUWI gmLhrZ sc-iMZFOo ccjrzp">
  <span class="sc-giIncl dvrzOe">
    <!-- Here's the label -->
    Opportunity name
  </span>
  <input type="text" name="name" id="field-name" class="sc-dkIXFM gWtCDz" value="sadfasdf">
</label>
```

PR #3445, opened 2021-03-31
Add Trade Agreement fields to event form

ccp92:
## Description of change

Two new fields have been added to the Events form:
`Are there related Trade Agreements`
`Related Trade Agreements`

A feature flag of `relatedTradeAgreements` toggles these fields.

## Test instructions

Two new fields should be required fields when the feature flag is enabled.
When disabled, the fields should now show and should not be required.

### Before
image

### After
image

## Checklist

### DIFF
--- a/src/apps/events/controllers/edit.js
+++ b/src/apps/events/controllers/edit.js
@@ -57,6 +57,7 @@ async function getEditOptions(req, createdOn, currentAdviser) {
 async function renderEditPage(req, res, next) {
   try {
     const eventData = transformEventResponseToFormBody(res.locals.event)
+    const featureFlags = res.locals.features
 
     const eventId = get(eventData, 'id', '')
     const eventName = get(eventData, 'name')
@@ -73,7 +74,7 @@ async function renderEditPage(req, res, next) {
     )
 
     const eventForm = buildFormWithStateAndErrors(
-      eventFormConfig(assign({}, { eventId }, options)),
+      eventFormConfig(assign({}, { eventId }, options), featureFlags),
       mergedData,
       get(res.locals, 'form.errors.messages')
     )
--- a/src/apps/events/controllers/edit.js
+++ b/src/apps/events/controllers/edit.js
@@ -57,7 +57,6 @@ async function getEditOptions(req, createdOn, currentAdviser) {
 async function renderEditPage(req, res, next) {
   try {
     const eventData = transformEventResponseToFormBody(res.locals.event)
-    const featureFlags = res.locals.features
 
     const eventId = get(eventData, 'id', '')
     const eventName = get(eventData, 'name')
@@ -74,7 +73,9 @@ async function renderEditPage(req, res, next) {
     )
 
     const eventForm = buildFormWithStateAndErrors(
-      eventFormConfig(assign({}, { eventId }, options), featureFlags),
+      eventFormConfig(assign({}, { eventId }, options), {
+        ...res.locals.features,
+      }),
       mergedData,
       get(res.locals, 'form.errors.messages')
     )
--- a/src/apps/events/controllers/edit.js
+++ b/src/apps/events/controllers/edit.js
@@ -1,5 +1,5 @@
 /* eslint camelcase: 0 */
-const { assign, merge, get, pickBy } = require('lodash')
+const { merge, get, pickBy } = require('lodash')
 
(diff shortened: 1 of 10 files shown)


src/apps/events/transformers.js
```diff
@@ -198,7 +205,22 @@ function transformEventFormBodyToApiRequest(props) {
...
+        related_trade_agreements = null
+      } else {
+        related_trade_agreements = castCompactArray(
+          props.related_trade_agreements
+        )
+      }
+    } else if (props.related_trade_agreements_exist === 'false') {
+      related_trade_agreements = []
+    }
+  }
```
paulgain, line 223: This looks somewhat verbose, you could improve the function by returning something similar to this:
```js
function castRelatedTradeAgreements() {
  return props.related_trade_agreements_exist === 'true' 
    ?  !props.related_trade_agreements.length ? null 
    : castCompactArray(props.related_trade_agreements) : []
}
```

src/apps/events/transformers.js
```diff
@@ -184,9 +184,16 @@ function transformEventResponseToFormBody(props = {}) {
...
+function transformEventFormBodyToApiRequest(props, featureFlags) {
   const teamsArray = castCompactArray(props.teams)
   const related_programmes = castCompactArray(props.related_programmes)
+  let related_trade_agreements
+  if (featureFlags && featureFlags.relatedTradeAgreements) {
+    castRelatedTradeAgreements()
+  } else {
+    related_trade_agreements = []
+  }
+
```
paulgain, line 196: Same as above.

paulgain (commented):
Hi @ccp92 looks good, I've added a few minor comments.

src/apps/events/controllers/edit.js
```diff
@@ -73,7 +74,7 @@ async function renderEditPage(req, res, next) {
     )
 
     const eventForm = buildFormWithStateAndErrors(
-      eventFormConfig(assign({}, { eventId }, options)),
+      eventFormConfig(assign({}, { eventId }, options), featureFlags),
```
tomkingston, line 77: as Paul mentioned, the convention here is to use the spread operator rather than assign

src/apps/macros.js
```diff
@@ -139,6 +139,47 @@ const globalFields = {
...
+  tradeAgreements: {
+    macroName: 'MultipleChoiceField',
+    name: 'related_trade_agreements',
+    label: 'Related Trade Agreements',
+    initialOption: '-- Select related trade agreement --',
+    options: [
+      {
+        label: 'No Related Trade Agreement',
+        value: '0b371c90-e48a-40ea-a536-0968fb181f6c',
+      },
```
tomkingston, line 152: As discussed on the call before, having a "No related trade agreement" option is an anti-pattern - you could even select that alongside some trade agreements. If you want to specify no trade agreements, you should send an empty array.

ccp92:
@tomkingston and @paulgain I think I've addressed your comments if you could please re-review 🙏

src/apps/events/controllers/edit.js
```diff
@@ -73,7 +73,9 @@ async function renderEditPage(req, res, next) {
     )
 
     const eventForm = buildFormWithStateAndErrors(
-      eventFormConfig(assign({}, { eventId }, options)),
+      eventFormConfig(assign({}, { eventId }, options), {
+        ...res.locals.features,
+      }),
```
tomkingston, line 78: I see the assign was in the existing code, but this can be re-written as:
```
const eventForm = buildFormWithStateAndErrors(
    eventFormConfig({ eventId, ...options}), res.locals.features),
    mergedData,
    ...
```

tomkingston: Since it was existing, feel free to leave the `assign` in, but there isn't a lot of point of spreading `res.locals.features` and then putting it back together again (just use `res.locals.features` instead of `{ ...res.locals.features }`).

ccp92: Ah ok, that makes more sense, I didn't clock that you meant the options variable there.
I can update it to use the spread on options as it's good practice to be tidying up code you're touching 👍

tomkingston (approved):
Nice one :+1:

paulgain (approved):
@ccp92 :muscle:

PR #3465, opened 2021-04-09
Add fakers to generate functional test data

tomkingston:
## Description of change

Firebreak project to refactor the sandbox.

This uses intercept instead of the sandbox. Factories are used to generate test fixtures instead of the static sandbox fixtures.

## Checklist


test/functional/cypress/factories/investment-projects.js
```diff
@@ -0,0 +1,31 @@
...
+  ...overrides,
+})
+
+export const investmentProjectsFactory = (overrides, length = 1) => {
+  let list = []
+  for (let i = 0; i < length; i++) {
+    list.push(investmentProjectFactory(overrides))
+  }
+  return list
+}
```
tomkingston, line 29: I'm planning to refactor this so it does not need to be repeated for every factory generating a list of items - ideally to also use a declarative style instead of the for loop.

test/functional/cypress/specs/dashboard-new/project-stage-tag.spec.js
```diff
@@ -0,0 +1,48 @@
...
+
+describe('Dashboard - Investment project stages', () => {
+  beforeEach(() => {
+    cy.setFeatureFlag(
+      'layoutTesting:9010dd28-9798-e211-a939-e4115bead28a',
+      true
+    )
+    cy.intercept('POST', '/api-proxy/v3/search/investment_project', {
+      body: {
+        count: 0,
```
tomkingston, line 11: This should be one now

test/functional/cypress/specs/dashboard-new/reminders-spec.js
```diff
@@ -1,17 +1,35 @@
...
+  const tomorrow = addDays(new Date(), 1)
+  const proposition1 = propositionFaker({
+    name: 'University Proposition',
+    investment_project: {
+      name: 'University buildings',
+      project_code: 'DHP-00007004',
+      id: '18750b26-a8c3-41b2-8d3a-fb0b930c2270',
+    },
+    deadline: tomorrow,
+  })
```
tomkingston, line 18: We don't actually need to specify these here (the faker will generate these anyway), but I thought it would be best to be explicit as these are all checked in the tests.

src/client/utils/date-utils.js
```diff
@@ -1,5 +1,9 @@
 import moment from 'moment'
-import { endOfToday, differenceInCalendarDays } from 'date-fns'
+import {
+  endOfToday,
+  format as formatFNS,
```
paulgain, line 4: We can't use `format` per se as it's already defined with the file. We need to refactor this file to remove `moment`.

test/functional/cypress/fakers/investment-project-summary.js
```diff
@@ -0,0 +1,65 @@
...
+  Object.fromEntries(
+    investmentProjectStages.map((stage) => [
+      snakeCase(stage.name),
+      {
+        label: stage.name,
+        id: faker.datatype.uuid(),
+        value: faker.datatype.number({ min: 0, max: 50 }),
+      },
+    ])
+  )
```
tomkingston, line 36: We might want to add a value override to this faker at a later point so that we can test specific values on the summary widget (we would also export this function), but for now that seems premature as it will only be needed when those new tests are added.

ian-leggett (approved):

paulgain (approved):

paulgain (approved):

PR #3475, opened 2021-04-13
Component/dismissable status message

peterhudec:
## Description of change

Adds the new `DismissableStatusMessage` component, which adds a cancel button to the `StatusMessage`.

## Test instructions

1. Find the component in Storybook.
2. It should behave exactly as `StatusMessage`,
3. It should have an extra with a cross mark button on the right
4. The button should be aligned to the baseline of the message text
3. The button should be labelled "dismiss"
4. If a function is passed to the `onDismiss` prop, it should be called with an `ClickEvent` when the button is clicked

## Screenshots

Screen Shot 2021-04-13 at 9 25 24 AM

### DIFF
--- a/src/client/components/StatusMessage/Dismissable.jsx
+++ b/src/client/components/StatusMessage/Dismissable.jsx
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types'
 import styled from 'styled-components'
 
 import StatusMessage from '.'
+import TextLikeButton from '../TextLikeButton'
 
 const StyledStatusMessage = styled(StatusMessage)({
   display: 'flex',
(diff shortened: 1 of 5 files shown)


src/client/components/StatusMessage/Dismissable.jsx
```diff
@@ -0,0 +1,29 @@
...
+  alignItems: 'baseline',
+  justifyContent: 'space-between',
+})
+
+const DismissableStatusMessage = React.forwardRef(
+  ({ children, onDismiss, ...props }, ref) => (
+    <StyledStatusMessage {...props} ref={ref}>
+      {children}
+      <TextLikeButton tabIndex={0} onClick={onDismiss} aria-label="dismiss">
+        {'\u{2715}'}
```
elcct, line 19: Could a comment that this is an X sign help?

elcct (approved):

ambrozic (approved):

PR #3481, opened 2021-04-16
Add UK Opportunities CollectionList page to investments app

LaurenQurashi:
## Description of change
This PR does a lot, so in general ... 

This PR adds the third tab to the investments app, which displays UK opportunities. This tab should show an unfiltered version of the CollectionList component, which lists all of the current UK opportunities. Filtering and downloading functionality should come in a later PR. This tab also adds links to the pre-existing create an opportunity functionality, as well as the opportunity details pages.

This PR also adds: 
- A refactor if the file structure between Opportunities list, and Opportunities Details to make a clear distinction between the two. 
- Amends a bug relating to the `addItemUrl` prop on the CollectionList component. The CollectionListHeader had the functionality to take the prop but the CollectionList component was missing the code to pass it down. 
- Adds a fix to the faulty pluralisation on the DataDownloadHeader. This header had some simple logic that would just add an 's' to the end of the `entityName` prop passed in, in order to make the grammar in the header correct. I've added an optional `entityNamePlural` prop you can pass in if your entity has a non-conforming name - ie Opportunity/Opportunities. I've also added storybook functionality to the basic story to demonstrate.

## Test instructions
After booting up your local, navigate to the investments app. You should see three tabs on this page now, with the final tab being UK Opportunities. When clicking this tab, you should see an unfiltered collection list of current opportunities. They should have 'updated on' metadata, along with dynamic location badges, which either show a singular location or should say "Multiple UK sites" if more than one location is available. When clicking on the title of the opportunity, you should be taken to that opportunity's details page. When clicking on the 'Add Opportunity' button in the header, you should be redirected to the `/investments/opportunities/create/` page. 

### Before

Screenshot 2021-04-20 at 15 56 22

### After

Screenshot 2021-04-26 at 12 52 24

## Checklist


src/apps/investments/client/opportunities/UnfilteredLargeCapitalOpportunityCollection.jsx
```diff
@@ -0,0 +1,51 @@
...
+      taskProps={collectionListTask}
+      collectionName="Opportunity"
+      items={results}
+      count={count}
+      onPageClick={onPageClick}
+      activePage={page}
+      isComplete={isComplete}
+      baseDownloadLink="/investments/opportunities/export"
+      entityName="opportunitie"
+      addItemUrl="/investments/opportunities/create"
```
LaurenQurashi, line 39: Not sure what to do about this. The code that handles pluralising depending on list length just adds an S to the end of the entity name if there's more than one and I'm not sure what logic is bet to implement the prevention of this.

src/apps/investments/client/opportunities/LargeCapitalOpportunityCollection.jsx
```diff
@@ -0,0 +1,5 @@
+import React from 'react'
+
+import Unfiltered from './UnfilteredLargeCapitalOpportunityCollection'
+
+export default ({ ...props }) => <Unfiltered {...props} />
```
LaurenQurashi, line 5: Here I'm creating a holder component. The Profiles tab seemed to follow the same convention - it checks for a flag present and depending on the flag it would either show the unfiltered version or the filtered version. Filters will be added in another ticket so for now the component just shows the unfiltered version.

src/apps/investments/client/opportunities/state.js
```diff
@@ -1,6 +1,7 @@
 export const TASK_GET_OPPORTUNITY_DETAILS = 'TASK_GET_OPPORTUNITY_DETAILS'
+export const TASK_GET_OPPORTUNITIES_LIST = 'TASK_GET_OPPORTUNITIES_LIST'
 
-export const ID = 'opportunityDetail'
+export const ID = 'opportunitiesList'
```
LaurenQurashi, line 4: Wasn't too sure about this either. I found it tricky to figure out what exactly this ID allows stuff to do? The profiles tab has a similar set up to the Opportunities tab and a lot of the patterns were copied for this. Profiles seemed to also only have one ID called `profilesList` but it is the only id available in the state for all their tasks. Is there a convention to naming this ID? I'm not sure whether it's right that the tasks in the opportunities details page also uses this ID? 🤷‍♀️

src/apps/investments/controllers/opportunities.js
```diff
@@ -13,4 +13,19 @@ const renderOpportunitiesView = (req, res, next) => {
...
+        props: {
+          opportunityId: req.params.opportunityId,
+        },
+      })
+  } catch (error) {
+    next(error)
+  }
+}
+
+module.exports = { renderOpportunitiesView, renderOpportunityDetailsView }
```
LaurenQurashi, line 31: Here I accidentally changed the original Details view controller to show the List view instead, then realised that was a silly thing to do as it broke the old Details view, so made a new one. This is what the controller for the list view looks like. 

```
const renderOpportunitiesView = (req, res, next) => {
  try {
    res
      .breadcrumb('UK Opportunities')
      .render('investments/views/opportunities', {
        heading: 'UK Opportunities',
        props: {
          opportunityId: req.params.opportunityId,
        },
      })
  } catch (error) {
    next(error)
  }
}
```

src/apps/investments/transformers/opportunities.js
```diff
@@ -46,4 +51,32 @@ const transformInvestmentOpportunity = ({
...
+}) {
+  function filteredLocationBadge(locationList) {
+    var locationNamesArray = getArrayNames(locationList)
+    var locationBadgeText = []
+    if (locationNamesArray.length > 1) {
+      locationBadgeText.push({ text: 'Multiple UK sites' })
+    } else if (locationNamesArray.length == 1) {
+      locationBadgeText.push({ text: locationNamesArray[0].toString() })
+    }
+    return locationBadgeText
```
LaurenQurashi, line 68: If there's a better way to do this please let me know! I struggled to get a more concise version to work but I would be grateful for any refactoring offered up on this! 😄

src/apps/investments/views/opportunities.njk
```diff
@@ -1,10 +1,21 @@
...
+      <div class="govuk-grid-column-full">
+        {% component 'react-slot', {
+            id: 'large-capital-opportunity-collection',
+            props: props
+          }
+          %}
+      </div>
+    </div>
+  </div>
+</div>
```
LaurenQurashi, line 20: Just copying the profile tab set up for the nunjucks that holds the react component.

src/apps/investments/views/opportunity-details.njk
```diff
@@ -0,0 +1,8 @@
+{% extends "_layouts/collection.njk" %}
+
+{% block content %}
+{% component 'react-slot', {
+  id: 'opportunity-details',
+  props: props
+} %}
+{% endblock %}
```
LaurenQurashi, line 8: When I created the Opportunities tab, I overwrote the details tab stuff, so had to make a new one once I realised.

src/client/components/CollectionList/CollectionHeader.jsx
```diff
@@ -40,6 +40,7 @@ function CollectionHeader({
       href={addItemUrl}
       buttonColour={GREY_3}
       buttonTextColour={BLACK}
+      data-test="add-item-url-button"
     >
```
LaurenQurashi, line 44: I added this data-test attribute so that I can grab the button in the functional Cypress tests

src/client/components/CollectionList/index.jsx
```diff
@@ -37,6 +38,7 @@ const CollectionList = ({
           <CollectionHeader
             totalItems={count}
             collectionName={collectionName}
+            addItemUrl={addItemUrl}
```
LaurenQurashi, line 41: The CollectionHeader already had the ability to take this attribute, but the CollectionList component was just missing the functionality to pass it down. I added it as I needed the Add Opportunities button.

src/apps/investments/client/opportunities/UnfilteredLargeCapitalOpportunityCollection.jsx
```diff
@@ -0,0 +1,51 @@
...
+    <CollectionList
+      taskProps={collectionListTask}
+      collectionName="Opportunity"
+      items={results}
+      count={count}
+      onPageClick={onPageClick}
+      activePage={page}
+      isComplete={isComplete}
+      baseDownloadLink="/investments/opportunities/export"
+      entityName="opportunitie"
```
LaurenQurashi, line 38: Not sure what to do about this. The code that handles pluralising depending on list length just adds an S to the end of the entity name if there's more than one and I'm not sure what logic is best to implement the prevention of this.

src/client/components/CollectionList/CollectionHeader.jsx
```diff
@@ -40,6 +40,7 @@ function CollectionHeader({
       href={addItemUrl}
       buttonColour={GREY_3}
       buttonTextColour={BLACK}
+      data-test="add-item-url-button"
```
LaurenQurashi, line 43: I added this data-test attribute so that I can grab the button in the functional Cypress tests.

test/visual/specs/data-hub-pages-spec.js
```diff
@@ -61,13 +61,15 @@ describe('Data Hub', () => {
...
-  //   await hideScrollBarIE()
-  //   await browser.imageDiff.take()
-  //   const result = await browser.imageDiff.validate()
-  //   assert.equal(result, 0)
-  // })
+  it('investments page', async () => {
+    await browser.url('/investments/projects')
+    await hideScrollBarIE()
+    const filter = await $('#field-stage')
+    await filter.waitForDisplayed({ timeout: 10000 })
```
LaurenQurashi, line 68: Pippo and I added this to force the tests to wait for the filters to load and then take the screenshots. This is because it would regularly fail by taking a screenshot too quickly and then complain that there's no stuff on the page

src/apps/investments/client/opportunities/UnfilteredLargeCapitalOpportunityCollection.jsx
```diff
@@ -0,0 +1,51 @@
...
+} from '../../../../client/actions'
+
+const LargeCapitalOpportunityCollection = ({
+  page,
+  count,
+  results,
+  onPageClick,
+  isComplete,
+}) => {
+  const collectionListTask = {
```
ian-leggett, line 18: If there is only one use case I would just inline this into line 30

src/apps/investments/client/opportunities/LargeCapitalOpportunityCollection.jsx
ian-leggett: I can't see why we need this file. The profile tab uses a ternary whereas yours just imports one file, renames it and forwards the props onto the renamed component. We can remove this file and just export one instead.

src/apps/investments/client/opportunities/state.js
ian-leggett: This is important so all the state changes that relate to this feature fall under one id and not just spread on the root which could quite easily have some very strange effects to the UI. I can demo this to you if you like 👍

src/apps/investments/transformers/opportunities.js
```diff
@@ -1,8 +1,13 @@
+const { investments } = require('../../../lib/urls')
```
ian-leggett, line 1: We need to remove this file, we can't really do anymore data transformation in this layer as we are moving towards a client side app. We can apply some of this logic in the React layer.

src/apps/investments/transformers/opportunities.js
ian-leggett: See my comment above 👍

src/client/components/CollectionList/CollectionHeader.jsx
ian-leggett: We should lift this test hook higher up to the root of the collection header as you want to test wether this button renders within the context of the component.

test/functional/cypress/specs/investments/investment-opportunity-details.js
```diff
@@ -62,7 +62,7 @@ describe('UK Opportunity with complete data', () => {
   })
 
   it('should display required field tags', () => {
-    cy.get('#opportunities').should('contain', 'Complete')
-    cy.get('#opportunities').should('contain', 'Complete')
+    cy.get('#opportunity-details').should('contain', 'Complete')
```
ian-leggett, line 65: We can't use id's here we need to use data attributes such as `data-test`

test/functional/cypress/specs/investments/investment-opportunity-details.js
```diff
@@ -47,8 +47,8 @@ describe('UK Opportunity with missing data', () => {
   })
 
   it('should display required field tags', () => {
-    cy.get('#opportunities').should('contain', '7 fields required')
-    cy.get('#opportunities').should('contain', '3 fields required')
+    cy.get('#opportunity-details').should('contain', '7 fields required')
+    cy.get('#opportunity-details').should('contain', '3 fields required')
```
ian-leggett, line 51: We can't use id's here we need to use data attributes such as `data-test`

ian-leggett (changes requested):

test/functional/cypress/specs/investments/investment-opportunity-details.js
LaurenQurashi: Will add this change into the current outstanding PR around the opportunities details page! 🎉

src/apps/investments/client/opportunities/Details/reducer.js
```diff
@@ -0,0 +1,40 @@
...
+
+export default (state = initialState, { type, result }) => {
+  switch (type) {
+    case INVESTMENT_OPPORTUNITY_DETAILS__LOADED:
+      return {
+        ...state,
+        ...transformInvestmentOpportunityDetails(result),
+      }
+    default:
+      return state
```
faissalMT, line 38: This pattern appears in a few places in this PR, unless you intend to expand this in future it might be clearer to represent this as a conditional return. I usually go by the rule of thumb that if I have less than 7 conditions I don't use a `switch`.

faissalMT (approved):
Just one minor issue but nothing that's a showstopper as far as I can tell

src/apps/investments/client/opportunities/Details/reducer.js
LaurenQurashi: Thank you so much for the review! That's a good point, I think in the next PR I'm going to be making it will be expanding, however, I'm not sure if it's more than 7 conditions so I'll swap it out for a conditional 😎 Thanks!

src/apps/investments/client/opportunities/List/UnfilteredLargeCapitalOpportunityCollection.jsx
```diff
@@ -0,0 +1,49 @@
...
+} from '../../../../../client/actions'
+
+const LargeCapitalOpportunityCollection = ({
+  page,
+  count,
+  results,
+  onPageClick,
+  isComplete,
+}) => {
+  return (
```
ian-leggett, line 18: We can use an implicit return here

src/apps/investments/client/opportunities/List/tasks.js
```diff
@@ -0,0 +1,43 @@
...
+              value: format(new Date(parseISO(created_on)), DATE_LONG_FORMAT),
+            },
+          ],
+        }
+      }
+      data = {
+        count: data.count,
+        results: data.results.map(transformInvestmentOpportunitiesList),
+      }
+      return data
```
ian-leggett, line 41: I would remove line 37 and just return 
```
{
        count: data.count,
        results: data.results.map(transformInvestmentOpportunitiesList),
      }
```

src/client/components/DownloadDataHeader/__stories__/DownloadDataHeader.stories.jsx
```diff
@@ -6,7 +6,14 @@ import DownloadDataHeader from '..'
 const downloadDataHeaderStories = storiesOf('DownloadDataHeader', module)
 
 const BasicDownloadDataHeader = () => {
-  return <DownloadDataHeader count={5} downloadLink="/#dl" />
+  return (
```
ian-leggett, line 9: We can use an implicit return here

ian-leggett (approved):

PR #3488, opened 2021-04-21
Add the ValidatedForm component and related fields

peterhudec:
## Description of change

Adds the `ValidatedForm` component together with compatible field components `InputField`, `TextareaField`, `RadiosField` and `SelectField` (`TypeaheadField` will come in due course). The PR is quite big, but if you subtract the docstring and Storybook lines, its just above 400 lines.

The `ValidatedForm` is a precursor to the `TaskForm` — a composition of the `ValidatedForm` and `Task` components, which should rid us of all the copy-pasting of the task related boilerplate. Its PR is already waiting in the queue.

### The [existing form components](https://github.com/uktrade/data-hub-frontend/tree/master/src/client/components/Form)

Are problematic, because:

* They use the React context which
   * Violates the Redux architecture
   * Tightly couples the [field components](https://github.com/uktrade/data-hub-frontend/tree/master/src/client/components/Form/elements) to the form preventing them to be used standalone
* The form's progress state is tied to the duration of a possibly async `onSubmit` function, which mechanism can never be used without violating the Redux architecture i.e. there's no way the form progress state can be tied to the progress of a _task_ that was started by the form submission. The only solution is to make an XHR request directly in the `onSubmit` handler, which violates the architecture. The issue is described in [this ticket](https://trello.com/c/C0qlxm1N/542-fix-the-flawed-isloading-mechanism-of-forms)
* Returning a string from the `onSubmit` handler sets the return value to `window.location.href` — another architecture violation
* The implementation is unnecessarily complicated, the [`MultiInstanceForm`](https://github.com/uktrade/data-hub-frontend/blob/master/src/client/components/Form/MultiInstanceForm.jsx) employs [10 different action types](https://github.com/uktrade/data-hub-frontend/blob/master/src/client/components/Form/reducer.js#L3-L14)
* Doesn't take care of the submit and secondary buttons although it should, forcing you to paste the following markup boilerplate in every instance fostering visual inconsistency and [cluttering the codebase](https://github.com/uktrade/data-hub-frontend/search?q=%3CFormActions%3E)
  ```jsx
   <FormActions>
     <Button>Submit</Button>
     <Link href="...">Cancel</Link>
   </FormActions>
  ```
* Most features are completely undocumented. One needs to dig in the code to find out how it's supposed to work. The only examples in storybook are those of form fields, which use the `FormStateful` (13 usages), only to make the fields work with no demonstration of the form features. There's no documentation whatsoever of the `MultiInstanceForm` (8 usages)

With all these issues, it would be wise to build the `TaskForm` on the existing form components and it's much simpler to write a new form component than to try to fix all the issues of the current ones.

### The `ValidatedForm` proposed in this PR

* Is fully compliant with the Redux architecture: no internal state, no context
* Abstracts away validation, submit and secondary action buttons and multiple steps
* The implementation is rather simple: the form itself only needs [3 action types](https://github.com/uktrade/data-hub-frontend/blob/dc6c88337707c8bda887992c1d01b5ea82fd9dd5/src/client/actions.js#L146-L148) (one more action is used by two of the four fields each)
* The submit mechanism is no different to that of the native `<form/>` apart from the fact, that it is only submitted if all fields are valid
* The fields are connected to the form explicitly and since the only coupling is the shape of their props, they are totally independent from the form
* [Fully documented](https://github.com/uktrade/data-hub-frontend/blob/dc6c88337707c8bda887992c1d01b5ea82fd9dd5/src/client/components/ValidatedForm/index.jsx#L203-L326)

There are currently only 4 compatible field implementations: `InputField`, `TextareaField`, `RadiosField` and `SelectField`, but I'm already working on a compatible `TypeaheadField` and more importantly a `TaskTypeahead` and `TaskTypeaheadField` (the current async `Typeahead` requires direct XHR calls in its `loadOptions` prop, which is yet another architecture violation).

## Example

This is a simple two-step `ValidatedForm` example from the docstring which also demonstrates inter-field validation. It's also present in the Storybook.

```jsx
<ValidatedForm
  id="foo"
  onSubmit={(e, values) => {
    e.preventDefault()
    alert(JSON.stringify(values, null, 2))
  }}
  validators={{
    username: (x) =>
      x?.length > 7 || Error('Must be at least 8 characters long'),
    password: (x) =>
      x?.match(/(?=.*[A-Z]+)(?=.*\d+)/) ||
      Error('Must contain an uppercase character and a digit'),
    passwordConf: (x, { password }) =>
      x === password || Error('Passwords must match'),
  }}
>
  {(field) => (
    <>
      <h2>Step 1</h2>
      <Input label="Username" {...field('username')} />
    </>
  )}
  {(field) => (
    <>
      <h2>Step 2</h2>
      <Input type="password" label="Password" {...field('password')} />
      <Input
        type="password"
        label="Repeat password"
        {...field('passwordConf')}
      />
    </>
  )}
</ValidatedForm>
```

## Test instructions

The only way to test this is in the ValidatedForm section in the Storybook.


paulgain:
Hi @peterhudec, this looks really good 👍 

I have a few questions and issues I've come across when playing with the forms in Storybook:

* Regarding the _Next_ and _Back_ buttons, I think the _Back_ button needs to be a link for consistency.
* After successful form submission, we need to be routed to a consumer-supplied URL.
* We should notify the user if they bail out early through a user journey (i.e. multi-step form), we currently show a basic confirmation dialog by _listening_ for the `window.beforeunload` event, somethinig like _Leave site? Changes that you made may not be saved._ At this point we would also need to _cleanup_ the _Store_ ensuring not to keep hold of stale values after the user has bailed.
* Step API calls - we need the ability to make API calls within any step of a multi-step user journey (we currently do this in the _Add company_ user journey to retrieve a DIT region based on a UK postcode) this is important, and should be factored into the form architecture as soon as possible, rather than painfully retrofitting it at a later point in time.
* Form validation - consider using something like Joi: https://www.npmjs.com/package/joi which is very powerful, seasoned, and popular with ~3.4 million weekly downloads.
* I've found an issue with the multi-step form in Storybook... do not enter any form values and simply submit, the form correctly invalidates, now add some characters to the `Textarea` field (knowing it will invalidate), notice the red border around the element is removed on keypress as if your input was validated on the fly (when it hasn’t), now correct your input (knowing it will validate) and submit the form for a second time and the validation error correctly is removed - hope this makes sense.
* I think we should create a directory called **legacy** and move all previous form implementations (x3) into that directoy as a precondition to this PR being approved and subsequently merged. This will stop devs from being confused with which form implementation they should be using.

PR #3490, opened 2021-04-22
Add trade agreements to interactions

faissalMT:
## Description of change

This PR adds trade agreement functionality to interactions, including a new component to implement a multiselect pattern found in the Nunjucks forms.

This relies on the following api PR: https://github.com/uktrade/data-hub-api/pull/3451

~~@ian-leggett mentioned that it would be good to improve the accessibility on this in order to help fix some accessibility issues elsewhere, I'll try to get to this when I'm free.~~ this will now be dealt with after this goes live as part of the broader piece of work on improving accessibility.  

## Test instructions

- Enable the "trade-agreement-interaction-v4-endpoint" feature flag in the API and switch to the branch in the aforementioned PR
- Navigate to a company, choose 'add interaction', select 'trade agreement'
- You should see trade agreement specific guidance text
- Select the 'trade agreements' service and another service
- You should see a question about trade agreements, answer yes
- Add some trade agreements

### Before
Screenie_6
Screenie_5

### After

Screenie_1
Screenie_7

## Checklist


ian-leggett (commented):
Just had a good look over this and it does looks great and works well but there are a couple of things that I think we need to look at.

1. The new component
This PR creates a new component which basically wraps the typeahead and gives you options to add another value. This might be a bit of pre-optimisation but I would probably rename this to `FieldAddAnother` and abstract the typeahead as its very similar to a pattern we know from MOJ - https://moj-design-system.herokuapp.com/components/add-another we actually have this in nunjucks form `./src/templates/_macros/form/add-another.njk`.

I am also a little worried about the accessibility as we already know that we need to fix accessibility issues on the typeahead so does this add to our problems or does it actually solve some of our problems, maybe it solves some and we need to use this for all multiple select typeaheads. We just need to be careful with adding this, be good to get more input from other devs.

2. Moving away from node

We need to remove all the code from the node layer and handle API calls and data transformations in the React/client layer. If we don’t do this then we are just adding to the technical debt we already have. Once this is all moved over to the client layer the testing also becomes easier and you don’t need sandbox.

Happy to pair on some of this work

faissalMT:
Thanks for taking a look at this @ian-leggett, Miranda needs this pretty urgently, and I don't think we'll be able to manage that by Thursday. I can record these tasks on our board and we should be able to get to them right after this piece of work if that's alright?

ian-leggett:
> Thanks for taking a look at this @ian-leggett, Miranda needs this pretty urgently, and I don't think we'll be able to manage that by Thursday. I can record these tasks on our board and we should be able to get to them right after this piece of work if that's alright?

After our discussion I understand that this feature is urgent and for that reason we are going to accept the tech debt. I have created 2 tickets in the maintenance backlog that address the issues I raised, these are now ready to be picked up.

ian-leggett (approved):

src/apps/interactions/apps/details-form/client/InteractionDetailsForm.jsx
```diff
@@ -99,11 +103,22 @@ const InteractionDetailsForm = ({
                 initialValues={initialValues}
                 submissionError={saveInteractionTask.errorMessage}
                 onSubmit={(values) => {
+                  let newValues = { ...values }
+                  if (
+                    !(
+                      values.service ===
+                        TRADE_AGREEMENT_IMPLEMENTATION_ACTIVITY &&
+                      values.has_related_trade_agreements
+                    )
```
ccp92, line 112: This could possibly be extracted out into a constant to elaborate on what the comparison is as this doesn't read very nicely.

ccp92 (approved):
One minor comment which is not a blocking change for me 👍

PR #3493, opened 2021-04-23
Removed codeclimate

ambrozic:
## Description of change

No one seems to be particularly interested about codeclimate, we have lost access to codeclimate dashboard but most of all it is removed because security concerns.


cgsunkel:
We can delete this file as well - https://github.com/uktrade/data-hub-frontend/blob/master/.codeclimate.yml

cgsunkel (approved):

cgsunkel:
It's also worth mentioning that the Codeclimate checks haven't actually run since January as the webhook was removed.

tomkingston (approved):

PR #3509, opened 2021-04-28
Make FieldTradeAgreementList more generic and rename to FieldAddAnother

faissalMT:
## Description of change

This PR cleans up part of the work on that was done in https://github.com/uktrade/data-hub-frontend/pull/3490
It makes the field generic, allowing it to be used outside of the context of trade agreements as well as enabling more complex subforms and improving accessibility.

## Test instructions

If you go to FieldAddAnother in Storybook selecting add or remove will now let you know what item you're adding or removing. Other changes in this PR are not immediately user-facing.  

## Checklist


package.json
```diff
@@ -129,6 +129,7 @@
     "numeral": "^2.0.6",
     "nunjucks": "^3.2.3",
     "nunjucks-markdown": "^2.0.1",
+    "ordinal": "^1.0.3",
```
ian-leggett, line 132: Maybe we should use a vanilla approach rather than using ordinal. Be good to not have another dependancy if we can help it.

ian-leggett (commented):
This looks really good and works well, just a couple of things.
1. If we can remove the dependancy that would be good
2. Should you be able to add another and select the same option as the first?

src/apps/interactions/apps/details-form/client/StepInteractionDetails.jsx
```diff
@@ -259,13 +260,30 @@ const StepInteractionDetails = ({
...
-              placeholder="-- Search trade agreements --"
-              options={relatedTradeAgreements}
-            />
+              item-name="trade agreement"
+            >
+              {({ value, onChange, error }) => (
+                <Typeahead
+                  name="related_trade_agreements"
+                  inputId="related_trade_agreements"
+                  label={''}
```
paulgain, line 274: Is this actually required? If so, consider using `label=""`

package.json
paulgain: This is a tiny utility, essentially it's just these two files. We could easily follow that!
https://github.com/dcousens/ordinal.js/blob/master/indicator.js
https://github.com/dcousens/ordinal.js/blob/master/index.js

paulgain (commented):
@faissalMT just a couple of comments.

paulgain (approved):

PR #3540, opened 2021-05-06
Add opportunity edit functionality - name only

LaurenQurashi:
## Description of change

This PR is the first in a series of PRs towards adding edit functionality to a UK Opportunity. This particular PR allows for an edit to be made for the name only. If this work is accepted, then the next PR will work on building up the Opportunity Details form, and the Opportunity requirements form. 

This PR contains tech debt. It's my current understanding that the Form component has a bug in it, which means that it doesn't actually wait for the Task it holds to finish, before doing the next thing which can result in things not updating as quickly as we would want. My PR uses this broken form component, as there's no agreed alternative. 

## Test instructions

Go to investments > UK opportunities and click on an opportunity. Click on the opportunity details toggle, and click the Edit button at the bottom of the section. Fill out a new name and click Submit. You should be redirected back to the section with the summary table in it, showing your new name.

### Before

_Add a screenshot_

### After

_Add a screenshot_

## Checklist


src/apps/investments/client/opportunities/Details/Opportunities.jsx
```diff
@@ -49,84 +55,126 @@ const RequiredFields = (fieldCount) => {
...
+                      isEditing={isEditingRequirements}
+                      onEdit={onRequirementsEdit}
+                    />
+                  </>
+                ),
+              },
+            }}
+          />
+          <StyledToggle
+            variant={VARIANTS.SECONDARY}
```
tomkingston, line 151: Once you rebase, you will be able to get rid of the variants (the default variant is the same so just import ToggleSection)

src/apps/investments/transformers/opportunities.js
```diff
@@ -1,6 +1,10 @@
-const getArrayNames = (data) => data.map((d) => d.name)
 const getArrayNamesAndIds = (data) =>
-  data.map((d) => ({ name: d.name, id: d.id }))
+  data.map((d) => ({ label: d.name, value: d.id }))
+
+const getNameAndId = (data) =>
+  data ? { value: data.id, label: data.name } : {}
+
+const idName2valueLabel = ({ id, name }) => ({ value: id, label: name })
```
tomkingston, line 7: The casing on this looks a bit strange - shouldn't it be `idName2ValueLabel`... or even better `idNameToValueLabel`

src/apps/investments/client/opportunities/Details/tasks.js
```diff
@@ -1,7 +1,60 @@
 import { apiProxyAxios } from '../../../../../client/components/Task/utils'
+import { OPPORTUNITY_FIELD_NAME } from './constants'
 
-export function getOpportunityDetail(id) {
+const idName2valueLabel = ({ id, name }) => ({ value: id, label: name })
```
tomkingston, line 4: As before, the casing on this looks a bit strange - `idNameToValueLabel` would be best

tomkingston (commented):
Hi Lauren - left a few comments. My main query is on why you are nesting everything under `details` rather than using the old pattern which appears to be a bit simpler?

It looks like the commits could also be tidied up a bit to avoid WIP / fix up commit messages.

src/apps/investments/client/opportunities/Details/reducer.js
```diff
@@ -32,7 +53,61 @@ export default (state = initialState, { type, result }) => {
     case INVESTMENT_OPPORTUNITY_DETAILS__LOADED:
       return {
         ...state,
-        ...transformInvestmentOpportunityDetails(result),
+        details: { ...transformInvestmentOpportunityDetails(result) },
```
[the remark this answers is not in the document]
LaurenQurashi: spoke on slack - Details is nested due to the final wireframe having 5 tabs, details being one of them, decided to keep the nesting 😎

peterhudec (changes requested):
Getting there, but there are some issues. Although not introduced by this PR, but heavily affecting it is the massive reducer. It's tightly coupling all the components together. It needs to be broken down into smaller reducers owned by the respective sub-components.

peterhudec:
The form seems to work fine in that it updates the name, but there is a slight UX defficiency:

When the form is submitted, the user is returned back to the read only view, which however is still showing the old name. The user must refresh the page to see the updated name.

PR #3554, opened 2021-05-11
Remove conditional rendering of Trade Agreement fields for Interactions

ccp92:
## Description of change

Previously, the Trade Agreement fields were only visible for Trade Agreement option interactions.
This field is now available regardless of the type of interaction.

There have also been minor changes with regards to casing/wording of some fields and guidance text.

## Test instructions

Attempting to add an interaction of any kind, you should now have a new mandatory field of "Related Trade Agreements".
If "Yes" is selected, you will see a multi-select field for specifying the trade agreements.
To test, enable the feature flag: `trade-agreement-interaction-v4-endpoint`

### Before

image

### After

image

## Checklist


test/selectors/interaction-form.js
```diff
@@ -43,4 +43,5 @@ module.exports = {
     export: '#field-' + EXPORT_INTEREST_STATUS.EXPORTING_TO,
     noInterest: '#field-' + EXPORT_INTEREST_STATUS.NOT_INTERESTED,
   },
+  hasRelatedTradeAgreementsNo: '#field-has_related_trade_agreements [value=no]',
```
ian-leggett, line 46: Ideally we need to stop adding more selectors to these files as we have all decided to opt for the `data-test` attribute instead as a way of targeting elements. However I understand that these tests are quite old and ideally it all needs a bit of a refactor.

ian-leggett (approved):

test/selectors/interaction-form.js
chloehkwong1: @ian-leggett  Thanks for the feedback. Yes, we just followed the existing pattern for those tests so were unaware that the `data-test` attribute is the preferred way now. Happy to change it to use `data-test` but it will look slightly out of place with the existing tests. Thoughts?

ian-leggett: @chloehkwong1 I think this is fine for now as ideally we need to remove this whole file but this will probably mean a larger refactor to all of the tests around the interaction form.

chloehkwong1: @ian-leggett Okay sounds good. Thanks :)

src/apps/interactions/labels.js
```diff
@@ -29,8 +29,8 @@ const interaction = {
   service_delivery_status: 'Service status',
   grant_amount_offered: 'Grant offered',
   has_related_trade_agreements:
-    'Does this interaction relate to a trade agreement?',
-  related_trade_agreements: 'Trade agreement(s)',
+    'Does this interaction relate to a named trade agreement?',
```
chloehkwong1, line 32: @ian-leggett I've just made a small wording change based of our show and tell yesterday. Is this okay with you?

src/apps/interactions/labels.js
```diff
@@ -29,8 +29,8 @@ const interaction = {
...
+  related_trade_agreements: 'Named trade agreement(s)',
```
chloehkwong1, line 33: @ian-leggett I've just made a small wording change based of our show and tell yesterday. Is this okay with you?

paulgain (approved):

PR #3581, opened 2021-05-19
Added companies collection to new react companies page

tomkingston:
## Description of change

Adds the collection list to the new react companies page. Note that this does not include filters or sort by. You may also notice that other elements need some work before this is fully ready (for example the download header should say "companies" instead of "companys"), but this is just to give a platform to build out the other filters etc.

## Test instructions

Go to `/companies/react` and see the new page. Note that many features including the filters are not there yet.

### Before

Screenshot from 2021-05-19 14-02-09

### After

Screenshot from 2021-05-19 13-57-41

## Checklist

### DIFF
--- a/src/apps/companies/transformers/company-to-list-item.js
+++ b/src/apps/companies/transformers/company-to-list-item.js
@@ -4,9 +4,6 @@ const { get } = require('lodash')
 const labels = require('../labels')
 const urls = require('../../../lib/urls')
 
-const { addressToString } = require('../../../common/addresses')
-const { formatLongDate, formatMediumDateTime } = require('../../../common/date')
-
 module.exports = function transformCompanyToListItem({
   id,
   name,
@@ -25,27 +22,19 @@ module.exports = function transformCompanyToListItem({
   }
 
   const meta = []
-  const metadata = []
-  const badges = []
 
   if (trading_names && trading_names.length) {
     meta.push({
       label: labels.hqLabels.trading_names,
       value: trading_names,
     })
-    metadata.push({
-      label: labels.hqLabels.trading_names,
-      value: trading_names.join(', '),
-    })
   }
 
   if (sector) {
-    const sectorMeta = {
+    meta.push({
       label: 'Sector',
       value: get(sector, 'name'),
-    }
-    meta.push(sectorMeta)
-    metadata.push(sectorMeta)
+    })
   }
 
   if (address && address.country) {
@@ -54,7 +43,6 @@ module.exports = function transformCompanyToListItem({
       type: 'badge',
       value: get(address, 'country.name'),
     })
-    badges.push({ text: get(address, 'country.name') })
   }
 
   if (uk_based) {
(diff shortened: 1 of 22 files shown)


ccp92 (approved):
Really like the faker pattern for the intercept on the api call 👍

src/apps/companies/transformers/company-to-list-item.js
```diff
@@ -22,19 +25,27 @@ module.exports = function transformCompanyToListItem({
   }
 
   const meta = []
+  const metadata = []
```
chloehkwong1, line 28: Hi Tom, sorry if this is a silly question, but what is the difference between `metadata`  and `meta`? Is `meta` being phased out?

test/functional/cypress/fakers/addresses.js
```diff
@@ -0,0 +1,16 @@
+import faker from 'faker'
+
+const addressFaker = (overrides = {}) => ({
+  line_1: faker.address.streetAddress(),
+  line_2: faker.address.streetName(),
+  town: faker.address.city(),
+  county: faker.address.county(),
+  postcode: 'AB1 2DE',
```
chloehkwong1, line 8: Why has the postcode been hard-coded to a UK postcode when the country is randomised? If this was intentional, shouldn't country be fixed for UK?

test/functional/cypress/fakers/companies.js
```diff
@@ -0,0 +1,34 @@
...
+import { listFaker } from './utils'
+
+/**
+ * Generate fake data for a single company.
+ *
+ * Starts by generating data based on the json schema, adds some defaults and
+ * merges in overrides.
+ */
+const companyFaker = (overrides = {}) => ({
+  ...jsf.generate(apiSchema.components.schemas.Company),
```
chloehkwong1, line 16: VSCode flagged `jsf` with this error: `'jsf' is deprecated.ts(6385)`. Not sure if this is an issue but thought I'd let you know.

chloehkwong1 (commented):
Nice work! I learnt a lot from the faker stuff :) Just left a few small comments

test/functional/cypress/fakers/addresses.js
tomkingston: Yeah, the faker library didn't appear to have a generator for uk postcodes so I just hardcoded one - ideally this would be randomised too, but it doesn't really affect anything and probably not really worth the effort at this point. If we need to test international postcodes, we can just set those explicitly.

test/functional/cypress/fakers/companies.js
tomkingston: That's interesting... the [docs](https://github.com/json-schema-faker/json-schema-faker) say: 

> You will not longer be able to call jsf() and get a fully-dereferenced result. It will just generate given refs and inline ones, nothing else.
> - jsf.generate() is the sync-version, with partial dereferencing through given refs, etc.

So `jsf.generate(...` should be fine, but we don;t want to use `jsf(...`

chloehkwong1 (approved):
Looks good to me :)

test/functional/cypress/fakers/addresses.js
tomkingston: I have just updated this now to use `faker.address.zipcode` since this is potentially for anywhere in the world. So hopefully it looks a bit more random.

PR #3584, opened 2021-05-20
Fix/axios address lookup

tomkingston:
## Description of change

Rolls back hawk auth to use Request Promise - the axios code has been kept to fix later (since we need to migrate away from request promise).

## Test instructions

You need to connect to the staging environment backend as the dev env appears to be broken.

Try to add a company - when you reach the page where you select DIT sector and region it shouldn't break!

## Checklist

### DIFF
--- a/src/lib/__test__/axios-hawk-request.test.js
+++ b/src/lib/__test__/axios-hawk-request.test.js
@@ -1,7 +1,11 @@
+/**
+ * Tests for axios version of hawk request
+ */
 const config = require('../../config')
 const rewire = require('rewire')
+const { StatusCodeError } = require('../errors')
 
-const modulePath = '../hawk-request'
+const modulePath = '../axios-hawk-request'
 
 const testDataHubCredentials = {
   id: 'test-key-id',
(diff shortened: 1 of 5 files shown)


src/lib/__test__/axios-hawk-request.test.js
```diff
@@ -0,0 +1,194 @@
...
+  //     )
+  //   ).to.be.rejectedWith(Error)
+
+  //   expect(authenticateStub).to.be.calledOnceWith(
+  //     error,
+  //     testDataHubCredentials,
+  //     testClientHeaderArtifacts,
+  //     { payload: '{}' }
+  //   )
+  // })
```
paulgain, line 170: Why are we commenting out this test?

paulgain (commented):
This works as expected on Staging. Just a single comment.

src/lib/__test__/axios-hawk-request.test.js
tomkingston: Hi Paul - sorry, I should have marked the PR as `Work In Progress` - the test is commented out because it fails, however we do not actually use `axios-hawk-request.js` anywhere - it's just in there to make it easy to restore the axios code once we get it working again. Essentially, I could remove `axios-hawk-request.js` and `axios-hawk-request.test.js` but have kept them in to make it easier to restore them.

I will have a go at fixing this test this morning.

tomkingston: This should be ok now

paulgain: @tomkingston thanks for the explanation and fix, much appreciated.

paulgain (approved):

ian-leggett (approved):

ambrozic (approved):

PR #3616, opened 2021-06-01
Contacts collection list (React rewrite)

paulgain:
## Description of change
This is the next step in rewriting the contacts list in React, this PR simply includes the paginated list of contacts, the filters/sorting and everything else will be developed in future iterations. This is the second PR in the series, following on from the [first](https://github.com/uktrade/data-hub-frontend/pull/3583).

## Test instructions
Navigate to: `/contacts/react`

### Before
contacts
_Add a screenshot_

### After
Screenshot 2021-06-01 at 08 58 00

## Checklist


src/apps/contacts/client/transformers.js
```diff
@@ -0,0 +1,72 @@
...
+  company,
+  modified_on,
+  archived,
+  company_sector,
+  primary,
+  telephone_countrycode,
+  telephone_number,
+  email,
+} = {}) => {
+  if (!id || (!first_name && !last_name)) {
```
tomkingston, line 29: I know this is copied from the old transformer, but I'm wondering why `first_name` and `last_name` need to be checked - they are both required fields on the api, so if you have an ID, they should also be present.

src/client/components/CollectionList/CollectionItem.jsx
```diff
@@ -82,15 +74,7 @@ const CollectionItem = ({
...
-      {metadata && metadata.length > 4 ? (
-        <>
-          <StyledDetails summary={summaryMessage}>
-            <Metadata rows={metadata} />
-          </StyledDetails>
-        </>
-      ) : (
-        <Metadata rows={metadata} />
-      )}
+      <Metadata rows={metadata} />
```
tomkingston, line 77: The old code does appear a bit unnecessary, but it does perform a function (it hides the number of metadata rows when there are too many metadata fields) - I see @LaurenQurashi wrote that originally, perhaps we should check the original justification before deleting this?

test/functional/cypress/fakers/contacts.js
```diff
@@ -0,0 +1,99 @@
...
+  'Creative and Media',
+]
+
+const DIT_TEAM_NAMES = [
+  'Digital Data Hub - Live Service',
+  'Aberdeen City Council',
+  'Healthcare UK',
+]
+
+const getFakePerson = () => {
```
tomkingston, line 27: For consistency this should probably be called `userFaker` - since it can be used by lots of different collection lists it should probably also have its own faker file too ie `fakers/users.js`. You might also want to include the `dit_team` field in this faker.

tomkingston (approved):
Approved but I think we should really check what the justification for hiding the metadata before we delete that bit. There are also a couple of other minor comments.

src/client/components/CollectionList/CollectionItem.jsx
paulgain: I'm fully aware of what the code does, there really is no need to explain what is does. I'm working on a better solution so the collection lists can continue to work as expected without this limitation.

src/apps/contacts/client/transformers.js
paulgain: Both fields are required on the API **now**, that may not have been always the case. I'm curious to know why there are additional safe guards around these fields. Data Hub hasn't always had great data,  can you please run a DB query to ensure every contact has a `first_name` and `last_name` before I remove anything.

tomkingston: I'll see if I can access the database a bit later, but just looking through the git history, those fields were defined as required as soon as they were created back in 2016 (https://github.com/uktrade/data-hub-api/commit/2d708ea199e96a57dd668c8cd79741ab5efb5b11), so I wonder whether those checks were added for some sort of testing thing. I expect the safest thing is to just keep them in.

src/client/components/CollectionList/CollectionItem.jsx
tomkingston: Hey, sorry, I only added that explanation for a bit of context on the comment and for my own reference!

src/apps/contacts/client/transformers.js
paulgain: That's interesting, I'll run the query and see what comes back. I agree that it would be good to remove these additional checks.

paulgain: Confirmed,  all DH contacts in our DB have both `first_name` and `last_name` defined. I'll now remove these superfluous checks.

test/functional/cypress/fakers/contacts.js
paulgain: It doesn't make sense to include the `dit_team` field in the `userFaker` as not every user is coupled to a DIT team.

src/apps/contacts/client/transformers.js
paulgain: After running another DB query (including empty strings) there appears to be six Data Hub contacts that do not have a first name defined, even though the first name field is mandatory on the contact form. I've asked Live Services if they can request this from them. For the time being we will need to keep the `first_name` check in place.

paulgain: Live services have now added the six missing `first_name` fields, there's no need to test for any of the following three fields: `id`, `first_name` and `last_name`.

src/apps/contacts/client/state.js
```diff
@@ -0,0 +1,25 @@
...
+const collectionListPayload = (paramProps) => {
+  return Object.fromEntries(
+    Object.entries(searchParamProps(paramProps)).filter((v) => v[1])
+  )
+}
+
+export const state2props = ({ router, ...state }) => {
+  const queryProps = qs.parse(router.location.search.slice(1))
+  const filteredQueryProps = collectionListPayload(queryProps)
+  const obj = {
```
ian-leggett, line 17: Really minor but we could just return the object rather using a variable

src/client/components/CollectionList/CollectionItem.jsx
```diff
@@ -112,14 +101,14 @@ CollectionItem.propTypes = {
     })
   ),
   type: PropTypes.string,
+  metadataRenderer: PropTypes.func,
 }
 
 CollectionItem.defaultProps = {
```
ian-leggett, line 107: We should remove these default props and add defaults when deconstructing from line 48

src/apps/companies/apps/exports/client/ExportsHistory/index.jsx
```diff
@@ -16,6 +17,20 @@ const Wrapper = styled('div')`
   margin-top: ${SPACING.SCALE_3};
 `
 
+const StyledDetails = styled(Details)`
+  margin: ${SPACING.SCALE_3} 0 0 0;
+`
+
+const metadataRenderer = (metadata) => {
```
tomkingston, line 24: I guess this is pretty much a component at this point so I am wondering whether you should name it in that style, but then again it is also a function so I suppose it is whatever your preference is.

Rather than having a metadataRenderer, an alternative option might have been to just send the max items (4) as a prop - then if it is undefined you could show the default (non-StyledDetails version).

ian-leggett (approved):

src/apps/companies/apps/exports/client/ExportsHistory/index.jsx
paulgain: Initially I had this as an anonymous function, I then decided on a function expression as it feels a little cleaner, not cluttering the CollectionList. It's fine as a function, no different to a react Hook, bar the state. Happy to change the name if you can think of something more appropriate.

I did think about this approach, passing a single prop is not a problem, but overtime that doesn't scale, we'll end up passing lots of props which makes it harder to maintain. I think it's better to pass a renderer, for reusability and abstracting everything away and making it the responsibility of the consumer to define what should be rendered, keeping the CollectionItem as clean as possible.

Personally, I would like to see us define our own custom CollectionItem, and pass that to the CollectionList, if one isn't passed as a prop we fallback to the default, there maybe some duplication of code, however, we could mitigate that by composition. This approach ensures we keep the CollectionList and CollectionItem free of unnecessary conditionals and props explosion as the application grows over time.

PR #3621, opened 2021-06-02
Tidy up Investments collection list pages

ian-leggett:
## Description of change

Since the collection lists were converted into React we need to remove and re-organise a few files. We simply need to delete/modify some of the old files, these files will typically be middleware and transformers that were originally used in the node layer. Any data transformations we still use need to be moved into the client area and exported using es6 modules syntax. 

I also noticed some transformations happening in reducers so I have moved this closer to the API so they now sit alongside the task files.

## Test instructions
All investments collection list should function as before

1. View Investments projects `/investments/projects`
2. View Investments profiles /investments/profiles`
3. View UK opportunities `investments/opportunities`
4. View opportunities details page `/investments/opportunities` then click on an opportunity

## Checklist


src/client/utils/date-utils.js
```diff
@@ -35,13 +34,13 @@ export const formatWithTime = (dateTimeStr) => {
   return moment(dateTimeStr).format('D MMM YYYY, h:mma')
 }
 
-export const isDateValid = (year, month, day, format = DATE_FORMAT_LONG) => {
+export const isDateValid = (year, month, day, format = 'YYYY-MM-DD') => {
```
tomkingston, line 37: Every function appears to use a different date format... it would be nice if all these formats were assigned to constants which then get used in the functions so it's easy to tell the difference. Moment and date-fns use different formats, but perhaps this would be an opportunity to be explicit about which one is which ie:
```
const DATE_FORMAT_LONG = 'd MMMM yyyy'
const DATE_FORMAT_LONG_MOMENT = 'DD MMM YYYY'

const DATE_FORMAT_SHORT = 'yyyy-MM'
const DATE_FORMAT_SHORT_MOMENT = 'YYYY-MM'
```

This is probably out of scope for this PR, but would be nice to pick this up at some point - possibly as part of the work to remove moment.

src/apps/investments/client/projects/transformers.js
```diff
@@ -0,0 +1,45 @@
...
+  ]
+
+  const metadata = [
+    { label: 'Investor', value: investor_company.name },
+    { label: 'Sector', value: sector ? sector.name : '' },
+    {
+      label: 'Estimated land date',
+      value:
+        estimated_land_date &&
+        format(parseISO(estimated_land_date), 'MMMM yyyy'),
```
tomkingston, line 30: It would be nicer if this just used a format function from date-utils - but I think that can wait for a future PR

src/apps/investments/client/profiles/transformers.js
```diff
@@ -0,0 +1,19 @@
...
+import { DATE_FORMAT_LONG } from '../../../../client/utils/date-utils'
+
+const transformLargeCapitalProfiles = ({ investor_company, created_on }) => ({
+  headingText: investor_company.name,
+  headingUrl: companies.investments.largeCapitalProfile(investor_company.id),
+  itemId: investor_company.id,
+  metadata: [
+    {
+      label: 'Updated on',
+      value: format(new Date(parseISO(created_on)), DATE_FORMAT_LONG),
```
tomkingston, line 14: It would be nice if this just needed a single function from date utils, but that can wait for another time.

tomkingston (approved):
Looks good - I did notice that you cannot download the investment opportunities, but I expect this is a limitation of the heroku setup.

The date-utils also look like they could do with some refactoring but probably best to do that at a later point when moment gets removed.

src/apps/investments/client/profiles/transformers.js
ian-leggett: Makes sense, just refactored this 👍

src/apps/investments/client/projects/transformers.js
ian-leggett: Makes sense, just refactored this 👍

src/client/utils/date-utils.js
ian-leggett: I have added new constants for some of the date formats and added notes that display examples of those formats. Moment is still there but I have managed to clean up a couple of places where its being called. Its now clearer than ever that we have inconsistent date formats dotted around the app.

src/apps/investments/transformers/__test__/collection.test.js
```diff
@@ -15,20 +15,63 @@ describe('Investment project transformers', () => {
...
+
+      it('should list the stage as a badge', () => {
+        expect(this.transformedItem.meta[0]).to.deep.equal({
+          label: 'Stage',
+          type: 'badge',
+          value: this.rawItem.stage,
+        })
+      })
+
+      it('should list the invstment type as a badge', () => {
```
chloehkwong1, line 38:
```suggestion
      it('should list the investment type as a badge', () => {
```

src/apps/investments/client/projects/labels.js
```diff
@@ -0,0 +1,11 @@
+const labels = {
+  investmentProjectMetaItemLabels: {
```
tomkingston, line 2: We already know that these are project labels, so this `investmentProjectMetaItemLabels` namespacing seems a bit redundant.... however I see that is in existing code, so no need to do anything unless you want to

src/apps/investments/transformers/__test__/collection.test.js
```diff
@@ -15,20 +15,63 @@ describe('Investment project transformers', () => {
...
+      })
+
+      it('should list the sector', () => {
+        expect(this.transformedItem.meta[3]).to.deep.equal({
+          label: 'Sector',
+          value: this.rawItem.sector,
+        })
       })
 
       it('should list the esimated land date as date', () => {
```
chloehkwong1, line 61:
```suggestion
      it('should list the estimated land date as date', () => {
```

chloehkwong1 (approved):

src/apps/investments/client/projects/labels.js
ian-leggett: Good spot, just removed it

src/apps/investments/client/projects/transformers.js
```diff
@@ -0,0 +1,48 @@
...
+      label: 'Estimated land date',
+      value:
+        estimated_land_date &&
+        format(estimated_land_date, DATE_FORMAT_LONG_NO_DAY),
+    },
+  ].filter((metadata) => metadata.value)
+
+  return {
+    id,
+    type: 'investments/project',
```
paulgain, line 39: I've removed `type` from the `CollectionItem`.

src/apps/investments/transformers/__test__/collection.test.js
ian-leggett: Good spot thanks 👍

src/apps/investments/transformers/__test__/collection.test.js
ian-leggett: Good spot thanks 👍

src/apps/investments/transformers/__test__/collection.test.js
```diff
@@ -15,20 +15,63 @@ describe('Investment project transformers', () => {
...
+      })
+
       it('should provide the name', () => {
-        expect(this.transformedItem.headingText).to.equal(this.rawItem.name)
+        expect(this.transformedItem.name).to.equal(this.rawItem.name)
+      })
+
+      it('should provide the type', () => {
+        expect(this.transformedItem.type).to.equal('investments/project')
+      })
```
paulgain, line 28: Same as above.

src/apps/investments/client/projects/transformers.js
ian-leggett: Nice thanks just removed it 👍

src/apps/investments/transformers/__test__/collection.test.js
ian-leggett: Updated 👍

paulgain (approved):
LGTM

PR #3654, opened 2021-06-08
Filter company list by status

tomkingston:
## Description of change

Filters the new company collection list by status (Active / Inactive). Note that this actually relates to whether or not a company is archived (which is used in the url).

## Test instructions

Go the new react companies collection list page (/companies/react). There should be the option to filter by status (active / inactive). This should update the url accordingly (adding archived[0]=true / archived[1]=false etc). A chip should appear above the list of companies showing Active / Inactive.

## Screenshots

Screenshot from 2021-06-08 12-39-54

## Checklist

### DIFF
--- a/src/apps/companies/client/filters.js
+++ b/src/apps/companies/client/filters.js
@@ -5,46 +5,44 @@ import {
 import * as labels from './labels'
 import { COMPANY_STATUS_OPTIONS } from './metadata'
 
-export const buildSelectedFilters = (queryParams, metadata) => {
-  return {
-    selectedCountries: buildOptionsFilter({
-      options: metadata.countryOptions,
-      value: queryParams.country,
-      categoryLabel: labels.COUNTRY,
-    }),
-    selectedHeadquarterTypes: buildOptionsFilter({
-      options: metadata.headquarterTypeOptions,
-      value: queryParams.headquarter_type,
-      categoryLabel: labels.HEADQUARTER_TYPE,
-    }),
-    selectedName: buildInputFieldFilter({
-      value: queryParams.name,
-      categoryLabel: labels.COMPANY_NAME,
-    }),
-    selectedSectors: buildOptionsFilter({
-      options: metadata.sectorOptions,
-      value: queryParams.sector_descends,
-      categoryLabel: labels.SECTOR,
-    }),
-    selectedUkRegions: buildOptionsFilter({
-      options: metadata.ukRegionOptions,
-      value: queryParams.uk_region,
-      categoryLabel: labels.UK_REGION,
-    }),
-    selectedCompanyStatuses: buildOptionsFilter({
-      options: COMPANY_STATUS_OPTIONS,
-      value: queryParams.archived,
-      categoryLabel: labels.COMPANY_STATUS,
-    }),
-    selectedExportToCountries: buildOptionsFilter({
-      options: metadata.countryOptions,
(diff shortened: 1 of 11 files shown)


paulgain:
**HTTP 400**
The company search endpoint (same as contacts) only accepts a boolean for `archived`, it doesn’t accept an array of booleans.

Screenshot 2021-06-08 at 13 58 37

src/apps/companies/client/tasks.js
```diff
@@ -11,15 +11,17 @@ import { transformResponseToCompanyCollection } from './transformers'
...
-function getCompanies({ limit = 10, page, ...rest }) {
-  let offset = limit * (parseInt(page, 10) - 1) || 0
+const getCompanies = ({ limit = 10, page, archived, ...rest }) => {
+  const offset = limit * (parseInt(page, 10) - 1) || 0
+  let options = { limit, offset, ...rest }
+
+  // The archived param only needs to be present when a single option is picked
+  if (archived && archived.length === 1) {
+    options.archived = archived[0] === 'true'
+  }
```
paulgain, line 21: It's not the responsibility of the task to transform the data to the API, this should be done in the transformer, here's an example:https://github.com/uktrade/data-hub-frontend/blob/master/src/apps/contacts/client/transformers.js#L14

paulgain (commented):
I've refactored contacts which improves readability and ensures separation of concerns, take a look when you have a moment. We should roll these changes out to companies as well (separate PR is fine).

src/apps/companies/client/tasks.js
```diff
@@ -11,15 +11,12 @@ import { transformResponseToCompanyCollection } from './transformers'
 
 const handleError = (error) => Promise.reject(Error(error.response.data.detail))
 
-function getCompanies({ limit = 10, page, ...rest }) {
-  let offset = limit * (parseInt(page, 10) - 1) || 0
+const getCompanies = ({ limit = 10, page, ...rest }) => {
```
paulgain, line 14: The problem with the `...rest` parameter is that it conceals all the API POST fields making it impossible to read this file.

src/apps/companies/client/filters.js
```diff
@@ -0,0 +1,50 @@
...
+      value: queryParams.export_to_countries,
+      categoryLabel: labels.CURRENTLY_EXPORTING_TO,
+    }),
+    selectedFutureCountriesOfInterest: buildOptionsFilter({
+      options: metadata.countryOptions,
+      value: queryParams.future_interest_countries,
+      categoryLabel: labels.FUTURE_COUNTRIES_OF_INTEREST,
+    }),
+  }
+}
```
paulgain, line 50: No need for the explicit return:
```javascript
export const buildSelectedFilters = (queryParams, metadata) => ({ ... })
```

paulgain (commented):
Just a couple of comments.

paulgain (approved):

src/apps/companies/client/tasks.js
```diff
@@ -31,7 +28,7 @@ function getCompanies({ limit = 10, page, ...rest }) {
  *
  * @returns {promise} - the promise containing a list of options for each category
  */
-function getCompaniesMetadata() {
+const getCompaniesMetadata = () => {
```
ian-leggett, line 31: We can use an implicit return here

PR #3685, opened 2021-06-16
Set default sort and filters for company collection list

tomkingston:
## Description of change

- When you get to the new company collection list page with no filters applied, a default sortby of modified on descending is applied and default status of active (non-archived) is applied.

- You can now sort results by various options

- I have also refactored some of the constants.

- Furthermore, I have fixed some of the sandbox fixtures to ensure that each item rendered by a collection list has a unique id - this stops some warnings about non-unique keys appearing in the console during testing.

**NOTE: I have not yet added tests for this as I want to get these in with the new refactored test format that Paul is currently working on**

## Test instructions

Go to '/companies/react' - when no filters are applied, the list should be automatically filtered by "Active" and sorted by recently updated. 

Applying the sort by filter should sort the results appropriately.

## Checklist

[//]: # "When submitting a PR make sure the code review guidelines have been satisfied.
https://github.com/uktrade/data-hub-frontend/blob/master/docs/Code%20review%20guidelines.md"

- [ ] Has the branch been rebased to master?
- [ ] Automated tests (Any of the following when applicable: Unit, Functional or End-to-End)
- [ ] Manual compatibility testing (Browsers: Chrome, Firefox, IE11, Safari)

PR #3687, opened 2021-06-17
Add sort to the contact collection list

paulgain:
## Description of change
Add sort to the contacts collection list

## Test
`/contacts/react`

## Checklist


src/apps/contacts/client/filters.js
```diff
@@ -33,30 +24,30 @@ export const buildSelectedFilters = (
...
     options: ukRegionOptions,
     value: company_uk_region,
-    categoryLabel: UK_REGION,
+    categoryLabel: 'UK region',
   }),
   selectedStatus: buildOptionsFilter({
     options: statusOptions,
     value: archived,
-    categoryLabel: STATUS,
+    categoryLabel: 'Status',
```
tomkingston, line 51: Although I think we probably could just hardcode the labels - these are actually used these in two places, `ContactsCollection.jsx` and `filters.js` - so, I think it probably is a good idea to put them into a `LABELS` constant to make it easy to chaneg them consistently - I have done that in this PR: https://github.com/uktrade/data-hub-frontend/pull/3685/files

tomkingston (approved):
Looks good - just a comment on the labels

test/functional/cypress/specs/contacts/sort-spec-react.js
```diff
@@ -0,0 +1,112 @@
...
+
+describe('Contact Collections Sort', () => {
+  context('Default sort', () => {
+    const contactsList = contactsListFaker(10)
+    before(() => {
+      cy.intercept('POST', '/api-proxy/v3/search/contact', {
+        body: {
+          count: contactsList.length,
+          results: contactsList,
+        },
```
tomkingston, line 12: Sorry, just one other comment, but do we actually need to mock the response - you are already testing the results in `collection-react-spec.js`

src/apps/contacts/client/filters.js
paulgain: They were originally in as constants, however, I didn't realise that the `ContactsCollection.jsx` list also had the same labels. Thanks.

ian-leggett (approved):

test/functional/cypress/specs/contacts/sort-spec-react.js
paulgain: Yes we do, otherwise the are a few errors in the console during the tests. It's due to our new translator and the sandbox [endpoint](https://github.com/uktrade/data-hub-frontend/blob/master/test/sandbox/routes/v3/search/contact.js) having a bunch of conditionals we're moving away from.

tomkingston: Sounds good, yeah, I had to update some ids in sandbox fixtures to avoid some of those console log errors on mine

ian-leggett (approved):

PR #3688, opened 2021-06-17
Add the Interactions kind filter to the new React interactions collection list

ian-leggett:
## Description of change

Added interaction kind filter to the new interactions React page allowing users to filter out interactions by kind/type

## Test instructions

This should work the same as the current interaction kind filter on the interaction collection list

Go to `/interactions/react`

### Before
Screenshot 2021-06-17 at 12 03 05

## Checklist

### DIFF
--- a/src/apps/interactions/client/InteractionsCollection.jsx
+++ b/src/apps/interactions/client/InteractionsCollection.jsx
@@ -2,8 +2,13 @@ import React from 'react'
 import { connect } from 'react-redux'
 
 import { INTERACTIONS__LOADED } from '../../../client/actions'
+import { INTERACTION, SERVICE_DELIVERY } from './constants'
 
-import { FilteredCollectionList } from '../../../client/components'
+import {
+  FilteredCollectionList,
+  RoutedCheckboxGroupField,
+  CollectionFilters,
+} from '../../../client/components'
 
 import { ID, state2props, TASK_GET_INTERACTIONS_LIST } from './state'
 
@@ -22,17 +27,40 @@ const InteractionCollection = ({
       onSuccessDispatch: INTERACTIONS__LOADED,
     },
   }
-
   return (
     <FilteredCollectionList
       {...props}
-      collectionName="Interaction"
+      collectionName={INTERACTION.label}
       sortOptions={optionMetadata.sortOptions}
       taskProps={collectionListTask}
       selectedFilters={selectedFilters}
       baseDownloadLink="/interactions/export"
       entityName="interactions"
-    ></FilteredCollectionList>
+      defaultQueryParams={{
+        sortby: 'date:desc',
+        page: 1,
+      }}
+    >
+      <CollectionFilters>
+        <RoutedCheckboxGroupField
+          legend="Kind"
+          name="kind"
+          qsParam="kind"
+          options={[
+            {
+              label: INTERACTION.label,
+              value: INTERACTION.value,
+            },
+            {
+              label: SERVICE_DELIVERY.label,
+              value: SERVICE_DELIVERY.value,
+            },
+          ]}
+          selectedOptions={selectedFilters.selectedKind}
+          data-test="status-filter"
+        />
+      </CollectionFilters>
+    </FilteredCollectionList>
   )
 }
 
--- a/src/apps/interactions/client/InteractionsCollection.jsx
+++ b/src/apps/interactions/client/InteractionsCollection.jsx
@@ -30,7 +30,7 @@ const InteractionCollection = ({
   return (
     <FilteredCollectionList
       {...props}
-      collectionName={KIND_OPTIONS.interaction.label}
+      collectionName="interaction"
       sortOptions={optionMetadata.sortOptions}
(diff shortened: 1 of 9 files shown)


src/apps/interactions/client/filters.js
```diff
@@ -0,0 +1,27 @@
+import { KIND_OPTIONS, LABELS } from './constants'
+
+const buildOptionsFilter = ({ options = [], value, categoryLabel }) =>
+  options
+    .filter((option) => value && value.includes(option.value))
+    .map(({ value, label }) => ({
+      value,
+      label,
+      categoryLabel,
+    }))
```
paulgain, line 10: This can be brought in from `src/client/filters.js`

src/apps/interactions/client/InteractionsCollection.jsx
```diff
@@ -22,17 +27,40 @@ const InteractionCollection = ({
       onSuccessDispatch: INTERACTIONS__LOADED,
     },
   }
-
   return (
     <FilteredCollectionList
       {...props}
-      collectionName="Interaction"
+      collectionName={KIND_OPTIONS.interaction.label}
```
tomkingston, line 33: I think just keep this as "Interaction" - the options shouldn't be related to the collection name

test/functional/cypress/specs/interaction/filter-react-spec.js
```diff
@@ -0,0 +1,98 @@
...
+    // Default query params
+    sortby: 'date:desc',
+    ...queryParams,
+  })
+
+const taskParams = {
+  limit: 10,
+  offset: 0,
+  sortby: 'date:desc',
+}
```
paulgain, line 26: I think we agreed on `minimumPayload`

test/functional/cypress/support/assertions.js
```diff
@@ -506,6 +506,16 @@ const assertElementsInOrder = ({ parentElement, expectedIdentifiers }) => {
...
 
+/**
+ * Assert the expected payload to the API
+ */
+
+const assertPayload = (apiRequest, expectedParams) => {
+  cy.wait(apiRequest).then(({ request }) => {
+    expect(request.body).to.deep.equal(expectedParams)
+  })
+}
```
paulgain, line 517: Nice 👍

src/apps/interactions/client/constants.js
```diff
@@ -0,0 +1,14 @@
...
+export const KIND_OPTIONS = {
+  interaction: {
+    label: 'Interaction',
+    value: 'interaction',
+  },
+  serviceDelivery: {
+    label: 'Service delivery',
+    value: 'service_delivery',
+  },
+}
```
tomkingston, line 14: I think you could just define this as an array as shown below. This would also mean the order would be consistent when the constant is used.

```
export const KIND_OPTIONS = [
    { label: 'Interaction', value: 'interaction' },
    { label: 'Service Delivery', value: 'service_delivery' }  
]
```

https://github.com/uktrade/data-hub-frontend/pull/3685/files#diff-4349e0cd10327a15439d865a8f6661bb5f174c93fbd86a1fec69691d1f44d6d6R14-R19

paulgain (approved):
Just a few comments

src/apps/interactions/client/state.js
```diff
@@ -2,25 +2,29 @@ import { omitBy, isEmpty } from 'lodash'
...
+
 export const ID = 'interactionsList'
 
-const getFilteredQueryParams = (router) => {
-  const queryParams = router.location.search.slice(1)
-  const filteredQueryParams = omitBy({ ...qs.parse(queryParams) }, isEmpty)
+import { buildSelectedFilters } from './filters'
+
+const parseQueryString = (queryString) => {
+  const queryStringObject = omitBy({ ...qs.parse(queryString) }, isEmpty)
```
tomkingston, line 11: I know this is in the existing contacts page, but it's nice if we can avoid suffixing variable names with type names - I think this is actually just `queryProps`

src/apps/interactions/client/transformers.js
```diff
@@ -26,8 +28,8 @@ const formatParticipants = (dit_participants) =>
 
 const getbadgeLabel = (type, hasFeedback = false) => {
   const badges = {
-    interaction: 'Interaction',
-    service_delivery: 'Service delivery',
+    interaction: KIND_OPTIONS.interaction.label,
+    service_delivery: KIND_OPTIONS.serviceDelivery.label,
```
tomkingston, line 32: I think this might not be needed... but would need to investigate

tomkingston: Actually - I expect it is needed... I just don't think you should have to navigate the options object to get here - perhaps you could define the two options as LABELS constants before putting them into the KIND_OPTIONS array

test/functional/cypress/specs/interaction/filter-react-spec.js
ian-leggett: Good spot changed 👍

tomkingston (approved):

PR #3711, opened 2021-06-21
Refactor contact form to react

peterhudec:
## Description of change

Refactors the add/edit contact page from Nunjucks to React. Along the way, I have:

- Tweaked the `Form/elements/FieldAddress` component so that it conditionallyrenders a dropdown for its country field
- Fixed an issue in the `ReferrerLink` component, which was breaking some tests
- Added the `superHeading` option to the `LocalHeader` component
- Introduced the `State` component
- Introduced the `Resource` component together with its `Resource/Company`, `Resource/Contact` and `Resource/Countries` derivations two of which are used in the `ContactForm`
- Introduced the `Form/elements/FieldCountrySelect` which uses `Resource/Company`
- Introduced the `email` form validator in `src/client/components/Form/validators.js`
- Added the `idProp` parameter to `multiInstance` so that the `Resource` component could be implemented with it
- Added the `checkRadioGroup` and `getSubmitButtonByLabel` Cypress commands (we are not using Cypress Testing Library)

I realize that this PR is way beyond the 400 LOC limit, but the only other way would be way to introduce the new components in separate PRs without actually being used anywhere, which would raise a lot of questions and require lenghty explanation. But should this be really hard to review, it can still be broken up into smaller PRs.

## Test instructions

Nothing should change by this pr. The add/edit contact form should work as before.


src/client/components/ContactForm/index.jsx
```diff
@@ -0,0 +1,353 @@
...
+              acceptsDitEmailMarketing,
+              addressSameAsCompany,
+              primary,
+              ...values
+            }) => {
+              const payload = {
+                ..._.mapKeys(values, (v, k) => _.snakeCase(k)),
+                // The checkbox's value is for some odd reason at index 1
+                accepts_dit_email_marketing:
+                  acceptsDitEmailMarketing[1] === 'yes',
```
faissalMT, line 100: What's at index 0? It's probably worth figuring out why this happens in case it affects functionality here

src/client/components/ContactForm/index.jsx
```diff
@@ -0,0 +1,353 @@
...
+              county,
+              country,
+              postcode,
+              acceptsDitEmailMarketing,
+              addressSameAsCompany,
+              primary,
+              ...values
+            }) => {
+              const payload = {
+                ..._.mapKeys(values, (v, k) => _.snakeCase(k)),
```
faissalMT, line 97: It's not immediately obvious what this is doing, so it might be worth putting it into a function

src/client/components/ContactForm/index.jsx
```diff
@@ -0,0 +1,353 @@
...
+                // The checkbox's value is for some odd reason at index 1
+                accepts_dit_email_marketing:
+                  acceptsDitEmailMarketing[1] === 'yes',
+                primary: primary === 'yes',
+                company,
+                address_same_as_company: addressSameAsCompany === 'yes',
+                // The API is complaining if we send the address fields when address_same_as_company is true
+                ...(addressSameAsCompany !== 'yes' && {
```
faissalMT, line 105: I've never thought of doing this before, I like it

src/client/components/ContactForm/index.jsx
```diff
@@ -0,0 +1,353 @@
...
+                  validate={validators.email}
+                />
+                <FieldCheckboxes
+                  name="acceptsDitEmailMarketing"
+                  options={[
+                    {
+                      value: 'yes',
+                      label: 'The company contact does accept email marketing',
+                      hint:
+                        !!values.accepts_dit_email_marketing?.length &&
```
faissalMT, line 218: Is the coercion needed here?

src/client/components/Form/elements/FieldAddress/index.jsx
```diff
@@ -235,24 +237,30 @@ const FieldAddress = ({
...
   hint: PropTypes.node,
+  apiEndpoint: PropTypes.string.isRequired,
+  onSelectUKAddress: PropTypes.func,
+  isCountrySelectable: PropTypes.any,
+  // Country is only required if isCountrySelectable is falsy, but this can't be
+  // expressed with propTypes
   country: PropTypes.shape({
     id: PropTypes.string.isRequired,
     name: PropTypes.string.isRequired,
   }).isRequired,
```
faissalMT, line 263: If country is conditionally required and there's no way to express that it's only required in certain circumstances I would remove the `.isRequired` here just to avoid confusing warnings

test/functional/cypress/specs/interaction/details-form-spec.js
```diff
@@ -812,10 +812,9 @@ describe('Contact loop', () => {
         .find('input')
         .should('have.attr', 'value', 'Test if values is restored')
 
-      cy.get('.c-message--success').should(
-        'contain',
-        'You added Json Russel.You can now continue recording the interaction.'
-      )
+      // We are not expecting John Doe here, because the mocked sandbox response
+      // returns Json Russel
```
faissalMT, line 816: There's a few comments in this PR - like this one - that I presume are WIP comments?

src/client/components/ContactForm/index.jsx
peterhudec: Ok, that was actually caused by me, when I was always wrapping the inital value in an array. It's fixed now.

src/client/components/ContactForm/index.jsx
peterhudec: Yes, React renders numbers and ignores booleans, so if we didn't coerce here, it would render the digit 0.

test/functional/cypress/specs/interaction/details-form-spec.js
peterhudec: No, it's there to explai why we are expecting "Jonh Russel" in the message even though we filled out the form with "John Doe" a couple of lines above.

paulgain:
The validation is not quite right, this is a great opportunity to correct the previous behaviour. Rather than informing the user "This field may not be null" can we come up with something better? 

null


Here's an example from the Design System: https://design-system.service.gov.uk/components/error-summary. In addition, take a look at how we did the 'add a company' user journey: `/companies/create`

better

peterhudec:
@paulgain we plan to improve the messages in later PRs. This one should be a pure refactoring from Nunjucks to React and as such it should avoid changing behaviour if possible. In fact we have changed the wording of the flash messages, but only because the original messages would require the form to keep track of where the user came from.

paulgain (commented):
LGTM, just a couple of comments.

src/client/components/ContactForm/index.jsx
```diff
@@ -0,0 +1,356 @@
...
+              acceptsDitEmailMarketing,
+              addressSameAsCompany,
+              primary,
+              ...values
+            }) => {
+              const payload = {
+                ...keysToSnakeCase(values),
+                accepts_dit_email_marketing:
+                  acceptsDitEmailMarketing[0] === 'yes',
+                primary: primary === 'yes',
```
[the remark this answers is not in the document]
peterhudec: @paulgain thanks for the feedback.

1. I agree, but unfortunately both the `FieldCheckboxes` and `FieldRadios` require the value to be a string so it can't be boolean. This is because we use use two radio buttons instead of a single checkbox for binary questions and because we don't have a `FieldCheckbox` field, only a `FieldCheckboxes`.
2. That's a good point, I will do it.
3. That didn't occur to me and it actually seems to work, but only with `addressSameAsCompany` and `primary` and not with `accepts_dit_email_marketing`, which must be boolean. Also, it only works one way -- the API sends all of them back as booleans so it still needs to be converted to 'yes' and 'no'.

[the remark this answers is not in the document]
paulgain: @peterhudec I feel your pain.

faissalMT (approved):
Just two minor comments, everything else looks good to me

src/client/components/ContactForm/index.jsx
```diff
@@ -0,0 +1,357 @@
...
+                    {
+                      value: NO,
+                      label: NO,
+                      children: (
+                        <fieldset>
+                          <legend>Contact address</legend>
+                          <FieldAddress
+                            apiEndpoint="/api/postcodelookup"
+                            isCountrySelectable={true}
+                            forcePostcodeLookupButton={true}
```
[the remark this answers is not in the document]
peterhudec: The `FieldAddress` component only shows the postcode lookup button when the country is UK. But in this case, we want to always have the button, because this is how it is in the form I'm rewriting to React and I don't want to introduce any UX changes. So I introduced this flag which when truthy, will force the postcode lookup button to always be rendered, regardless of the selected country.

ian-leggett (approved):
After our discussion I understand why the PR is so large. To me this PR does 2 things:

1. Refactors a contact form
2. Introduces a new component which wraps the task component giving us a higher level to push/pull data without all the boilerplate

Ideally this should be split into at least two PR's but I understand its hard to build the second without having a use case. All good from me but we must share this new component with other devs so we can all adopt this pattern going forward.

PR #3724, opened 2021-06-25
Event type filter

tomkingston:
## Description of change

Adds an event type filter to the new react event collection page

## Test instructions

It should now be possible to filter by event type - the event types should be populated as a checkbox group. Event type chips should appear above and the url should include the selected event_types

## Screenshots

Screenshot from 2021-06-25 11-48-30

## Checklist


test/functional/cypress/fakers/event-types.js
```diff
@@ -0,0 +1,22 @@
+import faker from 'faker'
+
+import { listFaker } from './utils'
+
+/**
+ * Generate fake data for an event type.
+ */
+const eventTypeFaker = (overrides = {}) => ({
+  id: faker.datatype.uuid(),
+  name: faker.lorem.words(),
```
paulgain, line 10: Why not mock the actual event names and then randomly select one? If we're going to mock, it should at least look like the real data.

test/functional/cypress/specs/events/filter-react-spec.js
```diff
@@ -211,26 +219,81 @@ describe('events Collections Filter', () => {
     })
   })
 
+  context('Event Type', () => {
+    const element = '[data-test="event-type-filter"] fieldset'
+    const { id: eventTypeId, name: eventTypeLabel } = randomChoice(eventTypes)
```
paulgain, line 224: The renaming of variables makes your code harder to read, it's a level of indirection that's not required. We should only take this approach when there is a name clash.

test/functional/cypress/specs/events/filter-react-spec.js
```diff
@@ -211,26 +219,81 @@ describe('events Collections Filter', () => {
...
+      })
+    })
+  })
+
   context('Remove all filters', () => {
     before(() => {
       const ukCountryId = '80756b9a-5d95-e211-a939-e4115bead28a'
       const adviserId = 'e83a608e-84a4-11e6-ae22-56b6b6499611'
       const southEastRegionId = '884cd12a-6095-e211-a939-e4115bead28a'
+      const { id: eventTypeId } = randomChoice(eventTypes)
```
paulgain, line 276: Same as above.

test/functional/cypress/specs/events/filter-react-spec.js
tomkingston: Personally I find this easier to read - id is used quite a bit lower down in the code and it can be quite difficult to understand that `id` is relating to an eventType once you get that far... perhaps it would be best if I didn't expand the object and went with `const eventType = randomChoice(eventTypes)` and then use `eventType.id` lower down?

test/functional/cypress/fakers/event-types.js
tomkingston: Can do

tomkingston: I think if this is the case, we might as well just use a fixture rather than a faker though

test/functional/cypress/specs/events/filter-react-spec.js
paulgain: I just find it odd to rename `id` to `eventTypeId ` when the context is `Event Type`.
Yeah, that sounds like a sensible approach, although wouldn't it be just `const event = ...` I'm not sure what the `Type` suffix adds? It seems the event label sitting above the checkboxes has had an influence.

paulgain (approved):

ian-leggett (approved):

PR #3747, opened 2021-06-29
Add address areas to parseAddress()

faissalMT:
## Description of change

Some functionality that was taking advantage of `parseAddress()` was missing area functionality, meaning that unverified functionality was missing areas. This adds the feature flag `address-area-unverifed-match` to control this behaviour.

What's been marked as having not been covered by tests is due to this being behind a feature flag, since I don't know of a way to change feature flags during Cypress runtime. The functionality when the feature flag is enabled is tested however.

## Test instructions

When doing company matching for a US company the state should appear on every page under 'located at', likewise with a Canadian company the same should happen with the province.

### Before

Screenie_2
Screenie_4

### After

Screenie_1
Screenie_3

## Checklist


mforner13 (approved):
Looks good 👍

src/apps/companies/apps/match-company/controllers.js
```diff
@@ -49,7 +57,12 @@ function getCountryCode(company, countries) {
   )
 }
 
-function parseAddress(dnbCompany, countries, prefix = '') {
+function parseAddress(
+  dnbCompany,
+  countries,
+  prefix = '',
+  isAddressAreaEnabled = false
+) {
```
tomkingston, line 65: This function signature is getting quite big now - I wonder whether it would be worth encapsulating the input arguments as an object so it is a bit clearer what each argument means when you call it, at the moment, you have to refer back to the function to know what the empty string for the third argument is when you do this:
```
            parseAddress(
              company.address,
              countries,
              '',
              isAddressAreaEnabled
            ),
```

With an object you wouldn't even need to pass the prefix since a blank string is already the default, so:
```
            parseAddress({
              dnbCompany: company.address,
              countries,           
              isAddressAreaEnabled
            })
```

or alternatively, something like this:

```
            parseAddress(company.address, {
              countries,           
              isAddressAreaEnabled
            })
```

This would involve updating the relevant function calls across the project though, so only do it if you think it's worth it.

tomkingston (approved):
Looks good - I've added a possible suggestion, but up to you on what to do with it

PR #3781, opened 2021-07-07
Testing library activity feed

vfarah-if:
## Description of change

This PR

Rewrites the Activity feed functional tests using Testing library commands
Improves the markup structures for better accessibility

## Test instructions

_What should I see?_

### Before

_Add a screenshot_

### After

_Add a screenshot_

## Checklist


test/functional/cypress/specs/companies/activity-feed-spec.js
```diff
@@ -27,36 +15,19 @@ describe('Company activity feed', () => {
...
 
     it('should display "Refer this company" button', () => {
-      cy.get(
-        selectors
-          .companyCollection()
-          .interaction.referButton(fixtures.company.venusLtd.id)
-      ).should('have.text', 'Refer this company')
-    })
-
-    it('should not display the activity feed', () => {
```
vfarah-if, line 52: removed this as the test below covers it implicitly by virtue of that message showing

test/functional/cypress/specs/companies/activity-feed-spec.js
```diff
@@ -240,37 +193,17 @@ describe('Company activity feed', () => {
...
-      cy.get('@maxemailCampaign')
-        .find('h3')
-        .eq(1)
-        .should('have.text', 'British Business Network Update - December 2020')
-
-      cy.get('@maxemailCampaign')
-        .find('ul')
-        .first()
-        .should('have.text', '15 Dec 2020')
+      cy.findAllByRole('heading', {
```
vfarah-if, line 196: This may seem hacky but as this received multiple responses, this seemed the only way to make sure it doesn't fail, which is the only bad side about not querying a specific selector path

test/functional/cypress/specs/companies/activity-feed-spec.js
```diff
@@ -66,44 +37,34 @@ describe('Company activity feed', () => {
...
     it('should not display the "Add interaction" button', () => {
-      cy.contains('Add interaction').should('not.exist')
+      cy.findByRole('button', { name: /Add interaction/i }).should('not.exist')
     })
 
     it('should not display "Refer this company" button', () => {
-      cy.contains('Refer this company').should('not.exist')
-    })
-
-    it('should display the activity feed', () => {
```
vfarah-if, line 97: Removed because test below does the same implicitly

src/client/components/ActivityFeed/activities/card/CardHeader.jsx
```diff
@@ -106,20 +106,24 @@ const CardHeader = ({
...
 
-      <StyledMetaItems>
-        {startTime && <ListItem>{format(startTime)}</ListItem>}
+      <StyledMetaItems role="list">
+        {startTime && <ListItem role="listitem">{format(startTime)}</ListItem>}
 
         {badge && (
-          <ListItem>
+          <ListItem role="listitem">
             <Badge borderColour={badge.borderColour}>{badge.text}</Badge>
```
vfarah-if, line 127: Important to note that lists where the biggest issue area for me in the sense that searching by semantically implied role, actual configured role didn't work most of the time and it seemed mostly because the lists were nested in other lists or divs maybe muddying the water and even though I could see a lovely trail of names in the cypress log, I found it difficult to understand how to fix this so it would always reliably come back with the list item I was searching for

LaurenQurashi (approved):

PR #3782, opened 2021-07-07
Add scrollable checkbox options to Storybook

ian-leggett:
## Description of change
As we are building out the new React collection lists we noticed that the long lists of checkbox options you get on the likes of the interactions list page looks strange in comparison to the old component. This is because the new design has no obvious container around the options and no fixed scroll so its hard to just by looking at it that you can scroll the options.

The changes we have made are an iteration towards this example you can see on the Gov.uk website - https://www.gov.uk/search/all?keywords=cars&order=relevance in the next iteration we look at adding the input field so the user can search (similar to how the typeahead behaves)
Screenshot 2021-07-07 at 17 03 45

For now this does the following:

1. You can now specify a scrollable area via the new property `visibleHeight`. This enables the developer to customise how large/high the container is before it overflows with a scroll. As labels can vary in size this is very hard to set a height that suits all scenarios so setting a height is the best option for now.
2. The scroll will always visible
3. As the user clicks on the checkboxes a count will appear (scrollable version only)

## Test instructions
1. Run Storybook and look at the "Overflow with scroll" option for "CheckboxGroupField"

## Screenshots
Screenshot 2021-07-07 at 16 52 39

_Add a screenshot_

## Checklist

### DIFF
--- a/src/client/components/CheckboxGroupField/index.jsx
+++ b/src/client/components/CheckboxGroupField/index.jsx
@@ -1,8 +1,9 @@
 import React, { useState, useEffect } from 'react'
 import MultiChoice from '@govuk-react/multi-choice'
+import PropTypes from 'prop-types'
 import { GREY_2, YELLOW } from 'govuk-colours'
 import styled, { css } from 'styled-components'
-import { FONT_SIZE } from '@govuk-react/constants'
+import { FONT_SIZE, SPACING } from '@govuk-react/constants'
 
 import Checkbox from '../Checkbox'
 import FieldWrapper from '../Form/elements/FieldWrapper'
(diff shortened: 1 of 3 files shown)


src/client/components/CheckboxGroupField/index.jsx
```diff
@@ -74,6 +105,7 @@ const CheckboxGroupField = ({
   selectedOptions = [],
   onChange = () => null,
   id,
+  visibleHeight = 0,
```
paulgain, line 108: 1. It would be good if we could add some `propTypes` for `CheckboxGroupField`.
2. Can we add a comment above the function as to what `visibleHeight` actually does:
    `@param {number} visibleHeight - some description`.
3. I thought we all agreed on `maxScrollHeight`?

src/client/components/CheckboxGroupField/index.jsx
```diff
@@ -54,6 +49,42 @@ const StyledFieldWrapper = styled(FieldWrapper)`
...
+        }
+      }
+      ${checkboxGroupElementStyles}`
+      : checkboxGroupElementStyles}
+`
+
+const SelectedCount = styled('span')`
+  font-size: ${FONT_SIZE.SIZE_14};
+  display: block;
+  padding: 5px 0;
```
tomkingston, line 79: Is there a const that should be used instead of 5px?

tomkingston (approved):

src/client/components/CheckboxGroupField/index.jsx
ian-leggett: Points 1 - 3 all done and sorry I forgot we spoke about naming it `maxScrollHeight ` all changed 👍

paulgain (approved):

PR #3797, opened 2021-07-12
Move all date transformations into single date util file

cgsunkel:
## Description of change

There are currently numerous date transformations happening all over the place in the FE, and it was agreed that all of these should be moved into a centralised location.

As part of this work I've created several new date utilities and refactored several old ones. Note that many of the dates will only transform correctly under very specific circumstances so several specific formatting/validation functions were created to handle these cases. I've also updated the list of formats in `constants` in order to accomodate the constants defined in `date-utils`.

Most of the 80 files changed are just updating imports and replacing `date-fns` functions with their equivalent in the new `util` file.

## Test instructions

Pages with dates should render as normal.

## Checklist


src/client/utils/date-utils.js
```diff
@@ -25,24 +27,33 @@ const normaliseAndFormatDate = (year, month, day) => {
   return day ? `${yearAndMonth}-${padZero(day)}` : yearAndMonth
 }
 
+export const isDateValid = (date) => {
+  return isValid(parseISO(date))
+}
+
 export const format = (dateStr, dateFormat = 'dd MMM yyyy') =>
-  dateStr ? formatFNS(parseISO(dateStr), dateFormat) : null
+  isDateValid(dateStr) ? formatFNS(parseISO(dateStr), dateFormat) : null
```
cgsunkel, line 35: This should help to reduce the number of `date-fns`-related Sentry errors

src/apps/investments/transformers/project.js
```diff
@@ -111,8 +111,9 @@ function transformFromApi(body) {
   const estimatedLandDate = body.estimated_land_date
   if (!isEmpty(estimatedLandDate)) {
     const date = new Date(estimatedLandDate)
+    const month = date.getMonth() + 1
```
cgsunkel, line 114: `date-fns` would not accept this date as being valid, so as the year is extracted using native date methods I've used them for the month as well

src/client/utils/date-utils.js
```diff
@@ -25,24 +27,33 @@ const normaliseAndFormatDate = (year, month, day) => {
...
+  return format(new Date())
 }
 
 export const formatWithTime = (dateTimeStr) => {
-  return formatFNS(parseISO(dateTimeStr), 'd MMM yyyy, h:mmaaa')
+  return format(dateTimeStr, 'd MMM yyyy, h:mmaaa')
 }
 
-export const isDateValid = (year, month, day, format = DATE_FORMAT_LONG) => {
+export const isNormalisedDateValid = (
```
cgsunkel, line 45: This was renamed as the majority of our date validations don't need to be formatted first

src/common/date.js
```diff
@@ -1,30 +1,45 @@
...
-function formatLongDate(dateString = []) {
-  if (dateString) {
-    return format(parseDateString(dateString), DATE_LONG_FORMAT)
-  }
+function isDateValid(date) {
+  return isValid(parseISO(date))
+}
 
-  return null
+function isUnparsedDateValid(date) {
```
cgsunkel, line 25: Sometimes `date-fns` rejects valid dates if `parseISO` is used. This function was created in order to meet these edge cases

src/common/date.js
```diff
@@ -1,29 +1,51 @@
...
+  return formatDate(dateString, DATE_MEDIUM_FORMAT)
 }
 
 function formatMediumDateTime(dateString) {
-  if (dateString) {
-    return format(parseDateString(dateString), DATE_TIME_MEDIUM_FORMAT)
+  return formatDate(dateString, DATE_TIME_MEDIUM_FORMAT)
+}
+
+function formatDateWithoutParsing(date, dateFormat, parseISOInFormat = true) {
```
faissalMT, line 44: It might be worth splitting out the `parseISOInFormat` functionality into another function, this reads to me like a function that's doing two slightly different things

paulgain:
@cgsunkel I think this is a great opportunity to have a **single** date file, having multiple date files is confusing.

If we move all the date functions into a file called `/src/client/utils/date.js` and export each function in the CommonJS format we can simply `require/import` those functions whether you're working server-side or client-side.

`/src/client/utils/date.js`
```javascript
function today() { .. }

module.exports = {
  today,
}
```

**Server (Node/CommonJS)**
This may look odd but over time it's all going anyway!
```javascript
const { today } = require('./client/utils/date.js')
```

**Client (React/ES6)**
```javascript
const { today } = from './client/utils/date.js'
```

I feel this is a far simpler approach, there's no duplication of functions and a centralised file for all dates. As already mentioned the node code will move to the browser over time, at that point we should convert all the functions to be ES6 modules with the benefit of not touching any of the calling code.

**Once everything is in the browser**
`/src/client/utils/date.js`
```javascript
// We go from this:
function today() { .. }
// to this
export const today = () => { .. }
```

Finally, the `/src/client/utils/date.js` file should house generic functions only, any functions like `isDNBChangeRequestValid` should be moved out to the area it's called from.

cgsunkel:
> @cgsunkel I think this is a great opportunity to have a **single** date file, having multiple date files is confusing.

@paulgain I've now combined the two files, let me know if you're happy with it

faissalMT (approved):

paulgain (approved):
@cgsunkel this is looking a lot cleaner, great job 👏 

**Can we add a comment at the top of the file to inform others what the strategy/thinking is:** 
1. Why we’re exporting these functions as CommonJS modules
2. Mention the date functions are used in both the client-side and server-side code
3. Mention it allows us to house all the dates in a single file removing any duplication and confusion of multiple date files
4. Mention all exported functions within `date.js` will eventually be converted to ES6 modules

PR #3798, opened 2021-07-13
Add state and province filters to company collection list

tomkingston:
## Description of change

Adds state and province filters to new react company collection list

## Test instructions

Go to /companies/react - there should be new filters for US states and Canadian Provinces. Selecting these should update the url and show a chip.

## Checklist

### DIFF
--- a/src/apps/companies/client/constants.js
+++ b/src/apps/companies/client/constants.js
@@ -3,6 +3,8 @@ export const LABELS = {
   companyName: 'Company name',
   sector: 'Sector',
   country: 'Country',
+  usState: 'US state',
+  canadianProvince: 'Canadian province',
   ukPostcode: 'UK postcode',
   ukRegion: 'UK region',
   companyStatus: 'Status',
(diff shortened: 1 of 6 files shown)


src/apps/companies/client/constants.js
```diff
@@ -34,3 +36,8 @@ export const SORT_OPTIONS = [
     value: 'latest_interaction_date:desc',
   },
 ]
+
```
tomkingston, line 39: I wasn't sure how to deal with these country ids - they are needed to filter the administrative areas (they are also used to filter on the nunjucks implementation). Doesn't seem ideal having them hardcoded but I think theyt are needed. 
I was also wondering whether I should store these in a more generic file location so that other pages / components could use them.

paulgain: I don't see them being imported anywhere else other than the task, therefore, I would be inclined to move them both to the task.

paulgain (approved):

ccp92 (approved):

PR #3799, opened 2021-07-14
Filter the service metadata correctly

ian-leggett:
## Description of change

Currently the services in the new react interactions collection list filter are not being filtered correctly as per legacy version. By default `filterDisabled: true` gets passed to the API as default and we needed to override this to `false` in order to get the right list back.

## Test instructions

1. Go to `interactions/react`
2. Compare the list of options in the services filter against the options in the live version `/interaction`

They should be exactly the same.

## Checklist

### DIFF
--- a/src/apps/interactions/client/transformers.js
+++ b/src/apps/interactions/client/transformers.js
@@ -86,29 +86,21 @@ export const transformWasPolicyfeedBackProvidedToApi = (
 export const filterServiceNames = (services) => {
   if (!services) return
 
-  const excludedServiceStrings = [
+  const excludedParentServices = [
     'A Specific DIT Export Service or Funding',
     'A Specific Service',
     'Enquiry or Referral Received',
     'Enquiry Received',
   ]
-
   const filteredServiceNames = services
     .map((service) => {
-      const splitServiceName = service.label.split(' : ')
-      const name =
-        splitServiceName[1] &&
-        excludedServiceStrings.includes(splitServiceName[0])
-          ? splitServiceName[1]
-          : service.label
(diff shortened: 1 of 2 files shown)


src/apps/interactions/client/transformers.js
```diff
@@ -82,3 +82,33 @@ export const transformResponseToCollection = ({ count, results = [] }) => ({
...
+    .map((service) => {
+      const splitServiceName = service.label.split(' : ')
+      const name =
+        splitServiceName[1] &&
+        excludedServiceStrings.includes(splitServiceName[0])
+          ? splitServiceName[1]
+          : service.label
+      return { ...service, label: name }
+    })
+    .sort(function (a, b) {
```
ccp92, line 106: Could this be an arrow function to match the style used in the map?

ccp92 (approved):

src/apps/interactions/client/transformers.js
paulgain: I find this quite difficult to read, we could break it down a bit so it's easier to read:
```javascript
  const excludedParentServices = [
    'A Specific DIT Export Service or Funding',
    'A Specific Service',
    'Enquiry or Referral Received',
    'Enquiry Received',
  ]

  const filteredServiceNames = services
    .map((service) => {
      const [parent, child] = service.label.split(' : ')
      const isParentExcluded = excludedParentServices.includes(parent)
      const label = child && isParentExcluded ? child : service.label
      return { ...service, label }
    })
    .sort((a, b) => a.label.localeCompare(b.label))
```

paulgain (approved):

PR #3800, opened 2021-07-14
Resolve issue with unverified companies not saving with undefined area

ccp92:
## Description of change

Resolves an issue where unverified companies with no address area were failing to save correctly.

## Test instructions

Go to add an unverified company that is not located in the united states or Canada. The company should save correctly without and address area (i.e. State or Province) correctly.

### Before

n/a

### After

n/a

## Checklist

### DIFF
--- a/src/apps/companies/apps/add-company/transformers.js
+++ b/src/apps/companies/apps/add-company/transformers.js
@@ -62,4 +62,5 @@ const transformToCreateDnbCompanyInvestigation = (formData, companyId, res) => {
 module.exports = {
   transformToDnbStubCompany,
   transformToCreateDnbCompanyInvestigation,
+  transformFormData,
 }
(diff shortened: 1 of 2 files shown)


ian-leggett (approved):

src/apps/companies/apps/add-company/transformers.js
```diff
@@ -15,6 +15,17 @@ const transformFormData = (
   res,
   countryAsObject = true
 ) => {
+  const sanitiseArea = (area) => {
```
mforner13, line 18: Could use a ternary operator here instead which would make it slightly shorter

```
  const sanitiseArea = (area) => {
    area === undefined
      ? {
          area: undefined,
        }
      : {
          area: { id: area },
        }
  }
  ```

But doesn't really make it any clearer so up to you :)

mforner13 (approved):

src/apps/companies/apps/add-company/transformers.js
ccp92: Yeah, had the same train of thought but think that the if block read a little more clearly.

PR #3802, opened 2021-07-16
Create large capital oppprtunity requirements form

mforner13:
## Description of change

This PR adds the ability for users to edit the requirements of a large capital investment opportunity. Currently, these details are displayed in a table but there is no functionality to edit. 

For the form, the `<MulitiInstanceForm>` component is used, which is rendered once the appropriate metadata has been collected to populate the form field options - for example, the investment types.

As there weren't previously any tests around the requirements table, this PR also includes testing that the correct data is displayed for when requirements are both complete and incomplete.

**TO NOTE**
Currently the large capital opportunity endpoint `v4/large-capital-opportunity/${opportunityId}` `investment_types` field returns an array, although there can only be one investment type per opportunity. Once this has been amended on the API, the structure of this field will be refactored in the form's initial value and in the task payload.

## Test instructions

- Go to `investments/opportunities`
- Pick any opportunity
- Click the 'Edit' button in the requirements section
- Fill out the form (no fields are mandatory)

### Before

Screenshot 2021-07-16 at 12 55 52

### After

**Table view**

Screenshot 2021-07-16 at 12 54 40

**Form view**

screenshot-localhost_3000-2021 07 16-12_53_52

## Checklist


ccp92 (approved):

elcct (approved):

chloehkwong1 (approved):
Looks good! 👍🏼  Just one thing - would you typically not feature flag something like this? And if not, then when would you put in a feature flag? Thanks :)

mforner13:
> Looks good! 👍🏼 Just one thing - would you typically not feature flag something like this? And if not, then when would you put in a feature flag? Thanks :)

I don't think we'd feature flag this at the form can be used by users straight away. I think the team's approach has been very iterative, so first display the tables of information, then give users the ability to edit the information. This way value is brought to the users quicker, rather than if we hid everything behind a feature flag and released it all together. :)

chloehkwong1:
>  I don't think we'd feature flag this as the form can be used by users straight away. I think the team's approach has been very iterative, so first display the tables of information, then give users the ability to edit the information. This way value is brought to the users quicker, rather than if we hid everything behind a feature flag and released it all together. :)

That makes sense, thanks for explaining :)

PR #3812, opened 2021-07-19
Swap out nunjucks interaction collection list for react version

ian-leggett:
## Description of change

This removes the nunjucks version of the interactions collection list and replaces it with the react version. 

Note: you will see commented out code, the reason for us doing this is just incase we need to roll it back easily. After a week we will re-visit the code and if the users are happy we will remove the comments, delete files and unit tests that are no longer needed.

## Test instructions

1. Visit `/interactions` all filters and functionality should work as it did with legacy version

## Checklist


src/apps/interactions/router.js
```diff
@@ -1,38 +1,36 @@
...
-const urls = require('../../lib/urls')
-
 const {
   DEFAULT_COLLECTION_QUERY,
   APP_PERMISSIONS,
   QUERY_FIELDS,
   QUERY_DATE_FIELDS,
 } = require('./constants')
-const { renderInteractionList } = require('./controllers/list')
+// const { renderInteractionList } = require('./controllers/list')
```
elcct, line 9: Maybe good to add a `TODO: ` comment so other devs are aware of why this is commented out and when should be removed.

elcct (approved):

paulgain:
@ian-leggett looks good 😊

There are only two small issues I've come across, this shouldn't prevent this form going live today.

**Teams**
When the user initially inserts their cursor into the **Teams** textfield it would be good to fetch the first x items and display them in the dropdown. Currently the user has to either a) enter a space or b) start typing characters before they see a subset of teams.

**Company One List Group Tier**
When selecting one of the tiers the checkboxes jump, not a problem if you're only selecting one tier, however, if you try to select two tiers close together you miss the second one due to the jump. The jump occurs due to the **x selected** being inserted into the DOM.

paulgain (approved):

PR #3816, opened 2021-07-20
CDS-234 Fix validation for US and Canadian area field

chloehkwong1:
## Description of change

Previously, the add company and edit company forms were not validating against the state/province fields when they were present.

## Test instructions

Go to Add a new company to US/Canada. Fill in all address fields apart from State/Province and attempt to submit, you should receive an error asking you to fill in that field.

Go to Edit an existing US/Canadian company's business details. Empty the State/Province field and attempt to submit, you should receive an error asking you to fill in that field.

### Before

image

### After

image

## Checklist

### DIFF
--- a/test/functional/cypress/specs/companies/add-company-spec.js
+++ b/test/functional/cypress/specs/companies/add-company-spec.js
@@ -96,6 +96,46 @@ const goToUKCompanySectorAndRegionPage = () => {
   cy.get(selectors.companyAdd.continueButton).click()
 }
 
+const manualAddUSOrCanadianCompanyForm = () => {
+  cy.get(selectors.companyAdd.newCompanyRecordForm.organisationType.charity)
+    .parent()
+    .should('be.visible')
+  cy.get(
+    selectors.companyAdd.newCompanyRecordForm.organisationType
+      .governmentDepartmentOrOtherPublicBody
+  )
+    .parent()
+    .should('be.visible')
+  cy.get(
+    selectors.companyAdd.newCompanyRecordForm.organisationType.limitedCompany
+  )
+    .parent()
+    .should('be.visible')
+  cy.get(
+    selectors.companyAdd.newCompanyRecordForm.organisationType
+      .limitedPartnership
+  )
+    .parent()
+    .should('be.visible')
+  cy.get(selectors.companyAdd.newCompanyRecordForm.organisationType.partnership)
+    .parent()
+    .should('be.visible')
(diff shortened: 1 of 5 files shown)


test/functional/cypress/specs/companies/add-company-spec.js
```diff
@@ -801,5 +742,90 @@ describe('Add company form', () => {
...
+    before(() => {
+      goToCandianCompanySearchResultsPage()
+
+      cy.get(selectors.companyAdd.entitySearch.cannotFind.summary).click()
+      cy.get(
+        selectors.companyAdd.entitySearch.cannotFind.stillCannotFind
+      ).click()
+    })
+
+    it('should display the manual entry form', () => {
```
mforner13, line 769: If this is essentially the same as the manual entry form test when adding a new US company (line 696), I'd say either:

1. Can the common fields be extracted into an `assertCommonFormFields` const (doesn't have to be called that :))
2. Is it necessary to test that the whole form is rendered in both cases? Is it possible one could fail and the other pass? If not, then I'd say repeating the assertions of the common fields for both countries is unnecessary

mforner13 (commented):
Just one comment about the tests :)

test/functional/cypress/specs/companies/add-company-spec.js
chloehkwong1: Hi @mforner13 , thanks for the review!  Good point - we have now extracted it into a const `manualAddUSOrCanadianCompanyForm`.

mforner13 (approved):
Looks good 👍

test/functional/cypress/specs/companies/add-company-spec.js
```diff
@@ -81,6 +96,46 @@ const goToUKCompanySectorAndRegionPage = () => {
   cy.get(selectors.companyAdd.continueButton).click()
 }
 
+const manualAddUSOrCanadianCompanyForm = () => {
```
tomkingston, line 99: The name of this function suggests that it is taking an action to manually add a US or Canadian Company, however, from what I understand, it is just asserting that all the relevant fields are visible - so perhaps something like `assertUsOrCanadianCompanyFormFields` would be a more accurate description?

tomkingston (approved):
One comment on the name of the `manualAddUSOrCanadianCompanyForm` function, but otherwise all looks good :)

test/functional/cypress/specs/companies/add-company-spec.js
chloehkwong1: @tomkingston Thanks, good idea. I've changed it :)

PR #3821, opened 2021-07-22
Add detailed header to opportunity details page

mforner13:
## Description of change

This PR creates a new generic component `LocalHeaderDetails` which adds a row of data to a `LocalHeader` component. 

This is then used to create the local header for the opportunity details page, which loads details of an opportunity into the page header after the task to get this data from the API succeeds (same task which fetches data for the whole page).

## Test instructions

_What should I see?_

- Go to `/investments/opportunities` and pick any opportunity
- The detailed header should appear on the page.

To test in storybook

- Go to the LocalHeaderDetails section
- There should be an example usage of the component alongside a README for it

### Before
Screenshot 2021-07-22 at 12 39 41

### After
Screenshot 2021-07-22 at 12 40 16

## Checklist


src/apps/investments/client/opportunities/Details/OpportunityDetailsHeader.jsx
```diff
@@ -0,0 +1,83 @@
...
+      ? 'Not yet defined'
+      : ukRegions.length > 1
+      ? 'Multiple'
+      : `${ukRegions.map((v) => v.label)}`
+  const assetClass =
+    assetClasses.length == 0
+      ? 'Not yet defined'
+      : assetClasses.length > 1
+      ? 'Multiple'
+      : `${assetClasses.map((c) => c.label)}`
```
faissalMT, line 29: Nested ternary operators can be hard to read, the code in these two assignments is also quite similar so it might be worth putting it into a function

faissalMT (commented):
Looks good, just one minor thing

ian-leggett (commented):
There seems to be 2 duplicate API calls for this page 
Screenshot 2021-07-26 at 11 00 57

src/apps/investments/views/opportunity-details.njk
```diff
@@ -1,8 +1,15 @@
...
+
 {% block content %}
-{% component 'react-slot', {
-  id: 'opportunity-details',
-  props: props
-} %}
+  {% component 'react-slot', {
+    id: 'opportunity-details',
+    props: props
+  } %}
```
ian-leggett, line 14: Ideally all the react needs to mounted in one place rather than 2 mounts via two react slots. Can we merge this or is this too much at this stage? Maybe use a different layout from nunjucks, something that exposes more of the page.

src/client/components/LocalHeaderDetails/index.jsx
```diff
@@ -0,0 +1,46 @@
...
+  display: inline-block;
+  padding-right: ${SPACING.SCALE_5};
+`
+
+const StyledHeaderListLabel = styled('label')`
+  color: #6f777b;
+`
+
+const LocalHeaderDetails = ({ items }) => {
+  return (
```
ian-leggett, line 21: We can use an implicit return here

src/client/components/LocalHeaderDetails/index.jsx
```diff
@@ -0,0 +1,46 @@
...
+  background-color: ${GREY_4};
+`
+
+const StyledHeaderList = styled('li')`
+  display: inline-block;
+  padding-right: ${SPACING.SCALE_5};
+`
+
+const StyledHeaderListLabel = styled('label')`
+  color: #6f777b;
```
ian-leggett, line 17: Should this colour come from gov-colours?

src/apps/investments/client/opportunities/Details/OpportunityDetailsHeader.jsx
mforner13: Good thinking! Have updated - let me know what you think :)

src/apps/investments/views/opportunity-details.njk
mforner13: Merging it into one would be much easier as then wouldn't need the opportunity local header to be listening out for the task to get the API data too! I had thought that we don't have a layout atm which would allow this, but will investigate :)

mforner13:
> There seems to be 2 duplicate API calls for this page
> Screenshot 2021-07-26 at 11 00 57

@ian-leggett  This looks like it's an existing bug being caused by the `formSaved` prop in [opportunities.jsx](https://github.com/uktrade/data-hub-frontend/blob/master/src/apps/investments/client/opportunities/Details/Opportunities.jsx#L92). As it looks like this prop was a workaround to get the summary table to render after the form submitted, i will work to refactor this mechanism without the duplicate API calls in my next PR :)

src/apps/investments/client/opportunities/Details/OpportunityDetailsHeader.jsx
faissalMT: https://github.com/uktrade/data-hub-frontend/pull/3821/files#diff-079c8f04b3b2b10b52cc41503a99ce9a53563b1a5e48f7a2e20bb567be1f0846R20-R26 might work better as just a set of if statements with multiple returns. Since there's only one item in the array it might also be clearer to `[0]` it instead of `.map()`ing

src/apps/investments/views/opportunity-details.njk
mforner13: It was possible! Have updated to be within one slot :)

src/apps/investments/client/opportunities/Details/OpportunityDetailsHeader.jsx
mforner13: You're right that is more readable :) Have updated with both suggestions

faissalMT (approved):

ian-leggett (approved):

PR #3847, opened 2021-07-30
Add change status form for opportunities

cgsunkel:
## Description of change

This PR will allow the user to change the status of an opportunity.

## Test instructions

Go to an opportunity page and replace `details` with `status`.

## Screenshots
I'll put a screenshot in when I'm finished.

## Checklist


src/apps/investments/client/opportunities/Details/OpportunityChangeStatus.jsx
```diff
@@ -0,0 +1,137 @@
...
+
+  useEffect(() => {
+    if (state[SAVED]) {
+      window.location.href =
+        urls.investments.opportunities.details(opportunityId)
+    }
+  })
+
+  return (
+    <Task.Status
```
cgsunkel, line 58: This is needed to get the opportunity name for the breadcrumb

src/apps/investments/client/opportunities/Details/OpportunityDetailsHeader.jsx
```diff
@@ -26,10 +26,23 @@ const OpportunityDetailsHeader = ({ details }) => {
     }
   }
 
+  const getStatus = (status) => {
+    if (status.label) {
+      return status.label
+    }
+    return 'Unassigned'
+  }
+
```
peterhudec, line 35: I would get rid of this function. It's only used once and it can be replaced by a more readabe inline fallback:
```js
<span>
  {status.label || 'Unassigned'}
  {' - '}
  <a href={urls.investments.opportunities.status(id)}>change</a>
</span>
```
Or even better do the fallback when destructuring. Looks like you are only using the label from the `status`:
```js
const OpportunityDetailsHeader = ({
  id,
  status: { label: statusLabel = 'Unassigned' },
  detailsFields: { createdOn, name, ukRegions, assetClasses, opportunityValue },
}) => {
  // ...
}
```
And then just:
```js
<span>
  {statusLabel}
  {' - '}
  <a href={urls.investments.opportunities.status(id)}>change</a>
</span>
```

src/apps/investments/client/opportunities/Details/constants.js
```diff
@@ -5,6 +5,7 @@ const ESTIMATED_RETURN_RATE_FIELD_NAME = 'estimated_return_rate'
 const TIME_HORIZONS_FIELD_NAME = 'time_horizons'
 const CLEARED_REFERENCE = '02d6fc9b-fbb9-4621-b247-d86f2487898e'
 const ISSUES_IDENTIFIED_REFERENCE = '9beab8fc-1094-49b4-97d0-37bc7a9de631'
+const STATUS_FIELD_NAME = 'status'
```
peterhudec, line 8: This is an overkill. The constant is only used once. It would be much more readable hardcoded inline:
```js
<FieldRadios
  name="status"
  options={opportunityStatusOptions}
  data-test="types-of-status"
  required="You must select a status"
/>
```

src/apps/investments/client/opportunities/Details/reducer.js
```diff
@@ -82,6 +87,14 @@ export default (state = initialState, { type, result }) => {
...
         },
       }
+    case INVESTMENT_OPPORTUNITY__STATUS_METADATA_LOADED:
+      return {
+        ...state,
+        metadata: {
+          ...state.metadata,
+          opportunityStatuses: result.opportunityStatuses,
+        },
+      }
```
peterhudec, line 97: You wouldn't need this if you used the resource.

src/apps/investments/client/opportunities/Details/state.js
```diff
@@ -3,11 +3,15 @@ export const TASK_GET_OPPORTUNITY_DETAILS_METADATA =
...
+export const TASK_GET_OPPORTUNITY_STATUS_METADATA =
+  'TASK_GET_OPPORTUNITY_STATUS_METADATA'
 
 export const TASK_SAVE_OPPORTUNITY_DETAILS = 'TASK_SAVE_OPPORTUNITY_DETAILS'
 export const TASK_SAVE_OPPORTUNITY_REQUIREMENTS =
   'TASK_SAVE_OPPORTUNITY_REQUIREMENTS'
+export const TASK_SAVE_OPPORTUNITY_STATUS = 'TASK_SAVE_OPPORTUNITY_STATUS'
 
 export const ID = 'opportunityDetails'
+export const SAVED = 'edit-saved'
```
peterhudec, line 15: Although this one is used twice, I still find it much more readable hardcoded.

src/apps/investments/client/opportunities/Details/tasks.js
```diff
@@ -99,3 +99,20 @@ export function saveOpportunityRequirements({ values, opportunityId }) {
...
+}
+
+export const getStatusMetadata = () =>
+  Promise.all([
+    apiProxyAxios.get(
+      '/v4/metadata/large-capital-opportunity/opportunity-status'
+    ),
+  ]).then(([{ data: opportunityStatuses }]) => ({
+    opportunityStatuses: opportunityStatuses.map(idNameToValueLabel),
+  }))
```
peterhudec, line 118: You would get this for free if you used the `resourceFactory`.

src/client/actions.js
```diff
@@ -153,6 +153,10 @@ export const INVESTMENT_OPPORTUNITY__DETAILS_CHANGE =
   'INVESTMENT_OPPORTUNITY__DETAILS_CHANGE'
 export const INVESTMENT_OPPORTUNITY__REQUIREMENTS_CHANGE =
   'INVESTMENT_OPPORTUNITY__REQUIREMENTS_CHANGE'
+export const INVESTMENT_OPPORTUNITY__STATUS_METADATA_LOADED =
+  'INVESTMENT_OPPORTUNITY__STATUS_METADATA_LOADED'
```
peterhudec, line 157: You wouldn't need this if you used the resource.

src/client/components/LocalHeaderDetails/index.jsx
```diff
@@ -36,7 +36,7 @@ LocalHeaderDetails.propTypes = {
   items: PropTypes.arrayOf(
     PropTypes.shape({
       label: PropTypes.string.isRequired,
-      value: PropTypes.string.isRequired,
+      value: PropTypes.oneOfType([PropTypes.string, PropTypes.node]).isRequired,
```
peterhudec, line 39: `PropTypes.node` already includes `PropTypes.string` so it can just be `value: PropTypes.node`.

test/functional/cypress/specs/investments/opportunity-status-form-spec.js
```diff
@@ -0,0 +1,59 @@
...
+      cy.contains('Fully funded').should('be.visible')
+      cy.contains('Suspended').should('be.visible')
+      cy.contains('Abandoned').should('be.visible')
+    })
+  })
+  context('when submitting the form', () => {
+    it('should give an error if a status is not selected', () => {
+      cy.contains('Save').click()
+      cy.get('#form-errors').should('be.visible')
+      cy.get('#form-errors').should('contain', 'You must select a status')
```
peterhudec, line 39: It's better to not rely on implementation details like `#form-errors` in tests.

peterhudec (changes requested):
Looks good, but you could simplify it quite a bit if you used the `Resource` component and some other suggestions.

src/apps/investments/client/opportunities/Details/OpportunityChangeStatus.jsx
```diff
@@ -0,0 +1,137 @@
...
+    },
+    { text: 'Change status' },
+  ]
+
+  useEffect(() => {
+    if (state[SAVED]) {
+      window.location.href =
+        urls.investments.opportunities.details(opportunityId)
+    }
+  })
```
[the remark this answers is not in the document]
peterhudec: Actually, there's a ~better~ simpler way how to do the redirect. If you used `FormStateful` you can return a URL from its `onSubmit` handler and the form will redirect there.

src/apps/investments/client/opportunities/Details/OpportunityChangeStatus.jsx
```diff
@@ -0,0 +1,117 @@
...
+            />
+            <Task>
+              {(getTask) => {
+                const saveStatusTask = getTask(TASK_SAVE_OPPORTUNITY_STATUS, ID)
+                return (
+                  <OpportunityStatusResource>
+                    {(opportunityStatuses) => (
+                      <>
+                        <Main>
+                          <MultiInstanceForm
```
peterhudec, line 69: Better use the `FormStateful` instead of `MultiInstanceForm`. You will have a progress indicator and redirect mechanism for free. Sorry, I have only found out that the `MultiInstanceForm` doesn't support the redirect now.

peterhudec (changes requested):
Almost there. Only a couple of things:

`OpportunityStatus` resource name should be plural.

You are still using a _task_ in the form's `onSubmit`, but unfortunately the forms `onSubmit` mechanism is not compatible with the Redux architecture and thus also with the task abstraction. Using a task in the form results in a bug when the task promise rejects (axios request fails). You can try this out by tweaking the URL on [this line](https://github.com/uktrade/data-hub-frontend/pull/3847/files#diff-d9aa4bfff4b59d75f544cfd4cfb256265f3921031ef07f31320f217b6c352abaR105) to something invalid. When you submit the form, nothing will happen, whereas the user should be made aware that the request failed by way of an error message.

You would see the bug even better if you used `FormStateful` instead of `MultiInstanceForm`, which has built in a progress indicator and also has a redirect mechanism built in (although one which violates the Redux architecture) where to redirect, you only need to return the redirect path from the `onSubmit` handler.

If you used the `FormStateful` with the builtin mechanism, you would see that the form always redirects after a fraction of a second, regardless of how long the task's promise takes to resolve or whether the the task rejects. This will result in two types of inconsistency:

- If the request fails the user would be redirected without realizing that there was an error, but would see (if attentive enough) that the status has not changed
- If the request takes longer, the user would be redirected and still see the old status, whereas the status actually changed, but after the redirect, the user might think that something is not working, but if she refreshed the page, she would see the updated status.

Because of this, it is best to use `FormStateful` which will give you the loading indicator and redirect mechanism for free, but instead of starting a task in its `onSubmit` you should pass an asynchronous function to it and do a direct axios call in it.

PR #3868, opened 2021-08-06
Update Data Hub dependencies

paulgain:
## Description of change
This fix was born out of a dependabot [PR](https://github.com/uktrade/data-hub-frontend/pull/3855)

**@babel/plugin-proposal-class-properties**
After bumping `@babel/preset-env from 7.14.7 to 7.14.9` and running Storybook (no problems with the client) there was an error `Missing class properties transform` (scroll right). 
```
npm run storybook
...
...
...
ERROR in ./src/client/components/ActivityFeed/activities/card/CardTable.jsx
Module build failed (from ./node_modules/babel-loader/lib/index.js):
SyntaxError: /Users/paulgain/Dev/data-hub-frontend/src/client/components/ActivityFeed/activities/card/CardTable.jsx: Missing class properties transform.
  30 |
  31 | class DetailsRow extends React.PureComponent {
> 32 |   static propTypes = {
     |   ^
  33 |     header: PropTypes.string.isRequired,
  34 |     children: PropTypes.node,
  35 |   }
```
This indicates we required the `@babel/plugin-proposal-class-properties` plugin to transform class properties, which we have, in fact, we have it twice as it [ships](https://github.com/babel/babel/blob/master/packages/babel-preset-env/package.json#L23) with `@babel/preset-env`. To avoid future red herrings and confusion I've removed the duplicate. 

However, that didn't fix the error, the fix was to upgrade Storybook to the latest version along with other dependencies which can be seen below. If we can, it would be great to ensure both `react` and `react-dom` are at the same version (or as close as) otherwise it may cause us problems going forward, the same applies to `@babel/core` and `@babel/preset-env`.

You may also notice that `@storybook/addon-knobs` has only been upgraded to `6.3.0` instead of `6.3.6` like its compatriots, this is because it's now [deprecated](https://github.com/storybookjs/storybook/discussions/15060). We should be using [`storybook/controls`](https://storybook.js.org/docs/react/essentials/controls) which is out of scope for this PR and has breaking changes, Storybook authors are going to maintain `@storybook/addon-knobs` until `v7` giving everyone time to migrate.

**I realise we are holding back from upgrading Storybook for a separate reason, however, @cgsunkel is away this week so I can't ask him why and I'm also away next week so won't be able to speak to him.**

## Checklist


cgsunkel (changes requested):
This looks good and looks like we've fixed this WSOD issue finally, however there are a couple of things that need to be updated before this can go in:

- The rest of the `@babel` dependencies should be updated to 7.15.0 for consistency
- The Dependabot ignore list needs to be updated, I think we should just remove one of the Storybook packages and leave the rest. This is because if one comes through they should all be updated anyway and having a PR come up for each is wasteful. We should follow the same approach with the `@babel` dependencies.
- [This section](https://github.com/uktrade/data-hub-frontend/blob/master/docs/Dependabot.md#storybook) of the Dependabot docs needs to be updated to say that the `knobs` package can't be updated anymore. We should also amend it to include the `@babel` packages.

cgsunkel:
Storybook released a new version last night, so we should also upgrade all the Storybook dependencies to 6.3.7

cgsunkel (approved):

ian-leggett (approved):

PR #3891, opened 2021-08-10
Add id's to sort options for tracking in Google analytics...

ian-leggett:
## Description of change

We need to know how many people are using the two sort options (Stage and Sort by) available on the new dashboard. In order to track this usage we need id's added to each option inside the two select drop down elements.

## Test instructions

Nothing should change apart from id's should now exist in the markup related to these two select drop downs.

1. Inspect the generated mark up

## Screenshots
Screenshot 2021-08-10 at 14 52 52
Screenshot 2021-08-10 at 15 02 16

## Checklist

### DIFF
--- a/src/client/components/MyInvestmentProjects/InvestmentListFilter.jsx
+++ b/src/client/components/MyInvestmentProjects/InvestmentListFilter.jsx
@@ -1,4 +1,5 @@
 import React from 'react'
+import { kebabCase } from 'lodash'
 
 import { Select } from '../../components'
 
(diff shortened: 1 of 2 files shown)


src/client/components/MyInvestmentProjects/InvestmentListFilter.jsx
```diff
@@ -10,7 +11,7 @@ const InvestmentListFilter = ({ options, onChange }) => (
     }}
   >
     {options.map(({ id, name }, index) => (
-      <option value={id} aria-label={name} key={index}>
+      <option value={id} aria-label={name} key={index} id={kebabCase(name)}>
```
tomkingston, line 14: Why use the kebabed name when you can just use the id?

tomkingston (approved):
Looks good, but wondering why you kebab case the name when you could just use the id?

src/client/components/MyInvestmentProjects/InvestmentListFilter.jsx
tomkingston: Ahhhh, I see - presumably this is to avoid using uuids as the id which should make it a bit clearer in analytics

cgsunkel (approved):

PR #3898, opened 2021-08-12
Add Postcode Validation to postcode field when country is United States or Canada

ccp92:
## Description of change

This PR adds front-end validation to the add/edit company form to mark Postcode as a mandatory field when the country is United States or Canada.
Also corrects the Post Code terminology to match the regional equivalents (ZIP code for US and postal code for Canada)

## Test instructions

Go to add a new company/ edit an existing company that has a country set as United States or Canada.
If you attempt to submit the form without completing the postcode field, the field should be highlighted with an error message.

### Before

image

### After
image
image

## Checklist


src/client/components/Form/elements/FieldAddress/index.jsx
```diff
@@ -155,6 +155,17 @@ const FieldAddress = ({
...
 
+  const postcodeLabel = () => {
+    if (isUS) return 'Zip Code'
+    if (isCanada) return 'Postal Code'
+    return 'Postcode (optional)'
+  }
+
+  const postcodeErrorMessage = () => {
+    if (isUS) return 'Zip Code is required'
+    if (isCanada) return 'Postal Code is required'
```
mforner13, line 166: The common pattern for errors for a text input field is `Enter a <thing you're asking for>` 

So I think these should be `Enter a zip code` and `Enter a postal code`.

Guidance on error messages in the design system [here](https://design-system.service.gov.uk/components/error-message/).

In other places in this form it misses out the 'a' which definitely seems grammatically incorrect :p e.g. `Enter town or city` and `Enter postcode`, so if you've got a content designer on hand could check if they're cool with us updating the others.

mforner13 (changes requested):

src/client/components/Form/elements/FieldAddress/index.jsx
chloehkwong1: @mforner13 thanks for this! We'll look into updating the others at the same time :)

mforner13 (approved):

PR #3901, opened 2021-08-13
Duplicate contact warning

peterhudec:
## Description of change

Adds a warning message when a user is about to create a new contact with an email already used in some other contact.

## Test instructions

1. Go to a company page of your choice
2. Click on the _Company contacts_ tab
3. Click on the _Add contact_ button
4. Fill out the form with a unique _email_
5. Submit the form
6. The form should redirect you to the details page of the newly created contact
7. Repeat steps 1 to 5, except in step 4 don't use a unique _email_, but use the same email as before
8. After the form was submitted there should appear a duplicate email message in a blue box on top of the form
9. Not heeding the warning, submit the form again
10. The form should redirect you to the details page of the newly created contact with a duplicate email
11. This flow should also work 

## Screenshots

Screen Shot 2021-08-13 at 4 59 17 PM


test/cypress/support/commands.js
```diff
@@ -285,3 +285,21 @@ Cypress.Commands.add('runA11y', (context = null, options = null) => {
...
+Cypress.Commands.add(
+  'checkRadioGroup',
+  (label, option) =>
+    cy.contains(label).parent().find(`[aria-label="${option}"]`).check()
+)
+
+Cypress.Commands.add(
+  'clickSubmitButton',
+  (label) => cy.get('form').contains('button', label).click()
+)
```
mforner13, line 305: Missing new line at end of file :)

test/end-to-end/cypress/specs/DIT/contacts-spec.js
```diff
@@ -31,4 +31,56 @@ describe('Contacts', () => {
       .and('contain', 'company.contact@dit.com')
       .and('contain', 'Rua Candido Portinari, Numero 521, Campinas, Brazil')
   })
+
+  it('Create with duplicate email', () => {
```
mforner13, line 35: Would this test be better placed in the functional suite? I'd thought that that would be where we test the various scenarios that could occur, and the e2e suite was more for the e2e happy path - although I may be wrong! 😋 

If you do move it - the 'it' could more descriptive of expected behaviour like the test above :) 

E.g. `Should show warning message when adding contact with duplicate email address`

test/sandbox/routes/v3/contact/contact.js
```diff
@@ -22,6 +22,10 @@ const validateContactForm = function (formData) {
 }
 
 exports.contact = function (req, res) {
+  // This is here to allow creation of new contacts. The email must contact "new"
```
mforner13, line 25: Should `contact` be `contain`? :)

mforner13 (commented):
Just a question around the tests & v minor corrections

test/end-to-end/cypress/specs/DIT/contacts-spec.js
peterhudec: So here I'm testing the whole flow when a user creates a contact with a unique email and subsequently creates another contact with the same email. This requires the contact to actually be persisted and also reflected in the `/v3/contact?email=...` query. The functional tests with the sandbox backend are pretty much stateless so it's impossible to write this kind of tests in the functional suite. Or it was impossible, when I wrote the e2e tests, because I had to introduce some conditionals to the `/v3/contact` sandbox endpoint when it behaves slightly differently when it receives a request with an `email` param. So it actually could be implemented in the functional suite, but such a test would be of much less value.

The e2e tests are sitting at the top of the proverbial testing pyramid and the functional tests are right below them. But I don't think the way we organize the tests into functional and e2e is in line with the [general understanding of these two concepts](https://www.atlassian.com/continuous-delivery/software-testing/types-of-software-testing). Our e2e tests fit more the general description of functional tests and our functional tests seem to be just som sort of a temporary technical workaround to allow testing of parts of the UI which depend on existence of a HTTP server. If we had the technical solution which we have now in the e2e tests we would never have introduced the sandbox and we would not have any functional suite.

> If you do move it - the 'it' could more descriptive of expected behaviour like the test above :)
> E.g. Should show warning message when adding contact with duplicate email address
I agree with this ☝️ will fix it

mforner13: Ahhh ofc! I hadn't considered you'd need to persist the data between requests! That makes sense - thanks Peter 😄

mforner13 (approved):
Looks good 👍

elcct (approved):

PR #3930, opened 2021-08-23
Add Opportunities field to interaction form

LaurenQurashi:
## Description of change

This PR introduces the addition of the Opportunities fields into the create interactions form. Recording this information comes in the shape of 2 questions on the Interactions form, a yes/no as to whether the interaction relates to an opportunity, then a typeahead that lets you choose one opportunity it's attached to.  This is necessary so that we can link an interaction to an opportunity, and also is the precursor to work around allowing users to see interactions related to an opportunity on the opportunity's end of things. 

## Test instructions

Head to the deployment of the branch below - go to companies, click on a company and make a new interaction. You should see a field at the bottom of the main form (after the first step of the create interaction journey) which asks if this interaction is related to a large capital opportunity. Click yes, and you should see a typeahead field come up, allowing you to search for and select existing opportunities. You should be able to submit the form and be redirected to a table that shows you the details of the interaction you've just created. You should see in this table the details of the opportunity you selected, along with a working link taking you to the correct opportunity. You should be able to click on edit interaction and see your previous selections in the fields as initial values. 

## Screenshots

Screenshot 2021-09-09 at 16 56 37
Screenshot 2021-09-09 at 17 00 22

## Checklist


src/apps/interactions/apps/details-form/client/InteractionDetailsForm.jsx
```diff
@@ -137,7 +138,11 @@ const InteractionDetailsForm = ({
...
                         <StepInteractionDetails
                           companyId={companyId}
+                          interactionId={interactionId}
                           activeEvent={initialValues.event}
+                          relatedOpportunity={
+                            initialValues.large_capital_opportunity
+                          }
                           onOpenContactForm={(e) => {
                             e.preventDefault()
                             openContactFormTask.start({
```
LaurenQurashi, line 148: I add in the interaction ID in here to pass it through to form step, as I need it for the initialvalue formatter

src/apps/interactions/apps/details-form/client/StepInteractionDetails.jsx
```diff
@@ -167,17 +173,27 @@ const StepInteractionDetails = ({
...
   )
+  const opportunityBooleanFormatter = (interactionId, opportunityId) => {
+    let hasOppInitialValue
+    if (interactionId) {
+      hasOppInitialValue = opportunityId ? OPTION_YES : OPTION_NO
+    } else {
+      hasOppInitialValue = null
+    }
+    return hasOppInitialValue
+  }
```
LaurenQurashi, line 193: Not sure if this is the best name, but without being too long it's the best I could think of. Essentially it's in charge of setting the initial value for the has_related_opportunity question. The other fields in this form save them on the API but I thought that was a bit of a waste of space. It checks if the interaction has an id, if not then it's a new one and can have an empty initial value. If it's got an id, then it checks whether the answer should be yes or no, based on the existence of an opportunity id on the interaction.

src/apps/interactions/apps/details-form/client/tasks.js
```diff
@@ -135,8 +135,9 @@ export function saveInteraction({ values, companyIds, referralId }) {
     related_trade_agreements: transformArrayOfOptions(
       values.related_trade_agreements
     ),
+    large_capital_opportunity: values.large_capital_opportunity?.value,
   }
-
+  delete values.has_related_opportunity
```
LaurenQurashi, line 140: This isn't ideal, but due to the way the form works, it sends off all the values of the form to the API but I don't want to store the `has_related_opportunity` boolean. Other fields that work in this way with the conditional rendering of a further question all store this value so it works for them. Has anyone got any thoughts about the way I do this? feels a bit wrong to just throw it out last minute :(

src/apps/interactions/labels.js
```diff
@@ -31,6 +31,9 @@ const interaction = {
   has_related_trade_agreements:
     'Does this interaction relate to a named trade agreement?',
   related_trade_agreements: 'Named trade agreement(s)',
+  has_related_opportunity:
+    'Does this interaction relate to a large capital opportunity?',
+  large_capital_opportunity: 'Related large capital opportunity',
```
LaurenQurashi, line 36: This is required for the logic to make the little table on the creation of an interaction. I thought I'd include the has_related_opportunity question as well in case it's used elsewhere, but can also understand not including it until it's definitely needed.

src/apps/interactions/transformers/interaction-response-to-view.js
```diff
@@ -5,10 +5,15 @@ const labels = require('../labels')
 const { getDataLabels } = require('../../../lib/controller-utils')
 const groupExportCountries = require('../../../lib/group-export-countries')
 
-function transformEntityLink(entity, entityPath, noLinkText = null) {
+function transformEntityLink({
+  entity,
+  entityPath,
+  urlAffix = '',
+  noLinkText = null,
+}) {
```
LaurenQurashi, line 13: I had to amend this function. It's responsible for making the links in the completion summary table thing, however it only accepted links that were "pathway/id" and didn't accommodate for anything after the id. The opportunities have `/details/` after their id so it would break. Do you reckon `urlAffix ` is a good name for this? it's just stuff that's supposed to go after the ID but I don't know if that's clear enough.

src/apps/interactions/transformers/interaction-response-to-view.js
```diff
@@ -107,9 +113,15 @@ function transformInteractionResponseToViewRecord(
...
+      transformEntityLink({
+        entity: companies[0],
+        entityPath: 'companies',
+      }),
     contacts: contacts.map((contact) =>
-      transformEntityLink(contact, 'contacts')
+      transformEntityLink({
+        entity: contact,
+        entityPath: 'contacts',
+      })
```
LaurenQurashi, line 124: This is just me changing the existing uses of `transformEntityLink` to take in an object.

test/end-to-end/cypress/specs/DA/add-interaction-spec.js
```diff
@@ -49,6 +49,8 @@ describe('DA add Investment Project interaction', () => {
         .select('Investment - Services')
```
LaurenQurashi, line 49: The tests for interactions were exceptionally confusing and in abundance, if you pay particular attention to anything in this PR, please let it be these tests! Would love to know what I've missed or could do better! for context, I followed the trade agreements logic to decipher whether the opportunities were needed or not.

test/functional/cypress/specs/interaction/details-form-spec.js
```diff
@@ -849,6 +880,7 @@ describe('Adding an interaction from a referral', () => {
         were_countries_discussed: 'yes',
         kind: 'interaction',
         event: null,
+        large_capital_opportunity: 'a84f8405-c419-40a6-84c8-642b7c3209b2',
         export_countries: [
           {
             country: '6e6a9ab2-5d95-e211-a939-e4115bead28a',
```
LaurenQurashi, line 886: In the code, next to this file, there's one called `details-spec.js`. It seems to be testing that the little summary table thing renders the right items, which I would imagine should include my opportunities, however, each context doesn't really relate to rendering a thing if the thing exists, it's more under specific edge cases like if there's documents. Have I missed out a test that checks whether this table renders my opportunity if applied?

test/selectors/interaction-form.js
```diff
@@ -44,4 +44,5 @@ module.exports = {
     noInterest: '#field-' + EXPORT_INTEREST_STATUS.NOT_INTERESTED,
   },
   hasRelatedTradeAgreementsNo: '#field-has_related_trade_agreements [value=no]',
+  hasRelatedOpportunityNo: '#field-has_related_opportunity [value=no]',
 }
```
LaurenQurashi, line 48: Thanks for reviewing! I really appreciate it 🌻

src/apps/interactions/apps/details-form/client/StepInteractionDetails.jsx
tomkingston: You could just call `hasOppInitialValue` `hasValue` since that should be specific enough inside the scope of this function.... but then you could make the function a bit more declarative and avoid that altogether (untested):

```
    <FieldRadios
          inline={true}
          name="has_related_opportunity"
          legend="Does this interaction relate to a large capital opportunity?"
          required="Answer if this interaction relates to a large capital opportunity"
          options={OPTIONS_YES_NO}
          initialValue={interactionId &&  relatedOpportunity ? OPTION_YES : OPTION_NO}
        />
```

src/apps/interactions/apps/details-form/client/tasks.js
tomkingston: At the moment you are creating a side-effect by manipulating one of the function inputs - it would be better to create a new object based on values that leaves out `has_related_opportunity`, which you can do with something like this: 
```
const { has_related_opportunity, ...cleanedValues } = values
...
    export_countries: transformExportCountries(cleanedValues),
```

src/apps/interactions/transformers/interaction-response-to-view.js
tomkingston: I think `urlSuffix` is slightly less ambiguous but affix is ok

src/apps/interactions/transformers/interaction-response-to-view.js
```diff
@@ -5,10 +5,15 @@ const labels = require('../labels')
...
+function transformEntityLink({
+  entity,
+  entityPath,
+  urlAffix = '',
+  noLinkText = null,
+}) {
   return entity
     ? {
-        url: `/${entityPath}/${entity.id}`,
+        url: `/${entityPath}/${entity.id}` + urlAffix,
```
tomkingston, line 16: You can continue with the string interpolation rather than concatnation, ie: 
```
    url: `/${entityPath}/${entity.id}${urlAffix}`,
```

test/selectors/interaction-form.js
```diff
@@ -44,4 +44,5 @@ module.exports = {
     noInterest: '#field-' + EXPORT_INTEREST_STATUS.NOT_INTERESTED,
   },
   hasRelatedTradeAgreementsNo: '#field-has_related_trade_agreements [value=no]',
+  hasRelatedOpportunityNo: '#field-has_related_opportunity [value=no]',
```
tomkingston, line 47: You should probably also test that everything happens as expected when you click `hasRelatedOpportunityYes`. I think there is also a plan to move away from these selectors constants files and use either Peter's new framework or inline `[data-test=...]` selectors, but I think what you have done is the correct approach for now.

tomkingston (commented):
Looks good - I've left a couple of comments but nothing major

src/apps/interactions/apps/details-form/client/tasks.js
mforner13: Such a smart way of doing it!! Thanks Tom 😄

src/apps/interactions/apps/details-form/client/StepInteractionDetails.jsx
mforner13: Thank you that works! Just needed brackets around the ternary :)

src/apps/interactions/apps/details-form/client/tasks.js
mforner13: Noticed that there was already a `FIELDS_TO_OMIT` array at the top of tasks.js which is omitted from the payload in the request - so have added the field to this array to follow the existing pattern :)

test/selectors/interaction-form.js
mforner13: @tomkingston there's a test added to the functional suite for selecting yes and then a large capital opportunity - do you think it needs to be added to the end to end suite too? :)

chloehkwong1 (approved):
Looks really good to me! I'm happy with only the functional suite testing the yes opportunity case as it doesn't make sense to me that you'd repeat the same test twice for e2e too, but let me know if this is the wrong way of thinking :)

test/functional/cypress/fixtures/investment/large-capital-opportunity-complete.json
```diff
@@ -1,3 +1,4 @@
 {
```
PippoRaimondi, line 1: another way of getting this information is by accessing the fixture used to render the UI. This would then reduce maintenance if something changed in that fixture:

You could require the fixtures (if not already) and access them like this:

```
const largeCapitalOpportunity = require('../../../../sandbox/fixtures/ test/sandbox/fixtures/v4/investment/large-capital-opportunity-complete.json

largeCapitalOpportunity.id
largeCapitalOpportunity.name
```

PippoRaimondi (approved):
left a tiny comment and pre approved

test/functional/cypress/fixtures/investment/large-capital-opportunity-complete.json
mforner13: Good point @PippoRaimondi - have removed the two cypress fixture files and only used the sandbox fixtures :) Thanks!

PR #3956, opened 2021-09-06
CDS-186 Adding US/Canada specific address fields to Contacts

chloehkwong1:
## Description of change

This PR updates the address field to remove the forced postcode field, and adds support for dynamically showing and hiding fields based on country in order to enable state & province for company contacts.  
This work required that changes be made whereever the address field is used, since it no longer saves data to an area field but rather to two distinct fields: areaUS and areaCanada, and it is now the responsibility of the form submission to combine these according to other data in the form. This was done due to complexities involved with ensuring old data didn't persist when changing.  

## Test instructions
In Django Admin, add two features flags `address-postcode-company-required-field` and `contacts-area-field`. 
Either:  
1. Go to a US/Canadian company and add a contact.
or
2. Go to a US/Canadian contact and click edit contact.

Then:
Answer 'No' to the question 'Is the contact's address the same as the company address?'

You should be able to see an extra field _State_ for US contacts, and _Province_ for Canadian contacts. 
For US companies, there will be a _ZIP Code_ field which is mandatory. For Canadian companies, there will be a _Postal Code_ field which is mandatory.

When editing an existing contact, the previously saved area should appear in the state/province field.  

For UK contacts, the find postcode button should show and function and the state/province field should not appear.

### Before
**Canadian Company**
image

**US Company**
image

### After
**Canadian Company**
image

**US Company**
image

## Checklist


src/client/components/Form/elements/FieldAddress/index.jsx
```diff
@@ -63,18 +60,31 @@ const FieldAddress = ({
...
+  const [isUK, setIsUK] = useState(country?.id === UNITED_KINGDOM_ID)
+  const [isUS, setIsUS] = useState(country?.id === UNITED_STATES_ID)
+  const [isCanada, setIsCanada] = useState(country?.id === CANADA_ID)
+
   useEffect(() => {
     setIsLoading(isSubmitting && isAreaFilterSubmitting)
   }, [isSubmitting, isAreaFilterSubmitting])
 
+  useEffect(() => {
+    if (country_form_value && isCountrySelectable) {
```
Staberinde, line 81: Can I check that my understanding of the logic of this line is correct; if country can be mutated within the form then we need the `useEffect` so that we can re-evaluate this code every time the country is changed. Otherwise these statements are evaluated somewhere else on the initial render of this form?

src/client/components/Form/elements/FieldAddress/index.jsx
```diff
@@ -63,18 +60,31 @@ const FieldAddress = ({
...
+      setIsUK(country_form_value === UNITED_KINGDOM_ID)
+      setIsUS(country_form_value === UNITED_STATES_ID)
+      setIsCanada(country_form_value === CANADA_ID)
+    }
+  }, [country_form_value, isUK, isUS, isCanada])
```
Staberinde, line 86: Do you need `isUK`, `isUS` and `isCanada` as part of the `deps` argument here. Don't we just want this function to update on mutation of `country_form_value`?

test/sandbox/fixtures/v3/feature-flag/feature-flag.json
```diff
@@ -82,5 +82,9 @@
   {
     "code": "address-postcode-company-required-field",
     "is_active": true
+  },
+  {
+    "code": "contacts-area-field",
```
Staberinde, line 87: Can you document this feature flag in the confluence doc please https://uktrade.atlassian.net/wiki/spaces/CDS/pages/2753298445/Feature+Flags#State%2FProvince-meta-epic---phase-2

Is this the only PR in which you've introduced it? If so I wonder if it's worth making the feature flag more descriptive so it doesn't get confused with the feature flag Faissal's implemented for the backend (assuming we want them to be different?) What are your thoughts on this?

src/client/components/Form/elements/FieldAddress/index.jsx
chloehkwong1: Yes, I believe so. We only want the `useEffect` to run when the country is changed on the form. @vfarah-if double check?

src/client/components/Form/elements/FieldAddress/index.jsx
mforner13: Agreed ^ since the effect is updating the `isCountry` values anyway, seems like the effect doesn't need to be run when these props change.

mforner13 (commented):
Looks good, just the remaining comments from Toby :)

test/sandbox/fixtures/v3/feature-flag/feature-flag.json
faissalMT: Yes, this is the only PR where this has been introduced. There's no feature flag for the backend, choosing whether or not v4 is used is done by the frontend, not the backend.

src/client/components/Form/elements/FieldAddress/index.jsx
vfarah-if: The linter will complain if you don't do this so by virtue of using the state setters otherwise we can override the linter

src/client/components/Form/elements/FieldAddress/index.jsx
vfarah-if: So the only need for this is to listen to the aliased country changes

faissalMT: `useEffect()` should already only trigger when `country_form_value` is changed because of it being specified as a dependent variable. On the initial render they'll just be going in with their initial value, which is decided based on the props passed in

mforner13 (approved):

test/functional/cypress/specs/companies/add-company-spec.js
```diff
@@ -736,6 +735,9 @@ describe('Add company form', () => {
 
     it('should display the manual entry form', () => {
       assertUSOrCanadianCompanyFormFields()
+      cy.get(selectors.companyAdd.newCompanyRecordForm.areaUS).should(
+        'be.visible'
+      )
```
Staberinde, line 740:
```suggestion
      cy.get(selectors.companyAdd.newCompanyRecordForm.areaUS).should(
        'be.visible'
      )
      cy.get(selectors.companyAdd.newCompanyRecordForm.areaCanada).should(
        'not.be.visible'
      )
``` 
Just thinking out loud here

test/functional/cypress/specs/companies/add-company-spec.js
```diff
@@ -765,6 +767,10 @@ describe('Add company form', () => {
 
     it('should display the manual entry form', () => {
       assertUSOrCanadianCompanyFormFields()
+      cy.get(selectors.companyAdd.newCompanyRecordForm.areaCanada).should(
+        'be.visible'
+      )
```
Staberinde, line 772:
```suggestion
      cy.get(selectors.companyAdd.newCompanyRecordForm.areaCanada).should(
        'be.visible'
      )
      cy.get(selectors.companyAdd.newCompanyRecordForm.areaUS).should(
        'not.be.visible'
      )
```
Again, just thinking out loud

test/functional/cypress/specs/contacts/create-and-edit.jsx
```diff
@@ -166,8 +192,95 @@ describe('Create contact form', () => {
...
+        .find('option:selected')
+        .should('have.text', 'Massachusetts')
+    })
+
+    describe('when a UK company is submitted', () => {
+      it('should explicitly submit area as null', () => {
+        cy.intercept('PATCH', `/api-proxy/v4/contact/${EDIT_UK_CONTACT_ID}`).as(
+          'editContactResponse'
+        )
+        cy.log(`/api-proxy/v4/contact/${EDIT_UK_CONTACT_ID}`)
```
Staberinde, line 211: Might be worth removing this before merging

test/functional/cypress/specs/contacts/create-and-edit.jsx
```diff
@@ -166,8 +192,95 @@ describe('Create contact form', () => {
...
+    })
+
+    describe('when changing countries', () => {
+      describe('when changing a Canadian contact country to UK', () => {
+        it('should clear the province value', () => {
+          cy.intercept(
+            'PATCH',
+            `/api-proxy/v4/contact/${EDIT_US_CONTACT_ID}`
+          ).as('editContactResponse')
+          cy.log(`/api-proxy/v4/contact/${EDIT_US_CONTACT_ID}`)
```
Staberinde, line 232: Same as above, might be worth removing this before merging

test/functional/cypress/specs/contacts/create-and-edit.jsx
```diff
@@ -166,8 +192,95 @@ describe('Create contact form', () => {
...
+
+          cy.wait('@editContactResponse').then((xhr) => {
+            expect(xhr.request.body.address_area).to.equal(null)
+          })
+        })
+      })
+
+      describe('when a US contact has a US state and changes to Canada', () => {
+        it('should clear the province value', () => {
+          cy.log(`/api-proxy/v4/contact/${EDIT_US_CONTACT_ID}`)
```
Staberinde, line 249: Same as above, might be worth removing this before merging

test/functional/cypress/specs/contacts/create-and-edit.jsx
```diff
@@ -166,8 +192,95 @@ describe('Create contact form', () => {
...
+          cy.get('#postcode').type('NE16 386')
+          cy.get('[data-test="submit"]').click()
+
+          cy.contains('Select a province')
+        })
+      })
+
+      describe('when a UK contact is switched to US', () => {
+        it('the state should be empty', () => {
+          cy.log(`/api-proxy/v4/contact/${EDIT_UK_CONTACT_ID}`)
```
Staberinde, line 266: Same as above, might be worth removing this before merging

Staberinde (commented):
Looks good! Just a couple of thoughts

src/apps/companies/apps/edit-company/client/EditCompanyForm.jsx
```diff
@@ -33,57 +34,104 @@ function EditCompanyForm({
...
+            areaUS: areaUS(formInitialValues?.address?.area),
+            areaCanada: areaCanada(formInitialValues?.address?.area),
+          },
+        }
+      }}
+    >
+      {({ submissionError }) => {
+        if (submissionError) {
+          // eslint-disable-next-line no-console
+          console.error(submissionError)
```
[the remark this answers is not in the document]
faissalMT: I put this here because unlike the rest of the clientside, everything relating to submission errors was being silently caught which is atypical and will catch people by surprise and made it annoying for me to debug.

src/apps/companies/apps/edit-company/client/EditCompanyForm.jsx
```diff
@@ -33,57 +34,104 @@ function EditCompanyForm({
       return urls.companies.businessDetails(company.id)
     }
 
+    values.address = { ...values.address, area: null }
+    values.area = null
```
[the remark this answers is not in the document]
faissalMT: It's being set twice in the form because all the address fields are present at both levels before submission and it's not clear to me why, so I've kept that pattern to avoid any potential surprises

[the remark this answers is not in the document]
Staberinde: That's fair enough; i'd do the same in that position. We can revisit this pattern as part of a technical debt task

Staberinde (approved):

src/apps/companies/apps/edit-company/client/EditCompanyForm.jsx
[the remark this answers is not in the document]
Staberinde: Sure, I'm not going to block the PR on this.

PR #3960, opened 2021-09-07
Company contacts collection list

paulgain:
## Description of change

Complete rewrite the company contact collection list in React.

## Test instructions

Go to `/companies/<company-id>/contacts`

## Screenshots
Screenshot 2021-09-07 at 08 29 56

## Checklist


src/apps/contacts/client/transformers.js
```diff
@@ -12,53 +12,41 @@ const getTelephoneNumber = (telephone_countrycode, telephone_number) =>
...
-  company,
-  modified_on,
-  archived,
-  company_sector,
-  primary,
-  telephone_countrycode,
-  telephone_number,
-  email,
-} = {}) => {
+export const transformContactToListItem = (companyId) => (contact) => {
```
ian-leggett, line 15: Really small but you could deconstruct "contact" to remove repetition below.

ian-leggett (approved):

ian-leggett:
Looks really good. Left one small comment thats all. I think an old test maybe running as its looking for a class selector `c-entity-list__item` hence the failing test.

ian-leggett (commented):
Looks really good, should we test the payload like we did in the other collection lists? Also I think an old test maybe running as its looking for a class selector `c-entity-list__item` hence the failing test.

src/apps/contacts/client/transformers.js
paulgain: Yep, it was originally deconstructed, however, I feel it's easier to read using the `contact` reference as it reduces the the function line count by 14.

src/client/components/FilteredCollectionList/index.jsx
```diff
@@ -37,6 +37,7 @@ const FilteredCollectionList = ({
   addItemUrl,
   defaultQueryParams,
   titleRenderer = null,
+  width = 'two-thirds',
```
tomkingston, line 40: The investment opportunities page uses the collection list, but didn't need to set a width param (https://www.datahub.trade.gov.uk/investments/opportunities), so might be nice if passing width can be avoided (ie the css works out the width according to the space available), but not worth spending much time on

tomkingston (approved):

src/client/components/FilteredCollectionList/index.jsx
paulgain: Good spot! I've removed the width setting altogether, initially it was set to `two-thirds` as the other third was consumed by the filters, but really there's no need to set the width whatsoever.

PR #3964, opened 2021-09-09
Fix cypress browser size issue

PippoRaimondi:
Fix cypress browser size issue


mforner13 (approved):
Looks good and fixes the issue when I run locally 👍

elcct (approved):

PR #3965, opened 2021-09-09
Display interactions linked to an opportunity

mforner13:
## Description of change

**What the PR does**

This PR adds tab navigation to the page which displays an individual large capital investment opportunity, and in the new `Interactions` tab, displays the interactions linked to the opportunity.

**Implementation details**

- As the tab navigation wraps the content of the opportunity page, I have moved the task which gets the opportunity details to the tabnav component, as well as mounting the tab nav component rather than the opportunities component.
- I have used the react-router `<Route>` component to give certain components access to the `location` and `history` props, rather than connecting the components to this slice of state.
- I have renamed some components related to opportunities, as with the introduction of the tabs, calling the opportunity details table `OpportunityDetails` no longer worked.

**To note**
- The commit `Improve the Resource abstraction`, has been cherry-picked from [this PR](https://github.com/uktrade/data-hub-frontend/pull/3986), and can be removed once that PR is merged.
- Don't be alarmed by the number of changes! Well over 1000 lines are from new fixtures, and more from the aforementioned commit.

## Test instructions

- On the heroku deployment of this branch go to [this opportunity](https://data-hub-fro-feature-di-ccdmez.herokuapp.com/investments/opportunities/503e38aa-d2ba-46f1-9ff6-0dca44b822c1/details)
- Click on the interactions tab
- You should see a list of interactions linked to this opportunity
- To see the pagination there needs to be more than 10 interactions against an opportunity, you can set the `large_capital_opportunity` field on interactions in django admin to see this

### Before

Screenshot 2021-09-22 at 12 04 21

### After

Screenshot 2021-09-22 at 12 04 48

## Checklist


src/apps/investments/views/opportunity-details.njk
```diff
@@ -2,7 +2,7 @@
 
 {% block body %}
   {% component 'react-slot', {
-    id: 'opportunity-details',
+    id: 'opportunity-tab-nav',
```
ian-leggett, line 5: I find the naming of the ID a little confusing as the actual thing that mounts on top of the HTML is more than an tab nav. It feels like the previous ID was named better.

src/client/index.jsx
```diff
@@ -48,7 +48,7 @@ import EventsCollection from '../apps/events/client/EventsCollection.jsx'
 import InteractionsCollection from '../apps/interactions/client/InteractionsCollection'
 import InvestmentProjectsCollection from '../apps/investments/client/projects/ProjectsCollection.jsx'
 import CompanyProjectsCollection from '../apps/investments/client/projects/CompanyProjectsCollection.jsx'
-import Opportunities from '../apps/investments/client/opportunities/Details/Opportunities.jsx'
+import OpportunityTabNav from '../apps/investments/client/opportunities/Details/OpportunityTabNav'
```
ian-leggett, line 51: As per comment around the ID this feels likes its more than just a tab navigation. It mounts something bigger, IMO its the opportunity details

ian-leggett (commented):
Just a couple of comments around the naming of the high level component, it feels like this should be named something like `Opportunities` or `OpportunityDetails`. 

Looking at `src/client/index.jsx` lines 381 down could we combine these:

```
      <Mount selector="#opportunity-tab-nav">
        {(props) => <OpportunityTabNav {...props} />}
      </Mount>
      <Mount selector="#opportunity-details">
        {(props) => <Opportunities {...props} />}
      </Mount>
      <Mount selector="#opportunity-details-header">
        {(props) => <OpportunityDetailsHeader {...props} />}
      </Mount>
```



The other thing is that since you are making the call client side you could have used cy.intercept in the tests rather than sandbox.

src/apps/investments/views/opportunity-details.njk
mforner13: Agreed - have changed to just `opportunity`, rather than `opportunity-details` to avoid confusion with the opportunityDetails component (which now represents the Details tab).

src/client/index.jsx
mforner13: Great point! Have updated the top level 'page' component to be just `Opportunity`.

test/functional/cypress/specs/investments/opportunity-interactions-spec.js
```diff
@@ -0,0 +1,105 @@
...
+    it('should show the current and number of pages', () => {
+      cy.contains('Page 1 of 2').should('be.visible')
+    })
+    it('should allow sort by oldest interactions', () => {
+      cy.get('[data-test="sortby"] select').select('date')
+      cy.url().should('contain', '?sortby=date')
+    })
+    it('should allow sort by recently created interactions', () => {
+      cy.get('[data-test="sortby"] select').select('-date')
+      cy.url().should('contain', '?sortby=-date')
```
mforner13, line 83: @ian-leggett is it here that I could use `cy.intercept` on the GET request rather than check the url? :)

ian-leggett (approved):

test/functional/cypress/specs/investments/opportunity-interactions-spec.js
ian-leggett: No it would be close to when you visit the url, you then stub it with `intercept` something like 
```
    cy.intercept('POST', ENDPOINT GOES HERE).as('apiRequest')
      cy.visit(URL GOES HERE)
      assertPayload('@apiRequest', JSON RESPONSE GOES HERE) << this is used instead of sandbox
```

We can pair when you get back if you like.

src/client/components/CollectionList/index.jsx
```diff
@@ -53,14 +54,16 @@ const CollectionList = ({
...
+          {includeDownloadHeader && (
+            <RoutedDownloadDataHeader
+              count={count}
+              maxItems={maxItemsToDownload}
+              data-test="download-data-header"
+              baseDownloadLink={baseDownloadLink}
+              entityName={entityName}
+              entityNamePlural={entityNamePlural}
+            />
+          )}
```
dredmonds, line 66: I just simply notice, `includeDownloadHeader` was always set to `true`. Then implemented the logic `RoutedDownloadDataHeader` to renders if `true`. I think doesn't make sense to put a logic because it always `true`.

mforner13: This line in the props `includeDownloadHeader = true` sets the default value to `true`, meaning that it can be overridden and set to false when using the CollectionList component. I needed to set the default to be true, as all collection lists so far are using the download header, so without a default I would need to set the prop to true in all existing instances of the component :)

[Here's where I set it to false](https://github.com/uktrade/data-hub-frontend/blob/c7d98542bd44f2e63b463167c35b4e82aab67d7c/src/apps/investments/client/opportunities/Details/OpportunityInteractions.jsx#L53) in the collection list I add in this PR

dredmonds (approved):

ian-leggett (approved):

PR #3963, opened 2021-09-08
Project summary filters

tomkingston:
Superseded by #3984 

## Description of change

Makes the investment project summary component stay consistent with the filters picked in the listing component by using the new "summary" data from the api endpoint.

## Test instructions

With the personalised dashboard user feature flag enabled, go to the dashboard and select filters for investment projects - the chart should update to reflect the selected filters. It is also worth checking what happens when you navigate to the companies list tab and back again.

### Before

image

image

### After

image

image

## Checklist

[//]: # "When submitting a PR make sure the code review guidelines have been satisfied.
https://github.com/uktrade/data-hub-frontend/blob/master/docs/Code%20review%20guidelines.md"

- [ ] Has the branch been rebased to master?
- [ ] Automated tests (Any of the following when applicable: Unit, Functional or End-to-End)
- [ ] Manual compatibility testing (Browsers: Chrome, Firefox, IE11, Safari)

PR #3984, opened 2021-09-14
Investment summary links

tomkingston:
## Description of change

Everything from https://github.com/uktrade/data-hub-frontend/pull/3963

Plus:
- Implements a new custom legend for the investments summary pie chart
- Adds updated colours to pie chart

## Test instructions

See https://github.com/uktrade/data-hub-frontend/pull/3963

## Screenshots

image

## Checklist


src/client/components/MyInvestmentProjects/InvestmentProjectSummary.jsx
```diff
@@ -0,0 +1,45 @@
...
+  return {
+    summary:
+      data &&
+      data.map(({ label, value, ...rest }, index) => ({
+        ...rest,
+        colour: segmentColours[index % segmentColours.length],
+        value,
+        id: label,
+      })),
+  }
```
paulgain, line 26: I find this quite difficult to read, clearly there's name clashing occurring, hence the aliasing, it would be great to rename the variables/constants that are causing the problem. Not sure why we have to test for `data` (L19), ideally we would ensure both `summary` and `unfilteredSummary` are set to an empty array within the `initialState` object defined within the reducer.

tomkingston: I'll see what I can do - the name clashing is because the two summaries come from exactly the same endpoint, but one of them does not have any filters applied - I might be able to do something in the check investments reducer though.

ian-leggett: I agree this does get a bit confusing. I have created alias's throughout and put some meaning behind them so you know what it is you are pulling out of the store. Hopefully this reads better.

ian-leggett (approved):

paulgain (approved):

PR #4123, opened 2021-11-01
Remove these preset plugins as they are part of ES2018 and ES2020

craigjbass:
## Description of change

These plugins for babel are in @babel/preset-env in ES2018 and ES2020 so shouldn't be needed separately.

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,8 +66,6 @@
   "dependencies": {
     "@babel/core": "^7.15.8",
     "@babel/helper-call-delegate": "^7.12.13",
-    "@babel/plugin-proposal-object-rest-spread": "^7.15.6",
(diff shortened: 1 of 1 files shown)


cgsunkel (approved):

Staberinde:
@craigjbass Did you intentionally generate a v2 lockfile rather than v1 here? My only concern is that I think v2 is only compatible with npm > 7.0.0 which is a requirement some people (including myself) may not currently meet.

Staberinde (approved):
Looking into this it looks like v2 is backwards compatible with v1. Strange that there are no deletions given that you're removing two packages, but maybe these are captured in the lockfile still even though part of the standard lib

PR #4124, opened 2021-11-01
TaskForm component

peterhudec:
## Description of change

Adds the new `TaskForm` component, which should behave exactly as the existing `FormStateful` and `MultiinstanceForm` in terms of its relationship with the form fields (note, thtat the two existing form components are not fully compatible with each other). Where the `TaskForm` differs is in how it handles the form submission.

The `TaskForm` doesn't have an `onSubmit` property, but instead starts a _task_ when submitted with the form values passed to the task as payload. The form will render a `LoadingBox` overlay on top of the form while the _task_ is in progress, which also handles the rejection of the task. The form can also resolve it's initial values with a task. For more details refer to the [component's docstring](https://github.com/uktrade/data-hub-frontend/blob/component/task-form/src/client/components/Task/Form/index.jsx#L289-L340) and [Storybook examples](https://github.com/uktrade/data-hub-frontend/blob/component/task-form/src/client/components/Task/Form/__stories__/TaskForm.stories.jsx#L12).

Apart from the `TaskForm`, the PR also:

- Improves the `Analytics` component
- Introduces the `Effect` component
- Wraps the `ErrorSummary` component in `React.forwardRef`
- Fixes incompatibilities in the `MultiinstanceForm`
- Improves the `HardRedirect` component
- Improves the `Task/LoadingBox` component
- Allows the `TaskStatus` error view to be _dismissable_
- Introduces the `Wrap` component

This PR doesn't include application of the component to keep the (already big) size of the PR manageable, but you can see it applied here:

-  #4127
- #4128
- #4143
- #4142


src/client/components/Analytics/types.d.ts
```diff
@@ -0,0 +1,15 @@
...
+  action: string,
+  label: string,
+  extra?: Record<string, any>,
+) => void
+
+export type Props = {
+  children: (pushAnalytics: PushAnalytics) => JSX.Element
+}
+
+export type Analytics = (props: Props) => JSX.Element
```
vfarah-if, line 15: Peter this may be a mistake that you have the file with a typescript extension, but I love it :)

peterhudec: No, it's deliberate. These types are referenced in the `TaskForm` docstring and as a result, you will have nice intellisense for the props in VSCode. It's a non-invasive usage of typescript.

src/client/components/Task/utils.js
```diff
@@ -19,8 +19,8 @@ export const delay = curry((duration, task, payload) =>
   ]).then(([result]) => result)
 )
 
-export const catchApiError = ({ response: { data } }) =>
-  Promise.reject(data.detail || data)
+export const catchApiError = ({ response }) =>
+  Promise.reject(response?.data?.detail || response?.data)
```
vfarah-if, line 23: Should this not cater for **response?.data?.detail || response?.text || response?.statusText || message** which will resolve a server side failure we had with field validation on Event

vfarah-if (commented):
Let's chat more on Monday about some ideas and exploration about what I was questioning. Maybe we need edge case tests as cypress unit tests?

craigjbass (approved):
We can safely ignore codecov job failures, they need relaxing.

ian-leggett (approved):

vfarah-if (approved):

PR #4127, opened 2021-11-02
Refactor the Add company form with TaskForm

peterhudec:
## Description of change

Refactors the _add company form_ with the `TaskForm` component.

## Test instructions

1. Go to `/companies/create`
2. Tweak the [`postAddDnbCompany `](https://github.com/uktrade/data-hub-frontend/blob/master/src/apps/companies/apps/add-company/controllers.js#L68-L83) request handler of the form's 
2. Try out all the possible user journeys of the form
3. They should all work as before
4. When you submit the form, the POST request should fail
5. You should see an error message overlaid on top of the form

### Before

No error handling of rejected requests
Screen Shot 2021-11-02 at 1 36 47 PM

### After

Screen Shot 2021-11-02 at 1 31 58 PM

### DIFF
--- a/src/apps/companies/apps/add-company/client/AddCompanyForm.jsx
+++ b/src/apps/companies/apps/add-company/client/AddCompanyForm.jsx
@@ -4,13 +4,11 @@
 
 import React, { useEffect } from 'react'
 import PropTypes from 'prop-types'
-import axios from 'axios'
 import { get } from 'lodash'
 
 import {
   FieldRadios,
   FieldSelect,
-  FormStateful,
   Step,
 } from '../../../../../client/components'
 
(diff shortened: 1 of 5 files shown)


paulgain:
Hi @peterhudec,

When going all the way through to the last page of the 'Add company' journey the DIT region is looked up and set, however, the form is unaware as it incorrectly displays a validation error:

unaware-region-is-selected
.

peterhudec:
> Hi @peterhudec,
> 
> When going all the way through (both company lookup and manual company input) to the last page of the 'Add company' journey the DIT region is looked up and set, however, the form is unaware as it incorrectly displays a validation error:

Thanks @paulgain I have noticed it too. The problem was in how the field's `initialValue` prop was treated by the form. It is now fixed.

paulgain (approved):

tomkingston (approved):

PR #4159, opened 2021-11-17
Proof of concept mocha based unit tests.

craigjbass:
## Description of change

Adds mocha-based unit tests that can render react components against jsdom.

For this level of testing I would expect us to be testing behaviours i.e. "when this button was clicked, this text is now visible", and not visual aspects e.g. "this button is red".

*Update:* I noticed we're using styled-components in conjunction with variables and if statements - which means that often the styling applied is "behaviour". In 84b7c19 I added a PoC of how to assert computed styles

[ ] Add to circle ci jobs

### DIFF
--- a/test/component/mocha/Badge.test.js
+++ b/test/component/mocha/Badge.test.js
@@ -1,0 +1,43 @@
+import React from 'react'
+import Badge from 'client/components/Badge'
+import { act } from 'react-dom/test-utils'
+import { expect } from 'chai'
+import ReactDOM from 'react-dom'
+import './bootstrap'
+
+const mount = (component) => {
+  const container = document.createElement('div')
+  document.body.appendChild(container)
+  act(() => {
+    ReactDOM.render(component, container)
+  })
+  return container
+}
+
+describe('Badge', () => {
+  const badgeText = 'example'
+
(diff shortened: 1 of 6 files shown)


test/component/mocha/test.test.js
```diff
@@ -0,0 +1,29 @@
...
+
+require('./bootstrap.js')
+
+const SomeJSX = (props) => <button onClick={props.onClick}></button>
+
+describe('SomeJSX', () => {
+  it('should be clickable', (done) => {
+    var a = 0
+
+    const container = document.createElement('div')
```
craigjbass, line 14: I think a lot of the boilerplate here can be tidied up

test/component/mocha/Badge.test.js
```diff
@@ -0,0 +1,43 @@
+import React from 'react'
+import Badge from 'client/components/Badge'
+import { act } from 'react-dom/test-utils'
+import { expect } from 'chai'
+import ReactDOM from 'react-dom'
+import './bootstrap'
+
+const mount = (component) => {
```
craigjbass, line 8: I'm not sure about this as an abstraction, I just wanted to demonstrate that it's possible!

PR #4190, opened 2021-12-03
Navigate events using react router mostly

vfarah-if:
## Description of change

Test react router and events where there should only be one server side request and all functionality should work like it used to

## Test instructions

Working with local reloads like before

### Before

image

### After

image

## Checklist


src/client/modules/Events/EventDetails/index.jsx
```diff
@@ -60,37 +60,38 @@ const EventDetails = ({
...
-        payload: eventId,
+        payload: id,
         onSuccessDispatch: EVENTS__DETAILS_LOADED,
       }}
     >
-      {() =>
-        name && (
-          <>
-            <LocalHeader
+      {() => {
```
vfarah-if, line 74: Remove the brace once the flashMessage details have been resolved

src/client/modules/Events/EventDetails/index.jsx
```diff
@@ -45,8 +42,11 @@ const EventDetails = ({
   service,
   archivedDocumentsUrlPath,
   disabledOn,
-  flashMessages,
 }) => {
+  const { id } = useParams()
+  const { state } = useLocation()
+  // const id = 'c8e6dc0e-4a1d-41f0-ac0a-fed19c776fa9'
+  // const state = undefined
```
vfarah-if, line 49: Remove when done

src/apps/events/controllers/events.js
```diff
@@ -1,4 +1,4 @@
-const renderEventsView = async (req, res, next) => {
+const renderReactifiedEventsView = async (req, res, next) => {
```
tomkingston, line 1: Since this will remain the only events view I think you should just leave this as `renderEventsView` to avoid any confusion.

src/client/modules/Events/EventDetails/tasks.js
```diff
@@ -1,7 +1,7 @@
 import { apiProxyAxios } from '../../../components/Task/utils'
 import { transformResponseToEventDetails } from '../transformers'
 
-export const getEventDetails = (eventId) =>
+export const getEventSummaryDetails = (eventId) =>
```
tomkingston, line 4: Why is this now called `getEventSummaryDetails` - why not keep it as `getEventDetails`? The api endpoint doesn't suggest this is any different to standard event details. (Same for the related task names etc).

src/client/modules/Events/EventForm/EventFormFields.jsx
```diff
@@ -20,7 +20,7 @@ import { validateStartDateBeforeOrEqualToEndDate } from './validators'
 import urls from '../../../../lib/urls'
 
 export const EventFormFields = ({ values }) => {
-  return values && values.metadata ? (
+  return (
```
tomkingston, line 23: One of Ian's favourite PR comments was always to make these implicit returns (ie `export const EventFormFields = ({ values }) => (<>...</>)`... however, that does mean changing all the indentation as soon as you want to do a bit more than that in the future, so I'm not bothered either way.

src/client/modules/Events/EventForm/__stories__/EventForm.stories.jsx
```diff
@@ -14,5 +15,9 @@ storiesOf('Task/Form/Event Form/Update Event Form', module)
     options: { theme: undefined },
   })
   .add('Edit event', () => (
```
tomkingston, line 17: I don't think the EventForm should really be in storybook which is I would expect to be reserved for re-usable components rather than 'modules' - storybook could end up quite cluttered with all the modules.... if you do want it, perhaps you could add it to a `modules` namespace so that it doesn't appear with all the components etc.

tomkingston:
Noticed this when clicking through to an event....
image
The header / layout etc should stay visible after the click.

src/client/modules/Events/EventDetails/tasks.js
vfarah-if: This is ambiguous with EventDetails in another space and seemed clearer to understand because it is event collection, event summary, event details so this helped to make a clear difference between the two

src/client/modules/Events/EventForm/__stories__/EventForm.stories.jsx
vfarah-if: Okay cool I will move it to Modules. I liked this because it is an isolated space for doing this in a standalone environment and is a good example for doing this,

src/client/modules/Events/EventForm/EventFormFields.jsx
vfarah-if: Fixed to follow standard

src/apps/events/controllers/events.js
vfarah-if: The main reason for this is to move this out and share it for all things that use react as this will be the same default template going forward, but no problem I can put that the way it was

src/client/modules/Events/EventDetails/tasks.js
vfarah-if: Rolledback

tomkingston (approved):

paulgain (approved):
I really don't like the idea of passing `useReactRouter` down the collection list, I also understand that we're moving from one tech stack to another, I guess we'll have to swallow this until everything has migrated across at which point we'll have an SPA and can then remove the conditionals currently going in.

paulgain (changes requested):
**Just come across this issue when adding an attendee to an event**

unable-to-add-attendee

src/client/components/Task/Form/index.jsx
```diff
@@ -212,7 +216,8 @@ const _TaskForm = ({
                                           redirectMode === 'soft' &&
                                             redirectTo &&
                                             history.push(
-                                              redirectTo(result, values)
+                                              redirectTo(result, values),
+                                              { messages: getMessages() }
```
mforner13, line 220: Could you just use the message that's been passed in on the flashMessage prop from line 210 here, rather than calling session storage? I.e:
`{messages: message}`

src/client/modules/Events/EventForm/index.js
```diff
@@ -0,0 +1,81 @@
+import React from 'react'
```
mforner13, line 1: Seems like this file was accidentally turned to `.js` from `.jsx`

src/client/components/Task/Form/index.jsx
vfarah-if: I was worried about throwing away other potential message, so to guarantee this does not happen, I take it from the queue that can be generated from other tabs or stuff

mforner13 (approved):
Looks good pending Paul's approval :)

paulgain (approved):

PR #4198, opened 2021-12-06
Remove vue.js :question:

craigjbass:
## Description of change

Does this pass the tests

That is the question


craigjbass:
Basically, no

PR #4201, opened 2021-12-06
Accessible multi/single-select component

tomkingston:
**Note: since this is on storybook only, it is not possible to write any functional tests for it, so the test coverage is low. I am currently looking at adding a few component-level tests, but think it would be best to do that in another PR to keep the size of this one down (obviously that's not very TDD, but should make this a bit more manageable)**

## Description of change

Adds an accessible multi / single select typeahead to storybook. This is based on accessible multi-select components designed by Sarah Higley from Microsoft here: 

*Research*:
https://www.24a11y.com/2019/select-your-poison-part-2/ 

*Multiselect*:
https://codepen.io/smhigley/pen/GRgjRVN

*Single select*:
https://codepen.io/smhigley/pen/BayzXbO

This is intended to replace the existing typeahead which is not very accessible. 

This is a work in progress being done in multiple stages and will eventually replace all the existing typeaheads but for now it is just on storybook. 

As part of the next stage, this should be combined with the work on the Task Typeahead to load options from an API.

## Test instructions

Go to storybook and check the Typeahead2 stories. Make sure that these are accessible.

## Screenshots

image

image

## Checklist


vfarah-if (approved):

LaurenQurashi (approved):
This looks so good! Thank you for taking on the challenge 🎉

PR #4222, opened 2021-12-14
Refactor AddRemoveListForm to use TaskForm

faissalMT:
## Description of change

Refactors AddRemoveListForm from using FormStateful to use the new TaskForm

## Test instructions
Adding and removing from a list should continue to work as normal.

## Checklist


src/client/components/Task/Form/index.jsx
```diff
@@ -66,6 +66,7 @@ const _TaskForm = ({
   touched = {},
   steps = [],
   initialStepIndex = 0,
+  submitButtonProps = {},
```
mforner13, line 69: If we keep this in and don't add the data-tests to the taskform as standard, could you add this to the docstring which gives explanations for all props in this file, and also to the Props object in the typescript file? :)

src/client/components/CompanyLists/AddRemoveFromListForm.jsx
```diff
@@ -4,70 +4,83 @@ import Link from '@govuk-react/link'
...
+      submissionTaskName="Add or remove from list"
+      initialValues={initState}
+      transformPayload={(list) => ({ list, token, companyId })}
+      redirectTo={() => cancelLinkUrl}
+      submitButtonProps={{ 'data-test': 'submit-button' }}
+      actionLinks={[
+        {
+          children: 'Cancel',
+          href: cancelLinkUrl,
+          linkProps: { 'data-test': 'cancel-button' },
```
mforner13, line 37: What do you think about adding these as standard to the task form rather than creating new props to pass in? :) I know the text can change on the button/link themselves, but they're always serving the same purpose and seems like it would be useful to be able to access them by their data-test all the time

mforner13 (commented):
Let me know what you think :)

cgsunkel (approved):

src/client/components/CompanyLists/AddRemoveFromListForm.jsx
faissalMT: Hmm, yeah that sounds like a good idea given this pattern seems so common, it'll probably also make testing simpler and the props more uniform

mforner13 (approved):
👍

PR #4231, opened 2021-12-17
Remove old form components

cgsunkel:
## Description of change

Now that all the forms have been refactored to use `TaskForm`, the components `Form`, `FormStateful` and `MultiInstanceForm` can now be removed. As part of this work I've also refactored the various Storybook stories using `FormStateful` to use `TaskForm` instead.

## Checklist


src/client/components/Resource/index.jsx
```diff
@@ -97,7 +97,6 @@ Resource.propTypes = {
   name: PropTypes.string.isRequired,
   id: PropTypes.string.isRequired,
   children: PropTypes.func,
-  taskStatusProps: PropTypes.shape(_.omit(Task.Status.propTypes, 'name', 'id')),
```
cgsunkel, line 100: This line has been removed as it was preventing the FE from loading once `MultiInstanceForm` had been removed from the provider.

docs/Data Fetching Building Blocks.md
```diff
@@ -243,7 +243,7 @@ it's options, the form will not show a validation error for the field even thoug
 the field is required and the form will be submitted.
 
 This due to the mechanism of how the form communicates with its fields. The form
```
mforner13, line 245: Not from your changes, but since we're updating this paragraph - there's a typo in the first line `This is due to...`

src/client/components/Form/elements/FieldWrapper/usage.md
```diff
@@ -1,4 +1,4 @@
-FieldInput
+FieldWrapper
```
mforner13, line 1: Good spot 👍

src/client/components/Task/Form/index.jsx
```diff
@@ -283,9 +307,12 @@ const _TaskForm = ({
...
                             {/*
                             We don't want to render the submit button when the form
-                            has multiple steps as the steps come with a built-in submit button
+                            has multiple steps as the steps come with a built-in submit button.
+
+                            We also don't want the submit button to show up in Storybook stories
+                            for individual form fields.
                             */}
-                            {!steps.length && (
+                            {!steps.length && !inStorybook && (
```
mforner13, line 315: The field element stories currently show a submit button, so it seems like instead of adding this new prop, you could just remove the submit buttons from the stories and let the default 'Save' button show? I think better than creating logic just for showing in storybook.

src/client/components/Form/elements/__stories__/FieldSelect.stories.jsx
```diff
@@ -74,5 +78,5 @@ storiesOf('Forms/Select', module)
           <pre>{JSON.stringify(form, null, 2)}</pre>
```
mforner13, line 78: E.g. remove `<Button>Submit</Button>`. It does mean the save button shows below the JSON that's showing the state of the form, but I don't think that matters tbh :)

src/client/components/Resource/index.jsx
mforner13: I don't think we'd want to remove this all together as then we'll be missing some key proptypes. The resource and task status shouldn't be affected by MultiInstance - do you know what specifically caused this to no longer work?

cgsunkel: This was failing as these props are coming through as undefined. There are two variants of this line (the second is `taskStatusProps: PropTypes.shape(Task.Status.propTypes)`) and both resulted in the same error.

tomkingston (approved):

src/client/index.jsx
```diff
@@ -3,6 +3,7 @@ import ReactDOM from 'react-dom'
 import * as Sentry from '@sentry/browser'
 import { Switch } from 'react-router-dom'
 
+import './components'
```
cgsunkel, line 6: This import has been added as otherwise Webpack isn't able to find the components folder. It worked previously as the provider was importing `MultiInstanceForm` from the main directory instead of the form folder within it.

However, now that the `MultiInstanceForm` import has been removed we need to import the directory here instead.

mforner13 (approved):
Great work cleaning it all up! 👍 🎉

PR #4247, opened 2021-12-23
`Form` refactoring cleanup

cgsunkel:
## Description of change

This PR cleans up a few bits and pieces left (mosly by me) during the Form refactoring work, which are listed below:

- When the component was renamed, the actions weren't renamed to match. The action names have been changed from `TASK_FORM__` to just `FORM__`.
- When the `CreateUKInvestmentOpportunity` form was refactored, a reference to the old reducer was left in the provider. This has been removed as the reducer it was referring to has been deleted.
- The formatting of the Storybook stories was inconsistent with the rest of Storybook. This has been fixed, and I also added a description and list of props to the basic examples story.
- The `actionLinks` prop is no longer required, so it has been removed.

## Checklist


src/apps/my-pipeline/client/PipelineForm.jsx
```diff
@@ -152,7 +152,8 @@ PipelineForm.propTypes = {
   redirectTo: PropTypes.func.isRequired,
   flashMessage: PropTypes.func.isRequired,
   submitButtonLabel: PropTypes.string,
-  actionLinks: PropTypes.array,
+  cancelRedirectTo: PropTypes.func.isRequired,
```
faissalMT, line 155: `cancelRedirectTo` should not be required as the cancel button will hide itself if it is not specified.

src/client/components/Form/__stories__/basic-example.md
```diff
@@ -75,4 +92,25 @@ This is a basic example of all the capabilities of `Form`. It:
...
+
+### Properties
+Prop | Required | Default | Type | Description
+:--- | :------- | :------ | :--- | :----------
+`analyticsFormName` | true | `````` | string | A string to identify the form in Google Analytics events. Should be in camelcase format e.g. editCompany (this is by request of the performance analysis team)
+`id` | true | `````` | string | ID of the task that should be started when the form is submitted and when initial values are loaded
+`submissionTaskName` | true | `````` | string | Name of the task that should be started when the form is submitted
+`analyticsData` | false | `````` | object | A function which takes the values of the form as an argument, and which returns an object containing additional - non-sensitive - data to be passed to Google Analytics.
+`cancelButtonLabel` | false | Cancel | string | The label for the built-in cancel button
+`cancelRedirectTo` | false | `````` | string | A function which returns a URL to redirect to when the cancel button is clicked. If unset the cancel button will not display.
```
faissalMT, line 105: The type here should be a function

faissalMT (approved):
Just two minor things, looks good otherwise

mforner13 (approved):

PR #4252, opened 2021-12-31
Fixed company creation when telephone number is empty

ambrozic:
## Description of change

When `telephone_number` field was submitted as empty string, dnb service would return `400 Bad Request` because serializer does not allow blank values. Fix now removes `telephone_number` field from payload when value is empty string.

Jira ticket [CPS-290](https://uktrade.atlassian.net/browse/CPS-290)

### DIFF
--- a/src/apps/companies/apps/add-company/client/AddCompanyForm.jsx
+++ b/src/apps/companies/apps/add-company/client/AddCompanyForm.jsx
@@ -4,7 +4,7 @@
 
 import React, { useEffect } from 'react'
 import PropTypes from 'prop-types'
-import { get } from 'lodash'
+import { get, omit } from 'lodash'
 
 import {
   FieldRadios,
@@ -66,6 +66,13 @@ function AddCompanyForm({
     return {}
   }
 
+  function transformPayload(values) {
+    return {
+      ...omit(values, [!values?.telephone_number && 'telephone_number']),
+      csrfToken,
+    }
+  }
+
   return (
     <Form
       id="add-company-form"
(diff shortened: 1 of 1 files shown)


tomkingston (approved):

elcct (approved):

src/apps/companies/apps/add-company/client/AddCompanyForm.jsx
```diff
@@ -66,14 +66,21 @@ function AddCompanyForm({
     return {}
   }
 
+  function transformPayload(values) {
+    return {
+      ...omit(values, [!values?.telephone_number && 'telephone_number']),
```
craigjbass, line 71: This took me a moment to understand what it's doing. The use of && to passthru a truthy string was astonishing to me, (even though || to do the same in Ruby is kinda normal), combined with a not in there. I got there in the end, but I do wonder whether this could be written in a way that would be more easily understood by a mixed seniority team.

That said I'm not blocking this PR.

```suggestion
      ...omit(values, values?.telephone_number ? [] : ['telephone_number']),
```

craigjbass (approved):

src/apps/companies/apps/add-company/client/AddCompanyForm.jsx
ambrozic: i used array spreading primarily because it seems to be quite common throughout this code repository. kids this days spread their arrays like nothing.

craigjbass: scary

PR #4260, opened 2022-01-04
Replace typeaheads - swap out old typeaheads for the new one

tomkingston:
## Description of change

Uses the new accessible typeahead instead of the old ones. 

Note that the old typeahead is still in place in a few places around datahub - these can be replaced in a later ticket once we know people are happy with these on the collection list pages.

This also required a few changes ot the new typeahead to deal with RoutedTypeaheads / onChange events etc.

## Test instructions

Go to the various collection list pages and try out the typeaheads. These should now be using the accessible typeaheads.

### Before

image

### After

image

## Checklist


test/sandbox/routes/v4/metadata/index.js
```diff
@@ -160,7 +160,13 @@ exports.ukRegion = function (req, res) {
 }
 
 exports.administrativeArea = function (req, res) {
-  res.json(administrativeArea)
+  res.json(
```
tomkingston, line 163: Added this so that the dom doesn't render hundreds of admin areas in dropdowns (slowing it down a lot) - this is now more consistent with the rel behaviour

src/client/components/Typeahead2/Typeahead.jsx
```diff
@@ -164,15 +166,17 @@ const Typeahead = ({
...
 }) => {
+  const initialValue = value || defaultValue
   useEffect(() => {
     onInitialise({
       isMulti,
-      value: value || defaultValue,
+      value: initialValue,
     })
-  }, [])
+  }, [JSON.stringify(initialValue), isMulti])
```
mforner13, line 179: What does passing `JSON.stringify(initialValue)` do over just passing `initialValue` here out of interest?

src/client/components/RoutedAdvisersTypeahead/index.jsx
```diff
@@ -10,18 +10,20 @@ import Task from '../Task'
 import { parseAdviserData } from '../../../common/formatAdviser'
 
 const fetchAdvisers = (onlyShowActiveAdvisers) => {
-  return throttle(
-    (searchString) =>
-      axios
+  return throttle((searchString) => {
+    if (searchString.length) {
```
mforner13, line 14: I think this logic needs to be added to the `fetchTeams` function in `RoutedTeamsTypeahead` too. I was just testing out the filters on the different collection pages and the Teams filter on the interactions collection page is loading all options on click into the input field which is very slow, rather than waiting for the user to enter a search string (which is the current behaviour for this field).

mforner13 (commented):
Just one suggestion and a question :)

src/client/components/Typeahead2/Typeahead.jsx
tomkingston: Ahh, yes, I was having an issue here and this was the fix. The problem was to do with deep comparisons - since initialValue is an Array of items, the array coming in was actually a different instance with exactly the same items.... therefore the effect was being triggered for every re-render. By stringifying it, the code knows that the value is actually the same.

tomkingston: Stackoverflow on it here: https://stackoverflow.com/questions/59467758/passing-array-to-useeffect-dependency-list

src/client/components/RoutedAdvisersTypeahead/index.jsx
tomkingston: Ahhh, thanks, I will check that out

tomkingston: Should be fixed now 👍

mforner13 (approved):
Looks good to me 👍

LaurenQurashi (approved):

PR #4263, opened 2022-01-07
Add sagas to handle flash messages in session storage

LaurenQurashi:
## Description of change

This PR updates the way that the frontend interacts with session storage in order to pass flash messages between pages.

### The problem we're solving

Previously, the `FlashMessages` component got messages directly from the session and was also responsible for clearing them. This caused issues because when submitting a form on a page which used the react local header (containing `<FlashMessages>`), clearMessages() would be called as the local header is re-rendered before redirecting, thus removing the flash message from session storage before it was rendered on the next page.

### Our solution

This PR fixes this issue, by using redux sagas to handle side effects - ie. interactions with our session storage.

One saga takes an action dispatched on successful submission of the form and adds the message from the body of that action to the session storage.

The other saga awaits an action to be dispatched (which happens on render of the FlashMessage component within LocalHeader), then dispatches an action to add the message from session storage into the state.

To account for when a redirect is made via react router (so there is no page refresh), the useEffect hook which dispatches the action in the FlashMessages component has a cleanup method which clears the flashMessages section of state when the component is unmounted.

### TO NOTE Testing

We have not added any additional tests in this PR, as the expected behaviour of the flash message component hasn't changed. The existing functional tests check that messages appear after form submissions, but we have no tests to check the implementation detail. 

Any thoughts on whether we should add tests to check our implementation would be greatly appreciated! 

## Test instructions

No visual changes.

You can check the behaviour of the flash messages when redirecting with react router by:
1. Editing an event
2. Checking the flash message appears on the event details page
3. Click the 'Edit event' button
4. Check the flash message no longer appears in the local header

## Checklist


src/client/components/LocalHeader/reducer.js
```diff
@@ -0,0 +1,17 @@
+import {
+  FLASH_MESSAGE__ADD_TO_STATE,
+  FLASH_MESSAGE__CLEAR_FROM_STATE,
+} from '../../actions'
+
+export default (state = {}, { type, flashMessages }) => {
+  switch (type) {
+    case FLASH_MESSAGE__ADD_TO_STATE:
+      return {
```
tomkingston, line 9: I suppose this works because you only have flashMessages in this store, but perhaps it would be more robust to include other parts of state ie, something like this?

```
export default (state = {flashMessages: []}, { type, flashMessages }) => {
...
case FLASH_MESSAGE__ADD_TO_STATE:
    return {
        ...state, 
        flashMessages,
    }
case FLASH_MESSAGE__CLEAR_FROM_STATE:
    return {
        ...state,
        flashMessages: []
    }

tomkingston (approved):

src/client/components/LocalHeader/reducer.js
mforner13: Agreed!

cgsunkel (approved):

PR #4276, opened 2022-01-11
Replace remaning typeaheads

tomkingston:
## Description of change

- Replaces the FieldTypeaheads to use the new accessible Typeahead
- Swaps out the add another form component on the add interaction form for an accessible multi-select typeahead

## Test instructions

Form typeahead should work as expected.
The "Related named trade agreement(s)" field on the interation form (/interactions/<id>/edit) should show a multi-select typeahead when "Yes" is selected. Options should be saved as expected.

### Before

image

### After

image

## Checklist


src/apps/my-pipeline/client/PipelineForm.jsx
```diff
@@ -36,101 +36,99 @@ function PipelineForm({
...
-          <FieldDate
-            format="short"
-            label="Expected date for win (optional)"
-            hint="For example 11 2020"
-            name="expected_win_date"
-          />
-        </>
-      )}
-    </Form>
+    <div className="govuk-!-width-two-thirds">
```
LaurenQurashi, line 39: Out of pure interest, how come you've had to add this in?

LaurenQurashi (approved):

src/apps/my-pipeline/client/PipelineForm.jsx
tomkingston: The old code was adding `className="govuk-!-width-two-thirds"` to each of the inputs which not only meant more code, but also messed with the styling of the new typeaheads. By setting the container width all the child elements just expand to that size.

paulgain: Although this is very convenient, should we really be using `className` to layout elements? Instead, can we use a set of grid system components such as [this](https://govuk-react.github.io/govuk-react/?path=/docs/grid-col--column-two-thirds-one-third). As you know, we already have this as a dependency. It would be good to get a general consensus on how we deal with laying out components going forward.

paulgain (commented):
Looks good Tom, just a couple of comments.

src/apps/my-pipeline/client/PipelineForm.jsx
tomkingston: I've reverted to the old code which applies this override to each input (and amended the typeahead so that this doesn't break it). I do think that using override classes is also quite messy but we can wait for some time in the future to implement some layout components like you mentioned.

paulgain (approved):

PR #4289, opened 2022-01-17
Add investment user journey rewrite

paulgain:
## Description of change
This is a rewrite of the add investment user journey with many improvements (see below). We've also added a new "Add investment project" button on the `/investments` page.

Both user journeys are comprised of the same form and a number of steps, at each step investment details are collated, temporarily stored and validated before going onto the next step, there is a max of three steps from the `/investments` page and only two steps from the `/companies` page. The last step on both user journeys contains the majority of form components, once completed the user submits the investment to the API, upon success, they are redirected to the investment details page: `/investments/projects/<investment-id>/details`

### Test

User journey from `/investments`
1. Click "Add investment project"
2. Search for a foreign source of equity (STEP 1)
3. From the results select a company
4. Choose the investment type and continue (FDI/Non-FDI/Commitment to Invest) (STEP 2)
5. Fill out all form fields and validate (FE form validation) by submitting the form (STEP 3)

User journey from `/companies`
1. Search for a foreign source of equity (`e.g. /companies/<foreign-company-id>`)
2. Click the Investment tab
3. Click "Add investment project"
4. Identical to number 4 above (STEP 1)
5. Identical to number 5 above (STEP 2)

### Improvements
1. Removed the 'Add investment project' button from all UK companies as you can't add an investment to a UK company, from UR users always search for a foreign company and then add an investment to the company. We've removed the same button in Production, we've hidden it behind a feature flag, released on the 13th January and no one has complained thus far, which so far confirms our analysis and suspicions that it's simply unused.
2. Removed the 'Will this company be the source of foreign equity investment?' question after the user has selected the investment type. From observing UR sessions and speaking with Yvonne users always select 'Yes', we think the question is redundant and can be safely removed.
3. Improved form validation, at present the user receives generalised validation error messages such as 'This field may not be blank' or 'This field is required', now they receive very specific error messages 'Enter a project name', 'Enter a project description' which improves the UX when things go wrong. We've also improved the validation around the estimated land date and actual land date. 
4. Rolled out type aheads allowing users to search for data quicker.
5. As we're moving over to a SPA, you'll notice there are no more page reloads allowing users to add investments quicker.
6. Significantly better search results after switching the endpoint, we now go via Elasticsearch where the results actually make sense.
7. Contains the latest type aheads which improves accessibility from previous versions, the selected options now appear as labels directly above the field
8. Reduction of tech debt
9. Improved UX especially when choosing a contact whilst on a user journey. Previously, if a company didn't have a contact you would have to create one separately in a new browser tab, go back to the first tab and refresh the page pulling in the newly created contact. Now you have a link that takes you to the add a contact page, after completion you are redirected back to where you came from making it easy for users to continue on their journey. 

**Unfortunately this a fairly large PR, we're replacing existing functionality and must maintain the same urls, therefore, we are unable to hide this new functionality behind express routes or feature flags. I've grouped each of the commits logically so it should be fairly easy to follow, there is very little complexity, you can play with both user journeys via the deployment.**

There are only a few basic functional tests, I'm am about to build the rest out.

## Checklist


src/client/components/Form/elements/FieldDate/index.jsx
```diff
@@ -60,9 +60,13 @@ const getValidator =
...
-    return !isValid && (!isDateEmpty || required)
-      ? required || 'Enter a valid date'
-      : null
+    if (required && isDateEmpty) {
+      return required
+    }
+
+    if (!isValid && !isDateEmpty) {
+      return invalid || 'Enter a valid date'
+    }
```
paulgain, line 69: Yes, this could be a ternary but for readability this is a better option.

src/apps/investments/client/projects/create/constants.js
```diff
@@ -0,0 +1,6 @@
+export const OPTION_YES = 'yes'
```
vfarah-if, line 1: (Minor) We have these constants defined and littered all over the code base - maybe we should import from one space and utilise one set representing this. You can pick from src/common/constants.js or src/apps/constants.js

src/apps/investments/client/projects/create/reducer.js
```diff
@@ -0,0 +1,33 @@
...
+}
+
+export default (state = initialState, { type, result, searchTerm }) => {
+  switch (type) {
+    case INVESTMENT__COMPANY_INVESTMENT_COUNT:
+      return {
+        ...state,
+        companyInvestmentCount: result.count,
+      }
+    case INVESTMENT__SEARCH_COMPANY_LIST_LOADED:
```
vfarah-if, line 20: Reducers and connect are being removed for the TaskForm patterns and ways of abstracting this

src/apps/investments/client/projects/create/tasks.js
```diff
@@ -0,0 +1,137 @@
+import axios from 'axios'
+
```
vfarah-if, line 2: **apiProxyAxios** within client/components/Task/utils for utilising a shared error handling, however I may not be understanding the underlying reason for creating bespoke errorhandling (see storybook for examples Peter created)

src/apps/investments/client/projects/tasks.js
```diff
@@ -1,12 +1,11 @@
 import axios from 'axios'
```
vfarah-if, line 1: See comments above about axios

craigjbass (approved):

vfarah-if (approved):

src/apps/investments/client/projects/create/reducer.js
paulgain: Agreed, however, for `INVESTMENT__COMPANY_INVESTMENT_COUNT` I needed to style my own loading indicator so it sits central within a row of the summary table. At present, this is the only way to achieve this.
Regarding `INVESTMENT__SEARCH_COMPANY_LIST_LOADED` this is a POST request, the new patterns you mention only deal with GET requests. Going forward, it would be good to makes these changes within the new patterns but is out of scope for this PR.

src/apps/investments/client/projects/create/constants.js
paulgain: Agreed, it needs tidying up. We had the same problem with dates which you can read about [here](https://github.com/uktrade/data-hub-frontend/pull/3797#issuecomment-886659492.).
Is it OK if we do this in a separate PR? For now I'm just going to reference `src/apps/constants.js`

PR #4347, opened 2022-02-14
Restore FieldAddAnother component fixing all accessibility issues

vfarah-if:
## Description of change

Restore the FieldAddAnother component showing this working with multiple fields as a repeater for multiple components.

## Checklist


cgsunkel (approved):

src/client/components/Form/elements/FieldAddAnother/index.jsx
```diff
@@ -0,0 +1,122 @@
...
+`
+const FieldAddAnother = ({
+  name,
+  label,
+  children,
+  'data-test-prefix': dataTestPrefix,
+  'item-name': ariaItemName,
+  initialChildCount = 0,
+}) => {
+  const [internalValue, setInternalValue] = useState(
```
tomkingston, line 28: I don't particularly mind this myself, but I thought that architeturally we avoid internal state and use redux for everything?

tomkingston (commented):
This actually looks good to me and it makes me think how much easier /nicer everything seems when we use internal state.... however, I think this probably does go against the redux architecture that Peter wanted for datahub, so would be useful to know what your thinking is on this before approving?

src/client/components/Form/elements/FieldAddAnother/index.jsx
vfarah-if: Ah its a very good point and I kept this the way it was in the original, but will fix this as a code exercise

src/client/components/Typeahead/Typeahead.jsx
```diff
@@ -260,7 +260,7 @@ const Typeahead = ({
       <InputWrapper>
         <AutocompleteInput
           {...inputProps}
-          id={name}
+          id={inputProps.inputId || name}
```
vfarah-if, line 263: Remove this

src/client/components/Form/elements/FieldAddAnother/FieldAddAnother.jsx
```diff
@@ -0,0 +1,148 @@
...
+  padding-bottom: ${SPACING.SCALE_1};
+`
+const FieldAddAnother = ({
+  name,
+  legend,
+  children,
+  'data-test-prefix': dataTestPrefix,
+  'item-name': ariaItemName,
+  initialChildGroupCount = 1,
+  fieldGroupIds,
```
mforner13, line 42: I might be missing something stupid - but how does the component know what `fieldGroupIds` is if the component isn't connected to that slice of state?

src/client/components/Form/elements/FieldAddAnother/FieldAddAnother.jsx
```diff
@@ -0,0 +1,148 @@
...
+}
+
+FieldAddAnother.propTypes = {
+  'data-test-prefix': PropTypes.string,
+  'item-name': PropTypes.string.isRequired,
+  name: PropTypes.string.isRequired,
+  legend: PropTypes.string.isRequired,
+  children: PropTypes.func,
+  initialChildGroupCount: PropTypes.number,
+  // multiInstance internal properties
```
mforner13, line 112: Calling these the 'props from redux state' or something similar could make it easier to understand :) (Even though it is multiinstance that's handling them

src/client/components/Form/elements/FieldAddAnother/FieldAddAnother.jsx
```diff
@@ -0,0 +1,148 @@
...
+
+const StyledLink = styled('div')`
+  padding-bottom: ${SPACING.SCALE_1};
+`
+const FieldAddAnother = ({
+  name,
+  legend,
+  children,
+  'data-test-prefix': dataTestPrefix,
+  'item-name': ariaItemName,
```
mforner13, line 40: Can these just be set in camelcase? Or are the kebab case versions being applied directly anywhere? :)

src/client/components/Form/elements/FieldAddAnother/usage.md
```diff
@@ -0,0 +1,82 @@
+# FieldAddAnother
+
+### Description
+
+This is a form component renders a list in a form. The fields of this list are specified by a function based as a child to the component that is called with the arguments `{groupIndex}`, where:
```
mforner13, line 5: Think this needs to be reworded to make it a bit clearer. Maybe something like: 

```
This is a form field component which enables rendering of multiple instances of the same field(s).
The component takes as its children a function which returns the field(s) to be repeated. The argument to this function is the `{groupIndex}`, where:
```

src/client/components/Form/elements/FieldAddAnother/usage.md
```diff
@@ -0,0 +1,82 @@
...
+
+- ***groupIndex*** is the the index of FieldAddAnother component item
+
+The ***initialValues*** is an array of grouped items. The ***transformArrayToObject*** function that will flatten this for the ***form*** component to deliver the values to the relative Field type components. 
+
+The **initialChildGroupCount** is an attribute informing the component of how many items need to be synchronised within the *FieldAddAnother* children and is primarilly used in edit mode when showing values within the *FieldAddAnother* component. So in the example below there are two groups of adviser roles [adviser_0, role_0] and (adviser_1, role_1) and so the *initialChildGroupCount* should be ***2***. By default this is ***1*** and when no data is being assigned will show no component details until the Add another button is clicked.
```
mforner13, line 11: 'how many instances of the child components need to be rendered...' ? :)

mforner13 (approved):
👍

PR #4400, opened 2022-03-03
Add an `npm` command to run selected unit tests

Harriethw:
## Description of change

I would like to be able to run selected unit tests rather than the whole
suite, so that we can do TDD more effectively.

I was going to edit the existing `test:unit` command but as this is used
by a couple of other commands where variables are passed in, it seems
like we need a new command.

## feedback

Maybe it's possible to do this already with the existing command? 
I'm not a master at shell commands so mainly copied this from [SO](https://stackoverflow.com/questions/51401352/if-else-on-arguments-in-npm-run-script) -
maybe there's a nicer way of doing it? 
Is there a nicer name for the command?

## Test instructions

run `npm run test:unit:files ./src/apps/events/attendees/controllers/__test__/create.test.js ./src/apps/omis/controllers/__test__/create.test.js` and you should see tests running for just those files

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
     "lint:js": "npx eslint --ext .jsx,.js ./test ./src ./assets",
     "lint:js:fix": "npx eslint --ext .jsx,.js ./test ./src ./assets --fix",
     "test": "run-s test:*",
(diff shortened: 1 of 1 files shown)


craigjbass (approved):
Nice!!

vfarah-if (approved):

PR #4440, opened 2022-03-14
Add a robots noindex meta tag.

craigjbass:
## Description of change

Adds a robots noindex meta tag to every page.


cgsunkel (approved):

mforner13 (changes requested):
Sorry just realised. For this to show on every page - there are multiple template files with varying levels of nunjucks in you'll need to add it in too. Will just find them

mforner13 (approved):
I was wrong - they all extend from this one 😄

PR #4445, opened 2022-03-15
Log output and errors from `node.js` server

Harriethw:
## Description of change

We want to be able to see console logs and errors from the `node.js`
server, which is started when the webpack build is done.

This commit sets the `stdout` and `stderr` of the child
process[1] spawned in `StartServerAfterBuild` to inherit from the
parent.

[1] https://nodejs.org/api/child_process.html#optionsstdio

## Test instructions

After running `npm run develop`:
1. Console log something in a controller function and you should see it in your terminal. 
2. Cause a compilation error to happen in a controller and you should see that in the terminal.

## Checklist

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -20,7 +20,7 @@ const StartServerAfterBuild = () => {
         } else {
           server = spawn(
             "npx nodemon --inspect --ignore 'src/**/__test__/**/*'",
-            { stdio: 'pipe', shell: true }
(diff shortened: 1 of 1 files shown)


paulgain:
Consider using `stdio: ['pipe', 'inherit', 'inherit']` for an elegant fix.

Harriethw:
@paulgain nice one, thanks for the tip! I've updated the commit now.

cgsunkel (approved):

PR #4476, opened 2022-03-25
Adding Activity Stream activities to a Contact

Harriethw:
## Description of change

We want to show all activities from Activity Stream for a Contact under the Contact Interactions tab, so that we can surface Aventri data for users and advisers can have a full picture of the contact's activity.

This story began with wanting to just pull in Aventri data, but we found with a query to Activity Stream, and by re-using much of the Activity Feed architecture, we could pull in other external and Data Hub data sources.

NOTE: We'll be making iterative releases and this is the very first chunk, which we're hiding behind a user feature flag. It's not intended for production users, but will allow us to do some testing and see how it looks with prod data.

## Test instructions

There is currently only Aventri data for one contact, John Doe, and only on the dev environment.
If you are running locally, you'll need to:

1. point at dev environment
2. add the `user-contact-activities` user feature flag to your user (you can do this in the Django admin -> Adviser > Other)
3. Visit John Doe: http://localhost:3000/contacts/68be55fc-544d-4d9c-a264-977b917f1fd3/interactions and see the data coming through 

### Before

Screenshot 2022-03-25 at 16 02 50

### After
(with feature flag active on user, should look the same to users without flag)

Screenshot 2022-03-25 at 16 03 11

## Checklist


src/apps/companies/apps/activity-feed/es-queries/contact-activity-query.js
```diff
@@ -0,0 +1,66 @@
...
+            },
+          },
+        ],
+      },
+    },
+    sort: [
+      {
+        published: {
+          order: 'desc',
+          unmapped_type: 'long',
```
Harriethw, line 57: I just noticed that I had copied this `unmapped_type` field in from another query without knowing what it does.. turns [out it won't sort if the field is `null`](https://stackoverflow.com/questions/71332151/if-all-fields-for-a-column-are-null-opensearch-does-not-include-that-field-and). I think that makes sense, because if it doesn't have a date for published we probably don't want to retrieve it?

src/client/components/ActivityFeed/activities/AventriAttendee.jsx
```diff
@@ -0,0 +1,43 @@
...
+    registrationStatus: activity.object['dit:aventri:registrationstatus'],
+  }
+}
+
+export default class AventriAttendee extends React.PureComponent {
+  static propTypes = {
+    activity: PropTypes.object.isRequired,
+  }
+
+  static canRender(activity) {
```
Harriethw, line 21: note: we had to make this component a class so that we could have this static `canRender` function called by `Activity.jsx`

src/apps/contacts/client/ContactInteractionsApp.jsx
```diff
@@ -0,0 +1,44 @@
...
+import styled from 'styled-components'
+import { TASK_GET_CONTACT_INTERACTIONS, ID, state2props } from './state'
+import { CONTACTS__INTERACTIONS_LOADED } from '../../../client/actions'
+import Task from '../../../client/components/Task'
+import Activity from '../../../client/components/ActivityFeed/Activity'
+import { SPACING } from '@govuk-react/constants'
+
+const ContactInteractionsList = styled('ol')`
+  list-style-type: none;
+  padding: 0;
```
Harriethw, line 12: note: the design for this is also WIP

test/sandbox/routes/v4/activity-feed/activity-feed.js
```diff
@@ -35,6 +35,15 @@ const DATA_HUB_AND_EXTERNAL_ACTIVITY = [
 const VENUS_LTD = 'dit:DataHubCompany:0f5216e0-849f-11e6-ae22-56b6b6499611'
 
 exports.activityFeed = function (req, res) {
+  // Activities by contact
+  var isContactActivityQuery = get(
+    req.body,
+    "query.bool.must[0].bool.should[1].bool.must[1].term['object.dit:emailAddress']"
```
Harriethw, line 41: note: this seems like a really brittle way to add logic to a stubbed api 😓 but we're following the existing patterns and weren't sure how else to do it.

src/apps/companies/apps/activity-feed/es-queries/contact-activity-query.js
chloehkwong1: I agree based on my understanding of reading the [docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/sort-search-results.html)! I think if we don't have it the search request will fail if there is no date? Which we probably don't want?

test/functional/cypress/specs/contacts/interactions-spec.js
```diff
@@ -0,0 +1,37 @@
...
+
+  context(
+    'when viewing a contact with data on Activity Stream with the feature flag enabled',
+    () => {
+      before(() => {
+        cy.setUserFeatures(['user-contact-activities'])
+        cy.visit(urls.contacts.contactInteractions(contactId))
+      })
+
+      it('should display the Activity Stream activities', () => {
```
Harriethw, line 15: note: we're not testing all the branching logic of whether it shows different types of activities here and the different content within the cards: 
1. that logic for every activity that isn't Aventri should in theory be covered by the Activity Feed tests
2. We're not sure what the Aventri stuff is going to look like yet, for now we just want to test if it renders or not, I think?

src/client/components/ActivityFeed/activities/AventriAttendee.jsx
craigjbass: I think you can add functions to functional components 

`const B = () => <span />`

`B.canRender = (activity) => true`

Harriethw: that is cool!

test/sandbox/routes/v4/activity-feed/activity-feed.js
craigjbass: I think the smell is that our contract here is an open search query. Not something we can solve here I think

src/client/components/ActivityFeed/activities/AventriAttendee.jsx
Harriethw: changed to a function 😎

craigjbass (approved):

vfarah-if (approved):

PR #4491, opened 2022-03-29
DET-85 refactor Contact Interaction task to handle errors properly

Harriethw:
## Description of change

The `Task` component [does not handle generic Errors](https://github.com/uktrade/data-hub-frontend/blob/99d66f94319bb6cfcae5f33a027d8c645ce0cea2/src/client/components/Task/saga.js#L42) as we originally assumed. It looks like we have to handle them by rejecting the Task's Promise and returning a message string.

It doesn't look particularly user-friendly atm, so I expect in future we'll want to render a nice-looking error component

## Test instructions

_What should I see?_

### Before

The Task would fail in the background and the 'loading' component shown continuously.

### After

If `fetchActivityFeed` throws an error:
Screenshot 2022-03-29 at 11 42 45

## Checklist

### DIFF
--- a/src/apps/companies/apps/activity-feed/controllers.js
+++ b/src/apps/companies/apps/activity-feed/controllers.js
@@ -160,7 +160,7 @@ async function fetchActivitiesForContact(req, res, next) {
       next(error)
     })
 
-    let activities = results?.hits.hits.map((hit) => hit._source)
+    let activities = results.hits.hits.map((hit) => hit._source)
 
     res.json({ activities })
   } catch (error) {
--- a/src/apps/companies/apps/activity-feed/controllers.js
+++ b/src/apps/companies/apps/activity-feed/controllers.js
@@ -160,9 +160,7 @@ async function fetchActivitiesForContact(req, res, next) {
       next(error)
     })
 
-    let activities = results
-      ? results.hits.hits.map((hit) => hit._source)
-      : { error: 'error' }
+    let activities = results?.hits.hits.map((hit) => hit._source)
 
     res.json({ activities })
   } catch (error) {
--- a/src/apps/contacts/client/tasks.js
+++ b/src/apps/contacts/client/tasks.js
@@ -2,4 +2,7 @@ import axios from 'axios'
 import urls from '../../../lib/urls'
 
 export const getContactInteractions = (contactId) =>
-  axios.get(urls.contacts.activity.data(contactId)).then(({ data }) => data)
+  axios
+    .get(urls.contacts.activity.data(contactId))
+    .then(({ data }) => data)
+    .catch(() => Promise.reject('Unable to load Contact Interactions'))
--- a/src/apps/contacts/client/ContactInteractionsApp.jsx
+++ b/src/apps/contacts/client/ContactInteractionsApp.jsx
@@ -30,15 +30,11 @@ const ContactInteractionsApp = ({ contactId, activities }) => (
     {() =>
       activities && (
         <ContactInteractionsList>
-          {activities?.error ? (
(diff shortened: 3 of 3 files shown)


src/apps/companies/apps/activity-feed/controllers.js
```diff
@@ -160,9 +160,7 @@ async function fetchActivitiesForContact(req, res, next) {
       next(error)
     })
 
-    let activities = results
-      ? results.hits.hits.map((hit) => hit._source)
-      : { error: 'error' }
+    let activities = results?.hits.hits.map((hit) => hit._source)
```
craigjbass, line 163: Do you want activities to always be a list and never be null/undefined? Think with the ?. there is a chance for this to resolve to a null-like value

Harriethw: hmm yeah it's a bit odd. Ideally we wouldn't need this `?` at all but I'm not entirely confident we're definitely catching the errors.

I think that the way that the `Task` works if you want to avoid a 'flash' of data before the call to the API starts loading, [you have to have a check within it](https://github.com/uktrade/data-hub-frontend/blob/feature/DET-85-refactor-error-handling/src/apps/contacts/client/ContactInteractionsApp.jsx#L31 ). So I guess there's three states - it's null/undefined in which case we render nothing as the task in theory hasn't started, it's an empty array in which case there are no activities, or it's an array of something.  

But we could check that the array is empty and just live with having a potential render of '0 activities' before it starts loading.

craigjbass: I think this file is server-side JS, so the result of this function will only ever be used after the task has started. 

I think it's weird that the Task works like that, I'd expect the public interface of Task to have require placeholder JSX for the known states to avoid the conditional on the client side.

Harriethw: > I think it's weird that the Task works like that, I'd expect the public interface of Task to have require placeholder JSX for the known states to avoid the conditional on the client side. 

I haven't actually dug into it, so might take a look

> I think this file is server-side JS, so the result of this function will only ever be used after the task has started

I'm thinking that we should remove the `?` and rely on the try block to do its job. It doesn't really make sense to return `null` or an empty array if `results` don't exist, as the API call will have failed and we should either catch that or blow up.

chloehkwong1 (approved):

src/apps/contacts/client/tasks.js
```diff
@@ -2,4 +2,7 @@ import axios from 'axios'
 import urls from '../../../lib/urls'
 
 export const getContactInteractions = (contactId) =>
-  axios.get(urls.contacts.activity.data(contactId)).then(({ data }) => data)
+  axios
+    .get(urls.contacts.activity.data(contactId))
+    .then(({ data }) => data)
+    .catch(() => Promise.reject('Unable to load Contact Interactions'))
```
cgsunkel, line 8: I think we should stick with the `Error occurred while loading activities.` text here until we know what the final name of this page will be

cgsunkel (approved):

dredmonds (approved):

src/apps/contacts/client/tasks.js
```diff
@@ -2,4 +2,7 @@ import axios from 'axios'
 import urls from '../../../lib/urls'
 
 export const getContactInteractions = (contactId) =>
-  axios.get(urls.contacts.activity.data(contactId)).then(({ data }) => data)
+  axios
```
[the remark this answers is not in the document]
cgsunkel: We can't use `api-proxy` with these calls

onyela-o (approved):

PR #4504, opened 2022-04-04
Add pagination to Contact Interactions page

chloehkwong1:
## Description of change

## Test instructions

_What should I see?_

### Before

_Add a screenshot_

### After

_Add a screenshot_

## Checklist


dredmonds (approved):

dredmonds (dismissed):
Dismissed the approval review which undermine the internal process of PR.

src/apps/contacts/client/ContactInteractionsApp.jsx
```diff
@@ -1,11 +1,23 @@
 import React from 'react'
```
tomkingston, line 1: It looks like this was done on a previous PR, however, I haven't seen these file nams appended with `App` before so `ContactInteractions.jsx` would fit the naming convention a bit better. Also, I believe the plan was to move to the new client modules folder - so this would ideally be [here](https://github.com/uktrade/data-hub-frontend/tree/master/src/client/modules/Interactions) along with the reducer and tasks. 

Same with ActivityFeedApp.

src/apps/contacts/client/ContactInteractionsApp.jsx
```diff
@@ -17,28 +29,68 @@ const ContactInteractionsList = styled('ol')`
...
-        </ContactInteractionsList>
-      )
-    }
-  </Task.Status>
-)
+const StyledSpan = styled('span')`
+  color: ${GREY_1};
+`
+
+const ContactInteractionsApp = ({
```
tomkingston, line 36: Again, this should probably just be called `ContactInteractions` to fit the naming convention.

tomkingston (approved):
Looks good - just the comments about the naming and location of some of the files (which was actually done in a previous PR)

src/apps/contacts/client/ContactInteractionsApp.jsx
tomkingston: I think part of the plan to move to the modules folder was also to add these to the router, so maybe it's best to leave in this location for now (but drop the `App` bit)

cgsunkel: We're going to rename these files to `ContactActivities` in a future PR (which I'm going to start after merging this), so will include these changes with that work.

Harriethw: yep we originally added the `App` because we were following the `ActivityFeedApp` pattern, but agree it's not right, and we weren't really sure where it belongs... there's a high chance that the two could become the same shared component as we continue building.

>Also, I believe the plan was to move to the new client modules folder - so this would ideally be [here](https://github.com/uktrade/data-hub-frontend/tree/master/src/client/modules/Interactions) along with the reducer and tasks. 

Would def be interested in understanding where things should be and what the plans are if there's any documentation on it. The 'Interactions' tab for a Contact is going to be re-named 'Activities' because it's not just `Interactions` anymore, so not sure it makes sense to live in that module? Would it make more sense in the `src/client/components` folder where `ActivityFeedApp` currently lives?

tomkingston: I'm not sure if there are any docs around on this - before he left, Ian began work to put everything on to react routes (instead of using a hybrid of express and react with the mount component stuff) - unfortunatley that requires quite a bit of work in some places where the express controller is doing a lot so at the moment we have a bit of both going on.
Yeah, it might make sense for that activities stuff to go into the components, but if it is the whole page, somewhere like `modules/contacts/activities/index.jsx` might be a good place - I actually thought calling it "pages" made more sense than modules!

Harriethw: thanks @tomkingston that makes sense! As Chris said, we having a ticket to rename the component so I think moving it would make sense in a future PR.

src/client/components/Form/index.jsx
```diff
@@ -121,6 +121,7 @@ const _Form = ({
           ])
         )
       )
+      window.scrollTo({ top: 0 })
```
mforner13, line 124: Have been thinking we should do this for ages! Hooray 🎉

test/functional/cypress/specs/contacts/interactions-spec.js
```diff
@@ -19,15 +40,33 @@ describe('Contact interactions', () => {
...
+
+      it('should display the total number of activites', () => {
+        cy.get('#contact-interactions-app').contains('1,233 activities')
+      })
+
+      context('when there are more than 10 activities', () => {
+        it('should be possible to page through', () => {
+          cy.get('[data-page-number="2"]').click()
+          cy.get('#contact-interactions-app').contains(
+            `${dataHubActivities.hits.hits[0]._source.object['dit:subject']}`
```
mforner13, line 52: I may have misunderstood, but is this checking the logic correctly if this is the first object in the activities fixture? 

Would finding the text that says 'Page 2 of x' be more reliable, as then we're not testing sandbox logic?

mforner13 (commented):
Just one suggestion for a test :)

test/functional/cypress/specs/contacts/interactions-spec.js
Harriethw: agree it is a bit convoluted but the thinking was that we wanted to make sure everything is hooked up from React land to the controller function - so that the page number gets converted to the [`from` value in controller ](https://github.com/uktrade/data-hub-frontend/blob/feat/DET-99-paginate-contact-activities/src/apps/companies/apps/activity-feed/controllers.js#L154) and that gets passed through to the api. If we just tested that on the page number I think that might not provide coverage for the controller function?

I would probably usually aim for that granularity in a unit test but I've been told we shouldn't be writing new unit tests?

mforner13: Have spoken and makes sense, as a different fixture is returned based on the `from` value being properly calculated 😄

mforner13 (approved):
Looks good 👍

test/functional/cypress/specs/contacts/interactions-spec.js
cgsunkel: I believe the intention around this test was to cover the redux logic used for pagination and that activities were being paginated/returned correctly.

I will add an assertion for the 'Page 2' text

Harriethw: @cgsunkel please could you also add a comment above line 52 as we thought that might be a good idea to make it crystal clear. Something like:
`//checks that a second page of data is being returned from sandbox`

PR #4508, opened 2022-04-04
Remove progress indication from filter typeaheads and fix teams filter

elcct:
## Description of change

This fixes a visual bug where filter typeaheads in collection pages would briefly disappear showing "Loading advisers" or "Loading teams" progress indicator after selecting a filter. Since the progress is also shown in the main collection list, it is not necessary to show it in individual filters. An option to bypass showing progress has been added to `Task.Status` component.

Additionally it fixes Teams collection filter, where `getTeamNames` used axios incorrectly causing component to crash.

## Test instructions

Go to Investment collection page and select an Adviser. The adviser typeahead shouldn't be disappearing.

Go to Interaction collection page and select a Team. The team typeahead shouldn't be disappearing and component no longer crashes after making a selection.

## Checklist

### DIFF
--- a/src/apps/investments/client/projects/ProjectsCollection.jsx
+++ b/src/apps/investments/client/projects/ProjectsCollection.jsx
@@ -1,6 +1,5 @@
 import React from 'react'
 import { connect } from 'react-redux'
-
 import urls from '../../../../lib/urls'
 import {
   CollectionFilters,
@@ -49,7 +48,6 @@ const ProjectsCollection = ({
   const adviserListTask = {
     name: TASK_GET_INVESTMENTS_PROJECTS_ADVISER_NAME,
     id: INVESTMENT_PROJECTS_ID,
-    progressMessage: 'Loading advisers',
     startOnRender: {
       payload: payload.adviser,
       onSuccessDispatch: INVESTMENTS__PROJECTS_SELECTED_ADVISERS,
--- a/src/client/components/RoutedAdvisersTypeahead/index.jsx
+++ b/src/client/components/RoutedAdvisersTypeahead/index.jsx
@@ -32,7 +32,7 @@ const RoutedAdvisersTypeahead = ({
   loadOptions = fetchAdvisers(onlyShowActiveAdvisers),
   ...props
 }) => (
-  <Task.Status {...taskProps}>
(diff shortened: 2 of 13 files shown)


tomkingston (commented):
Looks good, especially for the teams fix, but I'm not sure this fixes the jumping issue that @paulgain has identified when I try it in heroku

paulgain:
> Looks good, especially for the teams fix, but I'm not sure this fixes the jumping issue that @paulgain has identified when I try it in heroku

@elcct this is a huge improvement, it's great to see the adviser label and previously selected advisors remaining in situ after selecting other advisers, previously both label and advisor tags would vanish and then reappear during an API call creating an unwanted flashing effect. However, I see two issues here:

1. After selecting an advisor such as `Paul Gain, Digital Data Hub - Live Service` the entire name is displayed in a tag but then sooner after it snaps back to `Paul Gain` creating an unwanted effect.
2. Removing the loading indicator is not the correct approach, after all we're making an API call to the `/adviser/<adviser-id>` endpoint, we would also need to handle any error should the call fail (we may be doing that already). I think we need a different type of loading indicator that doesn't consume vertical space, ideally it would be a thin horizontal line that loads left to right - we can do this piece of work in a separate PR.

elcct:
@paulgain 

1. It seems like the chips are built from typeahead option label that includes teams and then after an API call they are built just with a name. I thought of a simplest workaround to include an extra attribute `chipLabel` that would be used when available to build chips and this way they can be made consistent in this scenario - when typeahead options are created, the `label` for the drop down contains name and a team and `chipLabel` only contains a name.
2. I agree. This seemed a nice idea at first, but given that a requests are being sent after user selection, there should be an indication of that. The proposed loading indicator would be very nice, but I think that would require much more work - creating the component itself and some refactoring of `Task.Status` like passing the progress status to children so the progress could be displayed underneath the filter header - indeed something for a separate PR. I thought of an alternative solution to simply use existing `LoadingBox` component that overlays the typeahead when async requests are being made. This way it is still not jumping, but also there is a visual feedback about the request.

paulgain (approved):

tomkingston (approved):

PR #4513, opened 2022-04-05
Rename contact interactions tab to activities

cgsunkel:
## Description of change

Now that the contact interactions feed contains more than just interactions, it has been decided to rename the page to 'Contact Activities'. These changes are behind the `user-contact-activities` flag for the moment.

## Test instructions

Go to a contact with the feature flag enabled. Everything should be as it was before but all mentions of 'interactions' apart from the local nav have been changed.

## Checklist


src/apps/contacts/router.js
```diff
@@ -49,11 +49,7 @@ router.get('/:contactId/audit', getAudit)
 
 router.get('/:contactId/documents', renderDocuments)
 
-router.use(
-  '/:contactId/interactions',
-  setInteractionsDetails,
-  interactionsRouter
-)
+router.use('/:contactId/activities', setInteractionsDetails, interactionsRouter)
```
Harriethw, line 52: just an initial thought without running the branch - the activities stuff needs to still be behind a feature flag, but I guess the router/url changes will be visible to anyone? Maybe we will have to do anything related to the route in a future PR

src/client/modules/Contacts/ContactActivity/ContactActivity.jsx
```diff
@@ -45,13 +51,20 @@ const ContactInteractionsApp = ({
   return (
     <GridRow>
       <GridCol setWidth="full">
+        <StyledSectionHeader>Contact activity</StyledSectionHeader>
+
+        <p>
+          Activity could include a meeting, call, email, event or other
```
chloehkwong1, line 57: I'm being very very pedantic but maybe it should be "_An_ activity ..." rather than just "Activity"?

cgsunkel: I think this is what was written in the design, will change it

Harriethw:
Thanks @cgsunkel 💫   - it looks good apart from the router changes, as we don't want users without the feature flag to see any different. I think the easiest thing would be to keep the route and LOCAL_NAV the same, [but we could conditionally swap the breadcrumb value here](https://github.com/uktrade/data-hub-frontend/blob/47cd098df71799071d09b289498c7eea047cd42c/src/apps/interactions/controllers/list.js#L119 ). Something like this?

```
const isContactActivitiesFeatureOn = res.locals.userFeatures?.includes(
      'user-contact-activities'
    )

const breadcrumbTitle = isContactActivitiesFeatureOn
      ? 'Activity'
      : 'Interactions'

res.breadcrumb(breadcrumbTitle).render(view, {
      actionButtons,
      ....
```

lmk if you want to pair on anything!

src/apps/interactions/controllers/list.js
```diff
@@ -116,7 +124,7 @@ function renderInteractionsForEntity(req, res, next) {
         ]
       : undefined
 
-    res.breadcrumb('Activity').render(view, {
+    res.breadcrumb(breadcrumbTitle).render(view, {
```
Harriethw, line 127: minor: I think now we have that `isContactActivitiesFeatureOn` variable we can pass that into the breadcrumbs on line 132 rather than checking the locals again:

```
    res.breadcrumb(breadcrumbTitle).render(view, {
      actionButtons,
      props: {
        contactId: contactId,
      },
      isContactActivitiesFeatureOn,
    })
   ```

Harriethw (approved):
LGTM 👍  once the fixup commits are fixedup I think it's ready, made a minor comment but don't think it should block merging if you don't fancy doing it.

chloehkwong1 (approved):

PR #4514, opened 2022-04-06
Persist my investment projects filters with sessionStorage

tomkingston:
## Description of change

Persists the filters and sort on the personalised homepage so that users do not have to set them again when they navigate elsewhere and come back.

## Test instructions

Go to personalised homepage and select some filters / sort - navigate to other pages and come back and the filters should stay applied.

## Checklist


mforner13 (commented):
Looks good! Some functional tests around it would be good if possible :)

src/client/components/MyInvestmentProjects/index.jsx
```diff
@@ -83,35 +83,58 @@ const MyInvestmentProjects = ({
...
         label="Stage"
         options={STAGE_OPTIONS}
-        initialValue={stage}
-        onChange={(event) => onStageChange(event.target.value)}
+        initialValue={window.sessionStorage.getItem('myProjects.stageFilter')}
+        onChange={(event) => {
+          window.sessionStorage.setItem(
+            'myProjects.stageFilter',
+            event.target.value
+          )
```
paulgain, line 91: Wasn't there talk of moving this to a Saga? I think this is what's classified as `impure`. If not, it may be a good idea to abstract these out into a couple of functions so we can reuse the functionality, it will also be easier to test. I've used `loadItem` but maybe it should be called `loadState`:

```javascript
export const loadItem = (item) => {
  try {
    const strItem = window.sessionStorage.getItem(item)
    return JSON.parse(strItem)
  } catch (err) {
    // Catch error
  }
}

export const saveItem = (key, item) => {
  try {
    const strItem = JSON.stringify(item)
    window.sessionStorage.setItem(key, strItem)
  } catch (err) {
    // Catch error
  }
}
```

src/client/components/MyInvestmentProjects/index.jsx
```diff
@@ -178,39 +186,73 @@ MyInvestmentProjects.propTypes = {
...
-  onStageChange: (stage) =>
+  onStageChange: (stage) => {
     dispatch({
       type: MY_INVESTMENTS__STAGE_CHANGE,
       stage,
       page: 1,
-    }),
-  onStatusChange: (status) =>
+    })
+    dispatch({
```
tomkingston, line 206: We need two different dispatch calls because the second uses the new saga

paulgain (approved):
Nice work Tom 👍

PR #4546, opened 2022-04-27
Stop interactions with no service breaking the company activity feed

cgsunkel:
## Description of change

There is currently a live issue where the company activity feed breaks if an interaction without a service comes through [(example in dev](https://www.datahub.dev.uktrade.io/companies/b570f2de-1638-4308-9149-5ff17b64d18e/activity)). This is because some of the transformers for the new interactions card (which run regardless of whether the feature flag is active or not) are reliant on a service being present.

The relevant functions now only run if a service exists, and return `null` if it doesn't. In addition to this, service information no longer renders on the new contact activity feed where interactions have no service.

## Test instructions

With your local FE pointing to dev, go to [the example company listed above](http://localhost:3000/companies/b570f2de-1638-4308-9149-5ff17b64d18e/activity). The page should render properly.

If you want to inspect the contact activity feed changes, enable the `user-contact-activities` feature flag and go to [this contact](http://localhost:3000/contacts/715e8a58-6515-4d52-a4b5-84cb49566b01/interactions?featureTesting=user-contact-activities). You should see an interaction with no service information.

## Checklist

### DIFF
--- a/test/functional/cypress/specs/contacts/activity-spec.js
+++ b/test/functional/cypress/specs/contacts/activity-spec.js
@@ -85,12 +85,6 @@ describe('Contact activity', () => {
           )
         })
 
-        it('should display the communication channel', () => {
-          cy.get('[data-test=interaction-activity]').contains(
-            'Communication channel:'
-          )
-        })
-
         it('should display the advisers with email', () => {
           cy.get('[data-test=interaction-activity]')
             .contains(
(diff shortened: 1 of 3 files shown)


test/functional/cypress/specs/contacts/activity-spec.js
```diff
@@ -85,12 +85,6 @@ describe('Contact activity', () => {
           )
         })
 
-        it('should display the communication channel', () => {
```
Harriethw, line 88: I think we need to have a test for cases where there is a communication channel and where there is not, and the same for `service`.

cgsunkel: I only took this approach because this is an urgent issue and needs to be merged/released quickly. This could be something that goes into the other PR?

Harriethw: @cgsunkel ok, that makes sense

Harriethw (approved):

PR #4561, opened 2022-05-05
Change land date filters to month-date format.

elcct:
## Description of change

This changes land date filters on investment collection page to be in month year format.

Note: this may not work in Firefox, but we are okay with that.

## Test instructions

Visit investment collection page and unroll Land date details section to see updated filters.

### Before

land dates before

land date before labels

### After

land dates

land dates after labels

## Checklist


src/apps/investments/client/projects/tasks.js
```diff
@@ -1,21 +1,25 @@
...
+  transformInvestmentProjectToListItem,
+  transformLandDateFilters,
+} from './transformers'
 
 const handleError = (error) => Promise.reject(Error(error.response.data.detail))
 
 export const getProjects = ({ limit = 10, page, companyId, ...rest }) => {
   let offset = limit * (parseInt(page, 10) - 1) || 0
 
+  const transformedRest = transformLandDateFilters(rest)
```
Harriethw, line 14: minor/opinion: could this be re-named to `transformedParams` to be a bit more specific?

src/client/filters.js
```diff
@@ -1,6 +1,7 @@
-const { formatLongDate } = require('./utils/date')
+const { formatLongDate, formatShortDate } = require('./utils/date')
 
 const getDateLabel = (value) => (value ? `${formatLongDate(value)}` : '')
+const getShortDateLabel = (value) => (value ? `${formatShortDate(value)}` : '')
```
Harriethw, line 4: minor: I *think* as the format function returns a string we don't need to add string interpolation here?

Harriethw (approved):
LGTM :)

src/apps/investments/client/projects/transformers.js
```diff
@@ -4,10 +4,38 @@ const {
...
+
+const { addressToString } = require('../../../../client/utils/addresses')
+
+const {
+  DATE_SHORT_FORMAT,
+  DATE_LONG_FORMAT_3,
+} = require('../../../../common/constants')
+
+const formatMonthYearDate = (date) =>
+  formatFns(parse(date, DATE_SHORT_FORMAT, new Date()), DATE_LONG_FORMAT_3)
```
paulgain, line 19: It would be good if we could move this into `src/client/utils/date.js` so we have a central place for dates and encourage reusability.

paulgain (approved):

src/apps/investments/client/projects/tasks.js
elcct: I prefer to keep it that way, as it tells me the new variable comes from `rest`, so the code is easier to follow.

src/client/filters.js
elcct: Good point! Fixed. Thank you

src/apps/investments/client/projects/transformers.js
elcct: I have moved it now. Thank you!

PR #4577, opened 2022-05-18
Set NPM Engine to the latest NPM

vfarah-if:
## Description of change

Update to use the fix that was not resolving the overrides

## Test instructions

Make sure the front

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,7 +61,8 @@
   },
   "license": "MIT",
   "engines": {
-    "node": "16.14.0"
+    "node": "16.15.0",
+    "npm": "8.10.0"
   },
   "dependencies": {
     "@govuk-react/constants": "^0.10.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -61,7 +61,7 @@
   },
   "license": "MIT",
   "engines": {
-    "node": "16.14.0"
+    "node": "16.15.0"
```
craigjbass, line 64: I wouldn't change this without changing the buildpacks. This sets the `node.js` version not the `npm` version

craigjbass: https://docs.npmjs.com/cli/v8/configuring-npm/package-json#engines

You can specify node and npm versions that are enforced locally.

```{
  "engines": {
    "node": "16.14.0",
    "npm": "XXX"
  }
}```

vfarah-if: Nice one, 8.10.0 actually did the job, rather than the one that comes with the latest node

craigjbass (approved):

tomkingston (approved):

PR #4583, opened 2022-05-24
Display Aventri events attended status

chloehkwong1:
## Description of change

This PR pulls through a field from Activity Stream called `virtual_event_attendance` which is set to "Yes" when virtual attendance has been confirmed, or an empty string if it hasn't been/ the information is unknown.

If virtual attendance has been confirmed, the 'Attended' status will be shown at the end of the event name. If not, then nothing will be shown. 

## Test instructions
- Go to Data Hub dev django admin and turn on the user feature flag `user-contact-activities` on your adviser profile.
- Using branch `feature/DET-181-display-aventri-attended-status` and the dev api go to the contact John Doe http://localhost:3000/contacts/68be55fc-544d-4d9c-a264-977b917f1fd3/interactions?featureTesting=user-contact-activities
- If you scroll down, you should be able to see three events. One with nothing at the end, and two with the Attended status. 

## Screenshots
**Before**
image

**After**
image

## Checklist

- [X] Has the branch been rebased to master?

### DIFF
--- a/src/client/components/ActivityFeed/activities/AventriAttendee.jsx
+++ b/src/client/components/ActivityFeed/activities/AventriAttendee.jsx
@@ -3,6 +3,8 @@ import PropTypes from 'prop-types'
 import CardUtils from './card/CardUtils'
 import { ACTIVITY_TYPE } from '../constants'
 import { formatStartAndEndDate } from '../../../utils/date'
+import { GREY_1 } from 'govuk-colours'
+import styled from 'styled-components'
 
 import ActivityCardWrapper from './card/ActivityCardWrapper'
 import ActivityCardSubject from './card/ActivityCardSubject'
@@ -18,6 +20,12 @@ const transformAventriAttendee = (activity) => ({
     activity.object['dit:aventri:virtual_event_attendance'] == 'Yes',
 })
 
+const StyledSpan = styled('span')`
+  & > span {
+    color: ${GREY_1};
+  }
+`
+
 export default function AventriAttendee({ activity }) {
   const { eventName, date, isVirtualAttendanceConfirmed } =
     transformAventriAttendee(activity)
@@ -28,9 +36,9 @@ export default function AventriAttendee({ activity }) {
       <ActivityCardSubject>
         {eventName}
         {isVirtualAttendanceConfirmed && (
-          <span>
+          <StyledSpan>
             : <span>{VIRTUAL_EVENT_ATTENDANCE_STATUS}</span>
-          </span>
+          </StyledSpan>
         )}
       </ActivityCardSubject>
       <ActivityCardMetadata
--- a/src/client/components/ActivityFeed/activities/AventriAttendee.jsx
+++ b/src/client/components/ActivityFeed/activities/AventriAttendee.jsx
@@ -9,14 +9,15 @@ import ActivityCardSubject from './card/ActivityCardSubject'
 import ActivityCardMetadata from './card/ActivityCardMetadata'
 import ActivityCardLabels from './card/ActivityCardLabels'
 
+const VIRTUAL_EVENT_ATTENDANCE_STATUS = 'Attended'
+
 const transformAventriAttendee = (activity) => ({
   eventName: activity.eventName,
   date: formatStartAndEndDate(activity.startDate, activity.endDate),
   isVirtualAttendanceConfirmed:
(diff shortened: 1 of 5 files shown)


src/client/components/ActivityFeed/activities/card/ActivityCardSubject.jsx
```diff
@@ -17,10 +17,17 @@ const StyledActivitySubject = styled('h3')`
...
   }
+  & > span > span {
+    color: ${GREY_1};
+  }
 `
 
 const ActivityCardSubject = ({ children }) => {
-  return <StyledActivitySubject>{children}</StyledActivitySubject>
+  return (
+    <span>
```
Harriethw, line 27: I think we can remove this enclosing span, and make the styling `& > span`

src/client/components/ActivityFeed/activities/AventriAttendee.jsx
```diff
@@ -9,17 +9,34 @@ import ActivityCardSubject from './card/ActivityCardSubject'
...
 
   return (
     <ActivityCardWrapper dataTest="aventri-activity">
       <ActivityCardLabels service="event" kind="aventri service delivery" />
-      <ActivityCardSubject>{eventName}</ActivityCardSubject>
+      <ActivityCardSubject
+        isVirtualAttendanceConfirmed={isVirtualAttendanceConfirmed}
+      >
+        {eventName}
+        {isVirtualAttendanceConfirmed == 'Yes' ? (
```
Harriethw, line 32: I wonder if we can simplify this a bit, so that the logic is handled by the transformer. something like this:

```
const transformAventriAttendee = (activity) => ({
  eventName: activity.eventName,
  date: formatStartAndEndDate(activity.startDate, activity.endDate),
  isVirtualAttendanceConfirmed:
    activity.object['dit:aventri:virtual_event_attendance'] == 'Yes',
})

....

  return (
    <ActivityCardWrapper dataTest="aventri-activity">
      <ActivityCardLabels service="event" kind="aventri service delivery" />
      <ActivityCardSubject>
        {eventName}
        {isVirtualAttendanceConfirmed && (
          <span>
            : <span>{VIRTUAL_EVENT_ATTENDANCE_STATUS}</span>
          </span>
        )}
      </ActivityCardSubject>
     ...
    
   ```

File: test/functional/cypress/specs/contacts/activity-spec.js [hunk 3]
```diff
@@ -139,9 +139,9 @@ describe('Contact activity', () => {
       })
 
       context('when viewing a Contact with Aventri activity', () => {
-        it('should diplay event name from Aventri ', () => {
+        it('should display event name and virtual event attendance from Aventri ', () => {
```
cgsunkel, line 142: It would be nice to have a test for a non-attended event as well

cgsunkel (approved):
I have one minor comment but looks good otherwise so will approve

src/client/components/ActivityFeed/activities/card/ActivityCardSubject.jsx
```diff
@@ -17,10 +17,17 @@ const StyledActivitySubject = styled('h3')`
     text-decoration: none;
     color: ${BLUE};
   }
+  & > span > span {
```
Harriethw, line 20: minor/opinion: i'm a bit nervous about styling an anonymous nested span as it might unintentionally affect content put in, but I'm not sure exactly how we'd solve it. Maybe by styling the element in `AventriAttendee` as that's the only place we need this styling atm?

Harriethw (commented):
Nice work guys! I think it looks really good mostly, just had a few comments/feedback. Sorry I've left them a bit abruptly, let me know if you'd like to go through them and we can find some time to pair.

src/client/components/ActivityFeed/activities/card/ActivityCardSubject.jsx
dredmonds: Done and considered the comment, which reduced a nested span to display `Attended`.

src/client/components/ActivityFeed/activities/card/ActivityCardSubject.jsx
chloehkwong1: We've removed this span now thanks @Harriethw !

Harriethw (approved):
nice work, thanks both!

PR #4592, opened 2022-05-30
Add component tests for new components

cgsunkel:
## Description of change

This PR adds component tests for two of the new components I added recently (`ArchivePanel` and `ContactLocalHeader`). These components have also been added to Storybook. This work was not done in the original PRs as they were quite large.

## Test instructions

The two new components should be visible in Storybook. Everything else should work as before.

## Checklist


src/client/components/ArchivePanel/index.jsx
```diff
@@ -61,7 +61,7 @@ const ArchivePanel = ({
 
 ArchivePanel.propTypes = {
   isArchived: PropTypes.bool.isRequired,
-  archivedBy: PropTypes.object.isRequired,
+  archivedBy: PropTypes.object,
```
cgsunkel, line 64: Setting this to be required was causing a console error for automatically archived records (which don't have an `archived_by` prop).

test/functional/cypress/specs/contacts/details-spec.js
```diff
@@ -269,36 +261,4 @@ describe('View contact details', () => {
       })
     })
   })
-
-  context('when the activities feature flag is set', () => {
```
cgsunkel, line 273: These tests (and the other ones that were removed) are now covered by the component test.

santoshdasa12345 (approved):

test/component/cypress/specs/ContactLocalHeader.test.jsx
```diff
@@ -0,0 +1,163 @@
...
+const notPrimaryContact = require('../../../sandbox/fixtures/v3/contact/contact-incomplete-details-uk.json')
+
+const companyName = primaryContact.company.name
+const companyLink = urls.companies.details(primaryContact.company.id)
+const contactName = primaryContact.name
+const addInteractionUrl = urls.companies.interactions.create(
+  primaryContact.company.id
+)
+
+describe('ContactLocalHeader', () => {
```
mforner13, line 19: Equally with this test, I think testing the local header component itself rather than an implementation of it would be better, as we would then cover all uses of LocalHeader.

mforner13 (changes requested):
Some comments on writing stories/tests for `LocalHeader` rather than the contact header. Also - are there any functional tests testing the archive panel which can now be removed, if covered by the component tests? :)

src/client/components/ContactLocalHeader/__stories__/ContactLocalHeader.stories.jsx
```diff
@@ -0,0 +1,42 @@
+import React from 'react'
```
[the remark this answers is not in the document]
cgsunkel: The `CompanyLocalHeader` has an entry in Storybook so I don't see why this shouldn't have one as well. There are a few other things in Storybook that aren't reusable outside of certain sub-apps.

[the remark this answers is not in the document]
mforner13: I think it's a mistake to have the companyLocalHeader in storybook, so don't think that's a good example to follow. As a general rule, if it won't be reused then there's not much utility having it in Storybook as it's just duplicating the implementation code. Whereas having the LocalHeader in storybook would be useful as it's a widely reused component, and could then mean we can remove the company local header entry too :)

mforner13 (approved):

PR #4593, opened 2022-05-31
Add `userFeatureFlags` to global React store

Harriethw:
## Description of change

We have a story coming up where we will need to check a user's feature
flags on the Events collection page. Because the current function for
checking [user feature flags is done via the Express routing middleware](https://github.com/uktrade/data-hub-frontend/blob/master/src/middleware/user-features.js#L43-L44),
we need a way to check the flags when routing is done via React.

I noticed that the `userFeatures` middleware was doing two things - setting the `userFeatures` in `res.locals` if it was not already set, and then adding the specified user feature flag to the path. I also noticed that we are setting environment-level feature flags at [the server level by default](https://github.com/uktrade/data-hub-frontend/blob/71e5139bfb7c8c42e75a738830f7fa4baf65f6a7/src/server.js#L135) and that we have a function for adding global props to the redux store - for things [like `modulePermissions` and `clientAdviserId`](https://github.com/uktrade/data-hub-frontend/blob/master/src/middleware/react-global-props.js#L11)

I'm proposing that we 
1. decouple the checking and setting of userFeatures in `res.locals` so that they are always set
2. add the user feature flags to the react global store, so that we will be able to access them in React-land
e.g. a `mapState2Props` function would look in the state and pass the value to component props like so:

```
...

 const isContactActivitiesFeatureOn =
    state.userFeatureFlags.includes(
        'user-contact-activities'
      )

return {
 ...state[ID],
 isContactActivitiesFeauteOn,
}

```

## Feedback

* I feel like there must be a simpler approach - in previous projects I've worked on this would have been done in a hook or context provider purely in React land, but I'm not sure I really grasp how to do it in this repo. I guess another way could be having a component that runs a Task to get the user features and passes them to children... 
* testing - the e2e tests were failing because they weren't able to make an authorisedrequrest to the `whoami` endpoint, so I'm failing the function silently, not sure how to put a test framework around this, or if it's needed, as we will test in functional tests when user feature flag is implemented? like we are doing here: https://github.com/uktrade/data-hub-frontend/blob/master/test/functional/cypress/specs/contacts/activity-spec.js#L11 
* I'm not an expert on Express, is this the right way to use middleware?
* naming - I notice I'm not really following the existing naming convention in the app for middleware - e.g. `features` instead of `setFeatures` but went with more verbose as it seemed better to be explicit

## Test instructions

_What should I see?_

## Checklist

### DIFF
--- a/src/middleware/react-global-props.js
+++ b/src/middleware/react-global-props.js
@@ -17,6 +17,7 @@ module.exports = () => {
         ...res.locals.PERMITTED_APPLICATIONS.map(({ key }) => key),
       ]),
       currentAdviserId: req.session?.user?.id,
+      userFeatureFlags: res.locals.userFeatures,
     }
     next()
   }
--- a/src/apps/contacts/router.js
+++ b/src/apps/contacts/router.js
@@ -26,7 +26,7 @@ const interactionsRouter = require('../interactions/router.sub-app')
 const {
   fetchActivitiesForContact,
 } = require('../companies/apps/activity-feed/controllers')
-const userFeatures = require('../../middleware/user-features')
+const { checkUserFeatures } = require('../../middleware/user-features')
 
 router.get(urls.contacts.index(), renderContactsView)
 router.get(['/create', '/:contactId/edit'], createAndEdit)
@@ -43,7 +43,7 @@ router.use(
 router.get('/:contactId', redirectToFirstNavItem)
 router.get(
   '/:contactId/details',
-  userFeatures('user-contact-activities'),
+  checkUserFeatures('user-contact-activities'),
   getDetails
 )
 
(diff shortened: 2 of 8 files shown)


src/middleware/react-global-props.js
```diff
@@ -17,6 +17,7 @@ module.exports = () => {
         ...res.locals.PERMITTED_APPLICATIONS.map(({ key }) => key),
       ]),
       currentAdviserId: req.session?.user?.id,
+      userFeatureFlags: res.locals.features,
```
elcct, line 20: I am not sure if `.features` stores user feature flags. I think that contains global feature flags. User feature flags seem to be coming from `res.locals.userFeatures` filled by `user-feature` middleware for apps that use it.

Harriethw: and that function is called here

Harriethw: looks like it includes our user flag `user-contact-activities`. maybe the key in the store should just be `featureFlags` instead of specifically `userFeatureFlags`?

Harriethw:
closing for another solution

PR #4536, opened 2022-04-22
Add new design for interaction activities

cgsunkel:
## Description of change

As part of the ongoing work to re-design the contact activity feed, the interaction card has been redesigned. This change is only visible for users with the relevant feature flag enabled and does not affect the company feed (or any other ActivityFeeds).

Screenshot 2022-04-22 at 15 20 02

## Test instructions

With the `user-contact-activities` flag enabled and your local FE connected to the dev environment, go to http://localhost:3000/contacts/715e8a58-6515-4d52-a4b5-84cb49566b01/interactions?featureTesting=user-contact-activities. You should see various interactions showcasing various options available with the new design.

## Screenshots
If the feature flag is disabled, no changes will be visible.

### Before (with feature flag enabled)

Screenshot 2022-04-22 at 15 13 02

### After

Screenshot 2022-04-22 at 15 20 15

## Checklist

[//]: # "When submitting a PR make sure the code review guidelines have been satisfied.
https://github.com/uktrade/data-hub-frontend/blob/master/docs/Code%20review%20guidelines.md"

- [ ] Has the branch been rebased to master?
- [ ] Automated tests (Any of the following when applicable: Unit, Functional or End-to-End)
- [ ] Manual compatibility testing (Browsers: Chrome, Firefox, Edge, Safari)

PR #4602, opened 2022-06-08
Add component test for new interaction card

cgsunkel:
## Description of change

This PR adds a component test for the new interaction `ActivityFeed` card introduced in #4536.

The bulk of this PR is made up of [very similar tests](https://github.com/uktrade/data-hub-frontend/blob/675231ec759b4ebc2323828baab5cd4aaccd5c71/test/component/cypress/specs/ActivityFeed/InteractionActivity.test.jsx#L284) I was unable to abstract, and if/when the remaining activity cards are switched to the new card design some of the assertions can be reused.

## Test instructions

The new test should pass. Nothing else should change.

## Checklist

### DIFF
--- a/src/client/components/ActivityFeed/activities/card/ActivityCardLabels.jsx
+++ b/src/client/components/ActivityFeed/activities/card/ActivityCardLabels.jsx
@@ -45,7 +45,7 @@ const ActivityCardLabels = ({ theme, service, kind }) => (
 ActivityCardLabels.propTypes = {
   theme: PropTypes.string,
   service: PropTypes.string,
-  kind: PropTypes.string,
+  kind: PropTypes.string.isRequired,
 }
 
 export default ActivityCardLabels
--- a/src/client/components/ActivityFeed/activities/card/ActivityCardLabels.jsx
+++ b/src/client/components/ActivityFeed/activities/card/ActivityCardLabels.jsx
@@ -35,11 +35,9 @@ const ActivityCardLabels = ({ theme, service, kind }) => (
       )}
     </TagColumn>
     <TagColumn>
-      {kind && (
-        <Tag data-test="activity-kind-label" colour="grey">
-          {kind}
-        </Tag>
-      )}
+      <Tag data-test="activity-kind-label" colour="grey">
+        {kind}
+      </Tag>
     </TagColumn>
   </TagRow>
 )
--- a/test/component/cypress/specs/ActivityFeed/InteractionActivity.test.jsx
+++ b/test/component/cypress/specs/ActivityFeed/InteractionActivity.test.jsx
@@ -1,0 +1,709 @@
+import React from 'react'
+import { mount } from '@cypress/react'
+
+import Interaction from '../../../../../src/client/components/ActivityFeed/activities/Interaction'
+import urls from '../../../../../src/lib/urls'
+import { getServiceText } from '../../../../../src/client/components/ActivityFeed/activities/InteractionUtils'
+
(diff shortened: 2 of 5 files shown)


src/client/components/ActivityFeed/activities/card/ActivityCardLabels.jsx
```diff
@@ -35,11 +35,9 @@ const ActivityCardLabels = ({ theme, service, kind }) => (
       )}
     </TagColumn>
     <TagColumn>
-      {kind && (
```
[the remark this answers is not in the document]
Harriethw: should we update the proptypes to show that `kind` is required?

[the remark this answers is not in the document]
cgsunkel: I have done this. I wonder if we should have similar behaviour for the `theme` label (set it to always display 'Other' if it isn't one of the three options).

test/component/cypress/specs/ActivityFeed/InteractionActivity.test.jsx
```diff
@@ -0,0 +1,709 @@
...
+      it('should render the kind label', () => {
+        assertKindLabel()
+      })
+
+      it('should render the interaction subject', () => {
+        assertInteractionSubject()
+      })
+
+      it('should render the interaction notes', () => {
+        assertNotes()
```
Harriethw, line 315: question/clarification: it's really great to see this stuff getting covered by tests, but I'm wondering if we need the repetition in these 'incomplete' contexts? imo we don't need to, for example, cover that the `subject` is being rendered when the `theme` is missing because there is no dependency there. The logic branches off at the point where we check `theme &&`  in `ActivityCardLabels`. We already have a test for whether most of these elements render in the first context, so I'm not sure we're getting any value from checking them again here - I'd be happy to chat about this though as I might be misunderstanding the components :)

cgsunkel: I wasn't sure whether to include these or not, I will remove some of these tests

cgsunkel: I have done this. I have kept the tests around related elements appearing (e.g. where the theme is missing only I've kept the tests around the service/kind labels) and removed all the others.

test/component/cypress/specs/ActivityFeed/InteractionActivity.test.jsx
```diff
@@ -0,0 +1,601 @@
...
+}
+
+const assertKindLabel = (expectedText = 'interaction') => {
+  assertText('[data-test=activity-kind-label]', expectedText)
+}
+
+const assertServiceLabel = (
+  service = interactionServices.specificDITService
+) => {
+  assertText('[data-test=activity-service-label]', getServiceText(service))
```
Harriethw, line 561: another bit of feedback that I'd be happy to talk through in person:
definitely really good to get that chunky `getServiceText` function covered by code, I'm wondering if you think unit tests have a place here?  I would usually test a function like this that is just transforming strings with a unit test for all the use cases - e.g. I pass 'Investment - IST Aftercare Offered (IST use only)' in then I get `IST Aftercare` back - and then the integration tests can test once that we called the function. Maybe something that can be looked at in a future PR though as I know this has been a bit long running! Also I know that there's some debate about unit tests 🙃

Harriethw (approved):

elcct (approved):

PR #4663, opened 2022-06-23
Update e2e docs

Harriethw:
## Description of change

I found some issues with the existing instructions when trying to run e2e tests locally.

After running `start-uat.sh` it was possible to successfully create an object, e.g. an Interaction, but it was not possible to find that object again when going through a `search` endpoint (e.g a Collection List page). Network requests all returned 200 so it wasn't because of a network or auth failure.

This could be because the `start-uat.sh` script doesn't create a celery instance, so opensearch is not updating? Or the difference in running `docker compose up` instead of `docker compose run`?  Changing the existing `docker-compose` file worked for me, but there may be a more automated way of doing it.

## Checklist


Harriethw:
closing as this solution doesn't seem to work for everyone

PR #4732, opened 2022-07-12
Add search local header component to substitute nunjucks header

PippoRaimondiDIT:
## Description of change

Use react local header version instead of nunjucks.

## Test instructions

All tests should function as it used to.

### Before

Screenshot 2022-07-12 at 18 09 52

### After

Screenshot 2022-07-12 at 18 10 10

## Checklist


src/client/components/SearchLocalHeader/index.jsx
```diff
@@ -0,0 +1,82 @@
...
+
+const StyledHeader = styled('div')`
+  padding-bottom: ${SPACING.SCALE_5};
+  background-color: ${GREY_LEGACY_HEADER};
+  padding-top: ${SPACING.SCALE_6};
+`
+
+const SearchLocalHeader = ({ csrfToken, flashMessages }) => (
+  <StyledHeader
+    aria-label="local header"
```
Harriethw, line 47: question: super cool to see aria labels and roles being used! I don't have much experience with them, do you have any resources for when we should be applying them? I understood that they were intended for interactive elements, as you have used the `search` role on the input below, does this label and role provide any use for our users? just wondering if it would make more noise for them, or if 'local header' has a meaning to them...

cgsunkel (approved):
LGTM, I have left one very small comment but will pre-approve

src/client/components/SearchLocalHeader/index.jsx
PippoRaimondiDIT: Other than the search button we will have flash messages being displayed in this component for example... so it could have been a bit odd if flash message was associated to search button component only... that was my line of thought...

src/client/components/SearchLocalHeader/index.jsx
```diff
@@ -0,0 +1,82 @@
...
+import Main from '@govuk-react/main'
+
+import urls from '../../../lib/urls'
+import SearchButton from '../SearchButton'
+import { BLUE } from 'govuk-colours'
+
+import FlashMessages from '../LocalHeader/FlashMessages'
+
+const GREY_LEGACY_HEADER = '#dee0e2'
+const GREY_LEGACY_PLACEHOLDER_TEXT = '#505a5f'
```
paulgain, line 16: It's best to have our colours in one central place. Normally we import colours in from [govuk-colours](https://github.com/penx/govuk-colours/blob/master/src/index.js) but it's missing a few key colours from the GOV.UK Design System [modern colour palette](https://github.com/alphagov/govuk-frontend/blob/main/src/govuk/settings/_colours-palette.scss#L37). Therefore, we have a [local file](https://github.com/uktrade/data-hub-frontend/blob/feature/rr-388-update-localheader/src/client/utils/colors.js) which defines what's missing plus a few legacy colours that you could reuse.

So  `GREY_LEGACY_HEADER` can be swapped out for this [guy](https://github.com/uktrade/data-hub-frontend/blob/feature/rr-388-update-localheader/src/client/utils/colors.js#L36).

And `GREY_LEGACY_PLACEHOLDER_TEXT` can be swapped out for this [guy](https://github.com/uktrade/data-hub-frontend/blob/feature/rr-388-update-localheader/src/client/utils/colors.js#L25).

Note that `#505a5f` is a modern colour.

paulgain (approved):

PR #4764, opened 2022-07-26
Add event name filter to new events collection list

chloehkwong1:
## Description of change
This PR adds an Event Name filter to the new Events Collection page. 
It works by modifying the query that is sent to Activity Stream. 

## Test instructions
- In Django dev API, add the user feature flag user-event-activities to your adviser profile.
- Bring up this branch on your local dev frontend and navigate to http://localhost:3000/events
- You should see an Event name filter on the left hand side. 
- In the filter box type 'project zeus' and press enter. It should add project zeus to the query params in the url and it should only be displaying events called Project Zeus. 
- In the filter box backspace everything and press enter. All the events should reappear. 

### Before
image

### After
image

## Checklist

### DIFF
--- a/src/apps/companies/apps/activity-feed/controllers.js
+++ b/src/apps/companies/apps/activity-feed/controllers.js
@@ -355,7 +355,7 @@ async function fetchAventriAttendees(req, res, next) {
   }
 }
 
-const filtersQueryBuilder = (name) => {
+const eventsColListQueryBuilder = (name) => {
   // return query if name is present, else return null
   const eventNameFilter = name
     ? {
@@ -383,7 +383,7 @@ async function fetchAllActivityFeedEvents(req, res, next) {
     const allActivityFeedEventsResults = await fetchActivityFeed(
       req,
       allActivityFeedEventsQuery({
-        filtersQuery: filtersQueryBuilder(name),
+        fullQuery: eventsColListQueryBuilder(name),
         sort:
           EVENT_ACTIVITY_SORT_OPTIONS[sortBy] ||
           EVENT_ACTIVITY_SORT_OPTIONS['modified_on:desc'],
@@ -409,5 +409,5 @@ module.exports = {
   fetchAventriEvent,
   fetchAllActivityFeedEvents,
   fetchAventriAttendees,
-  filtersQueryBuilder,
+  eventsColListQueryBuilder,
 }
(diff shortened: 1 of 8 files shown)


test/functional/cypress/specs/events/filter-spec.js
```diff
@@ -42,411 +44,443 @@ const eventTypeEndpoint = '/api-proxy/v4/metadata/event-type'
 const ukRegionsEndpoint = '/api-proxy/v4/metadata/uk-region'
 
 describe('events Collections Filter', () => {
```
chloehkwong1, line 46: I added a contexts for without feature flag and with feature flag which is why it looks like I've modified the whole file. My actual changes are on lines 456 onwards.

src/apps/companies/apps/activity-feed/controllers.js
```diff
@@ -354,13 +355,35 @@ async function fetchAventriAttendees(req, res, next) {
...
+    ? {
+        match: {
+          'object.name': name,
+        },
+      }
+    : null
+
+  const filtersArray = [eventNameFilter]
+
+  // do not pass in filter if its value is null
```
elcct, line 370: I think this comment is redundant. The filter is self explanatory.

elcct (approved):

src/apps/companies/apps/activity-feed/controllers.js
```diff
@@ -354,13 +355,35 @@ async function fetchAventriAttendees(req, res, next) {
...
+  const cleansedFiltersArray = filtersArray
+    .filter((filter) => filter != null)
+    .map((filter) => filter)
```
Harriethw, line 373: question: do you need to map through a second time? doesn't filter return an array already?

src/apps/companies/apps/activity-feed/controllers.js
chloehkwong1: Hi @elcct good point! I've removed this comment and the previous one and it was also pretty obvious what was happening. Thanks!

src/apps/companies/apps/activity-feed/controllers.js
chloehkwong1: Hi @Harriethw good spot thank you! It was a bit of legacy code from when I was filtering the keys in an object and then  mapping through to return an array of only the object values. I got rid of the object format now so yeh I don't need that second line anymore but totally didn't clock until you pointed it out :)

richmadetech (approved):

PR #4765, opened 2022-07-27
Explore flaky tests with quick solutions

Harriethw:
## Description of change

Fix for two flaky tests:

1. `events/filter-spec` - the command to select the first typeahead option was being triggered before the option had rendered, due to the data being throttled/network latency. This is resolved by intercepting the network request and directly returning mock data, rather than making a 'real' request to the sandbox. We aren't losing any test value here because it's still mocking the same logic as before, just without network latency. 
2.  `events/edit-spec` - this test was flakey as it would sometimes check form values before the form had been populated by an API request, the solution is to wait for that request to complete which seems to give it time to populate the form. 

## Test instructions

No flakey tests being logged for this **fix/flakey-event-filter-spec** branch

## Checklist

- [x Has the branch been rebased to master?

### DIFF
--- a/test/functional/cypress/specs/interaction/filter-spec.js
+++ b/test/functional/cypress/specs/interaction/filter-spec.js
@@ -195,6 +195,7 @@ describe('Interactions Collections Filter', () => {
       selectFirstAdvisersTypeaheadOption({
         element: advisersFilter,
         input: adviser.name,
+        mockAdviserResponse: false,
       })
       cy.wait('@adviserListApiRequest')
       cy.wait('@adviserApiRequest')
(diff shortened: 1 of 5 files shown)


test/functional/cypress/specs/interaction/filter-spec.js
```diff
@@ -203,6 +204,7 @@ describe('Interactions Collections Filter', () => {
       assertTypeaheadOptionSelected({
         element: advisersFilter,
         expectedOption: adviser.name,
+        mockAdviserResponse: false,
```
Harriethw, line 207: @vfarah-if FYI this `mockAdviserResponse` and the one on line 252 are being passed into a function that doesn't ask for it (`assertTypeaheadOptionSelected`, which is different to `selectFirstAdvisersTypeaheadOption`)

dredmonds (approved):

cgsunkel (approved):

PR #4778, opened 2022-08-01
Add start date filters to new events collection page

chloehkwong1:
## Description of change
This PR adds in 'Earliest start date' and 'Latest start date' filters to the new events collection page via modification to the Activity Stream OpenSearch query. 

The page has also been refactored so that the header and filters match the layout in the existing events collection list. 

## Test instructions
- In Django dev API, add the user feature flag user-event-activities to your adviser profile.
- Bring up this branch on your local dev frontend and navigate to http://localhost:3000/events
- You should see an 'Earliest start date' filter and a 'Latest start date' filter on the left hand side.
- It should work if you add dates in either/both these filters and should also add to query string in the url. 

### Before
image

### After
image

## Checklist

### DIFF
--- a/src/apps/companies/apps/activity-feed/__test__/controllers.test.js
+++ b/src/apps/companies/apps/activity-feed/__test__/controllers.test.js
@@ -545,7 +545,7 @@ describe('Activity feed controllers', () => {
         }
 
         const actualQuery = activityFeedEventsQuery({
-          fullQuery: eventsColListQueryBuilder(name),
+          fullQuery: eventsColListQueryBuilder({ name: name }),
           from,
           size,
           sort: EVENT_ACTIVITY_SORT_OPTIONS['modified_on:desc'],
@@ -583,7 +583,7 @@ describe('Activity feed controllers', () => {
         const actualQuery = activityFeedEventsQuery({
           from,
           size,
-          fullQuery: eventsColListQueryBuilder(name),
+          fullQuery: eventsColListQueryBuilder({ name: name }),
           sort: EVENT_ACTIVITY_SORT_OPTIONS['modified_on:desc'],
         })
 
@@ -641,7 +641,10 @@ describe('Activity feed controllers', () => {
           const endDate = undefined
 
           const actualQuery = activityFeedEventsQuery({
-            fullQuery: eventsColListQueryBuilder(startDate, endDate),
+            fullQuery: eventsColListQueryBuilder({
+              startDate: startDate,
+              endDate: endDate,
+            }),
             sort: EVENT_ACTIVITY_SORT_OPTIONS['modified_on:desc'],
           })
 
@@ -653,11 +656,14 @@ describe('Activity feed controllers', () => {
           const endDate = '2022-12-12T08:39:06'
 
           const actualQuery = activityFeedEventsQuery({
-            fullQuery: eventsColListQueryBuilder(startDate, endDate),
+            fullQuery: eventsColListQueryBuilder({
+              startDate: startDate,
+              endDate: endDate,
+            }),
             sort: EVENT_ACTIVITY_SORT_OPTIONS['modified_on:desc'],
           })
 
-          expect(expectedEsQuery).to.deep.equal(actualQuery)
+          expect(expectedEsQuery(startDate, endDate)).to.deep.equal(actualQuery)
         })
 
         it('builds the right query when the start and end date is selected', () => {
@@ -665,7 +671,10 @@ describe('Activity feed controllers', () => {
           const endDate = '2022-12-12T08:39:06'
 
           const actualQuery = activityFeedEventsQuery({
-            fullQuery: eventsColListQueryBuilder(startDate, endDate),
+            fullQuery: eventsColListQueryBuilder({
+              startDate: startDate,
+              endDate: endDate,
+            }),
             sort: EVENT_ACTIVITY_SORT_OPTIONS['modified_on:desc'],
           })
 
@@ -727,7 +736,11 @@ describe('Activity feed controllers', () => {
       }
 
       const actualQuery = activityFeedEventsQuery({
-        fullQuery: eventsColListQueryBuilder(name, startDate, endDate),
+        fullQuery: eventsColListQueryBuilder({
+          name: name,
+          startDate: startDate,
+          endDate: endDate,
+        }),
         sort: EVENT_ACTIVITY_SORT_OPTIONS['modified_on:desc'],
       })
 
--- a/src/apps/companies/apps/activity-feed/__test__/controllers.test.js
+++ b/src/apps/companies/apps/activity-feed/__test__/controllers.test.js
@@ -1,20 +1,21 @@
 const activityFeedEsFixtures = require('../../../../../../test/unit/data/activity-feed/activity-feed-from-es.json')
+const allActivityFeedEvents = require('../../../../../../test/sandbox/fixtures/v4/activity-feed/all-activity-feed-events.json')
 const buildMiddlewareParameters = require('../../../../../../test/unit/helpers/middleware-parameters-builder')
 const companyMock = require('../../../../../../test/unit/data/company.json')
 const {
   DATA_HUB_ACTIVITY,
   EXTERNAL_ACTIVITY,
   DATA_HUB_AND_EXTERNAL_ACTIVITY,
-  EVENT_ACTIVITY_SORT_OPTIONS,
 } = require('../constants')
 const { eventsColListQueryBuilder } = require('../controllers')
-const activityFeedEventsQuery = require('../es-queries/activity-feed-all-events-query')
 
 describe('Activity feed controllers', () => {
   let fetchActivityFeedStub,
+    fetchAllActivityFeedEventsStub,
     getGlobalUltimateHierarchyStub,
     controllers,
     middlewareParameters
+
   describe('#fetchActivityFeedHandler', () => {
     before(() => {
       fetchActivityFeedStub = sinon.stub().resolves(activityFeedEsFixtures)
@@ -511,6 +512,7 @@ describe('Activity feed controllers', () => {
       })
     })
   })
+
   describe('#eventsColListQueryBuilder', () => {
     context('check query builder when filtering on event name', () => {
       it('builds the right query when being filtered by event name', () => {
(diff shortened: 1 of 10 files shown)


src/client/components/NewFilteredCollectionList/index.jsx
```diff
@@ -0,0 +1,114 @@
...
+
+import Task from '../Task'
+import CollectionSort from '../CollectionList/CollectionSort'
+
+import RoutedPagination from '../Pagination/RoutedPagination'
+import CollectionHeader from '../CollectionList/CollectionHeader'
+import ActivityList from '../ActivityFeed/activities/card/ActivityList'
+import Activity from '../ActivityFeed/Activity'
+
+const NewFilteredCollectionList = ({
```
paulgain, line 19: What is the thinking behind the name `NewFilteredCollectionList`? Is the plan to replace all instances of `FilteredCollectionList`?

src/apps/companies/apps/activity-feed/__test__/controllers.test.js
```diff
@@ -583,12 +585,163 @@ describe('Activity feed controllers', () => {
...
+          ).to.deep.equal(actualQuery)
+        })
+
+        it('builds the right query when the end date is selected', () => {
+          const earliestStartDate = undefined
+          const latestStartDate = '2022-12-12T08:39:06'
+          const from = 0
+          const size = 10
+
+          const actualQuery = activityFeedEventsQuery({
```
Harriethw, line 659: question: this seems like a unit test for `eventsColListQueryBuilder`, and I'm wondering as it's a unit test if we need to include the `activityFeedEventsQuery` as well? I guess if it's included it's technically testing both functions?

src/apps/companies/apps/activity-feed/__test__/controllers.test.js
```diff
@@ -583,12 +585,163 @@ describe('Activity feed controllers', () => {
...
+          fullQuery: eventsColListQueryBuilder({ name: name }),
           sort: EVENT_ACTIVITY_SORT_OPTIONS['modified_on:desc'],
         })
 
         expect(expectedEsQuery).to.deep.equal(actualQuery)
       })
     })
+
+    context(
+      'check dsl query when filtering on event start and end date',
```
Harriethw, line 597: question: should this be within it's own `describe` block? It's currently in `#filtersQueryBuilder` which is maybe another function than the ones this is testing?

src/client/components/NewFilteredCollectionList/index.jsx
```diff
@@ -0,0 +1,114 @@
...
+  defaultQueryParams,
+  allActivityFeedEvents,
+}) => {
+  const totalPages = Math.ceil(
+    Math.min(total, maxItemsToPaginate) / itemsPerPage
+  )
+
+  return (
+    <Route>
+      {({ history, location }) => {
```
Harriethw, line 35: minor: I think using a `Route` to get the history location is an old way of doing it, have you tried using the `useLocation`/`useHistory` hooks? I think there are examples of them elsewhere in the app  (e.g. in the Form)

src/client/components/NewFilteredCollectionList/index.jsx
```diff
@@ -0,0 +1,114 @@
...
+                {sortOptions && (
+                  <CollectionSort
+                    sortOptions={sortOptions}
+                    totalPages={totalPages}
+                  />
+                )}
+                <Task.Status {...taskProps}>
+                  {() => (
+                    <ActivityList>
+                      {allActivityFeedEvents?.map((event, index) => {
```
Harriethw, line 66: question: just thinking if this is a component that's currently specifically for activity feed events, should it have a more specific name, and then it can be made more generic if needed?

src/client/components/NewFilteredCollectionList/index.jsx
chloehkwong1: @paulgain thank you for bringing this up, it's a good point. No, the plan is to only replace the instance of `FilteredCollectionList` used for the Events collection list as there as been a redesign and it uses a different component for the card styling. Maybe in the future when we remove the feature flag we can work to combine it into the existing `FilteredCollectionList` but for now I just created a new file so the existing component doesn't become too messy. 

I've renamed this new component to `EventsFilteredCollectionList` as that's a more accurate representation of what it actually is.

src/apps/companies/apps/activity-feed/__test__/controllers.test.js
chloehkwong1: @Harriethw, i've now changed all the tests to only test the `eventsColListQueryBuilder` function, except for one test at the end which tests the whole query being sent to activity stream. Thanks for bringing this up :)

src/apps/companies/apps/activity-feed/__test__/controllers.test.js
chloehkwong1: @Harriethw I've now changed the tests to have two describe blocks: `#eventsColListQueryBuilder` and `#allActivityFeedEventsQuery` to separate out the two functions being tested. Thanks!

src/client/components/NewFilteredCollectionList/index.jsx
paulgain: @chloehkwong1 `EventsFilteredCollectionList` is without doubt more helpful. Any future changes to `NewFilteredCollectionList ` would have meant updating the name a second time to `NewNewFilteredCollectionList`. :-)

chloehkwong1: @paulgain Definitely agree, thanks :D

src/client/components/NewFilteredCollectionList/index.jsx
chloehkwong1: @Harriethw I think this is similar to Paul's comment? I've renamed it to `EventsFilteredCollectionList` for now, unless you have an alternative suggestion?

src/client/components/NewFilteredCollectionList/index.jsx
chloehkwong1: @paulgain, I've renamed to `ActivityStreamCollectionList` following Harriet's comment!

src/client/components/NewFilteredCollectionList/index.jsx
chloehkwong1: hey @Harriethw I think for now I'll keep it this way just for simplicity and look to refactor in the future! :)

src/client/components/NewFilteredCollectionList/index.jsx
chloehkwong1: I've renamed this to `ActivityFeedFilteredComponent` now :)

src/apps/companies/apps/activity-feed/__test__/controllers.test.js
```diff
@@ -511,53 +512,149 @@ describe('Activity feed controllers', () => {
...
+          },
+        ]
+
+        it('builds the right query when start date selected', () => {
+          const earliestStartDate = '2022-11-01T08:39:06'
+          const latestStartDate = undefined
+          const actualQuery = eventsColListQueryBuilder({
+            earliestStartDate: earliestStartDate,
+            latestStartDate: latestStartDate,
+          })
```
paulgain, line 581: This could be slightly improved for better readability:
```          
const actualQuery = eventsColListQueryBuilder({
    earliestStartDate,
    latestStartDate,
})
```

src/apps/companies/apps/activity-feed/__test__/controllers.test.js
```diff
@@ -511,53 +512,149 @@ describe('Activity feed controllers', () => {
...
+          ).to.deep.equal(actualQuery)
+        })
+
+        it('builds the right query when the end date is selected', () => {
+          const earliestStartDate = undefined
+          const latestStartDate = '2022-12-12T08:39:06'
+          const actualQuery = eventsColListQueryBuilder({
+            earliestStartDate: earliestStartDate,
+            latestStartDate: latestStartDate,
+          })
```
paulgain, line 594: Same applies here

src/apps/companies/apps/activity-feed/__test__/controllers.test.js
```diff
@@ -511,53 +512,149 @@ describe('Activity feed controllers', () => {
...
+          ).to.deep.equal(actualQuery)
+        })
+
+        it('builds the right query when the start and end date is selected', () => {
+          const earliestStartDate = '2022-11-01'
+          const latestStartDate = '2022-12-12T08:39:06'
+          const actualQuery = eventsColListQueryBuilder({
+            earliestStartDate: earliestStartDate,
+            latestStartDate: latestStartDate,
+          })
```
paulgain, line 607: ...and here

src/apps/companies/apps/activity-feed/controllers.js
```diff
@@ -370,26 +374,42 @@ const eventsColListQueryBuilder = (name) => {
...
     const from = (page - 1) * ACTIVITIES_PER_PAGE
 
     const allActivityFeedEventsResults = await fetchActivityFeed(
       req,
       allActivityFeedEventsQuery({
-        fullQuery: eventsColListQueryBuilder(name),
+        fullQuery: eventsColListQueryBuilder({
+          name: name,
+          earliestStartDate: earliestStartDate,
+          latestStartDate: latestStartDate,
```
paulgain, line 409: ...and here.

test/functional/cypress/specs/events/filter-spec.js
```diff
@@ -458,25 +458,117 @@ describe('events Collections Filter', () => {
...
       const queryParamWithName = 'name=Big+Event'
-      const queryParamEmpty = 'name='
 
-      it('should not add anything to the query param when the page is first loaded', () => {
-        cy.url().should('not.include', queryParamEmpty)
+      context('should filter from user input', () => {
+        before(() => {
+          cy.intercept(
+            'GET',
+            `${urls.events.activity.data()}?sortBy=modified_on:desc&name=Big+Event&page=1`
```
paulgain, line 471: Ideally we don't want to be using `urls.*` in place of API endpoints, `urls` are specific to express routes only. If you search for `cy.intercept` in your IDE you will see we define API endpoints as a constant if it's used more than once, otherwise, use it inline.

test/functional/cypress/specs/events/filter-spec.js
```diff
@@ -458,25 +458,117 @@ describe('events Collections Filter', () => {
...
-        cy.get(element).type(`${eventName}{enter}`)
-        cy.url().should('include', queryParamWithName)
+      after(() => {
+        cy.get(element).clear()
+      })
+    })
+
+    context('Start date', () => {
+      const earliestStartElement = '#field-earliest_start_date-1'
+      const latestStartElement = '#field-latest_start_date-1'
```
paulgain, line 509: Try avoiding IDs, we prefer to use the `data-test` attribute selectors for consistency.

paulgain (commented):
Looks good, just a few minor comments.

src/apps/companies/apps/activity-feed/__test__/controllers.test.js
chloehkwong1: @paulgain done :)

test/functional/cypress/specs/events/filter-spec.js
chloehkwong1: Hi @paulgain, this **is** an express route we are testing on the activity feed controller, not the activity feed api endpoint!

paulgain (approved):

richmadetech (approved):

PR #4783, opened 2022-08-02
Add pagination to Aventri attendees page

onyela-o:
## Description of change

This ticket adds pagination to the Aventri event attendees pages. We have replicated the behaviour on Data hub events attendees pages.

## Test instructions

- Ensure you have `user-event-activities` feature flag on. You can do this is dev django admin.
- Get this branch up locally 
- Visit this Aventri event page (as it's the only one with 10 or more attendees): http://localhost:3000/events/aventri/3333/attendees?featureTesting=user-event-activities

You should see:
- Number of attendees 
- Page X of Y
- URL with sort by and page as parameters
- Pagination on the bottom

### Before
image

### After
image

## Checklist

### DIFF
--- a/src/client/modules/Events/EventAventriAttendees/index.jsx
+++ b/src/client/modules/Events/EventAventriAttendees/index.jsx
@@ -23,9 +23,8 @@ import ActivityList from '../../../components/ActivityFeed/activities/card/Activ
 const EventAventriAttendees = ({
   aventriAttendees,
   aventriEventData,
-  selectedSortBy,
-  selectedPage,
-  page,
+  payload,
+  page = parseInt(page || 1, 10),
   totalAttendees,
   itemsPerPage = 10,
   maxItemsToPaginate = 10000,
@@ -64,7 +63,7 @@ const EventAventriAttendees = ({
               id={ID}
               progressMessage="Loading Aventri attendees"
               startOnRender={{
-                payload: { aventriEventId, selectedSortBy, selectedPage },
+                payload: { aventriEventId, ...payload },
                 onSuccessDispatch: EVENTS__AVENTRI_ATTENDEES_LOADED,
               }}
             >
@@ -121,10 +120,7 @@ const EventAventriAttendees = ({
           )
         }
       </CheckUserFeatureFlag>
-      <RoutedPagination
-        initialPage={page}
-        items={totalAttendees}
-      ></RoutedPagination>
+      <RoutedPagination initialPage={page} items={totalAttendees} />
     </DefaultLayout>
   )
 }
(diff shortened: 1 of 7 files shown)


test/functional/cypress/specs/events/aventri-attendees-spec.js
```diff
@@ -111,6 +122,32 @@ describe('Aventri event attendees', () => {
...
+            )}?sortBy=first_name:desc&page=1`
+          ).as('firstNameZ-A')
+          cy.visit(urls.events.aventri.attendees(existingEventId))
+        })
+        it('should be possible to page through', () => {
+          cy.get('[data-page-number="2"]').click()
+          cy.get('[data-test=pagination-summary]').contains('Page 2 of 4')
+          // cy.get('[data-test="aventri-attendee"]')
+          //   .find('[data-test="data-hub-attendee-name"]')
+          //   .should('exist')
```
vfarah-if, line 147: (suggestion) Remove comment if not needed

vfarah-if (approved):

test/functional/cypress/specs/events/aventri-attendees-spec.js
onyela-o: Oops forgot about this. Thanks Vince!

src/client/modules/Events/EventAventriAttendees/index.jsx
```diff
@@ -38,72 +50,108 @@ const EventAventriAttendees = ({
...
-                        </li>
-                      ))}
-                    </ActivityList>
-                  </GridCol>
-                </GridRow>
-              )}
-            </Task.Status>
-          )
+    <Route>
+      {({ history, location }) => {
```
Harriethw, line 59: question: have you tried using the `useHistory` and `useLocation` hooks? we use them elsewhere in the app instead of using a `Route` component (e.g. in the Form)

src/client/modules/Events/EventAventriAttendees/index.jsx
```diff
@@ -38,72 +50,108 @@ const EventAventriAttendees = ({
...
+        const qsParams = qs.parse(location.search.slice(1))
+
+        const initialPage = parseInt(qsParams.page, 10)
```
Harriethw, line 62: question: is there a reason we need to get an `initialPage` from the query string here? Just looking at how we do it elsewhere and we use the state2props to check the params and set the state... e.g. [for ContactActivity](https://github.com/uktrade/data-hub-frontend/blob/master/src/client/modules/Contacts/ContactActivity/state.js#L16-L17). I think this makes sense because then the component can be 'dumb' and not have to think about state management?

Harriethw: oh i see you are getting the page in state :) in that case can we replace `initialPage` with `page`?

richmadetech (approved):

src/client/modules/Events/EventAventriAttendees/index.jsx
onyela-o: Had a look, we weren't too sure how to implement it but will add a ticket to refactor later if needed :)

PR #4802, opened 2022-08-08
Update validation error messages for no recent interaction

paulgain:
## Description of change
Provide better validation error messages around the no recent interaction settings form. Specifically, a user is only allowed to enter values between 1 and 365.

## Test instructions
Go to `/reminders/settings/no-recent-interaction` and choose **Yes** to the question "Do you want to get reminders for projects with no recent interaction?" then select "Add another" and generally play with the input field by adding negative numbers, zero, 365 (the max number of days) and any old gobbledygook.

## Screenshots
Screenshot 2022-08-08 at 13 17 07

## Checklist

### DIFF
--- a/src/client/modules/Reminders/Settings/NoRecentInteractionForm.jsx
+++ b/src/client/modules/Reminders/Settings/NoRecentInteractionForm.jsx
@@ -35,11 +35,11 @@ const StyledFieldInput = styled(FieldInput)({
   width: 50,
 })
 
-const MAX_DAYS = 32767 // Set on the API endpoint (DB constraint)
+const MAX_DAYS = 365
 const POSITIVE_INT_REGEX = /^[0-9]+$/
-const EMPTY_ERR_MSG = 'Enter when you want to get reminders for your projects'
-const MIN_ERR_MSG = 'Enter a whole number that’s 1 or higher, like 25'
-const MAX_ERR_MSG = `Enter a whole number that’s less than or equal to ${MAX_DAYS}`
+const ERROR_MESSAGE = 'Enter a whole number that’s between 1 and 365, like 25'
+const EMPTY_ERROR_MESSAGE =
+  'Enter when you want to get reminders for your projects'
 
 const isPositiveInteger = (value) => POSITIVE_INT_REGEX.test(value)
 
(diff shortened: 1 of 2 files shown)


vfarah-if (approved):

cgsunkel (approved):

src/client/modules/Reminders/Settings/NoRecentInteractionForm.jsx
```diff
@@ -35,11 +35,11 @@ const StyledFieldInput = styled(FieldInput)({
   width: 50,
 })
 
-const MAX_DAYS = 32767 // Set on the API endpoint (DB constraint)
+const MAX_DAYS = 365
 const POSITIVE_INT_REGEX = /^[0-9]+$/
-const EMPTY_ERR_MSG = 'Enter when you want to get reminders for your projects'
-const MIN_ERR_MSG = 'Enter a whole number that’s 1 or higher, like 25'
-const MAX_ERR_MSG = `Enter a whole number that’s less than or equal to ${MAX_DAYS}`
+const ERROR_MESSAGE = 'Enter a whole number that’s between 1 and 365, like 25'
```
elcct, line 40: I would probably rename it to `OUT_OF_RANGE_ERROR_MESSAGE` so it is clearer what's the error is about, but given it is a small file, probably it should be okay.

elcct (approved):

src/client/modules/Reminders/Settings/NoRecentInteractionForm.jsx
paulgain: Yeah, good point. If the user enters anything that's <1 or >365 then that makes perfect sense, however, the same error message is used as a catch all so should a user type in any letter or special characters then we show the same error message. For example, if the user types in the character 't' then it's not out of bounds, it's a type error. We should be OK with what we have.

paulgain: What do you think of this, or is this overkill?

```js
const ERROR_MESSAGE = 'Enter a whole number that’s between 1 and 365, like 25'
const OUT_OF_BOUNDS_ERROR_MESSAGE = ERROR_MESSAGE
const TYPE_ERROR_MESSAGE = ERROR_MESSAGE
...
validate={(value) => {
  if (isEmpty(value)) {
    return EMPTY_ERROR_MESSAGE
  } else if (isPositiveInteger(value)) {
    const val = parseInt(value, 10)
    return val > 0 && val <= MAX_DAYS
      ? null
      : OUT_OF_BOUNDS_ERROR_MESSAGE
  } else {
    return TYPE_ERROR_MESSAGE
  }
}}

```

PR #4813, opened 2022-08-11
Refactor feature specs

Harriethw:
## Description of change

We're doing some investigation into some of the older feature specs to see if they could be refactored. 

`company-lists/list-spec` is one of our longest tests, sometimes [taking 3.5 minutes to run](https://dashboard.cypress.io/projects/w97se2/runs/30408/specs)
This PR has moved some of the logic of checking the Table renders correctly into a component test. This has the benefit of running in parallel, and means we don't need to check every cell of the table for every change we make in the functional test. 

`interactions/profiles-filter-spec` is also [long at runs of 2.36](https://dashboard.cypress.io/projects/w97se2/runs/30408/specs) 
This PR removes some of the repetitive cases in the test - as far as I can tell they didn't seem to add any value to the tests in terms of logic or contract testing. - e.g. once we have tested that a country name is sent to the backend and shown in chips etc., we know it's successfully covered the Task logic, which just sends the query to the API. That logic won't change if we test with another country, if I understand correctly.  [the code coverage hasn't decreased by removing them](https://codecov.io/github/uktrade/data-hub-frontend/commit/1a2d84f7a3d6405465a69cb67030b9fb35125b6a). Please let me know if I'm missing something!

The refactoring seems to have knocked off respectively ~3 minutes and ~1 minute for the individual specs, and just over a minute off the overall test running 🥳 

## Test instructions

_What should I see?_

### Before

_Add a screenshot_

### After

_Add a screenshot_

## Checklist

### DIFF
--- a/test/functional/cypress/specs/pipeline/my-pipeline-spec.js
+++ b/test/functional/cypress/specs/pipeline/my-pipeline-spec.js
@@ -154,27 +154,6 @@ function assertAcrossTabs(callback) {
 }
 
 describe('My pipeline app', () => {
-  context('when query string is used by analytic team', () => {
-    before(() => {
-      cy.visit(
-        `${urls.pipeline.index()}?
-        some_source=averylongquery&othersource=anotherlongquery`
-      )
-    })
-
-    it('should render the sub tab nav', () => {
-      cy.get('[data-test="pipelineSubTabNav"]').within(() => {
-        cy.contains('To do').should('have.attr', 'aria-selected', 'true')
-        cy.contains('In progress')
-        cy.contains('Done')
-      })
-    })
-
-    it('should render the item counter', () => {
-      cy.contains('4 projects')
-    })
-  })
-
   context('When viewing the to do status', () => {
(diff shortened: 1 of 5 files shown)


test/functional/cypress/specs/pipeline/my-pipeline-spec.js
```diff
@@ -154,27 +154,6 @@ function assertAcrossTabs(callback) {
 }
 
 describe('My pipeline app', () => {
-  context('when query string is used by analytic team', () => {
```
chloehkwong1, line 157: Just wondering why these tests have been removed as you didn't give a description in the PR description!

cgsunkel: From the commit description:
> This test doesn't seem to be testing any logical branches.

Harriet is on leave so I can't ask them for further details but I agree with what they've done here

cgsunkel (approved):

chloehkwong1 (approved):

PR #4831, opened 2022-08-17
[RR-506] Ensure address is only shown when country is selected

PippoRaimondiDIT:
## Description of change

This PR handles the below:

- Address fields shouldn’t be displayed before a country is selected.
- When Canada is the country selected, Province drop down’s placeholder text should be in secondary text colour (GDS grey) and it should say Select only.
- When Canada is the country selected, Postal Code needs to be updated with Postal code (optional).
- When United States is the country selected, State drop down’s placeholder text should be in secondary text colour (GDS grey) and it should say Select only.
- When United States is the country selected, ZIP Code needs to be updated with ZIP code (optional) 
- When UK is the country selected, it should be rebuilt as shown in the prototype.

## Test instructions

All test suites should pass as usual.

### Before

Screenshot 2022-08-19 at 11 45 03

### After

Screenshot 2022-08-19 at 11 44 16
Screenshot 2022-08-19 at 11 43 57

## Checklist

### DIFF
--- a/src/client/components/Form/elements/FieldAddress/index.jsx
+++ b/src/client/components/Form/elements/FieldAddress/index.jsx
@@ -220,6 +220,32 @@ const FieldAddress = ({
     if (isUK) return 'Enter a postcode'
   }
 
+  let renderPostcode
+
+  const renderUKPostcode = (
+    <StyledFieldPostcode
+      type="search"
+      name="postcode"
+      label={postcodeLabel()}
+      required={postcodeErrorMessage()}
+      maxLength={10}
+      validate={postcodeValidator}
+    />
+  )
+
+  if (country_form_value !== UNITED_KINGDOM_ID) {
+    renderPostcode = (
+      <StyledFieldPostcode
+        type="text"
+        name="postcode"
+        label={postcodeLabel()}
+        required={postcodeErrorMessage()}
+        maxLength={null}
+        validate={postcodeValidator}
+      />
+    )
+  }
+
   return (
     <FieldWrapper {...{ label, legend, hint, name }} showBorder={true}>
       {isCountrySelectable ? (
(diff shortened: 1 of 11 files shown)


src/client/components/Form/elements/FieldAddress/index.jsx
```diff
@@ -205,90 +217,122 @@ const FieldAddress = ({
...
+          <>
+            {renderUsStateField()}
+            {renderCanadaProvinceField()}
+            {administrativeAreaSearchError && (
+              <StatusMessage>
+                Error occurred while retrieving Administrative Areas.
+              </StatusMessage>
+            )}
+          </>
+          {!isUK && renderPostcode}
```
paulgain, line 333: It would be better to render everything inline for better readability:
```
{!isUK && country_form_value !== UNITED_KINGDOM_ID && (
  <StyledFieldPostcode
    type="text"
    name="postcode"
    label={postcodeLabel()}
    required={postcodeErrorMessage()}
    maxLength={null}
    validate={postcodeValidator}
  />
)}

```
This way you can easily read under what conditions the styled element is shown. It also avoids the `let renderPostcode`.

paulgain (approved):
LGTM, just a minor comment that needs addressing.

vfarah-if (approved):

src/client/components/Form/elements/FieldAddress/index.jsx
```diff
@@ -40,6 +40,18 @@ const SyledDiv = styled('div')`
   padding-bottom: ${SPACING.SCALE_5};
 `
 
+const StyledButtonWrapper = styled('div')`
+  margin-bottom: -22px;
+  margin-left: 10px;
+`
+
+const SyledRowDiv = styled('div')`
```
elcct, line 48: That seems like a typo.

src/client/components/Form/elements/FieldAddress/index.jsx
PippoRaimondiDIT: good point, by debugging a bug, I left the above as is, but it's not nice... Fixed it! thanks for the feedback!

src/client/components/Form/elements/FieldAddress/index.jsx
PippoRaimondiDIT: nice catch! fixed!

PR #4870, opened 2022-09-02
Fix security warnings

Harriethw:
## Description of change

Two commits for three different security warnings

**2. remove `originalURL` to stop possible tampering**
The use of possible user input to redirect was causing the CodeQL alert
[1][2] The redirect function doesn't actually need the original url as it
will just auto redirect to that anyway.

[1] https://codeql.github.com/codeql-query-help/javascript/js-server-side-unvalidated-url-redirection/
[2] https://github.com/uktrade/data-hub-frontend/security/code-scanning/2

**3. stop polynomial redo on email regex**
This regex would have been very slow if the user entered a huge amount
of text. One way to stop this is checking the length doesn't exceed the
max length of an email address.

## Test instructions

Under the checks -> codeQL report for this PR, it should show three alerts fixed.  https://github.com/uktrade/data-hub-frontend/pull/4870/checks?check_run_id=8159587500

## Checklist

### DIFF
--- a/src/apps/__test__/middleware.test.js
+++ b/src/apps/__test__/middleware.test.js
@@ -286,7 +286,7 @@ describe('Apps middleware', () => {
 
       expect(this.nextSpy).to.not.have.been.called
       expect(this.resMock.redirect).to.be.calledWith(
-        '/sub-app?filter=apple&sortby=sweetness'
+        '?filter=apple&sortby=sweetness'
       )
     })
 
--- a/src/apps/middleware.js
+++ b/src/apps/middleware.js
@@ -79,7 +79,7 @@ function setLocalNav(items = [], appendBaseUrl = true) {
 function setDefaultQuery(query = {}) {
   return function handleDefaultRedirect(req, res, next) {
     if (isEmpty(req.query)) {
-      return res.redirect(`${req.originalUrl}?${queryString.stringify(query)}`)
+      return res.redirect(`?${queryString.stringify(query)}`)
     }
     next()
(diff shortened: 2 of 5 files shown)


src/apps/__test__/middleware.test.js
```diff
@@ -267,47 +267,6 @@ describe('Apps middleware', () => {
     })
   })
 
-  describe('setDefaultQuery()', () => {
```
cgsunkel, line 270: There is a second Omis `CollectionList` that wasn't Reactified (viewable at the path `/omis/reconciliation?sortby=payment_due_date%3Aasc&status=quote_accepted`), possibly because it isn't accessible from within Data Hub.

I'm not sure if there are any tests around this second list so it would be good to verify that this middleware isn't being used by that list

Harriethw: oh good spot, it is being used. 
out of interest, how do users access/get directed to that link?

cgsunkel: I know it is used but I'm not sure how users access it (I believe it's shared within the OMIS team)

Harriethw: cool, I've put the code back in and solved the issue in the first commit https://github.com/uktrade/data-hub-frontend/pull/4870/commits/60e48a135eeeb05f2b3cd322df1b5073fd62d50e

cgsunkel (approved):

vfarah-if (approved):

PR #4871, opened 2022-09-02
Anonymise specific collection list filter values before sending to analytics

paulgain:
## Description of change
Currently when a user applies filters (company name, lead ITA, advisers, teams, etc) on the Collection Lists pages the filter value ends up in Google Analytics (GA). These type of filters need to be changed to the filter heading. For example, if the user applies a filter such Lead ITA and selects a name Joe Bloggs then Lead ITA should be sent to GA, not Joe Bloggs.

## Checklist

### DIFF
--- a/src/client/components/FilteredCollectionList/index.jsx
+++ b/src/client/components/FilteredCollectionList/index.jsx
@@ -30,6 +30,13 @@ const filtersToAnalytics = (filters) =>
       ])
   )
 
+const getSelectedFilters = (filters) =>
+  Object.fromEntries(
+    Object.entries(filters).filter(
+      ([, value]) => value && value.options && value.options.length
+    )
+  )
+
 const FilteredCollectionList = ({
   results = [],
   summary = null,
@@ -49,6 +56,7 @@ const FilteredCollectionList = ({
   addItemUrl,
   defaultQueryParams,
   titleRenderer = null,
+  sanitizeFiltersForAnalytics = null,
   useReactRouter = false,
 }) => {
   const totalPages = Math.ceil(
@@ -113,7 +121,13 @@ const FilteredCollectionList = ({
                                 onClick={() => {
                                   pushAnalytics({
                                     event: 'filterResultClick',
-                                    extra: filtersToAnalytics(selectedFilters),
+                                    extra: {
+                                      ...filtersToAnalytics(selectedFilters),
+                                      ...(sanitizeFiltersForAnalytics &&
+                                        sanitizeFiltersForAnalytics(
+                                          getSelectedFilters(selectedFilters)
+                                        )),
+                                    },
                                   })
                                 }}
                               />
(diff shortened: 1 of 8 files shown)


src/client/components/FilteredCollectionList/index.jsx
```diff
@@ -113,7 +121,14 @@ const FilteredCollectionList = ({
                                 onClick={() => {
                                   pushAnalytics({
                                     event: 'filterResultClick',
-                                    extra: filtersToAnalytics(selectedFilters),
+                                    extra: {
+                                      ...filtersToAnalytics(selectedFilters),
```
Harriethw, line 125: is this call to `filtersToAnalytics` still sending the filter data to GA? or is it doing something different?

src/client/modules/Companies/CollectionList/index.jsx
```diff
@@ -91,6 +93,17 @@ const CompaniesCollection = ({
         entityName="company"
         entityNamePlural="companies"
         addItemUrl="/companies/create"
+        sanitizeFiltersForAnalytics={({
+          name,
+          ukPostcode,
+          leadItaOrGlobalAccountManagers,
+        }) => ({
+          ...(name ? sanitizeFilter(name) : {}),
```
Harriethw, line 101: could these ternary operations use the `&&` operator to avoid spreading an empty object? 
I'm also wondering if the `sanitizeFilter` function could check whether the filter exists?

Harriethw:
is there any test coverage for what we're pushing to the data layer? might be good to have some, if only to make it easier to understand what's going between all the different functions :)

src/client/components/FilteredCollectionList/index.jsx
paulgain: No, it takes the filters and translates them so they can be read by GA, take a [look](https://github.com/uktrade/data-hub-frontend/blob/87758ded0fec9f2846c6eafa7bd34bcb3726bf3b/src/client/components/FilteredCollectionList/index.jsx#L23) for your self.

src/client/modules/Companies/CollectionList/index.jsx
paulgain: _"I'm also wondering if the sanitizeFilter function could check whether the filter exists?"_  The filter is guaranteed to exist because that's what [this](https://github.com/uktrade/data-hub-frontend/blob/87758ded0fec9f2846c6eafa7bd34bcb3726bf3b/src/client/components/FilteredCollectionList/index.jsx#L33) function does - it gets all the selected filters, these filters are then passed to the `sanitizeFiltersForAnalytics(selectedFilters)` function so there's no need for an explicit check.

paulgain:
> is there any test coverage for what we're pushing to the data layer? might be good to have some, if only to make it easier to understand what's going between all the different functions :)

We don't test what's being written to the data layer other than looking in GA/GTM to ensure what's being written comes out the other end.

src/client/modules/Companies/CollectionList/index.jsx
paulgain: _"could these ternary operations use the && operator to avoid spreading an empty object?"_
Good idea 👍

marijnkampf (approved):

cgsunkel (approved):

PR #4895, opened 2022-09-12
Expand settings depending on the reminder url you come from

PippoRaimondiDIT:
## Description of change

The intent of this PR is to ensure the according settings is expanded depending on the URL you come from.

## Test instructions

Included functional tests, all functional tests should pass.

### Before

Screenshot 2022-09-12 at 09 06 04

### After

Screenshot 2022-09-12 at 09 06 16

## Checklist

### DIFF
--- a/src/client/modules/Reminders/Settings/index.jsx
+++ b/src/client/modules/Reminders/Settings/index.jsx
@@ -45,8 +45,17 @@ const StyledHomeLink = styled(Link)({
 const RemindersSettings = () => {
   const location = useLocation()
   const qsParams = qs.parse(location.search.slice(1))
-  const openESL = get(qsParams, 'estimated_land_date', false)
-  const openNRI = get(qsParams, 'no_recent_interaction', false)
+  const previousURL = document.referrer
+  const estimatedLandDateExpand =
+    previousURL && previousURL.includes('estimated-land-date')
+  const noRecentInteractionExpand =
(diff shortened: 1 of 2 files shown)


cgsunkel (approved):

src/client/modules/Reminders/Settings/index.jsx
```diff
@@ -42,11 +42,28 @@ const StyledHomeLink = styled(Link)({
   display: 'block',
 })
 
+const openSettings = (previousUrlType, queryParamType, qsParams) => {
+  const previousURL = document.referrer
```
paulgain, line 46: We shouldn't be accessing `window.document` directly as it violates the architecture we have in place. We prefer to user a Redux Saga instead. You may find this helpful: https://github.com/uktrade/data-hub-frontend/blob/main/src/client/components/Analytics/saga.js#L27

PippoRaimondiDIT: ok, thanks for the review... I will try to figure out how to use saga to accomplish the above...

PippoRaimondiDIT: I've matched the way we identify settings coming from email to simplify the logic... we are no longer using window object now..

paulgain: Nice, a big improvement.

PR #4982, opened 2022-10-17
Remove pipeline tests for 404 error

cgsunkel:
## Description of change

I would like to remove the 404 error tests from the pipeline specs for the following reasons:
- We do not test for incorrect IDs in any other area of the codebase. There is nothing special about the pipelines page that requires this to be tested.
- They are repeatedly failing in `main` and in other branches. This is causing us to spend time re-running the functional suite when we shouldn't need to.
- The tests [don't refelect reality](https://www.datahub.dev.uktrade.io/my-pipeline/INCORRECT_PIPELINE/archive) since the new error handling was introduced.

Based on the above points I believe we can remove these tests.

## Checklist

### DIFF
--- a/test/functional/cypress/specs/pipeline/my-pipeline-archive-spec.js
+++ b/test/functional/cypress/specs/pipeline/my-pipeline-archive-spec.js
@@ -48,17 +48,6 @@ function assertForm() {
 }
 
 describe('Archive pipeline item form', () => {
-  context('When the pipeline item id is incorrect', () => {
-    before(() => {
-      cy.visit(urls.pipeline.archive('INCORRECT-PIPELINE-ID'))
-    })
-
-    it('should render a 404 error message', () => {
-      cy.contains('Could not load PipelineItem')
-      cy.contains('Error: Not Found')
-    })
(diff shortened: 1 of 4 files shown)


marijnkampf:
Are these basically flaky tests you want to remove or do they fail consistently on certain environments?

paulgain (approved):

cgsunkel:
> Are these basically flaky tests you want to remove or do they fail consistently on certain environments?

They are extremely flaky tests (they fail constantly in all environments I've tried them in)

marijnkampf (approved):

paulgain:
@cgsunkel a few observations: 

Generic error handling (running the app locally).
error

1. The technical error messaging is fine as it's not in the production environment. 
2. Does it make sense to revisit this page with a designer to improve the page as it doesn't really fit into Data Hub.
3. Peter, has error handling built into the Task components, we also have error handling on the Node/Express layer. Vince has introduced generic error handling. I'm concerned we're handling errors in different parts of the app and we should have a single mechanism to deal with error handling not 3.

PR #4987, opened 2022-10-18
Add saga for dismissing banner

PippoRaimondiDIT:
## Description of change

The intent of this PR is the initial work around adding dismissible banners feature. To start with we would like to have the ability to read/write a key/value from localstorage.

## Test instructions

_What should I see?_

### Before

_Add a screenshot_

### After

_Add a screenshot_

## Checklist


src/client/actions.js
```diff
@@ -214,3 +214,10 @@ export const FIELD_ADD_ANOTHER__ADD = 'FIELD_ADD_ANOTHER__ADD'
 export const FIELD_ADD_ANOTHER__REMOVE = 'FIELD_ADD_ANOTHER__REMOVE'
 
 export const DATA_HUB_FEED__FETCHED = 'DATA_HUB_FEED__FETCHED'
+
+export const IS_BANNER_DISMISSED__SAVE_TO_LOCALSTORAGE =
+  'IS_BANNER_DISMISSED__SAVE_TO_LOCALSTORAGE'
+export const IS_BANNER_DISMISSED__GET_FROM_LOCALSTORAGE =
+  'IS_BANNER_DISMISSED__GET_FROM_LOCALSTORAGE'
+export const IS_BANNER_DISMISSED__ADD_TO_STATE =
```
paulgain, line 222: Ideally we don't want to name constants starting with `IS` as `is` is normally associated with booleans. I would recommend the following:
 ```js
    BANNER_DISMISSED__WRITE_TO_LOCALSTORAGE
    BANNER_DISMISSED__READ_FROM_LOCALSTORAGE
    BANNER_DISMISSED__UPDATE_STATE
```
I know there is a `*__ADD_TO_STATE` already defined, however, we're really updating the state.

File: src/client/components/LocalHeader/sagas.js [hunk 2]
```diff
@@ -43,3 +46,25 @@ export function* readFlashMesages() {
...
+    )
+    saveToLocalStorage(isBannerDismissed)
+  }
+}
+
+/* Saga to read localstorage state and add to redux state. */
+export function* readIsBannerDismissedFromLocalStorage() {
+  while (true) {
+    yield take(IS_BANNER_DISMISSED__GET_FROM_LOCALSTORAGE)
+    const isBannerDismissedState = getFromLocalStorage()
```
paulgain, line 64: I think `isBannerDismissed` is more accurate.

src/client/components/LocalHeader/utils.js
```diff
@@ -0,0 +1,11 @@
...
+  window.localStorage.setItem('isBannerDismissed', isBannerDismissed)
+}
+
+export const getFromLocalStorage = () => {
+  try {
+    return window.localStorage.getItem('isBannerDismissed') || true
+  } catch {
+    return {}
+  }
+}
```
paulgain, line 11: It's worth noting that writing a boolean to `localStorage` will return as a string when reading the same key:
```js
window.localStorage.setItem('isBannerDismissed', true)
// Returns the string 'true' not the boolean that was originally set.
window.localStorage.getItem('isBannerDismissed')
```

I would use this instead as this does the conversion for you.
```js
window.localStorage.getItem('isBannerDismissed') === 'true' // true
```

paulgain: I would also use a constant instead of two duplicate strings:
```js
export const BANNER_DISMISSED = 'bannerDismissed'

export const writeToLocalStorage = (isBannerDismissed = true) => {
  window.localStorage.setItem(BANNER_DISMISSED, isBannerDismissed)
}

export const readFromLocalStorage = () => {
  try {
    return window.localStorage.getItem(BANNER_DISMISSED) === 'true'
  } catch {
    return {}
  }
}

```

paulgain (commented):
Looks good, just a few comments.

src/client/actions.js
PippoRaimondiDIT: sorted!

src/client/components/LocalHeader/sagas.js
PippoRaimondiDIT: sorted!

paulgain (approved):

cgsunkel (approved):

PR #5121, opened 2022-12-07
Use correct company URL for company interactions

cgsunkel:
## Description of change

The URL for interactions in a company activity feed is currently the direct URL for the interaction (which was being taken directly from ActivityStream), whereas it should be the using the URL from the interaction sub-app within companies.

## Test instructions

Go to a company with interactions and click one from the activity feed. The breadcrumbs on the interaction page should display `Companies > Interaction` rather than `Interactions > Interaction`.

### Before

Screenshot 2022-12-05 at 14 45 12

### After

Screenshot 2022-12-05 at 14 45 05

## Checklist

### DIFF
--- a/src/client/components/ActivityFeed/activities/InteractionUtils.js
+++ b/src/client/components/ActivityFeed/activities/InteractionUtils.js
@@ -1,5 +1,6 @@
 import { get, includes } from 'lodash'
 import { STATUS, BADGES, INTERACTION_SERVICES } from '../constants'
+import urls from '../../../../lib/urls'
 
 const getStatus = (activity) => {
   const apiStatus = get(activity, 'object.dit:status')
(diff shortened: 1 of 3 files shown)


src/client/components/ActivityFeed/activities/InteractionUtils.js
```diff
@@ -22,6 +23,12 @@ const isServiceDelivery = (activity) => {
   return includes(activityTypes, 'dit:ServiceDelivery')
 }
 
+const getCompanyInteractionUrl = (activity) => {
```
chopkinsmade, line 26: Is the `attributedTo` property always returned as an array with at least 1 item? We could add some checks here to avoid throwing any UI errors if this value doesn't come back exactly as expected from the API

chopkinsmade (approved):

src/client/components/ActivityFeed/activities/InteractionUtils.js
cgsunkel: Yes we always get this from ActivityStream in this format

rachaelcodes (approved):

PR #5214, opened 2023-01-18
Improve app performance by removing duplicate API calls

paulgain:
## Description of change
We're starting to make multiple API calls to the same endpoint on the same page, specifically the `/v4/reminder/summary` endpoint. I've refactored parts of the code so that the ubiquitous `NotificationAlert` component is the only component to make a call to that endpoint and saves the results in the Redux store. Any connected components that require reminder summary data such as the personalised dashboard (i.e. the notification count), the reminder summary section and the reminder notifications menu can access the store instead of making the same API call. 

**The page load time saved in removing a duplicate API call is negligible (small amount of data transmitted over the wire), however, it's best to make these changes now rather than later in case more API calls to the same endpoint are unknowingly added.**

API call:
```code
curl --location --request GET 'https://<environment>/v4/reminder/summary' \
--header 'Authorization: Bearer <your-token>'
```

Response:
```json
{
    "count": 2,
    "investment": {
        "estimated_land_date": 0,
        "no_recent_interaction": 1,
        "outstanding_propositions": 1
    },
    "export": {
        "no_recent_interaction": 0,
        "new_interaction": 0
    }
}
```
redux-store

Screenshot 2023-01-19 at 11 56 19

## Test instructions
When viewing the Personalised Dashboard or the Reminders notification menu there should only be a single API call to the `/v4/reminder/summary` endpoint.

## Checklist


src/client/components/NotificationAlert/index.jsx
```diff
@@ -40,11 +40,11 @@ const NotificationAlert = ({ count, remindersURL }) => (
       alt="An image of a bell with the notification count overlaid"
     />
     <Task.Status
-      name={TASK_GET_NOTIFICATION_ALERT_COUNT}
+      name={TASK_GET_REMINDER_SUMMARY}
```
chopkinsmade, line 43: PR looks good, and if the answer to following question is no just resolve this comment, but do we have any way of triggering a Task on app start? Instead of waiting for the notification component to render, which in turn triggers the `TASK_GET_REMINDER_SUMMARY` task, we could trigger the task as soon as react is initialised

chopkinsmade (approved):

src/client/components/NotificationAlert/index.jsx
paulgain: It's a good point, at present we don't. Take a look at the [Task](https://github.com/uktrade/data-hub-frontend/blob/main/src/client/components/Task/index.jsx) component and you will see we only start the task on render. Going forward, we need to make a number of API calls upfront, one would be to get the adviser, another would be to get feature flags/groups another would be the notifications. As we slowly migrate over to a SPA this will make it easier - but we're not there yet.

elcct (approved):

PR #5215, opened 2023-01-18
Update ActivityCardLabels.jsx

bau123:
Add labels to datahub events

## Description of change

Adds labels to datahub events

## Test instructions

Labels on datahub events

### Before
Screenshot 2023-01-18 at 11 15 30

### After
Screenshot 2023-01-25 at 11 43 49

## Checklist


onyela-o (commented):
Hello!
So I think we have shortened versions of the theme and service name for the labels instead of the full version e.g. A specific DIT export service or funding becomes something like 'Export service and funding' - I can't remember where the list is though :( 

EDIT: Found them I think src/client/components/ActivityFeed/constants.js and src/client/components/ActivityFeed/activities/InteractionUtils.js might be useful

PaulWheatcroft (approved):

src/client/components/ActivityFeed/activities/DataHubEvent.jsx
```diff
@@ -21,10 +22,18 @@ export default function DataHubEvent({ activity: event }) {
...
+  const serviceTypeSplit = serviceType.split(':')
+  const theme = serviceTypeSplit[0]
+  const service = serviceTypeSplit[1]
   return (
     <ActivityCardWrapper dataTest="data-hub-event">
       <ActivityCardSubject dataTest="data-hub-event-name">
+        <ActivityCardLabels
+          theme={theme}
+          service={service}
+          kind={typeOfEvent}
```
onyela-o, line 35: If the `typeOfEvent` is null/empty - I think maybe the behaviour we would want would be for the kind label not to show at all, rather than say 'Not set'

src/client/components/ActivityFeed/activities/DataHubEvent.jsx
```diff
@@ -21,10 +22,18 @@ export default function DataHubEvent({ activity: event }) {
   const organiser = eventObject['dit:organiser']?.name || 'Not set'
   const serviceType = eventObject['dit:service']?.name || 'Not set'
   const leadTeam = eventObject['dit:leadTeam']?.name || 'Not set'
-
+  const typeOfEvent = eventObject['dit:eventType']?.name || 'Not set'
+  const serviceTypeSplit = serviceType.split(':')
```
onyela-o, line 26: This line might fail if the serviceType is set to 'Not set' as it will try and split it on a ':' and it doesn't exist in that case

src/client/components/ActivityFeed/activities/DataHubEvent.jsx
```diff
@@ -21,10 +22,18 @@ export default function DataHubEvent({ activity: event }) {
...
+  const theme = serviceTypeSplit[0]
```
onyela-o, line 27: Could do destructuring here instead so it's a bit neater like:
`const [ theme, service ] = serviceType.split(':')`
:)

onyela-o (changes requested):

src/client/components/ActivityFeed/activities/DataHubEvent.jsx
bau123: Added

src/client/components/ActivityFeed/activities/DataHubEvent.jsx
bau123: Added

src/client/components/ActivityFeed/activities/DataHubEvent.jsx
bau123: All service should have a . so safe to split

bau123 (commented):
Changes added

src/client/components/ActivityFeed/activities/DataHubEvent.jsx
```diff
@@ -21,10 +23,19 @@ export default function DataHubEvent({ activity: event }) {
   const organiser = eventObject['dit:organiser']?.name || 'Not set'
   const serviceType = eventObject['dit:service']?.name || 'Not set'
   const leadTeam = eventObject['dit:leadTeam']?.name || 'Not set'
+  const typeOfEvent = eventObject['dit:eventType']?.name || undefined
+  const [, service2] = serviceType.split(' : ')
```
onyela-o, line 27: Maybe rename this variable - I think we try and avoid naming something with a number

src/client/components/ActivityFeed/activities/DataHubEvent.jsx
```diff
@@ -21,10 +23,19 @@ export default function DataHubEvent({ activity: event }) {
...
+  const serviceText = getServiceText(serviceType)
+  const serviceOther = getServiceOtherText(service2)
```
onyela-o, line 29: Not sure about the serviceOther name - is it additional/extra text? Maybe serviceAdditionalText or something might be clearer

src/client/components/ActivityFeed/constants.js
```diff
@@ -105,6 +105,53 @@ export const INTERACTION_SERVICES = {
   'Trade Agreement Implementation Activity': 'Implementation',
 }
 
+export const INTERACTION_SERVICEOTHER = {
```
onyela-o, line 108: INTERACTION_SERVICE_OTHER (or whatever the name will be) to keep the casing consistent :)

onyela-o (approved):
A few comments about naming but I'll approve so it doesn't block you again

PR #5238, opened 2023-01-23
RR-831-refactor-reminder-urls

chopkinsmade:
## Description of change

- Move the logic for determining the default reminder link logic from notification bell component to the reminder component
- If no reminder type is selected, redirect to the default based on user feature flags

## Test instructions

This logic requires the use of `export-notifications` and the `investment-notifications` feature groups. 
Disabling `investment-notifications` and navigating to http://localhost:3000/reminders should redirect to the first export reminders link.
Enabling `investment-notifications` and navigating to http://localhost:3000/reminders should redirect to the first investment reminders link.

### Before

chrome_kuLQDnmHF8

### After

chrome_tIj7TlwJtt

## Checklist

### DIFF
--- a/src/client/modules/Reminders/RemindersLists.jsx
+++ b/src/client/modules/Reminders/RemindersLists.jsx
@@ -1,10 +1,8 @@
 import React from 'react'
-import { Redirect, useParams } from 'react-router-dom'
 import { SPACING, MEDIA_QUERIES } from '@govuk-react/constants'
 import styled from 'styled-components'
 import { Link } from 'govuk-react'
 
-import { connect } from 'react-redux'
 import { DefaultLayout } from '../../components'
 import RemindersMenu from './RemindersMenu'
 
@@ -23,7 +21,6 @@ import {
   COMPANIES_NO_RECENT_INTERACTIONS,
   COMPANIES_NEW_INTERACTIONS,
 } from './constants'
-import { state2props } from './state'
 
 const Container = styled('div')({
   [MEDIA_QUERIES.DESKTOP]: {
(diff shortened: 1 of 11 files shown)


src/client/modules/Reminders/RemindersLists.jsx
```diff
@@ -66,8 +68,12 @@ const HomeLink = styled(Link)({
   },
 })
 
-const RemindersLists = () => {
+export const RemindersLists = ({ defaultUrl }) => {
   const { reminderType } = useParams()
+  if (!reminderType) {
+    return <Redirect to={{ pathname: defaultUrl }} />
+  }
+
```
paulgain, line 76: All though we need a redirect it shouldn't be the lists responsibility. Is it possible to move this check up a level to the routing level?

PaulWheatcroft (approved):

src/client/modules/Reminders/RemindersLists.jsx
chopkinsmade: I've moved the logic up a level, so the redirect is now handled in the <RemindersRoutes /> icon

src/client/modules/Reminders/RemindersRoutes.jsx
```diff
@@ -1,22 +1,16 @@
...
-      </Route>
-      <Route path={`${path}/settings/:reminderType`}>
-        <RemindersForms />
-      </Route>
-      <Route path={`${path}/:reminderType`}>
-        <RemindersLists />
-      </Route>
-    </Switch>
-  )
+export const RemindersRoutes = ({ defaultUrl }) => {
```
paulgain, line 7: As the routes have now been removed from here does it make sense to still call it `RemindersRoutes`?

chopkinsmade: I had to leave some routing for the `RemindersLists` in here to work with the url params and redux state. Could rename it to `RemindersListRouting` / `RemindersListsRoutes`?

paulgain: Maybe just `Reminders`? as the actual `Route` no longer exists.

paulgain (approved):

PR #5274, opened 2023-02-10
role added to alert to resolve accessibility issue

PaulWheatcroft:
## Description of change

Status messages do not get read out for screen reader users upon first loading the page. This is due to not having a role="alert" attribute placed upon the element and instead having aria-live="polite". This change removes aria-live="polite" and replaces it with ole="alert" for all alerts.

## Test instructions

N/A

## Screenshots
No change
## Checklist

### DIFF
--- a/src/client/components/StatusMessage/index.jsx
+++ b/src/client/components/StatusMessage/index.jsx
@@ -21,8 +21,8 @@ StatusMessage.propTypes = {
 
 StatusMessage.defaultProps = {
   colour: BLUE,
+  role: 'alert',
   'data-test': 'status-message',
-  'aria-live': 'polite',
 }
 
 export default StatusMessage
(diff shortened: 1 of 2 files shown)


richmadetech (approved):

src/client/components/StatusMessage/index.jsx
```diff
@@ -21,8 +21,8 @@ StatusMessage.propTypes = {
 
 StatusMessage.defaultProps = {
   colour: BLUE,
+  role: 'alert',
   'data-test': 'status-message',
-  'aria-live': 'polite',
```
chopkinsmade, line 25: Does the `aria-live` need to be there to meet any other accessibility requirements? Can both `aria-live` and `role` be present or do they conflict?

chopkinsmade: Found the answer to my question, yes they can conflict so you're right removing it https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions#roles_with_implicit_live_region_attributes

chopkinsmade (approved):

PR #5287, opened 2023-02-15
Link an ARIA attribute to an ID on the page

paulgain:
## Description of change
**Unlinked ARIA attribute**
An `aria-labelledby` attribute was found which is not linked to any label on the page.

**Solution**
Ensure that the `aria-labelledby` attributes refer to an ID that exists on the same page

### Before
before

### After
after

## Checklist

### DIFF
--- a/src/client/components/TabNav/index.jsx
+++ b/src/client/components/TabNav/index.jsx
@@ -236,7 +236,7 @@ const TabNav = ({
             <StyledTabpanel
               role="tabpanel"
               tabIndex={-1}
-              aria-labelledby={createId(id, selectedIndex)}
+              aria-labelledby={createId(id, selectedIndex, true)}
               data-test="tabpanel"
             >
               {getContent(tabs, tabKeys, selectedIndex)}


cgsunkel (approved):

dredmonds (approved):

src/client/components/TabNav/index.jsx
```diff
@@ -236,7 +236,7 @@ const TabNav = ({
             <StyledTabpanel
               role="tabpanel"
               tabIndex={-1}
-              aria-labelledby={createId(id, selectedIndex)}
+              aria-labelledby={createId(id, selectedIndex, true)}
```
chopkinsmade, line 239: There was this PR https://github.com/uktrade/data-hub-frontend/pull/5289 that removed the `aria-labelledby` and replaced with an `aria-label`. Not sure if there is any overlap here in the accessibility report?

chopkinsmade (approved):

src/client/components/TabNav/index.jsx
paulgain: Thanks for pointing that out. Looking at both issues I think they are unrelated. 

This PR relates to a tab component, where each tab has a child component. When a user selects a tab the child component is shown. The `aria-labelledby` attribute allows us to specify an element ID to refer to another element in the DOM as this element’s label, in this particular case, the children are taking the parents label.

The other PR refers to a button (AKA a Chip) which relates to filtering, here, there's no need for the `aria-labelledby` attribute as per the report.

PR #5357, opened 2023-03-08
RR-849-add-form-scaffolding

chopkinsmade:
## Description of change

- Added a new page with an empty form to add new export pipeline items, along with the routing needed to access the page
- The same form will be used by both the add and edit pages, so a separate ExportFormFields.jsx component has been created ready

## Test instructions

Use the url `http://localhost:3000/export/create?companyId=c1ea5ee7-c6f5-4dd1-8c5e-785c5a115719` to load the new page (you can use any valid company id). There should be a breadcrumb with the company name, and the same company name as a sub heading.
The form will be empty, but clicking cancel will redirect to the company page that is provided in the `?companyId` query string

To test the error screen, use the url `http://localhost:3000/export/create?companyId=not_real`

### Valid company:
image

### Invalid company
image

## Checklist

### DIFF
--- a/src/client/modules/ExportPipeline/ExportForm/ExportFormAdd.jsx
+++ b/src/client/modules/ExportPipeline/ExportForm/ExportFormAdd.jsx
@@ -1,0 +1,98 @@
+import React from 'react'
+import { connect } from 'react-redux'
+import { useLocation } from 'react-router-dom'
+import { ERROR_COLOUR } from 'govuk-colours'
+import { H3 } from '@govuk-react/heading'
+
+import Task from '../../../components/Task'
+import urls from '../../../../lib/urls'
+import { COMPANY_LOADED } from '../../../actions'
+import { DefaultLayout } from '../../../components'
+import { TASK_GET_COMPANY_DETAIL } from '../../Companies/CompanyDetails/state'
+import { state2props } from './state'
+import { ID as COMPANY_DETAILS_ID } from '../../Companies/CompanyDetails/state'
+import { StatusMessage } from '../../../../client/components/'
+import ExportFormFields from './ExportFormFields'
+
+const DISPLAY_ADD_EXPORT = 'Add export'
+
+function useQuery() {
+  const { search } = useLocation()
+
+  return React.useMemo(() => new URLSearchParams(search), [search])
+}
+
+function ErrorHandler() {
+  return (
+    <StatusMessage
+      colour={ERROR_COLOUR}
+      aria-labelledby="company-load-error-summary-title"
+      role="alert"
+      data-test="company-load-error"
+    >
+      <H3 id="company-load-error-summary-title">Error loading company</H3>
+    </StatusMessage>
+  )
+}
+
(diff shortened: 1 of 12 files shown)


src/client/modules/ExportPipeline/ExportForm/ExportFormAdd.jsx
```diff
@@ -0,0 +1,98 @@
...
+    >
+      <Task.Status
+        name={TASK_GET_COMPANY_DETAIL}
+        id={COMPANY_DETAILS_ID}
+        progressMessage="Loading company details"
+        startOnRender={{
+          payload: companyId,
+          onSuccessDispatch: COMPANY_LOADED,
+        }}
+        renderError={ErrorHandler}
```
paulgain, line 85: There's no need to define this on a Task, the component has it baked in.

src/client/modules/ExportPipeline/ExportForm/ExportFormFields.jsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react'
+
+import Form from '../../../components/Form'
+import { FormLayout } from '../../../../client/components'
+import { FORM_LAYOUT } from '../../../../common/constants'
+import urls from '../../../../lib/urls'
+import { TASK_SAVE_EXPORT } from './state'
+
+const ExportFormFields = ({ companyId, analyticsFormName }) => {
+  return (
```
paulgain, line 10: Favour implicit returns over explicit returns:
```js
const ExportFormFields = ({ companyId, analyticsFormName }) => (<>...</>)
```

src/client/modules/ExportPipeline/ExportForm/ExportFormAdd.jsx
```diff
@@ -0,0 +1,98 @@
...
+      text: 'Home',
+    },
+    {
+      link: urls.companies.index(),
+      text: 'Companies',
+    },
+    { text: 'Add export' },
+  ]
+
+  if (Object.keys(company).length > 0) {
```
paulgain, line 51: Can you not just test for the company?
```js
if (company) { ... }
```

paulgain (commented):
Looks good, although the breadcrumbs section looks verbose.

src/client/modules/ExportPipeline/ExportForm/ExportFormAdd.jsx
paulgain: That's the problem, it shouldn't be defaulting to an empty object - see my comment below.

paulgain (approved):

src/client/modules/ExportPipeline/ExportForm/ExportFormAdd.jsx
chopkinsmade: I'll merge this in with a custom error handler, as there is a wider conversation to be had about how we handle errors from the API when we aren't going via an express route. Will look at some refactoring for the edit version of this form, as that will also trigger a direct API call avoiding express

PR #5371, opened 2023-03-13
Add sorting logic into opensearch query and remove workaround from co…

simon-mt:
## Description of change

The activity feed opensearch queries did not sort in the correct order when viewing a company - some activities did not (and should not) have a startTime, yet the startTime was being used as the sort criteria. This caused unpredictable, and changeable, behaviour in the order of activities displayed.

By changing the opensearch sorting criteria, a second, unnecessary sort operation in the front end controller could be removed.

## Test instructions

The activities should always be in date descending order and be consistently displayed in that order.

### Before

image

### After

image


src/apps/companies/apps/activity-feed/es-queries/data-hub-and-activity-stream-services-query.js
```diff
@@ -1,3 +1,19 @@
+const sortCriteria = {
```
marijnkampf, line 1: Would it make sense to refactor this into a generic method that is used across the different queries?

simon-mt: @marijnkampf Pushed change that addresses this. I've deliberately left the tests to not use the shared function because if a breaking change is made to the sortCriteria the tests would still pass.

dredmonds (approved):

src/apps/companies/apps/activity-feed/es-queries/data-hub-and-activity-stream-services-query.js
simon-mt: Resolving as PR now approved.

PR #5398, opened 2023-03-20
Feature/rr 791 export owner

chopkinsmade:
## Description of change

- Added current advisor name to the global props, as this is needed along with the current advisor id for the owner dropdown
- Added the new owner field to the export item form
- Added logic to transform api responses into values needed by the form field components `{id, name} => {value, label}`
- Added logic to transform form field values back into values needed by the api `{value, label} => {id, name}`
- Where possible in cypress tests, switch from `beforeEach()` to `before()` for visiting pages to improve performance

### Adding

image

### Editing

image

## Checklist

### DIFF
--- a/src/client/modules/ExportPipeline/transformers.js
+++ b/src/client/modules/ExportPipeline/transformers.js
@@ -1,14 +1,14 @@
 const mapApiToField = ({ id, name }) => ({ value: id, label: name })
 
-export const transformFormValuesForAPI = ({ id, title, owner }) => ({
+export const transformFormValuesForAPI = ({ company, id, title, owner }) => ({
+  company,
   id,
   title,
-  owner: {
-    id: owner.value,
-  },
+  owner: owner.value,
 })
 
-export const transformAPIValuesForForm = ({ id, title, owner }) => ({
+export const transformAPIValuesForForm = ({ company, id, title, owner }) => ({
+  company: company.id,
   id,
   title,
   owner: mapApiToField(owner),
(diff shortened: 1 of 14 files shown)


test/functional/cypress/support/assertions.js
```diff
@@ -345,7 +346,11 @@ const assertFieldInput = ({
           .should('have.text', hint || '')
           .next()
     )
-
+    .then(($el) => {
+      //in the scenario where we don't need to validate what the hint is, but a hint is still
+      //being rendered, skip over the hint without validating it to get to the next element
+      return ignoreHint && value ? cy.wrap($el).next() : undefined
```
chopkinsmade, line 352: Investigate why single line `=>` without a `return` statement doesn't seem to work, possibly something related to Cypress chaining

src/client/modules/ExportPipeline/transformers.js
```diff
@@ -0,0 +1,20 @@
...
+export const transformAPIValuesForForm = (initialValues) => {
+  if (!initialValues) {
+    return initialValues
+  }
+
+  return {
+    ...initialValues,
+    owner: mapApiToField(initialValues.owner),
+  }
+}
```
paulgain, line 20: It would be great to improve the readability of the code above as it's unclear as to what is being translated, something like the following:

```js
const mapApiToField = ({ id, name }) => ({ value: id, label: name })

export const transformFormValuesForAPI = ({ id, title, owner }) => ({
  id,
  title,
  owner: {
    id: owner.value,
  },
})

export const transformAPIValuesForForm = ({ company, owner, title }) => ({
  ...company,
  title,
  owner: mapApiToField(owner),
})
```

And then in ExportFormFields.jsx
```js
initialValues={
  initialValues && transformAPIValuesForForm(initialValues)
}
```

Then you can do away with all the checks

paulgain (commented):
Looks good, just a comment around the transforming of data.

santoshdasa12345 (approved):

PR #5532, opened 2023-04-20
Feature/rr 951 add e2e test for export pipeline

chopkinsmade:
## Description of change

Add a new e2e test that performs the following steps and assertions:

- Adds a new export pipeline item
- Check the new export pipeline item appears in the list of export pipeline items
- Edit the new export pipeline and change some field values
- Delete the new export pipeline item
- Check the deleted item no longer appears in the list of export pipeline items

## Checklist

### DIFF
--- a/test/end-to-end/cypress/specs/DIT/export-pipeline-spec.js
+++ b/test/end-to-end/cypress/specs/DIT/export-pipeline-spec.js
@@ -1,0 +1,81 @@
+const fixtures = require('../../fixtures')
+const { exportFaker } = require('../../../../functional/cypress/fakers/export')
+const { exportPipeline } = require('../../../../../src/lib/urls')
+const {
+  fill,
+  fillTypeahead,
+  fillSelect,
+} = require('../../../../functional/cypress/support/form-fillers')
+
+describe('Export pipeline', () => {
+  const company = fixtures.company.create.defaultCompany(
+    'export pipeline testing'
+  )
+  const contact = fixtures.contact.create(company.pk)
+  const newExport = exportFaker({ destination_country: { name: 'Greece' } })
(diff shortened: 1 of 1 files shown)


test/end-to-end/cypress/specs/DIT/export-pipeline-spec.js
```diff
@@ -0,0 +1,92 @@
...
+      cy.loadFixture([company])
+      cy.loadFixture([contact])
+    })
+
+    beforeEach(() => {
+      cy.intercept('POST', '/api-proxy/v4/export').as('createExport')
+      cy.intercept(
+        'GET',
+        '/api-proxy/v4/export?limit=10&page=1&offset=0&archived=false'
+      ).as('listExport')
```
paulgain, line 29: If this is an e2e test then we're calling the API directly, we shouldn't be intercepting any API calls. We only do that in the functional tests.

chopkinsmade: I'm not mocking any responses with the `cy.intercept` here, it's just spying on the request without applying any modifications so I can wait for the api call to complete

paulgain: Are you sure we actually need it? None of our other e2e test use `cy.intercept`?

paulgain (approved):

test/end-to-end/cypress/specs/DIT/export-pipeline-spec.js
chopkinsmade: I needed to grab the id from the api of the created export, so the spying helped there as I could inspect the response the get that new id.

PR #5776, opened 2023-06-23
Add an interaction kind to investment and trade agreement

paulgain:
## Description of change
When a user creates an interaction and they choose a `theme` (Interaction or Trade agreement) the API expects a `kind` to be set as "interaction". Previously, this was done in the `onChange` event handlers, but the value set was quickly removed in subsequent form updates. Now we perform a check just before saving the form, and if applicable, set the `kind` there.

## Checklist

### DIFF
--- a/src/apps/interactions/apps/details-form/client/StepInteractionType.jsx
+++ b/src/apps/interactions/apps/details-form/client/StepInteractionType.jsx
@@ -20,7 +20,6 @@ const getKindHandler = (setFieldValue) => (e) => {
 const getNoKindHandler = (setFieldValue) => (e) => {
   setFieldValue('service', '')
   setFieldValue('theme', e.target.value)
-  setFieldValue('kind', KINDS.INTERACTION)
 }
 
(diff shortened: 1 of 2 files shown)


src/apps/interactions/apps/details-form/client/StepInteractionType.jsx
```diff
@@ -20,7 +20,6 @@ const getKindHandler = (setFieldValue) => (e) => {
 const getNoKindHandler = (setFieldValue) => (e) => {
   setFieldValue('service', '')
   setFieldValue('theme', e.target.value)
-  setFieldValue('kind', KINDS.INTERACTION)
```
paulgain, line 23: Removing this line as it doesn't work as you would expect. Initially the `kind` field is set within the form values object, however, after subsequent form updates the value is quickly overwritten (actually the key is removed).

dredmonds (approved):

cgsunkel (approved):

PR #5801, opened 2023-07-03
Read/write the personalised dashboard tab selection from/to localstorage

paulgain:
## Description of change
We want to store the current personalised dashboard tab selection into `window.localstorage` so the user can continue working where they left off.

## Test instructions
_Preconditions: you need the personalised dashboard group feature flag._

Go to the personalised dashboard where the investment projects tab is selected by default

1. Select any of the other three tabs (say exports)
2. Kill the browser
3. Go to the personalised dashboard and the exports tab is now selected

## Checklist


src/client/utils/localStorage.js
```diff
@@ -0,0 +1,23 @@
...
+
+export const readFromLocalStorage = (key, parse = false) => {
+  const item = window.localStorage.getItem(key)
+  try {
+    return parse ? JSON.parse(item) : item
+  } catch (error) {
+    console.error(`Unable to read ${key} from local storage: ${error}`)
+    return null
+  }
+}
```
paulgain, line 23: Going forward it would be good to  just use these two functions to read and write to `localstorage`. All other instances in the codebase (i think there's just another two) need to be calling these functions - this work will be done in separate PRs. The idea is to have a central file that contains the keys and read/write functions to `localStorage` instead of them spread throughout the codebase. It will also make it easier to track all `localStorage` keys and avoids overwriting existing ones.

Usage:
`YOUR_KEY` should be defined in the same file, under the keys section.
```js
// String
writeToLocalStorage(YOUR_KEY, value)
readFromLocalStorage(YOUR_KEY)

// Object
writeToLocalStorage(YOUR_KEY, value, true)
readFromLocalStorage(YOUR_KEY, true)
```

backslashbaker (approved):

CaitBarnard (approved):

PR #5857, opened 2023-07-17
Merge the dashboard into the personalised dashboard

paulgain:
## Description of change
This works merges the dashboard into the personalised dashboard leaving a single dashboard.

**Dashboard tabs**
1. My companies lists
2. My referrals
3. Export projects

**Personalised dashboard tabs (now the only version)**
1. Company lists
2. Investment projects
3. Referrals
4. Export projects

### DAs and LEPs
Previously, DAs and LEPs were not allowed to view the tabs on the dashboard. We have removed that restriction so they can view all 4 tabs (was 3 now 4). However, when one of those users selects either the 'Company lists' or 'Referrals' tab they will get an error message informing them they don't have the permission to perform the action, this is enforced by the API. Previously, a handful of DAs and a significant amount of LEPs have been given a feature flag group which gives them access to the personalised dashboard where they get to view the same four tabs, they also receive the same error when clicking on tabs they don't have permissions for, for those users these changes will be invisible.

To determine if the user has the permissions to view the tabs we used the following check which has now been removed:
`const canViewTabs = userPermissions.includes('company_list.view_companylist')`.  Typically users that have that permission to view the tabs will have the group `DIT_staff`. Take a look at how [permissions](#permissions) work in Data Hub.

### Combined dashboard

**Users with no investment projects**

no-investment-projects

**Users with no investment projects and have subscribed to export reminders:**

no-invesmtents-has-export-reminders

**Users that have investments projects but haven't subscribed to investment or export reminders:**

investment-no-reminders

**Users that have investments projects and have subscribed to investment reminders:**

investments-with-reminders

**Users that have investments projects and have subscribed to both investments and export reminders:**

investments-with-both-reminders

### Window: localStorage
As Data Hub now remembers the last dashboard tab the user was previously on we shouldn't hear any complaints about their tab should be first in the row of tabs (the old default selection).

### Permissions  <a href="#permissions" id="permissions"/>

Each advisor is assigned a team (there are 1,350 teams)
A team is associated with a team role (there are 25 team roles)
A team role is assigned one or more groups (there are 10 groups)
A group contains a list of permissions (there are 602 permissions)

Example:
The advisor Paul Gain is assigned the team _Digital Data Hub - Live Service_
That team has the role _UKTI HQ Department_
That role is assigned the group _DIT_staff_
That group has a list of 36 permissions.

## Checklist


backslashbaker (approved):

src/lib/urls.js
```diff
@@ -346,6 +346,9 @@ module.exports = {
         '/projects/create/investment-type/info/:anchor'
       ),
     },
+    dashboard: {
+      projects: url('/investment-projects'),
+    },
```
paulgain, line 351: `projects` needs to be moved to the other `dashboard` object. However, making such a change touches a lot of files. This will be done in a subsequent PR.

src/client/components/DataHubFeed/index.jsx
```diff
@@ -28,48 +26,44 @@ const Date = styled('div')({
...
-          </Link>
-        </>
-      )}
-      {!items.length && <Paragraph>No updates available</Paragraph>}
-    </FeedContainer>
-  )
-}
+const DataHubFeed = ({ items, feedLimit = 5 }) => (
+  <FeedContainer>
+    <H3>{"What's new"}</H3>
```
chopkinsmade, line 31: Could get rid of the braces here now the conditional is removed

paulgain: Good spot. Thanks.

PR #5868, opened 2023-07-19
Reactify edit project summary form

cgsunkel:
## Description of change

The form for editing the investment project details summary has been reactified. A substantial amount of legacy code from old investment implementations has also been cleared out.

As a result of this work there is now a small amount of duplication in the fields used by the create project form and this one. As this PR is already very large, this work has been done in #5872 and will be put up for review separately once this has been merged in.

## Test instructions

It should still be possible to edit an investment project's details.

### Before

Screenshot 2023-07-19 at 16-13-18 Department for Business and Trade

### After

Screenshot 2023-07-19 at 16-14-04 Department for Business and Trade

## Checklist


src/client/modules/Investments/Projects/EditProjectSummary.jsx
```diff
@@ -0,0 +1,360 @@
...
+    : {
+        borderLeft: `${SPACING.SCALE_1} solid ${GREY_2}`,
+        marginLeft: SPACING.SCALE_4,
+      }),
+}))
+
+const checkReferralSourceAdviser = (currentAdviser, referralSourceAdviser) =>
+  currentAdviser === referralSourceAdviser
+
+const EditProjectSummary = ({ projectId, currentAdviser }) => (
```
chopkinsmade, line 73: Having nested Resources in this ways can slow down the rendering of the page, as each resource waits for the previous one before starting the network request. You can hopefully see it in this screenshot, where the request for referral-source-activity waits for investment-type to finish, then the investment resource (the guid) is waiting on both of the previous requests to finish. The network requests for the filters need to wait for all 3 Resources to finish, but then they all run at the same time as they are at the same level in React

An improvement could be to move the InvestmentTypesResource and ReferralSourceActivityResource inside the  InvestmentResource, and wrap them around the relevant Field components so they run at the same time as the other filter requests


image

test/end-to-end/cypress/specs/DIT/audit-spec.js
```diff
@@ -75,8 +78,9 @@ describe('Investment Project', () => {
   })
 
   it('should display name of the person who made investment project record changes', () => {
-    cy.get('[data-test="submit"]').click()
-    cy.get(selectors.message.successful).should('be.visible')
+    cy.wait(1000)
```
chopkinsmade, line 81: Instead of adding an arbitrary 1 second wait, could you use a `cy.intercept()` along with a `cy.wait()` to wait for one of the api requests to complete before moving onto clicking the button?
If you use `cy.intercept()` with just a matching url, and no response body, it acts in a spying mode where the real api response is returned but you can get your test to wait. Something like
```
cy.intercept('api-proxy/v4/metadata/investment-specific-programme').as('apiRequest')
cy.wait('@apiRequest')
```
Always waiting for 1 second might introduce a flaky test if one of those endpoints takes 1.1 seconds during a test run

chopkinsmade (approved):

PR #5943, opened 2023-08-07
On company tree move tags to below title when flexing

marijnkampf:
## Description of change

On company tree if the company name and tags are too wide to fit on a single line, the tags should be below the name. 

## Test instructions

View the company tree. The tags should either be on right right hand side next to the name (if there is space) or tags should be below the name.

### Before

image

### After

image

## Checklist

### DIFF
--- a/src/client/modules/Companies/CompanyHierarchy/styled.jsx
+++ b/src/client/modules/Companies/CompanyHierarchy/styled.jsx
@@ -227,6 +227,7 @@ export const GridColTags = styled(GridCol)`
 
 export const GridColHeader = styled(GridCol)`
   setwidth: two-thirds;
+  float: left;
   padding-left: 0px;
 `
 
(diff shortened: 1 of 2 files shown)


src/client/modules/Companies/CompanyHierarchy/styled.jsx
```diff
@@ -227,6 +227,7 @@ export const GridColTags = styled(GridCol)`
 
 export const GridColHeader = styled(GridCol)`
```
chopkinsmade, line 228: Not sure how much GridCol inherits from flexbox, but is it possible in this scenario to force the tags to the left?

marijnkampf: That's (probably) possible, but then wouldn't it look weird if there is a long Company name and there are only a few tags? See first child (Samir .... ) in screenshot below.

image

chopkinsmade (approved):

PR #6008, opened 2023-08-24
feature/CLS2-375-reactify-edit-one-list

chopkinsmade:
## Description of change

Move the logic for the edit one list page from a nunjucks controller into a React component. The tasks and form have been copied without modification, and a new Task used to get the one list information that was previously loaded in the controller

The existing controller and nunjucks have all been removed

## Test instructions

Use any exiting company, for example http://localhost:3000/companies/375094ac-f79a-43e5-9c88-059a7caa17f0/edit-one-list, you should be able to edit and make changes to the one list team in the same way as you can on the dev site

## Screenshots

chrome_izNXhWYXl1

## Checklist


src/client/modules/Companies/CoreTeam/EditOneList.jsx
```diff
@@ -0,0 +1,87 @@
+import React from 'react'
+import { useParams, useLocation } from 'react-router-dom'
+import { ID, TASK_GET_ONE_LIST_DETAILS, state2props } from './state'
+import { connect } from 'react-redux'
```
paulgain, line 4: For consistency, can we keep group all third-party/external dependencies from local dependencies (also grouped) with a line break separating the two.

src/client/modules/Companies/CoreTeam/EditOneListForm.jsx
```diff
@@ -0,0 +1,78 @@
...
+  company,
+  oneListTiers,
+  formInitialValues,
+  returnUrl,
+}) => (
+  <Form
+    id="edit-one-list"
+    name={TASK_SAVE_ONE_LIST_DETAILS}
+    initialValues={formInitialValues}
+    submissionTaskName={TASK_SAVE_ONE_LIST_DETAILS}
```
paulgain, line 29: There's a props warning in the console and unfortunately  we have too many of these throughout the app. We can fix it by adding the following line:
```js
cancelRedirectTo={() => urls.companies.businessDetails(company.id)}
```
Ideally, we would have a "Back" button or "Cancel without saving" button. At present the user has to go back via the browser back button or via the bread crumbs.

paulgain (approved):

src/client/modules/Companies/CoreTeam/tasks.js
```diff
@@ -0,0 +1,91 @@
...
+    response = await Promise.reject(
+      e.message ||
+        (e.errors &&
+          e.errors.one_list_tier &&
+          Object.values(e.errors.one_list_tier).join(', ')) ||
+        e.toString()
+    )
+  }
+  response.data
+}
```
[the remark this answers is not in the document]
chopkinsmade: I've moved the inner functions onto their own level, and got rid of the `let response` as we didn't need the response value anyway.

I've left the `if / else if / else` statement, as depending on the choices the user makes in the form there are 3 possible api combinations. If there were only 2 we could go for a ternary, but with the 3 options it reads easier using if else

src/client/modules/Companies/CoreTeam/EditOneListForm.jsx
chopkinsmade: I've added that cancel redirect to get rid of the warning. I've looked into the cancel button not appearing, and it looks like the `<Form>` component hides the cancel button when it's a multistep form. As this is just a reactify ticket I'll leave that as it is for now. 
We've had conversations about needing to make changes to that form in future as not all options are available to users on the radio button page, so can be looked at then

CaitBarnard (approved):

PR #6039, opened 2023-09-01
Add filter for sub-sector to companies page

CaitBarnard:
## Description of change

Add sub-sector filter to companies page and interactions page

## Test instructions

On the companies page or interactions page, under filters there should be sub-sectors filter. You can search for sub-sectors and filter by them. It should work regardless of what the sectors filter is set to. For example, sector filtered to Aerospace and sub-sector filtered to Music will show companies/interactions that are aerospace or music

## Screenshots
Screenshot 2023-09-01 at 14 57 51

Screenshot 2023-09-01 at 15 03 53

## Checklist


chopkinsmade (approved):

src/client/modules/Companies/CollectionList/tasks.js
```diff
@@ -35,6 +36,11 @@ const getCompanies = ({
   sortby = 'modified_on:desc',
 }) => {
   const administrativeAreas = [...us_state, ...canadian_province]
+  if (sub_sector_descends) {
+    sector_descends = sector_descends
```
paulgain, line 40: We shouldn't be assigning this as you're mutating the data. It's best practise to assign to new variable - no mutation.

src/client/modules/Interactions/CollectionList/tasks.js
```diff
@@ -57,13 +64,19 @@ const getInteractions = ({
...
   was_policy_feedback_provided,
   policy_areas,
   policy_issue_types,
   company_one_list_group_tier,
   dit_participants__team,
-}) =>
-  apiProxyAxios
+}) => {
+  if (sub_sector_descends) {
+    sector_descends = sector_descends
```
paulgain, line 75: Same applies here.

src/client/modules/Companies/CollectionList/tasks.js
```diff
@@ -36,18 +36,19 @@ const getCompanies = ({
...
-    sector_descends = sector_descends
-      ? [...sector_descends, ...sub_sector_descends]
-      : [...sub_sector_descends]
-  }
+  const combined_sectors =
+    sub_sector_descends || sector_descends
+      ? (sector_descends ? [...sector_descends] : []).concat(
+          sub_sector_descends ? [...sub_sector_descends] : []
+        )
+      : undefined
```
paulgain, line 44: This maybe easier to read and matches your original logic:

```js
const combined_sectors =
    sub_sector_descends && sector_descends
      ? [...sector_descends, ...sub_sector_descends]
      : sub_sector_descends
      ? [...sub_sector_descends]
      : []
```
As we're returning an array try to avoid returning `undefined`. Whatever is handling the `combined_sectors` array can simply loop over it without having to perform an `undefined` check.

CaitBarnard: We're trying to avoid having to return the empty array as we don't want sector_descends to be sent in the payload unless it has a value set. Would something like this work?
```Javascript
  const combined_sectors =
    sub_sector_descends && sector_descends
      ? [...sector_descends, ...sub_sector_descends]
      : sector_descends
      ? [...sector_descends]
      : sub_sector_descends
      ? [...sub_sector_descends]
      : undefined
```

paulgain: I think we posted at the same time. Looks a little crazy to be honest, try breaking it down for better readability.

paulgain: In functional programming this would be classified as a pure function.

paulgain: To be clear:

```js
  const combineSectors = (sector_descends, sub_sector_descends) =>
    sector_descends
      ? [...sector_descends, ...sub_sector_descends]
      : [...sub_sector_descends]

  ...

  sector_descends: sub_sector_descends
      ? combineSectors(sector_descends, sub_sector_descends)
      : undefined,
```

CaitBarnard: We're missing the case of having sectors only: `[...sector_descends]`. How would you suggest fitting that in? I agree, it's a little chaotic

paulgain: This is better than a ternary for readability, if we get to the end of the function we simply return an implicit `undefined`. There's no need to copy the last two arrays as we're simply passing them on to the API without mutating them.

```js
  const getSectors = (sector_descends, sub_sector_descends) => {
    if (sector_descends && sub_sector_descends) {
      return [...sector_descends, ...sub_sector_descends]
    }
    if (sub_sector_descends) {
      return sub_sector_descends
    }
    if (sector_descends) {
      return sector_descends
    }
  }

  ...

  sector_descends: getSectors(sector_descends, sub_sector_descends)
```

paulgain: Or adapting your ternary:

```js
  const combined_sectors =
    sub_sector_descends && sector_descends
      ? [...sector_descends, ...sub_sector_descends]
      : sector_descends
      ? sector_descends
      : sub_sector_descends
      ? sub_sector_descends
      : undefined
```

Either of these are acceptable, go with either one, what reads better to you?

CaitBarnard: I've changed it to the first suggestion as I think that's more understandable. Thanks for your help on this!

paulgain (approved):

PR #6046, opened 2023-09-04
Salary range disabled

bau123:
## Description of change

Change the average salary edit/create field in an investment. Salary ranges that have been disabled before the creation of the investment project can no longer be selected for the investment. However for existing investments with a salary range that is disabled before the project was created, it should still be displayed but can no longer be selected. Still can be selected in django admin 

## Test instructions

The salary range that is disabled before the project was created is no longer displayed on the page. Go to django page and disable a salary range before an investment project is created. Go on to the investment page and edit the salary range, the range should no longer be an option to select

### Before

image

image

image

### After
image

Disabled before

image

image

Disabled after

image

image

## Checklist


src/client/components/Form/elements/ResourceOptionsField/index.jsx
```diff
@@ -47,8 +47,11 @@ const ResourceOptionsField = ({
   payload,
   // Only required when Resource is not specified
   taskName,
+  date,
   resultToOptions = idNamesToValueLabels,
   interceptOption = (x) => x,
+  interceptOptionFilter = (x) =>
+    x.disabledOn === null || new Date(x.disabledOn) > new Date(date),
```
paulgain, line 54: You don't need to define `interceptOptionFilter`, you should use `interceptOption ` instead. The function passed to `interceptOption` will be called once for each of the options:

For example:
```js
<ResourceOptionsField
  interceptOption={(x) =>
    x.disabledOn === null || new Date(x.disabledOn) > new Date(date)
  }
/>
```

paulgain: The docs say:
```js
 * @param {Props['interceptOption']} [props.interceptOption] - A function that
 * will be called for each option already resolved with {props.resultToOptions}
 * and should return the tweaked option. Allows you to tweak the optios e.g.
 * to add children or hint to it.
```

So you need to return the option, not the date.

paulgain: Actually, overriding the `resultToOptions` is what we need. Here we can provide a function and return a filtered list of options based on your date logic. This is better as you don't need to touch the internals of `ResourceOptionsField` whatsoever :sunglasses:.

For example:
```js
<ResourceOptionsField
  ...
  resultToOptions={(result) =>
    // Returning true includes the option, returning false will exclude the option
    result.filter((x) => x.disabledOn ? new Date(x.disabledOn) > new Date(date) : true );
  }
  ...
/>
```

paulgain:
Looking good @bau123 

Ideally we need three tests around the investment project `created_on ` date.

1. The day before the option `disabled_on` date.
2. The day of the option `disabled_on` date.
3. The day after the option `disabled_on` date.

Something like this:
```
// The investment project
"created_on": "2016-02-05T12:00:00Z"
"created_on": "2016-03-05T12:00:00Z"
"created_on": "2016-04-05T12:00:00Z"

// The disabled option
"disabled_on": "2016-03-05T12:00:00Z"
```

paulgain (approved):

PippoRaimondiDIT (approved):

PR #6088, opened 2023-09-20
Reactify investment collection nav

cgsunkel:
## Description of change

The local nav for switching between the three investment `CollectionLists` has been reactified. The three `CollectionList`s have also been moved into the `modules` directory without being changed further. I would recommend reviewing this PR by commit to hide all the noise caused by this move.

This PR also contains a fix for `project-edit-recipient-spec`

## Test instructions

Go to the investments index page. You should be able to switch between the three `CollectionList` tabs and use them as before.

## Checklist


src/client/modules/Investments/InvestmentsRedirect.jsx
```diff
@@ -0,0 +1,10 @@
+import React from 'react'
+import { Redirect } from 'react-router-dom'
+
+import urls from '../../../lib/urls'
+
+const InvestmentsRedirect = () => {
+  return <Redirect to={urls.investments.projects.index()} />
```
paulgain, line 7: No need for the explicit return.

paulgain (commented):
* Ideally the 3 tabs need to go full width to match what we currently have.
* The investor profiles filters loading indicator needs switching out for the skeleton loading indicator (I realise you’re just moving code around but we want consistency).
* Previously the H1 page title on each of the 3 pages was “Projects”, “Investments” and “UK Opportunities” now it’s a static  “Investments” across all three pages.

cgsunkel:
>     * Previously the H1 page title on each of the 3 pages was “Projects”, “Investments” and “UK Opportunities” now it’s a static  “Investments” across all three pages.

I'm not sure if it's possible to restore these titles because these lists are now being loaded as a single page with the header only being rendered once. I used `Investments` to match the other high level list pages.

paulgain:
> > ```
> > * Previously the H1 page title on each of the 3 pages was “Projects”, “Investments” and “UK Opportunities” now it’s a static  “Investments” across all three pages.
> > ```
> 
> I'm not sure if it's possible to restore these titles because these lists are now being loaded as a single page with the header only being rendered once. I used `Investments` to match the other high level list pages.

Take a look at the reminders list as a working example: https://www.datahub.staging.uktrade.io/reminders. All you need to do is read the params to update the title. Here is a code example: https://github.com/uktrade/data-hub-frontend/blob/main/src/client/modules/Reminders/RemindersLists.jsx

src/client/modules/Investments/InvestmentCollections.jsx
```diff
@@ -7,16 +7,26 @@ import ProjectsCollection from './Projects/ProjectsCollection'
...
-    pageTitle="Investments"
+    heading={getCurrentTab(location.pathname)}
+    pageTitle={`${getCurrentTab(location.pathname)} - Investments`}
     breadcrumbs={[
       { link: urls.dashboard.index(), text: 'Home' },
       {
         link: urls.investments.index(),
         text: 'Investments',
       },
+      { text: getCurrentTab(location.pathname) },
```
paulgain, line 29: Calling `getCurrentTab` 3 times per page render in inefficient. It's best to call it once, save the value into a `const` and then use that value in the 3 places it's required. The algorithm can also be improved as `opportunities` will have to wait while `includes` iterates over the pathname of both `/projects` and `/profiles`. 

I recommend using a regex and an object instead of the above. The regex simply returns everything after the second `/`. For example: if `location.pathname` is `/investments/opportunities` then `PATH.exec(location.pathname)[0]` returns `opportunities` which then maps nicely to the key in the map returning `UK Opportunities`. 

Something like this:

```
const PATH = /([^\/]+$)/

const pathToPageTitleMap = {
  projects: 'Projects',
  profiles: 'Profiles',
  opportunities: 'UK Opportunities',
}

const InvestmentCollections = ({ ...props }) => {
  const path = PATH.exec(location.pathname)[0]
  const pageTitle = pathToPageTitleMap[path]
  return (
     <DefaultLayout
       ...
     />
  )
```

paulgain (approved):

PR #6095, opened 2023-09-21
Reactify edit OMIS quote information form

cgsunkel:
## Description of change

The form for editing the information for an OMIS quote has been reactified.

## Test instructions

Go to an OMIS order, scroll down to `Information for the quote` and click the Edit button. You should be able to edit both fields. You should also not be able to enter today's date (or any date before this) in the delivery date field.

### Before

Screenshot 2023-09-21 at 14 36 42

### After

Screenshot 2023-09-21 at 14 36 49

## Checklist


backslashbaker (approved):

src/client/modules/Omis/validators.js
```diff
@@ -0,0 +1,15 @@
...
+    if (values.delivery_date.year != '') {
+      return isDateInFuture(
+        transformDateObjectToDateString(values.delivery_date)
+      )
+        ? null
+        : 'Delivery date of work must be in the future'
+    }
+  }
+  return null
+}
```
paulgain, line 15: Consider simplifying your function:
1. I wouldn't pass `value` and `field` if you don't use them within your function.
2. There's no need to explicitly check for an empty string `!== ''` as empty strings are falsy. For example: if ('') { console.log('true') } else { console.log('false') } // false
3. If you only care about the year in your conditionals then test exactly that. I assume that if you have a `delivery_date.year` then you will have a `delivery_date`.
4. Ideally you would want to test this function to ensure it does what you think it does.

```
export const validateIfDateInFuture = (value, field, { values }) => {
  if (values?.delivery_date?.year) {
    return isDateInFuture(transformDateObjectToDateString(values.delivery_date))
      ? null
      : 'Delivery date of work must be in the future'
  }
  return null
}
```

cgsunkel: I added a test for this in the [edit-quote-spec](https://github.com/uktrade/data-hub-frontend/pull/6095/files#diff-487388c236a033f37263fea348bd163388c5a82051b34aed71dc2d25f14b8534R25). I will make the other suggested changes now

paulgain (approved):

PR #6108, opened 2023-09-26
Feature/cls2 430 view task detail

marijnkampf:
## Description of change

Users need to be able to view full details relating to a task.

This PR depends on the API branch feature/CLS2-462-tasks-API-should-return-details-for-investment-project-task

## Test instructions

Create an Investment Project Task and get the id of the related Task. View the Investment project task on http://localhost:3000/tasks/[taskId]. Make sure it's the id of the related Task and not the Investment Project Task (see steps below)

- [Create a Investment Project Task in the API.](http://localhost:8000/admin/task/investmentprojecttask/add/)
  - Create a Task for the Investment Project Task image 
     - Only the fields from Title onwards need completing image
  -  Either create or select an Investment Project.
- Get the id of the task by editing the related task image

Visit: http://localhost:3000/tasks/[taskId] to view the details of the Investment Project Task on the front end.

### Before

n/a

### After

image
## Checklist

[//]: # 'When submitting a PR make sure the code review guidelines have been satisfied.
https://github.com/uktrade/data-hub-frontend/blob/main/docs/Code%20review%20guidelines.md'

- [ ] Has the branch been rebased to main?
- [ ] Automated tests (Any of the following when applicable: Unit, Functional or End-to-End)
- [ ] Manual compatibility testing (Browsers: Chrome, Firefox, Edge, Safari)

PR #6109, opened 2023-09-27
Cls2 429 add task

backslashbaker:
## Description of change

Form to allow users to create tasks relating to investment projects

## Test instructions

Navigate to URL investments/projects/{projectid}/tasks/create and fill in form

### After

Screenshot 2023-09-27 at 12 31 05

## Checklist


src/client/modules/Investments/Projects/Tasks/InvestmentProjectTask.jsx
```diff
@@ -0,0 +1,174 @@
...
+                    legend="Task due date"
+                    required="Specify task due date"
+                    options={taskDueDateOptions.map((option) => ({
+                      ...option,
+                      ...(option.label === 'Custom date' && {
+                        children: <FieldDueDate />,
+                      }),
+                      ...option,
+                      ...option,
+                      ...option,
```
paulgain, line 108: This looks a little crazy,  do we really need to copy the object `option` over 4 times?

src/client/modules/Investments/Projects/Tasks/InvestmentProjectTask.jsx
```diff
@@ -0,0 +1,174 @@
...
+                  <FieldRadios
+                    name="taskRemindersEnabled"
+                    legend="Do you want to set a reminder for this task?"
+                    required="Specify reminder"
+                    options={OPTIONS_YES_NO.map((option) => ({
+                      ...option,
+                      ...(option.label === 'Yes' && {
+                        children: <FieldReminder />,
+                      }),
+                      ...option,
```
paulgain, line 120: Again, this doesn't look right as there's too much copying.

src/client/modules/Investments/Projects/Tasks/InvestmentProjectTask.jsx
```diff
@@ -0,0 +1,174 @@
...
+    hint="For example 28 11 2025"
+    required="Enter a date"
+    defaultValue={initialValue}
+  />
+)
+
+const FieldReminder = ({ initialValue = null }) => (
+  <GridRow>
+    <GridCol style={{ flexGrow: '0' }}>
+      <StyledFieldInput
```
chopkinsmade, line 155: image
Some styling issues on the validation here

src/client/modules/Investments/Projects/Tasks/InvestmentProjectTask.jsx
```diff
@@ -0,0 +1,174 @@
...
+    name="customDate"
+    hint="For example 28 11 2025"
+    required="Enter a date"
+    defaultValue={initialValue}
+  />
+)
+
+const FieldReminder = ({ initialValue = null }) => (
+  <GridRow>
+    <GridCol style={{ flexGrow: '0' }}>
```
paulgain, line 154: For consistency, we don't apply inline styles.

src/client/modules/Investments/Projects/Tasks/InvestmentProjectTask.jsx
```diff
@@ -0,0 +1,174 @@
...
+      <StyledFieldInput
+        className="govuk-input--width-2"
```
paulgain, line 156: We don't use `className`, for consistency please use `styled-components`.

chopkinsmade, line 156: We don't usually use class names, we use styled components. If this is to set the width of the input you can do that in your styled component

src/client/modules/Investments/Projects/Tasks/tasks.js
```diff
@@ -0,0 +1,10 @@
+import { apiProxyAxios } from '../../../../components/Task/utils'
+import { transformFormValuesForAPI } from './transformers'
+
+export const saveInvestmentProjectTask = ({ values, investmentProject }) => {
+  let request, endpoint
```
chopkinsmade, line 5: You don't need these variables declared as `let` just to set them below. You can use the values directly like 
```
return apiProxyAxios.post(
    '/v4/investmentprojecttask',
    transformFormValuesForAPI(values, investmentProject)
  )
```

src/client/modules/Investments/Projects/Tasks/transformers.js
```diff
@@ -0,0 +1,49 @@
...
+    title: taskTitle,
+    description: taskDescription,
+    due_date: getDueDate(taskDueDate, customDate),
+    email_reminders_enabled: taskRemindersEnabled === OPTION_YES,
+    reminder_days: taskReminderDays ? parseInt(taskReminderDays) : null,
+    advisers: [],
+  },
+})
+
+const getOneMonthFromNow = () => {
```
chopkinsmade, line 27: There a utility functions in src/client/utils/date.js that might be able to use here, instead of doing custom date manipulation

src/client/modules/Investments/Projects/Tasks/transformers.js
```diff
@@ -0,0 +1,49 @@
...
+  const date = new Date()
+  date.setMonth(date.getMonth() + 1)
+  return date.toISOString().slice(0, 10)
+}
+
+const getOneWeekFromNow = () => {
+  const date = new Date()
```
chopkinsmade, line 34: Same as above src/client/utils/date.js

src/lib/urls.js
```diff
@@ -444,6 +444,9 @@ module.exports = {
         '/projects/:projectId/remove-associated'
       ),
       evaluation: url('/investments', '/projects/:projectId/evaluation'),
+      task: {
```
chopkinsmade, line 447: can this be `tasks` instead, as the list of task page will be using this as well

src/client/modules/Investments/Projects/Tasks/transformers.js
```diff
@@ -0,0 +1,49 @@
...
+    id: investmentProject.id,
+    name: investmentProject.name,
+  },
+  task: {
+    title: taskTitle,
+    description: taskDescription,
+    due_date: getDueDate(taskDueDate, customDate),
+    email_reminders_enabled: taskRemindersEnabled === OPTION_YES,
+    reminder_days: taskReminderDays ? parseInt(taskReminderDays) : null,
+    advisers: [],
```
chopkinsmade, line 23: The adviser being sent should be the current logged in adviser, this is mandatory in the API but isn't working as expected at the moment, I'll look into why it isn't returning an error. 

The logged in advisor id is stored in redux as `currentAdviserId`

src/client/modules/Investments/Projects/Tasks/tasks.js
```diff
@@ -0,0 +1,10 @@
...
+  request = apiProxyAxios.post
+  endpoint = '/v4/investmentprojecttask'
+  values.investmentProject = investmentProject
+  return request(endpoint, transformFormValuesForAPI(values, investmentProject))
+}
```
paulgain, line 10: This is not easy to read, there's too much assignment/indirection with no benefit. Consider using the following:
```
export const saveInvestmentProjectTask = ({ values, investmentProject }) => {
  const body = transformFormValuesForAPI(values, investmentProject)
  return apiProxyAxios.post('/v4/investmentprojecttask', body)
}
```

You're mutating the values object with the assignment `values.investmentProject = investmentProject`. 
Mutation can lead to unexpected and hard-to-debug issues, for consistency with the rest of the codebase it's recommended to create a new object and copy `values` into it then set the field you care about. You could do this in `saveInvestmentProjectTask` (as you have done) or in the `transformFormValuesForAPI` (the preferred approach) function.

In addition, the function `transformFormValuesForAPI` destructures a single object, however, you call it with two objects.

src/client/modules/Investments/Projects/Tasks/InvestmentProjectTask.jsx
```diff
@@ -0,0 +1,174 @@
...
+                  <FieldRadios
+                    name="taskDueDate"
+                    legend="Task due date"
+                    required="Specify task due date"
+                    options={taskDueDateOptions.map((option) => ({
+                      ...option,
+                      ...(option.label === 'Custom date' && {
+                        children: <FieldDueDate />,
+                      }),
+                      ...option,
```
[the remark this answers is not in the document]
CaitBarnard: Looks like they're neccessary

src/client/modules/Investments/Projects/Tasks/InvestmentProjectTask.jsx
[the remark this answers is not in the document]
CaitBarnard: Yeah, it's a little confusing to read but there are 2 options and then the `...(option.label...` is the additional option when you select yes. Similarly on the due date field there are four `...options` and then the additional field under custom

src/client/modules/Investments/Projects/Tasks/InvestmentProjectTask.jsx
```diff
@@ -0,0 +1,174 @@
...
+import { useParams } from 'react-router-dom/cjs/react-router-dom.min'
+
+import {
+  Details,
+  GridCol,
+  GridRow,
+  Label,
+  ListItem,
+  UnorderedList,
+} from 'govuk-react'
```
paulgain, line 12: Space required to divide third-party dependencies and local dependencies.

src/client/modules/Investments/Projects/Tasks/InvestmentProjectTask.jsx
CaitBarnard: This was an edge-case as we had to scale the input without the validation also scaling. It's the best we could manage but we're happy to discuss it if you have ideas?

src/client/modules/Investments/Projects/Tasks/transformers.js
```diff
@@ -0,0 +1,49 @@
...
+  const date = new Date()
+  date.setMonth(date.getMonth() + 1)
+  return date.toISOString().slice(0, 10)
+}
+
+const getOneWeekFromNow = () => {
+  const date = new Date()
+  date.setDate(date.getDate() + 7)
+  return date.toISOString().slice(0, 10)
+}
```
paulgain, line 37: We may already have these functions defined in `src/client/utils/date.js`, if not, please move both functions there so they can be reused.

paulgain (commented):
Looks good, just a few comments.

src/client/modules/Investments/Projects/Tasks/InvestmentProjectTask.jsx
```diff
@@ -0,0 +1,177 @@
...
+  { label: 'Custom date', value: 'custom' },
+  { label: '1 week', value: 'week' },
+  { label: '1 month', value: 'month' },
+  { label: 'No due date', value: 'none' },
+]
+
+const InvestmentProjectTask = ({ currentAdviserId }) => {
+  const { projectId } = useParams()
+  return (
+    <InvestmentResource id={projectId}>
```
chopkinsmade, line 57: I can see the other pages in investment have the same ordering, however if we move the `<InvestmentResource` inside the `<DefaultLayout` component we get a nicer display with some breadcrumbs instead of just an error:
image

paulgain: I agree, you cannot load a page where it's blank and all you see is a loading indicator. We need to see the local header and breadcrumbs while you're fetching the data. Take a look at the following form which illustrates the point (refreshing the page will make you understand.)

`https://www.datahub.dev.uktrade.io/export/create?companyId=ee8b5e64-2bd0-4a24-8a6f-3c410627f471`

At Chris has correctly pointed out the `DefaultLayout` needs to be at the top of the component heiracrchy.

src/client/modules/Investments/Projects/Tasks/tasks.js
```diff
@@ -0,0 +1,16 @@
...
+  currentAdviserId,
+}) => {
+  values.investmentProject = investmentProject
+  values.currentAdviserId = currentAdviserId
+
+  return apiProxyAxios.post(
+    '/v4/investmentprojecttask',
+    transformFormValuesForAPI(values, investmentProject, currentAdviserId)
+  )
+}
```
paulgain, line 16: **You have resolved my original comment without making the changes requested.**

Can you please read my original comment. Again, it's not good to explicitly assign a field to an existing object as you're mutating the data (an existing `values` object in this case).

```
  values.investmentProject = investmentProject
  values.currentAdviserId = currentAdviserId
```

This function should look like:
```
export const saveInvestmentProjectTask = ({
  values,
  investmentProject,
  currentAdviserId,
}) => {
  return apiProxyAxios.post(
    '/v4/investmentprojecttask',
    transformFormValuesForAPI(values, investmentProject, currentAdviserId)
  )
}
```

The transformer should look like:
```
export const transformFormValuesForAPI = (
  formValues,
  investmentProject,
  currentAdviserId
) => {
  return {
    investment_project: {
      id: investmentProject.id,
      name: investmentProject.name,
    },
    task: {
      title: formValues.taskTitle,
      description: formValues.taskDescription,
      due_date: getDueDate(formValues.taskDueDate, formValues.customDate),
      email_reminders_enabled: formValues.taskRemindersEnabled === OPTION_YES,
      reminder_days: formValues.taskReminderDays
        ? parseInt(formValues.taskReminderDays)
        : null,
      advisers: [currentAdviserId],
    },
  }
}
```
Why? Because, now I can see from the function arguments what comes from where, compare this to what you have where it's all just bundled into a single object and then destructured immediately.

paulgain:
I'm unsure of the requirements back it appears you can add a date from the past and successfully save:
past-date

src/client/modules/Investments/Projects/Tasks/InvestmentProjectTask.jsx
chopkinsmade: Marijns PR has this for the Task page as well if you want another example. https://github.com/uktrade/data-hub-frontend/pull/6108/commits/6cb291ff4544286d1e948a1dc1e0f2d1f2c937aa

src/client/modules/Investments/Projects/Tasks/InvestmentProjectTask.jsx
```diff
@@ -0,0 +1,161 @@
+import React, { useState } from 'react'
+import styled from 'styled-components'
+import { useParams } from 'react-router-dom'
+import { connect } from 'react-redux'
+
+import { Details, ListItem, UnorderedList, Input } from 'govuk-react'
```
paulgain, line 6: This needs to move up one line so it's grouped with the other third-party imports.

src/client/modules/Investments/Projects/Tasks/InvestmentProjectTask.jsx
```diff
@@ -0,0 +1,161 @@
...
+      ]}
+      useReactRouter={false}
+    >
+      {!investmentProject ? (
+        <InvestmentResource id={projectId}>
+          {(investmentResource) => {
+            setInvestmentProject(investmentResource)
+          }}
+        </InvestmentResource>
+      ) : (
```
paulgain, line 65: If you need to fetch data for the form then you should use the form prop `initialValuesTaskName`.

Here's an example you can search for:

```
`state.js`
export const TASK_GET_INTERACTION_INITIAL_VALUES =
  'TASK_GET_INTERACTION_INITIAL_VALUES'


Your form
initialValuesTaskName={TASK_GET_INTERACTION_INITIAL_VALUES}

`src/client/tasks.js`
[TASK_GET_INTERACTION_INITIAL_VALUES]:
	addInteractionFormTasks.getInitialFormValues,

// Your local task
export async function getInitialFormValues({ investmentId }) {
  return apiProxyAxios
    .get(`v4/some/endpoint/${investmentId}`)
    .then(({ data }) => data) // your investment
}

```

src/client/modules/Investments/Projects/Tasks/InvestmentProjectTask.jsx
```diff
@@ -0,0 +1,161 @@
...
+const taskDueDateOptions = [
+  { label: 'Custom date', value: 'custom' },
+  { label: '1 week', value: 'week' },
+  { label: '1 month', value: 'month' },
+  { label: 'No due date', value: 'none' },
+]
+
+const InvestmentProjectTask = ({ currentAdviserId }) => {
+  const { projectId } = useParams()
+  const [investmentProject, setInvestmentProject] = useState(undefined)
```
paulgain, line 43: We have Redux, there's no need to use `useState`  here.

paulgain (commented):
Guys, its looking and behaving really good. Just a couple of comments that need addressing.

src/client/modules/Investments/Projects/Tasks/validators.js
```diff
@@ -0,0 +1,13 @@
+const { transformDateObjectToDateString } = require('../../../../transformers')
+const { isDateInFuture } = require('../../../../utils/date')
+
+export const validateIfDateInFuture = (values) => {
+  return isDateInFuture(transformDateObjectToDateString(values))
+    ? null
+    : 'Enter a date in the future'
+}
```
paulgain, line 8: No need for the explicit return:

```
export const validateIfDateInFuture = (values) =>
  isDateInFuture(transformDateObjectToDateString(values))
    ? null
    : 'Enter a date in the future'
```

CaitBarnard: I thought this as well but it failed without 🤔

src/client/modules/Investments/Projects/Tasks/state.js
```diff
@@ -0,0 +1,6 @@
+export const TASK_SAVE_INVESTMENT_PROJECT_TASK =
+  'TASK_SAVE_INVESTMENT_PROJECT_TASK'
+
+export const state2props = ({ currentAdviserId }) => {
+  return { currentAdviserId: currentAdviserId }
+}
```
paulgain, line 6: This can be simplified to: 

`export const state2props = ({ currentAdviserId }) => ({ currentAdviserId })`

src/client/modules/Investments/Projects/Tasks/tasks.js
```diff
@@ -0,0 +1,13 @@
...
+export const saveInvestmentProjectTask = ({
+  values,
+  investmentProject,
+  currentAdviserId,
+}) => {
+  return apiProxyAxios.post(
+    '/v4/investmentprojecttask',
+    transformFormValuesForAPI(values, investmentProject, currentAdviserId)
+  )
+}
```
paulgain, line 13: For consistency, favour the implicit return over the explicit return:

```
export const saveInvestmentProjectTask = ({
  values,
  investmentProject,
  currentAdviserId,
}) =>
  apiProxyAxios.post(
    '/v4/investmentprojecttask',
    transformFormValuesForAPI(values, investmentProject, currentAdviserId)
  )
```

src/client/modules/Investments/Projects/Tasks/InvestmentProjectTask.jsx
chopkinsmade: The `setState` is being used here to store the result of the `<InvestmentResource>`, which was added because of the issue with the `<DefaultLayout>`  being inside the `<InvestmentResource>` component and not showing when there was an error with the API call. Using the `setState` lets the result of the `<InvestmentResource>` be accessible to the breadcrumbs and page title.

I can think of a few ways we solve this issue as we move to using Resource components more:

1. Use this approach of combining Resource components with setState, though I can see it becoming repetitive if used more widely
2. Avoid using Resource components for page level data loading like this page, and the other investment project pages. Instead use redux and the `Task` component. This would solve the above issue, but does require redux actions, state and reducers to be written negating the benefit of a Resource that handles all that on your behalf.
3. Come up with some new component that combines Resources and Page level loading to allow the layout to be used and display errors with the datahub framework loading.

paulgain: I understand this topic is very subjective, here's my thoughts:

The Resource component is not meant for every use case, I think we may have spoken about this before. I don't understand the fascination with these components, I understand they come without the boilerplate and they save us time but they also come with a lot of inflexibility and poor UX when using a bunch of them together (more on this further down). If you're having to use the `InvestmentResource` in this manner,  (i.e. `setState`) you're using it wrong and clearly violating the redux architecture.

1. We really don't want to roll this option out for the reasons just given.
2. I agree, we could use a `Task` and the associated boilerplate this would fix the problem. We could also use the `InvestmentResource` and add the form inside of that, something like:
```js
<DefaultLayout
  ...
  >
  <InvestmentResource>
    {(investment) => {
      // update the breadcrumb here
      <FormLayout ...>
        <Form>...</Form>
      <FormLayout />
    }}
  </InvestmentResource>
</DefaultLayout>
```
This would mean not showing the form until we had a response from the API, only then could the breadcrumbs be updated, but at least the local header and breadcrumbs would be visible while we wait. The problem with this approach is that if in the future we add more components that require data, we need to add more Resource components, then you get the vertical line of loading indicators, one for each component, in this scenario the UX is poor in my opinion - the Resource components do not address this problem, it's great from an engineering standpoint but not from a user's perspective.

If you use a `Task` instead, we can make multiple API calls in parallel, this approach future proofs as we can easily add more API calls and the user only views a single loading indicator compared to a vertical line of loading indicators lighting up the page like a Christmas tree.

3. Agreed, unsure what this would look like, it would need to address the issues above though.

chopkinsmade (approved):

paulgain (approved):

PR #6112, opened 2023-09-27
Reactify OMIS edit internal information page

cgsunkel:
## Description of change

The form for editing the internal information for an OMIS order has been reactified. It's also now possible to save multiple service types (this was the intended functionality but when adding multiple service types in the current Nunjucks form only the first is saved).

## Test instructions

Go to an OMIS order. It should be possible to edit the internal information fields.

### Before

Screenshot 2023-09-27 at 22-15-48 Department for Business and Trade

### After

Screenshot 2023-09-27 at 22-15-32 Edit internal information - SDE234_91 - Orders (OMIS) - DBT Data Hub

## Checklist


src/client/modules/Omis/tasks.js
```diff
@@ -1,7 +1,6 @@
 import { apiProxyAxios } from '../../components/Task/utils'
 
-export const updateOrder = (values) =>
-  apiProxyAxios.patch(
-    `v3/omis/order/${values.id}?_csrf=${values.csrfToken}`,
-    values
-  )
+export const updateOrder = (values) => {
+  const useToken = values.csrfToken ? `?_csrf=${values.csrfToken}` : ''
+  return apiProxyAxios.patch(`v3/omis/order/${values.id}` + useToken, values)
```
chopkinsmade, line 5: You've already got string interpolation setup here you can use instead of the string concatonation:
`return apiProxyAxios.patch(``v3/omis/order/${values.id}${useToken}``, values)`

src/client/modules/Omis/EditInternalInformation.jsx
```diff
@@ -0,0 +1,117 @@
...
+  transformArrayForTypeahead,
+  transformObjectForTypeahead,
+} from '../Investments/Projects/transformers'
+import { transformInternalInformationForApi } from './transformers'
+import { idNamesToValueLabels } from '../../utils'
+
+const EditInternalInformation = () => {
+  const { orderId } = useParams()
+  return (
+    <OrderResource id={orderId}>
```
chopkinsmade, line 30: This needs a wider discussion, but we noticed yesterday on a different ticket that when a `<Resource>` component is at the top level and the `<DefaultLayout>` is a child component, when an error occurs in the `<Resource>`  the user gets an error message without any of the wider datahub page like header, navbar, etc. Simply moving the `<Resource>` to be a child of `<DefaultLayout>` isn't always possible, as in this PR the order is needed to construct the breadcrumbs and build the page title

We tried to get around this in this PR by using local `useState` to store the Resource item, but there might be a better way to approach this: https://github.com/uktrade/data-hub-frontend/pull/6108/files#diff-3f38e8d33d91e4a3f79784220d9ff30555ed11d3fc54764f2a6d508ffe7a23b6R66

test/functional/cypress/specs/omis/edit-internal-info-spec.js
```diff
@@ -0,0 +1,75 @@
...
+      assertFieldTextarea({
+        element,
+        label:
+          'Specific people or organisations the company does not want DBT to talk to (optional)',
+        value: null,
+      })
+    })
+  })
+
+  it('should submit form successfully', () => {
```
chopkinsmade, line 61: A good assert to add to this test is checking the body of the POST request, there is an example in this [PR ](https://github.com/uktrade/data-hub-frontend/pull/6109/files#diff-ebd3ecdecd398bc8096703b0bd3c868e41dfad9c95f3adf4aed844aab06e9ac0R108)but searching for `assertPayload` in the codebase will give loads of examples

chopkinsmade (approved):

dredmonds (approved):

src/client/modules/Omis/transformers.js
```diff
@@ -16,3 +17,24 @@ export const transformQuoteInformationForApi = ({
...
+  const infoValues = {
+    id: orderId,
+    service_types: service_types.map((x) => x.value),
+    sector: checkIfItemHasValue(sector?.value),
+    further_info,
+    existing_agents,
+    contacts_not_to_approach,
+  }
+
+  return infoValues
```
paulgain, line 39: There's no value in setting this object to a `const`. For consistency, can we just return the object please.

src/client/modules/Omis/EditInternalInformation.jsx
paulgain: I wonder if we could arrive at a consensus for the imports, the length of some imports are longer than others. I was thinking ordering them by length where the longest is at the top and the shortest at the bottom. It would make the imports easier to read compared to a jagged approach. It would be good to hear what others think.

src/client/modules/Omis/EditInternalInformation.jsx
```diff
@@ -0,0 +1,117 @@
...
+} from '../../components/Resource'
+import urls from '../../../lib/urls'
+import { TASK_EDIT_OMIS_INTERNAL_INFORMATION } from './state'
+import { FORM_LAYOUT } from '../../../common/constants'
+import {
+  transformArrayForTypeahead,
+  transformObjectForTypeahead,
+} from '../Investments/Projects/transformers'
+import { transformInternalInformationForApi } from './transformers'
+import { idNamesToValueLabels } from '../../utils'
```
paulgain, line 25: I was wondering if we could arrive at a consensus for the imports, the length of some imports are longer than others. I was thinking ordering them by length where the longest is at the top and the shortest at the bottom. It would make the imports easier to read compared to a jagged approach. It would be good to hear what others think.

PR #6119, opened 2023-10-02
Reactify OMIS payment reconciliation page

cgsunkel:
## Description of change

The form for reconciling an OMIS payment has been reactified.

## Test instructions

With your local FE pointing to dev, go to the [OMIS reconciliation list](http://localhost:3000/omis/reconciliation?sortby=payment_due_date%3Aasc&status=quote_accepted) and click on an order. You should be able to reconcile the payment. Note that some of the orders within dev don't have invoices, which will raise an error. There aren't any orders in production at the `quote_accepted` stage without invoices, so this can be disregarded.

View an order with the status set to `draft` or `quote_awaiting_acceptance` (you can find these by unticking all the status filters on the reconciliation list). You should see a message stating that the payment cannot be reconciled.

View an order with the status set to `paid` or `complete`. You should see a message stating that the order has already been paid for and a link to the receipt.

View an order with the status set to `cancelled`. You should see a message stating that the payment cannot be reconciled because the order has been cancelled.

### Before

Screenshot 2023-10-02 at 12-38-33 Department for Business and Trade

### After

Screenshot 2023-10-02 at 12-41-08 Reconcile a payment - CTU296_22 - Orders (OMIS) - DBT Data Hub

## Checklist


src/client/modules/Omis/PaymentReconciliation.jsx
```diff
@@ -0,0 +1,222 @@
...
+                    <Table.Row>
+                      <Table.CellHeader>Order reference</Table.CellHeader>
+                      <Table.Cell>{order.reference}</Table.Cell>
+                    </Table.Row>
+                    <Table.Row>
+                      <Table.CellHeader>
+                        Amount (excluding VAT)
+                      </Table.CellHeader>
+                      <Table.Cell>
+                        {currencyGBP(invoice.subtotalCost / 100)}
```
cgsunkel, line 138: For some reason the OMIS costs are stored in pence rather than pounds. Therefore in order to display the correct value they need to be changed to the value in pounds.

package.json
```diff
@@ -81,7 +81,6 @@
     "basic-auth": "^2.0.1",
     "body-parser": "^1.20.2",
     "case": "^1.6.3",
-    "chrono-node": "^2.6.5",
```
cgsunkel, line 84: This dependency was only being used in the controllers for this page and the quote information form (which was already reactified in #6095).

paulgain:
After the success of the fire break week shouldn't we be favouring component tests over functional tests?

src/client/modules/Omis/PaymentReconciliation.jsx
```diff
@@ -0,0 +1,222 @@
...
+              text: 'Orders (OMIS)',
+            },
+            {
+              link: urls.omis.order(order.id),
+              text: order.reference,
+            },
+            { text: 'Reconcile a payment' },
+          ]}
+        >
+          {(order.status === 'draft' ||
```
paulgain, line 73: I feel any `status` on this page should be a constant. Constants are preferred at they can be exported/imported to other areas of the code and reused. You have less chance of making a typo as you only define it once and you can view all statuses together. When magic strings are defined within a React component this means they're instantiated on every page render, obviously the strings here are small so any performance gain is negligible.

In addition, you only have to instantiate a string once rather than once per page render.

cgsunkel:
> After the success of the fire break week shouldn't we be favouring component tests over functional tests?

This work was done before the Firebreak had started which is why this was done with a functional test. I will update this...

src/client/modules/Omis/constants.js
```diff
@@ -0,0 +1,6 @@
+export const STATUS_DRAFT = 'draft'
+export const STATUS_QUOTE_AWAITING_ACCEPTANCE = 'quote_awaiting_acceptance'
+export const STATUS_CANCELLED = 'cancelled'
+export const STATUS_PAID = 'paid'
+export const STATUS_COMPLETE = 'complete'
+export const STATUS_QUOTE_ACCEPTED = 'quote_accepted'
```
paulgain, line 6: You can save yourself a bunch of imports buy grouping them into an object.
```
export const STATUS = {
  DRAFT:  'draft',
  ...
}

import STATUS from './some/path'
...
order.status === STATUS.DRAFT
```

paulgain (approved):

test/functional/cypress/specs/omis/reconciliation-spec.js
```diff
@@ -92,15 +99,60 @@ describe('Load reconciliation collection view', () => {
...
+      cy.get('[data-test="something-wrong-details"]')
+        .click()
+        .should('contain', 'Is something wrong with the payment?')
+        .should(
+          'contain',
+          'View the order to contact an adviser in the UK or contact omis.orders@digital.trade.gov.uk'
+        )
+    })
+
+    it('should not submit the form with no values', () => {
```
chopkinsmade, line 118: If you wanted to make this functional test smaller, you could extract the Form in https://github.com/uktrade/data-hub-frontend/pull/6119/files#diff-160eaf543dd5af39f298278cff0f97aefbc889a65c45a2319863a755b683a64cR179 into it's own component, and then add a component test for these form validation checks.
This functional test can then be used for the happy path of asserting the payload matched

chopkinsmade (approved):

PR #6285, opened 2023-11-23
Cls2 550 completed tasks project investment screen

samuele-mattiuzzo:
## Description of change

[JIRA ticket](https://uktrade.atlassian.net/browse/CLS2-550)

This change allows the DataHub users to be able to see the completed tasks on the investment project screen. It removes the `archived` filter from the API calls (returning both completed and not completed Tasks) and instead shows a `COMPLETED` tag both in the list and details. It also removes the `Mark as complete` button when needed and converts the back button to a link as per requested.

## Test instructions

- When viewing an Investment Project, add a task. The task should be marked as not archived (no tag)
- Add a second task for good measure
- Pick the first task, view its details page and mark as completed
- This will redirect you to the tasks list page. Two tasks should still be visible (as opposed to just one). One of the two tasks should have a completed tag
- View the details page for the completed task. The completed tag should be under its title and the `mark as complete` button should not be available

## Screenshots

Screenshot 2023-11-23 at 14 58 48
Screenshot 2023-11-23 at 14 58 59
Screenshot 2023-11-23 at 14 59 06

## Checklist


src/client/modules/Tasks/TaskDetails/index.jsx
```diff
@@ -2,26 +2,51 @@ import React from 'react'
...
   const { taskId } = useParams()
 
   const company = getCompany(task)
   const taskTitle = task ? task.title : ''
+  const archived = getArchivedStatus(task)
 
   return (
     <DefaultLayout
-      heading={taskTitle}
+      heading={{
```
chopkinsmade, line 38: Adding all this to the `heading` property loses the h1 formatting and introduces large gaps around the title. Looking at the `DefaultLayout` component, there is also a `subheading` property that could be used to show the Tag and leave the heading prop to just have the title

src/client/modules/Tasks/TaskDetails/TaskButtons.jsx
```diff
@@ -43,27 +43,26 @@ export const TaskButtons = ({ task, editUrl }) => (
...
+            buttonColour={GREY_3}
+            buttonTextColour={TEXT_COLOUR}
+            as={Link}
+            href={editUrl}
+            data-test="edit-form-button"
+          >
+            Edit
+          </Button>
+        )}
+        <Link
```
chopkinsmade, line 57: Changing this to a `Link` might be in the designs, but it looks strange when next to the other buttons as it doesn't align properly. Can you check with Andy if this is desired to have buttons combined with a link, and if it is we'll need to fix the alignment

image

src/client/modules/Investments/Projects/transformers.js
```diff
@@ -284,16 +285,29 @@ export const transformInvestmentProjectToListItem = ({
...
   id,
   title,
   dueDate,
   advisers,
+  archived,
 } = {}) => ({
   id,
   headingUrl: urls.tasks.details(id),
-  headingText: title,
+  headingText: getTaskHeading(archived, title),
```
chopkinsmade, line 310: Using the `headingText` property causes the tags to be rendered inside an `<a>` tag with additional padding around them. `CollectionItem` has a `subheading` prop that could be used instead for displaying these tags

samuele-mattiuzzo:
@chopkinsmade I'll address your comments + make sure the tests are passing. I'm not keen on mark this as ready for review yet though, that unit test still fails :/

samuele-mattiuzzo:
@chopkinsmade after our chat, I reverted the back link to a back button as it looks better, agreed. I'll talk with Andy separately. I'll mark this as ready for review, there seem to be transient tests failing so I'll keep an eye on it. The unit test failing is still a mistery

src/client/modules/Tasks/TaskDetails/index.jsx
```diff
@@ -18,10 +27,18 @@ const TaskDetails = ({ task }) => {
 
   const company = getCompany(task)
   const taskTitle = task ? task.title : ''
+  const archivedTag = task?.archived ? (
```
chopkinsmade, line 30: Same as above, can do away with the conditional
```
const archivedTag = task?.archived && (
    <StyledTag colour="green" data-test="activity-kind-label">
      COMPLETED
    </StyledTag>
  )```

chopkinsmade (approved):

chopkinsmade (approved):

backslashbaker (approved):

PR #6464, opened 2024-01-26
Company layout cleanup

cgsunkel:
## Description of change

This is a cleanup PR for the company layout refactoring work.

The following small changes have been made:
- The old `CompanyLayout` component has been deleted and replaced by the new one.
- The `flashMessages` prop has been removed as these aren't being passed down through the controllers anymore.
- The add/remove from company list form has now been refactored to remove the need for a specific `returnUrl` prop to be passed down from all the company pages.
- The `CompanyLocalHeader` now gets the `csrfToken` from the state instead of needing it to be passed down.
- The export collection lists now use the skeleton placeholder.
- A couple of other very minor changes.

## Test instructions

Company pages should still work as normal.

## Checklist


src/apps/company-lists/controllers/add-remove.js
```diff
@@ -18,7 +18,6 @@ async function handleAddRemoveCompanyToList(req, res, next) {
       listsToUpdate.push(addOrRemoveCompany(req, listId, id))
     }
     await Promise.all(listsToUpdate)
-    req.flash('success', 'Lists changes for this company have been saved.')
```
cgsunkel, line 21: This flash message is already being generated in the form

src/client/components/CompanyLocalHeader/LocalHeaderCompanyLists.jsx
```diff
@@ -43,18 +43,17 @@ const StyledAddButton = styled(StyledCompanyListButton)`
   border-bottom: 3px solid ${GREY_3_LEGACY};
 `
 
-export const LocalHeaderCompanyLists = ({ results, company, returnUrl }) => {
-  const queryString = returnUrl ? `${returnUrl}` : `/companies/${company.id}`
+export const LocalHeaderCompanyLists = ({ results, companyId }) => {
   const handleClickAddRemove = () => {
-    window.location.href = `/companies/${company.id}/lists/add-remove?returnUrl=${queryString}`
+    window.location.href = `/companies/${companyId}/lists/add-remove?returnUrl=${window.location.pathname}`
   }
```
peterhudec, line 49: This is perhaps a good opportunity to replace this approach with the buttons being links?

src/client/components/CompanyLocalHeader/state.js
```diff
@@ -1,3 +1,7 @@
 export const TASK_GET_LISTS_COMPANY_IS_IN = 'TASK_GET_LISTS_COMPANY_IS_IN'
 export const ID = 'companyListsCompanyIn'
 export const state2props = (state) => state[ID]
+
+export const companyState2Props = (state) => ({
+  csrfToken: state.csrfToken,
```
peterhudec, line 6: I'm pretty sure that we don't need to pass around CSRF tokens in React code that is making API calls. CSRF is only needed if the form is submitted in the classical way and the request is handled by the express backend.

If it's so, then we don't need connect the component to redux.

peterhudec (approved):
Looks good. I left a couple of comments.

src/client/components/CompanyLocalHeader/state.js
cgsunkel: The section that uses this token (the company unarchive URL) is still using Express unfortunately

PR #6479, opened 2024-02-01
Update export wins endpoint to use a hyphen instead of an underscore

paulgain:
## Description of change
1. Updated the export wins endpoint to use a hyphen instead of an underscore.
2. Update the company url so when clicking the company name when on the unconfirmed tab the user will be taken to the company details page.

## Test instructions
In the dev environment go to `/exportwins/unconfirmed` where the page should load unconfirmed export wins.

## Checklist

### DIFF
--- a/src/client/modules/ExportWins/Confirmed/ConfirmedWinsTable.jsx
+++ b/src/client/modules/ExportWins/Confirmed/ConfirmedWinsTable.jsx
@@ -44,7 +44,7 @@ export const ExportWinsTable = ({ exportWins }) => (
           <Table.Cell>{currencyGBP(total_expected_export_value)}</Table.Cell>
           <NoWrapCell>{formatMediumDate(date)}</NoWrapCell>
           <NoWrapCell>
-            {formatMediumDate(customer_response.created_on)}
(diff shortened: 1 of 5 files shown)


src/client/components/Resource/ExportWins.js
```diff
@@ -1,3 +1,3 @@
 import { createCollectionResource } from './Resource'
 
-export default createCollectionResource('Export Wins', 'v4/export-win')
+export default createCollectionResource('Export Wins', 'v4/export_win')
```
ince-dbt, line 3: I think you already got the right endpoint as we discussed this with Pawel this afternoon and we will be using export-win instead of export_win. So, I dont think this changes is needed. See [his PR here.](https://github.com/uktrade/data-hub-api/pull/5149#discussion_r1475054331)

paulgain: Thanks, I've updated the code in other places as well. This PR will fail the tests until the API endpoint changes have been merged, whether that's in this [PR](https://github.com/uktrade/data-hub-api/pull/5149) or another.

dredmonds (approved):

PR #6534, opened 2024-02-19
Simplified redux setup

peterhudec:
## Description of change

Simplifies Redux setup by:

* Removing the `src/client/middleware.js` module, because only one of the three things it exported was actually a middleware.
* Removed the `history`, `store` and `sagaMiddleware` props from `DataHubProvider`, because they only ever were passed the same values
* Generalized the way initial data gets into Redux state from the Nunjucks `react-slot` macro and how the useless "Unexpected key ..." error (thrown by Redux's `combineReducers`) is suppressed

Along the way I realized that the reasons why we had the `spa-base-path` Express middleware are no longer valid and the middleware can be removed, so I removed it alongside the doc that was explaining its usage.

## Test instructions

There should be no noticable difference. All existing tests should still be passing.


paulgain (approved):
Good to see the back of `src/middleware/spa-base-path.js`.

cgsunkel (approved):

PR #6567, opened 2024-02-29
Adds the notFoundView prop to Resource

peterhudec:
## Description of change

Adds the `notFoundView` prop to the `Resource` component.

## Test instructions

This should be a non-breaking change with no effect if the prop is not set.


cgsunkel:

PR #6589, opened 2024-03-07
Renamed task.error to task.isError

peterhudec:
## Description of change

This PR changes how `Task` handles rejected values.

* `task.error` has been renamed to `task.isError`
* `task.error` now contains the value with which the task rejected

## Test instructions

There should be no visible change.


paulgain:
@peterhudec looks good, I would have thought `hasError` is more appropriate than `isError`?

ince-dbt (approved):

PR #6631, opened 2024-03-19
Fix/export wins review css

peterhudec:
## Description of change

This PR fixes a number of styling issues in `/exportwins/review/:token`, whose Nunjucks template doesn't extend from `govuk/template.njk` and which is using some components that still rely on global CSS provided by the govuk template.


cgsunkel:

PR #6639, opened 2024-03-22
Added the ExportWinsMovingBanner to NavBar

peterhudec:
## Description of change

This PR adds the "Important: Export Wins is moving to Data Hub on 3 April 2024" banner to all pages right underneath the main navbar.

## Test instructions

1. Go to any Data Hub page
2. You should see a banner saying "Important: Export Wins is moving to Data Hub on 3 April 2024" right underneath the main navbar.
3. The link in the banner should go to https://data-services-help.trade.gov.uk/data-hub/updates/announcements/export-wins-is-moving-to-data-hub/

### Before

Screenshot 2024-03-22 at 12 07 49

### After

Screenshot 2024-03-22 at 12 07 30


src/client/components/DataHubHeader/ExportWinsMovingBanner.jsx
```diff
@@ -0,0 +1,32 @@
...
+    textDecoration: 'underline',
+  },
+})
+
+const ExportWinsMovingBanner = () => (
+  <Root>
+    <strong>Important:</strong>{' '}
+    <Link
+      target="_blank"
+      href="https://data-services-help.trade.gov.uk/data-hub/updates/announcements/export-wins-is-moving-to-data-hub/"
```
cgsunkel, line 25: I think this URL should be part of the URLs submodule

test/functional/cypress/specs/export-win/announcement-banner-spec.js
```diff
@@ -0,0 +1,25 @@
...
+        '/omis',
+        '/support',
+      ].forEach((page) => {
+        it(`"${page}" page`, () => {
+          cy.visit(page)
+
+          cy.contains(
+            'CompaniesContactsEventsInteractionsInvestmentsOrdersSupport' +
+              'Important: Export Wins is moving to Data Hub on 3 April 2024'
+          )
```
cgsunkel, line 20: Can a `data-test` be used here instead of having to check the nav bar text as well as the banner?

cgsunkel (commented):
I have a couple of minor comments, looks good otherwise

test/functional/cypress/specs/export-win/announcement-banner-spec.js
peterhudec: Although it doesn't really look nice, this actually a much more thorough test thant with `data-test`, because it asserts that the banner appears right after the navigation. To make a similar assertion with `data-test`, the test would have to rely on concrete DOM structure, which ist best to avoid.

cgsunkel (approved):

paulgain (approved):

PR #6671, opened 2024-04-01
Add export wins links to `/export`

peterhudec:
## Description of change

Add export wins links to `/export`.

## Test instructions

1. Add your DataHub account to the `export-wins` feature flag group in the way of your preference.
2. Navigate to `/export`
3. You should see the _View export wins_ button
4. If you click the button, it should navigate you to `/exportwins/rejected`
5. You should not see the button if your account is not a member of the `export-wins` feature flag group.

### Before

With no exports:

Screenshot 2024-04-01 at 14 53 23

With some exports:

Screenshot 2024-04-01 at 14 56 44

### After

With no exports:

Screenshot 2024-04-01 at 14 52 56

With some exports:

Screenshot 2024-04-01 at 14 57 11


paulgain:
Looks good, just a couple of comments:

All of the export win features that are hidden behind a user feature group and inside the test file `/export-win/user-feature-group-spec.js`. It would be good to keep everything in there for consistency. Before we go live we can just delete the file as the tests won't be required after that.

The link `Remove all filters` is now permanent, previously it was only shown when filters have been applied. This is inconsistent with the design at the time and how the link generally works throughout Data Hub. Take a look at Investments, Orders, Interactions etc for a working example. Companies and Contacts apply the archived filter by default so it's acceptable for `Remove all filters` to appear when the page loads.

ince-dbt (approved):

PR #6697, opened 2024-04-09
Delete Export Wins details page

paulgain:
## Description of change
Removed the details page from Export Wins as it's no longer required, we're using the summary page instead.

## Checklist


src/client/modules/ExportWins/Shared.jsx
```diff
@@ -0,0 +1,35 @@
...
+    Export wins
+  </Link>
+)
+
+export const VerticalSpacer = styled.div`
+  display: flex;
+  flex-direction: column;
+  gap: ${SPACING.SCALE_1};
+  margin-bottom: ${SPACING.SCALE_5};
+`
```
peterhudec, line 35: This could even go to `src/client/components` but I would remove the bottom margin from it as that should not be the responsiblilty of this component, but of the component which is using this.

test/component/cypress/specs/ExportWins/CustomerFeedback.cy.jsx
```diff
@@ -41,26 +41,42 @@ const dummyExportWin = (prefix, booleans) => ({
...
+    {
+      testTitle: 'All true',
+      win: dummyExportWin('all-true', true),
+      company: dummyCompany(),
+    },
+    {
+      testTitle: 'All false',
+      win: dummyExportWin('all-false', false),
+      company: dummyCompany(),
+    },
```
peterhudec, line 59: I would just hardcode `{id: 123}` here as the id can be anything and the `dummyCompany` function makes one thing that there's hole lot more to it than there really is.

The `dummyExportWin` is a function because it's parametrized.

```suggestion
    {
      testTitle: 'All true',
      win: dummyExportWin('all-true', true),
      company: {id: 123},
    },
    {
      testTitle: 'All false',
      win: dummyExportWin('all-false', false),
      company: {id: 123},
    },
```

peterhudec (approved):
Looks good, just to suggestions.

elcct (approved):

PR #6759, opened 2024-05-01
DPM 616 filters on investment profiles

marijnkampf:
## Description of change

Fix loading of metadata for filters on [Investments > Profiles > Investor profiles](http://localhost:3000/investments/profiles?page=1) on DBT Platform

## Test instructions

Filters should load correctly on the page.

### Before

image

### After

image

## Checklist

- [x] Has the branch been rebased to migration-deploy?

### DIFF
--- a/src/client/modules/Investments/Profiles/tasks.js
+++ b/src/client/modules/Investments/Profiles/tasks.js
@@ -1,4 +1,6 @@
+import urls from '../../../../lib/urls'
 import { apiProxyAxios } from '../../../components/Task/utils'
+import { getMetadataOptions } from '../../../metadata'
 
 import transformLargeCapitalProfiles from './transformers'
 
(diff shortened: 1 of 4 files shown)


peterhudec (commented):
It all looks good, but I'd like to know exactly what was the actual issue. Because from the screenshot it looks like the error was actually in the API. If it was caused by the frontend, I would expect to see a _400 Bad Request_ error.

Changes in this PR don't seem to change what was sent to the API, except perhaps the added `{filterDisabled: false}` to the `/v4/metadata/uk-region` request. Was that the cause of the error?

marijnkampf:
> Changes in this PR don't seem to change what was sent to the API, except perhaps the added `{filterDisabled: false}` to the `/v4/metadata/uk-region` request. Was that the cause of the error?

There is a generic call to metadata that I altered to add a query parameter (which fixed the issue). The code in this PR wasn't using the generic call previously. Having changed this to the generic call it now adds a query parameter which avoid the issues with Hawk on the API.

peterhudec:
Ok, I can see now that the `getMetadataOptions` function has the extra query string param in the [migration-deploy](https://github.com/uktrade/data-hub-frontend/tree/migration-deploy) branch. It makes sense now.

PR #6760, opened 2024-05-01
Fix/content security policy

peterhudec:
## Description of change

This PR tightens the _Content Security Policy (CSP)_ headers to the strictes possible level. For the app to continue to work with this CSP, a couple of things had to be changed:

### Test coverage

The Istanbul library we are using for instrumenting our code for code coverage is adding code which falls into the `unsafe-eval` category and thus violates our new, strict CSP policy.

Also, before this PR, all tests were run on CI with the `NODE_ENV` set to `"development"` instead of `"test"`. We want to develop the application with the production CSP headers to detect code/dependencies that violate it as early as possible.

So in this PR...

* Code is now instrumented for coverage only when `NODE_ENV` is set to `"test"`
* All tests on CI are run with `NODE_ENV=test`. For that I have introduced the new `./test.env` file, which is just a copy of `./sample.env` with `NODE_ENV=test`
* All Cypress tests on CI now test the application compiled with `npm run build:for-test-coverage`, which is the same as `npm run build` with `NODE_ENV=test`
* The `getHawkHeader` function doesn't anymore treat specially the case when `NODE_ENV=test`
* The `test/end-to-end/cypress/specs/{DA,LEP}/permission-spec.js` were removed because it was pretty much impossible to figure out why they were failing with `NODE_ENV=test` and because it turned out that the only thing they tested was that the API responds with 403 in certain cases, which is already covered in the API tests. They were not testing any frontend logic.

### Our code that violated the strict CSP policy

We use a handful of inline `<script>` and `<style>` tags which fall into the `unsafe-inline` category. To make them compliant with the new strict CSP policy, they need to have a _nonce_ attribute, whose value is generated (unique for each response) in the `headers` middleware and which is added to the CSP header.

Another problem was that the `NotificationAlert` component was using the the bell icon SVG compiled as a data URI in an `` tag's `src` attribute, which falls into the `unsafe-eval` category. The solution was to let React render the SVG directly, which required some tweaks to the icon's sizing.

### Dependencies violating the strict CSP policy

These dependencies are violating the strict CSP policy:

* `styled-components` — add inline `<style>` tags, which falls under the `unsafe-inline` category
* `qs` — uses code which falls into the `unsafe-eval` category
* `react-use` — some functions of the library add an iniline `<style>` tag, which falls under the `unsafe-inline` category
* The Google Tag manager snippet is an inline `<script>` tag and thus falls into the `unsafe-eval` category

Both `styled-components` and `qs` could be fixed by letting Webpack know about the _nonce_ we are using in the CSP policy, which is not a very well documented mechanism.

The problem with `react-use` could be solved by only importing the single function `useSearchParam` we are using directly from `react-use/lib/useSearchParam`. 

To make the Google Tag manager work, special directives to allow sources from specific Google domains had to be added to the CSP policy.

## Test instructions

* The application should work exactly as before
* except that if you now use any of the following in the source code, the code should be rejected by the browser with a CSP `EvalError: ...` error:
   * Code evaluation from string, e.g. `eval("3 + 5")`, `Function("3 + 5")`, etc
   * Inline `<script>` and `<style>` tags without a correct `nonce` atribute
   * `data:*` URL's in CSS

## Screenshots

The application should work exactly as before


src/client/components/NotificationAlert/index.jsx
```diff
@@ -36,10 +36,7 @@ const NotificationAlert = ({ count, remindersURL }) => (
     href={remindersURL}
     id="notification-bell-count"
   >
-    <StyledImage
-      src={BellSVG}
-      alt="An image of a bell with the notification count overlaid"
```
cgsunkel, line 41: Will removing the alt text introduce an accessibility issue?

tests.env
```diff
@@ -0,0 +1,113 @@
+# This is a sample config file shared between FE and API instances.
+# This file is also used with CircleCI tests.
```
cgsunkel, line 2: We should probably remove this equivalent line from the original `sample.env` file if we won't be using it in CircelCI anymore.

It may be worth replacing it with some text reminding us to update this file when `sample.env` is updated.

cgsunkel (approved):
I have a couple of very minor comments, everything else looks good so will approve

src/templates/_includes/google-tag-manager-snippet.njk
```diff
@@ -1,22 +1,10 @@
...
-window.dataLayer.push({
-  userId: '{{user.id}}',
-  userUserId: '{{user.sso_user_id}}',
-})
-if ('{{userFeatureGTMEvent}}') {
-  window.dataLayer.push({
-    name: '{{userFeatureGTMEvent.name}}',
-    event: '{{userFeatureGTMEvent.event}}',
-  })
-}
```
paulgain, line 13: In the past we were asked to put these in by the GA guys, here at DBT. Now we have tightened the CSP is there another way of doing this instead of just removing them?

peterhudec: Oh, I thought this was part of the copy pasted code snippet. I'm gonna have a look at it.

paulgain: Something like this (not tested):

```js
const { v4: uuid } = require('uuid')

const logger = require('../config/logger')
const config = require('../config')

const STS_MAX_AGE = 180 * 24 * 60 * 60

module.exports = function headers(
  req,
  res,
  next,
  {
    excludedPaths = ['/css', '/javascripts', '/images'],
    nonceGenerator = uuid,
    mode = config.env,
  } = {}
) {
  // Check if the request URL matches any excluded paths
  const isExcluded = excludedPaths.some((path) => req.url.includes(path))

  // Set CSP nonce
  res.locals ||= {}
  res.locals.cspNonce = nonceGenerator()
  const selfAndNonce = `'self' 'nonce-${res.locals.cspNonce}'`

  // Configure CSP policy based on environment
  const scriptSrc =
    mode === 'test'
      ? `${selfAndNonce} 'unsafe-eval' https://*.googletagmanager.com`
      : `${selfAndNonce} https://*.googletagmanager.com`

  // Add GA domain to script source
  const googleAnalyticsSrc = 'https://www.googletagmanager.com'

  // Set Content-Security-Policy header globally
  res.set(
    'Content-Security-Policy',
    [
      `default-src ${selfAndNonce}`,
      `script-src ${scriptSrc} ${googleAnalyticsSrc}`,
      `img-src 'self' https://*.google-analytics.com https://*.googletagmanager.com`,
      `connect-src 'self' https://*.google-analytics.com https://*.analytics.google.com https://*.googletagmanager.com`,
    ].join(';')
  )

  // Add other security headers for non-excluded paths
  if (!isExcluded) {
    logger.http('Headers middleware -> Adding response headers')
    res.set('Cache-Control', 'no-cache, no-store, must-revalidate, private')
    res.set('Pragma', 'no-cache')
    res.set('X-Frame-Options', 'DENY')
    res.set('X-Content-Type-Options', 'nosniff')
    res.set('X-XSS-Protection', '1; mode=block')
    res.set('Strict-Transport-Security', `max-age=${STS_MAX_AGE}`)
  }

  next()
}
```

tests.env
peterhudec: I was not very happy with having a copy of the template either, but for some reason I thought that the values in `.env` override the env vars, but it's actually the other way round so I just added the override of `NODE_ENV` in CI config and removed this file.

src/client/components/NotificationAlert/index.jsx
peterhudec: Fixed it by adding the same text to the [`<title>` and `<desc>` children of the SVG](https://github.com/uktrade/data-hub-frontend/blob/af6fba3b9295f7657ef81f9ff63fdc15c8075431/src/client/components/NotificationAlert/BellIcon.jsx#L22-L27) as suggested [here](https://css-tricks.com/accessible-svgs/#aa-2-inline-svg).

PR #6837, opened 2024-05-24
Restore the reverted CSP policy PR after fixing degraded performance of functional tests

peterhudec:
## Description of change

This PR restores the CSP policy PR (#6760) which has been reverted in #6828 because it tripled the duration of the `functional_tests` CI job. This was caused by [calling `uuid()` in the `headers` middleware](https://github.com/uktrade/data-hub-frontend/blob/a155efbdc7cc8a6d88beaaf09a7a264e3adf92be/src/middleware/headers.js#L15).

This PR solves the problem by [completely unplugging the `headers` middleware](https://github.com/uktrade/data-hub-frontend/blob/a155efbdc7cc8a6d88beaaf09a7a264e3adf92be/src/server.js#L52-L54) when `NODE_ENV=test`. It is safe to do so, because none of the headers—except `Content-Security-Policy`—have any effect on the tests. To allow code coverage, the `Content-Security-Policy` header had to be weakened anyway, which was pretty much equivalent to completely unplugging it.

This PR also fixes some flaky tests which kept failing more often than not.

## Test instructions

* All the requirements of #6760 apply here and
* The `functional_tests` CI job should take around 15 minutes to pass and
* The whole CI pipeline should take around 20 minutes to pass


dredmonds (approved):

PippoRaimondiDIT (approved):
Nice one!

PR #6862, opened 2024-05-31
Chore/dependencies 2024 05 31

elcct:


cgsunkel (approved):

package.json
```diff
@@ -179,7 +179,7 @@
     "chai": "^4.4.1",
     "chai-as-promised": "^7.1.2",
     "chai-subset": "^1.6.0",
-    "cypress": "^13.9.0",
+    "cypress": "^13.10.0",
```
oliverjwroberts, line 182: Just a reminder that you will also need to update the version of Cypress used in `Dockerfile.dependencies`. Steps to do so are in `docs/Dependabot.md` and `docs/Docker.md`.

oliverjwroberts (commented):
Couple of additional changes required for the Cypress version bump, but otherwise looks good to me.

package.json
elcct: It is now updated. Thank you!

paulgain (approved):

oliverjwroberts (approved):

PR #6895, opened 2024-06-07
Deprecate contact activity integration with activity stream

PippoRaimondiDIT:
## Description of change

The scope of this PR is to return an empty contact activity result and deprecate the integration with activity stream for contact activity.

Post migration task will be worked on to consume contact activity from data-flow instead


p3dr0migue1 (approved):

cgsunkel (changes requested):
Instead of just showing nothing we still need to show data on these pages, particularly for the overview page. This is the approach agreed with Rowan and the PM's.

cgsunkel:
Closing as I'm going to re-do this work

PR #6991, opened 2024-07-10
Migration/dpm 130 performance testing playwright

marijnkampf:
## Description of change

## Test instructions

# Performance E2E tests
The performance E2E tests use Playwright. Any current tests should only run non-mutating, non-destructive tests.

In `playwright.config.ts` you can define which environment to run it against by settings the `baseURL`.

To run against your local environment spin it assumes it is running already (`npm run develop`).

If this is the first time running playwright you may need to run `npm install` to ensure you've got everything you need. 

Run tests headless:
`npx playwright test`

Run tests in Playwright Test UI:
`npx playwright test --ui`

## Checklist

- [x] Has the branch been rebased to migration-deploy?


playwright.config.ts
```diff
@@ -0,0 +1,113 @@
...
+  /* Opt out of parallel tests on CI. */
+  workers: process.env.CI ? 1 : undefined,
+  /* Reporter to use. See https://playwright.dev/docs/test-reporters */
+  reporter: 'html',
+  /* Shared settings for all the projects below. See https://playwright.dev/docs/api/class-testoptions. */
+  use: {
+    /* Base URL to use in actions like `await page.goto('/')`. */
+    // baseURL: 'http://127.0.0.1:3000',
+    // baseURL: 'https://dev.datahub.uktrade.digital/',
+    baseURL: 'https://datahub.prod.uktrade.digital',
```
p3dr0migue1, line 31: I would probably put this value in a variable like `prodUrl`, `devUrl`, `stagingUrl` so that you are not always manually editing the string itself.

p3dr0migue1 (approved):

playwright.config.ts
marijnkampf: I've added the most common base URLs to a constant so you can use e.g. `baseURL: BASE_URLS.LOCALHOST`.

peterhudec (commented):
Getting there 😅
There's a couple of things:

- I would not check for the existence of the auth file
- It's not necessary to wait for the requests as most of the [Playwright assertions are _auto retrying_](https://playwright.dev/docs/test-assertions#auto-retrying-assertions), but more importantly, the requests are implementation details and we don't want them to leak into e2e tests.
- I would remove most of the comments and only keep those which explain _why_
- We already have `./test/end-to-end/cypress`, so these could perhaps go to `./test/end-to-end/playwright`?
- I would also hide the `npx playwright test` and `npx playwright test --ui` behind NPM commands

PR #7003, opened 2024-07-17
Add the progressDelay task start option

peterhudec:
## Description of change

This PR adds the `progressDelay` option to starting a _task_, which delays the transition of a _task_ to its _progress_ state by the number of milliseconds the option was set to. The default value is 100ms. This is to prevent progress indicator to be display for a fraction of a seconds for tasks that resolve very quickly or immediately.

## Test instructions

1. Add a task to the `Provider` which resolves in 1 second
    ```js
    const Provider = createProvider({
      foo: () => new Promise(resolve => setTimeout(resolve, 1000, 'RESULT')),
      // ...other tasks
    })
    ```
3. Call the task with `progressDelay` set to both shorter and longer than 1 second
    ```js
    <Task>
      {(t) => (
        <>
          <button
            onClick={() => t('foo', 'shouldSeeProgress').start({
              payload: 123,
              progressDelay: 800,
            })}
          >
            Start with progress delay 800ms
          </button>
          <button
            onClick={() => t('foo', 'shouldNotSeeProgress').start({
              payload: 123,
              progressDelay: 1200,
            })}
          >
            Start with progress delay 1200ms
          </button>
          <pre>{JSON.stringify({
            shouldSeeProgress: t('foo', 'shouldSeeProgress'),
            shouldNotSeeProgress: t('foo', 'shouldNotSeeProgress'),
          }, null, 2)}</pre>
        </>
      )}
    </Task>
    ```
4. The task started with `progressDelay: 800` should be in progress state for 200ms 
5. The task started with `progressDelay: 1200` should never get to progress state

## Screenshots

This should not affect how the application works, except that tasks which resolve in less than 100ms there should be no progress indicator visible.


cgsunkel:

PR #7005, opened 2024-07-18
Fix/api proxy transfer encoding

peterhudec:
## Description of change

Adds the `transfer-encoding: chunked` HTTP header to `/api-proxy` headers. Without this, some endpoints were responding with 403 when deployed to AWS.


cgsunkel (approved):
This needs to go into the `main` branch rather than the `paas` one

p3dr0migue1 (approved):

PR #7188, opened 2024-09-27
Fix/edit interaction styling

peterhudec:
## Description of change

Fixes a misaligned pipeline graphic next to a sub-select in the edit interaction form's Service section.

## Test instructions

1. Go to `/interactions/<id>/edit`
2. Select a value from the select field right below the _Service_ heading (yes, the field doesn't have a label), other than _Other advice and information_
3. There should appear a sub-select (again, no label)
4. The sub field should have an L-shaped pipeline graphic to its left leading from the select above.
5. The pipeline should be nicely aligned so, that it leads to the vertical middle of the sub-select
6. The carret on the very right of the sub-field should stay aligned in the same way as in other selects at all times

### Before

Screenshot 2024-09-27 at 15 35 01

### After
Screenshot 2024-10-01 at 17 22 23


dredmonds (approved):
Looking good to me, and would be better to add an additional  snapshot(screen capture) comparison.

ince-dbt (approved):
LGTM

PR #7359, opened 2024-11-28
Retry tests up to 2 additional times upon failure

oliverjwroberts:
## Description of change

This PR enables retries for any failing tests, up to an additional 2 times. 

The thinking was this will alleviate the challenges we are having with flakey tests while we work on longer term fixes for the offending tests. Rather than needing to re-run the entire test suite upon a failure, Cypress will automatically retry individual tests. 

Retries are recorded in the results, so we will be able to keep visibility of these flakey tests while simultaneously increasing the chance of passing test suites. If the tests continue to fail after 3 attempts, the test suite will fail in the usual fashion. 

For local testing, developers may wish to set this value to 0 while testing new changes.

See the [cypress docs](https://docs.cypress.io/app/guides/test-retries) for more information.

## Checklist

### DIFF
--- a/cypress.config.js
+++ b/cypress.config.js
@@ -27,4 +27,8 @@ module.exports = defineConfig({
     baseUrl: 'http://localhost:3000',
     supportFile: 'test/cypress/support/index.js',
   },
+  retries: {
+    runMode: 2, // number of retries when running `cypress run`
+    openMode: 2, // number of retries when running `cypress open`
+  },
 })


baarkerlounger:
I worry this would help make it easier to ignore the problem rather than fix it properly so would vote against, but don't let that block this if other people are for.

oliverjwroberts:
> I worry this would help make it easier to ignore the problem rather than fix it properly so would vote against, but don't let that block this if other people are for.

@baarkerlounger that was one of my concerns too. I'll give others a chance to comment before merging anything...

cgsunkel (approved):
I'm happy with this, we don't know when the test fix tickets will be picked up (or how long they will take to complete) and this will be helpful in the meantime

samuele-mattiuzzo (approved):

PR #7360, opened 2024-11-29
Company name can be an empty string when generating fake data

DeanElliott96:
## Description of change

The type function expects a value and with the fake json data generator the company name could be an empty string. This change specifies the company name for the test and creates a new test for adding a task without a company name (it's optional).

Example use with the type function (line 118). For the add-task-spec, it tried to type in a company name which was an empty string.
https://github.com/uktrade/data-hub-frontend/blob/91040d6681918e81d93949e540a714df84bc0cf4/test/cypress/support/commands.js#L114-L123

Original Error:
```
Add company task
       When visiting the create a task page
         add task button should send expected values to the api:
     CypressError: `cy.type()` cannot accept an empty string. You need to actually type something.
```

## Checklist

### DIFF
--- a/test/functional/cypress/specs/tasks/add-task-spec.js
+++ b/test/functional/cypress/specs/tasks/add-task-spec.js
@@ -154,11 +154,18 @@ describe('Add company task', () => {
       cy.get('h1').should('have.text', `Add task`)
     })
 
-    it('add task button should send expected values to the api', () => {
+    it('add task button should send expected values with company to the api', () => {
+      company.name = 'a company name'
       cy.intercept(`/api-proxy/v4/company?*`, { results: [company] })
       cy.get('[data-test=assigned-to-me]').click()
       fillTypeahead('[data-test=field-company]', company.name)
       assertTaskForm(null, [myAdviserId], company.id, dashboard.myTasks())
+    })
+
+    it('add task button should send expected values without company to the api', () => {
+      cy.intercept(`/api-proxy/v4/company?*`, { results: [company] })
+      cy.get('[data-test=assigned-to-me]').click()
+      assertTaskForm(null, [myAdviserId], null, dashboard.myTasks())
     })
   })
 })
--- a/test/functional/cypress/specs/tasks/add-task-spec.js
+++ b/test/functional/cypress/specs/tasks/add-task-spec.js
@@ -155,11 +155,16 @@ describe('Add company task', () => {
     })
 
     it('add task button should send expected values with company to the api', () => {
-      company.name = 'a company name'
-      cy.intercept(`/api-proxy/v4/company?*`, { results: [company] })
+      const companyWithName = { ...companyFaker(), name: 'a company name' }
+      cy.intercept(`/api-proxy/v4/company?*`, { results: [companyWithName] })
       cy.get('[data-test=assigned-to-me]').click()
-      fillTypeahead('[data-test=field-company]', company.name)
-      assertTaskForm(null, [myAdviserId], company.id, dashboard.myTasks())
+      fillTypeahead('[data-test=field-company]', companyWithName.name)
+      assertTaskForm(
+        null,
+        [myAdviserId],
+        companyWithName.id,
+        dashboard.myTasks()
+      )
     })
 
     it('add task button should send expected values without company to the api', () => {


paulgain:
@DeanElliott96 This is a generated file and shouldn't be edited manually. Ideally, it should be regenerated frequently to stay aligned with the API. If we make edits now, they will likely be lost the next time the file is generated.

test/functional/cypress/fakers/companies.js
```diff
@@ -18,6 +18,7 @@ import { UNITED_KINGDOM_ID, CANADA_ID } from '../../../../src/common/constants'
 const companyFaker = (overrides = {}) => ({
   ...jsf.generate(apiSchema.components.schemas.Company),
   id: faker.string.uuid(),
+  name: faker.company.name(),
```
peterhudec, line 21: This is a quick fix, but the proper fix should be to remove faker from the test. We should not use anything random in tests. If we do, we are asking for trouble.

DeanElliott96: I was generally under the impression to use fake data but I've been doing some googling and it seems like its not the right way to go about testing.

I'm going to keep looking into it!

But yeah let's get a quick fix in for now and changing faker is a much larger ticket!

paulgain: I wonder if we could keep the fakers but ensure that any randomness introduced is controlled (e.g. by using seeds).

DeanElliott96: Would you prefer if I hard coded a name for the company here or just go with the faker for now?

peterhudec: Another problem with faker is that it generates inconsistent/nonsensical data e.g. the flakiness of the `test/functional/cypress/specs/export-pipeline/add-export-spec.js` is caused by the faker generating a company which has `archived: true` and `archived_on: null` at the same time, which then crashes the FE code which assumes that if `archived == true`then `archived_on` will be a string.

This kind of a dependency can be expressed in some type systems like TypeScript:
```ts
type Archived = {
  archived: true
  archivedOn: string
}

type NotArchived = {
  archived: false
  archivedOn: null
}

type Company = (Archived | NotArchived) & {
  id: string
  name: string
}

const company1: Company = {
//    ^^^^^^^
// Type '{ id: string; name: string; archived: true; archivedOn: null; }' is not assignable to type 'Company'.
//   Types of property 'archivedOn' are incompatible.
//     Type 'null' is not assignable to type 'string'.(2322)
  id: 'foo',
  name: 'bar',
  archived: true,
  archivedOn: null,
}

const company2: Company = {
//    ^^^^^^^
// Type '{ id: string; name: string; archived: false; archivedOn: string; }' is not assignable to type 'Company'.
//   Types of property 'archivedOn' are incompatible.
//     Type 'string' is not assignable to type 'null'.(2322)
  id: 'foo',
  name: 'bar',
  archived: false,
  archivedOn: '2024-11-29',
}
```

And also in OpenSchema
```yaml
components:
  schemas:
    Company:
      allOf:
        -
          type: object
          properties:
            id:
              type: string
            name:
              type: string   
        -
          oneOf:
            - 
              type: object
              properties:
                archived:
                  type: boolean
                  enum: [true]
                archived_on:
                  type: string
            -
              type: object
              properties:
                archived:
                  type: boolean
                  enum: [false]    
```
Which yields this:
Screenshot 2024-11-29 at 15 38 51

But it might be challenging to get DRF to produce such a schema and also `json-schema-faker` [doesn't seem to support this nor plans to](https://github.com/json-schema-faker/json-schema-faker/issues/84).

DeanElliott96: It does look like DRF is deprecating their schema generator in favour of third party packages:
https://www.django-rest-framework.org/topics/documenting-your-api/#built-in-openapi-schema-generation-deprecated

I'll have a look at the suggestions and see what kind of schema they generate.

baarkerlounger: Is this definitely incorrect at the backend data model level? 

i.e. does anything currently prevent us from legitimately saving a Company record without a name? Do we definitely never want to do that?

oliverjwroberts: @baarkerlounger the `Company` model has the `name` field defined as `name = models.CharField(max_length=MAX_LENGTH)`, so we are not allowing null values or empty strings on a database level. I believe the serialiser follows suit.

I think, however, the `apiSchema` in the frontend does not enforce those restrictions?

baarkerlounger: What about these? https://dev.datahub.uktrade.digital/companies?archived[0]=false&sortby=modified_on:desc&page=1

DeanElliott96: CharField can be empty string, docs on the blank and null values here:
https://docs.djangoproject.com/en/5.1/ref/models/fields/#django.db.models.Field.null

DeanElliott96: I'm going to update the test to test with a company name as that's what that test is testing. Then add a new test to test without a company name (it's optional for this task). I will also remove the faker change.

oliverjwroberts: Apologies, I confused myself, `blank=True` is only for validation; you are correct, the DB will still allow it.

test/functional/cypress/specs/tasks/add-task-spec.js
```diff
@@ -154,12 +154,19 @@ describe('Add company task', () => {
       cy.get('h1').should('have.text', `Add task`)
     })
 
-    it('add task button should send expected values to the api', () => {
+    it('add task button should send expected values with company to the api', () => {
+      company.name = 'a company name'
       cy.intercept(`/api-proxy/v4/company?*`, { results: [company] })
       cy.get('[data-test=assigned-to-me]').click()
       fillTypeahead('[data-test=field-company]', company.name)
```
DeanElliott96, line 161: This was the failing line, it calls [type()](https://docs.cypress.io/api/commands/type) which expects a string however as the company name can be empty, it crashes when passed an empty string.

DeanElliott96:
@peterhudec @paulgain @baarkerlounger @oliverjwroberts Ready for review again please 😄

test/functional/cypress/specs/tasks/add-task-spec.js
```diff
@@ -154,12 +154,19 @@ describe('Add company task', () => {
       cy.get('h1').should('have.text', `Add task`)
     })
 
-    it('add task button should send expected values to the api', () => {
+    it('add task button should send expected values with company to the api', () => {
+      company.name = 'a company name'
```
baarkerlounger, line 158: Does this get overwritten by the line below?

test/functional/cypress/specs/tasks/add-task-spec.js
paulgain: I don't think we should be mutating the company object here. It's far better to create a new object with the desired property:
```js
const company = { ...companyFaker(), name: 'a company name' }

```

baarkerlounger: Good point, you don't want the tests to be order dependent

DeanElliott96: Nice catch! I thought the company object was created in the `beforeEach` but its created in the `describe`! Fixed now.

paulgain (approved):

PR #7370, opened 2024-12-02
Fix/manage randomness in tests

peterhudec:
## Description of change

This PR prevents direct import of `faker` and `json-schema-faker` and only allows usage of their seeded, _pseudo-random_ instances from `test/sandbox/utils/random.js` to ensure tests are deterministic.

The only reason why the `utils/random.js` module is located undeer `test/sandbox` is that it's the only place from which it can be imported both from within the _sandbox_ sub-project and the rest of the codebase.

All tests seem to be consistently passing when I rerun the whole pipeline 10 times. In the 3 failed runs, only component tests failed because of a [bug in Cypress component test mode](https://github.com/cypress-io/cypress/issues/28644), which is not in any way related to randomness.
Screenshot 2024-12-04 at 12 14 02

## Test instructions

1. Run a test which makes use of `faker` or `json-schema-faker` (one of the many fakers that we have)
2. No matter how many times you run the test, the "random" value generated by these utilities should be identical between test runs
3. Importing `faker` or `json-schema-faker` directly should be a linter error


paulgain:
@peterhudec looking good. 

There's a small caveat with dates and time which I've outlined below for awareness.

```js
const { faker } = require('@faker-js/faker')

const seed = 12345

// Generate the first sequence
faker.seed(seed)
console.log(faker.string.uuid()) // e52399fa-babf-4014-8acd-fbab757bfac2
console.log(faker.company.name()) // Turner, Bailey and Kulas
console.log(faker.date.past()) // 2024-06-14T17:40:01.175Z

// Reset the seed to repeat the same sequence
faker.seed(seed)
console.log(faker.string.uuid()) // e52399fa-babf-4014-8acd-fbab757bfac2
console.log(faker.company.name()) // Turner, Bailey and Kulas
console.log(faker.date.past()) // 2024-06-14T17:40:01.177Z (notice you won't get the exact time, this is 2 milliseconds greater)
```

paulgain (approved):

peterhudec:
> There's a small caveat with dates and time which I've outlined below for awareness.

God spot @paulgain. I guess there's not much we can do about this, but it is not likely to cause issues as we are not relying on the details of the generated values in tests. In general, it's better to prefer hardcoded values in tests or if we really want to test a function against a (pseudo) random sample of its domain, we should use a _generative testing_ library like [fast-check](https://fast-check.dev/).

cgsunkel (approved):

PR #7300, opened 2024-11-07
Bump date-fns to latest

cgsunkel:
## Description of change

`date-fns` has been upgraded from v2.3.0 to v4.1.0. In order to carry out this upgrade several changes have been made to our date handling functions (mostly relating to whether certain dates need to be parsed before transformation).

For more details around the changes to the package see their [changelog](https://github.com/date-fns/date-fns/blob/main/CHANGELOG.md).

## Test instructions

Pages with dates should load as before.

## Checklist

[//]: # 'When submitting a PR make sure the code review guidelines have been satisfied.
https://github.com/uktrade/data-hub-frontend/blob/main/docs/Code%20review%20guidelines.md'

- [ ] Has the branch been rebased to main?
- [ ] Automated tests (Any of the following when applicable: Unit, Functional or End-to-End)
- [ ] Manual compatibility testing (Browsers: Chrome, Firefox, Edge, Safari)

PR #7372, opened 2024-12-04
Fix export win dates

paulgain:
## Description of change
This PR fixes the dates on export wins following the `date-fns` [upgrade](https://github.com/uktrade/data-hub-frontend/pull/7300).

A new utility file `date-utils.js` has been created containing a single function `formatDate(dateISO, dateISOFormat)`. This function takes a date in either ISO standard format (`2024-12-04`) or a date-time format (`2024-12-04T10:41:45.425717Z`), and a format constant, to return the desired date format. For example:

```js
import { formatDate, DATE_FORMAT_MEDIUM_WITH_TIME,  DATE_FORMAT_COMPACT } from './date-utils'

formatDate('2024-12-04T10:41:45.425717Z', DATE_FORMAT_MEDIUM_WITH_TIME) // 4 Dec 2024, 10:41am
formatDate('2024-12-04', DATE_FORMAT_COMPACT) // 04 Dec 2024
```

This approach improves both readability (each constant has docstrings) and maintainability compared to the previous implementation. I plan to refactor the rest of the codebase to adopt this new method. Specifically, I’ll clean up `dates.js` by removing the outdated formatting functions and replacing them with calls to formatDate. The current state of `dates.js` is convoluted, and this change should make the codebase more manageable going forward. Eventually, we aim to consolidate all date-related functionality into a single, well-organised file.

## Test instructions
Go to: `/exportwins/pending`, `/exportwins/confirmed` and `/exportwins/rejected`.

### Before
pending

rejected

### After (Pending)
pending-after

### After (Rejected)
rejected-after

## Checklist

### DIFF
--- a/src/client/utils/__test__/date-utils.test.js
+++ b/src/client/utils/__test__/date-utils.test.js
@@ -1,0 +1,72 @@
+import {
+  formatDate,
+  DATE_FORMAT_FULL,
+  DATE_FORMAT_FULL_DAY,
+  DATE_FORMAT_COMPACT,
+  DATE_FORMAT_ISO,
+  DATE_FORMAT_MEDIUM,
+  DATE_FORMAT_MEDIUM_WITH_TIME,
+  DATE_FORMAT_YEAR_MONTH,
+  DATE_FORMAT_MONTH_YEAR,
+  DATE_FORMAT_MONTH_ABBR_YEAR,
+  DATE_FORMAT_DAY_MONTH,
+  DATE_FORMAT_INTERACTION_TIMESTAMP,
+} from '../date-utils'
+
+describe('formatDate', () => {
+  const date = '2024-12-04'
+  const time = 'T10:41:45.425717Z'
+  const dateAndTime = `${date}${time}`
+
+  it("should render '04 Dec 2024' (default format)", () => {
+    expect(formatDate(date)).to.equal('04 Dec 2024') // the default
+  })
+
+  it("should render '04 Dec 2024' (DATE_FORMAT_COMPACT)", () => {
+    expect(formatDate(date, DATE_FORMAT_COMPACT)).to.equal('04 Dec 2024')
+  })
+
+  it("should render '4 Dec 2024' (DATE_FORMAT_MEDIUM)", () => {
+    expect(formatDate(dateAndTime, DATE_FORMAT_MEDIUM)).to.equal('4 Dec 2024')
+  })
+
+  it("should render '4 Dec 2024, 10:41am' (DATE_FORMAT_MEDIUM_WITH_TIME)", () => {
+    expect(formatDate(dateAndTime, DATE_FORMAT_MEDIUM_WITH_TIME)).to.equal(
+      '4 Dec 2024, 10:41am'
+    )
+  })
+
+  it("should render '4 December 2024' (DATE_FORMAT_FULL)", () => {
+    expect(formatDate(date, DATE_FORMAT_FULL)).to.equal('4 December 2024')
+  })
+
+  it("should render 'Wed, 04 Dec 2024' (DATE_FORMAT_FULL_DAY)", () => {
+    expect(formatDate(date, DATE_FORMAT_FULL_DAY)).to.equal('Wed, 04 Dec 2024')
+  })
+
+  it("should render '2024-12-04' (DATE_FORMAT_ISO)", () => {
+    expect(formatDate(date, DATE_FORMAT_ISO)).to.equal('2024-12-04')
+  })
+
+  it("should render '2024-12' (DATE_FORMAT_YEAR_MONTH)", () => {
+    expect(formatDate(date, DATE_FORMAT_YEAR_MONTH)).to.equal('2024-12')
+  })
+
+  it("should render 'December 2024' (DATE_FORMAT_MONTH_YEAR)", () => {
+    expect(formatDate(date, DATE_FORMAT_MONTH_YEAR)).to.equal('December 2024')
+  })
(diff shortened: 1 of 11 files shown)


src/client/utils/__test__/date-utils.test.js
```diff
@@ -0,0 +1,92 @@
...
+
+  // Invalid dates
+
+  it('should render "Invalid date: "', () => {
+    expect(formatDate('')).to.equal('Invalid date: ')
+  })
+
+  it('should render "Invalid date: 2024-32-44"', () => {
+    expect(formatDate('2024-32-44')).to.equal('Invalid date: 2024-32-44')
+  })
```
peterhudec, line 91: I would prefer if these cases threw errors. This way the users will see the error messages rendered in place of the date, but not formatted as an error and with no notification for us through Sentry that there was an error.

src/client/utils/date-utils.js
```diff
@@ -0,0 +1,131 @@
...
+ * Day and month format with abbreviated month name.
+ * Example: 04 Dec
+ */
+const DATE_FORMAT_DAY_MONTH = 'dd MMM'
+
+/**
+ * Interaction timestamp format with single-digit day and month.
+ * Example: 2024-12-4
+ */
+const DATE_FORMAT_INTERACTION_TIMESTAMP = 'y-MM-d'
```
peterhudec, line 67: Do we really need all of these? Does it make any UX sense to sometimes use zero paddding and sometimes not? Or to sometimes use ISO format and sometimes the localised format? We seems to be inconsistent in how we render dates even within the context of a small component:
Screenshot 2024-12-05 at 12 25 14

src/client/utils/date-utils.js
```diff
@@ -0,0 +1,131 @@
...
+  let parsedDate
+  try {
+    parsedDate = parseISO(date)
+  } catch (error) {
+    return `Error parsing date: ${date}`
+  }
+
+  if (parsedDate.toString() === 'Invalid Date') {
+    return `Invalid date: ${date}`
+  }
```
peterhudec, line 113: As I mentioned before, I would not catch the `parseISO(date)` call and would throw if `parsedDate` is invalid.

test/component/cypress/specs/ExportWins/WinsConfirmedList.cy.jsx
```diff
@@ -92,7 +92,7 @@ describe('WinsConfirmedList', () => {
             )
           )}`
         )
-      cy.get(items).eq(2).should('have.text', 'Date won: 1 May 2023')
+      cy.get(items).eq(2).should('have.text', 'Date won: May 2023')
```
peterhudec, line 95: Are we removing the day from the date won?

peterhudec (commented):
Looks good, the only issue I see is that we are not throwing an error when `formatDate` is called with ivalid input.

src/client/utils/__test__/date-utils.test.js
paulgain: The issue with throwing errors is that if any dates on the page are invalid, the entire page might fail to load, resulting in a generic global error message accompanied by a stack trace. This approach would not provide a user-friendly inline error message while allowing the rest of the page to load.

test/component/cypress/specs/ExportWins/WinsConfirmedList.cy.jsx
paulgain: This is aligned with the latest design. It seems we need a UX/Design conversation here.

src/client/utils/__test__/date-utils.test.js
paulgain: By throwing an error when encountering an invalid date users would see this which is not ideal.

Screenshot 2024-12-05 at 16 46 47

src/client/utils/date-utils.js
paulgain: If we ignore throwing errors for now, how about:

```js
function formatDate(date, dateFormat = DATE_FORMAT_COMPACT) {
  if (!date) {
    return `Invalid date: ${date}`
  }

  const parsedDate = parseISO(date)

  if (!isValid(parsedDate)) {
    return `Invalid date: ${date}`
  }

  return format(parsedDate, dateFormat)
}
```

paulgain:
@peterhudec, we need to collaborate with the UX to standardise date formatting across Data Hub. I agree there are significant inconsistencies in the codebase, likely due to changes in designers over time. However, this PR won’t address those issues, as resolving them would require a more comprehensive, root-and-branch approach.

src/client/utils/date-utils.js
paulgain: Revised version:

```js
function formatDate(date, dateFormat = DATE_FORMAT_COMPACT) {
  let parsedDate
  try {
    parsedDate = parseISO(date)
  } catch (error) {
    return `Error parsing date: ${date}`
  }

  if (!isValid(parsedDate)) {
    return date === '' ? 'Invalid date: empty string' : `Invalid date: ${date}`
  }

  return format(parsedDate, dateFormat)
}
```

paulgain: We've settled on this revised version 😄. It will throw an error if the date is invalid (unlikely, as these dates come from the API) or if it hasn’t been set correctly in our tests.

```js
function formatDate(dateISO, dateISOFormat = DATE_FORMAT_COMPACT) {
  return format(parseISO(dateISO), dateISOFormat)
}
```

elcct (approved):

peterhudec (approved):

PR #7395, opened 2024-12-12
Remove `formatWithoutParsing` function and replace it with `formatDate`

paulgain:
## Description of change
This PR simplifies the use of date utilities by replacing all instances of `formatWithoutParsing` with `formatDate`. Importantly, it introduces the practice of importing `date-fns` functions directly in source files, bypassing the centralised `date.js` module. This change reduces the reliance on redundant wrappers like those in `date.js`, making the codebase cleaner and more maintainable. In a subsequent PR, the remaining `date.js` wrappers and imports from `date-fns` will be removed entirely.

## Checklist

### DIFF
--- a/src/apps/interactions/apps/details-form/client/tasks.js
+++ b/src/apps/interactions/apps/details-form/client/tasks.js
@@ -29,7 +29,12 @@ import {
   transformExportCountriesToGroupStatus,
 } from '../../../../transformers'
 
-import { formatWithoutParsing } from '../../../../../client/utils/date'
+import {
+  formatDate,
+  DATE_FORMAT_DAY,
+  DATE_FORMAT_MONTH,
+  DATE_FORMAT_YEAR,
+} from '../../../../../client/utils/date-utils'
 
 const { transformValueForAPI } = require('../../../../../client/utils/date')
 
@@ -139,9 +144,9 @@ const transformInteractionToValues = (interaction, companyId, investmentId) => {
       transformObjectToOption(adviser)
     ),
     date: {
-      day: formatWithoutParsing(date, 'dd'),
-      month: formatWithoutParsing(date, 'MM'),
-      year: formatWithoutParsing(date, 'yyyy'),
+      day: formatDate(date, DATE_FORMAT_DAY),
+      month: formatDate(date, DATE_FORMAT_MONTH),
+      year: formatDate(date, DATE_FORMAT_YEAR),
     },
     companies: [companyId],
     investment_project: investmentId,
@@ -237,9 +242,9 @@ export async function getInitialFormValues({
       companies: [companyId],
       investment_project: investmentId,
       date: {
-        day: formatWithoutParsing(date, 'dd'),
-        month: formatWithoutParsing(date, 'MM'),
-        year: formatWithoutParsing(date, 'yyyy'),
+        day: formatDate(date, DATE_FORMAT_DAY),
+        month: formatDate(date, DATE_FORMAT_MONTH),
+        year: formatDate(date, DATE_FORMAT_YEAR),
       },
       contacts:
         referral && referral.contact
--- a/src/config/nunjucks/filters.js
+++ b/src/config/nunjucks/filters.js
@@ -34,16 +34,12 @@ const {
 const { newlineToBr } = require('../../lib/text-formatting')
 const { joinPaths } = require('../../lib/path')
 const {
-  formatWithoutParsing,
-  isUnparsedDateValid,
-  isDateValid,
-  format,
-} = require('../../client/utils/date')
+  formatDate: dateFormatter,
+  DATE_FORMAT_FULL,
+  DATE_FORMAT_MEDIUM_WITH_TIME,
+} = require('../../client/utils/date-utils')
+
 require('numeral/locales/en-gb')
-const {
-  DATE_LONG_FORMAT_1,
-  DATE_TIME_MEDIUM_FORMAT,
-} = require('../../common/constants')
 
 numeral.locale('en-gb')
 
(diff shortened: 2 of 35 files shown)


baarkerlounger (approved):

src/apps/interactions/apps/details-form/client/tasks.js
```diff
@@ -29,7 +29,12 @@ import {
   transformExportCountriesToGroupStatus,
 } from '../../../../transformers'
 
-import { formatWithoutParsing } from '../../../../../client/utils/date'
+import {
+  formatDate,
+  DATE_FORMAT_DAY,
+  DATE_FORMAT_MONTH,
+  DATE_FORMAT_YEAR,
+} from '../../../../../client/utils/date-utils'
```
peterhudec, line 38: I only have realized this when I was actually using the new `formatDate`, how _uergonomic_ if not annoying is to have to import the format constants separately. It makes a dev to move up and down in the module. It would be much more practical to define the constants as an object and constrain the second parameter of `formatDate` to only be a key of the format map.

This can even be typed with TypeScript in JSDoc so that we get the formats suggested by intellisense for a really smooth developer experience. And since we are delegating to the `date-fns` `format`, we can reuse the type of its first argument (`Parameters<format>[0]` which is really `Date | number | string`) for our `date` parameter and that way we will have the type always in sync with the type accepted by the library, should they ever change it.

```js
const { format, parseISO } = require('date-fns')

/**
 * @param {Parameters<format>[0]} date - The date to be formatted
 * @param {keyof typeof PRESETS} formatPreset - The date
 */
const formatDate = (date, formatPreset) =>
  format(date, formatPreset)
```
Screenshot 2024-12-13 at 12 20 20

test/functional/cypress/specs/dashboard/estimated-land-date-spec.js
```diff
@@ -24,13 +24,13 @@ const assertEstimatedLandDate = ({ index, date, countdown, colour }) => {
 
   cy.get('@estimatedLandDate')
     .find('[data-test="estimated-land-date-date"]')
-    .should('have.text', formatWithoutParsing(date, DATE_DAY_LONG_FORMAT))
+    .should('have.text', formatDate(date, DATE_FORMAT_FULL_DAY))
```
peterhudec, line 27: I don't think its the best idea to use `formatDate` in assertions for two reasons:
* Let's say that someone _accidentally_ updated `formatDate` so that it would always return `"foo"` (or some other, less severe bug), then this test would still pass. This is something like a circular proof — if `formatDate` is our only way of formatting then we are pretty much testing the function itself indirectly here and we are making assertion about the function by only using the function itself.
* Ideally, tests should be readable as a prose so that it would be easy for a dev to understand how should the _system under test_ behave. But here the dev doesn't really know what should the `text` look like and if they want to know, they have to understand what `formatDate(date, DATE_FORMAT_FULL_DAY)` does.

I would much prefer hardcoded values where possible in tests. I know that we probably have a lot of tests where the date is dynamic, but the solution is for all the dynamic/random values to be deterministic.

Of course, this is outside of the scope of this PR and doesn't need to be addressed here, but if we can replace the formatting expression for a hardcoded value and the test is still passing, I would do it.

peterhudec (approved):
Looks good, but I left a number of suggestions:

* The ergonomics of `formatDate` could be greatly improved if the format presets were a map and we leveraged TypeScript in JSDoc
* I wouldn't use `formatDate` in test assertions

Approving it anyway as I don't wanna block it while I'm gonna be off for until new year.

src/config/nunjucks/filters.js
```diff
@@ -158,29 +154,11 @@ const filters = {
...
-  formatDateTime: (value, dateFormat = DATE_TIME_MEDIUM_FORMAT) => {
-    if (!value) {
-      return value
-    }
-
-    if (!isDateValid(value)) {
-      return value
-    }
+  formatDate: (value, format = DATE_FORMAT_FULL) =>
+    dateFormatter(new Date(value), format),
```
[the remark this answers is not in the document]
paulgain: Both `formatDate` and `formatDateTime` are Nunjucks filter functions. They are used as shown below:

```
{% macro SubTitle (props) %}
  <h4 class="c-entity__subtitle">
    {% if props.value %}
      {% if props.type === 'date' %}
        {% set value = props.value | formatDate %}
      {% elif props.type === 'dateMonthYear' %}
        {% set value = props.value | formatDate('MMMM yyyy') %}
      {% elif props.type === 'datetime' %}
        {% set value = props.value | formatDateTime %}
      {% else %}
        {% set value = props.value %}
      {% endif %}
      {{ props.label }} {{ value }}
    {% else %}
      {{ props }}
    {% endif %}
  </h4>
{% endmacro %}
```
When using `props.value | formatDate`, the function defaults to a date format string, which is currently `d MMMM yyyy` in `filters.js`. However, the `formatDate` function from `date-utils` defaults to `dd MMM yyyy`. If both default formats were the same, we could simplify as follows:

```js
const { formatDate } = require('../../client/utils/date-utils')

const filters = {
  ...
  formatDate
  ...
}
```

Since the formats differ, I handle it like this instead (note that I've removed the unnecessary alias):

```js
const {
  formatDate,
  DATE_FORMAT_FULL,
  DATE_FORMAT_MEDIUM_WITH_TIME,
} = require('../../client/utils/date-utils')

  formatDate: (value, format = DATE_FORMAT_FULL) => // Maintain the same formatting d MMMM yyyy
    formatDate(new Date(value), format),

  formatDateTime: (value, dateFormat = DATE_FORMAT_MEDIUM_WITH_TIME) =>
    formatDate(value, dateFormat),
```
The priority should be to replace the Nunjucks templates and related code. Once that’s done, `filters.js` can be removed entirely, making this discussion unnecessary.

src/apps/interactions/apps/details-form/client/tasks.js
paulgain: This is a great idea! As you know we would need to create a `tsconfig.config` file for this to work and commit it to Gitub so all devs benefit. This is outside the scope of this PR, but I'm keen to implement this in a future PR.

test/functional/cypress/specs/dashboard/estimated-land-date-spec.js
paulgain: Thank you for the thoughtful feedback! You’re absolutely right that relying on formatDate in assertions risks masking bugs and makes the tests less readable. Using hardcoded values would improve clarity and prevent circular validation.

For this PR, I’ll leave it as is since it’s out of scope, but I’ll keep your suggestion in mind for future tests and refactoring to ensure they’re more deterministic and easier to understand.

PR #7412, opened 2024-12-18
TET-758: Only replace history if params has values

baarkerlounger:
## Description of change

Prior to V5 ` history.replace(url, { data: res.data })` used to strip trailing `?`s. It now doesn't causing a test to fail, this may be related to https://github.com/remix-run/history/issues/701.

This PR checks if we actually have any data in the params object before calling `history.replace()` at all, since that seems in like with the test intentions.

Alternative approaches:
We could adjust the test expectations (there's not much harm in leaving the trailing `?`), or we could strip trailing `?`s from `url` before passing it into `history.replace`.

## Checklist

### DIFF
--- a/assets/javascripts/lib/xhr.js
+++ b/assets/javascripts/lib/xhr.js
@@ -39,7 +39,7 @@ const XHR = {
   updateOutlet(res, params, win = window) {
     this.injectResponseInHtml(res.data)
 
-    if (params) {
+    if (params && Object.keys(params).length > 0) {
       const url = `?${queryString.stringify(params)}`
       try {
         history.replace(url, { data: res.data })
--- a/test/unit-client/assets/javascripts/lib/xhr.test.js
+++ b/test/unit-client/assets/javascripts/lib/xhr.test.js
@@ -40,7 +40,6 @@ describe('XHR', () => {
       const res = { data: {} }
       const params = {}
       XHR.updateOutlet(res, params)
-      // expect(Object.keys(params).length).to.equal(1)
       expect(history.location.search).to.equal('')
     })
     it('should perform page load if unable to pushState', () => {
--- a/test/unit-client/assets/javascripts/lib/xhr.test.js
+++ b/test/unit-client/assets/javascripts/lib/xhr.test.js
@@ -40,7 +40,7 @@ describe('XHR', () => {
       const res = { data: {} }
       const params = {}
       XHR.updateOutlet(res, params)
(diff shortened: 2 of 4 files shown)


santoshdasa12345 (approved):

test/unit-client/assets/javascripts/lib/xhr.test.js
```diff
@@ -40,6 +40,8 @@ describe('XHR', () => {
       const res = { data: {} }
       const params = {}
       XHR.updateOutlet(res, params)
+      if (Object.keys(params).length > 1) {
+      }
```
marijnkampf, line 44: Either there is a expect / assert missing, or these two lines are not needed.

assets/javascripts/lib/xhr.js
```diff
@@ -39,7 +39,7 @@ const XHR = {
   updateOutlet(res, params, win = window) {
     this.injectResponseInHtml(res.data)
 
-    if (params) {
+    if (params && Object.keys(params).length > 1) {
```
baarkerlounger, line 42: This should probably be `> 0` actually?

test/unit-client/assets/javascripts/lib/xhr.test.js
```diff
@@ -40,6 +40,8 @@ describe('XHR', () => {
       const res = { data: {} }
       const params = {}
       XHR.updateOutlet(res, params)
+      if (Object.keys(params).length > 1) {
```
baarkerlounger, line 43: What is this supposed to be doing?

assets/javascripts/lib/xhr.js
santoshdasa12345: resolved

test/unit-client/assets/javascripts/lib/xhr.test.js
santoshdasa12345: Added assertion..resolved

test/unit-client/assets/javascripts/lib/xhr.test.js
```diff
@@ -40,6 +40,7 @@ describe('XHR', () => {
       const res = { data: {} }
       const params = {}
       XHR.updateOutlet(res, params)
+      expect(Object.keys(params).length).to.equal(0)
```
baarkerlounger, line 43: What is this assertion supposed to be testing?

swenban (approved):

PR #7420, opened 2024-12-20
Render OMIS order quote `accepted_by` only if defined

paulgain:
## Description of change
Display the `accepted_by` field for OMIS order quotes only when it contains a valid object, preventing potential issues with null.

## Checklist

### DIFF
--- a/test/functional/cypress/specs/omis/quote-spec.js
+++ b/test/functional/cypress/specs/omis/quote-spec.js
@@ -214,6 +214,13 @@ describe('Order quote', () => {
     })
   })
 
+  context('When the quote has been accepted, but accepted_by is null', () => {
+    it('should not render accepted-by-name', () => {
+      cy.visit(urls.omis.quote(quoteAccepted.id))
+      cy.get('[data-test="accepted-by-name"]').should('not.exist')
+    })
+  })
+
   context('When the order has been cancelled', () => {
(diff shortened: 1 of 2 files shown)


test/functional/cypress/specs/omis/quote-spec.js
```diff
@@ -214,6 +214,13 @@ describe('Order quote', () => {
     })
   })
 
+  context('When the quote has been accepted, but accepted_by is null', () => {
+    it('should not render accepted-by-name', () => {
+      cy.visit(urls.omis.quote(quoteAccepted.id))
+      cy.get('[data-test="accepted-by-name"]').should('not.exist')
+    })
+  })
```
paulgain, line 222: I'm unsure under what conditions this would be true: "When the quote has been accepted, but accepted_by is null". Could this be a BE issue?

baarkerlounger: I'm also not sure if that's a valid case - likely an edge case at best. Possibly if the user the quote was accepted by has since been deleted this would happen?

I think rendering the rest of the component as you're doing here is the best approach though. If it's invalid data that at least makes it easier to surface and have removed.

baarkerlounger (approved):

dredmonds (approved):

PR #7422, opened 2024-12-20
Feature/tet 812 add stova service tag

DeanElliott96:
## Description of change

Add a new event service tag for a new `Stova Event Service` service. Without it, there is an empty tag shown.
Also fixes a console warning for an invalid prop value.

This is required to be merged and released before: https://github.com/uktrade/data-hub-api/pull/5866

## Test instructions

The existing service tags allowing the new (to be released) Stova Event Service to be shown.

### Before

Screenshot 2024-12-20 at 12 23 28

### After

Screenshot 2024-12-20 at 12 23 37

## Checklist

### DIFF
--- a/src/client/modules/Events/transformers.js
+++ b/src/client/modules/Events/transformers.js
@@ -4,6 +4,7 @@ import urls from '../../../lib/urls'
 
 import { getDifferenceInDays, formatStartAndEndDate } from '../../utils/date'
 
+import { TAG_COLOURS } from '../../components/Tag'
 import {
   formatDate,
   DATE_FORMAT_FULL,
@@ -32,7 +33,7 @@ const transformEventToListItem = ({
   if (event_type) {
     tags.push({
       text: event_type.name,
-      colour: 'grey',
+      colour: TAG_COLOURS.GREY,
       dataTest: 'event-kind-label',
     })
   }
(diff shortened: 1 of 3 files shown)


src/client/modules/Events/transformers.js
```diff
@@ -40,7 +40,7 @@ const transformEventToListItem = ({
   if (service) {
     tags.push({
       text: getServiceText(service.name),
-      colour: 'default',
```
DeanElliott96, line 43: There is no default colour so this raises a console warning. It did default to the govBlue colour so I've just specified that one.

src/client/modules/Events/transformers.js
```diff
@@ -40,7 +40,7 @@ const transformEventToListItem = ({
...
+      colour: 'govBlue',
```
paulgain, line 43: Can we use the constants here, something like:
```js
import { TAG_COLOURS } from 'Tag'
...
colour: TAG_COLOURS.GOV_BLUE
```

DeanElliott96: Yes that would be better and would have avoided this warning.

paulgain (approved):

chopkinsmade (approved):

PR #7435, opened 2024-12-23
Add export project interactions list

paulgain:
## Description of change
This PR adds a paginated list of export projects interactions

## Test instructions
Go to: `export/<export-uuid>/interactions`

### Before
Screenshot 2024-12-23 at 12 02 50

### After
Screenshot 2024-12-23 at 12 00 45

## Checklist


src/client/components/Resource/Paginated.js
```diff
@@ -158,7 +158,7 @@ const PaginatedResource = multiInstance({
                       qsParamName={sortByQsParamName}
                     />
                   )}
-                  {result ? children(result.results) : null}
+                  {result ? children(result.results || result) : null}
```
baarkerlounger, line 161: No context but should this maybe be standardised on the backend?

baarkerlounger (approved):
Might be worth adding some tests for the missing branches still?

src/client/components/Resource/Paginated.js
paulgain: I'll speak to Peter in the New Year to get his thoughts, he wrote the component after all. I don't like `result.results` as it's a code smell.

paulgain:
> Might be worth adding some tests for the missing branches still?

This has been thoroughly tested! However, I recommend temporarily disabling Codecov, as the component tests haven’t been instrumented. This means Codecov doesn’t recognise that the lines are being tested, making its results misleading. Chris has made an attempt to address this issue [here](https://github.com/uktrade/data-hub-frontend/pull/7393).

DeanElliott96 (approved):

src/client/components/Resource/Paginated.js
peterhudec: The component assumes that the `result` has the schema of a page of paginated collection of results:

```js
{
  "count": 123,
  "next": "http://api.example.org/accounts/?offset=400&limit=2",
  "previous": "http://api.example.org/accounts/?offset=200&limit=2",
  // Current page
  "results": [
    {}, // Item in current page
    {}, // Item in current page
  ]
}
```
We want the interface of the `Resource.Paginated` component to be such as to pass us the current page into the `children` function:

```js
<Interactions.Paginated
  // ...
>
  {page =>
    // The page should be a list of items on the page,
    // that's the semantics of the Paginated component.
    // Anything else would not make much sense
  }
</Interactions.Paginated>
```
Not sure why we should need the fallback `result.results || result`. When the `result` doesn't have the `.results` property, it most likely is not a result of a paginated API call as the API is sticking to this convention. The fallback would only be helpful if the `result` was an array. Whatever we pass to `children` as an argument should be a list of items on the current page.

The `Interactions` resource component is configured to fetch data from the `/v4/interaction` endpoint which in deed respons with the above schema so in this particular case the fallback should not be necessary.

paulgain: @peterhudec thanks for the detailed response.

From memory I added the fallback in order to get the component tests to run. I may have set the Task up wrong. 

```js

    const interactionList = [
      ...
      interactionFaker(),
      interactionFaker(),
    ]

    cy.mountWithProvider(
      <ExportInteractionsList exportId={exportProject.id} />,
      {
        tasks: {
          Interactions: () => Promise.resolve(interactionList),
          Export: () => Promise.resolve(exportProject),
        },
      }
    )
```

PR #7459, opened 2025-01-08
Refactor date conversion functions and improve the documentation

paulgain:
## Description of change
This refactor removes four functions, adds another and improves the documentation. 

Added functions:
* `isoStringToDateParts`

Removed functions:
* `convertDateToFieldShortDateObject`
* `convertUnparsedDateToFieldShortDateObject`
* `convertDateToFieldDateObject`
* `convertUnparsedDateToFieldDateObject`

## Checklist


src/client/modules/ExportPipeline/__test__/transformers.test.js
```diff
@@ -108,7 +108,11 @@ describe('transformAPIValuesForForm', () => {
         title: 'a',
         owner: undefined,
         team_members: [],
-        estimated_win_date: { month: '', year: '' },
+        estimated_win_date: {
+          year: '',
+          month: '',
+          day: '',
+        },
```
paulgain, line 115: I realise that the estimated win date only has month and year, but we I don't think we need to differentiate between this and a full year. For example, when posting to an endpoint just month and year and a subsequent GET request the response has the day set to `1`.

src/client/modules/Investments/Projects/Details/EditProjectSummaryInitialStep.jsx
```diff
@@ -181,11 +177,7 @@ const EditProjectSummaryInitialStep = ({
         project={project}
       />
       <FieldActualLandDate
-        initialValue={
-          project.actualLandDate
-            ? convertDateToFieldDateObject(project.actualLandDate)
-            : ''
```
DeanElliott96, line 187: These changes and others were returning null or an empty string previously when there was no date. Now `isoStringToDateParts` returns `{ year: '', month: '', day: '' }`.

Is that expected and do these components expect and handle values like this?

DeanElliott96 (commented):
Looks good, just a question on the change in values when the date value is undefined.

src/client/modules/Investments/Projects/Details/EditProjectSummaryInitialStep.jsx
paulgain: @DeanElliott96 good point, thanks for bringing this up. 

Yes, this is the expected behaviour. The `isoStringToDateParts` function consistently returns an object with the structure: `{ year: '', month: '', day: '' }` when no valid date is provided. This approach ensures a consistent interface for the components consuming it, such as the `FieldDate` component.

Most calls to `isoStringToDateParts` set the initial value of `FieldDate` to an object with empty strings, so it aligns with the component's expectations. All the relevant tests are passing, which suggests the components are handling this as intended.

That said, if any issues arise or there's a preference to handle missing dates as `null`, it's straightforward to update the function to return `null` instead of `{ year: '', month: '', day: '' }` for invalid or missing inputs.

So we can go from this:

```js
function isoStringToDateParts(isoString) {
  const emptyDateParts = { year: '', month: '', day: '' }

  if (!isoString) {
    return emptyDateParts
  }

  const date = parseISO(isoString)

  return isValid(date)
    ? {
        year: date.getFullYear(),
        month: date.getMonth() + 1,
        day: date.getDate(),
      }
    : emptyDateParts
}
```

To this:

```js
function isoStringToDateParts(isoString) {
  if (!isoString) {
    return null 
  }

  const date = parseISO(isoString)

  return isValid(date)
    ? {
        year: date.getFullYear(),
        month: date.getMonth() + 1,
        day: date.getDate(),
      }
    : null
}
```

paulgain: @peterhudec it would be good to get your thoughts on this, Dean has raised a good point.

DeanElliott96: I think I would prefer what you have and then it's up to each component whether they want to handle undefined dates first themself (and return what they want as a default) or let the function handle it and handle the return of that function.

But for the existing components as long as they are working as expected then I don't see problems with its current implementation.

DeanElliott96 (approved):

PR #7537, opened 2025-02-12
Remove aventri code from frontend as replaced with Stova Events from Data Hub API

DeanElliott96:
## Description of change

Previously Aventri events and attendees were provided via a microservice. This has been replaced by Stova which is now ingested directly into Data Hub API and served via Data Hub API as well. These ingested events and attendees also behave like Data Hub Events and Data Hub atteendees (contacts). 

On Data Hub API, Stova events and Attendees are converted to Data Hub Events and Contacts which means the frontend does not need to do all the data mangling it was previously doing.

## Test instructions

Everything working as before as the aventri code being removed was not currently being used by the frontend.

## Checklist


src/client/components/ActivityFeed/activities/index.js
```diff
@@ -3,19 +3,13 @@ import CompaniesHouseCompany from './CompaniesHouseCompany'
 import DirectoryFormsApi from './DirectoryFormsApi'
 import MaxemailCampaign from './MaxemailCampaign'
 import HmrcExporter from './HmrcExporter'
-import AventriAttendee from './AventriAttendee'
 import DataHubEvent from './DataHubEvent'
```
cgsunkel, line 6: I think we can get rid of this one as well

src/client/modules/Events/CollectionList/index.jsx
```diff
@@ -238,23 +231,6 @@ const EventsCollection = ({
...
-              id="EventsCollection.aventriId"
-              label={LABELS.aventriId}
-              name="aventri_id"
-              qsParam="aventri_id"
-              hintText="For example, 100100100"
-              type="number"
-              onInput={maxLengthAventriIdValidation}
-              data-test="aventri-id-filter"
-            />
-          </FilterToggleSection>*/}
```
cgsunkel, line 257: I think we should bring this one back rather than deleting it (unless we don't want to allow users to filter by Stova ID?)

cgsunkel (commented):
Just a couple of comments

src/client/modules/Events/CollectionList/index.jsx
cgsunkel: Maybe we need to write a ticket to reinstate this

src/client/components/ActivityFeed/activities/index.js
DeanElliott96: Nice catch, it looks like we can, removed now!

src/client/modules/Events/CollectionList/index.jsx
cgsunkel: We can have a chat about this in tomorrow's standup perhaps?

DeanElliott96: Spoke with Tom and Beth on standup. They are happy to not reintroduce this feature unless asked for it.

cgsunkel (approved):

dredmonds (approved):
LGTM 👍

PR #7567, opened 2025-02-17
CLS2-1246 Adding is_high_value latest audit log entry to EYB search r…

stuart-mindt:
…esults

## Description of change

## Test instructions

_What should I see?_

### Before

_Add a screenshot_

### After

_Add a screenshot_

## Checklist

### DIFF
--- a/src/client/modules/Investments/EYBLeads/transformers.js
+++ b/src/client/modules/Investments/EYBLeads/transformers.js
@@ -14,6 +14,7 @@ export const transformLeadToListItem = ({
   landing_timeframe,
   proposed_investment_region,
   is_high_value,
+  audit_log,
 }) => {
   const tags = [
     {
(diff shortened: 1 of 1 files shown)


src/client/modules/Investments/EYBLeads/transformers.js
```diff
@@ -27,11 +28,49 @@ export const transformLeadToListItem = ({
...
+    )
+
+    return [
+      {
+        label: 'Value modified on',
+        value: formatDate(latestChange.timestamp, DATE_FORMAT_COMPACT),
+      },
+      {
+        label: 'Value change',
+        value: latestChange.changes.is_high_value[0]
```
oliverjwroberts, line 54: My only concern here is what should happen when the value changes to `null`? I know currently we don't have any leads that have `value==null` and I'm not sure its even possible from EYB side of things.

test/api-schema.json
```diff
@@ -9043,6 +9043,10 @@
               "type": "string"
             }
           },
+          "audit_log": {
+            "type": "string",
```
oliverjwroberts, line 9047: I know this was from the auto generated API schema, but should this be an array?

test/functional/cypress/specs/investments/eyb-leads-spec.js
```diff
@@ -245,6 +278,20 @@ describe('EYB leads collection page', () => {
         .eq(4)
         .should('contain', COMPANY_NAME_DEFAULT)
     })
+    it('should display the audit log metadata for each collection item correctly', () => {
```
oliverjwroberts, line 281: Is it worth also having a test to check the correct wording is displayed when the value goes from low to high?

src/client/modules/Investments/EYBLeads/transformers.js
stuart-mindt: Default value for is_high_value is false, never null. Chatted with Olly, happy :)

oliverjwroberts (approved):

PR #7577, opened 2025-02-26
Add initial company name typeahead search using dnb microservice

DeanElliott96:
## Description of change

Initial work from Spike to investigate if we can add a typeahead pointing at the dnb microservice when adding a new company.

## Test instructions

_What should I see?_

### Before

_Add a screenshot_

### After

https://github.com/user-attachments/assets/a32491ee-c061-4114-8990-2de420d48b29

## Checklist


src/client/components/Form/elements/FieldCompanyDnBTypeahead/index.jsx
```diff
@@ -0,0 +1,79 @@
...
+                  getTradingNames(result.dnb_company),
+                  getAddress(result.dnb_company),
+                ]),
+                data: result,
+              }))
+            )
+        } else {
+          return Promise.resolve([])
+        }
+      }, 500)}
```
baarkerlounger, line 64: What is 500 here?

DeanElliott96: It's part of the throttle function which only allows a request every 500ms.
https://lodash.com/docs/#throttle

baarkerlounger: Worth extracting into a named variable?

DeanElliott96: The docs and existing use of throttle don't have a named variable but if whoever is implementing it thinks it helps they could add it: https://uktrade.atlassian.net/browse/TET-993

PR #7672, opened 2025-04-11
The Typeahead now doesn't react on blur

peterhudec:
## Description of change

Fixes a bug when the _company name_ filter on the `/companies` page could be selected by clicking out of the input field which action creates a _filter tag_. Removing the _filter tag_ didn't reset the input field as expected.

The fix is in removing the `onBlur` event from the `Typeahead` component.

## Test instructions

1. Navigate to `/companies`
2. Type something into the _Company name_ field on top of the left hand panel
3. Click outside of the input or hit the TAB key to _blur_ the field
4. Nothing should happen
    - No filter tag should be created
    - The input field value should not be reset

### DIFF
--- a/src/client/components/RoutedInput/index.jsx
+++ b/src/client/components/RoutedInput/index.jsx
@@ -56,9 +56,6 @@ const RoutedInput = ({
           writeQs(e.target.value)
         }
       }}
-      onBlur={(e) => {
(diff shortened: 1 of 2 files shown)


dredmonds (approved):
LGTM

PR #7688, opened 2025-04-16
Remove the react-lines-ellipsis dependency

peterhudec:
## Description of change

This PR removes the [react-lines-ellipsis]() dependency which is not compatible with React 19 and replaces it with our own `LinesEllipsis` CSS based component.

Theres is a slight difference where the previous implementation added the ellipsis on the word boundary and now it's added to the end of the line disregarding words.

## Test instructions

1. Navigate to the page root `/`
2. Click on the _Company lists_ tab if not already selected
3. Look at the _Subject_ column of the table
4. Long cell values in that column should be truncated to maximum of 2 lines and the last line should be truncated with an ellipsis

### Before
Screenshot 2025-04-16 at 14 49 11

### After

Screenshot 2025-04-16 at 14 48 09


baarkerlounger (approved):
Happy with this from a technical perspective. Is the change of word boundary a deliberate design change?

elcct (approved):

peterhudec:
> Happy with this from a technical perspective. Is the change of word boundary a deliberate design change?
@baarkerlounger No, not deliberate, unfortunately currently CSS doesn't provide any way to fine controll where the ellipsis kicks in. The truncation is only used in the companies list and the design decision was made years ago and is probably at odds with the current design philosophy. Also it only kicks in if the company name or the interaction subject are longer than two lines.

I just took the liberty and traded the place where the ellipsis occurs off for removal of an abandoned library 😈

PR #7702, opened 2025-04-23
Make number of new jobs optional for FDI projects with no involvement

peterhudec:
## Description of change

This PR makes the _number of new jobs_ field in the _edit project values_ form optional for all _FDI_ projects with no _level of involvement_.

## Test instructions

1. Create a new _investment project_ with
    * _Investment type_ = _FDI -> [whatever]_ and
    * _Level of investor involvement_ = _No Involvement_
    Or find an existing project with the above properties and grab its ID
2. Navigate to `/investments/projects/{project-id}/edit-value`
4. There should be a _Number of new jobs (optional)_ field
5. You should be able to leave the field empty or put `0` into it as a value
6. Upon submitting the form, there should be no form error related to that field
7. Change the _Level of investor involvement_ of the project to any value other than _No Involvement_
8. Now the field should be required and have a label _Number of new jobs (required)_
9. If you leave the field empty or put `0` into it as a value and submit the form
10. You should see a form error related to the field saying that the field is required or that the value must not be `0`

### DIFF
--- a/src/client/components/Layout/EYBLeadLayout.jsx
+++ b/src/client/components/Layout/EYBLeadLayout.jsx
@@ -31,8 +31,6 @@ const GlobalStyles = createGlobalStyle`
 
 const LocalHeader = styled('div')`
   background-color: ${GREY_4};
-  paddingTop: SPACING.SCALE_5,
-  textAlign: 'center',
 `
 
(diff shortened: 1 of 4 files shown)


dredmonds (approved):
LGTM.

PR #7763, opened 2025-05-19
Chore/dependencies 19 05 25

baarkerlounger:
### Description of change
Weekly dependabot updates

### Checklist

* [x] Has this branch been rebased on top of the current `main` branch?

  <details>
  <summary>Explanation</summary>
  
  The branch should not be stale or have conflicts at the time reviews are requested.
  
  </details>

* [x] Is the CircleCI build passing?


package.json
```diff
@@ -123,7 +123,7 @@
     "react-string-replace": "^1.1.1",
     "react-to-print": "^3.1.0",
     "react-use": "^17.6.0",
-    "redis": "4.7.0",
+    "redis": "5.0.1",
```
oliverjwroberts, line 126: Interesting that the [docs](https://www.npmjs.com/package/redis) mention they don't _officially_ support Redis versions `<7.2`. I know we only just bumped the engine to 7.1 in prod, maybe worth bumping again, if available?

oliverjwroberts (approved):

package.json
baarkerlounger: I think 7.1 is the latest currently supported by [platform-tools] so think that'll need upstream changes first https://github.com/uktrade/platform-tools/blob/main/terraform/elasticache-redis/locals.tf#L3

PR #7765, opened 2025-05-20
Update contact message on view rejected win page

ince-dbt:
## Description of change

This is to reinstate the rejected page AS-IS

## Test instructions

Please go to /exportwins/rejected and select one of the rejected win in the list. You should see UI changes as per below

## Checklist

### DIFF
--- a/src/client/modules/ExportWins/Form/ExportWinForm.jsx
+++ b/src/client/modules/ExportWins/Form/ExportWinForm.jsx
@@ -42,14 +42,20 @@ const StyledParagraph = styled('p')({
   fontWeight: 'bold',
 })
 
-export const ContactLink = ({ sections = [], shouldPluralize = true }) => {
+export const ContactLink = ({
+  sections = [],
+  shouldPluralize = true,
+  hideSectionNames = false,
+}) => {
   const section = shouldPluralize
     ? `${pluralize('section', sections.length)}:`
     : 'section.'
+
   return (
     <>
       Contact <AccessibleLink href={`mailto:${EMAIL}`}>{EMAIL}</AccessibleLink>{' '}
-      if you need to update the {section} {sections.join(', ')}
+      if you need to update the {section}
+      {!hideSectionNames && ` ${sections.join(', ')}`}
     </>
   )
 }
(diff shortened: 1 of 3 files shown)


oliverjwroberts (approved):

src/client/modules/ExportWins/Form/ExportWinForm.jsx
```diff
@@ -42,14 +42,20 @@ const StyledParagraph = styled('p')({
   fontWeight: 'bold',
 })
 
-export const ContactLink = ({ sections = [], shouldPluralize = true }) => {
+export const ContactLink = ({
+  sections = [],
+  shouldPluralize = true,
+  hideSectionNames = false,
+}) => {
   const section = shouldPluralize
```
dredmonds, line 50: The ternary logic operator always gives a value of `true` as `shouldPluralize` set as default into `true` from de-structuring function bracket. It would be good to removed this initialisation as `ContactLink` component has dynamic `props` being called from **Line 352** and **Line 410**.

dredmonds (commented):
I put some comment into ternary logic operator implementation, it might be worth it to review.

src/client/modules/ExportWins/Form/ExportWinForm.jsx
ince-dbt: ContactLink is used in quite a few places beyond lines 352 and 410, and many of them don’t explicitly pass shouldPluralize. If we remove the default true, it’ll be undefined in those cases (which is falsy), and that would cause the message to incorrectly fall back to the singular “section.”

That would unintentionally change the UI in several spots, and since the current wording has already been signed off by the UCD team before, I’d prefer to leave the default as is for now — just to keep things consistent.
Happy to revisit if we later decide to make all usages explicit. Really appreciate the input!

ince-dbt:
> I put some comment into ternary logic operator implementation, it might be worth it to review.

I think this is something we can revisit a bit later on, given the current context. I’ve outlined a bit more detail in my response on your other comment.

baarkerlounger (approved):

